PR #4787, opened 2022-07-26
Benchmark -> benchmark in csv header

zliu41:
It is case sensitive and needs to be lower case.


kwxm (commented):
I was just about to merge this but I see that it has a conflict, probably because of my earlier PR that fixed the problem with the backup file.  That also fixes this problem (thanks to you pointing out the case problem), so maybe we can just close this?

zliu41:
Yep!

PR #4789, opened 2022-07-26
Fix CSV file output problem in cost model benchmarking

kwxm:
This fixes a problem where the CSV output from the cost model benchmarking process was being split into two files.  This is a quick fix: PLT-202 should provide a better solution when we get round to it.


kwxm:
I accidentally added an updated copy of the PLC specification that should have been in another branch. I'll close this and try again.

PR #4790, opened 2022-07-26
Fix CSV file output problem in cost model benchmarking

kwxm:
This fixes a problem where the CSV output from the cost model benchmarking process was being split into two files. This is a quick fix: PLT-202 should provide a better solution when we get round to it.

### DIFF
--- a/plutus-core/cost-model/budgeting-bench/CriterionExtensions.hs
+++ b/plutus-core/cost-model/budgeting-bench/CriterionExtensions.hs
@@ -1,3 +1,4 @@
+-- editorconfig-checker-disable-file
 {-# LANGUAGE LambdaCase #-}
 
 module CriterionExtensions (criterionMainWith, BenchmarkingPhase(..)) where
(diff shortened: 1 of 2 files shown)


plutus-core/cost-model/budgeting-bench/CriterionExtensions.hs
```diff
@@ -23,67 +22,74 @@ import System.FilePath ((<.>))
...
+           case phase of
+              Start -> do
+                 csvExists <- liftIO $ doesFileExist file
+                 if csvExists
+                 then liftIO $ renameFile file (file <.> "backup")
+                 else pure ()
+                 time <- liftIO getCurrentTime
+                 liftIO $ appendFile file $ "# Plutus Core cost model benchmark results\n"
+                 liftIO $ appendFile file $ "# Started at " ++ show time ++ "\n"
+                 writeCsv ("Benchmark","t","t.mean.lb","t.mean.ub","t.sd","t.sd.lb", "t.sd.ub")
```
zliu41, line 65: "Benchmark" should be "benchmark".

zliu41 (approved):
:+1:

PR #4793, opened 2022-07-28
haddock-combine: Include dependencies of hsdocs so all haddock links work

shlevy:
**Please do a proper merge, _not_ a squash/rebase, when merging. The parent commit of this PR is intentional**


shlevy:
Found a bug, temporarily converting to draft.

shlevy:
Working now.

nix/lib/haddock-combine.nix
```diff
@@ -7,14 +7,24 @@
 let
   hsdocs = builtins.map (x: x.doc) (builtins.filter (x: x ? doc) hspkgs);
 in
-runCommand "haddock-join" { buildInputs = [ hsdocs ]; } ''
-  # Merge all the docs from the packages. We don't use symlinkJoin because:
+runCommand "haddock-join"
+{
+  buildInputs = [ hsdocs ];
+  exportReferencesGraph = lib.concatLists
```
michaelpj, line 13: Seems like it needs a comment. What's going on here? Something clever for sure.

michaelpj:
> The parent commit of this PR is intentional

Why? What are you trying to do?

nix/lib/haddock-combine.nix
shlevy: Added

shlevy:
> > The parent commit of this PR is intentional
> 
> Why? What are you trying to do?

I want to be able to use these changes in `marlowe-cardano`, which is on an older version of `plutus`. This will allow us to directly use cb126ff6d3f20985b34cf19ca31fc7f80f3ae550 instead of creating a backport branch with a cherry-pick

shlevy:
> Comment line lengths are too long, see hydra.

I get line too long even if I put the Nix manual URL on its own line, is there any way to disable it for that line?

michaelpj:
Like so: https://github.com/input-output-hk/plutus/blob/master/nix/pkgs/haskell/extra.nix#L73

Or you  could disable it for the file, not a big deal.

PR #4794, opened 2022-07-28
Delay SECP256k1 builtins to Chang

zliu41:
This reverts commit 26fd20a901dba0d49e6427dc0a22b00aa019e3b7.


L-as:
Can someone explain why this keeps being reverted then unreverted and so on?

PR #4800, opened 2022-08-02
PLT-571 Fix failing agda conformance tests regarding returning error terms.

thealmarty:
As discussed(see the updated readme), I'm changing the agda implementation to throw an error instead of returning the `error` term.

- changed the agda implementation to throw an error instead of returning the `error` term.
- removed the fixed tests from the expected failure list.
- updated readme.
- fixed NEAT term level test that was failing because it's now throwing an error instead of returning the `error` term.  The fix is to catch the error and return the `error` term for those tests. It's a bit hacky but it's inline with what those tests do elsewhere. (Thanks @effectfully)

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-conformance/README.md
+++ b/plutus-conformance/README.md
@@ -17,11 +17,11 @@ The tests currently cover or will cover the Haskell and Agda implementation of:
 
 ## Adding/updating test outputs
 
-To update test outputs, use the accept test option of the tests. E.g., to have the test results overwriting the `.expected` files in the Haskell implementation test suite (`haskell-conformance`) , run:
+To update or add test outputs, use the accept test option of the tests. E.g., to have the test results overwriting the `.expected` files in the Haskell implementation test suite (`haskell-conformance`) , run:
 
 `cabal test haskell-conformance --test-options=--accept`
 
-There is also an executable (`add-test-output`) for adding test output with more refined options:
+There is also an executable (`add-test-output`) for adding test output to a specific directory:
 
 E.g., run
 
@@ -39,7 +39,7 @@ For the manual, run:
 
 ## Executable for Haskell implementation
 
-(WIP) `haskell-implementation` is an executable for Haskell implementation CLI testing/usage. 
+(WIP) `haskell-implementation` is an executable for Haskell implementation CLI testing/usage.
 
 ## The Plutus Conformance Test Suite Library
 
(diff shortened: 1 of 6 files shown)


plutus-conformance/README.md
```diff
@@ -39,25 +39,33 @@ For the manual, run:
...
-1. "parse error"
+### The CEK machine
+
+Currently we have tested the conformance of UPLC evaluation against our *Haskell* and *Agda* implementations of the CEK machine. Note that we are not testing conformance of a *reducer*. We are testing an *evaluator*. One noticeable difference between a reducer and an evaluator is that a reducer reduces an ill-defined term to an `error` term while an evaluator has a special *error state*. See section 6.1 of our specification for more details. <!--TODO add link to the spec when it's ready. -->
+
+### Expected outputs
+
+The expected output may contain:
+
+1."parse error"
```
effectfully, line 60: Why remove the space? Here and below.

plutus-metatheory/test/TestNEAT.hs
```diff
@@ -126,11 +126,12 @@ prop_Term tyG tmG = do
   -- turn it into an untyped de Bruijn term
   tmUDB <- withExceptT FVErrorP $ U.deBruijnTerm tmU
   -- reduce the untyped term
-  tmUDB' <- withExceptT (\e -> (Ctrex (CtrexTermEvaluationFail "untyped CEK" tyG tmG))) $ liftEither $ runUAgda tmUDB
+  tmUDB' <- withExceptT (\e -> Ctrex (CtrexTermEvaluationFail "untyped CEK" tyG tmG)) $
+    pure $ either (\_ -> U.Error ()) id $ runUAgda tmUDB
```
effectfully, line 130: Hm... I guess it can simply be 

```
    let tmUDB' = either (\_ -> U.Error ()) id $ runUAgda tmUDB
```

then? I didn't realize during out call that we pretend to handle every kind of error here.

effectfully (approved):

plutus-conformance/README.md
thealmarty: I was getting markdown linter warning. I'm not sure if this is a more correct way of enumerating? What do you think?

plutus-metatheory/test/TestNEAT.hs
thealmarty: actually that's true. Done.

plutus-conformance/README.md
effectfully: No idea. OK, do as you feel correct.

plutus-metatheory/src/Main.lagda
```diff
@@ -557,8 +557,9 @@ runU : TermU → Either ERROR TermU
 runU t = do
```
michaelpj, line 557: Now this `run` function has a different behaviour to all the other `run` functions in this module, right? I'd probably have called it something else, but :shrug:

thealmarty: I will add a comment but I expect we'll change the TCEK too when I'm there so it's just in the mean time.

PR #4805, opened 2022-08-08
PLT-580: Return more useful info from the `isScriptWellFormed` function

bezirg:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
@@ -3,7 +3,7 @@ module PlutusLedgerApi.Common.SerialisedScript
     ( SerialisedScript
     , scriptCBORDecoder
     , ScriptForExecution (..)
-    , isScriptWellFormed
+    , assertScriptWellFormed
     ) where
 
 import PlutusCore
@@ -13,10 +13,12 @@ import UntypedPlutusCore qualified as UPLC
 import Codec.CBOR.Decoding qualified as CBOR
 import Codec.CBOR.Extras
 import Codec.CBOR.Read qualified as CBOR
+import Control.Arrow ((>>>))
+import Control.Monad.Except
+import Data.Bifunctor
 import Data.ByteString.Lazy (fromStrict)
 import Data.ByteString.Short
 import Data.Coerce
-import Data.Either
 import Data.Set as Set
 import Prettyprinter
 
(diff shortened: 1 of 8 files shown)


plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
```diff
@@ -59,6 +61,14 @@ implies that it is (almost certainly) an encoded script and the script does not
...
+assertScriptWellFormed :: MonadError CBOR.DeserialiseFailure m
+                       => LedgerPlutusVersion
+                       -> ProtocolVersion
+                       -> SerialisedScript
+                       -> m ()
+assertScriptWellFormed lv pv = fromShort
+                             >>> fromStrict
+                             >>> CBOR.deserialiseFromBytes (scriptCBORDecoder lv pv)
+                             -- throw away the success result
+                             >>> second (const ())
```
bezirg, line 73: I could refrain from throwing away the result and just return `Right ScriptForExecution`: this would let me call this function inside `mkTermToEvaluate` , thus saving on duplicated code. 
I didn't do it because I was worried that I would be exposing `ScriptForExecution` to the api. I am listening for your opinions

bezirg: Also I didn't go wiht the isomorphic `Maybe DeserialiseFailure` because that's just plain confusing

michaelpj: I agree with your judgements here :+1: but maybe write it down for future reference?

michaelpj (approved):

plutus-ledger-api/src/PlutusLedgerApi/V2.hs
```diff
@@ -111,25 +111,32 @@ module PlutusLedgerApi.V2 (
...
 
-
 import PlutusCore.Data qualified as PLC
 import PlutusTx.AssocMap (Map, fromList)
 
 -- | Check if a 'Script' is "valid" according to a protocol version. At the moment this means "deserialises correctly", which in particular
 -- implies that it is (almost certainly) an encoded script and the script does not mention any builtins unavailable in the given protocol version.
-isScriptWellFormed :: ProtocolVersion -> SerialisedScript -> Bool
-isScriptWellFormed = Common.isScriptWellFormed PlutusV2
+assertScriptWellFormed :: MonadError CBOR.DeserialiseFailure m
```
thealmarty, line 135: does it make sense to name the function `assertDeserialisationSuccess` or something like that?

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
```diff
@@ -59,6 +61,14 @@ implies that it is (almost certainly) an encoded script and the script does not
 
 Note: Parameterized over the ledger-plutus-version since the builtins allowed (during decoding) differs.
 -}
```
thealmarty, line 63: does it make sense to name the function `assertDeserialisationSuccess` or something like that?

thealmarty (approved):

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
bezirg: Maybe, but I don't want to change the name too much because of the existing users of the API

PR #4806, opened 2022-08-08
PLT-556: Copy ADR setup from plutus-apps

zliu41:
Copied from plutus-apps.


michaelpj:
I think we should include the ADR-001 that says we're doing ADRs and gives the format, otherwise there's no reference for it.

zliu41:
@michaelpj updated.

michaelpj (approved):

PR #4810, opened 2022-08-12
add two new function to the V1 and V2 context modules

perturbing:
Hi, I would like to add two functions to both the V1 and V2 context modules. The functions I would like to introduce are similar to the already present `valueLockedBy` and `scriptOutputsAt` function. I want to add something similar for the inputs of a script (so the code is almost copied from those), these are called `scriptInputsAt` and `valueUnLockedBy`.


michaelpj:
We're gradually moving towards having most utility functions be defined in `plutus-script-utils` (https://github.com/input-output-hk/plutus-apps/tree/main/plutus-script-utils) and paring down `plutus-ledger-api` to just the pieces that are strictly used by the ledger. So probably these should be added over there to avoid making the situation worse!

perturbing:
Clear, I will have a look there and close this PR, thanks @michaelpj.

PR #4813, opened 2022-08-18
haddock-combine: Remove index/content link fixup.

shlevy:
These were broken links with the way haddock is folded into the sphinx build.


shlevy:
*Please do a proper merge retaining the parent commit for ease of integration with marlowe-cardano*

shlevy:
It looks like this might make sense when haddock-combined is used directly, rather than as part of the overall sphinx build. Couldn't find any broken links with this in place, and in particular the 2 links called out in the comment are broken currently.

shlevy:
Oh actually on a close read I see that the intent was to point to the *combined* index/contents pages. Currently that's not what happens but it's also not what happens with this PR. If that's important I can try to do it with a relative link.

michaelpj:
Okay! I wrote this a very long time ago, so I trust your analysis. This is 100% hacks :D

michaelpj (approved):

PR #4814, opened 2022-08-18
PLT-712: Update and add Haddock to some functions in `plutus-ledger-api`.

### LINKED ISSUE #4476: Undocumented invariants in types with exposed constructors
### Summary

Multiple types have undocumented invariants, and also expose their constructors, in modules that are not marked as unsafe or internal. The following is a (non-exhaustive) list of examples:

* `PubKey` (in all variants) must be 64 bytes long.
* `POSIXTime` must be non-negative.
* `Signature` must be 64 bytes long.
* `DatumHash` must be 28 bytes long.
* `TokenName` must be no longer than 32 bytes, and be hex encoded.
* `TxId` must be 28 bytes long.
* `ValidatorHash` must be 28 bytes long.
* `PubKeyHash` (in all variants) must 28 bytes long.

1. Attempt to construct any value of any of the listed types which does not follow the specified invariant(s).
2. Observe how the compiler accepts it.
3. Try to use it on-chain.
4. Watch things blow up.

### Actual Result

Your transactions will fail, usually by erroring out.

### Expected Result

If we can construct a value, it should be usable without erroring. If it has internal invariants, those should be documented at least; better still, the constructors should _not_ be exposed.

At minimum, all such invariants must be documented clearly at the definition site of the type, ideally with reasons why they are so. If this is for stuff that is external to the type, links should be provided in the Haddocks to the relevant information. However, in a more ideal world, the constructors of such types should also _either_ inhabit modules clearly marked as unsafe, _or_ not be exposed at all in favour of safe construction methods, such as smart constructors or quasi-quoters.

### System info

Plutus commit: 3f3f875764f2e329c8d738f99cdef0c1260ccde1

### LINKED ISSUE #4801: Plutus Ledger API haddocks incorrectly state that SHA256 is being used.
### Summary

According to a comment from @JaredCorduan [here](https://github.com/input-output-hk/cardano-ledger/issues/2941#issuecomment-1203012829), all hashes in the cardano-ledger should be blake2b. 

Areas where I _know_ that SHA256 is referenced in the haddocks:

- `ScriptHash`
-  `DatumHash`
- `ValidatorHash`
-  `RedeemerHash` 
- `MintingPolicyHash`
-  `StakeValidatorHash`
- `TxId`.

I'm not sure whether all of these _should_ be blake2b, but Jared has confirmed that at least TxId should be.

Look at the source in `src/PlutusLedgerApi/V1/Scripts.hs`, `src/PlutusLedgerApi/V1/Contexts.hs`, `src/PlututsLedgerAPI/V1/Tx.hs`.

### Actual Result

References to `Digest SHA256` and `(...) using a SHA256 hash` appear.

### Expected Result

These should most likely say blake2b, sometimes 224 and sometimes 256.

- Determine which hashes _should_ be blake2b, and which variants.
- `grep -rie "sha256"` and make sure that all of the relevant instances have been updated.

### System info

Latest master (commit 107d7e7ec5a1c033228c402ee221794117566d15)

thealmarty:
The purpose of this is to add/update the Haddock of some functions in the plutus-ledger-api. Closes #4476, closes #4801.

Haddock is added to types mentioned in the issues roughly following this format:

- What is it?
- Its hash algorithm, if any
- No. of bytes, if applicable
- Add warning that these types are simple types without any validation.
- Links to relevant info/spec

I've tried to verify the added info against the spec/code but @michaelpj and @JaredCorduan could you please confirm all the info are accurate. 

@JaredCorduan regarding [your comment](https://github.com/input-output-hk/plutus/issues/4476#issuecomment-1209294094), could you please let us know when the appendix is added so we can add links to the Haddock? Thanks.

I've included most types mentioned in #4476 and #4801. More Haddock is needed and we have follow up tickets to improve our Haddock. But there's demand to merge this part soon.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
@@ -35,7 +35,7 @@ data LedgerBytesError =
     deriving stock (Show)
     deriving anyclass (Exception)
 
-{- | Convert a hex encoded `Bytestring` to a `LedgerBytes`. May return a `LedgerBytesError`. -}
+{- | Convert a hex encoded `Bytestring` to a `LedgerBytes`. May return an error (`LedgerBytesError`). -}
 fromHex :: BS.ByteString -> Either LedgerBytesError LedgerBytes
 fromHex = fmap (LedgerBytes . P.toBuiltin) . asBSLiteral
     where
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
@@ -28,14 +28,14 @@ import PlutusTx
 import PlutusTx.Prelude qualified as P
 import Prettyprinter.Extras (Pretty, PrettyShow (..))
 
-{- | An error that is encountered when converting a `Bytestring` to a `LedgerBytes`. -}
+{- | An error that is encountered when converting a `ByteString` to a `LedgerBytes`. -}
 data LedgerBytesError =
     UnpairedDigit -- ^ Odd number of bytes.
     | NotHexit Char -- ^ Not a hex digit.
     deriving stock (Show)
     deriving anyclass (Exception)
 
-{- | Convert a hex encoded `Bytestring` to a `LedgerBytes`. May return an error (`LedgerBytesError`). -}
+{- | Convert a hex encoded `ByteString` to a `LedgerBytes`. May return an error (`LedgerBytesError`). -}
 fromHex :: BS.ByteString -> Either LedgerBytesError LedgerBytes
 fromHex = fmap (LedgerBytes . P.toBuiltin) . asBSLiteral
     where
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
@@ -28,12 +28,14 @@ import PlutusTx
 import PlutusTx.Prelude qualified as P
 import Prettyprinter.Extras (Pretty, PrettyShow (..))
 
+{- | An error that is encountered when converting a `Bytestring` to a `LedgerBytes`. -}
 data LedgerBytesError =
-    UnpairedDigit
-    | NotHexit Char
+    UnpairedDigit -- ^ Odd number of bytes.
+    | NotHexit Char -- ^ Not a hex digit.
     deriving stock (Show)
     deriving anyclass (Exception)
 
+{- | Convert a hex encoded `Bytestring` to a `LedgerBytes`. May return a `LedgerBytesError`. -}
 fromHex :: BS.ByteString -> Either LedgerBytesError LedgerBytes
 fromHex = fmap (LedgerBytes . P.toBuiltin) . asBSLiteral
     where
@@ -76,12 +78,17 @@ bytes = P.fromBuiltin . getLedgerBytes
 fromBytes :: BS.ByteString -> LedgerBytes
 fromBytes = LedgerBytes . P.toBuiltin
 
+{- | The `IsString` instance of `LedgerBytes` could throw an exception of `LedgerBytesError`. -}
 instance IsString LedgerBytes where
     fromString = unsafeFromEither . fromHex . fromString
 
+{- | The `Show` instance of `LedgerBytes` is its base16/hex encoded bytestring,
+decoded with UTF-8, unpacked to `String`. -}
 instance Show LedgerBytes where
     show = Text.unpack . encodeByteString . bytes
 
+{- | Encode a ByteString value in base16 (i.e. hexadecimal), then
+decode with UTF-8 to a `Text`. -}
 encodeByteString :: BS.ByteString -> Text.Text
 encodeByteString = TE.decodeUtf8 . Base16.encode
 
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
@@ -28,9 +28,12 @@ data StakingCredential
     -- a public key credential (`Crypto.PubKeyHash`) or
     -- a script credential (`Scripts.ValidatorHash`). Both are hashed with /BLAKE2b-244/. 28 byte.
     = StakingHash Credential
-    -- | NB: The fields should really be `Word64` `Natural` `Natural`,
+    -- | NB: The fields should really be all `Word64`, as they are implemented in `Word64`,
     -- but 'Integer' is our only integral type so we need to use it instead.
-    | StakingPtr Integer Integer Integer
+    | StakingPtr
+        Integer -- ^ the slot number
+        Integer -- ^ the transaction index (within the block)
+        Integer -- ^ the certificate index (within the transaction)
     deriving stock (Eq, Ord, Show, Generic)
     deriving anyclass (NFData)
 
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
@@ -24,8 +24,11 @@ import Prettyprinter (Pretty (..), (<+>))
 
 -- | Staking credential used to assign rewards.
 data StakingCredential
+    -- | The staking hash is the `Credential` required to unlock a transaction output. Either
+    -- a public key credential (`Crypto.PubKeyHash`) or
+    -- a script credential (`Scripts.ValidatorHash`). Both are hashed with /BLAKE2b-244/. 28 byte.
     = StakingHash Credential
-    -- | NB: The fields should really be Word64 / Natural / Natural,
+    -- | NB: The fields should really be `Word64` `Natural` `Natural`,
     -- but 'Integer' is our only integral type so we need to use it instead.
     | StakingPtr Integer Integer Integer
     deriving stock (Eq, Ord, Show, Generic)
@@ -44,13 +47,14 @@ instance PlutusTx.Eq StakingCredential where
         PlutusTx.&& c PlutusTx.== c'
     _ == _ = False
 
--- | Credential required to unlock a transaction output
+-- | Credentials required to unlock a transaction output.
 data Credential
   =
     -- | The transaction that spends this output must be signed by the private key.
+    -- Hashed with /BLAKE2b-244/. 28 byte. See `Crypto.PubKeyHash`.
     PubKeyCredential PubKeyHash
-    -- | The transaction that spends this output must include the validator script
-    -- and be accepted by the validator.
+    -- | The transaction that spends this output must include the validator script and
+    -- be accepted by the validator. Hashed with /BLAKE2b-244/. 28 byte. See `Scripts.ValidatorHash`.
   | ScriptCredential ValidatorHash
     deriving stock (Eq, Ord, Show, Generic)
     deriving anyclass (NFData)
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
@@ -28,7 +28,9 @@ data StakingCredential
     -- a public key credential (`Crypto.PubKeyHash`) or
     -- a script credential (`Scripts.ValidatorHash`). Both are hashed with /BLAKE2b-244/. 28 byte.
     = StakingHash Credential
-    -- | NB: The fields should really be all `Word64`, as they are implemented in `Word64`,
+    -- | The certificate pointer, constructed by the given
+    -- slot number, transaction and certificate indices.
+    -- NB: The fields should really be all `Word64`, as they are implemented in `Word64`,
     -- but 'Integer' is our only integral type so we need to use it instead.
     | StakingPtr
         Integer -- ^ the slot number
@@ -54,10 +56,10 @@ instance PlutusTx.Eq StakingCredential where
 data Credential
   =
     -- | The transaction that spends this output must be signed by the private key.
-    -- Hashed with /BLAKE2b-244/. 28 byte. See `Crypto.PubKeyHash`.
+    -- See `Crypto.PubKeyHash`.
     PubKeyCredential PubKeyHash
     -- | The transaction that spends this output must include the validator script and
-    -- be accepted by the validator. Hashed with /BLAKE2b-244/. 28 byte. See `Scripts.ValidatorHash`.
+    -- be accepted by the validator. See `Scripts.ValidatorHash`.
   | ScriptCredential ValidatorHash
     deriving stock (Eq, Ord, Show, Generic)
     deriving anyclass (NFData)
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
@@ -17,15 +17,17 @@ import Control.DeepSeq (NFData)
 import GHC.Generics (Generic)
 import PlutusLedgerApi.V1.Crypto (PubKeyHash)
 import PlutusLedgerApi.V1.Scripts (ValidatorHash)
-import PlutusTx qualified as PlutusTx
+import PlutusTx qualified
 import PlutusTx.Bool qualified as PlutusTx
 import PlutusTx.Eq qualified as PlutusTx
 import Prettyprinter (Pretty (..), (<+>))
 
--- | Staking credential used to assign rewards
+-- | Staking credential used to assign rewards.
 data StakingCredential
     = StakingHash Credential
-    | StakingPtr Integer Integer Integer -- NB: The fields should really be Word64 / Natural / Natural, but 'Integer' is our only integral type so we need to use it instead.
+    -- | NB: The fields should really be Word64 / Natural / Natural,
+    -- but 'Integer' is our only integral type so we need to use it instead.
+    | StakingPtr Integer Integer Integer
     deriving stock (Eq, Ord, Show, Generic)
     deriving anyclass (NFData)
 
@@ -44,8 +46,12 @@ instance PlutusTx.Eq StakingCredential where
 
 -- | Credential required to unlock a transaction output
 data Credential
-  = PubKeyCredential PubKeyHash -- ^ The transaction that spends this output must be signed by the private key
-  | ScriptCredential ValidatorHash -- ^ The transaction that spends this output must include the validator script and be accepted by the validator.
+  =
+    -- | The transaction that spends this output must be signed by the private key.
+    PubKeyCredential PubKeyHash
+    -- | The transaction that spends this output must include the validator script
+    -- and be accepted by the validator.
+  | ScriptCredential ValidatorHash
     deriving stock (Eq, Ord, Show, Generic)
     deriving anyclass (NFData)
 
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Crypto.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Crypto.hs
@@ -18,8 +18,12 @@ import PlutusTx.Lift (makeLift)
 import PlutusTx.Prelude qualified as PlutusTx
 import Prettyprinter
 
--- | The hash of a public key. This is frequently used to identify the public key, rather than the key itself.
--- Should be 28 bytes.
+{- | The hash of a public key. This is frequently used to identify the public key,
+rather than the key itself. Hashed with `BLAKE2b-224`. 28 bytes.
+This is a simple type without any validation, __use with caution__.
+You may want to add checks for its invariants. See the
+ [Shelly ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
+-}
 newtype PubKeyHash = PubKeyHash { getPubKeyHash :: PlutusTx.BuiltinByteString }
     deriving stock (Eq, Ord, Generic)
     deriving anyclass (NFData)
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Crypto.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Crypto.hs
@@ -19,6 +19,7 @@ import PlutusTx.Prelude qualified as PlutusTx
 import Prettyprinter
 
 -- | The hash of a public key. This is frequently used to identify the public key, rather than the key itself.
+-- Should be 28 bytes.
 newtype PubKeyHash = PubKeyHash { getPubKeyHash :: PlutusTx.BuiltinByteString }
     deriving stock (Eq, Ord, Generic)
     deriving anyclass (NFData)
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Crypto.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Crypto.hs
@@ -19,7 +19,8 @@ import PlutusTx.Prelude qualified as PlutusTx
 import Prettyprinter
 
 {- | The hash of a public key. This is frequently used to identify the public key,
-rather than the key itself. Hashed with `BLAKE2b-224`. 28 bytes.
+rather than the key itself. Hashed with /BLAKE2b-224/. 28 bytes.
+
 This is a simple type without any validation, __use with caution__.
 You may want to add checks for its invariants. See the
  [Shelly ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Crypto.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Crypto.hs
@@ -23,7 +23,7 @@ rather than the key itself. Hashed with /BLAKE2b-224/. 28 bytes.
 
 This is a simple type without any validation, __use with caution__.
 You may want to add checks for its invariants. See the
- [Shelly ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
+ [Shelley ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
 -}
 newtype PubKeyHash = PubKeyHash { getPubKeyHash :: PlutusTx.BuiltinByteString }
     deriving stock (Eq, Ord, Generic)
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Crypto.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Crypto.hs
@@ -24,6 +24,9 @@ newtype PubKeyHash = PubKeyHash { getPubKeyHash :: PlutusTx.BuiltinByteString }
     deriving stock (Eq, Ord, Generic)
     deriving anyclass (NFData)
     deriving newtype (PlutusTx.Eq, PlutusTx.Ord, PlutusTx.ToData, PlutusTx.FromData, PlutusTx.UnsafeFromData)
-    deriving IsString via LedgerBytes
-    deriving (Show, Pretty) via LedgerBytes
+    deriving
+        (IsString        -- ^ from hex encoding
+        , Show           -- ^ using hex encoding
+        , Pretty         -- ^ using hex encoding
+        ) via LedgerBytes
 makeLift ''PubKeyHash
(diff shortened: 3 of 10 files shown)


plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
```diff
@@ -17,15 +17,17 @@ import Control.DeepSeq (NFData)
...
-import PlutusTx qualified as PlutusTx
+import PlutusTx qualified
 import PlutusTx.Bool qualified as PlutusTx
 import PlutusTx.Eq qualified as PlutusTx
 import Prettyprinter (Pretty (..), (<+>))
 
--- | Staking credential used to assign rewards
+-- | Staking credential used to assign rewards.
 data StakingCredential
     = StakingHash Credential
```
JaredCorduan, line 27: did you want to mention that this hash is as 28 byte Blake2b 224?

plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
```diff
@@ -44,8 +46,12 @@ instance PlutusTx.Eq StakingCredential where
 
 -- | Credential required to unlock a transaction output
 data Credential
-  = PubKeyCredential PubKeyHash -- ^ The transaction that spends this output must be signed by the private key
-  | ScriptCredential ValidatorHash -- ^ The transaction that spends this output must include the validator script and be accepted by the validator.
+  =
+    -- | The transaction that spends this output must be signed by the private key.
+    PubKeyCredential PubKeyHash
```
JaredCorduan, line 51: did you want to mention that this hash is as 28 byte Blake2b 224?

plutus-ledger-api/src/PlutusLedgerApi/V1/Tx.hs
```diff
@@ -59,32 +59,40 @@ import PlutusLedgerApi.V1.Bytes
 import PlutusLedgerApi.V1.Crypto
 import PlutusLedgerApi.V1.Scripts
 import PlutusLedgerApi.V1.Value
-
--- | A transaction ID, using a SHA256 hash as the transaction id.
+{- | A transaction ID. Hashed with BLAKE2b-244. 28 byte.
```
JaredCorduan, line 62:
```suggestion
{- | A transaction ID. Hashed with BLAKE2b-256. 32 byte.
```

thealmarty:
We normally do a squash and merge but we can also do a rebase and merge for this one to preserve all the commits, if desired.

L-as:
Thank you!

plutus-tx/src/PlutusTx/IsData/Class.hs
```diff
@@ -46,6 +46,7 @@ class UnsafeFromData (a :: Type) where
     --
     -- When implementing this function, make sure to call 'unsafeFromBuiltinData'
     -- rather than 'fromBuiltinData' when converting substructures!
+    -- This is a simple type without any validation, __use with caution__.
```
bezirg, line 49: nitpicking: I don't know if your intentions is to have a newline but you need an empty comment line to have a newline. Right now, it is concatenated with the text above.

plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
```diff
@@ -28,12 +28,14 @@ import PlutusTx
 import PlutusTx.Prelude qualified as P
 import Prettyprinter.Extras (Pretty, PrettyShow (..))
 
+{- | An error that is encountered when converting a `Bytestring` to a `LedgerBytes`. -}
 data LedgerBytesError =
-    UnpairedDigit
-    | NotHexit Char
+    UnpairedDigit -- ^ Odd number of bytes.
+    | NotHexit Char -- ^ Not a hex digit.
```
bezirg, line 34: This will not be shown in haddock output because `LedgerBytesError` is not exported by this module. Which implies we don't have any (regression) tests for the `fromHex` function

plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
```diff
@@ -76,12 +78,17 @@ bytes = P.fromBuiltin . getLedgerBytes
...
 instance IsString LedgerBytes where
     fromString = unsafeFromEither . fromHex . fromString
 
+{- | The `Show` instance of `LedgerBytes` is its base16/hex encoded bytestring,
+decoded with UTF-8, unpacked to `String`. -}
 instance Show LedgerBytes where
     show = Text.unpack . encodeByteString . bytes
 
+{- | Encode a ByteString value in base16 (i.e. hexadecimal), then
+decode with UTF-8 to a `Text`. -}
```
bezirg, line 91: I have a feeling that going through Text and decodeUTF8 is redundant for showing base16 bytestrings.

Also do you know why we prepend '0x' only for tokennames? https://github.com/input-output-hk/plutus/blob/master/plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs#L115-L116

plutus-tx/src/PlutusTx/IsData/Class.hs
thealmarty: I didn't have that intention but that's a good idea.

plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
thealmarty: Maybe it's redundant, since `asBase16` is only used to convert to `String` it seems: https://github.com/input-output-hk/plutus/blob/15b9c416ededcd7184db3c4f741737cadb194966/plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs#L121-L125 Not sure if I should be fixing code in this PR though. I think we should do it in the clean up PR? There's a ticket to clean up `plutus-ledger-api`.

I have no idea why we prepend '0x' only for token names.

plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
thealmarty: Good point. But let's leave this here anyway.

plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
```diff
@@ -17,15 +17,20 @@ import Control.DeepSeq (NFData)
...
 
--- | Staking credential used to assign rewards
+-- | Staking credential used to assign rewards.
 data StakingCredential
+    -- | The staking hash is the `Credential` required to unlock a transaction output. Either
+    -- a public key credential (`Crypto.PubKeyHash`) or
+    -- a script credential (`Scripts.ValidatorHash`). Both are hashed with /BLAKE2b-244/. 28 byte.
     = StakingHash Credential
-    | StakingPtr Integer Integer Integer -- NB: The fields should really be Word64 / Natural / Natural, but 'Integer' is our only integral type so we need to use it instead.
+    -- | NB: The fields should really be `Word64` `Natural` `Natural`,
```
bezirg, line 31: Do you want to say whwat the word64 , natural , natural represent? I think it is the slot number, the transaction index inside that slot and the txout index inside that transaction. Is that correct @JaredCorduan  ?

plutus-ledger-api/src/PlutusLedgerApi/V1/Scripts.hs
```diff
@@ -269,50 +269,104 @@ newtype StakeValidator = StakeValidator { getStakeValidator :: Script }
 instance Haskell.Show StakeValidator where
     show = const "StakeValidator { <script> }"
 
--- | Script runtime representation of a @Digest SHA256@.
+{- | Script runtime representation of a 28 byte hash. Hashed with /BLAKE2b-224/.
+
+This is a simple type without any validation, __use with caution__.
+You may want to add checks for its invariants. See the
+ [Shelly ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
```
bezirg, line 276: typo: Shelley

plutus-ledger-api/src/PlutusLedgerApi/V1/Scripts.hs
```diff
@@ -269,50 +269,104 @@ newtype StakeValidator = StakeValidator { getStakeValidator :: Script }
...
     deriving stock (Generic)
     deriving newtype (Haskell.Eq, Haskell.Ord, Eq, Ord, ToData, FromData, UnsafeFromData)
     deriving anyclass (NFData)
 
--- | Script runtime representation of a @Digest SHA256@.
+{- | Script runtime representation of a 28 byte hash. Hashed with /BLAKE2b-224/.
+
+This is a simple type without any validation, __use with caution__.
+You may want to add checks for its invariants. See the
+ [Shelly ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
```
bezirg, line 293: typo: Shelley

plutus-ledger-api/src/PlutusLedgerApi/V1/Scripts.hs
```diff
@@ -269,50 +269,104 @@ newtype StakeValidator = StakeValidator { getStakeValidator :: Script }
...
     deriving stock (Generic)
     deriving newtype (Haskell.Eq, Haskell.Ord, Eq, Ord, ToData, FromData, UnsafeFromData)
     deriving anyclass (NFData)
 
--- | Script runtime representation of a @Digest SHA256@.
+{- | Script runtime representation of a 32 byte hash. Hashed with /BLAKE2b-256/.
+
+This is a simple type without any validation, __use with caution__.
+You may want to add checks for its invariants. See the
+ [Shelly ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
```
bezirg, line 310: typo

plutus-ledger-api/src/PlutusLedgerApi/V1/Scripts.hs
```diff
@@ -269,50 +269,104 @@ newtype StakeValidator = StakeValidator { getStakeValidator :: Script }
...
```
bezirg, line 327: typo

plutus-ledger-api/src/PlutusLedgerApi/V1/Scripts.hs
```diff
@@ -269,50 +269,104 @@ newtype StakeValidator = StakeValidator { getStakeValidator :: Script }
...
     deriving stock (Generic)
     deriving newtype (Haskell.Eq, Haskell.Ord, Eq, Ord, ToData, FromData, UnsafeFromData)
     deriving anyclass (NFData)
 
--- | Script runtime representation of a @Digest SHA256@.
+{- | Script runtime representation of a 28 byte hash. Hashed with /BLAKE2b-224/.
+
+This is a simple type without any validation, __use with caution__.
+You may want to add checks for its invariants. See the
+ [Shelly ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
```
bezirg, line 344: typo

plutus-ledger-api/src/PlutusLedgerApi/V1/Scripts.hs
```diff
@@ -269,50 +269,104 @@ newtype StakeValidator = StakeValidator { getStakeValidator :: Script }
...
```
bezirg, line 361: typo

plutus-ledger-api/src/PlutusLedgerApi/V1/Tx.hs
```diff
@@ -59,32 +59,41 @@ import PlutusLedgerApi.V1.Bytes
 import PlutusLedgerApi.V1.Crypto
 import PlutusLedgerApi.V1.Scripts
 import PlutusLedgerApi.V1.Value
+{- | A transaction ID. Hashed with BLAKE2b-256. 32 byte.
 
--- | A transaction ID, using a SHA256 hash as the transaction id.
+This is a simple type without any validation, __use with caution__.
+You may want to add checks for its invariants. See the
+ [Shelly ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
```
bezirg, line 66: typo

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -53,36 +53,47 @@ import Prelude qualified as Haskell
...
+import PlutusTx.These (These (..))
+import Prettyprinter (Pretty, (<>))
+import Prettyprinter.Extras (PrettyShow (PrettyShow))
 
+{- | ByteString representing the currency, hashed with /BLAKE2b-224/.
+It is empty for `Ada`, 28 bytes for `PolicyId`.
+
+This is a simple type without any validation, __use with caution__.
+You may want to add checks for its invariants. See the
+ [Shelly ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
```
bezirg, line 78: typo

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -91,18 +102,26 @@ currencyMPSHash (CurrencySymbol h) = MintingPolicyHash h
...
 currencySymbol = CurrencySymbol . PlutusTx.toBuiltin
 
--- | ByteString of a name of a token, shown as UTF-8 string when possible
+{- | ByteString of a name of a token, hashed with /BLAKE2b-256/.
+Shown as UTF-8 string when possible.
+Should be no longer than 32 bytes, empty for Ada.
+
+This is a simple type without any validation, __use with caution__.
+You may want to add checks for its invariants. See the
+ [Shelly ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
```
bezirg, line 111: typo

plutus-ledger-api/src/PlutusLedgerApi/V1/Crypto.hs
```diff
@@ -18,11 +18,20 @@ import PlutusTx.Lift (makeLift)
 import PlutusTx.Prelude qualified as PlutusTx
 import Prettyprinter
 
--- | The hash of a public key. This is frequently used to identify the public key, rather than the key itself.
+{- | The hash of a public key. This is frequently used to identify the public key,
+rather than the key itself. Hashed with /BLAKE2b-224/. 28 bytes.
+
+This is a simple type without any validation, __use with caution__.
+You may want to add checks for its invariants. See the
+ [Shelly ledger specification](https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf).
```
bezirg, line 26: typo

plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
bezirg: Ah no, it was just a general comment, not to be fixed by this PR (if it needs fixing)

plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
bezirg: Actually after reading the shelley spec, the `Word64` may even be wrong, because under "B.4 Payloads" of https://hydra.iohk.io/build/16861845/download/1/ledger-spec.pdf    it says :

> pointer address one 28-byte bytestring, and **three variable-length** unsigned integers

plutus-ledger-api/src/PlutusLedgerApi/V1/Time.hs
```diff
@@ -35,7 +35,8 @@ newtype DiffMilliSeconds = DiffMilliSeconds Integer
 
 makeLift ''DiffMilliSeconds
 
--- | POSIX time is measured as the number of milliseconds since 1970-01-01T00:00:00Z
+-- | POSIX time is measured as the number of milliseconds since 1970-01-01T00:00:00Z.
+-- Must be non-negative.
```
bezirg, line 39: Where does it say that it cannot be negative? Especially since POSIXTime is used as an `Interval` why the lower bound cannot be negative? @michaelpj

bezirg: POSIXTime is seconds in c (int32 or int64) or in Haskell internally (Integer that represents picoseconds)

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
bezirg: What is PolicyId?

bezirg: Is PolicyID same as MintingPolicyHash?

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -53,36 +53,47 @@ import Prelude qualified as Haskell
...
+        , Haskell.Show   -- ^ using hex encoding
+        , Pretty         -- ^ using hex encoding
+        ) via LedgerBytes
     deriving stock (Generic, Data)
     deriving newtype (Haskell.Eq, Haskell.Ord, Eq, Ord, PlutusTx.ToData, PlutusTx.FromData, PlutusTx.UnsafeFromData)
     deriving anyclass (NFData)
 
 {-# INLINABLE mpsSymbol #-}
--- | The currency symbol of a monetary policy hash
+-- | The `CurrencySymbol` of a monetary policy hash
```
bezirg, line 91: Oh yes, answering my own question policyid seems to be the same. Perhaps rename every PolicyId occurence to MintingPolicyHash

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
bezirg: Maybe we should make clear that this has nothing to do with the mintingpolicyhash and is a non-unique arbitrary string attach to the currencysymbol/mintpolicyhash

bezirg: Or does it have to be unique in the AssetClass?

plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
thealmarty: Yes I would like to put in more info if we have it.

plutus-ledger-api/src/PlutusLedgerApi/V1/Time.hs
thealmarty: I think we've confirmed that POSIXTime is in milliseconds. It was listed in one of the issues as an invariant that we should document. I thought it made sense from its existing Haddock because it's definitely after 1970?

plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
JaredCorduan: The triple represents: the slot number, the transaction index (within the block), and the certificate index (within the transaction).

In the spec, "three variable-length unsigned integers" refers to the wire spec, not the haskell types.

The implementation uses a `Word64` for all three of these.

plutus-ledger-api/src/PlutusLedgerApi/V1/Time.hs
bezirg: Yes confimed to be milliseconds. I thought perhaps to leave a note that this is not the same as Haskell's `Data.Time.Clock.POSIX.POSIXTime`

thealmarty: OK I'll remove it since we're not sure.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
thealmarty: An `AssetClass` is identified by a `CurrencySymbol` and a `TokenName` so I think `TokenName` can be anything for an `AssetClass`. A `Value` though, is a map from `CurrencySymbol`'s to a quantity of that currency, which is a map from `TokenName` to an `Integer`. So `CurrencySymbol`'s have to be unique and for each `CurrencySymbol` you need to have unique `TokenName`, otherwise there can be problems. But I think there is only one `TokenName` for each `CurrencySymbol` in an `AssetClass` so that should be fine. So the main thing is that `CurrencySymbol` have to be unique for the map to not lose info.

thealmarty: I've updated the Haddock.

thealmarty: I don't understand why `Value` isn't a map from `AssetClass` to `Integer` though.

bezirg (approved):

plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
```diff
@@ -28,12 +28,14 @@ import PlutusTx
 import PlutusTx.Prelude qualified as P
 import Prettyprinter.Extras (Pretty, PrettyShow (..))
 
+{- | An error that is encountered when converting a `Bytestring` to a `LedgerBytes`. -}
```
michaelpj, line 31:
```suggestion
{- | An error that is encountered when converting a `ByteString` to a `LedgerBytes`. -}
```

plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs
```diff
@@ -28,12 +28,14 @@ import PlutusTx
...
 data LedgerBytesError =
-    UnpairedDigit
-    | NotHexit Char
+    UnpairedDigit -- ^ Odd number of bytes.
+    | NotHexit Char -- ^ Not a hex digit.
     deriving stock (Show)
     deriving anyclass (Exception)
 
+{- | Convert a hex encoded `Bytestring` to a `LedgerBytes`. May return an error (`LedgerBytesError`). -}
```
michaelpj, line 38:
```suggestion
{- | Convert a hex encoded `ByteString` to a `LedgerBytes`. May return an error (`LedgerBytesError`). -}
```

plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
```diff
@@ -17,15 +17,23 @@ import Control.DeepSeq (NFData)
...
 
--- | Staking credential used to assign rewards
+-- | Staking credential used to assign rewards.
 data StakingCredential
+    -- | The staking hash is the `Credential` required to unlock a transaction output. Either
+    -- a public key credential (`Crypto.PubKeyHash`) or
+    -- a script credential (`Scripts.ValidatorHash`). Both are hashed with /BLAKE2b-244/. 28 byte.
     = StakingHash Credential
-    | StakingPtr Integer Integer Integer -- NB: The fields should really be Word64 / Natural / Natural, but 'Integer' is our only integral type so we need to use it instead.
+    -- | NB: The fields should really be all `Word64`, as they are implemented in `Word64`,
```
michaelpj, line 31: Can we include a sentence about what this constructor represents as well as the NB?

plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
```diff
@@ -42,10 +50,15 @@ instance PlutusTx.Eq StakingCredential where
...
     _ == _ = False
 
--- | Credential required to unlock a transaction output
+-- | Credentials required to unlock a transaction output.
 data Credential
-  = PubKeyCredential PubKeyHash -- ^ The transaction that spends this output must be signed by the private key
-  | ScriptCredential ValidatorHash -- ^ The transaction that spends this output must include the validator script and be accepted by the validator.
+  =
+    -- | The transaction that spends this output must be signed by the private key.
+    -- Hashed with /BLAKE2b-244/. 28 byte. See `Crypto.PubKeyHash`.
```
michaelpj, line 57: I don't think we need to repeat this here since it's on the documentation of the inner type.

plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs
```diff
@@ -42,10 +50,15 @@ instance PlutusTx.Eq StakingCredential where
...
+    PubKeyCredential PubKeyHash
+    -- | The transaction that spends this output must include the validator script and
+    -- be accepted by the validator. Hashed with /BLAKE2b-244/. 28 byte. See `Scripts.ValidatorHash`.
```
michaelpj, line 60: ditto

plutus-ledger-api/src/PlutusLedgerApi/V1/Scripts.hs
```diff
@@ -223,50 +223,104 @@ newtype StakeValidator = StakeValidator { getStakeValidator :: Script }
 instance Haskell.Show StakeValidator where
     show = const "StakeValidator { <script> }"
 
--- | Script runtime representation of a @Digest SHA256@.
+{- | Script runtime representation of a 28 byte hash. Hashed with /BLAKE2b-224/.
```
michaelpj, line 226: Perhaps it's closer to say it's the "Type representing a hash of a script"? And then similarly below but for hashes of validators, datums, etc. That's the main purpose of the type, to help you remember what kind of thing it's a hash of.

plutus-ledger-api/src/PlutusLedgerApi/V1/Tx.hs
```diff
@@ -48,32 +48,41 @@ import PlutusLedgerApi.V1.Bytes
 import PlutusLedgerApi.V1.Crypto
 import PlutusLedgerApi.V1.Scripts
 import PlutusLedgerApi.V1.Value
+{- | A transaction ID. Hashed with BLAKE2b-256. 32 byte.
```
michaelpj, line 51:
```suggestion
{- | A transaction ID, i.e. the hash of a transaction. Hashed with BLAKE2b-256. 32 byte.
```

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -53,36 +53,50 @@ import Prelude qualified as Haskell
...
-
+import PlutusTx.These (These (..))
+import Prettyprinter (Pretty, (<>))
+import Prettyprinter.Extras (PrettyShow (PrettyShow))
+
+{- | ByteString representing the currency, hashed with /BLAKE2b-224/.
+It is empty for `Ada`, 28 bytes for `MintingPolicyHash`.
+Forms an `AssetClass` along with `TokenName`.
+A `Value` is a map from `CurrencySymbol`'s to a map from `TokenName` to an `Integer`.
+So `CurrencySymbol` should be unique for `Value` to contain all currencies.
```
michaelpj, line 77: I don't understand this sentence?

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -91,18 +105,27 @@ currencyMPSHash (CurrencySymbol h) = MintingPolicyHash h
 currencySymbol :: BS.ByteString -> CurrencySymbol
 currencySymbol = CurrencySymbol . PlutusTx.toBuiltin
 
--- | ByteString of a name of a token, shown as UTF-8 string when possible
+{- | ByteString of a name of a token, hashed with /BLAKE2b-256/.
```
michaelpj, line 108: Token names aren't hashes! They're just arbitrary bytestrings (with a fixed length).

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -145,21 +170,28 @@ newtype AssetClass = AssetClass { unAssetClass :: (CurrencySymbol, TokenName) }
...
--- 'Value'. The effect of this is that the currencies in the 'Value' are "independent",
--- and are operated on separately.
---
--- Whenever we need to get the quantity of a currency in a 'Value' where there
--- is no explicit quantity of that currency in the 'Value', then the quantity is
--- taken to be zero.
---
--- See note [Currencies] for more details.
+{- | The 'Value' type represents a collection of amounts of different currencies.
+This is a map from 'CurrencySymbol's to a quantity of that currency.
```
michaelpj, line 174: kind of, it's a map from currency symbols to token names to quantities. I wouldn't say it's directly a map to quantities. also I wouldn't describe it as a map ;) it's a finitely-supported function! or the vector space analogy is fine

michaelpj (approved):

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
thealmarty:
```suggestion
```

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
thealmarty: let's not worry about it then.
```suggestion
```

kwxm:
Seems to be stuck in CI, so I'm going to merge anyway.

PR #4819, opened 2022-08-22
Add test coverage in CI

zeme-iohk:


zeme-iohk:
This is too old, reopening.

PR #4821, opened 2022-08-23
PLT-424: PlutusTx.Show instances for builtin types

zliu41:
ByteString and Integer can both be shown without new builtins (as suggested by @michaelpj and @effectfully), so we need no new builtin after all.

some notes:
- We show the base16-encoded ByteStrings. We _can_ replicate the behavior of `Prelude.show @ByteString`, but we'd have to implement all the complex logic in [showLitChar](https://hackage.haskell.org/package/base-4.17.0.0/docs/src/GHC.Show.html#showLitChar), for very little benefit.
- We simply add quotes for Strings, which is different from what `Prelude.show @String` does (as pointed out by @effectfully). We have no builtin that operates on `Char`, so we can't replicate the `Prelude.show` behavior.
- `Data` is shown by encoding it into a ByteString, and then showing the base16-encoded ByteString.


plutus-tx-plugin/test/TH/traceRepeatedly.plc.golden
```diff
@@ -1 +1,5 @@
-[Making my first int, Making my second int, Adding them up]
\ No newline at end of file
+[ Making my first int
+, Making my second int
+, Making my first int
```
zliu41, line 3: @michaelpj @effectfully any idea why this is shown twice? Looks like a bug to me.

plutus-tx/src/PlutusTx/List.hs
```diff
@@ -170,10 +170,9 @@ splitAt n xs
   where
     go :: Integer -> [a] -> ([a], [a])
     go _ []     = ([], [])
-    go 1 (y:ys) = ([y], ys)
-    go m (y:ys) = (y:zs, ws)
-      where
-        (zs, ws) = go (Builtins.subtractInteger m 1) ys
+    go m (y:ys)
+      | m `Builtins.equalsInteger` 1 = ([y], ys)
+      | otherwise = let (zs, ws) = go (Builtins.subtractInteger m 1) ys in (y:zs, ws)
```
effectfully, line 175: What do you think about making it a `case` instead of a `let`? The `let (..., ...) = ... in (..., ...)` pattern is usually used for extra laziness (for example to [tie the knot](https://wiki.haskell.org/Tying_the_Knot)), since matching in `let` is lazy. Plus, a `case` is more straightforward to read, since the reader doesn't need to account for the possibility of a variable appearing on both the LHS and the RHS of a `let`. Total nitpicking.

plutus-tx/src/PlutusTx/Show.hs
```diff
@@ -20,28 +21,78 @@ import PlutusTx.Bool
...
+        toDigits m =
+            let quotient = m `Builtins.quotientInteger` 10
+                remainder = m `Builtins.remainderInteger` 10
+                digit =
+                    if remainder `Builtins.lessThanInteger` 0
+                        then 0 `Builtins.subtractInteger` remainder
+                        else remainder
+             in if quotient `Builtins.equalsInteger` 0
+                    then [digit]
+                    else digit : toDigits quotient
```
effectfully, line 44: Can we have a property test that generates thousands of integers, shows each of them via this function, reads back the result via regular Haskell `Read` and checks that it's the same number?

plutus-tx/src/PlutusTx/Show.hs
```diff
@@ -20,28 +21,78 @@ import PlutusTx.Bool
...
 instance Show Builtins.BuiltinByteString where
-    {-# INLINEABLE show #-}
-    show = Builtins.showByteString
-
-instance Show Builtins.BuiltinData where
-    {-# INLINEABLE show #-}
-    show = Builtins.showData
+    {-# INLINEABLE showsPrec #-}
+    -- Base16-encode the ByteString and show the result.
+    showsPrec _ s = foldl alg id [0 .. Builtins.subtractInteger (Builtins.lengthOfByteString s) 1]
```
effectfully, line 66: Why `foldl` here and `foldr` in `Show Builtins.Integer`? `foldr` is better for function composition.

plutus-tx/src/PlutusTx/Show.hs
```diff
@@ -57,16 +108,20 @@ instance Show () where
...
+    showsPrec _ = showList (showsPrec 0)
+
+{-# INLINEABLE showList #-}
+showList :: forall a. (a -> ShowS) -> [a] -> ShowS
+showList showElem = \case
+    [] -> showString "[]"
+    x : xs ->
+        showString "["
+            . showElem x
+            . foldl alg id xs
```
effectfully, line 120: `foldl` here too.

plutus-tx-plugin/test/TH/traceRepeatedly.plc.golden
effectfully: Seems like the inliner considers `i1` and `i2` safe to inline, @michaelpj is in a better position to answer.

effectfully (approved):
Handling of chars, quoting etc feels rather hacky to me, but since it's not eternalized in a builtin, I think it's all fine.

plutus-tx/src/PlutusTx/Show.hs
zliu41: Sounds good, will do.

michaelpj (approved):
I'd like some tests that do this in PLC, maybe add to the Budget tests so we can also see how expensive this is.

plutus-tx/src/PlutusTx/Show.hs
zliu41: Can you elaborate why `foldr` is better for function composition? I didn't know it is. Is `f . (g . h)` better than `(f . g) . h` somehow?

plutus-tx/src/PlutusTx/List.hs
```diff
@@ -170,10 +170,9 @@ splitAt n xs
   where
     go :: Integer -> [a] -> ([a], [a])
     go _ []     = ([], [])
-    go 1 (y:ys) = ([y], ys)
```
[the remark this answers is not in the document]
zliu41: It's actually interesting because we also use `| n <= 0` in a few functions, and that's fine, but matching on number literals isn't. Any idea why? Is `Eq` more likely to be specialized than `Ord`?

plutus-tx-plugin/test/TH/traceRepeatedly.plc.golden
zliu41: > Because the let-binding is non-strict and it's used twice.

So there's no memoization in PLC evaluation? That's interesting

plutus-tx/src/PlutusTx/List.hs
michaelpj: you can also just add a bang pattern

plutus-tx/src/PlutusTx/Show.hs
michaelpj: I also think just defining the inverses ourselves is maybe a good idea. They might be useful? For testing at least.

plutus-tx-plugin/test/TH/traceRepeatedly.plc.golden
michaelpj: Non-strict bindings are compiled like this:
```
let nonstrict x = y in ... y ...
==>
let strict x = \_ -> y in ... (y ()) ...
```
it's ugly, and I somewhat wish we'd made strict bindings the default. It wouldn't have been that surprising and this behaviour is itself pretty surprising.

effectfully: I've managed to forget that we do it this way. I thought let-expressions were strict by default. That certainly makes it harder to explain to people the semantics of the language.

plutus-tx/src/PlutusTx/List.hs
effectfully: Or that.

plutus-tx/src/PlutusTx/Show.hs
effectfully: > Is `f . (g . h)` better than `(f . g) . h` somehow?

In Haskell, yes, the former reduces to `\x -> f ((g . h) x)`, while the latter to `\x -> (f . g) (h x)` and so the former can compute immediately if `f` doesn't force its argument while the latter has to reduce the inner composition first. Here's a test: the former is [able](https://ideone.com/UDYXdi) to handle an infinite list,  while the latter is [not](https://ideone.com/FWK7in).

Not that it's not only about infinite lists: you may want to get a `String`  lazily to print it as it gets computed, if the `String` is huge or it takes a lot of time to compute it fully right away.

In Plutus there's probably no difference due to the language being strict, though. So I guess `foldl` is equally fine.

I'd still prefer consistent usage of either `foldr` or `foldl`.

effectfully: > So I guess `foldl` is equally fine.

Although it may be slower or faster compared to `foldr`, hard to tell without benchmarking.

plutus-tx-plugin/test/TH/traceRepeatedly.plc.golden
effectfully: Yes, and results in more code bloat, which is one of the biggest problems so far. I'm more worried about semantics though. "By default bindings are lazy in a strict language compiled from a fully lazy language" -- good luck trying to figure out what the heck this all means.

Although if we pretended that Haskell had no influence on Plutus apart from the syntax, it would be equally unclear how to move a piece of code from off-chain to on-chain or vice versa. Maybe it is already unclear, though.

Those strictness concerns do suck.

bezirg: So couldn't we say that our `let`s are call-by-name and not Haskell's call-by-need? The problem is only observable when using `trace` right?

plutus-tx/src/PlutusTx/List.hs
bezirg: I was bitten by let assuming irrefutable patterns / lazy patterns before, so `case` I think is better to reason about evaluation

plutus-tx-plugin/test/TH/traceRepeatedly.plc.golden
michaelpj: We do make some compromises with Haskell's semantics. This is one of them. Maybe it was a bad idea. I thought that Haskellers would be too used to being able to move around let-bindings freely without consequences, which would no longer be true if they were strict by default...

plutus-tx/src/PlutusTx/Show.hs
zliu41: Sounds good, thanks. Changed `foldl` to `foldr`.

PR #4823, opened 2022-08-25
PLT-760: Add PlutusTx.List.fromRange

zliu41:
The plugin can't compile list range syntax, so adding a helper function.

### DIFF
--- a/plutus-tx/src/PlutusTx/List.hs
+++ b/plutus-tx/src/PlutusTx/List.hs
@@ -24,7 +24,8 @@ module PlutusTx.List (
     dropWhile,
     partition,
     sort,
-    sortBy
+    sortBy,
+    fromRange,
     ) where
 
 import PlutusTx.Bool (Bool (..), otherwise, (||))
@@ -258,3 +259,13 @@ sortBy cmp l = mergeAll (sequences l)
       | otherwise       = a:merge as' bs
     merge [] bs         = bs
     merge as []         = as
+
+{-# INLINEABLE fromRange #-}
+-- | @fromRange x y = [x..y]@. This is needed because the Plutus Tx compiler cannot
+-- compile the range syntax.
+fromRange :: Integer -> Integer -> [Integer]
+fromRange start = go []
+  where
+    go acc x
+      | x `Builtins.lessThanInteger` start = acc
+      | otherwise = go (x:acc) (x `Builtins.subtractInteger` 1)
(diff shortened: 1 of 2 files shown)


michaelpj (approved):

plutus-tx/src/PlutusTx/List.hs
```diff
@@ -258,3 +259,13 @@ sortBy cmp l = mergeAll (sequences l)
...
+
+{-# INLINEABLE fromRange #-}
+-- | @fromRange x y = [x..y]@. This is needed because the Plutus Tx compiler cannot
+-- compile the range syntax.
+fromRange :: Integer -> Integer -> [Integer]
+fromRange start = go []
+  where
+    go acc x
+      | x `Builtins.lessThanInteger` start = acc
+      | otherwise = go (x:acc) (x `Builtins.subtractInteger` 1)
```
effectfully, line 271: Do we have a policy on using accumulators vs non-tail-recursion? I see that `merge` above is non-tail-recursive, while this one uses an accumulator to achieve tail-recursion as opposed to doing something like

```
    go x
      | x `Builtins.greaterThanInteger` end = []
      | otherwise = x : go (x `Builtins.addInteger` 1)
```

Also should we use the standard `enumFromTo` name?

Is it beneficial for stuff in PlutusTx Prelude to be "Haskell-friendly"? Like not using an accumulator here would be "Haskell-friendly", because it would make the function productive (in the corecursion sense).

michaelpj: No clear policy. I think tail-recursion is still good, although it would be nice to demonstrate that with a test.

zliu41: @effectfully all very good points.

I tested the non-tail-recursive version and the generated PLC actually costs less cpu (tr: 11819821618, ntr: 11819745428), and it's better for Haskell so I'll update it.

> Also should we use the standard enumFromTo name?

Integer isn't enum but I do like `fromTo` better than `fromRange`.

effectfully: > I tested the non-tail-recursive version and the generated PLC actually costs less cpu (tr: 11819821618, ntr: 11819745428), and it's better for Haskell so I'll update it.

Thank you! I was always curious, but not to the point of getting bothered to check.

> Integer isn't enum but I do like `fromTo` better than `fromRange`.

In `base` it is:

Screenshot from 2022-08-25 18-30-33

zliu41: > In base it is:

Ah yes, of course, I confused it with Bounded

PR #4824, opened 2022-08-25
PLT-424: Add a budget test for PlutusTx.Show

zliu41:
Also optimized the `Show Integer` instance as @michaelpj suggested (if negative, take the inverse first), which is about 20% more efficient because it no longer checks the sign of each digit.


michaelpj (approved):
I was also thinking it would be nice to just have two tests for e.g. a simple function in a tail-recursive and non-tail recursive version.

zliu41:
> a simple function in a tail-recursive and non-tail recursive version.

Yeah, good idea.

effectfully (approved):

zliu41:
Sneaking in a commit (1f292c7) that changes some long names to symbols.

plutus-tx/src/PlutusTx/List.hs
```diff
@@ -172,7 +172,7 @@ splitAt n xs
     go :: Integer -> [a] -> ([a], [a])
     go _ []     = ([], [])
     go m (y:ys)
-      | m `Builtins.equalsInteger` 1 = ([y], ys)
+      | m == 1 = ([y], ys)
       | otherwise = case go (Builtins.subtractInteger m 1) ys of
```
bezirg, line 176: i think this could also be made `(m - 1)`

zliu41: This module can't import `PlutusTx.Numeric` because it would cause circular dependency.

PR #4830, opened 2022-08-29
PLT-724: Remove transaction inputs from plutus-ledger-api

zliu41:
`TxIn` is not part of `ScriptContext` and is not used by the ledger. If Plutus Tools needs it, it should be moved to plutus-apps.


koslambrou (approved):

koslambrou:
Some of these are already moved to plutus-apps.

michaelpj (approved):
Good.

bezirg (commented):
Yes looks good

PR #4831, opened 2022-08-29
Josfaj/plt 758 remove refs to secp256k1sig builtins

joseph-fajen:
This edit is limited to removing references to the two Secp256k1signature built-ins that were pulled out of the Vasil HF.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -22,8 +22,8 @@ if rtd_version not in ["stable", "latest"]:
 
 # -- Project information -----------------------------------------------------
 
-project = 'Plutus'
-copyright = '2021, IOHK'
+project = 'Plutus Programming and Reference Guide'
+copyright = '2022, IOHK'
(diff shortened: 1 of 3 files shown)


joseph-fajen:
I made a PR for PLT-758 that I meant to use only for deleting two references  to Secp256k1Signature built-ins that were pulled out of the Vasil HF from this page (https://plutus.readthedocs.io/en/latest/reference/cardano/language-changes.html). However, after committing and looking at the PR, I see that it also included my draft 2 of the updated landing page/intro for Plutus Core readthedocs. I don't really see that as a problem, but it means that draft 2 may get pushed live before intended. Although I think that's ok, even though I expect to make more revisions after some more discussion about it.

michaelpj (approved):
Removals LGTM, I'd somewhat prefer to merge this with just the intended changes.

PR #4837, opened 2022-09-01
Kwxm/parser/minor fixes

kwxm:
This is PLT-778.  The `uplc` parser was producing error messages like this:
```
uplc: ParseErrorB (ParseErrorBundle {bundleErrors = TrivialError 30 (Just (Tokens ('x' :| "+y)))\n"))) (fromList [Tokens ('b' :| "uiltin"),Tokens ('c' :| "on"),Tokens ('d' :| "elay"),Tokens ('e' :| "rror"),Tokens ('f' :| "orce"),Tokens ('l' :| "am")]) :| [], bundlePosState = PosState {pstateInput = "(program 1.2.3\n(lam x (lam y (x+y)))\n)", pstateOffset = 0, pstateSourcePos = SourcePos {sourceName = "test", sourceLine = Pos 1, sourceColumn = Pos 1}, pstateTabWidth = Pos 8, pstateLinePrefix = ""}})
```
That was just because of a missing call to the prettyprinter, and a one-line fix turns it into this
```
uplc: test:2:16:
  |
2 | (lam x (lam y (x+y)))
  |                ^^^^^^^
unexpected "x+y)))<newline>"
expecting "builtin", "con", "delay", "error", "force", or "lam"
```
which s a lot more readable. 

However, note that at the top it says `uplc: test:2:16:`.  That's because the parser parameter that tells it where the input's coming from was set to a default, so I also modified the code to get the file path in there.  The majority of the changes are for that and involve threading a filename (or some other identifier) through various functions.

### DIFF
--- a/plutus-core/executables/Common.hs
+++ b/plutus-core/executables/Common.hs
@@ -72,8 +72,7 @@ class Executable p where
 
   -- | Parse a program.
   parseProgram ::
-    T.Text ->
-      Either ParserErrorBundle (p PLC.SourcePos)
+    String -> T.Text -> Either ParserErrorBundle (p PLC.SourcePos)
 
   -- | Check a program for unique names.
   -- Throws a @UniqueError@ when not all names are unique.
@@ -92,7 +91,7 @@ class Executable p where
 
 -- | Instance for PLC program.
 instance Executable PlcProg where
-  parseProgram = PLC.runQuoteT . PLC.parseProgram
+  parseProgram inputName = PLC.runQuoteT . PLC.parseProgram inputName
   checkProgram = PLC.checkProgram
   serialiseProgramFlat nameType p =
       case nameType of
@@ -103,7 +102,7 @@ instance Executable PlcProg where
 
 -- | Instance for UPLC program.
 instance Executable UplcProg where
-  parseProgram = PLC.runQuoteT . UPLC.parseProgram
+  parseProgram inputName = PLC.runQuoteT . UPLC.parseProgram inputName
   checkProgram = UPLC.checkProgram
   serialiseProgramFlat nameType p =
       case nameType of
@@ -218,6 +217,10 @@ instance PrintBudgetState Cek.RestrictingSt where
 ---------------- Types for commands and arguments ----------------
 
 data Input       = FileInput FilePath | StdInput
+instance Show Input where
+    show (FileInput path) = show path
+    show StdInput         = "<stdin>"
+
 data Output      = FileOutput FilePath | StdOutput
 data TimingMode  = NoTiming | Timing Integer deriving stock (Eq)  -- Report program execution time?
 data CekModel    = Default | Unit   -- Which cost model should we use for CEK machine steps?
@@ -282,10 +285,10 @@ parseInput ::
 parseInput inp = do
     contents <- getInput inp
     -- parse the UPLC program
-    case parseProgram contents of
+    case parseProgram (show inp) contents of
       -- when fail, pretty print the parse errors.
       Left (err :: ParserErrorBundle) ->
-        errorWithoutStackTrace $ show err
+          errorWithoutStackTrace $ PP.render $ pretty err
       -- otherwise,
       Right p -> do
         -- run @rename@ through the program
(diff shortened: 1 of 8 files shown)


plutus-core/executables/Common.hs
```diff
@@ -282,10 +286,10 @@ parseInput ::
 parseInput inp = do
     contents <- getInput inp
     -- parse the UPLC program
-    case parseProgram contents of
+    case parseProgram (show inp) contents of
       -- when fail, pretty print the parse errors.
       Left (err :: ParserErrorBundle) ->
-        errorWithoutStackTrace $ show err
+          errorWithoutStackTrace $ PP.render $ pretty err
```
kwxm, line 292: This is the fix for the main problem.

plutus-core/plutus-core/test/Spec.hs
```diff
@@ -191,7 +191,7 @@ propParser = property $ do
                 (\p -> fmap (TextualProgram . void) (parseProg p))
     where
         parseProg :: T.Text -> Either ParserErrorBundle (Program TyName Name DefaultUni DefaultFun SourcePos)
-        parseProg p = runQuoteT $ parseProgram p
+        parseProg = runQuoteT . parseProgram "test"
```
kwxm, line 194: We don't actually have a file name at this point, so I just got it to do what it was doing before.

thealmarty: My thought was that `parseProgram` is for when we don't use the file name. Otherwise, you should use `parse`. Perhaps we can rename the old `parseProgram` to `parseProgramWithoutName`?

plutus-core/executables/Common.hs
thealmarty: I've pushed a commit that uses `errorBundlePretty` instead. It may be better for in case we do configure how we pretty print the parse errors later. It's up to you whether you take the commit or not, I'm fine either way. You can also use `display` here if you want to do this way.

plutus-core/plutus-core/test/Spec.hs
kwxm: Yes, maybe that's a good idea.  Once I started doing this it turned out that `parseProgram` was used in more places than I'd expected, so maybe exporting both a generic one and one that takes a name too is the right way to go.  I'll look at that tomorrow.

michaelpj: or just pipe it up further? it's easy enough for the caller to choose a name

thealmarty: there are cases that the name just isn't used/needed. For cases where the name is useful, one should use `parse`.

plutus-core/executables/Common.hs
kwxm: Ah, thanks.  I get lost in all these different ways of printing things.  I just copied what it said a few lines below, but that's for a completely different kind of error.

plutus-core/plutus-core/test/Spec.hs
kwxm: @thealmarty
> Perhaps we can rename the old parseProgram to parseProgramWithoutName?
I tried that but it was kind of annoying because we still have `parseTerm` and `parseType` which do similar things and it was inconsistent with them.  I've put `parseProgaram` back to what it was before and changed the name of  the method in `Excutable` to `parseNamedProgram` instead and got it to use `pare program`.  That's the only place where we use the name of the input, so that seems reasonably sensible.

plutus-core/plutus-core/src/PlutusCore/Parser.hs
```diff
@@ -78,10 +79,14 @@ term = choice $ map try
     , varTerm
     ]
 
--- | Parse a PLC program. The resulting program will have fresh names. The underlying monad must be capable
--- of handling any parse errors.
-parseProgram :: (AsParserErrorBundle e, MonadError e m, MonadQuote m) =>
-    Text -> m (Program TyName Name DefaultUni DefaultFun SourcePos)
+-- | Parse a PLC program. The resulting program will have fresh names. The
+-- underlying monad must be capable of handling any parse errors.  This passes
+-- "test" to the parser as the name of the input stream; to supply a name
```
thealmarty, line 84: thanks! That clarifies it well!

thealmarty (approved):

kwxm:
This has been stuck in CI for a few days.  I've run all the relevant tests manually and everything seems to be OK, so I'm going to merge it regardless.

PR #4838, opened 2022-09-01
Trivial change to get a branch for experimentation

kwxm:
I just want to experiment with the benchmarking machine, so ignore this.


kwxm:
That eventually gives this
Nix-error
(followed by many more lines of stack trace).

If I log in to the machine and run nix-shell then `cabal bench validation` works without any problems (apart from a couple of occurrences of "the impossible happened" which went away after repeating the command).

kwxm:
Don't need this any more.

PR #4839, opened 2022-09-01
Temporarily suppress broken haddock-join derivation

zeme-iohk:


doc/default.nix
```diff
@@ -23,6 +23,12 @@ stdenv.mkDerivation {
     recommonmark
   ];
   buildPhase = ''
+    # FIXME
+    # https://input-output.atlassian.net/browse/PLT-789
+    # https://hydra.iohk.io/build/18701775/nixlog/1
+    echo TODO > $out
+    exit 0
+
     cp -aR ${combined-haddock}/share/doc haddock
```
michaelpj, line 32: I think you could just comment out this line, no? That would prevent it from being built

michaelpj: So the problematic derivation just wouldn't be required.

zeme-iohk: I'd have to comment out lines 32 to 35, as they use the `haddock` folder.
But I can definitely leave `haddock-combine.nix` alone. Updated

PR #4841, opened 2022-09-01
PLT-64: plutus-conformance: tests for all error cases of evaluator

bezirg:
Adds more tests that exercice different cases the machines can fail.
I created one uplc example for every different kind of cek machine error or cek user error.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-conformance/test-cases/uplc/evaluation/failure/cons-bs-arg-overflow/cons-bs-arg-overflow.uplc
```diff
@@ -0,0 +1,2 @@
+-- the arg overflow'ed over the maxBound :: Word8
+(program 1.0.0 [(builtin consByteString) (con integer 256) (con bytestring #)])
```
michaelpj, line 2: I think we have a specific builtins folder for behaviour of specific builtins, which I think this is?

michaelpj: and some more below

plutus-conformance/test-cases/uplc/evaluation/failure/head/head.uplc
```diff
@@ -0,0 +1 @@
+(program 0.0.0 [(force (builtin headList)) [(builtin mkNilData) (con unit ())]])
```
michaelpj, line 1: comment?

michaelpj (commented):
This looks like good coverage of builtin errors, are we covering all the machine stuff? e.g. do we have applying a non-function to something?

plutus-conformance/test-cases/uplc/evaluation/failure/cons-bs-arg-overflow/cons-bs-arg-overflow.uplc
thealmarty: so we have a `builtins` folder and a `failure` folder. Not sure if it's a sensible organization? But if it is, then failures should all go in the `failure` folders.

michaelpj: I'm not sure that "failure" is such an interesting categorization. I think people are more likely to be interested in a class of behaviour, and then they're going to be interested in both successful and failing behaviour. so my money goes for "just put them all in the builtins folder"

bezirg:
> This looks like good coverage of builtin errors, are we covering all the machine stuff? e.g. do we have applying a non-function to something?

argExpected.uplc

plutus-conformance/test-cases/uplc/evaluation/failure/cons-bs-arg-overflow/cons-bs-arg-overflow.uplc
thealmarty: Sounds good to me.

bezirg:
@thealmarty can you check one last time?

thealmarty:
`consByteString` still failing due to Agda version as you pointed out. We need to fix it or add it to the expected failure list for now.

Could you clarify a bit more what we have tested on `verifyEcdsaSecp256k1Signature`? I think you included some of this info in the file names but did we test all of:
- invalid key (including key too long or too short or just invalid or just empty)
- invalid key with edge case messages (empty or just spaces or weird things like that)

bezirg:
> `consByteString` still failing due to Agda version as you pointed out. We need to fix it or add it to the expected failure list for now.
Thanks for that i fixed it i think.

> Could you clarify a bit more what we have tested on `verifyEcdsaSecp256k1Signature`? I think you included some of this info in the file names but did we test all of:
> 
> * invalid key (including key too long or too short or just invalid or just empty)
> * invalid key with edge case messages (empty or just spaces or weird things like that)

I only tested the `verify*Signature` builtins for things that can lead to (evaluation failure). For this to happen, the msg/key/sig has to have some format error. I did not add tests that exercise if the cek succees but the verification of the signature returns `False`.

bezirg:
> > `consByteString` still failing due to Agda version as you pointed out. We need to fix it or add it to the expected failure list for now.
> 
> Thanks for that i fixed it i think.
> 
> > Could you clarify a bit more what we have tested on `verifyEcdsaSecp256k1Signature`? I think you included some of this info in the file names but did we test all of:
> > 
> > * invalid key (including key too long or too short or just invalid or just empty)
> > * invalid key with edge case messages (empty or just spaces or weird things like that)
> 
> I only tested the `verify*Signature` builtins for things that can lead to (evaluation failure). For this to happen, the msg/key/sig has to have some format error. I did not add tests that exercise if the cek succeeds with the result of verification of the signature returning `False`.

In other words, I did not test for "invalid key with edge case messages (empty or just spaces or weird things like that)"

thealmarty (approved):

PR #4842, opened 2022-09-01
PLT-724: Remove Script type

zliu41:
It can be moved to plutus-apps if needed.

Added `serialiseUPLC`, `deserialiseUPLC` and `serialiseScript` functions for Flat serialisation.


plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
```diff
@@ -55,6 +60,38 @@ we only perform it in V2 and above.
 -- | Scripts to the ledger are serialised bytestrings.
 type SerialisedScript = ShortByteString
 
+{-| Note [Using Flat for serialsing/deserialising Script]
```
bezirg, line 63: typo serialising

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
```diff
@@ -55,6 +60,38 @@ we only perform it in V2 and above.
...
+format otherwise, we have defined the `serialiseUPLC` and `deserialiseUPLC` functions.
+
+Because Flat is not self-describing and it gets used in the encoding of Programs,
+data structures that include scripts (for example, transactions) no-longer benefit
+for CBOR's ability to self-describe it's format.
+-}
+
+serialiseUPLC :: UPLC.Program UPLC.DeBruijn DefaultUni DefaultFun () -> SerialisedScript
+serialiseUPLC =
+    -- See Note [Using Flat for serialsing/deserialising Script]
```
bezirg, line 78: ditto

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
```diff
@@ -55,6 +60,38 @@ we only perform it in V2 and above.
...
+    -- Currently, this is off because the old implementation didn't actually work, so we need to be careful
+    -- about introducing a working version
+    toShort . BSL.toStrict . serialise . SerialiseViaFlat
+
+deserialiseUPLC :: SerialisedScript -> UPLC.Program UPLC.DeBruijn DefaultUni DefaultFun ()
+deserialiseUPLC = unSerialiseViaFlat . deserialise . BSL.fromStrict . fromShort
+  where
+    unSerialiseViaFlat (SerialiseViaFlat a) = a
```
bezirg, line 86: I guess also `coerce @(SerialiseViaFlat _)` could be used here instead, although that wouldn't be very explicit

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
```diff
@@ -55,6 +60,38 @@ we only perform it in V2 and above.
...
+
+serialiseScript :: forall a. CompiledCode a -> SerialisedScript
```
bezirg, line 88: the explicit forall can be omitted here I think

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
```diff
@@ -55,6 +60,38 @@ we only perform it in V2 and above.
...
+serialiseScript = serialiseUPLC . toNameless . getPlc
+    where
+        toNameless :: UPLC.Program UPLC.NamedDeBruijn DefaultUni DefaultFun ()
```
bezirg, line 91: the `toNameless` traversal could be optimized away either by

1) coercing the program to Program FakeNamedDeBruijn which would break though the invariant that name=fakeName
2) write a newtype wrapper ThrowAwayNamedDeBrujin with deriving Flat via FakeNameDeBruijn and coerce to that

but maybe all these are not worth it because `serialiseScript` is run offline

doc/tutorials/BasicValidators.hs
```diff
@@ -82,16 +83,16 @@ validatePayment _ _ ctx =
...
 -- The serialized forms can be written or read using normal Haskell IO functionality.
 showSerialised :: IO ()
-showSerialised = print serializedDateValidator
+showSerialised = print serialisedDateValidator
 -- BLOCK6
 -- We can serialize 'CompiledCode' also
-serializedCompiledCode :: BS.ByteString
-serializedCompiledCode = Flat.flat $ alwaysSucceedsCompiled
+serialisedCompiledCode :: BS.ByteString
+serialisedCompiledCode = Flat.flat alwaysSucceedsCompiled
```
bezirg, line 95: This tutorial is confusing. We mention CBOR forSerialisedScript and DateValidator, but we use flat for serialisedCompiledCode and loadFromFile

We should say that the on-chain serialized format is in CBOR but the off-chain format for locally storing and executing plutus code is in flat.

Alternatively, we can cleanup the confusion between CompiledCode (Flat) and Script (CBOR). From a user's perspective, it is confusing to explain that there is CompiledCode but also on-chain Script and they both have different serialization formats; in the end they are both plutus scripts.

I propose to have the same serialization format for both off-chain and on-chain scripts (CBOR), but differentiate slightly between offline `DebugScript` and on-chain `Script`.

```
type DebugScript = Program NamedDeBruijn ... (serialized via CBOR-enveloped Flat)
type Script = Program DeBruijn (serialised via CBOR-enveloped Flat)
```

what do you think?

michaelpj: I think we should use CBOR everywhere and keep flat as an implementation detail.

zliu41: Yeah nobody outside of our team should need to know anything about Flat, or what CompiledCode is. All they need to know is `serialiseScript` (I probably should name it `serialiseCompiledCode`) serialises CompiledCode into CBOR.

I do agree that the tutorial is confusing. It shouldn't use Flat. In fact I don't even know whether this tutorial is useful at all. It's a pretty old tutorial and we should decide if we should keep it or come up with a better one as part of Joseph's work of re-organizing the readthedocs site.

As to the type aliases: there's already `SerialisedScript` that wraps `Program NamedDeBruijn`. We could have a type/newtype that wraps `Program DeBruijn`, but the ledger doesn't need it, and so I'd expect it to be defined in Plutus Tools.

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
zliu41: I see, so basically fusing the traversal with the flat serialisation.

It's a bit complicated, and this is not performance critical code as you said, so I'd leave it as it is.

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
zliu41: I prefer having explicit foralls. Happy to change it for consistent reasons, but we currently don't seem to have consistency on this.

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
zliu41: Yep.

effectfully (commented):
I've no idea what's going on here and why, so I'm not going to leave a review. Sorry!

kwxm:
Sorry, I'm off at the moment!

bezirg (approved):

L-as:
What's the point in not having an alias or newtype for `UPLC.Program UPLC.DeBruijn DefaultUni DefaultFun ()`? This is heavily inconvenient for downstream users and a huge breaking change.

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
L-as: Please export this function.

zliu41:
> What's the point in not having an alias or newtype

This package is the ledger API, not a script utils library, and so is not meant to offer anything not needed by cardano-ledger. There's a `plutus-script-utils` package in plutus-apps, which looks like a good place for such things as the `Script` newtype to live in.

zliu41:
> What about plutus-tx-plugin?

The `Script` newtype wraps a UPLC program, and is not specific to Plutus Tx. People who write their scripts using Plutarch or other tools may also want to use it. So I wouldn't put it in plutus-tx-plugin either.

L-as:
@zliu41 I mean, I don't think cardano-ledger uses plutus-tx-plugin.

zliu41:
> I mean, I don't think cardano-ledger uses plutus-tx-plugin.

Ah, yes. If you mean there are plutus-tx things in plutus-ledger-api that cardano-ledger doesn't need, you are quite right. Their dependency should be reversed - plutus-tx should depend on plutus-ledger-api. I've been wanting to do that for a while.

PR #4856, opened 2022-09-14
[Metatheory] [Builtins] Add support for 'chooseUnit'

effectfully:
This adds support for `chooseUnit` to the metatheory and a template for adding support for all other missing builtins. I've restructured `BUILTIN` to make it exhaustively match on builtins, so that we it's clear which builtins are handled and which are not.


effectfully:
@thealmarty note that adding support for `mkNilData` and `mkNilPairData` will require adding the type of lists to `Utils.TermCon`, i.e. it's much less trivial than adding support for `chooseUnit` not requiring any additions to `TermCon`.

thealmarty (approved):
Thanks!

PR #4862, opened 2022-09-20
Replace `PyF` with `string-interpolate`

zeme-iohk:
`PyF >= 0.11` is broken in a very weird way: https://github.com/guibou/PyF/issues/115
A fix seems unlikely to be found soon.
And `PyF <= 0.10` (current version) is broken on `ghc >= 9.2`
In order to allow the 9.2 migration to continue, we replace `PyF` with `string-interpolate`.
This until we either patch `PyF` or we decide that `string-interpolate` is good enough.

Differences between `PyF` and `string-interpolate` 
- `PyF` uses `{var}` syntax, `string-interpolate` uses `#{var}` syntax
- `PyF` QQ is called `fmt`, `string-interpolate` is called `__i` (not the prettiest) 
- `string-interpolate` does not have formatting options like `PyF`, it just uses the show instance.


michaelpj:
You saw the link to the 1`PyF` issue, right? https://github.com/guibou/PyF/issues/111

I didn't try to minimize it, but that might be a good next step.

zeme-iohk:
I was unable to replicate the bug, so I hoped that adding the 0.11 constraint Would Just Work :( 
What did you test it on? Which ghc/PyF versions? cabal repl or ghci?
I'm testing on ghc8107 and PyF 0.11.0.0 with cabal repl and it works fine... yet the nix build fails.

michaelpj:
I could replicate it with `cabal build` in the `nix-shell`, I think? I'll check.

michaelpj:
Yep, can reproduce.

michaelpj:
`cabal build plutus-tx-plugin`

zeme-iohk:
@zliu41 I've been informed you were the proponent of PyF.
PyF does have a more powerful and slimmer API, but again this is blocking the ghc9.2 upgrade.
Are you happy replacing it with `string-interpolate`?

michaelpj (approved):
I'm fine with this if Ziyang is!

guibou:
Hello, PyF author here.

I'm sorry to learn about your problem and I would like to find a solution suitable for your needs.

- I can fix PyF `<= 0.10` for ghc 9.2 (and 9.4), it should be easy.
- I can also, as proposed in https://github.com/guibou/PyF/issues/115, avoid the `unsafeRunTcm` trick, or at least set a flag to disable it. Actually, I'm using this "trick" in two contexts, one for error reporting with correct location, and one to detect if a variable is in scope, which was the problem observed in https://github.com/guibou/PyF/issues/111.

Do you have a reproduction of the issue which may help me try and see what solution would be the best?

zeme-iohk:
Hey @guibou thank you for reaching out.

Fixing PyF 0.10 for ghc >=9.2 would be great!

We can reproduce the error only when using PyF 0.11 as a build dependency using the patched GHC coming from [`haskell-nix`](https://github.com/input-output-hk/haskell.nix) on linux (macSO works fine it seems).

To reproduce you could enter the `nix-shell` (any revision behind the current master, which now uses `string-interpolate`) after setting PyF to ==0.11.0.0 in the cabal files,  and running e.g. `cabal build plutus-tx-plugin`.

But that is not at all a minimal reproducible example...

guibou:
Would you mind testing https://github.com/guibou/PyF/pull/117 which move the "dangerous" function under the cabal flag `-unsafe_error_reporting`.

You may want to use PyF with this flag set to `True` when developing locally (i.e. no cross compilation, better error reporting) and set the flag to `False` in your CI.

guibou:
> But that is not at all a minimal reproducible example...

That's minimal enough, I'll try that, thank you ;)

guibou:
https://hackage.haskell.org/package/PyF-0.11.1.0/changelog fixs the build in your codebase.

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationEvent.hs
```diff
@@ -143,17 +144,15 @@ data TestFailure
 renderTestFailure :: TestFailure -> String
 renderTestFailure = \case
     InvalidResult err -> display err
-    MissingCostParametersFor ver ->
-        [fmt|
-Missing cost parameters for {show ver}. Report this as a bug
-against the script dumper in plutus-apps."
-|]
+    MissingCostParametersFor ver -> [__i|
+        Missing cost parameters for #{show ver}.
```
bezirg, line 148: i think calling show is redundant here, unless the extra double-quotes around the version are needed.

plutus-tx-plugin/src/PlutusTx/Options.hs
```diff
@@ -88,27 +89,25 @@ newtype ParseErrors = ParseErrors (NonEmpty ParseError)
...
+        #{Text.intercalate "\n" (fmap renderParseError (toList errs))}
+    |]
 
 instance Exception ParseErrors
 
 renderParseError :: ParseError -> Text
 renderParseError = \case
     CannotParseValue k v tr ->
-        [fmt|Cannot parse value {v} for option {k} into type {show tr}.|]
+        [__i|Cannot parse value #{v} for option #{k} into type #{show tr}.|]
```
bezirg, line 102: ditto

plutus-tx-plugin/app/GeneratePluginOptionsDoc.hs
```diff
@@ -53,17 +53,16 @@ optionsTable =
...
-     - {show (pretty defaultValue)}
-     - {desc}
-|]
-  where
-    valueType = show tr
-    defaultValue = O.defaultPluginOptions ^. field
+genRow k (O.PluginOption tr _ field desc) = [__i|
+    * - ``#{k}``
+      - #{show tr}
+      - #{show (pretty defaultValue)}
```
bezirg, line 64: ditto

michaelpj:
@bezirg this is getting reverted

PR #4863, opened 2022-09-22
[Builtins] [Costing] Add 'paramMultiplyIntegerV2'

effectfully:
This shows how to add a new way of costing an already existing builtin.


michaelpj:
This would require the cost model to have parameters for both "versions" of the builtin, right? Even though one is unused.

effectfully:
> This would require the cost model to have parameters for both "versions" of the builtin, right? Even though one is unused.

Yes, but the unused ones are still used _somewhere_ and so they need to be stored _somewhere_, so we could probably as well store them here?

Anyways, what's the alternative? To have extensible built-in models, create those huge `BuiltinCostModelBase`-like records each duplicating 90% of the previous one as well as a whole lot of supporting code? Have "delta" records overloading behavior for certain builtins and complicating the logic by a huge margin? Like what else can we possibly do?

michaelpj:
That's why there was a big spike for thinking about it! I haven't thought about it much, but maybe there's a way? Maybe not, though.

effectfully:
@kwxm requesting your review just so you can take a look.

I think this PR has served its purpose of providing an example of having versioned costing, so I'm closing it.

PR #4864, opened 2022-09-26
Bump hackage.nix + index-state & Revert to using PyF

zeme-iohk:


michaelpj (approved):
LGTM so long as Hydra is happy!

PR #4866, opened 2022-09-27
PLT-956: Fix cost model tests

kwxm:
Looking at PLT-953 (moving the cost model code into a separate Cabal component/package), I noticed that the cost model tests (which aren’t run in CI because of sporadic failures) were failing most of the time.  These tests run the Haskell costing functions for builtins and compare the results with predictions obtained directly from the R models, checking that they agree to within a small margin of error.  

The reason that they were failing is because #4778 changed the costing code to take objects as arguments where they’d previously taken sizes: the new version calculates the sizes itself.  Unfortunately there was an existing `ExMemoryUsage` instance for `ExMemory` which meant that the testing code continued to compile, but the Haskell costing functions were being fed the sizes of sizes, not the sizes themselves.  The result of this was that the Haskell costing functions were generally being given arguments which were all 1, but the R functions were getting genuine size arguments and so the results of the two versions differed.  This is fixed by using a wrapper type with a special `ExMemoryUsage` instance which allows us to pass size arguments to the Haskell costing functions: the tests all pass now.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
@@ -186,7 +186,7 @@ instance (ExMemoryUsage a, ExMemoryUsage b) => ExMemoryUsage (a, b) where
 instance ExMemoryUsage SatInt where
     memoryUsage n = memoryUsage (fromIntegral @SatInt @Int n)
     {-# INLINE memoryUsage #-}
-deriving newtype instance ExMemoryUsage ExMemory
+
 deriving newtype instance ExMemoryUsage Unique
 
 -- See https://github.com/input-output-hk/plutus/issues/1861
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
@@ -18,7 +18,6 @@ module PlutusCore.Evaluation.Machine.ExMemory
 ) where
 
 import PlutusCore.Data
-import PlutusCore.Name
 import PlutusCore.Pretty
 import PlutusPrelude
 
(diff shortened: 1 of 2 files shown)


michaelpj (approved):

michaelpj:
I wonder if we could try doing them in CI again? What was going wrong?

michaelpj:
Oh, it's not running because it's a benchmark not a test :D I guess that was our workaround...

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
```diff
@@ -186,7 +186,7 @@ instance (ExMemoryUsage a, ExMemoryUsage b) => ExMemoryUsage (a, b) where
 instance ExMemoryUsage SatInt where
     memoryUsage n = memoryUsage (fromIntegral @SatInt @Int n)
     {-# INLINE memoryUsage #-}
-deriving newtype instance ExMemoryUsage ExMemory
+
```
kwxm, line 189: This was what caused the problems.  There are more `ExMemoryUsage` instances nearby that I'm wondering if we should get rid of.  We only really want instances for built-in types.

kwxm:
> I wonder if we could try doing them in CI again? What was going wrong?

I'm still not sure, but see the comment [here](https://github.com/input-output-hk/plutus/blob/607e73a506092ff33965ff867cb819054cf2106b/plutus-core/cost-model/test/TestCostModels.hs#L113) and the earlier comment from Simon just above that.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
michaelpj: Go for it, if it compiles then I'm happy, if it doesn't then that tells us about some suspicious stuff.

michaelpj:
Ugh, that does sound bad :(

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
kwxm: I've now also removed the instances for `SatInt` and `Unique`.

effectfully (approved):
Thanks for fixing!

kwxm:
I was wondering why we had instances of `ExMemoryUsage` for `Term` in PlutusIR and (typed) PlutusCore, and whether we could remove them, but @effectfully says they're needed in a couple of places so I'll leave them there.  I did remove the `ExMemoryUsage` instance for `Name` so that may mean that some versions of `Term` won't have the memory usage instance.  CI's not complaining though, so I'll leave it like that.

PR #4869, opened 2022-09-28
Internal library for inline-r (PLT-953)

kwxm:
This modifies `plutus-core.cabal` to put the `inline-r` dependencies in an internal library, which should stop `plutus-core` depending on `inline-r`, which is problematic for the node.

I'm not sure how to test if it solves the original problem, but it doesn't seem to break anything in `plutus-core`.


michaelpj:
Ah, this isn't going to solve the actual problem. We don't want `plutus-core` to depend on `inline-r` even _transitively_. So my idea was to have all the stuff that _uses_ `inline-r` in an internal library, and then use _that_ from e.g. `cost-model-tests` or whatever, while keeping the main library clear of it.

zliu41:
> We don't want plutus-core to depend on inline-r even transitively.

But `plutus-core:plutus-core` doesn't. It is `plutus-core:cost-model-test` and `plutus-core:generate-cost-model` that depend on `inline-r`, so what's the problem?

michaelpj:
There is no problem :facepalm:

PR #4873, opened 2022-09-29
std flake: import haskell.nix project devshell into main devshell.

jbgi:
this required bumping haskell.nix to latest master.


zeme-iohk:
Hey @jbgi thank you for this! 
Unfortunately running `nix develop` inside `__std__` triggers a cold build of `ghc-8.6.5`, among other things.
I'm trying to figure out a way to avoid that.

__std__/flake.nix
```diff
@@ -16,18 +16,18 @@
   # { inputs, cell }: inputs.nixpkgs, inputs.haskell-nix
   inputs = {
     nixpkgs = {
-      url = "github:NixOS/nixpkgs/34e4df55664c24df350f59adba8c7a042dece61e";
+      follows = "haskell-nix/nixpkgs";
```
michaelpj, line 19: This is a major change: we're deliberately not tying ourselves to the `haskell.nix` nixpkgs at the moment. I'd rather keep it that way unless we have a good reason not to.

michaelpj: And it forces you to fix the R thing!

__std__/nix/toolchain/library/pkgs.nix
```diff
@@ -11,33 +11,17 @@
...
 let
 
-  haskell-nix-bootstrap = import inputs.haskell-nix {
-
-    pkgs = import inputs.nixpkgs.path {
-      system = inputs.nixpkgs.system;
-    };
-
-    sourcesOverride = {
-      hackage = inputs.hackage-nix;
```
michaelpj, line 21: No! We absolutely definitely want to keep pinning `hackage.nix` separately from `haskell.nix`, it's far more convenient!

__std__/nix/toolchain/library/r-overlay.nix
```diff
@@ -2,26 +2,7 @@
 
 self: super: {
 
-  rPackages = super.rPackages.override {
```
michaelpj, line 5: are these overrides not needed?

__std__/nix/toolchain/library/r-overlay.nix
```diff
@@ -2,26 +2,7 @@
...
-      substituteInPlace configure --replace \
-        "-install_name libRblas.dylib" "-install_name $out/lib/R/lib/libRblas.dylib"
-      substituteInPlace configure --replace \
-        "-install_name libRlapack.dylib" "-install_name $out/lib/R/lib/libRlapack.dylib"
-      substituteInPlace configure --replace \
-        "-install_name libR.dylib" "-install_name $out/lib/R/lib/libR.dylib"
-    '';
-  });
+  # Use old R 4.1.3 due to https://github.com/tweag/HaskellR/issues/374
+  inherit (inputs.nixpkgs-r.legacyPackages) rPackages R;
```
michaelpj, line 6: This seems unlikely to work in general: our nixpkgs and this one could diverge on common libraries like `glibc`. I think we should probably instead override the R package to pin the version to an old one.

michaelpj (commented):
@jbgi it would be helpful if you didn't include unrelated changes in PRs like this :)

__std__/nix/toolchain/library/pkgs.nix
michaelpj: If you bump your index-state you need to bump hackage.nix. That will likely force you to rebuild a bunch of haskell stuff, but _not_ GHC. Whereas bumping haskell.nix can often force you to rebuild GHC, so is far more disruptive. I think everyone should pin them separately.

__std__/nix/toolchain/library/r-overlay.nix
jbgi: I think it has been fixed upstream with https://github.com/NixOS/nixpkgs/pull/108268#issuecomment-753688799
but I see that it was only for darwin, so maybe that is still needed for linux. Not sure how to check... ?

michaelpj: Don't try and fix the R thing here until we have working CI? :) Or fix it in the main nix setup by bumping nixpkgs and check that CI is happy, then port the changes over here. If we do it now and we don't know it works it's just a landmine for us later.

__std__/flake.nix
```diff
@@ -16,17 +16,21 @@
...
+      url = "github:NixOS/nixpkgs";
     };
     std = {
       url = "github:divnix/std";
       inputs.nixpkgs.follows = "nixpkgs";
     };
     haskell-nix = {
-      url = "github:input-output-hk/haskell.nix/e36a34c4abc02287ec79bf88a0c993f6b66f9945";
+      url = "github:input-output-hk/haskell.nix";
+      inputs = {
```
zeme-iohk, line 27: I had tried the exact same myself i.e. to move the `haskell-nix-bootstrap` logic of `__std__/nix/toolchain/library/pkgs.nix` into the flake inputs like you did, but it didn't seem to work, it was triggering a full rebuild of everything...
It is not doing it now? Ill go anche check myself

flake.nix
```diff
@@ -13,7 +13,7 @@
       flake = false;
     };
     haskell-nix = {
-      url = "github:input-output-hk/haskell.nix";
+      url = "github:input-output-hk/haskell.nix/devshells-noExactDeps";
```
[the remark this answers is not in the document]
jbgi: fixed

__std__/nix/plutus/devshells/plutus-shell.nix
```diff
@@ -9,11 +9,27 @@ inputs.std.std.lib.mkShell {
...
   imports = [
+    cell.library.plutus-project.devshell
     inputs.cells.toolchain.devshellProfiles.haskell
   ];
 
-  packages = pkgs.lib.optionals (!pkgs.stdenv.isDarwin) [
+  packages = [
+    pkgs.libsodium-vrf
+    pkgs.libsodium-vrf.dev
+    pkgs.secp256k1
```
michaelpj, line 19: Why are these necessary? The `haskell.nix` shell should provide them?

__std__/nix/plutus/devshells/plutus-shell.nix
```diff
@@ -9,11 +9,27 @@ inputs.std.std.lib.mkShell {
...
+  ] ++ pkgs.lib.optionals (!pkgs.stdenv.isDarwin) [
     pkgs.rPackages.plotly
     pkgs.R
   ];
+
+  env = [
```
michaelpj, line 25: Why do we need to do this? I would expect the devshell stuff to manage its own environment variables like this? Really needs a comment at least.

__std__/nix/plutus/library/plutus-project.nix
```diff
@@ -33,6 +33,11 @@ let
       name = "plutus";
     };
 
+    shell = {
```
michaelpj, line 36: We use an explicit invocation of `shellFor` which already sets this, so I think this does nothing?

__std__/nix/toolchain/library/cabal-project-index-state.nix
```diff
@@ -1,4 +1,3 @@
 { inputs, cell }:
 
-# TODO(std) revisit why we need this, and why we were parsing it from the actual cabal.project
-"2022-08-31T00:00:00Z"
+inputs.cells.plutus.library.plutus-project.index-state
```
michaelpj, line 3: please leave the todo, I'm not sure if this is actually equivalent to what we were doing before

__std__/nix/plutus/library/plutus-project.nix
```diff
@@ -191,4 +196,6 @@ let
 
 in
 
-project 
+project.appendOverlays [
+  library.haskell-nix.haskellLib.projectOverlays.devshell
```
michaelpj, line 200: Or is this redoing the whole shell setup? If we do that then there's a lot more that we need to do... It would really help if you explained what you were doing.

zeme-iohk: We are trying to get a working haskell [devshell](https://github.com/numtide/devshell) for building plutus -- replacing  `haskell-nix.shellFor`
Reason being that it plays nicely with the rest of `std` stuff, mostly `devshellProfiles`

__std__/nix/plutus/devshells/plutus-shell.nix
zeme-iohk: I was also expecting the [devshell](https://github.com/numtide/devshell) to do this automatically.

jbgi: devshell want all environment variables to be explicit. There is some limited support to ease this if we import https://github.com/numtide/devshell/blob/master/extra/language/c.nix though.

__std__/flake.nix
zeme-iohk: Turns out I must have done something wrong, this works.

__std__/nix/plutus/devshells/plutus-shell.nix
zeme-iohk: So I've tested the new shell and I can do a full `cabal build all` on my machine.

Following the `c.nix` example you posted, I reckon the `env` stuff should therefore be moved to 
`inputs.cells.toolchain.devshellProfiles.haskell` instead?

__std__/nix/plutus/devshells/plutus-shell.nix
zeme-iohk: @jbgi this in fact a little annoying, as `shellFor` would do it automatically.
Apart from this, the new setup seems to work

__std__/nix/plutus/devshells/plutus-shell.nix
jbgi: ah yeah. But could probably also be upstreamed in https://github.com/input-output-hk/haskell.nix/blob/7462e537efccfc4471908a5c96c5c72b2646f985/lib/project-overlays.nix#L16
Also working a patch to avoid explicitly listing libsodium/secp256k1.

michaelpj:
I do think that doing the nixpkgs bump and R fix on the main setup also is a good idea. That will also tell you if it works.

__std__/nix/plutus/devshells/plutus-shell.nix
michaelpj: Are we the first people to use this? :sweat_smile:

zeme-iohk:
I'm happy with this, than

zeme-iohk:
I'm pretty happy with this, thank you @jbgi

zeme-iohk (approved):

__std__/flake.nix
```diff
@@ -16,17 +16,23 @@
...
+        hackage.follows = "hackage-nix";
+        nixpkgs.follows = "nixpkgs";
+        # Ignore stackage (self reference)
```
michaelpj, line 30: Can you say a bit more? Or can we just leave out this line?

nix/overlays/r.nix
```diff
@@ -7,7 +7,13 @@ self: super: {
       });
     });
   };
-  R = super.R.overrideAttrs (oldAttrs: {
+  R = super.R.overrideAttrs (oldAttrs: rec {
```
michaelpj, line 10:
```suggestion
  # inline-r only builds with an old version of R: https://github.com/tweag/HaskellR/issues/374
  R = super.R.overrideAttrs (oldAttrs: rec {
```

michaelpj (approved):
Fine with this (remember to ditch the first commit since we don't need it now). I'm not sure where we're going with the shell expression since the new (?) thing doesn't have a lot of the stuff that the current one has, but I think Lorenzo knows where this is going so I'm happy.

__std__/flake.nix
jbgi: this was an "hack" to avoid downloading new stackage each time we update haskell.nix (since we don't use stackage).
But can be removed.

jbgi: But not sure why that was necessary, since inputs loading should lazy now. I'll remove for now.

PR #4870, opened 2022-09-29
New `std` + Flake CI - Part 1

zeme-iohk:
Added a new cell `automation` with a `hydraJobs` block, defining the `required-jobset` for hydra.

Currently our hydra setup is based on `release.nix`. 
With `std`, we'll need a flake-based hook.

Migration will follow these steps:

- [x] The ci-ops PR https://github.com/input-output-hk/ci-ops/pull/104 gets merged 
- [ ] The new dummy jobset in the top-level flake added by this PR can be built successfully 
- [ ] The current `release.nix` jobset is migrated to the top-level `flake.nix` 
- [ ] The new jobset in `__std__/nix/automation/hydraJobs/required-jobset` is migrated to the top-level `flake.nix`

PR #4877, opened 2022-10-01
New std + Flake CI - Part 1 

zeme-iohk:
Added a new cell automation with a hydraJobs block, defining the required-jobset for hydra.

Currently our hydra setup is based on release.nix.
With std, we'll need a flake-based hook.

Migration will follow these steps:

- [x]   The ci-ops PR https://github.com/input-output-hk/ci-ops/pull/104 gets merged
- [x]   Both old and new CI are triggered 
- [x]  New CI passes 

See https://github.com/input-output-hk/plutus/pull/4870


.github/workflows/test.yml
```diff
@@ -2,21 +2,21 @@ name: "Tests"
 on:
   pull_request:
 jobs:
```
zeme-iohk, line 4: This will be restored before merging

__std__/nix/plutus/library/plutus-project.nix
```diff
@@ -102,6 +102,8 @@ let
           # > Error: C stack usage  17556409549320 is too close to the limit
           # > Fatal error: unable to initialize the JI
           plutus-core.components.exes.generate-cost-model.buildable = lib.mkForce false;
+
+          plutus-core.components.benchmarks.cost-model-test.buildable = lib.mkForce false;
```
zeme-iohk, line 106: This is the only job left that fails. 
It fails with the same error as `generate-cost-model` above (stack use too close to the limit)

__std__/nix/toolchain/library/cabal-project-index-state.nix
```diff
@@ -1,4 +1,4 @@
 { inputs, cell }:
 
 # TODO(std) revisit why we need this, and why we were parsing it from the actual cabal.project
-"2022-08-31T00:00:00Z"
+"2022-09-26T00:00:00Z"
```
zeme-iohk, line 4: Reflects the current cabal.

__std__/nix/toolchain/packages/default.nix
```diff
@@ -4,8 +4,6 @@
 
   cabal-install = import ./cabal-install.nix block;
 
-  cabal-install-alt = import ./cabal-install-alt.nix block;
```
zeme-iohk, line 7: Obsolete

flake.nix
```diff
@@ -1,51 +1,263 @@
-# NOTE: This flake is only used for source pinning and has no outputs
```
zeme-iohk, line 1: flake.lock has been renamed to old.flake.lock
nx has been renamed to old.flake.nix 

the new flake+lock have been moved from __std__ to the top-level.
But git doesn't see that as a "move", but as a diff. 
So this is a diff between the current and new flake. It's a mess, it can be skipped over

old.flake.lock
```diff
@@ -0,0 +1,182 @@
+{
```
zeme-iohk, line 1: the old flake, unchanged. This file can be ignored

old.flake.nix
```diff
@@ -0,0 +1,55 @@
+# NOTE: This flake is only used for source pinning and has no outputs
```
zeme-iohk, line 1: Same as above

flake.nix
zeme-iohk: The reason this was done is: 
We need a way to test the new flake-based CI, so we need the new flake to be at the top-level.
Fortunately, the old flake was only used for source pinning, so it's ok to rename it and leave it as `old.flake.nix`.

__std__/nix/automation/hydraJobs/all.nix
```diff
@@ -0,0 +1,67 @@
...
+  native-plutus-jobs = make-haskell-jobs plutus-project;
+
+  windows-plutus-jobs = make-haskell-jobs plutus-project.projectCross.mingwW64;
+
+  other-jobs = {
+    plutus.shells = inputs.cells.plutus.devshells;
+    doc.shells = inputs.cells.doc.devshells;
+    doc.packages = inputs.cells.doc.packages;
+    doc.scripts = inputs.cells.doc.scripts;
+    # toolchain.packages = inputs.cells.toolchain.packages;
```
zeme-iohk, line 33: These still need some attention, working on them next

__std__/nix/plutus/library/plutus-project.nix
```diff
@@ -102,6 +102,8 @@ let
           # > Error: C stack usage  17556409549320 is too close to the limit
           # > Fatal error: unable to initialize the JI
           plutus-core.components.exes.generate-cost-model.buildable = lib.mkForce false;
+
+          # plutus-core.components.benchmarks.cost-model-test.buildable = lib.mkForce false;
```
zeme-iohk, line 106: Sadly this is the _only_ thing that fails!
It works in the old CI, not here.
See  https://hydra.iohk.io/eval/1284384

flake.nix
zeme-iohk: Unless...
What if we changed this line https://github.com/input-output-hk/ci-ops/blob/fcbd80ad860405ad947bc95eb707db10f081de53/jobsets/default.nix#L374
from `github:input-output-hk/plutus` to `github:input-output-hk/plutus/__std__` 
Then it should pick up the flake in `github:input-output-hk/plutus/__std__/flake.nix`
And so we can keep the original flake.nix as is in the top-level
What do you think @jbgi?

jbgi: No, think hydra will look at the `__std__` branch instead of subdirectory.
I think the way to go is to add the sub-flake as input to the main flake (with `url = path:./__std__`) and expose the `hydraJobs` attribute in the main flake.

michaelpj: > I think the way to go is to add the sub-flake as input to the main flake (with url = path:./__std__) and expose the hydraJobs attribute in the main flake.

If that works it would be nice!

__std__/nix/plutus/library/plutus-project.nix
michaelpj: Yes, that's it, see here for where we set the platforms: https://github.com/input-output-hk/plutus/blob/master/nix/pkgs/haskell/haskell.nix#L134

I thought the current thing was quite elegant in that it works off the existing nixpkgs `meta` attributes for platforms and works generically, but I won't cry if you bin it.

flake.nix
```diff
@@ -1,51 +1,304 @@
...
+#
+# This repository uses the std tool https://github.com/divnix/std.
+# Familiarity with std is required to be able to contribute effectively.
+# While official documentation for std can be found in its GitHub, this flake
+# has been thoroughly commented so as to quickstart new maintainers.
+#
+# Most of what there is to know about the nix code inside this repository
+# can be learned by reading this file. A second read will be needed as some
+# std-specific terms may be referenced first and defined later.
+# You may also refer to the glossary: https://divnix.github.io/std/glossary.html
```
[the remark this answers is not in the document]
michaelpj: ... why? shouldn't they be in sync anyway? 

also you'll need to fix `nix/sources.nix` which is how the old code gets the inputs out of here

zeme-iohk:
new CI is green 👍 
https://hydra.iohk.io/jobset/Cardano/plutus-flaked-pr-4877

__std__/nix/automation/hydraJobs/all.nix
```diff
@@ -0,0 +1,63 @@
+{ inputs, cell }:
+
+let
+  inherit (inputs.cells.toolchain.library) pkgs haskell-nix;
+  inherit (pkgs.stdenv) system;
```
michaelpj, line 5: Is this the approved way of getting the current system in std?

michaelpj:
Keep an eye on https://github.com/input-output-hk/haskell.nix/pull/1743, if Hamish makes my suggestions we'll be able to reuse quite a few generic functions for collecting the packages etc which will be quite nice!

nix/sources.nix
```diff
@@ -19,11 +19,24 @@ let
...
+      (fetcher-args lock.nodes.${inputs.__old__nixpkgs}.locked))
     {
       inherit system;
     };
+
+  tmpSources = builtins.mapAttrs
+    (_: node-name: pkgs.fetchzip (fetcher-args lock.nodes.${node-name}.locked))
+    inputs;
+
+  finalSources = {
```
michaelpj, line 31: ugly, but fair enough

__std__/nix/automation/hydraJobs/default.nix
```diff
@@ -0,0 +1,4 @@
+{ inputs, cell }@block: rec
+{
+  all = import ./all.nix block;
```
michaelpj, line 3: Looks like we end up with a kind of pointless `all` in the attribute names for the jobs, can we avoid that?

michaelpj:
Job names seem a big long, but I guess that's okay:
image

__std__/nix/automation/hydraJobs/all.nix
```diff
@@ -0,0 +1,63 @@
...
+    if system == "x86_64-linux" then {
+      plutus.windows = windows-plutus-jobs;
+      plutus.native = native-plutus-jobs;
+      other = other-jobs;
+    } else if system == "x86_64-darwin" then {
+      plutus.native = native-plutus-jobs;
+      other = other-jobs;
+    } else { };
+
+  # Hydra doesn't like these attributes hanging around in "jobsets": it thinks they're jobs!
```
michaelpj, line 47: Do you actually still need this? I thought this was an artifact of how `ci.nix` is setup?

__std__/nix/automation/hydraJobs/all.nix
```diff
@@ -0,0 +1,63 @@
...
+      plan-nix = project.plan-nix;
+    };
+
+  plutus-project = inputs.cells.plutus.library.plutus-project;
+
+  native-plutus-jobs = make-haskell-jobs plutus-project;
+
+  windows-plutus-jobs = make-haskell-jobs plutus-project.projectCross.mingwW64;
+
+  other-jobs = {
```
michaelpj, line 28: Looking at this makes me think: what if we defined `ciJobs` per cell and then just harvested them all? That's kind of what's going on here: each cell is defining some jobs. Then `automation` would go away as a cell. Not sure if you find that better or worse organized tbh!

michaelpj: I guess you would still need to define the top-level jobset and the required job :thinking:

zeme-iohk:
> Keep an eye on https://github.com/input-output-hk/haskell.nix/pull/1743
👍 

> Job names seem a big long, but I guess that's okay:
> image

Yes job names are kinda long.
We can't get rid of the system name.
I might have an idea how to get rid of the `all` bit, and I guess I could flatten the `other-jobs` too. I will try in the next PR.

PR #4881, opened 2022-10-04
Make cross-compilation work again.

jbgi:
To correctly eval the haskell.nix project we need the pkgs for the current
 system, not the cross system one.

 cf. https://github.com/input-output-hk/haskell.nix/issues/1666

### DIFF
--- a/__std__/nix/plutus/library/plutus-project.nix
+++ b/__std__/nix/plutus/library/plutus-project.nix
@@ -22,6 +22,8 @@ let
 
     compiler-nix-name = library.ghc-compiler-nix-name;
 
+    evalPackages = library.pkgs;
+
     # This is incredibly difficult to get right, almost everything goes wrong,
     # see https://github.com/input-output-hk/haskell.nix/issues/496
     src = library.haskell-nix.haskellLib.cleanSourceWith {
@@ -149,7 +151,7 @@ let
             platforms = lib.platforms.linux;
           };
 
-          # Werror everything. 
+          # Werror everything.
           # This is a pain, see https://github.com/input-output-hk/haskell.nix/issues/519
           plutus-core.ghcOptions = [ "-Werror" ];
 
@@ -191,4 +193,4 @@ let
 
 in
 
-project 
+project
--- a/nix/pkgs/haskell/haskell.nix
+++ b/nix/pkgs/haskell/haskell.nix
@@ -20,6 +20,7 @@ let
   r-packages = with rPackages; [ R tidyverse dplyr stringr MASS plotly shiny shinyjs purrr ];
   project = haskell-nix.cabalProject' ({ pkgs, ... }: {
     inherit compiler-nix-name;
+    evalPackages = pkgs.buildPackages.buildPackages;
     # This is incredibly difficult to get right, almost everything goes wrong,
     # see https://github.com/input-output-hk/haskell.nix/issues/496
     src = let root = ../../../.; in


nix/pkgs/haskell/haskell.nix
```diff
@@ -20,6 +20,7 @@ let
   r-packages = with rPackages; [ R tidyverse dplyr stringr MASS plotly shiny shinyjs purrr ];
   project = haskell-nix.cabalProject' ({ pkgs, ... }: {
     inherit compiler-nix-name;
+    evalPackages = pkgs.buildPackages.buildPackages;
```
michaelpj, line 23: comment, please! Also I don't think we want this: this makes it not possible to evaluate things on a different system from the build one, which we used to be able to do. So should we not actually use `evalPackages` here?

michaelpj: Indeed, isn't that the default? or is the issue that the default is broken?

michaelpj (commented):
I have some questions D:

michaelpj:
Thanks for the fix!

nix/pkgs/haskell/haskell.nix
jbgi: `evalPackages` default is `pkgs.buildPackages`. But apparently this not enough, probably because:
```
nix eval -f ./default.nix --arg crossSystem '{ config = "x86_64-w64-mingw32"; libc = "msvcrt"; }' plutus.haskell.project.pkgs.buildPackages.targetPlatform.system
"x86_64-windows"
```
while with `pkgs.buildPackages.buildPackages`
```
nix eval -f ./default.nix --arg crossSystem '{ config = "x86_64-w64-mingw32"; libc = "msvcrt"; }' plutus.haskell.project.pkgs.buildPackages.buildPackages.targetPlatform.system
"x86_64-linux"
```
But not sure why exactly. Maybe @hamishmack ?

hamishmack: Oh that is a good point.  Just using `.buildPackages` has caused problems in the past with caching.  IIRC anything that used python (including some surprising things like git) would include a dependency on `gdb` (I think it was `gdb`), but `buildPackages.gdb` would be the cross compiler `gdb` (not the native one).  The result was that git would recompile.

__std__/nix/plutus/library/plutus-project.nix
```diff
@@ -22,6 +22,8 @@ let
 
     compiler-nix-name = library.ghc-compiler-nix-name;
 
+    evalPackages = library.pkgs;
```
michaelpj, line 25: Should this not also use `library.pkgs.buildPackages.buildPackages`?

jbgi: no, in this case `library.pkgs` is the original pkgs (not from any `cross*` invocation), so the one we want.

michaelpj:
Also, I remembered that there's a simpler and clearer way to get this: `pkgs.pkgsBuildBuild` https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/stage.nix#L116

__std__/nix/plutus/library/plutus-project.nix
michaelpj: That doesn't make sense to me, aren't these two files literally the same thing just with one of them setup for the new infra? how are we doing cross builds in the `std` world then?

jbgi: with `plutus-project.projectCross.mingwW64`.

michaelpj: More attributes over functions :( What was wrong with the old way? I won't fight about it though.

jbgi: though, using `pkgs.pkgsBuildBuild` should also work and must be the same. so we can use it instead `library.pkgs`.

michaelpj: I'm very pro consistency if it doesn't cost us anything!!

nix/pkgs/haskell/haskell.nix
hamishmack: See https://github.com/input-output-hk/haskell.nix/pull/1739

__std__/nix/plutus/library/plutus-project.nix
jbgi: thanks. I'll close this PR then.

PR #4884, opened 2022-10-05
Setup cicero ci actions

Pacman99:
Export hydraJobs based on ci.nix and then use that to export tullia actions


nix/lib/ci.nix
```diff
@@ -1,6 +1,6 @@
+# Generic nixpkgs, use *only* for lib functions that are stable across versions
+{ pkgs ? import (import ../sources.nix { system = builtins.currentSystem; }).nixpkgs { } }:
```
michaelpj, line 2: maybe just drop the default value and explicitly pass it in `release.nix` also

nix/tullia.nix
```diff
@@ -0,0 +1,72 @@
...
+    common = {
+      config,
+      ...
+    }: {
+      preset = {
+        # needed on top-level task to set runtime options
+        nix.enable = true;
+
+        github-ci = {
+          enable = config.actionRun.facts != {};
```
michaelpj, line 18: what does this do? why is it set to what it is?

nix/tullia.nix
```diff
@@ -0,0 +1,72 @@
...
+    mkHydraJobTasks = __mapAttrs (_: mkHydraJobTask);
+
+    hydraJobTasks = mkHydraJobTasks (flakeOutputTasks ["hydraJobs" os] self);
+
+    ciTasks = taskSequence "ci/" hydraJobTasks (__attrNames hydraJobTasks);
+  in
+    ciTasks // {
+      "ci" = {lib, ...}: {
+        imports = [common];
+        after = [(lib.last (__attrNames ciTasks))];
```
michaelpj, line 51: background question: is this a graph-like dependency ordering? i.e. could we say `after = _attrNames ciTasks` to say that it has to run after all the other tasks, and then rely on `taskSequence` to ensure that all the others run sequentially? Or do we actually have to say it runs specifically after the last one?

nix/tullia.nix
```diff
@@ -0,0 +1,72 @@
...
+      io = ''
+        let github = {
+          #input: "${ciInputName}"
+          #repo: "input-output-hk/plutus"
+        }
+
+        #lib.merge
+        #ios: [
+          #lib.io.github_push & github,
+          { #lib.io.github_pr, github, #target_default: false },
```
michaelpj, line 67: What do these things do?

michaelpj (commented):
Looks pretty plausible! Would just be nice to get some more comments for those of us who aren't as familiar with what's going on.

And a generic question: do _all_ the tasks produce github statuses? How do we know, what controls that?

flake.nix
```diff
@@ -45,7 +41,37 @@
...
+  outputs = { self, utils, tullia, ... } @ inputs:
+    let
+      supportedSystems = [ "x86_64-linux" "x86_64-darwin" ];
+    in
+      utils.lib.eachSystem supportedSystems (system:
+        {
+          # TODO
+          # do it similar to cardano-node?
+          # or use flattenTree and filter by .system/.meta.platforms?
+          #   might be an option because there doesn't seem to be IFD in here (in contrast to
```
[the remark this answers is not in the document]
dermetfan: I put everything in system namespaces and set `supportedSystem` so that eval should work

nix/tullia.nix
dermetfan: Tullia tasks can run locally or on Cicero. When no facts are present we know that we are running locally and vice versa. When running locally, the current directory is already bind-mounted into the container, so we don't need to fetch the source from GitHub and we don't want to report a GitHub status.

nix/tullia.nix
dermetfan: Right, that should work too. I'll give it a try

nix/tullia.nix
dermetfan: This is the CUE expression that defines when to run the action. In particular, it uses tullia's library from `#lib` to build an expression that matches the JSON document that the GitHub webhook event sends on a push or PR.

michaelpj: Worth explaining! or at least adding a link to where to find out!

dermetfan:
> do all the tasks produce github statuses? How do we know, what controls that?

Yes, they all do. This is enabled by `preset.github-ci` which is enabled in `common`.

flake.nix
```diff
@@ -45,7 +41,37 @@
...
+          #   cardano-node)
+          #   but might be bad idea because it will break if they ever add IFD
+
+          # Does not follow flake output schema, only created to pass to tullia
+          hydraJobs = import ./ci.nix {
```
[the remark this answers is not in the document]
dermetfan: It doesn't need to be a valid set of hydra jobs. `flakeOutputTasks` just expects some arbitrarily-nested attrset that contains derivations. In fact, it cannot be: The flake output schema is `.#hydraJobs.<attr-maybe-nested>.<system> = drv`. But in order to evaluate which jobs exist we may need IFD that builds on another system. So instead I opted to put the system first and set `supportedSystems` to only the respective system, so that we can safely select `.#hydraJobs.<system>` and recurse into that.

nix/tullia.nix
```diff
@@ -0,0 +1,72 @@
...
+in rec {
+  tasks = let
+    inherit (self.inputs.tullia) flakeOutputTasks taskSequence;
+
+    common = {
+      config,
+      ...
+    }: {
+      preset = {
+        # needed on top-level task to set runtime options
```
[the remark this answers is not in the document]
dermetfan: The nix preset [does set](https://github.com/input-output-hk/tullia/blob/a05c75d87f0f3bc4625503ea4d40b8ce2725d8af/nix/preset/nix.nix#L23) some nix options. We don't in the plutus repo. Still, we have to keep this line in order to enable the nix preset at all.

nix/tullia.nix
dermetfan: There are no docs yet :sweat_smile:

michaelpj:
Is this actually getting run somewhere? How do we know it works? Can we test it locally?

michaelpj:
(Ideally we'd document the answers to those questions in CONTRIBUTING, which already says something about how e.g. Hydra works today)

dermetfan:
> Is this actually getting run somewhere?

Yes, on [cicero.ci.iog.io](https://cicero.ci.iog.io). Look for an action called [plutus/ci](https://cicero.ci.iog.io/action/15d635e6-b19a-4a82-9e66-439fff0733c8).

> How do we know it works?

It reports GitHub statuses. These also have a details link that will take you to our Cicero instance where you can inspect the logs etc.

For example:
`ci/ciJobs.x86_64-linux.linux.docs.cost-model-notes — Took 13s`

If you wish you can make only certain statuses required in the repo settings, or require the `ci` status, which is the aggregate of all others.

> Can we test it locally?

Yes, using `tullia run`. At least in theory. This is not very well tested at the moment and may be broken. Once we expect it to work we can link to the docs that we will hopefully have by then.

I added a few lines about that in CONTRIBUTING.

flake.nix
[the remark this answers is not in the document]
michaelpj: That all sounds sensible. I also like putting the system first!

CONTRIBUTING.adoc
```diff
@@ -493,6 +494,14 @@ These will be automatically retried, but if you're in a hurry Michael has permis
 Nondeterministic failures are very annoying.
 Michael also has permissions to restart failed builds.
 
+==== Cicero
+
+[Cicero](https://github.com/input-output-hk/cicero) is our new CI and automation system.
+It builds all the same jobs as Hydra by reading `ci.nix`.
+
+For each job, it we generate a [Tullia](https://github.com/input-output-hk/tullia) task.
+Those can be run locally and on Cicero. Run `tullia list` to see all available tasks and `tullia run $job` to run it.
```
michaelpj, line 503: We probably shouldn't say this unless we actually have `tullia` in the shell!

CONTRIBUTING.adoc
```diff
@@ -493,6 +494,14 @@ These will be automatically retried, but if you're in a hurry Michael has permis
 Nondeterministic failures are very annoying.
 Michael also has permissions to restart failed builds.
 
+==== Cicero
+
+[Cicero](https://github.com/input-output-hk/cicero) is our new CI and automation system.
```
michaelpj, line 499: also link to https://cicero.ci.iog.io/ and maybe to the main action itself?

CONTRIBUTING.adoc
dermetfan: @Pacman99 can I push that onto you? :grin:

Pacman99: done

michaelpj:
I see a lot of this in the logs:
```
warning: ignoring untrusted flake configuration setting 'extra-substituters'
warning: ignoring untrusted flake configuration setting 'extra-trusted-public-keys'
```
I guess that's _probably_ what you want, allowing the per-flake nix config settings probably is dangerous.

michaelpj:
The `details` links are very confusing. They all link to the same place, which appears to be the `plutus/ci` "action" (whatever that is). From close inspection of the logs, it does appear to actually be building all the constituents. I'd assumed that each task would get its own page? This way seems quite difficult to parse.

michaelpj:
I'm going to make some issues on cicero for you :)

michaelpj:
I think this one is my biggest complaint at the moment: https://github.com/input-output-hk/cicero/issues/59

michaelpj:
Also unfortunately this now conflicts with @zeme

zeme-iohk:
There is an example of std+tullia integration in cardano-world https://github.com/input-output-hk/cardano-world/blob/d22f50fc77d23e2612ca2b313a098dd0b48834d4/flake.nix#L194
I should be ok resolving any conflicts myself after this PR is merge an I'll ask you @dermetfan to review

zeme-iohk:
It's actually pretty sweet to have all jobs as GHAs. 
I wonder if it's normal that only the x86_64-linux jobs show up? 
Is it still a work in progress?

nix/tullia.nix
```diff
@@ -0,0 +1,72 @@
...
+      ciTasks = __mapAttrs
+        (_: flakeOutputTask: { ... }: {
+          imports = [ common flakeOutputTask ];
+
+          memory = 1024 * 8;
+          nomad.resources.cpu = 10000;
+        })
+        (flakeOutputTasks [ "ciJobs" system ] self);
+
+      ciTasksSeq = taskSequence "ci/" ciTasks (__attrNames ciTasks);
```
michaelpj, line 41: Also, does this mean that if a task fails, none of the ones after it will be run, even if they would succeed? That's _significantly_ worse for diagnosing issues than the current setup. 

Could we at least make it so that the different systems are run independently? They shouldn't share any deps so there won't be any risk of redoing work, and that's one of the most common ways for a subset of the builds to fail (system X is broken but not system Y).

michaelpj:
I mean, this PR conflicts with what just got merged. I guess you could push to this branch and fix conflicts if you want, @zeme-iohk ?

michaelpj:
>  I wonder if it's normal that only the x86_64-linux jobs show up?

Good point, that's definitely something missing.

Pacman99:
@zeme-iohk I can handle the merge conflicts, I'll go ahead and standardize this code too based on how it's done in cardano-world as you pointed out. 

>  I wonder if it's normal that only the x86_64-linux jobs show up?

Also there are no available darwin systems in cicero yet, so only linux tasks are run. But the nix code is all there for running x86_64-darwin tasks.

nix/tullia.nix
Pacman99: I believe the other tasks will keep running, the one that failed will just report a github status failure. @dermetfan can you confirm this?

And I think its already setup so different systems run independently. The `tullia` output has seperate exports for each system.

zeme-iohk:
OK then go ahead 😃

zeme-iohk:
note that `__std__/flake.nix` is obsolete and it's ok to delete it.

nix/tullia.nix
dermetfan: Unfortunately no, @michaelpj's concern is correct. We could start all tasks at the same time, in which case they would all finish independently of each other, but I purposely did not do that to keep the needed memory down. Maybe that's not necessary here, we could try.

michaelpj: Well, there's also the worry about them redoing work, since they'll all have shared Nix dependencies. I don't mind too much about them running sequentially, but it would be nice if we got failure reports for everything. Although actually maybe that's a bad idea: if there's a failing nix derivation that they all depend on, then trying to keep going would just result in building the same derivation over and over and failing each time, which could be bad/expensive. Hydra handles this more gracefully...

Pacman99:
I got it setup under the cicero std interface. I am running into a strange infinite recursion error. I've traced it through tullia and I believe its an issue with the current cells setup. Because tullia has to evaluate all the different hydraJobs, there may be somewhere that evaluation loops around. @zeme-iohk do you have any ideas of the cause?

zeme-iohk:
The `doc` and `plutus` cell depend on the `toolchain` cell. 
I'd be surprised if this were a problem?
It truth the `toolchain` cell also taps into the `plutus` cell in only one place (`library/cabal-project-index-state.nix` -- this is something that will be addressed), but I've tried removing that and it still runs into infinite recursion.

I've also tried something else: create an `automation/packages.nix` with a `dummy` derivation and use `outputs.${system}.hydraJobs = cell.packages.dummy;`
This *also* runs into infinite recursion.
Also...
`outputs.${system}.hydraJobs = cell;` --> infinite recursion.
And using `inputs.cells.automation` doesn't do it :P 
So it seems that the problem is that we cannot use the `cell` variable.
 
If this were the case, we have to create an additional cell (maybe `cloud`) to house `pipelines.nix`, which would tap into `automation.hydraJobs`. 
Not super happy about it, but if it's our only choice then be it. 

But before we do, it might be worth trying two things:
- Update tullia so that we don't get the std warning about using clades
- Ask the cicero/tullia developers to confirm our suspicion of `cell`

michaelpj:
I had some serious problems with infinite recursion when I set up `ci.nix` the first time. The issue I eventually tracked down is that you can get weird things where derivations have attributes that are attribute sets that can contain themselves or something. You need to be very careful how you traverse a nested attribute set to avoid this. There's a bunch of code in `lib/ci.nix` to deal with this. I saw that tullia has some similar code, but maybe it doesn't quite handle the same cases?

flake.nix
```diff
@@ -191,15 +197,19 @@
...
+        packages = inputs.std.harvest inputs.self [ "automation" "ciJobs" ];
       }
       {
         packages = inputs.std.harvest inputs.self [ "toolchain" "packages" ];
       }
       {
         packages = inputs.std.harvest inputs.self [ "toolchain" "scripts" ];
-      };
+      }
+      (inputs.tullia.fromStd {
```
michaelpj, line 209: Please write down whatever you found out about the infinite recursion. That's useful info, otherwise we'll inevitably try and merge this back in at some point in the future and be very confused!

flake.nix
```diff
@@ -111,7 +115,7 @@
         #
         # In this repository we have four cells:
         #   automation
-        #     Hydra jobsets and GHA tasks
+        #     Hydra jobsets, Cicero tasks, and GHA tasks
```
michaelpj, line 118: Should be in a new `cloud` section now

__std__/cells/cloud/pipelines.nix
```diff
@@ -0,0 +1,50 @@
...
+      imports = [ common flakeOutputTask ];
+
+      memory = 1024 * 8;
+      nomad.resources.cpu = 10000;
+    })
+    (flakeOutputTasks [ system "ciJobs" ] { outputs.${system}.ciJobs = ciJobs; });
+
+  ciTasksSeq = taskSequence "ci/" ciTasks (__attrNames ciTasks);
+in
+if system == "x86_64-linux" then
```
michaelpj, line 41: Probably needs a big comment explaining that we don't support darwin yet but we will (?)

michaelpj: What happens if we enable the darwin tasks? We'll just get a task that always fails? In fact, how does cicero know what kind of system to run a job on?

__std__/cells/cloud/pipelines.nix
```diff
@@ -0,0 +1,50 @@
...
+  ciTasks // # for running tasks separately
+  ciTasksSeq // # for running in an arbitrary sequence
+  {
+    "plutus/ci" = { lib, ... }: {
```
michaelpj, line 45: AIUI, if we _do_ enable this for darwin, we'd get two tasks called `plutus/ci`, and the _subjobs_ would have the system included? So not sure what we actually want to do there re naming.

michaelpj (approved):

__std__/cells/cloud/pipelines.nix
dermetfan, line 45: This is the plutus flake so why put "plutus" in the task's name?
```suggestion
    ci = { lib, ... }: {
```

__std__/cells/cloud/pipelines.nix
dermetfan: I also wonder — this should not be needed. Cicero looks at `.#cicero.${builtins.currentSystem}`.
```suggestion
```

Pacman99: No it is not needed, I just forgot to remove the line after some debugging

Pacman99:
Just addressed all the comments. I believe I fixed the cicero error, but I'll have to wait for the action to run to be sure of that.

For future reference, the solution to the infinite recursion issue was what @zeme-iohk suggested - to move actions and pipelines to a separate cloud cell.

__std__/cells/cloud/pipelines.nix
```diff
@@ -0,0 +1,60 @@
+{ cell
+, inputs
+}:
+/*
+  This code needs to be separate from any of the cells it uses
+  to prevent infinite recursion. Specifically the `flakeOutputTasks`
+  call triggers the error when called with any attribute that relies
+  on the cell that the pipelines are defined in.
+  That is why this is currently located in a separate cloud cell
+  and not the automations cell.
```
zeme-iohk, line 10: nit: automations -> automation

flake.nix
```diff
@@ -145,8 +152,8 @@
         #     Everything that is not a derivation goes here
         #     Includes functions, attrsets and simple literal values shared across cells
         #     These are not exposed to the flake
-        #   hydraJobs :: installables
-        #     Jobsets for our Hydra CI
+        #   ciJobs :: installables
+        #     Jobsets for our Hydra and Cicero CI
```
zeme-iohk, line 156: - We have five cell blocks -> we have eight cell blocks (line 140)
- Please add a 1-liner for the `pipelines.nix` and `actions.nix` cell blocks.

zeme-iohk:
Looks great 👍 
I just left a couple small comments

__std__/cells/cloud/pipelines.nix
Pacman99: @dermetfan nvm it looks like this is still needed, I'm not sure why but this invocation failed without it: https://cicero.ci.iog.io/invocation/a1be61dd-3452-4316-9e14-05748be735e4

Maybe I'm exporting the tasks wrong and darwin tasks are being lumped with linux tasks?

__std__/cells/cloud/pipelines.nix
Pacman99: I think it can still be considered one task just with two supported systems.

__std__/cells/cloud/pipelines.nix
Pacman99: I don't think this is a blocker, since cicero doesn't have darwin support yet

michaelpj: I'd still like this to have a comment, even if it's "we need this and we're not sure why"

Pacman99: oh sorry i completely missed your first message in this thread, ill add the comment in my next cicero PR

PR #4885, opened 2022-10-06
PLT-941: Plutus release process

zliu41:
This is a much stripped down version of the draft in the Google Doc, suitable to be published on GH.


doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+* plutus-core:pir
+* plutus-core:plc
+* plutus-core:uplc
+
+.Dependencies among Plutus packages
+image::plutus-packages.png[Plutus packages,500]
+
+== Version Scheme
+
+All above packages are versioned and released in sync. That is, whenever one package releases version X, so do all other packages. This greatly simplifies the release process compared to versioning and releasing the pacakges independently.
```
koslambrou, line 25: pacakges -> packages

koslambrou, line 25: pacakges -> packages

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+- A minor release (e.g., 3.5.2 -> 3.6.0) may contain changes such as new features, which technically can be breaking, but are unlikely to cause breakage in practice, and if it does, the fix should be straightforward (e.g., by renaming).
++
+A minor release is also allowed to change the observable behaviors of functions that users should not rely on. For example, if a function returns a list of things and makes no promise on the order of elements in the returned list, then a minor release may change the order.
+- A patch release (e.g., 3.5.2 -> 3.5.3) can only contain such things as simple bug fixes, performance improvements, and documentation updates.
+
+=== Version Qualifiers
+
+Version qualifiers, also known as pre-release versions, usually refer to alpha, beta, and release candidate. Typically, alpha and beta are pre-feature complete, while release candidates are post-feature complete.
+
+We do not tag or publish alpha or beta verions for libraries. Libraries are published on Hackage, and Hackage does not support version qualifiers. It is also uncommon for a library to publish pre-release versions (unless it is new or a complete rewrite).
```
koslambrou, line 39: > Libraries are published on Hackage, and Hackage does not support version qualifiers

Maybe not an argument, since you say below that we _do_ release candidates

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+== Release Process
+
+=== Major and Minor Releases
+
+Suppose we are releasing version `x.y.0`.
+
+1. Update version numbers in cabal files to `x.y.0` (if it is not already the case).
+- Version bounds for packages in the same repo should be set to `==x.y.*`.
+2. Make sure the change log is up to date.
+3. Tag `x.y.0-rc1` on master.
```
koslambrou, line 65: I would add a comment saying: "we don't create a release branch" just to clarify.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+4. If no release blocking issue is found, do a formal release and upload the packages to Hackage.
+- If issues are found, fix them on master, backport the fixes to `release/x.y`, and go to step 3.
+
+== Backwards Compatibility
+
+Plutus releases should meet two backwards compatibility criteria:
+
+1. *Each Plutus release should build with the latest version of Cardano API.*
++
+This makes it possible for Plutus Tools to upgrade to the new Plutus release, without needing a new Cardano API release.
```
koslambrou, line 95: I would change "Plutus Tools" to any downstream users.

Also, I would be good to know what kind of changes downstream users can use without an new Cardano API release. Most the time, the plutus version is heavily tighted to a particular Cardano API version, and we can't just upgrade plutus and expect our stuff to work

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+We do tag release candidates, e.g,. 3.5.2-rc1, because additional non-CI tests and checks may need to be carried out.
+These tests and checks either are too expensive to run in CI, or require manual steps (e.g., security audit), but are prerequisites of making a release.
+Tagging release candidates makes it easy to identify the correct commit to run these tests and checks against.
+
+For executables, we may publish alpha, beta or release candidates at our discretion.
+
+== Release Frequency
+
+We will begin by making a new major or minor release every 4 weeks, and evaluate and adjust the frequency in the future.
+Ad-hoc releases can be made upon request.
```
effectfully, line 51: Oh, interesting.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+== Release Process
+
+=== Major and Minor Releases
+
+Suppose we are releasing version `x.y.0`.
+
+1. Update version numbers in cabal files to `x.y.0` (if it is not already the case).
+- Version bounds for packages in the same repo should be set to `==x.y.*`.
+2. Make sure the change log is up to date.
```
effectfully, line 64: How serious are we about that? If serious, we need to make it a policy that every change that isn't trivial goes into the changelog. @asutherlandus @michaelpj also would like to hear your opinion.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+- This does not include tests that require building a new node, or integrating with other downstream projects.
+- In some cases we need input from domain experts and/or security audit (e.g., when adding a cryptography builtin). The right group of people should be involved and sign off on the tests and checks.
+5. If no release blocking issue is found, do a formal release and upload the packages to Hackage.
+- If issues are found, create a release branch `release/x.y`, fix the issues on master, backport the fixes to `release/x.y`, tag `x.y.0-rc2`, and go to step 4.
+- Why not just fix the issues on master and tag `x.y.0-rc2` from master?
+It is desirable to minimize the amount of change between `rc1` and `rc2`, because it may reduce the tests and checks that need to be performed against `rc2`.
+For instance, if `plutus-ledger-api` is the only package changed, there is no need to re-run tests on `plutus-core` or `plutus-tx`.
+Another example is if a security audit is done on `rc1`, and the changes in `rc2` does not modify the audited code, then the audit does not need to be re-done.
```
effectfully, line 75:
```suggestion
Another example is if a security audit is done on `rc1`, and the changes in `rc2` do not modify the audited code, then the audit does not need to be re-done.
```

effectfully (approved):
Very clear and sensible, thank you.

doc/RELEASE.adoc
joseph-fajen, line 39: typo: "verions" -> "versions"

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+- A patch release (e.g., 3.5.2 -> 3.5.3) can only contain such things as simple bug fixes, performance improvements, and documentation updates.
+
+=== Version Qualifiers
+
+Version qualifiers, also known as pre-release versions, usually refer to alpha, beta, and release candidate. Typically, alpha and beta are pre-feature complete, while release candidates are post-feature complete.
+
+We do not tag or publish alpha or beta verions for libraries. Libraries are published on Hackage, and Hackage does not support version qualifiers. It is also uncommon for a library to publish pre-release versions (unless it is new or a complete rewrite).
+However, we may annotate individual functions in a library as alpha or beta if we deem it appropriate.
+
+We do tag release candidates, e.g,. 3.5.2-rc1, because additional non-CI tests and checks may need to be carried out.
```
joseph-fajen, line 42: e.g,. -> e.g.,

doc/RELEASE.adoc
joseph-fajen, line 64: In this context I recommend using hyphens: "up-to-date"

joseph-fajen (approved):

doc/RELEASE.adoc
michaelpj, line 75: This section is mixing rationale with the process. I think we probably want to keep these "process descriptions" just telling you what to do, and move the rationale afterwards or in a footnote or something.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+1. Update version numbers in cabal files to `x.y.0` (if it is not already the case).
+- Version bounds for packages in the same repo should be set to `==x.y.*`.
+2. Make sure the change log is up to date.
+3. Tag `x.y.0-rc1` on master.
+4. Run all tests and checks.
+- This includes all tests that can be done with only the Plutus repo, such as unit tests, property-based tests, conformance tests and nightly tests.
+- This does not include tests that require building a new node, or integrating with other downstream projects.
+- In some cases we need input from domain experts and/or security audit (e.g., when adding a cryptography builtin). The right group of people should be involved and sign off on the tests and checks.
+5. If no release blocking issue is found, do a formal release and upload the packages to Hackage.
```
michaelpj, line 70: what's a "formal release"? I am a dumb reader, I thought this section was telling me how to do a release, but now I need to know how to do a formal release already?

michaelpj: Presumably it includes "tag the version as x.y.0", but that's not mentioned anywhere.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+- If issues are found, create a release branch `release/x.y`, fix the issues on master, backport the fixes to `release/x.y`, tag `x.y.0-rc2`, and go to step 4.
+- Why not just fix the issues on master and tag `x.y.0-rc2` from master?
+It is desirable to minimize the amount of change between `rc1` and `rc2`, because it may reduce the tests and checks that need to be performed against `rc2`.
+For instance, if `plutus-ledger-api` is the only package changed, there is no need to re-run tests on `plutus-core` or `plutus-tx`.
+Another example is if a security audit is done on `rc1`, and the changes in `rc2` does not modify the audited code, then the audit does not need to be re-done.
+
+6. After the release, update version numbers in cabal files to `x.y+1.0`.
+- This may not be the actual version of the next release - it could be `x+1.0.0`. We don't need to be precise here; we just need a version number higher than the latest release.
```
michaelpj, line 78: ... and it can never be a patch version, since patch versions are always and only made from release branches on-demand?

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+- This does not include tests that require building a new node, or integrating with other downstream projects.
+- In some cases we need input from domain experts and/or security audit (e.g., when adding a cryptography builtin). The right group of people should be involved and sign off on the tests and checks.
+5. If no release blocking issue is found, do a formal release and upload the packages to Hackage.
+- If issues are found, create a release branch `release/x.y`, fix the issues on master, backport the fixes to `release/x.y`, tag `x.y.0-rc2`, and go to step 4.
+- Why not just fix the issues on master and tag `x.y.0-rc2` from master?
+It is desirable to minimize the amount of change between `rc1` and `rc2`, because it may reduce the tests and checks that need to be performed against `rc2`.
+For instance, if `plutus-ledger-api` is the only package changed, there is no need to re-run tests on `plutus-core` or `plutus-tx`.
+Another example is if a security audit is done on `rc1`, and the changes in `rc2` does not modify the audited code, then the audit does not need to be re-done.
+
+6. After the release, update version numbers in cabal files to `x.y+1.0`.
```
michaelpj, line 77: This is a tricky corner. AIUI, there are two main strategies for bumping version numbers:
1. Bump them when you make a change that requires a bump. So if you do something that requires a new major version, you bump to the next major version. But this requires you to remember whether you've done a release since the last time you bumped - if you have you don't bump.
2. Bump them when you want to do a release. This requires you to go back through your changes and decide what kind of bump is merited.

It seems like you're suggesting a kind of hybrid of 1 and 2: the first point of this process made me think we're doing 2, but this makes me think that we're doing 1. Maybe worth discussing on its own! Personally I think 1 is probably fine, it's not that hard to remember what the last released version was so long as you only have one!

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+- This may not be the actual version of the next release - it could be `x+1.0.0`. We don't need to be precise here; we just need a version number higher than the latest release.
+
+=== Patch Releases
+
+1. If it is the first time we are making a patch release for version `x.y` (i.e., `x.y.0` -> `x.y.1`), a release branch may not exist for `x.y`. If so, create branch `release/x.y` from the `x.y.0` tag.
+- We create release branches lazily, because we do not expect to make many patch releases.
```
michaelpj, line 83: :+1:

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+2. Backport the needed fixes from master to `release/x.y`.
+3. Run all tests and checks.
+4. If no release blocking issue is found, do a formal release and upload the packages to Hackage.
```
michaelpj, line 86: again that "formal release" phrase...

doc/RELEASE.adoc
michaelpj, line 39: They're not published on Hackage, they're going to be published on CHaP.

doc/RELEASE.adoc
michaelpj: Hackage -> CHaP

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+6. After the release, update version numbers in cabal files to `x.y+1.0`.
+- This may not be the actual version of the next release - it could be `x+1.0.0`. We don't need to be precise here; we just need a version number higher than the latest release.
+
+=== Patch Releases
+
+1. If it is the first time we are making a patch release for version `x.y` (i.e., `x.y.0` -> `x.y.1`), a release branch may not exist for `x.y`. If so, create branch `release/x.y` from the `x.y.0` tag.
+- We create release branches lazily, because we do not expect to make many patch releases.
+2. Backport the needed fixes from master to `release/x.y`.
+3. Run all tests and checks.
```
michaelpj, line 85: This is inconsistent with before. Before we tagged `x.y.0-rc0` before doing this. Should we do that here? `x.y.z-rc0`? I think the two processes should be the same!

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+4. If no release blocking issue is found, do a formal release and upload the packages to Hackage.
+- If issues are found, fix them on master, backport the fixes to `release/x.y`, and go to step 3.
+
+== Backwards Compatibility
+
+Plutus releases should meet two backwards compatibility criteria:
+
+1. *Each Plutus release should build with the latest version of Cardano API.*
```
michaelpj, line 93: `cardano-api` doesn't depend on `plutus` directly (or shouldn't). Do you mean the _ledger_ should build with the released version?

michaelpj: Or no, I think you do mean this: you mean that _everything downstream_ should build, all the way to `cardano-api`.

michaelpj: We need to automate this check! Probably not too hard with CHaP.

michaelpj: Do we want this policy even for new _major_ versions? I buy it for minor versions, but kind of the point of major versions is that you break something, no?

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
++
+This makes it possible for Plutus Tools to upgrade to the new Plutus release, without needing a new Cardano API release.
++
+To do so, rather than making changes to the Plutus API that breaks Cardano API (e.g., changing the type of a function), we should temporarily keep both the old Plutus API and the new Plutus API, until a new Cardano API version is released that no longer depends on the old Plutus API.
```
michaelpj, line 97: This is a big change, but probably good.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
++
+This requirement will likely go away once Cardano API starts to make more frequent releases.
+
+2. *Plutus language behaviors should be preserved.*
++
+When we start developing a new Plutus language version (e.g., Plutus V3), initially the language version is considered "unreleased", and should be guarded behind a flag to make it unusable.
```
michaelpj, line 103:
```suggestion
When we start developing a new Plutus ledger language version (e.g., Plutus V3), initially the language version is considered "unreleased", and should be guarded behind a flag to make it unusable.
```

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+1. *Each Plutus release should build with the latest version of Cardano API.*
++
+This makes it possible for Plutus Tools to upgrade to the new Plutus release, without needing a new Cardano API release.
++
+To do so, rather than making changes to the Plutus API that breaks Cardano API (e.g., changing the type of a function), we should temporarily keep both the old Plutus API and the new Plutus API, until a new Cardano API version is released that no longer depends on the old Plutus API.
++
+This requirement will likely go away once Cardano API starts to make more frequent releases.
+
+2. *Plutus language behaviors should be preserved.*
```
michaelpj, line 101:
```suggestion
2. *Plutus ledger language behaviors should be preserved.*
```

doc/RELEASE.adoc
michaelpj: Interesting, we should work out how we're going to do this technically. Since I think we _do_ want to test this, and we maybe even want the ledger to be able to start putting together the scaffolding for the new version also. So if anything I think the flag should be in the _ledger_.

doc/RELEASE.adoc
michaelpj, line 101: I think that this section is not really about _releases_ per se. It's about how we manage new ledger language versions, which is kind of orthogonal. "New features should be guarded by feature flags" is not a _release_ consideration.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+Once it is ready to be used, we remove the flag in a Plutus release, making it usable.
+After this point, no more behavior changes (that is, the kind of changes marked "LV" in https://cips.cardano.org/cips/cip35[CIP-35]) are allowed.
++
+If a change does not require LV, but requires HF per CIP-35, then it is allowed, but it must be guarded behind an appropriate protocol version condition.
++
+Note that Plutus language versions are unrelated to package versions used for identifying releases.
+
+== Downstream Integration
+
+Some Plutus features and changes require integration testing on devnets and testnets. Such tests are typically performed by the node QA team or the ecosystem collaborators. Test scenarios related to the Plutus changes should be reviewed and signed off by the Plutus team.
```
michaelpj, line 114: At the moment this is dangling: it doesn't fit into the process. I think that's just true: this is part of the _node_ release process that we should be involved in. Not sure how to convey that.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+As stated before, if a bug is found at the integration test stage that requires fixes in Plutus, we should make a patch release using the process described above, as opposed to a major or a minor one. This is because integration tests are most likely done against a Cardano node release branch, and the only changes that should be merged into a release branch are fixes to release blocking bugs.
```
michaelpj, line 116:
```suggestion
As stated before, if a bug is found during the node release process that requires fixes in Plutus, we should make a patch release using the process described above, as opposed to a major or a minor one. That is because we want to minimize the changes to downstream, since this will occur when the Cardano node is already progressing through its release process.
```

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+The Plutus team should also reserve some additional capacity in supporting the QA team, ecosystem collaborators, and other open source issues and requests.
```
michaelpj, line 118: Again, seems out of scope. This is an Andrew problem not a release process concern.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+Lastly, we should publish a mapping from each Cardano node version to the Plutus version used by the node version. This helps identify which Plutus changes require a hard fork, and it also makes it easier for downstream projects to determine which Plutus version to depend on.
```
michaelpj, line 120: This should be an automated process that the node can do, right? Is this our problem?

michaelpj (commented):
Sorry, lots of comments :sweat_smile:

doc/RELEASE.adoc
michaelpj: Although maybe we want to say "new Plutus ledger language versions must always be relased in a _major_ version"?

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
+= Plutus Release Process
```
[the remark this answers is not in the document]
zliu41: I'd say definitely yes - I wish every open source project had a documented release process, so I don't need to guess things!

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+* plutus-ledger-api:plutus-ledger-api
+* plutus-tx:plutus-tx
+* plutus-tx-plugin:plutus-tx-plugin
+- Executables:
+* plutus-core:pir
+* plutus-core:plc
+* plutus-core:uplc
+
+.Dependencies among Plutus packages
+image::plutus-packages.png[Plutus packages,500]
```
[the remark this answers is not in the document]
zliu41: Ah, sounds interesting. I'll take a look at this.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+== Version Scheme
+
+All above packages are versioned and released in sync. That is, whenever one package releases version X, so do all other packages. This greatly simplifies the release process compared to versioning and releasing the pacakges independently.
+
+We follow https://semver.org[semantic versioning] for the packages, where version numbers have a `major.minor.patch` pattern with the following semantics:
```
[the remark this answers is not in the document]
zliu41: What we discussed before is that PVP is better for libraries and SemVer is better for executables (especially ones whose users include non-Haskell developers).

I don't think we need to spend too much time on easily reversible decisions like this. I'm happy to use PVP and see how it works out.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
++
+A minor release is also allowed to change the observable behaviors of functions that users should not rely on. For example, if a function returns a list of things and makes no promise on the order of elements in the returned list, then a minor release may change the order.
+- A patch release (e.g., 3.5.2 -> 3.5.3) can only contain such things as simple bug fixes, performance improvements, and documentation updates.
+
+=== Version Qualifiers
+
+Version qualifiers, also known as pre-release versions, usually refer to alpha, beta, and release candidate. Typically, alpha and beta are pre-feature complete, while release candidates are post-feature complete.
+
+We do not tag or publish alpha or beta verions for libraries. Libraries are published on Hackage, and Hackage does not support version qualifiers. It is also uncommon for a library to publish pre-release versions (unless it is new or a complete rewrite).
+However, we may annotate individual functions in a library as alpha or beta if we deem it appropriate.
```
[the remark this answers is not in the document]
zliu41: Unfortunately Haskell doesn't have a first-class way to do this, but I'm thinking either in Haddock or in module names.

doc/RELEASE.adoc
zliu41: For a library as upstream as Plutus, I think it is critical. Otherwise everyone downstream will need to slack us and ask for the right version to use!

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+=== Major and Minor Releases
+
+Suppose we are releasing version `x.y.0`.
+
+1. Update version numbers in cabal files to `x.y.0` (if it is not already the case).
+- Version bounds for packages in the same repo should be set to `==x.y.*`.
+2. Make sure the change log is up to date.
+3. Tag `x.y.0-rc1` on master.
+4. Run all tests and checks.
```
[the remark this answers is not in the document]
zliu41: It should be after tagging RC. If all tests and checks are already done, then we are ready to tag the actual release, not the RC.

doc/RELEASE.adoc
zliu41: Yeah, I agree we should tag a release candidate in this case too.

doc/RELEASE.adoc
zliu41: Yeah.. and minor versions shouldn't need this policy, because it's not supposed to be backwards compatible. So perhaps we don't need this policy at all!

zliu41: I rewrote this section in https://github.com/input-output-hk/plutus/pull/4885/commits/1c9f6a0431b35ab304ebddddd7e52d0aa4e5e645.

doc/RELEASE.adoc
zliu41: I rewrote this section in https://github.com/input-output-hk/plutus/pull/4885/commits/1c9f6a0431b35ab304ebddddd7e52d0aa4e5e645. A good example is if we make some improvements to `plutus-tx-plugin` and releases a new version, but it can't be used by Plutus Tools because the corresponding `plutus-core` doesn't build with Cardano API.

doc/RELEASE.adoc
zliu41: Agreed. Since the node release process hasn't been written, I just want to call it out. I'm not sure we've been signing off test scenarios related to plutus in `cardano-node-tests`, and we should.

doc/RELEASE.adoc
zliu41: Deleted.

doc/RELEASE.adoc
zliu41: Yeah, I'll delete this.
Since there's no node release process right now, I have a tendency to include the part of node release process that's relevant to plutus :smile:

doc/RELEASE.adoc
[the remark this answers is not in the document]
zliu41: Changed to PVP.

doc/RELEASE.adoc
zliu41: Correct.

doc/RELEASE.adoc
zliu41: I think 1 can be cumbersome. If we haven't bumped the major version, then every PR needs to be evaluated to see whether it should cause the major version to be bumped. I think we can go with 2, but instead of going back through the changes, just bump the major version (second PVP component, unless the change is big enough to justify bumping the first), unless we are certain that we only need to bump the minor version. What do you think?

doc/RELEASE.adoc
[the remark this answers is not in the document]
zliu41: I haven't figured out how to use it here - probably need some extensions somehow installed? I'll just remove it. It's not that important.

doc/RELEASE.adoc
zliu41: > I think that this section is not really about releases per se.

It is true that Plutus language version is independent of package version. But it is relevant to mention it here because before a language version is released, we can make changes to the behavior, but the behavior is "frozen" after a specific release that "publishes" the language version. This doc seems a good place to mention this.

> new Plutus ledger language versions must always be relased in a major version

I don't think so. New features usually only need bumping the minor version unless it is backwards incompatible.

doc/RELEASE.adoc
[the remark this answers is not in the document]
michaelpj: Could be fun to have graphs in future though :)

doc/RELEASE.adoc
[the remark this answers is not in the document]
michaelpj: I think this is maybe a situation of "let's not say we do it until we actually do it". If we start doing this then we can say so, but if we're just going to write this but never do it then it's pointless.

doc/RELEASE.adoc
```diff
@@ -0,0 +1,120 @@
...
+In all other cases, we always start a new major or minor release from master.
+
+== Release Process
+
+=== Major and Minor Releases
+
+Suppose we are releasing version `x.y.0`.
+
+1. Update version numbers in cabal files to `x.y.0` (if it is not already the case).
+- Version bounds for packages in the same repo should be set to `==x.y.*`.
```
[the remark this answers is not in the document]
michaelpj: Unsure - worst case we write a script that does some `sed`-ing. I think it wouldn't be too bad.

doc/RELEASE.adoc
michaelpj: I'm happy with either tbh! I think it would be nice to call this out as it's own thing, since I predict "when do I bump the versions?" to be a regular question!

doc/RELEASE.adoc
michaelpj: > But it is relevant to mention it here because before a language version is released, we can make changes to the behavior, but the behavior is "frozen" after a specific release that "publishes" the language version.

Right, but that's the same as other functionality: you can change it as you like before it's released, and after that you may need to do something else (e.g. release a major version). Language versions are even more finicky than that, but I think that's a language-version topic rather than a release topic per se. Anyway, just trying to avoid this doc becoming _too_ gigantic :D

doc/RELEASE.adoc
michaelpj: I guess I just lean towards "help the node people get stuff set up so their release process does what we want" rather than "optimistically declare that we're going to do it all" :sweat_smile:

doc/RELEASE.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+It is a good idea to avoid breaking the latest version of Cardano API in a new Plutus release.
+This makes it easy for downstream projects to update Plutus version without needing a new Cardano API release.
+
+For example, suppose we make some improvements to `plutus-tx-plugin` and make a new major release.
+Since all packages are released in sync, we also make a new major release for `plutus-core`.
+Although it is a major release, we should avoid making changes that is incompatible with the latest version of Cardano API.
+Otherwise, downstream projects such as Plutus Tools won't be able to use the new Plutus version and take advantage of the plugin improvements, until a new Cardano API version is published.
+
+To do so, rather than making changes to the Plutus API that breaks Cardano API (e.g., changing the type of a function), we can temporarily keep both the old Plutus API and the new Plutus API, until a new Cardano API version is released that no longer depends on the old Plutus API.
```
michaelpj, line 110: I'm still a bit confused by this. So when _can_ we break `cardano-api`? Apparently not on a major version so... never?

doc/RELEASE.adoc
```diff
@@ -0,0 +1,132 @@
...
+Version qualifiers, also known as pre-release versions, usually refer to alpha, beta, and release candidate.
+Typically, alpha and beta are pre-feature complete, while release candidates are post-feature complete.
+
+We do not tag or publish alpha or beta versions for libraries.
+Depending on an alpha or beta version of a library requires `source-repository-package`, which is quite inconvenient.
+It is also uncommon for a library to publish pre-release versions (unless it is new or a complete rewrite).
+However, we may annotate individual functions in a library as alpha or beta if we deem it appropriate.
+
+We do tag release candidates, e.g., 3.5.2.1-rc1, because additional non-CI tests and checks may need to be carried out.
+These tests and checks either are too expensive to run in CI, or require manual steps (e.g., security audit), but are prerequisites of making a release.
```
michaelpj, line 46: Another place where we could just reference the Release QA process. Generally I'm pretty in favour of having dedicated sections for stuff like this and just cross-referencing a lot.

michaelpj (approved):
Looking good!

doc/RELEASE.adoc
```diff
@@ -0,0 +1,132 @@
...
+1. Update version numbers in cabal files to `x.y.z.0` (if it is not already the case).
+- Version bounds for packages in the same repo should be set to `==x.y.z.*`.
+2. Make sure the change log is up-to-date.
+3. Tag `x.y.z.0-rc1` on master.
+- There's no need to create a release branch at this point.
+A release branch is only needed if bugs are found in `x.y.z.0-rc1` and we need to backport a fix from master.
+4. Run all tests and checks.
+- This includes all tests that can be done with only the Plutus repo, such as unit tests, property-based tests, conformance tests and nightly tests.
+- This does not include tests that require building a new node, or integrating with other downstream projects.
+- In some cases we need input from domain experts and/or security audit (e.g., when adding a cryptography builtin).
```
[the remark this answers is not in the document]
michaelpj: This is kind of important actually: at the moment this document isn't actionable, because it's not clear what you _actually_ do here. Ideally we'd have a checklist of tests or something.

doc/RELEASE.adoc
zliu41: The idea is not to break the _latest_ version of `cardano-api`. I'll clarify that this is a suggestion rather than a hard rule.

doc/RELEASE.adoc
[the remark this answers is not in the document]
zliu41: Great suggestion.

michaelpj:
This wasn't added to the TOC anywhere so it's not actually part of the site atm @zliu41

PR #4887, opened 2022-10-06
GHC 9.2.4

newhoggy:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


cabal.project
```diff
@@ -73,4 +75,13 @@ source-repository-package
...
+  tag: 3825d3abf75f83f406c1f7161883c438dac7277d
+
+-- Needed for ghc-9.2.4
+source-repository-package
+  type: git
+  location: https://github.com/input-output-hk/int-cast
+  tag: 20cc3a95383fc8d1c2ea3b3d691700036cecdeac
+
+constraints:
+  dependent-sum >= 0.7.1.0,
```
michaelpj, line 87: This is annoying. It would be much better if we could put this in a cabal file somewhere, but I'm not sure we can without adding a dead package dependency...

cabal.project
```diff
@@ -73,4 +75,13 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/Win32-network
-  tag: 3825d3abf75f83f406c1f7161883c438dac7277d
\ No newline at end of file
+  tag: 3825d3abf75f83f406c1f7161883c438dac7277d
+
+-- Needed for ghc-9.2.4
```
michaelpj, line 80: Hmm. Should we push a patched version to CHaP? upstream looks fairly dead. Perhaps we can remove the dependency, but it is useful

cabal.project
newhoggy: https://github.com/obsidiansystems/dependent-sum/issues/72

PR #4890, opened 2022-10-07
Remove obsolete CI scripts and shorten job names

zeme-iohk:
I think that we can do without all the `*-checker` scripts.
Originally they came as `default.nix:tests`
They are needed to make sure we don't commit unformatted files to master, or rather files that don't pass the `pre-commit` check.
However, we have a derivation named `pre-commit-check.nix`, which, when built, takes care of detecting all such errors. 
So, should it not be sufficient to `nix build .#pre-commit-check` as a GHA, and - maybe redundantly - include `pre-commit-check` also as job in Hydra?
The only difference is see is that these `*-checker` derivations have more informative error messages.
At the same time, the error messages provided by `pre-commit` itself are familiar and should be sufficient.

In addition this PR tries to shorten hydra job names.


.github/workflows/test.yml
```diff
@@ -20,3 +20,12 @@ jobs:
       - uses: actions/checkout@v3
       - uses: nixbuild/nix-quick-install-action@v17
       - run: nix-build default.nix -A tests --arg supportedSystems '[ builtins.currentSystem ]' --restrict-eval -I . --allowed-uris 'https://github.com/NixOS/nixpkgs https://github.com/input-output-hk https://github.com/NixOS/nixpkgs-channels' --option trusted-public-keys "hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= iohk.cachix.org-1:DpRUyj7h7V830dp/i6Nti+NEO2/nhblbov/8MW7Rqoo= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=" --option substituters "https://hydra.iohk.io https://iohk.cachix.org https://cache.nixos.org/"
+  pre-commit-check:
```
zeme-iohk, line 23: This would replace the `nix-tests` action above

__std__/nix/automation/hydraJobs.nix
```diff
@@ -25,37 +31,33 @@ let
...
+    if system == "x86_64-linux" then
+      { native = native-plutus-jobs; windows = windows-plutus-jobs; } // other-jobs
+    else if system == "x86_64-darwin" then
+      { native = native-plutus-jobs; } // other-jobs
+    else
+      { };
 
   # Hydra doesn't like these attributes hanging around in "jobsets": it thinks they're jobs!
-  filtered-jobs = lib.filterAttrsRecursive (n: _: n != "recurseForDerivations") jobs;
+  # filtered-jobs = lib.filterAttrsRecursive (n: _: n != "recurseForDerivations") jobs;
```
zeme-iohk, line 50: just an attempt at seeing if we still need this.

michaelpj:
Does `pre-commit-check` run on all the files in the repository? 

> In addition this PR tries to shorten hydra job names.

I strongly suggest only doing one thing at a time. It _really_ makes things easier.

flake.nix
```diff
@@ -299,6 +298,5 @@
       "hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ="
     ];
     allow-import-from-derivation = true;
-    accept-flake-config = true;
```
zeme-iohk, line 302: We don't want this in general as it's unsafe

michaelpj:
> Does `pre-commit-check` run on all the files in the repository? 

Answered my own question: it does seem to!

zeme-iohk:
> Does `pre-commit-check` run on all the files in the repository?
It runs on all relevant files (it skips gitignored files), but it runs regardless of whether they've been changed in the last commit. It's slightly inefficient. I can look into a way to avoid that 

> I strongly suggest only doing one thing at a time. It _really_ makes things easier.
Yes I'll split this PR into two

.github/workflows/test.yml
zeme-iohk: So, `nix build .#pre-commit-check` implies `pre-commit run --all-files`, no way around it. 
It respects .gitignore, but it checks all files regardless of whether they've been modified by this commit. 
It's disadvantage is that it's slightly inefficient, but still lightning fast in practice.
It's advantage is that it will catch weird edge cases like it did in https://github.com/input-output-hk/plutus/pull/4886

The alternative is to replace with 
`nix develop --command pre-commit`
Which will work as expected. 

Which to pick?

michaelpj:
No, checking all files is good! That's what we want for this test, I think.

flake.nix
```diff
@@ -195,7 +195,6 @@
       }
       {
         packages = inputs.std.harvest inputs.self [ "toolchain" "scripts" ];
-        library = inputs.std.harvest inputs.self [ "toolchain" "library" ];
```
michaelpj, line 198: is this backwards? you got rid of lots of stuff in `scripts` not `library`?

zeme-iohk: this was temporary and committed by mistake, we are not exposing `library` to the flake, though in a later PR I'll make a case for why we might want to do that

.github/workflows/test.yml
```diff
@@ -20,3 +20,12 @@ jobs:
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: nixbuild/nix-quick-install-action@v17
+      - run: nix build .#pre-commit-check --extra-experimental-features "nix-command flakes" --allowed-uris 'https://github.com/NixOS/nixpkgs https://github.com/input-output-hk https://github.com/NixOS/nixpkgs-channels' --option trusted-public-keys "hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= iohk.cachix.org-1:DpRUyj7h7V830dp/i6Nti+NEO2/nhblbov/8MW7Rqoo= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=" --option substituters "https://hydra.iohk.io https://iohk.cachix.org https://cache.nixos.org/"
```
michaelpj, line 31: You don't need to do all this. I think I failed to get this across, but there's a CLI option to say you want to accept the config, which is appropriate for running things in CI. See https://github.com/input-output-hk/haskell.nix/blob/master/.buildkite/pipeline.yml#L149

michaelpj (changes requested):
Couple of significant comments but then I think this is good to go.

zeme-iohk:
Still needs `--extra-experimental-features "nix-command flakes"` apparently...
Adding `extra-experimental-features = [ "nix-command" "flakes" ];` to `nixConfig` in flake.nix doesn't do it.

.github/workflows/test.yml
zeme-iohk: 🙏 💯

michaelpj (approved):

PR #4892, opened 2022-10-07
Make Hydra Job Names Shorter

zeme-iohk:

PR #4893, opened 2022-10-09
Nix Code Standardisation - Part 6 - Almost Done

zeme-iohk:
With this PR the new nix setup goes live.

Users of `nix-shell` will receive a deprecation message and will be prompted to switch to `nix develop`
However `nix-shell` will be maintained parallel with `nix develop` until the end of the month.

This is work in progress and can't be merged until these are resolved:
https://github.com/input-output-hk/haskell.nix/issues/1738 
https://github.com/input-output-hk/plutus/pull/4892 
https://github.com/input-output-hk/plutus/pull/4884


scripts/ci-plutus-benchmark.sh
```diff
@@ -78,6 +78,8 @@ echo -e "</details>"
 } > bench-compare-result.log
 
 nix-shell -p jq --run "jq -Rs '.' bench-compare-result.log >bench-compare.json"
+# TODO(std) check that this works
```
zeme-iohk, line 81: we should already be inside a shell with `jq` at this point...

flake.lock
```diff
@@ -129,6 +129,21 @@
         "type": "github"
```
zeme-iohk, line 129: Changes to `flake.lock` are caused by the fact that `std` was updated and that the hard-coded revision hash was removed from the flake inputs urls.

flake.nix
```diff
@@ -197,98 +197,6 @@
         packages = inputs.std.harvest inputs.self [ "toolchain" "scripts" ];
       };
 
-  # TODO(std) move this part of the doc (which doesn't need to reference the code
```
zeme-iohk, line 200: This has been moved to `__std__/README.md`

shell.nix
```diff
@@ -107,6 +116,7 @@ haskell.project.shellFor {
 
   shellHook = ''
     ${pre-commit-check.shellHook}
+    ${deprecation-warning}
   '';
 
   # This is no longer set automatically as of more recent `haskell.nix` revisions,
```
michaelpj, line 122: Is this in our new shell?

shell.nix
```diff
@@ -107,6 +116,7 @@ haskell.project.shellFor {
 
   shellHook = ''
     ${pre-commit-check.shellHook}
```
michaelpj, line 118: Is this in the new devshell?

michaelpj (approved):
Looks fine! Is the idea to rename `__std__` to `nix` also? Or only do that once we finally EOL the existing setup?

shell.nix
zeme-iohk: Yes, it's the deprecation warning reminding the team to switch to `nix develop`

shell.nix
zeme-iohk: The pre-commit-check is in the new shell

zeme-iohk:
> Looks fine! Is the idea to rename __std__ to nix also? Or only do that once we finally EOL the existing setup?

The folder name `nix` is already taken by the old setup, so we have to keep __std__ until EOL

shell.nix
michaelpj: No, I'm commenting on the bit about the locales.

michaelpj:
If this makes `nix develop` the new "blessed way", then it should also update all the CONTRIBUTING docs to talk about it instead, also.

shell.nix
zeme-iohk: Ah, good catch, will look into it

zeme-iohk:
I updated CONTRIBUTING.adoc 
Also I run this to make sure that the shells are in parity 
```
# using nix-shell
dump-old-paths() {
  locale -a  
  which haskell-language-server 
  which cabal 
  which hlint  
  which ghc  
  which Agda  
}
dump-old-paths > old.txt 

# using: nix develop 
# needs redlink because the devshell adds another layer of indirection 
dump-new-paths() {
  locale -a
  readlink $(which haskell-language-server)
  readlink $(which cabal)
  readlink $(which hlint)
  readlink $(which ghc) 
  readlink $(which Agda)
}
dump-new-paths > new.txt 
diff old.txt new.txt
```
Agda is different simply because of the explicit index-state. Removing it yields the same derivation.
ghc is different because of this https://github.com/input-output-hk/plutus/blob/1a90cd16bf0e21fbe79720c04ffa2c95eb6aaee5/__std__/nix/plutus/library/plutus-project.nix#L109
Removing that line also yields the same ghc.

zeme-iohk:
I had added a hard-coded `index-state = X` to Agda in order to suppress a warning that said: `index-state not specified, using latest known X`.
If I set Agda's `index-state = library.cabal-project-index-state` (coming from plutus' cabal.project) the derivation output is the same.
I've also tried removing all other occurrences of `index-state = library.cabal-project-index-state` (they were used in cabal-install and cabal-fmt`) and no extra builds were triggered.
So maybe we can do without? And rely on the index-state provided by haskell.nix?
I pushed my changes here.

zeme-iohk:
Applied requested changes

PR #4899, opened 2022-10-11
plutus-core.cabal: Remove un-needed dependencies

erikd:


erikd:
Most importantly it removes an `inline-r` dependency for the `plutus-core` library (which is a dependency of `ledger`).

michaelpj:
I'm confused. You removed `inline-r` from the `generate-cost-model` executable, not the `plutus-core` library, which doesn't depend on it. And the dependency is not unused: the build of `generate-cost-model` is now failing.

erikd:
I have been messing with this so much I am confusing myself.

PR #4904, opened 2022-10-12
Add with-inline-r cabal flag

erikd:


angerman:
So this currently fails to build 
```
plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetingDefaults.hs:67:6: error:
    • Exception when trying to run compile-time code:
        /build/plutus-root-plutus-core-lib-plutus-core-root/plutus-core/../cost-model/data/cekMachineCosts.json: openBinaryFile: does not exist (No such file or directory)
      Code: readJSONFromFile DFP.cekMachineCostsFile
    • In the Template Haskell splice
        $$(readJSONFromFile DFP.cekMachineCostsFile)
      In the expression: $$(readJSONFromFile DFP.cekMachineCostsFile)
      In an equation for 'defaultCekMachineCosts':
          defaultCekMachineCosts
            = $$(readJSONFromFile DFP.cekMachineCostsFile)
   |
67 |   $$(readJSONFromFile DFP.cekMachineCostsFile)
   |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Now, why does this fail, because we now try to reach outside of the package to find some files. This can not work if we isolate package (e.g. upload them to hackage).

Plutus-core now has a data dependency on `cost-model`, we could just define the result of the expression as something that's imported from `cost-model`, but that will mean we bring in `inline-r` transitively into `plutus-core` again?

I would argue the _generated_ `.json` files must live in `plutus-core`, while the `generator` (which depends on inline-r) lives outside?

If `cekMachineCosts.json` is shared, but not generated, we'd need to move that into a `plutus-data` package, and have `plutus-core` and `cost-model` both depend on it?

erikd:
Thanks @angerman . Looks like I need a different approach

michaelpj:
I like this approach. Just make sure to set it to `true` in our `cabal.project`.

erikd:
@michaelpj if its true in `cabal.project` it will fail for me. Shouldn't it be in `cabal.project.local` ?

michaelpj:
We have multiple problems:
1. Make sure the dep doesn't pop up downstream, e.g. in the ledger. It shouldn't anyway, but it does because of the cabal bug. This fix solves that.
2. Don't require `R` to work on `plutus` itself. I do not want to solve this now: the status quo works fine for us, turning this off by default will be a regression for our team.

erikd:
For your team, all that is required is a one time addition of  a line or two to `cabal.project.local` .

michaelpj:
- Fixed the flag so it actually worked :sweat_smile: 
- Set it on by default in cabal.project
    - Tested that turning it off in cabal.project.local works, it does
- Added a comment explaining the situation.

michaelpj:
> For your team, all that is required is a one time addition of a line or two to cabal.project.local .

For people not on the team for whom it's a problem, all that is required is a one time addition of a line or two to cabal.project.local. #of-people-on-my-team >> #of-people-doing-drive-by-work-on-plutus-without-nix

But that reminds me that I also need a note in CONTRIBUTING about this.

PR #4905, opened 2022-10-12
Add LOCALE_ARCHIVE to plutus devshell

zeme-iohk:
Also adds R to the shell regardless of system


zeme-iohk:
@kwxm can you try this one please? You seem to be the only one with `glibc`

michaelpj (approved):

PR #4906, opened 2022-10-12
Add pkg-config to the shell

zeme-iohk:
After integrating CHaP, `nix develop` stopped working on Darwin. 
That is, running `cabal build all` would yield:
```
Error: cabal: Could not resolve dependencies:
[__0] trying: plutus-core-1.0.0.0 (user goal)
[__1] next goal: inline-r (dependency of plutus-core)
[__1] rejecting: inline-r-0.10.5, inline-r-0.10.4, inline-r-0.10.3,
inline-r-0.10.2, inline-r-0.10.1, inline-r-0.10, inline-r-0.9.2,
inline-r-0.9.1, inline-r-0.9.0.2, inline-r-0.9.0.1, inline-r-0.9.0.0,
inline-r-0.8.0.1, inline-r-0.8.0.0, inline-r-0.7.3.0, inline-r-0.7.2.0,
inline-r-0.7.1.2, inline-r-0.7.1.1, inline-r-0.7.1.0, inline-r-0.7.0.0
(conflict: pkg-config package libR>=3.0, not found in the pkg-config database)
[__1] fail (backjumping, conflict set: inline-r, plutus-core)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: inline-r, plutus-core
```
Turns out that when `cabal.project` contains no `source-repository-package` stanzas then haskell.nix does not include `pkg-config` in its devshell's `buildInputs`.

### DIFF
--- a/__std__/cells/plutus/devshells/plutus-shell.nix
+++ b/__std__/cells/plutus/devshells/plutus-shell.nix
@@ -49,6 +49,7 @@ inputs.std.lib.dev.mkShell {
     pkgs.awscli2
     pkgs.bzip2
     pkgs.cacert
+    pkgs.pkg-config # TODO(std) Keep an eye on https://github.com/input-output-hk/plutus/pull/4906
 
     pkgs.rPackages.plotly
     pkgs.R


michaelpj:
I guess this is being fixed by https://github.com/input-output-hk/haskell.nix/pull/1751? Please add a note so we know to remove it when that's merged.

jbgi:
No, https://github.com/input-output-hk/haskell.nix/pull/1751 was for another issue, that occurred for shell with cross-compilation support.

michaelpj:
Okay, could one of you make a `haskell.nix` issue for this? It seems like a bug in the devshell support: it should have the same things as the main `haskell.nix` shell, which seems to do the right thing.

jbgi:
No the main haskell.nix shell had the same issue. `pkg-config` was removed from it. (hence removed from `devshell` as well).

zeme-iohk:
Actually it's a little subtler no? 
- `nix develop .#x86_64-darwin.plutus.library.plutus-project.shell` does *not* provide pkg-config 
- `nix develop .#x86_64-darwin.plutus.library.plutus-project.devshell` does *not* provide pkg-config either 
- but [project.shellFor](https://github.com/input-output-hk/plutus/blob/511611ce9234d3f34f50c322a0888fc2786c17e8/shell.nix#L113) does

PR #4907, opened 2022-10-13
Merge doc and toolchain cells into plutus cell

zeme-iohk:
- Why merge doc into plutus?
We now have two shells: one for doc and one for haskell.
All that buys us is that the doc shell is very lightweight compared to the haskell one.
However, those who work only on docs (Joseph) are also wanting to learn haskell
And those who work on plutus/haskell often want to edit the docs.
So, having two shells is just awkward and useless.
Having one shell makes the nix code easier to maintain also.

- Why merge toolchain into plutus?
The idea was to identify shared nix code that could be reused in `plutus-apps`.
However the extent to which this is actually possibile is unknown until plutus-apps is also standardised.
Also, the two cells are now mutually recursive, due to `cabal-project-index-state.nix`.
Which hints at the fact that the `toolchain` is not as "general" and independent of plutus as we would like.
So, putting everything inside a single cell again makes the code shorter and easier to reason about by removing one layer of indirection that in practice doesn't offer anything.
Still, sharing nix code with plutus-apps will be possibile regardless: the `plutus` cell will be used instead of `toolchain`.

- Why rename `scripts` to `devshellCommands`?
Because our scripts *are* devshell commands.
This makes them also very lightweight, as we assume that they are run inside the devshell, where we already have all our dependencies.


__std__/cells/plutus/packages/buildable-plutus-fragments.nix
```diff
@@ -0,0 +1,21 @@
+# Like the names suggests, this special derivation builds all available derivations
+# inside the plutus cell, and is used by the check-the-flake script.
```
michaelpj, line 2: Could we just use the ci jobset as a proxy for this?

michaelpj (commented):
There's one bit I really don't like but otherwise seems fine. I do still think it would be better to do one thing at a time...

__std__/cells/plutus/devshellCommands/fix-cabal-fmt.nix
```diff
@@ -0,0 +1,10 @@
+{ inputs, cell }:
+
+cell.library.pkgs.writeShellScriptBin "fix-cabal-fmt" ''
```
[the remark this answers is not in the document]
zeme-iohk: > Practically the whole selling point of nix is that derivations are standalone, and now this one isn't and it's back to the bad old world of implicit runtime dependencies :(

Can't argue with that. 
I couldn't resist the chance to make those scripts shorter... 
But this is asking for trouble. 
I'll rollback 👍

__std__/cells/plutus/devshells/plutus-shell.nix
```diff
@@ -10,44 +10,68 @@ inputs.std.lib.dev.mkShell {
...
-    inputs.cells.toolchain.devshellProfiles.common
   ];
 
   commands = [
     {
-      package = inputs.cells.toolchain.packages.fix-cabal-fmt;
+      package = cell.devshellCommands.check-the-flake;
+      category = "general commands";
+      help = "Build everything in the flake";
+    }
```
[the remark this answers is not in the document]
zeme-iohk: That's right. I'll rollback. However I'll move them to the `packages` cell block, no need for a separate `script` cell block.

__std__/cells/plutus/packages/buildable-plutus-fragments.nix
```diff
@@ -0,0 +1,21 @@
...
+
+{ inputs, cell }:
+
+let
+  inherit (cell.library.pkgs) lib releaseTools;
+
+  fragments =
+    cell.devshells //
```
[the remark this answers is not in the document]
zeme-iohk: Yes I was very mistaken about build times.
I say we just keep the `pre-commit` check as a GitHub action, and leave everything else in Cicero?

__std__/cells/plutus/devshells/plutus-shell.nix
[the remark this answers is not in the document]
michaelpj: that sounds fine yea

__std__/cells/plutus/packages/buildable-plutus-fragments.nix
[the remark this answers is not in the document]
michaelpj: Hydra gives bad errors if there's an evaluation failure, which is why we had this in the first place. Are the errors we'll get from Cicero better? I think probably, since we'll just see the eval error directly. So it may indeed just be fine.

.github/workflows/test.yml
```diff
@@ -32,6 +32,4 @@ jobs:
             experimental-features = nix-command flakes
             accept-flake-config = true 
       - name: Pre Commit Check
-        run: nix build .#pre-commit-check  
-      - name: Plutus Shell 
-        run: nix develop .#plutus-shell --build 
+        run: nix build .#pre-commit-check
```
zeme-iohk, line 35: `nix develop .#plutus-shell --build` takes too long (10 mins) and is very unlikely to break.
It will be checked by hydra anyway.

__std__/cells/plutus/packages/serve-readthedocs-site.nix
```diff
@@ -0,0 +1,17 @@
+{ inputs, cell }:
```
zeme-iohk, line 1: I couldn't get `git` to see these 3 as renames...

__std__/cells/plutus/packages/check-the-flake.nix
```diff
@@ -1,8 +1,7 @@
 { inputs, cell }:
 
 # TODO(std) path must be fixed once __std__ is brought to the toplevel.
-# TODO(std) make this part of CI
-# TODO(std) turn this into a single derivation using recurseForDerivations
```
zeme-iohk, line 5: I really find this script super useful the way it is.
It has helped me find bugs all over the place.
It also implicitly verifies the `one-derivation-per-file` rule and the `file-name == flake-fragment-name` rule.
I think turning it into a normal derivation using `recurseForDerivations` & co would make it worse.

.github/workflows/test.yml
michaelpj: Right, but this isn't what we did before, we checked that it _evaluated_! Is there a flake equivalent of checking whether an attribute evaluates?

michaelpj (approved):

__std__/cells/plutus/packages/check-the-flake.nix
zeme-iohk: `nix flake check` has this problem https://github.com/NixOS/nix/issues/4265
Running `nix flake check --system x86_64-darwin --show-trace` (both with and without the `--system` flag) returns :
```
error: a 'x86_64-linux' with features {} is required to build '/nix/store/b8z3p4b25dv1dc756852b63kk6n63l2y-cabal.project.drv', but I am a 'x86-64_darwin' with features {benchmark, big-parallel, nixos-test}

       … while realising the context of path '/nix/store/m76radakffd5n56x1iq15zj8jbkd063f-haskell-project-plan-to-nix-pkgs'

       at /nix/store/z0g3r63vhf8h1ac2fmcv7bygz5llnxmz-source/lib/import-and-filter-project.nix:18:13:

           17|   projectSubDir'' = if projectSubDir == "" then "" else projectSubDir + "/"; # With trailing /
           18|   project = import "${projectNix}${projectSubDir'}";
             |             ^
           19| in project // {


[...]


       … while checking the derivation 'devShells.x86_64-linux.default'

       at /nix/store/ldhr3ysw5vmry7iggzrv832bai80v080-cells/plutus/devshells/default.nix:9:3:

            8|
            9|   default = plutus-shell;
             |   ^
           10| }

       … while checking flake output 'devShells'
``` 

Odd that it's trying to check `x86_64-linux` even on Darwin and even with the `--system` flag 🤦 
The `--no-build` flag also does nothing ...

.github/workflows/test.yml
zeme-iohk: There doesn't seem to be a way to `nix flake check` a single *frament*

PR #4910, opened 2022-10-13
fix: make legacy ci code pure

Pacman99:
Since plutus-apps likely isn't going to be standardized by the time hydra goes down, I need to add tullia exports to that repo with the legacy ci code. That code relies on the ci.nix in this repo, so this is a fix to make it pure here.

### DIFF
--- a/ci.nix
+++ b/ci.nix
@@ -15,7 +15,8 @@ let
   # currently that is linux and darwin.
   systems = filterSystems supportedSystems;
   crossSystems =
-    let pkgs = (import ./default.nix { }).pkgs;
+    # System doesn't matter as long as we only use lib
+    let pkgs = (import ./default.nix { system = "x86_64-linux"; }).pkgs;
     in { inherit (pkgs.lib.systems.examples) mingwW64; };
 
   # Collects haskell derivations and builds an attrset:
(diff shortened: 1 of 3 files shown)


michaelpj:
> That code relies on the ci.nix in this repo, so this is a fix to make it pure here.

I'm sorry, what? How does it do that? Better to just vendor it there if it needs it.

Pacman99:
This line, https://github.com/input-output-hk/plutus-apps/blob/4474fac99868fad0b02fdd044716e54d2b5b054d/ci.nix#L11, imports the ci lib from this repo. Since plutus-apps isn't standardized I need to use the ci.nix in plutus-apps to export tullia jobs just like I originally did in the cicero PR here, before the standardization happened.

I never commited the fixes to make evaluation pure in this repo, so I'm doing that now.

michaelpj:
> This line, https://github.com/input-output-hk/plutus-apps/blob/4474fac99868fad0b02fdd044716e54d2b5b054d/ci.nix#L11, imports the ci lib from this repo. 

We can just copy it over there if necessary, no? And there's nothing forcing them to update their pin of `plutus` for now. I don't think we should regress here to support that.

Pacman99:
> We can just copy it over there if necessary, no?

Sounds good, I did just that

PR #4912, opened 2022-10-18
Fix Cicero CI

zeme-iohk:


michaelpj:
It would be cool to include the cicero evaluator command also, but let's merge this to get things unblocked hopefully.

PR #4915, opened 2022-10-20
PLT-834: ADR for a steppable CEK machine for debugging.

thealmarty:
This PR adds the proposed design details of the steppable CEK machine.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+
+As long as the debugging machine has the same type, we can alter `computeCek`/`returnCek` to be polymorphic over a type-level `Bool` specifying if we’re in debug mode or not. Then we demote it to the term level in the definition of `computeCek`/`returnCek` and branch on the `Bool` thus implementing different logic depending on whether we're in debug mode or not. This promotion to the type level allows us to statically instantiate the `Bool` in an instance and thus make GHC compile the whole worker of the CEK machine twice: once in debug mode and once in production mode. Theoretically, GHC will take care to remove all the dead debug code when in production mode.
+
+We currently decide to not take this approach because:
+
+- It's not clear whether we can implement it this way. Implementing a separate machine is simpler.
+- Some slowdown may still happen due to GHC optimizations being very unreliable.
+- It's difficult to detect slow downs due to this as we add more builtin functions.
+
+If the separate machine turns out to not provide correct debugging information, we may revisit this option again.
```
effectfully, line 50: Which is why we should generate traces in the debugging machine and in the production one (by piggybacking on the costing machinery) and test that those are the same.

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+    continue state
+continue (Returning !unbudgetedSteps ctx val) = do
+    state <- returnCekStep unbudgetedSteps ctx val
+    continue state
+continue (Terminating term) = pure term
+```
+
+### Alternative: use CPS
+
+This approach has been suggested to improve the performance of the debugging machine. The abstraction proposed above does not have good performance. To improve performance, we need to get rid of the creation of the state objects. This is because GHC probably doesn't have the capability to eliminate those, because the place where they're destructed is too far from where they're constructed. That means that instead we probably need to do something with continuations:
```
effectfully, line 107: > This approach has been suggested to improve the performance of the debugging machine. The abstraction proposed above does not have good performance.

For the production machine. For the debugging machine the performance of this approach is perfectly OK and we don't care about performance here much anyway.

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+The machine computation interprets each request into an `m` action.
+It is essentially a natural transformation from the suspension functor to `m`. This `m` will replace our current monad `CekM`.
+
+Suppose we define a type `SteppableCekM a` as our base monad `m`.
+Then the machine computation can be implemented as the following request handler function:
+
+```haskell
+handle :: RequestF a -> SteppableCekM a
+handle = \case
+  StepF state k -> step state >>= pure . k
```
effectfully, line 175: `step state >>= pure . k` is `k <$> step state` (the two below are similar).

effectfully (commented):
I don't understand the need for the `FreeT` stuff. I'm not against `FreeT` in general (e.g. I used free applicatives in a part of the compilation pipeline for Zonk), but I don't see why it's helpful here. Could you elaborate on that?

zliu41:
> I don't understand the need for the FreeT stuff. I'm not against FreeT in general (e.g. I used free applicatives in a part of the compilation pipeline for Zonk), but I don't see why it's helpful here. Could you elaborate on that?

It's one way of structuring the code. The debugger is naturally a coroutine, where one routine is the user and the other is the Cek machine, and they take turns to suspend and pass data and control to each other in a debugging session. So I suggested structure the code as a coroutine, as described in the linked paper. I think this should also play nicely when we implement Debug Adapter Protocol for the debugger (not part of MVP).

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+Proposed
+
+## Context
+
+In order to have a minimal viable product of a debugger for Plutus, we need a CEK machine that will give us more information for debugging than our current one.
+
+Implementing the steppable CEK machine is a non-trivial task and involves some design decisions:
+
+- should the debugging machine be a separate one?
+- should we modify the existing CEK machine to support stepping, or implement something entirely new?
```
zliu41, line 21: The first two bullet points are the same thing.

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Argument: A separate machine
+
+The debugging machine will give the most accurate information if it was the one in production. However, the drawback of that is that we have to care about its performance.
+
+We only care about performance if the steppable CEK machine will *replace* the production machine instead of being a separate one.
+
+Our current CEK machine's performance is sensitive to changes. For example, exporting the `computeCek` function from the module causes the CEK machine to become slower by up to 25%. Refactoring the machine in preparation for the stepper also shows significant slow downs. See [PR#4909](https://github.com/input-output-hk/plutus/pull/4909). We have enough evidence to show that replacing the machine without slow downs is impossible.
+
+We consciously chose to keep up the production CEK machine's performance by adding a separate machine for debugging. We will mitigate the drawback of this by keeping the implementation as close to the production machine as possible.
```
zliu41, line 36: Another benefit worth mentioning of a separate CEK machine is that it allows us to quickly iterate on it without worrying about performance regression.

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Alternative: polymorphic compute/return steps
+
+This approach has been suggested to potentially not introduce slow downs while not having a separate debugging CEK machine.
+
+As long as the debugging machine has the same type, we can alter `computeCek`/`returnCek` to be polymorphic over a type-level `Bool` specifying if we’re in debug mode or not. Then we demote it to the term level in the definition of `computeCek`/`returnCek` and branch on the `Bool` thus implementing different logic depending on whether we're in debug mode or not. This promotion to the type level allows us to statically instantiate the `Bool` in an instance and thus make GHC compile the whole worker of the CEK machine twice: once in debug mode and once in production mode. Theoretically, GHC will take care to remove all the dead debug code when in production mode.
```
zliu41, line 42: I don't think this approach actually works, since a steppable `computeCek`/`returnCek` won't have the same types as the normal ones. And I don't imagine the debug code to be a superset of the production code, which is what this approach assumes (hence using a Bool flag to turn debug code off in production).

Or maybe I misunderstood it; in either case, if you describe an approach here it is helpful to have some code snippet.

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+
+- should the debugging machine be a separate one?
+- should we modify the existing CEK machine to support stepping, or implement something entirely new?
+- how do we ensure performance and conformance?
+
+## Decision: A separate machine
+
+We came to the conclusion that the debugging machine should be a separate one. The main reason is that we don't want to slow down the production machine.
+
+## Argument: A separate machine
```
zliu41, line 28: This should be `###`.

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+
+The debugging machine will give the most accurate information if it was the one in production. However, the drawback of that is that we have to care about its performance.
+
+We only care about performance if the steppable CEK machine will *replace* the production machine instead of being a separate one.
+
+Our current CEK machine's performance is sensitive to changes. For example, exporting the `computeCek` function from the module causes the CEK machine to become slower by up to 25%. Refactoring the machine in preparation for the stepper also shows significant slow downs. See [PR#4909](https://github.com/input-output-hk/plutus/pull/4909). We have enough evidence to show that replacing the machine without slow downs is impossible.
+
+We consciously chose to keep up the production CEK machine's performance by adding a separate machine for debugging. We will mitigate the drawback of this by keeping the implementation as close to the production machine as possible.
+
+## Alternative: polymorphic compute/return steps
```
zliu41, line 38: ditto

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+
+So `computeStep` gets inlined, as before, at which point we hopefully know `k`, so GHC can optimize it properly. Except it's going to be more complicated than that because:
+
+1. We need to pass continuations for all the different cases, so three of them.
+2. The continuations are recursive with the main function! It's not clear how to make it work nicely.
+
+We currently aren't taking this approach because:
+
+- The performance of the debugging machine is not a main concern.
+- It's unclear how to make some parts work nicely.
```
zliu41, line 126: Can you elaborate on this?

zliu41 (commented):
It is worth mentioning that we can get away with using the same monad, `CekM`, in the steppable CEK machine, because we can convert it to/from IO via `unsafeSTToIO` and `unsafeIOToST` when we need IO capabilities.

doc/adr/0002-steppable-cek-machine.md
effectfully: See [this](https://github.com/input-output-hk/plutus/pull/4909#issuecomment-1286189092) comment. If you agree it seems to make sense, we can include its content in the proposal (or just a link to it).

effectfully: (it's only one, probably simplest, way of doing it. We could as well use `Either (Proxy a) (Proxy b)` instead of `Bool` and sneak into that argument any type information we want. Or have more arguments to the type class that only the `Bool` one)

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+We came to the conclusion that the debugging machine should be a separate one. The main reason is that we don't want to slow down the production machine.
+
+## Argument: A separate machine
+
+The debugging machine will give the most accurate information if it was the one in production. However, the drawback of that is that we have to care about its performance.
+
+We only care about performance if the steppable CEK machine will *replace* the production machine instead of being a separate one.
+
+Our current CEK machine's performance is sensitive to changes. For example, exporting the `computeCek` function from the module causes the CEK machine to become slower by up to 25%. Refactoring the machine in preparation for the stepper also shows significant slow downs. See [PR#4909](https://github.com/input-output-hk/plutus/pull/4909). We have enough evidence to show that replacing the machine without slow downs is impossible.
+
```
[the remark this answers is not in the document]
thealmarty: fair enough. I'll change it to "difficult" instead of "impossible".

doc/adr/0002-steppable-cek-machine.md
thealmarty: It seems Roman has proposed something that makes this approach irrelevant......not sure if I should still mention this.

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+## Context
+
+In order to have a minimal viable product of a debugger for Plutus, we need a CEK machine that will give us more information for debugging than our current one.
+
+Implementing the steppable CEK machine is a non-trivial task and involves some design decisions:
+
+- should the debugging machine be a separate one?
+- should we modify the existing CEK machine to support stepping, or implement something entirely new?
+- how do we ensure performance and conformance?
```
michaelpj, line 22: This isn't a decision!

How about this:

We have a decision to make about whether or not to write the debuggable machine as a separate machine or try to share code with the production machine. The _requirements_ that we need to satisfy are a) we must not compromise the performance of the production machine, and b) the debugging machine must behave the same as the production machine.

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Decision: A separate machine
+
+We came to the conclusion that the debugging machine should be a separate one. The main reason is that we don't want to slow down the production machine.
+
+## Argument: A separate machine
+
+The debugging machine will give the most accurate information if it was the one in production. However, the drawback of that is that we have to care about its performance.
+
+We only care about performance if the steppable CEK machine will *replace* the production machine instead of being a separate one.
```
michaelpj, line 32: I think there are a couple of things here:
- The "steppable machine" which is used in the debugger
- A hypothetical refactoring of the production machine which would allow its code to be used in both the debugger and the production machine.

I think it's confusing to say we were thinking of _replacing_ the production machine with the steppable one.

I think this sentence is rather: we only care about performance if the steppable machine and the production machine share code.

doc/adr/0002-steppable-cek-machine.md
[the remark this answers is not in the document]
michaelpj: Yeah, I don't think we've investigated this. We've made the very start of an investigation, but I don't think we've done anywhere near enough to justify this.

doc/adr/0002-steppable-cek-machine.md
michaelpj: I think we should try it, of course.

doc/adr/0002-steppable-cek-machine.md
michaelpj: Why would it not provide correct debugging information? I'm confused by the implication that they would behave differently for some reason?

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+
+We currently decide to not take this approach because:
+
+- It's not clear whether we can implement it this way. Implementing a separate machine is simpler.
+- Some slowdown may still happen due to GHC optimizations being very unreliable.
+- It's difficult to detect slow downs due to this as we add more builtin functions.
+
+If the separate machine turns out to not provide correct debugging information, we may revisit this option again.
+
+## Decision: implementing it as a coroutine system
```
michaelpj, line 52: Is this a coroutine system? I wouldn't have called it that but maybe it is?

doc/adr/0002-steppable-cek-machine.md
michaelpj: Exactly. This is only relevant if we're talking about refactoring the production machine to support both ways, then we need to make it fast. So it's another alternative alongside "polymorphic compute/return steps"

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+2. The continuations are recursive with the main function! It's not clear how to make it work nicely.
+
+We currently aren't taking this approach because:
+
+- The performance of the debugging machine is not a main concern.
+- It's unclear how to make some parts work nicely.
+
+If we find out that the debugging machine is too slow, we may want to revisit this approach again.
+
+### Coroutines in Haskell
```
michaelpj, line 130: Do we have a prototype of this? I'd rather see a prototype before deciding on this based on an abstract argument.

michaelpj: I also don't think this needs to go in here: if the decision is to have a separate machine, we can have a separate ADR once we decide how to structure it.

michaelpj (commented):
I'm a bit confused: it feels to me like we're at a point where we could have the start of a design doc, but not at the point where we can make a decision! In particular, we haven't tried very hard to refactor the CEK machine to make the code reusable. The thing that Marty and I wrote was really an extremely naive first approach, we can almost certainly do better and I don't think we should declare our investigation done until we have.

doc/adr/0002-steppable-cek-machine.md
thealmarty: OK it sounds like this (implementing a separate machine) will be the alternative anyway, so we're not deciding on that now. I'd rather include this as the alternative since it's already written anyway. We can of course have another ADR when it comes to the point that we have to revisit the alternative.

doc/adr/0002-steppable-cek-machine.md
michaelpj: Plus, later on you say we're going to make a decision about how to architect the steppable machine also. As I said there I think that could be an ADR since we've not even got a prototype yet

doc/adr/0002-steppable-cek-machine.md
michaelpj: I see, that bit was later, the structure got me a bit confused.

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,235 @@
...
+- should we modify the existing CEK machine to support stepping, or implement something entirely new?
+- how do we ensure performance and conformance?
+
+## Decision: A separate machine
+
+We came to the conclusion that the debugging machine should be a separate one. The main reason is that we don't want to slow down the production machine.
+
+## Argument: A separate machine
+
+The debugging machine will give the most accurate information if it was the one in production. However, the drawback of that is that we have to care about its performance.
```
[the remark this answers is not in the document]
thealmarty: The idea is that if the production machine was the debugging machine also, it certainly knew what it just did, and could give us that same information. But with a separate machine, this is not as guaranteed. I will make it more clear. This is along the same line of avoiding making mistakes.

doc/adr/0002-steppable-cek-machine.md
michaelpj: Sorry, I didn't quite follow that... What I'm suggesting is that we have one ADR for whether or not to use a separate machine, and one for how to structure the steppable machine, however we do it. Note that what you've written here about coroutines will I think be useful regardless, but it definitely belongs to a discussion of how to structure the steppable machine.

doc/adr/0002-steppable-cek-machine.md
thealmarty: The point is that they shouldn't! And we need to make sure of that. See my [other comment](https://github.com/input-output-hk/plutus/pull/4915#discussion_r1001934889).

doc/adr/0002-steppable-cek-machine.md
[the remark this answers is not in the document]
thealmarty: I still think it's very unlikely that performance won't be compromised but I agree it's worth investigating more. So I'll change the proposed approach and try more first.

doc/adr/0002-steppable-cek-machine.md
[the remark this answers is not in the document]
michaelpj: I read "it will give the most accurate information" as "even if we implement it correctly, it might somehow give less accurate (?) information", rather than "if we do it this way there is more scope for us to implement it wrongly and have bugs".

I also think it's a bit misleading to say that in the alternative where we share code they are the _same machine_. IMO they are two machines that _share code_. So I would say something like "if we share code between the steppable machine an the production machine there is less scope for us to make mistakes".

doc/adr/0002-steppable-cek-machine.md
michaelpj: The sentence didn't read to me like that. It read to me like "if we implement it correctly and somehow it gives wrong information, then we should...", rather than "if we struggle to implement the separate machine correctly, then we should..."

doc/adr/0002-steppable-cek-machine.md
thealmarty: OK I see. I can do that.

thealmarty: I'll open another PR for another ADR - this one will stay talking about how to structure the steppable machine. The new PR will have the ADR about how to have the debugging machine share code with the production machine without compromising performance.

thealmarty: Yeah I think that will be a lot clearer. Great suggestion!

michaelpj:
It sounds like I misunderstood our own process :sweat_smile: and the idea is that we can have draft ADRs that don't yet represent decisions. In that case I think I'm happy with this.

doc/index.rst
```diff
@@ -45,8 +45,10 @@ for writing Haskell code that can be compiled to Plutus Core.
 .. toctree::
    :caption: Architectural decision records
    :maxdepth: 1
+   :titlesonly:
 
-   adr/index
+   0001-record-architecture-decisions
+   0002-steppable-cek-machine
```
michaelpj, line 51: Did you mean to do this? I think this makes the other index file redundant since you just flattened the hierarchy?

doc/adr/0002-steppable-cek-machine.md
```diff
@@ -0,0 +1,189 @@
...
+
+In order to provide debugging information for each evaluation step, we need a steppable CEK machine. Implementing the steppable CEK machine is a non-trivial task and involves some design decisions. One decision to make is about whether we can share the code between the production and the debugging machine. That is not the scope of this ADR. See the next ADR for that.
+
+This ADR proposes a design for an implementation of a steppable CEK machine. Of course, this doesn't mean that this is the final decision. This means that the next step for us is to prototype the machine in this way - which we have reasons to believe will go well. We may adjust our proposed approach depending on how the prototyping goes.
+
+## Decision
+
+This section describes the proposed implementation of the debugging machine.
+
+We first **abstracts out the computation to "steps"** on our current machine. We then **implement a coroutine system** to add the debugging functionalities.
```
michaelpj, line 26:
```suggestion
We first **abstract out the computation to "steps"** on our current machine. We then **implement a coroutine system** to add the debugging functionalities.
```

michaelpj (approved):

doc/index.rst
thealmarty: I was just following the instruction on the ADR index.rst. Should I have?

thealmarty: if we don't want to I'll update index.rst too.

michaelpj: Which instruction? It says to add it to the TOC, which you did by adding it to `adr/index.rst`. Here you've replaced the TOC here that referenced that `index.rst` with something that lists all the ADRs directly, and I'm not sure why...

thealmarty: Oops, actually I think I got the files mixed up. Thanks.

doc/adr/0002-steppable-cek-machine.md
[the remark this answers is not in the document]
kwxm: If we were C programmers we might keep the two machines synchronised by using a shell script to extract relevant code from the real evaluator and then pasting it into the source for the debugging machine using `#include`.  Of course, we're far too principled to do anything like that ................. aren't we?

kwxm (approved):
This all looks sensible to me, and it seems to be in a state where we can get at least get started, and maybe pin things down a bit more firmly when we see how it works out.  Beginning with a separate machine seems like a good way to go since we should be able to get something working without worrying about breaking everything else.  That'll make it a lot easier to experiment, and once we've got something usable we might want to think about integration with the production machine (although it would still be useful to have a preliminary plan for that which we can keep in mind while we're implementing the steppable machine).

doc/adr/0002-steppable-cek-machine.md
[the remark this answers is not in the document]
michaelpj: Oh, we can do that in Haskell too with enough abuse of CPP :D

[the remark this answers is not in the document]
kwxm: > Oh, we can do that in Haskell too with enough abuse of CPP :D

Yes, I was trying to suggest that without saying it out loud!

bezirg (approved):

thealmarty:
@michaelpj please force merge this. Thanks.

PR #4916, opened 2022-10-21
PLT-1061: Script for updating versions and version bounds in cabal files

zliu41:
Usage: `./scripts/update-versions.sh x.y.z.w`. This updates versions in selected cabal files to `x.y.z.w`, and version bounds in all cabal files to `^>=x.y`.

### DIFF
--- a/scripts/update-versions.sh
+++ b/scripts/update-versions.sh
@@ -9,6 +9,9 @@ versioned_packages=(
   "plutus-tx-plugin"
 )
 
+additional_packages=( "$@" )
+versioned_packages+=( "${additional_packages[@]:1}" )
+
 versioned_packages_regex=$(printf "\|%s" "${versioned_packages[@]}")
 versioned_packages_regex="\(${versioned_packages_regex:2}\)"
 
--- a/scripts/update-versions.sh
+++ b/scripts/update-versions.sh
@@ -23,4 +23,12 @@ major_version="${components[0]}.${components[1]}"
 find . -regex ".*/$versioned_packages_regex\.cabal" -exec sed -i "s/\(^version:\s*\).*/\1$1/" {} \;
 
 # update version bounds in all cabal files
+# It looks for patterns like the following:
+#
+# - ", plutus-core"
+# - ", plutus-core:uplc"
+# - ", plutus-core ^>=1.0"
+# - ", plutus-core:{plutus-core, plutus-core-testlib}  ^>=1.0"
+#
+# and updates the version bounds to "^>={major version}"
 find . -name "*.cabal" -exec sed -i "s/\(, ${versioned_packages_regex}[^^]*\).*/\1 ^>=$major_version/" {} \;
--- a/scripts/update-versions.sh
+++ b/scripts/update-versions.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+versioned_packages="\(plutus-core\|plutus-ledger-api\|plutus-tx\|plutus-tx-plugin\)"
+
+IFS='.' read -r -a components <<< "$1"
+
+major_version="${components[0]}.${components[1]}"
+
+# update package versions in cabal files for versioned packages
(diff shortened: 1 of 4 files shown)


scripts/update-versions.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+versioned_packages="\(plutus-core\|plutus-ledger-api\|plutus-tx\|plutus-tx-plugin\)"
```
michaelpj, line 5: what about `word-array` and `prettyprinter-configurable`? They're dependencies of those packages, so if they change we need to release updated versions, which means we need to version them somehow. I think the lockstep consideration means we version them the same as everything else.

michaelpj: maybe "versioned_packages_regex" since that's what it is. You could have a bash array of the packages and then turn it into the regex, slightly easier to update in future

scripts/update-versions.sh
```diff
@@ -0,0 +1,15 @@
...
+
+IFS='.' read -r -a components <<< "$1"
+
+major_version="${components[0]}.${components[1]}"
+
+# update package versions in cabal files for versioned packages
+find . -regex ".*/$versioned_packages\.cabal" -exec sed -i "s/\(^version:\s*\).*/\1$1/" {} \;
+
+# update version bounds in all cabal files
+find . -name "*.cabal" -exec sed -i "s/\(, ${versioned_packages}[^^]*\).*/\1 ^>=$major_version/" {} \;
```
michaelpj, line 15: Worth commenting the regexes! This is something like "look for something which looks like ', foo <anything>', and replace the <anything> with '^>=$major_version"

michaelpj (approved):
I assume you tested this and it works okay!

scripts/update-versions.sh
michaelpj: I guess we could special case `word-array` and `prettyprinter-configurable` since they change so infrequently :thinking:

zliu41: Special casing them in the sense that we only release them if they are changed sounds good to me.

michaelpj: In that case how about making `update-versions` take the list of packages to update from its arguments, using the current list as a default if it's not supplied. That way if we do want to bump `word-array` we can just run `update-versions.sh word-array`.

zliu41: Yup, that's exactly what I'm doing.

zliu41: See [04f13fe](https://github.com/input-output-hk/plutus/pull/4916/commits/04f13fe3e08ccc2d94f98631bcfa6ac378b1dd42)

scripts/update-versions.sh
zliu41: Added a comment.

scripts/update-versions.sh
```diff
@@ -9,6 +9,9 @@ versioned_packages=(
   "plutus-tx-plugin"
 )
 
+additional_packages=( "$@" )
+versioned_packages+=( "${additional_packages[@]:1}" )
```
michaelpj, line 13: I was actually thinking something more like:
```
version=$1
shift
default_packages=...
if [ "$#" == "0" ]; then
   packages=default_packages
else
   packages=( "$@" )
```
That way if I've run `update-versions` already and I just need to do some specific ones then I can just do exactly those. Consider:
```
update-versions 1.2
update-versions 1.0.1 prettyprinter-configurable
# oops, updated plutus-tx-plugin to 1.0.1 now :(
```

zliu41: Ok, I guess we can make the script work that way. But in terms of versioning, my preference is to use the same version for everything, although we can skip releasing a version if it hasn't changed. Versioning packages independently makes everything more complicated without much benefit.

michaelpj: Oh yes, good point. I think that's fine then. I just think it's a nicer factoring to have effectively both `updateSpecificVersions` and `updateDefaultVersions`, but yeah, what's here is fine.

PR #4917, opened 2022-10-21
PLT-1061: 1.1.0.0 release

zliu41:


plutus-core/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Changelog for plutus-core
```
michaelpj, line 1: Is it worth adding these if we don't have any content yet? I think we should maybe start producing them before we do a release with a CHANGELOG that's basically TODO

michaelpj (approved):

michaelpj:
Note that I changed both `prettyprinter-configurable` and `word-array` for 9.2 support so they need releases.

plutus-core/CHANGELOG.md
zliu41: Yeah, they don't serve much useful purpose. removed.

zliu41:
Updated `prettyprinter-configurable` and `word-array`.

michaelpj (approved):

PR #4920, opened 2022-10-24
Bump CHaP after reset

andreabedini:
Index state 2022-10-15T00:00:00Z does not exist in CHaP anymore.


michaelpj:
Thanks yes, we do need this.

PR #4930, opened 2022-10-27
PLT-1084 ADR for how to share code between production and debugging CEK machines.

thealmarty:
I'm not sure how to add more details to this. We may just have to do a prototype? Feedback and suggestions welcome!

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


effectfully (changes requested):
Please add the `.rst` extension to the file (or `.md` or whatever), otherwise every paragraph becomes a super long line and it's really hard to read the text.

doc/adr/0003-sharing-prod-debugging-cek.md
```diff
@@ -0,0 +1,95 @@
...
+
+```haskell
+computeCekStep
+    :: forall uni fun s
+    . (Ix fun, PrettyUni uni fun, GivenCekReqs uni fun s)
+    => WordArray
+    -> Context uni fun
+    -> Closure uni fun
+    -> CekM uni fun s (CekState uni fun)
+```
```
effectfully, line 59: That's not exactly the type signature of `computeCekStep` from below (in particular, the below function doesn't have the constraints even though they are indeed in the scope of the function). I think it would be clear just not to give the type signature here as one can see the function below anyway.

doc/adr/0003-sharing-prod-debugging-cek.md
```diff
@@ -0,0 +1,95 @@
...
+
+```haskell
+enterComputeCek :: forall uni fun . Bool -> (Term NamedDeBruijn uni fun ()) -> CekM uni fun s (CekState uni fun)
```
effectfully, line 62: This should have at the very least the `Context` and `CekValEnv` arguments as per out discussion regarding reentrance.

doc/adr/0003-sharing-prod-debugging-cek.md
```diff
@@ -0,0 +1,95 @@
...
+    computeCek = if debug then computeCekDebug else computeCekStep
+    {-# NOINLINE computeCek #-}  -- Making sure the `if` is only evaluated once.
+
+    computeCekDebug 
+        :: Term NamedDeBruijn uni fun ()
+        -- | The `SteppableCekM` is the monad for the debugging CEK machine. It may be just `CekM`.
+        -> SteppableCekM uni fun s (CekState uni fun) 
+    computeCekDebug term = do
+        -- `step` is the function to do a computation step. The first step is always computing the given term with an empty closure, nothing spent on the budget etc.
+        state <- step (Computing (toWordArray 0) NoFrame (Closure term Env.empty))
```
effectfully, line 73: No, `computeCekStep` serves at the stepping function! It does one step for the debugging machine and full evaluation for the production machine, it's the whole point of this representation. It's not some other `step` function.

doc/adr/0003-sharing-prod-debugging-cek.md
```diff
@@ -0,0 +1,95 @@
...
+        ...
+
+    computeCek
```
effectfully, line 76: `computeCekStep`, not `computeCek`.

effectfully (commented):
> I'm not sure how to add more details to this. We may just have to do a prototype? Feedback and suggestions welcome!

I don't think any more details are needed, next step should definitely be prototyping. Just clarify that this is a somewhat vague idea that hasn't be tried out yet and that the implementer should use their own judgement.

effectfully (approved):

effectfully:
Eek, apparently I can't approve old changes without approving new ones. Whatever.

thealmarty:
Sorry for the mess and thanks for clearing some things up @effectfully. I hope the new commit is clearer.

doc/adr/0003-sharing-prod-debugging-cek.md
effectfully: `computeCekStep` matches on the term and calls `computeCek` on a subterm. It's `computeCek` that is defined differently depending on whether we're in the debug mode or not. In the debug mode `computeCek` immediately returns the current `CekState` and halts execution. In the production mode `computeCek` calls the original `computeCekStep`, i.e. in the production mode `computeCekStep` calls itself through the thin `computeCek` wrapper thus achieving recursion and replicating the old behavior of the CEK machine.

Is this a clearer description of the idea?

The choice of names is arbitrary and do feel free to call things differently.

thealmarty: Yes, thank you very much. I'll fix it up tomorrow.

zliu41 (changes requested):
Please clean it up!

thealmarty:
> Please clean it up!

Oops sorry. Should be good now. Ready for review!

doc/read-the-docs-site/adr/0003-sharing-prod-debugging-cek.md
```diff
@@ -0,0 +1,117 @@
...
+We enter either modes via `enterComputeCek`, which takes an extra `Bool` than our current implementation, to indicate whether we are in debugging mode or not:
+
+```haskell
+enterComputeCek 
+    :: forall uni fun s
+    . (Ix fun, PrettyUni uni fun, GivenCekReqs uni fun s)
+    => Bool
+    -> Context uni fun
+    -> Closure uni fun
+    -> CekM uni fun s (CekState uni fun)
```
zliu41, line 58: Does this function compute a single step, or all the way? It looks like it does the former, if so, can you show the function that does the latter?

doc/read-the-docs-site/adr/0003-sharing-prod-debugging-cek.md
```diff
@@ -0,0 +1,117 @@
...
+enterComputeCek debug ctx (Closure term env) =
+    computeCekStep (toWordArray 0) NoFrame (Closure term Env.empty) where
```
zliu41, line 60: `computeCekStep` should be `computeCek`?

doc/read-the-docs-site/adr/0003-sharing-prod-debugging-cek.md
```diff
@@ -0,0 +1,117 @@
...
+        -> Context uni fun
+        -> CekValue uni fun
+        -> CekM uni fun s (CekState uni fun) -- return a state instead of a term
+    returnCekStep !unbudgetedSteps NoFrame val = do
+    spendAccumulatedBudget unbudgetedSteps
+    pure $ Terminating $ dischargeCekValue val --wrap the term in the `Terminating` constructor when returning the term.
+    <other_clauses>
+```
+
+Because when we are not debugging, we are still using basically the same code as our current implementation, the performance should not be affected by much. (Given that the machine is tail-recursive, the additional wrapping of the returned term in the `Terminating` constructor will affect performance in a negligible way.)
```
zliu41, line 113: This doesn't seem like basically the same code to me: previously `computeCek` performs the whole evaluation in a tail recursion, and now the computation is broken into individual steps, and it may need to allocate `CekState` and immediately garbage-collect it at every step (although perhaps GHC can optimize that away? I don't know). Have you run the benchmark on the proposed approach?

michaelpj: I think the example isn't quite clear. The idea is that we will (through type-level trickery) end up with two versions of the machine, _specialised_ to the two values of `debug`. The one that is specialized to `False` should produce similar code, since it just tail-calls itself, and only actually constructs a `CekState` at the very end.

doc/read-the-docs-site/adr/0003-sharing-prod-debugging-cek.md
thealmarty: I thought it was `computeCekStep` because we are doing this init computation whether we are in debug mode or not?

doc/read-the-docs-site/adr/0003-sharing-prod-debugging-cek.md
thealmarty: @zliu41 This hasn't been prototyped yet so no I haven't run the benchmark on the proposed approach. I could do so though if you think I should. Let me know.

@michaelpj it's true, I didn't include the whole story. I'll add that as best I can.

thealmarty:
We need to decide whether we want to add even more details and work on a prototype of this or wrap this ADR up as is. I understand that there's a lot of missing pieces we still need to figure out but that may be easier when we prototype this. So maybe another ADR can document more of that and a decision. 

I think the consensus was that this is worth a try, and we should go ahead and prototype. I hope this has established enough to support that.

doc/read-the-docs-site/adr/index.rst
```diff
@@ -10,7 +10,7 @@ The goals are:
 
 * making decisions transparent to internal/external stakeholders and contributors.
 
-* getting feeback on decisions that we're about to make or have made
+* getting feedback on decisions that we're about to make or have made
 
 * providing external contributors a framework to propose architectural changes
```
zeme-iohk, line 15: you will want to add 
`0003-sharing-prod-debugging-cek`
below 
`0002-steppable-cek-machine`
on line 35

doc/read-the-docs-site/adr/0003-sharing-prod-debugging-cek.md
zliu41: Then when do we call `computeCek`?

doc/read-the-docs-site/adr/0003-sharing-prod-debugging-cek.md
```diff
@@ -0,0 +1,117 @@
...
+    computeCekDebug 
+        :: WordArray
+        -> Context uni fun
+        -> Closure uni fun
+        -> CekM uni fun s (CekState uni fun) 
+    computeCekDebug budget ctx (Closure term env) = 
+        pure $ Computing budget ctx (Closure term env)
+
+    -- in production mode, recursively calls `computeCek`, and thus `computeCekStep`, similar to our current production machine.
+    computeCekStep
```
zliu41, line 80: @thealmarty you don't need to prototype. It's just not very clear to me, and I don't think I know how to prototype this if I want to do so. My main question is how do we end up with the same code as we have now when `debug = False`.

So, I guess my confusion is because this function and `returnCekStep` are incorrectly named? Because they are used when `debug = False`, but we shouldn't compute one step at a time when not debugging.

zeme-iohk:
I have re-enabled the `read-the-docs-site build` in another PR (which builds these `adr` docs -- among others), and it's erroring out with:
```
Warning, treated as error:
       > /private/tmp/nix-build-read-the-docs-site.drv-0/source/adr/index.rst:29:toctree contains reference to nonexisting document 'adr/0003-sharing-prod-debugging-cek'
```
It seems that the `.md` extension isn't working.
I've tried https://cerodell.github.io/sphinx-quickstart-guide/build/html/markdown.html but it still won't work.
If you must use `.md` I can look further into a workaround, otherwise we can convert this document to `.rst`
The same must be done for `adr/0002-steppable-cek-machine.md` which was added yesterday.

michaelpj:
@zeme-iohk the source filter for the RTD derivation doesn't include `.md` files, it probably should.

doc/read-the-docs-site/adr/0003-sharing-prod-debugging-cek.md
thealmarty: As far as I understood, it's called in `computeCekStep`. It'll match on the `term`, and call either `computeCek` or `returnCek`. Do that sound right? I'll add a note if it does.

thealmarty:
@zeme-iohk it's not a huge deal to convert them to `.rst`. Let me know if we need to.

zliu41 (approved):
I think this is fine to merge. It's still not totally clear how this will work but we can explore more when we prototype it.

Another confusion that I have is that I don't see how much code this allows us to share between the production and the debug machine. What I'm seeing doesn't seem to be very different than keeping the `enterComputeCek` exactly as it currently is, and implementing a totally separate `enterComputeCekDebug`.

michaelpj:
> What I'm seeing doesn't seem to be very different than keeping the enterComputeCek exactly as it currently is, and implementing a totally separate enterComputeCekDebug.

There's a lot of code in `<other clauses>`! I don't quite see what you're not seeing :D we have one implementation that covers all the actual logic of processing terms etc, and it's usable in both usecases. How is that not sharing?

zliu41:
It's not a bad idea to say things like

> There's a lot of code in <other clauses>!

and

> As far as I understand this, this trick lets us inline the "step" functions and call them recursively like before.

explicitly! They just weren't obvious to me. Also not a bad idea to show some code snippet or pointers about the "type level Bool trickery".

michaelpj:
TBH this is why I think this really needs a prototype.

thealmarty:
OK I've added more details as suggested. If everyone's happy enough we can merge #4953 then this.

thealmarty:
> TBH this is why I think this really needs a prototype.

[Ticket](https://input-output.atlassian.net/browse/PLT-1196) created.

PR #4931, opened 2022-10-27
[Builtins] [Test] Add 'ScottToMetaUnit'

effectfully:
As per the today's discussion.

### DIFF
--- a/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
+++ b/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
@@ -264,7 +264,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
 
     toBuiltinMeaning _ver ScottToMetaUnit =
         makeBuiltinMeaning
-            (mempty
+            ((\_ -> ())
                 -- @(->)@ switches from the Rep context to the Type one. We could make @(->)@
                 -- preserve the current context, but there's no such notion in the current
                 -- elaboration machinery and we'd better not complicate it further just for the sake
--- a/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
+++ b/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
@@ -104,6 +104,7 @@ data ExtensionFun
     | IdFInteger
     | IdList
     | IdRank2
+    | ScottToMetaUnit
     -- The next four are for testing that costing always precedes actual evaluation.
     | FailingSucc
     | ExpensiveSucc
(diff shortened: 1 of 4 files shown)


plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
```diff
@@ -261,6 +262,18 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
                 => afa -> afa)
             (\_ _ -> ExBudget 1 0)
 
+    toBuiltinMeaning _ver ScottToMetaUnit =
+        makeBuiltinMeaning
+            (mempty
```
michaelpj, line 267: specifically, this is `(\_ -> ())`, right? a little confusing.

michaelpj: In particular, it can't *do* anything with the argument, right?

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
```diff
@@ -167,13 +168,22 @@ test_IdRank2 :: TestTree
...
                 $ integer
         typecheckEvaluateCekNoEmit def defaultBuiltinCostModelExt term @?= Right (EvaluationSuccess res)
 
+-- | Test that a builtin can be applied to a non-constant term.
+test_ScottToMetaUnit :: TestTree
+test_ScottToMetaUnit =
+    testCase "ScottToMetaUnit" $ do
+        let res = mkConstant @() @DefaultUni () ()
+            -- > scottToMetaUnit Scott.unitval
+            term = apply () (builtin () ScottToMetaUnit) Scott.unitval
```
michaelpj, line 184: What happens if you pass a term that _doesn't_ have the right shape? Nothing, presumably?

plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
effectfully: > specifically, this is `(\_ -> ())`, right? a little confusing.

Yeah, sorry, I guess I love to use `mempty` in this case too much, I'll fix it.

effectfully: > In particular, it can't do anything with the argument, right?

It can check if the argument is a constant and if it happens to be one, it can do all sorts of stuff with it (in particular, implement the behavior of any other builtin). In the branch that I DMed you some time ago it was also possible to apply the term that was passed to a builtin, but there's no such functionality on `master`. So on `master` it's only the constants stuff.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
effectfully: Yup, we don't check it's of the right shape. It's ensured by the type checker, though. But that's irrelevant for untyped evaluation, of course.

effectfully: In the branch where it's allowed to apply arguments of builtin functions we could've implemented

```
scottToMetaUnit f = f {()} ()
```

(provided we add handling of type application, which sounds doable), but I don't think we care about any of that to complicate the builtins machinery even more.

michaelpj: Maybe worth having a test that passing something else also succeeds?

michaelpj: Just to make the demonstration obvious.

effectfully: Will do, thanks.

effectfully:
@michaelpj comments addressed, please merge (CI is broken).

PR #4934, opened 2022-10-28
Josfaj/plt 815 pt2 links to v2 examples

joseph-fajen:
PLT-815 part 2: Added links to examples of Plutus V2 functionalities to Reference > Plutus on Cardano > Plutus language changes. 

See draft content rendered as html here: 
https://plutus--4934.org.readthedocs.build/en/4934/reference/cardano/language-changes.html

This is what the new links look like: 

**Examples**
- Plutus V2 functionalities
- How to use reference inputs
- How to use inline datums
- How to reference scripts
- How to use collateral outputs

This content that these links point to was created by Jordan Millar, a developer from Cardano Node and Thomas Vellekoop, a developer in the Education Team.

### DIFF
--- a/doc/reference/cardano/language-changes.rst
+++ b/doc/reference/cardano/language-changes.rst
@@ -8,13 +8,13 @@ Language versions
 
 See the documentation on :ref:`language versions <what_are_plutus_language_versions>` for an explanation of what they are.
 
-PlutusV1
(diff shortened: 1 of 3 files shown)


michaelpj (approved):
I'm fine with this so long as we're happy with these links being stable and sensible for users to look at!

PR #4936, opened 2022-10-28
Restructure the `doc` folder

zeme-iohk:
As discussed in an old Jira Story


michaelpj:
We're probably also going to need to change the config on RTD, but I can do that.

michaelpj:
Actually, I should add you as a project administrator, can you create a RTD account and send me your username?

zeme-iohk:
I updated the RTD config

PR #4937, opened 2022-10-28
Move to 9.2 as the default devshell

michaelpj:
That's the only way the plugin builds now, so it should be the default!

@zeme-iohk I got rid of the un-qualifed `plutus-project` and shell so it's exactly clear which one we are using at all times. I think that's more regular, especially if this is going to keep going!

The CI for this will probably be busted because I changed the tullia job names so that the 8.10 and 9.2 jobs are side by side rather than having one privileged. But I built the required job locally and it worked.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


zeme-iohk:
> I got rid of the un-qualifed plutus-project and shell so it's exactly clear which one we are using at all times

💯 👍

zeme-iohk (approved):

zeme-iohk (changes requested):
Ah wait, don't we want to update `ghc-compiler-nix-name.nix` too? 
So that the tools are also built with 9.2.4?

michaelpj:
Hrm, I still dislike the existence of that :p

zeme-iohk:
Ok so maybe it's time to take care of that and go about that build matrix.
I'm happy to do that.
I'm just saying that the default shell now has a mix of 9.2. and 8.10 stuff in it

michaelpj:
Yeah, that's totally busted. And it's a bit annoying again, e.g. ideally the 8.10 shell should have 8.10 HLS and the same for the 9.2 shell. That used to work but is now a bit annoying :thinking: 

I'll try and get this at least not totally wrong, may involve more build fiddling. I did think it was surprising that it worked so easily :D

michaelpj:
Okay, I just moved most stuff to use 9.2 for now and left a cautionary note about using 8.10.

michaelpj:
I'm just going to merge this: I built it locally and cicero won't like it because of https://input-output.atlassian.net/browse/CIC-246 agian.

PR #4941, opened 2022-10-31
PLT-901: Implement some basic benchmarks for script context usage

michaelpj:
Two cases:
1. Decode the script context and do some comparable work
2. Decode the script context then ignore it and do something cheap

Both run with a couple of different sizes of script context. This should show some significant changes with the sums and products work.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-benchmark/script-contexts/src/PlutusBenchmark/ScriptContexts.hs
```diff
@@ -0,0 +1,79 @@
...
+  else PlutusTx.traceError "Odd number of outputs"
+
+mkCheckScriptContext1Code :: ScriptContext -> PlutusTx.CompiledCode ()
+mkCheckScriptContext1Code sc =
+  let d = PlutusTx.toBuiltinData sc
+  in $$(PlutusTx.compile [|| checkScriptContext1 ||]) `PlutusTx.applyCode` PlutusTx.liftCode d
+
+-- This example aims to *force* the decoding of the script context and then ignore it entirely.
+-- This corresponds to the unfortunate case where the decoding "wrapper" around a script forces
+-- all the decoding work to be done even if it isn't used.
```
zliu41, line 65: Do you mean [this wrapper](https://github.com/input-output-hk/plutus-apps/blob/2b811787dc3ab20cc2b0d1bfedf6f5b9a745d5c3/plutus-script-utils/src/Plutus/Script/Utils/V2/Typed/Scripts/Validators.hs#L91-L98)? It doesn't seem to force `unsafeFromBuiltinData` is the script context isn't used?

michaelpj: Function application is strict...

zliu41: Ah right, no such thing as call by need.

zliu41 (approved):

thealmarty (approved):
This is an interesting way to use golden test! Can't wait to see the difference!

plutus-benchmark/plutus-benchmark.cabal
```diff
@@ -357,3 +357,41 @@ executable ed25519-throughput
     , plutus-core           ^>=1.1
     , plutus-tx             ^>=1.1
     , plutus-tx-plugin      ^>=1.1
+
+---------------- script contexts ----------------
+
+library script-contexts-internal
+  import:          lang
+
+  if impl(ghc <9.0)
```
kwxm, line 366: Why do we have this now?  I can't run these locally because of this but if I change it I get errors because other things also have this requirement.

plutus-benchmark/script-contexts/src/PlutusBenchmark/ScriptContexts.hs
```diff
@@ -0,0 +1,79 @@
...
+
+mkValue :: Int -> Value
+mkValue i = assetClassValue (assetClass adaSymbol adaToken) (fromIntegral i)
+
+-- This example decodes the script context (which is O(size-of-context) work), and then also
+-- does some work that's roughly proportional to the size of the script context (counting the
+-- outputs). This should be a somewhat realistic example where a reasonable chunk of work is
+-- done in addition to decoding.
+{-# INLINABLE checkScriptContext1 #-}
+checkScriptContext1 :: PlutusTx.BuiltinData -> ()
```
kwxm, line 49: The names could be more meaningful, but I'm just saying that for the sake of saying something.

plutus-benchmark/script-contexts/src/PlutusBenchmark/ScriptContexts.hs
```diff
@@ -0,0 +1,79 @@
...
+checkScriptContext1 d =
+  let !sc = PlutusTx.unsafeFromBuiltinData d
```
kwxm, line 51: What happens without the bang?

plutus-benchmark/script-contexts/src/PlutusBenchmark/ScriptContexts.hs
```diff
@@ -0,0 +1,79 @@
...
+
+-- This example aims to *force* the decoding of the script context and then ignore it entirely.
+-- This corresponds to the unfortunate case where the decoding "wrapper" around a script forces
+-- all the decoding work to be done even if it isn't used.
+{-# INLINABLE checkScriptContext2 #-}
+checkScriptContext2 :: PlutusTx.BuiltinData -> ()
+checkScriptContext2 d =
+  let (sc :: ScriptContext) = PlutusTx.unsafeFromBuiltinData d
+  in case sc of
+    !_ ->
```
kwxm, line 71: Also, why is this using the bang in a different way from `cheeckScriptContext1`?  Maybe I'm just being obtuse: I think there's something about compilation and strictness that I'm missing.

kwxm (commented):
Looks fine.  I would have liked to experiment with it a bit to better understand what's going on, but I couldn't build it.

kwxm:
> Two cases:
> 
>     1. Decode the script context and do some comparable work
> 
>     2. Decode the script context then ignore it and do something cheap

I suppose it might have been interesting to have "Don't decode the script

kwxm:
Sorry, I seem to have inadvertently closed this.  I have no idea how that happened.

plutus-benchmark/script-contexts/src/PlutusBenchmark/ScriptContexts.hs
michaelpj: I'll comment. I was having trouble with GHC being too clever.

michaelpj:
> This is an interesting way to use golden test! Can't wait to see the difference!

We actually do this quite a bit! We have a bunch of other tests that do this to, they're a bit like benchmarks but because the output is 100% deterministic we can check in the output and explicitly accept changes this way.

> I suppose we could have had "Don't decode the script context and do something cheap" too, just to see how much effort decoding the context takes.

Yeah, although in this case I think I'm more interested in the delta for doing things _with_ the script context?

kwxm:
> Yeah, although in this case I think I'm more interested in the delta for doing things _with_ the script context?

If I can get it to build I'll try doing that myself just to satisfy my curiosity.

PR #4948, opened 2022-11-02
Fix Broken Haddock Derivation

zeme-iohk:


nix/cells/plutus/packages/read-the-docs-site.nix
```diff
@@ -5,7 +5,7 @@ let
 in
 
 pkgs.stdenv.mkDerivation {
-  name = "read-the-docs-site";
+  name = "aaaaaaaa__read-the-docs-site";
```
zeme-iohk, line 8: So that cicero runs this first ... will be reverted before merging

michaelpj: I don't think this makes any difference any more, we switched to only having one job and then nix schedules everything.

zeme-iohk: Ok 👍

doc/read-the-docs-site/adr/index.rst
```diff
@@ -30,5 +30,4 @@ The general process for creating an ADR is:
    :maxdepth: 1
    :titlesonly:
 
-   0001-record-architecture-decisions
-   0002-steppable-cek-machine
```
zeme-iohk, line 34: I will bring this back once https://github.com/input-output-hk/plutus/pull/4930#issuecomment-1301769289 has been settled

PR #4950, opened 2022-11-02
Remove plutus-tx dependency from plutus-benchmark-script-contexts-tests

zeme-iohk:
This fixes CI


michaelpj:
Huh, sorry. I'm not sure how that happened.

PR #4953, opened 2022-11-03
Fix read-the-doc-site derivation

zeme-iohk:


doc/read-the-docs-site/conf.py
```diff
@@ -82,7 +82,8 @@
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = [
     'haddock', # Otherwise it tries to pick up the README.md's in the Haddock doc!
-    'README.md'
+    'README.md',
+    'stdspan.md'
```
zeme-iohk: Otherwise we get:
```
Warning, treated as error:
       > /private/tmp/nix-build-read-the-docs-site.drv-0/source/srcspan.md:document isn't included in any toctree
```
Which makes me think: is `srcspan.md` supposed to be in the `read-the-docs-site` folder? 🤔

michaelpj: probably not. I think this might have gone wrong when you rearranged things?

zeme-iohk: I thought so too but I just went through the history and originally it was in the `doc` folder so... @bezirg what would be a better place for this file? Maybe `doc/notes`?

michaelpj: yes

michaelpj (approved):
LGTM, let's just move srcspan.md out

michaelpj:
Do you need to add it to the CI jobs somewhere? Presumably it's not in there now otherwise master would be breaking?

zeme-iohk:
Yes, the `read-the-docs-site` derivation is still commented out, that's why master isn't breaking.
Ill move `srcspan.md` to `notes` then merge this

michaelpj:
> Yes, the read-the-docs-site derivation is still commented out, that's why master isn't breaking.

But then shouldn't we uncomment it in this PR if it's now not broken?

zeme-iohk:
It's already uncommented.
We have the `/tmp/index.json: permission denied` bug being fixed in https://github.com/input-output-hk/plutus/pull/4948
And we have this. 
This is required to get #4948 to pass

PR #4956, opened 2022-11-04
Setup scriv and adjust release scripts

michaelpj:
This sets up `scriv` to be used in the repo.

- Add `scriv` to the devshell and setup `changelog.d` folders.
- Worked on the release scripts. Now there are three:
    - `update-version.sh` updates the version for a package
    - `assemble-changelog.sh` assembles the changelog for a package
    - `prepare-release.sh` calls the above two for a list of packages
- Documented the above in the release doc and CONTRIBUTING


zeme-iohk:
This look great!
Would be nice to turn `prepare-release.sh update-version.sh assembre-changelog.sh` into derivations and create devshell commands? 
I kinda want to get rid the `scripts` folder altogether (and it's gonna happen now that we are moving benchmarking to cicero)

michaelpj:
> Would be nice to turn prepare-release.sh update-version.sh assembre-changelog.sh into derivations and create devshell commands?

I'm not sure how I feel about that... I mean, it doesn't really add anything (except discoverability) because they don't need to be built, and means you can't use them from outside nix. I guess I continue to want the ability to add things to the devshell menu without having to provide the command as a derivation right there...

zeme-iohk:
> I'm not sure how I feel about that... I mean, it doesn't really add anything (except discoverability) because they don't need to be built

Sure, but l think that discoverability might be enough of a reason to do this.. 
As well as not having a ./scripts folder with random & unrelated stuff in it.

>and means you can't use them from outside nix.

You already can't use them outside nix, since they rely on `scriv`, which we provision through nix.

> I guess I continue to want the ability to add things to the devshell menu without having to provide the command as a derivation right there...

Mm we could do something like this:

```
  commands = [
    {
      package = cell.library.make-devshell-command ./scripts/update-release.sh 
      ...
    }
```
Where `cell.library.make-devshell-command` should be easy enough to implement.

But then we are back to the discussion of implicit runtime dependencies:
In this case `./scripts/update-release.sh` relies on `scriv` which is defined in the devshell, and if we removed it would break the script.

michaelpj:
> You already can't use them outside nix, since they rely on scriv, which we provision through nix.

You can install `scriv` yourself, which is easy enough :shrug: 

I do agree that I'm being inconsistent here about whether or not I care about implicit runtime dependencies :D IDK. And I guess I'm okay with `fix-stylish-haskell` and co being only available via nix, so :shrug:

zeme-iohk:
It just would've been nice to see them in the devshell menu, that's all

zeme-iohk (approved):

michaelpj:
Folks who haven't commented, any thoughts?

thealmarty (approved):

michaelpj:
> So it's not for every PR? It's a case by case basis? We use our own judgment if the PR merits a change fragment or not?

Great point! I should definitely mention this.

zeme-iohk (approved):

michaelpj:
@zliu41 ^^

scripts/update-version.sh
```diff
@@ -0,0 +1,40 @@
...
+  echo "$USAGE"
+fi
+
+set -euo pipefail
+
+PACKAGE=$1
+VERSION=$2
+
+versioned_packages_regex=$(printf "\|%s" "${versioned_packages[@]}")
+versioned_packages_regex="\(${versioned_packages_regex:2}\)"
```
zliu41, line 20: Not needed

zliu41 (commented):
Nice. Does this complete https://input-output.atlassian.net/browse/PLT-980?

michaelpj:
Last call for thoughts, @kwxm and @bezirg

CONTRIBUTING.adoc
```diff
@@ -364,6 +364,50 @@ Set it up if you can, it's relatively easy to do.
...
+==== How to write a changelog entry
+
+The basic idea is that you write a changelog "fragment" in the `changelog.d` directory.
+When we do a release, these will be collected into the main `CHANGELOG.md`.
+Usually we don't edit `CHANGELOG.md` directly.
+
+You can make a changelog fragment using `scriv create` in the package directory, but you can also just create the fragment directly with an editor.
+A fragment is a markdown file beginning with a header giving the category of change.
+Currently these are:
+- Removed
```
kwxm, line 389: Are these supposed to be displayed as an itemised list?  I just see
```
A fragment is a markdown file beginning with a header giving the category of change. 
Currently these are: - Removed - Added - Changed - Deprecated - Fixed - Security
```

scripts/prepare-release.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+
+# for some reason you can't just put a heredoc in a variable...
```
kwxm, line 3: You could make this into a shell function, like
```
usage () {
  echo "$(basename $0) VERSION [PACKAGE...]
  Prepares to release PACKAGEs at VERSION. If no PACKAGEs are provided,
  prepares to release the default packages."
}
```

and then just say `usage` later.

scripts/prepare-release.sh
```diff
@@ -0,0 +1,44 @@
...
+read -r -d '' USAGE << EOF
+prepare-release.sh VERSION [PACKAGE...]
```
kwxm, line 5:
```suggestion
$(basename $0) VERSION [PACKAGE...]
```

scripts/update-version.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+# for some reason you can't just put a heredoc in a variable...
```
kwxm, line 3: Again, you could use a shell function here.

scripts/update-version.sh
```diff
@@ -0,0 +1,37 @@
...
+read -r -d '' USAGE << EOF
+update-version.sh PACKAGE VERSION
```
kwxm, line 5:
```suggestion
$(basename $0) PACKAGE VERSION
```

scripts/assemble-changelog.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+# for some reason you can't just put a heredoc in a variable...
```
kwxm, line 3: See later comments.

bezirg (approved):

scripts/update-version.sh
```diff
@@ -0,0 +1,37 @@
...
+# It looks for patterns like the following:
+#
+# - ", plutus-core"
+# - ", plutus-core:uplc"
+# - ", plutus-core ^>=1.0"
+# - ", plutus-core:{plutus-core, plutus-core-testlib}  ^>=1.0"
+#
+# and updates the version bounds to "^>={major version}"
+echo "Updating version bounds on $PACKAGE to '>=$major_version'"
+find . -name "*.cabal" -exec sed -i "s/\(, $PACKAGE[^^]*\).*/\1 ^>=$major_version/" {} \;
```
kwxm, line 37: A courageous use of `sed -i` there.

scripts/prepare-release.sh
```diff
@@ -0,0 +1,44 @@
...
+  Prepares to release PACKAGEs at VERSION. If no PACKAGEs are provided,
+  prepares to release the default packges.
```
kwxm, line 7:
```suggestion
  prepares to release the default packages.
```

kwxm (approved):
Looks fine.  Let's see how it works out in practice.

scripts/assemble-changelog.sh
```diff
@@ -0,0 +1,19 @@
...
+read -r -d '' USAGE << EOF
+assemble-changelog.sh PACKAGE VERSION
+  Assembles the changelog for PACKAGE at VERSION.
+EOF
+
+if [ "$#" == "0" ]; then
+  echo "$USAGE"
+fi
```
kwxm, line 11: Oh wait.  Don't you need `exit 1` or something here?  I think that just prints out the message and keeps going.  It's similar in the other scripts.

michaelpj:
Thanks Kenneth, all good comments, applied.

PR #4957, opened 2022-11-04
Move benchmarks to cicero

zeme-iohk:


nix/cells/cloud/actions.nix
```diff
@@ -3,13 +3,33 @@
 ,
 }:
 {
-  "plutus/ci" = {
-    task = "ci";
+  # "plutus/ci" = {
```
zeme-iohk, line 6: Temporarily disabled so that we only trigger the "plutus/benchmark" pipeline

nix/cells/cloud/packages.nix
```diff
@@ -0,0 +1,9 @@
+# TODO(std) delete this file before merging
```
zeme-iohk, line 1: Just for testing;
```
mkdir out 
cd out 
nix build ../#x86_64-darwin.cloud.packages.test-run-plutus-benchmark
./result/bin/plutus-benchmark-runner
```

nix/cells/cloud/pipelines.nix
```diff
@@ -54,6 +55,20 @@ let
       }));
 
   ciTasksSeq = taskSequence "ci/" ciTasks (__attrNames ciTasks);
+
+  benchmark = { config, ... }:
+    let
+      runner = cell.library.plutus-benchmark-runner {
+        PR_NUMBER = "4956";
```
zeme-iohk, line 62: @Pacman99 here is where you pass me the params: Please let PR_NUMBER be a string

Pacman99:
Commenting to test issue_comment webhook

nix/cells/cloud/pipelines.nix
```diff
@@ -54,6 +55,20 @@ let
       }));
 
   ciTasksSeq = taskSequence "ci/" ciTasks (__attrNames ciTasks);
+
+  benchmark = { config, ... }:
+    let
+      runner = cell.library.plutus-benchmark-runner {
+        PR_NUMBER = config.actionRun.facts.github.github_body.pull_request.number;
+        PR_COMMIT_SHA = config.actionRun.facts.github.github_body.pull_request.head.sha;
+        BENCHMARK_NAME = "nofib";
```
zeme-iohk, line 64: @Pacman99 Eventually we will want the benchmark name here too

michaelpj:
FYI it should be the component name, e.g. https://github.com/input-output-hk/plutus/pull/4330#issuecomment-1042916848

Pacman99:
> FYI it should be the component name, e.g. [#4330 (comment)](https://github.com/input-output-hk/plutus/pull/4330#issuecomment-1042916848)

I think @zeme-iohk may have changed the benchmark script, to just take the benchmark name.

michaelpj:
Pretty sure it's still taking a component name: https://github.com/input-output-hk/plutus/pull/4957/files#diff-e6a0334d313629a99c5dcd944408ebef5538f58420d2772889baea43c4129726R67

nix/cells/automation/library/plutus-benchmark-runner.sh
```diff
@@ -29,40 +24,47 @@ if [ -z "$PR_NUMBER" ] ; then
...
 if [ -z "$BENCHMARK_NAME" ] ; then
    echo "[ci-plutus-benchmark]: 'BENCHMARK_NAME' is not set! Exiting"
    exit 1
 fi
+if [ -z "$GITHUB_TOKEN" ] ; then
+   echo "[ci-plutus-benchmark]: 'GITHUB_TOKEN' is not set! Exiting"
+   exit 1
+fi
+
+echo "[ci-plutus-benchmark]: Cloning the plutus repository"
```
michaelpj, line 40: Do we need to do this? This makes the script much harder to use locally. Before the checkout happened outside the test runner, can we not arrange for that to happen in the same way that we do for normal CI runs?

zeme-iohk: IIRC when I discussed this with @pacman I learned that in cicero we start with an empty directory and so we have to checkout the repo ourselves each time. Hence why I added those git commands to the script. 
If there is a way to workaround that we could probably shave off some time? 
As for running locally, we can indeed remove that check and maybe echo a warning

Pacman99: Just removed it, tullia can do this for us with the git preset

nix/cells/automation/library/plutus-benchmark-runner.sh
```diff
@@ -81,9 +68,11 @@ EOF
...
-
-echo "[ci-plutus-benchmark]: Posting results to GitHub ..."
-curl -s -H "Authorization: token $(</run/keys/buildkite-github-token)" -X POST -d "{\"body\": $(<bench-compare.json)}" "https://api.github.com/repos/input-output-hk/plutus/issues/${PR_NUMBER}/comments"
+jq -Rs '.' bench-compare-result.log >bench-compare.json
+if [ -v GITHUB_TOKEN ] ; then
+   echo "[ci-plutus-benchmark]: Posting results to GitHub ..."
+   # TODO(std) we'll need to change this
+   curl -s -H "Authorization: token $(<$GITHUB_TOKEN)" -X POST -d "{\"body\": $(<bench-compare.json)}" "https://api.github.com/repos/input-output-hk/plutus/issues/${PR_NUMBER}/comments"
+else
+   echo "[ci-plutus-benchmark]: GITHUB_TOKEN not set, not posting results to GitHub"
```
michaelpj, line 77: This should probably be a failure. Are we going to get a commit status or some way of knowing if the job failed?

Pacman99: I did it like this because you wanted the script to be able to be run locally, I assume you wouldn't want to have to get a github token if your just getting info on the command line.

michaelpj: That's a great point. Maybe I'm just talking nonsense, I'm no longer sure what should be true of this script :grimacing:  I guess leave it like this for now.

michaelpj:
Hmm, I don't see any jobs appearing in the action? But the comment reaction is appearing :thinking:

nix/cells/automation/pipelines.nix
```diff
@@ -39,10 +44,46 @@ in
...
+      command.text = "${runner}/bin/plutus-benchmark-runner";
+      nomad.templates = [
+        {
+          destination = "/secrets/cicero/github/token";
+          data = ''{{with secret "kv/data/cicero/github"}}{{.Data.data.token}}{{end}}'';
+        }
+      ];
+
+      memory = 1024 * 8;
+      nomad.resources.cpu = 10000;
```
angerman, line 80: Are there default values? What does 8G, and 10000 mean? How do they differ from the default?

nix/cells/automation/pipelines.nix
```diff
@@ -39,10 +44,46 @@ in
...
+
+      runner = cell.library.plutus-benchmark-runner {
+        PR_NUMBER = toString prFact.pull_request.number;
+        PR_COMMIT_SHA = prRevision;
+        BENCHMARK_NAME = lib.removePrefix "/benchmark " commentFact.comment.body;
+        GITHUB_TOKEN = "/secrets/cicero/github/token";
+      };
+    in
+    {
+      # Not importing common, github preset is not needed here
```
angerman, line 64: This comment is very hard to parse and understand

angerman:
@zeme-iohk _great_ job!

michaelpj:
How did this end up posting on a different PR?? https://github.com/input-output-hk/plutus/pull/4982#issuecomment-1331099879

michaelpj:
Anyway, it now seems to be failing on the benchmark comparison script... apparently `/usr/bin/env bash` doesn't work???

nix/cells/automation/pipelines.nix
dermetfan: [Default](https://developer.hashicorp.com/nomad/docs/job-specification/resources#resources-parameters) is 300 MB, 100 MHz

nix/cells/automation/pipelines.nix
```diff
@@ -39,10 +46,49 @@ in
...
+    let
+      fact = config.actionRun.facts.${cloud.library.actions.benchmark.input}.value.github_body;
+      prNumber = toString fact.issue.number;
+
+      prRevision = lib.getExe (pkgs.writeScript "get-pr-rev" ''
+        curl \
+          -H "Accept: application/vnd.github+json" \
+          https://api.github.com/repos/input-output-hk/plutus/pulls/${prNumber} \
+          | jq -r .base.sha
+      '');
```
dermetfan, line 59:
```suggestion
      prRevision = pkgs.writeScript "get-pr-rev" ''
        ${lib.getExe pkgs.curl} \
          -H "Accept: application/vnd.github+json" \
          https://api.github.com/repos/input-output-hk/plutus/pulls/${prNumber} \
          | ${lib.getExe pkgs.jq} -r .base.sha
      '';
```

- `pkgs.writeScript` builds the script at `$out`, hence `lib.getExe` is not necessary and actually breaks this; use `writeScriptBin` to get `$out/bin/get-pr-rev`
- interpolate full path to curl and jq in here so that they will be available at runtime

michaelpj:
Hooray!

michaelpj:
I think that noise threshold is higher than we used to have, I think our noise was <1% before and that looks like <2%, but that might be tolerable.

Pacman99 (approved):
This should be good to go

michaelpj (approved):

PR #4960, opened 2022-11-06
PLT-998: Update PLC specification for deferred unlifting

kwxm:
This updates the specification to defer all type checks for builtin application until applications become saturated. Previously we had a relation ~ which said what it meant for a single argument to compatible with what was specified in a signature. This was used immediately when the builtin was applied to a monomorphic argument, even if further arguments were still expected. The ~ relation has now been removed and we only use the ≈ relation, which compares an entire sequence of values with a signature when a builtin application becomes saturated.

In detail,

 *   The definition of compatibility of arguments with signatures in the second paragraph of 4.3.1 has been clarified slightly.
 *   The definition of the relation ~ (immediate type compatibility) in 4.3.2 has been deleted; the old section 4.3.3 now becomes 4.3.2.
 *   The verbal description of type well-formedness in the first paragraph of 5.1 has been simplified, as has the detailed definition in the paragraph "Well-formed partial applications" on pages 14 and 15.
  *   In Figure 7, in the rules for application of a partial builtin applications to a value, the condition V ~ next(A) has been replaced by next(A) ∈ U_* ∪ V_#, and similarly in the definition of the CEK machine in Figure 10.

The main effect of this is that an application like [(builtin modInteger) (con string "blue")] would previously have caused an immediate error but is now a valid value. An error will occur when (and only when) a second argument is supplied.

See also https://github.com/input-output-hk/plutus/blob/0fe1e41596caec0e9a02c65d1c1573d78c4ce019/doc/read-the-docs-site/adr/0004-deferred-unlifting.md


kwxm:
This has already been reviewed elsewhere, so I'll merge it without further review.

PR #4961, opened 2022-11-07
Generate plutus-metatheory Intermediate .hs files in dist-newstyle

zeme-iohk:


plutus-metatheory/plutus-metatheory.cabal
```diff
@@ -251,197 +260,6 @@ library
     MAlonzo.Code.Untyped.RenamingSubstitution
     MAlonzo.Code.Utils
     MAlonzo.RTE
-    Opts
-    Raw
-    Scoped
-    Untyped
-
-  autogen-modules:
```
michaelpj, line 259: Are you sure we can remove this? It sounds like we should have to keep it: https://cabal.readthedocs.io/en/stable/cabal-package.html#pkg-field-autogen-modules

plutus-metatheory/Setup.hs
```diff
@@ -1,49 +1,109 @@
...
 
-The only way is to rely on Agda to do the change detection, and run it every time.
-That means we need a pre-build hook. But then it won't run before Haddock!
+Finally, the order in which the modules are listed in exposed-modules matters a lot!
+The MAlonzo.Code files must be listed first, otherwise cabal will fail with:
+setup: can't find source for MAlonzo/Code/* in src
 
-So: we do both! Since Agda is reasonably good at not redoing work, we mostly don't
-pay the cost twice on the first configure/build sequence.
+TODO Newer (> 3.6.3.0) versions of the cabal library introduced a ppOrdering field in PreProcessor
```
michaelpj, line 46: cool

plutus-metatheory/Setup.hs
```diff
@@ -1,49 +1,109 @@
...
+import Distribution.Simple.Program.Types qualified as D
+import Distribution.Types.BuildInfo qualified as D
+import Distribution.Types.ComponentLocalBuildInfo qualified as D
+import Distribution.Types.LocalBuildInfo qualified as D
+import Distribution.Verbosity qualified as D
+
+import Data.IORef (IORef, newIORef, readIORef, writeIORef)
+import System.IO.Unsafe (unsafePerformIO)
+
+{-
```
michaelpj, line 15: great comment

plutus-metatheory/Setup.hs
```diff
@@ -1,49 +1,109 @@
...
-Agda changes.
+Again if more than one .lagda file was modified, and we run cabal build, we only need to
+invoke agda once, with all subsequent invokations being noops, but still kind of slow.
+This is why we use the agdaProgramStatus IORef: to cut down compilation times.
 
-The only way is to rely on Agda to do the change detection, and run it every time.
-That means we need a pre-build hook. But then it won't run before Haddock!
+Finally, the order in which the modules are listed in exposed-modules matters a lot!
+The MAlonzo.Code files must be listed first, otherwise cabal will fail with:
+setup: can't find source for MAlonzo/Code/* in src
```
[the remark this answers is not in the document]
michaelpj: I can't figure out how this can happen without everything being broken all the time :sweat_smile:

plutus-metatheory/Setup.hs
```diff
@@ -1,49 +1,109 @@
...
--- Doing this before building ensures that the generated Haskell files are always up-to-date, since cabal won't
--- track the dependency on the source Agda files, but Agda should do
-preBuild' :: Args -> BuildFlags -> IO ()
-preBuild' _ _ = runAgda
+    runAgda :: IO ()
+    runAgda = D.runProgram D.verbose agdaProgram
+      [ "--compile-dir", D.buildDir lbi
+      , "--compile"
+      , "--ghc-dont-call-ghc"
+      , "--local-interfaces"
```
michaelpj, line 104: At the moment, this flag means that we put the generated Agda interface files next to the source files. Really, those are also auto-generated and would ideally go in `dist-newstyle` also. I don't know if there's a way to stop it doing that. I tried removing this option and it still put them next to the source files :thinking:

michaelpj (approved):
Nice!

thealmarty (approved):
Verified that this fixed the issue.

plutus-metatheory/Setup.hs
```diff
@@ -1,49 +1,109 @@
...
+  { D.platformIndependent = True
+  , D.runPreProcessor = D.mkSimplePreProcessor preProcessors
+  }
+  where
+    preProcessors :: FilePath -> FilePath -> D.Verbosity -> IO ()
+    preProcessors _ _ _ = do
+      status <- readIORef agdaProgramStatus
+      case status of
+        NotRun -> do
+          putStrLn "***** running agda preprocessor..."
```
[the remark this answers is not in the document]
zeme-iohk: `D.notice` is the only one that works (I've tried `D.warn` and `D.debug` but they don't print anything)

plutus-metatheory/plutus-metatheory.cabal
zeme-iohk: From the docs: commands like sdist will fail if the autogenerated modules are not listed in `autogen-modules`.
This is irrelevant to us I think? 
Still, I put them back just in case

plutus-metatheory/Setup.hs
zeme-iohk: Yes `--local-interfaces` seems to be a noop for me also. 🤔

plutus-metatheory/Setup.hs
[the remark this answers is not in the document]
michaelpj: Well, `notice` seems like the right log level anyway?

plutus-metatheory/plutus-metatheory.cabal
michaelpj: Unclear to me, maybe we want to sdist occasionally :thinking: I think it's good hygiene to retain sdist-ability.

plutus-metatheory/Setup.hs
michaelpj: Seems like it's something to do with some agda-specific build directory: https://github.com/agda/agda/pull/3987/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR9

PR #4963, opened 2022-11-07
PLT-1130: Add a Note on why we need GHC's PreInlineUnconditionally

zliu41:


plutus-tx-plugin/src/PlutusTx/Plugin.hs
```diff
@@ -121,6 +121,22 @@ plugin = GHC.defaultPlugin { GHC.pluginRecompile = GHC.flagRecompile
              : pluginPass
              : rest
 
+{- Note [GHC.sm_pre_inline]
+We run a GHC simplifier pass before the plugin, in which we turn on `sm_pre_inline`. The
```
michaelpj, line 125: I'm kind of confused here. We set `sm_inline=False`, and this new `sm_pre_inline` flag only appeared in GHC 9.2. Did they split `PreInlineUnconditional` into its own pass or something?

zliu41: I don't think `sm_inline` affects PreInlineUnconditionally, rather it only affects call site inlining. In GHC 8, PreInlineUnconditionally is only controlled by `Opt_SimplPreInlining` in DynFlags, and it is on by default.

kwxm:
^ Sorry for intruding, I was just looking for a PR that I could try this in.

zliu41:
@kwxm there's no [agent](https://buildkite.com/organizations/input-output-hk/agents) in the plutus queue to run the benchmarking job.

kwxm:
> @kwxm there's no [agent](https://buildkite.com/organizations/input-output-hk/agents) in the plutus queue to run the benchmarking job.

There was a `buildkite-agent` running on the benchmarking machine that vanished some months ago but reappeared a few days ago (I have no idea why), so I was curious to see if the whole thing worked now. Apparently not.  Thanks, and I'll leave your PR alone now.

plutus-tx-plugin/src/PlutusTx/Plugin.hs
michaelpj: I guess... I still feel confused after reading this? Perhaps we need a note with a slightly broader purpose that explains all the inliner settings we're setting and why, which sadly probably requires explaining a bit more of what GHC does? it seems like you know this already, so maybe we can just write it down

zliu41: @michaelpj Please see if this is clearer.

michaelpj (approved):
Nice!

PR #4964, opened 2022-11-09
Follow the Cardano Engineering Handbook

michaelpj:
If you haven't already, please take a read and check that you're okay with everything! https://input-output-hk.github.io/cardano-engineering-handbook/

- Adds a badge showing that we do
- Fixes a few issues with licenses
- Adds a section on project roles and responsibilities - this is just my read of the "social truth", we can modify/keep it up to date as we go!


CONTRIBUTING.adoc
```diff
@@ -610,3 +610,18 @@ The "details" link for the task goes to a page for the entire action and you wil
...
+
+- The regular contributors to the Haskell code, all of whom can review and merge PRs are:
+    - @michaelpj
+    - @effectfully
+    - @kwxm
+    - @bezirg
+    - @thealmarty
+    - @zliu41
+- The maintainer of the documentation is @joseph-fajen.
+- The maintainer of the Agda code is @jmchapman.
```
thealmarty, line 624: this may change soon?

zliu41 (approved):

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1 @@
+See https://github.com/input-output-hk/cardano-engineering-handbook/blob/main/CODE-OF-CONDUCT.md
```
thealmarty, line 1: 👍🏾

thealmarty (approved):

thealmarty:
Do we support windows and do we say we support it? The handbook said that we should say what OS we support.

michaelpj:
> Do we support windows and do we say we support it?

We support it in that we cross-compile to it. I don't think the Handbook says we have to state that, I think it just says we have to exercise it in our CI. Maybe we should suggest that people state it? Although ideally you do that in some standard way in your package metadata or something :thinking:

zeme-iohk (approved):

CONTRIBUTING.adoc
michaelpj: Yeah, I wasn't sure whether it was more accurate to say this or to say that it's unmaintained. I think in practice this is still accurate: if we had a big problem with it we'd go running for James!

effectfully (approved):
I'll review the Cardano Engineering Handbook separately.

CONTRIBUTING.adoc
thealmarty: also there's a new person?

CONTRIBUTING.adoc
```diff
@@ -610,3 +610,18 @@ The "details" link for the task goes to a page for the entire action and you wil
...
+
+== Project roles and responsibilities
+
+- The regular contributors to the Haskell code, all of whom can review and merge PRs are:
+    - @michaelpj
+    - @effectfully
+    - @kwxm
+    - @bezirg
+    - @thealmarty
+    - @zliu41
```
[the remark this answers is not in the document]
michaelpj: I think we could do, but I don't know how helpful it is :thinking: Probably if we wanted to go to the effort of doing that, we should use CODEOWNERS or similar, then it actually does something useful too. (I wouldn't suggest that we _require_ reviews from CODEOWNERS, but we could use it to suggest people).

CONTRIBUTING.adoc
michaelpj: yeah, but they won't be doing much for a bit. we can change this!

CONTRIBUTING.adoc
```diff
@@ -610,3 +610,18 @@ The "details" link for the task goes to a page for the entire action and you wil
...
+- The maintainer of the documentation is @joseph-fajen.
+- The maintainer of the Agda code is @jmchapman.
+- If you have a technical dispute that you need help resolving, you can ask @michaelpj.
```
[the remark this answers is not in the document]
michaelpj: I felt somewhat uncomfortable writing this, but I think it's the truth and I think it's useful to state it.

PR #4965, opened 2022-11-09
publish documents in CI

dermetfan:
- move tasks into automation cell
- simplify CI task (just run `nix build` instead of using `flakeOutputTasks`)
- extract input names to separate file to make it harder to mess them up
- add action that uploads the PDFs to cicero's artifact storage
  - [successful run](https://cicero.ci.iog.io/run/776e759c-22f7-4653-8baa-e032906a3896)

Pre-submit checklist:
- Branch
    - ~~Tests are provided (if possible)~~
    - ~~Relevant tickets are mentioned in commit messages~~
- PR
    - ~~(For external contributions) Corresponding issue exists and is linked in the description~~
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


flake.nix
```diff
@@ -153,7 +153,7 @@
       }
       (inputs.tullia.fromStd {
         actions = inputs.std.harvest inputs.self [ "cloud" "actions" ];
-        tasks = inputs.std.harvest inputs.self [ "cloud" "pipelines" ];
+        tasks = inputs.std.harvest inputs.self [ "automation" "pipelines" ];
```
michaelpj, line 156: What's the rationale? Maybe we need to update the comments above explaining the purposes of the cell blocks?

nix/cells/automation/pipelines.nix
```diff
@@ -0,0 +1,73 @@
...
+    preset.github-ci.sha =
+      config.preset.github-ci.lib.readRevision inputs.cells.cloud.library.actions.ci.input null;
+
+    command.text = ''
+      nix build -L .#${lib.escapeShellArg system}.automation.ciJobs.required
+    '';
+  };
+
+  publish-unraveling-recursion-paper = { config, pkgs, lib, ... }: {
+    imports = [ common ];
```
michaelpj, line 42: This is potentially going to provision a massively OTT machine for just publishing an artifact, unsure if that's a problem or those resource requirements are real or what.

nix/cells/automation/pipelines.nix
```diff
@@ -0,0 +1,73 @@
...
+in
+{
+  ci = { config, lib, ... }: {
+    imports = [ common ];
+
+    preset.github-ci.sha =
+      config.preset.github-ci.lib.readRevision inputs.cells.cloud.library.actions.ci.input null;
+
+    command.text = ''
+      nix build -L .#${lib.escapeShellArg system}.automation.ciJobs.required
```
michaelpj, line 37: :+1:

nix/cells/automation/pipelines.nix
```diff
@@ -0,0 +1,73 @@
...
+
+      jq > value.json --null-input \
+        --arg revision "$(${lib.escapeShellArg config.preset.github-ci.sha})" \
+        '{"plutus/unraveling-recursion-paper": $revision}'
+
+      http --multipart --ignore-stdin "$CICERO_WEB_URL/api/run/$NOMAD_JOB_ID/fact" \
+        value@value.json binary@result/unraveling-recursion.pdf
+    '';
+
+    dependencies = with pkgs; [ jq httpie ];
```
michaelpj, line 60: Why not just splice these into `command.text`?

nix/cells/cloud/library.nix
```diff
@@ -0,0 +1,8 @@
+{ cell
+, inputs
+}: {
+  actions = {
```
michaelpj, line 4: I don't really understand the purpose of this file. What difference does this make? Comments please!

nix/cells/cloud/actions.nix
```diff
@@ -21,4 +19,26 @@
...
+    task = "publish-unraveling-recursion-paper";
+    io = ''
+      let push = {
+        #lib.io.github_push
+        #input: "${cell.library.actions.paper.input}"
+        #repo: "input-output-hk/plutus"
+        inputs: _final_inputs
+      }
+
+      _final_inputs: inputs
```
michaelpj, line 33: this all looks weird but I assume you know what you're doing. Maybe could do with some comments :)

nix/cells/cloud/actions.nix
```diff
@@ -21,4 +19,26 @@
...
+      inputs: {
+        push.inputs
+
+        "CI passed": match: {
```
michaelpj, line 37: This means this won't run unless the CI job passed, I assume? What links it to that? I don't see anything that sets "ok" in the CI action? Should the CI action publish some specific fact?

nix/cells/cloud/library.nix
zeme-iohk: Seems like we can just hard-code those strings in, which I would be in favour of.

nix/cells/automation/pipelines.nix
```diff
@@ -0,0 +1,73 @@
...
+  publish-unraveling-recursion-paper = { config, pkgs, lib, ... }: {
+    imports = [ common ];
+
+    preset.github-ci.sha =
+      config.preset.github-ci.lib.readRevision inputs.cells.cloud.library.actions.paper.input null;
+
+    command.text = ''
+      set -x
+
+      nix build -L .#${lib.escapeShellArg system}.plutus.packages.unraveling-recursion-paper
```
zeme-iohk, line 50: Here is the full list of papers.
Note that we can do without the `${lib.escapeShellArg system}`
```



nix build .#plutus-report
nix build .#plutus-core-spec
nix build .#extended-utxo-spec

nix build .#unraveling-recursion-paper
nix build .#system-f-fin-agda-paper
nix build .#eutxo-paper
nix build .#utxoma-paper
nix build .#eutxoma-paper
```

michaelpj: > Note that we can do without the ${lib.escapeShellArg system}

That's because the flake CLI magically inserts it for you, though? I'd rather be explicit, tbh.

nix/cells/cloud/library.nix
dermetfan: Do so if you want, that's fine with me. This is just to extract some common strings to variables to avoid possible future typos.

michaelpj: That's fine, but worth a comment. I also don't know why these magic strings are what they are :grimacing:

flake.nix
dermetfan: Tullia tasks are in the automation cell in every other stdized repo and [this](https://github.com/input-output-hk/plutus/blob/3a7af5130920661567de0de8fe73fa4d22276201/nix/cells/cloud/pipelines.nix#L9-L10) comment indicated that it was only in the cloud cell because it had to be, but that condition has been fixed.

The logic behind that is that tullia tasks can be run locally (not in the cloud), so they belong in the automation cell. Cicero actions can only run in the cloud (and happen to run tullia tasks) so they belong in the cloud cell.

michaelpj:
How do we get a link to the artifact? For Hydra we had a stable link that always linked to the latest version of the artifact, which we could just stick in the readme and forget about.

nix/cells/automation/pipelines.nix
dermetfan: To have a normal-looking script instead of `${lib.getExe pkgs.jq}`, personal preference I guess.

nix/cells/cloud/actions.nix
dermetfan: `#lib.io.github_push` publishes

```
{
  "ok": true,
  "revision": …
}
```

[by default](https://github.com/input-output-hk/tullia/blob/86c97fe01e69070ae7f68e745f9bf6335f483711/lib/github.cue#L116-L122).

nix/cells/cloud/library.nix
dermetfan: They are just arbitrary names used to refer to the input fact in the tullia task.

nix/cells/automation/pipelines.nix
dermetfan: Right, the publish task should need much less because it can just pull the artifact from the cache.

nix/cells/automation/pipelines.nix
```diff
@@ -0,0 +1,73 @@
...
+
+    dependencies = with pkgs; [ jq httpie ];
+
+    nomad.templates = [
+      {
+        destination = "${config.env.HOME}/.netrc";
+        data = ''
+          machine cicero.ci.iog.io
+          login cicero
+          password {{with secret "kv/data/cicero/api"}}{{.Data.data.basic}}{{end}}
```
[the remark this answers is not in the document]
dermetfan: Yes but not now, priority is way too low

nix/cells/automation/pipelines.nix
dermetfan: report and core spec are the exact same:

```
❯ diff ./nix/cells/plutus/packages/plutus-*
4,5c4,5
<   name = "plutus-core-spec";
<   description = "Plutus core specification";
---
>   name = "plutus-report";
>   description = "plutus report";
```

michaelpj: Well, that's a bug.

However, I mostly want a parametric version of this. Then you can leave it to us to make sure it's building the right set of artifacts :)

dermetfan:
We don't have stable links to artifacts yet, but will. For now...

... using the web UI:

1. go to cicero start page
2. click on action named plutus/publish-unraveling-recursion-paper
3. click on "Nomad Job ID" of the first entry in the "Runs" table
4. click on fact ID in top right in the facts table
5. click on binary sha

... scripting:

```
id=$(echo '{"plutus/document/unraveling-recursion-paper": string}' | http https://cicero.ci.iog.io/api/fact/match/latest | jq -r .id)
curl https://cicero.ci.iog.io/api/fact/"$id"/binary > paper.pdf
```

michaelpj:
How long do those artifacts exist for? When will one of those links stop working, effectively?

dermetfan:
Never. Unless we implement fact GC at some point, but we'll also add some way or rules to pin certain facts.

nix/cells/cloud/actions.nix
michaelpj: I should learn CUE I guess :thinking:

michaelpj:
> Never. Unless we implement fact GC at some point, but we'll also add some way or rules to pin certain facts.

Well, that's also somewhat scary in its own rights but okay :sweat_smile:

nix/cells/cloud/actions.nix
Pacman99: Does this not cause a cyclic dependency on action? Thats what happened to me when I tried to do something similar with the comment action.

nix/cells/automation/pipelines.nix
michaelpj: It's already got quite a lot of splicing garbage in there, but as you like :)

nix/cells/automation/pipelines.nix
```diff
@@ -0,0 +1,103 @@
...
+          jq > ${lib.escapeShellArg k}.json --null-input \
+            --arg revision "$revision" \
+            --arg document ${lib.escapeShellArg k} \
+            '{"plutus/document/\($document)": $revision}'
+
+          http --multipart --ignore-stdin "$CICERO_WEB_URL/api/run/$NOMAD_JOB_ID/fact" \
+            value@${lib.escapeShellArg k}.json \
+            binary@result${
+              lib.optionalString (i != 0) "-${toString i}"
+            }/${lib.escapeShellArg documents.${k}}
```
michaelpj, line 79: Took me quite a lot of staring to work out what's going on here. You're getting the `result` symlinks and using them as the things to provide as the result, I think? Are we guaranteed that they come out in the right order? That's quite non-obvious to me, and I don't see it guaranteed anywhere in the docs.

I wonder if it would be simpler to just do the whole operation once per document, including calling `nix build`. Then we can use `-o` to get a specific known output link.

Also something it took me a while to realise and I think is maybe worth documenting: I was wondering why we can't just splice in a reference to the document here and avoid doing this convoluted dance. But then I realised that that would make us build the doc when we build this task definition, which is _not_ when we want it to happen. So we have to do it like this.

michaelpj: That said, this has lingered on a bit and we can refactor this afterwards I guess.

dermetfan: > Are we guaranteed that they come out in the right order?

AFAIK yes.

`nix build .#{a,b,c}` will result in

```
result → /nix/store/…-a
result-1 → /nix/store/…-b
result-2 → /nix/store/…-c
```

This is documented in `nix build --help` in the "Description" section.

zeme-iohk:
I've updated the links in the README.adoc
Enabling auto-merge now.

PR #4969, opened 2022-11-15
Add full R setup to the devshell

zeme-iohk:


zeme-iohk:
As requested by @bezirg

nix/cells/plutus/library/make-plutus-shell.nix
```diff
@@ -143,9 +143,8 @@ inputs.std.lib.dev.mkShell {
     pkgs.cacert
     pkgs.pkg-config # TODO(std) Keep an eye on https://github.com/input-output-hk/plutus/pull/4906
 
-    # More R libraries
-    pkgs.rPackages.plotly
-  ];
+    # R libraries
+  ] ++ cell.library.r-with-packages;
```
michaelpj, line 147: IMO this name doesn't follow the usual conventions. I'd call it `r-packages`. I expect `x-with-packages` to be a single derivation for `x` wrapped to have some packages available. This is just a list. I did a double-take when I saw this line because I thought "surely you can't be using ++ on that since it's just a single derivation". Yes, if only we had types, etc....

michaelpj (approved):
One naming nit otherwise LGTM.

bezirg:
does not work for me yet, so LNGTM

bezirg:
works now, LGTM

bezirg (approved):

nix/cells/plutus/library/make-plutus-shell.nix
zeme-iohk: Latest commit addresses this 👍

PR #4970, opened 2022-11-16
PLT-1057: update R and inline-r. Bump hackage-nix haskell-nix

bezirg:
Tested with&without nix, with ghc9.2 & ghc 8.10 and R 4.2, R 4.1.

I did not test on Mac and Win. Perhaps it does not work in Win.

Bumping hackage-nix was not required, but I left it there.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -133,8 +133,6 @@ If you use Nix, these tools are provided for you via `nix develop`, and you do *
 ====
 * We rely on forked or new versions of some system libraries.
 ** You can read the https://github.com/input-output-hk/cardano-node/blob/master/doc/getting-started/install.md[Cardano node documentation] to find out how to install these.
-* You may need an old version of R installed to build everything
-** If you don't want to do this and you don't need to build those components, you can turn off the `plutus-core` `with-inline-r` flag in your `cabal.project.local`.
 * You may get different versions of packages.
 ** This *shouldn't* happen, but we can't guarantee it.
 * We are not currently enabling the Nix integration for these tools, so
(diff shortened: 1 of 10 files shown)


bezirg:
This is the patch to make inline-r work for our nix:

https://github.com/bezirg/HaskellR/commit/3e7498be0f3c01cd80e5435ae1f2ccc94ad6b0df

It is ugly and there might be a better way. I don't know if we should upstream it to tweag/HaskellR

michaelpj:
Maybe the problem is that we don't set `R_LIBS_SITE`? I think I'd rather work around that on our side, i.e. set `R_LIBS_SITE` for the derivations that use R and the shell, then we don't have to fork `inline-r`. And maybe having the `r-with-packages` as a build input should set that variable?

michaelpj:
I see a bunch of stuff relating to `R_LIBS_SITE` in nixpkgs, so maybe it's already doing that?

bezirg:
@zeme-iohk told me that this is easier. I know it is not the best approach.  @michaelpj if you know how to do it please go ahead, my nix skills are not good

michaelpj:
Maybe we need to tell nix to build `inline-r` with our wrapped R and not the unwrapped one from nixpkgs? That would make sense.  But if the R used by `inline-r` will pick up libs from the lib path, then setting that environment up properly should also fix it?

michaelpj:
If you think that it'll be accepted upstream then I'm fine with it. I just want a) an explanatory comment since it sounds like this was quite tricky to figure out, b) some pathway to not depending on a patched version, either by it going upstream or by us working around the need for the patch.

kwxm:
FYI I went through [the process of adding a new builtin](https://github.com/input-output-hk/plutus/blob/master/plutus-core/cost-model/CostModelGeneration.md#adding-a-new-built-in-function-and-a-suitable-costing-function) in this branch (in connection with another issue) and all of the R stuff appeared to work correctly.

bezirg:
ready to review again

michaelpj (approved):
Great

PR #4973, opened 2022-11-21
Typo in the comments (ECDSA)

iquerejeta:
The public key of `verifyEcdsaSecp256k1Signature` is 33 bytes rather than 32. The rest of the description is correct.


kwxm:
Thanks for spotting this @iquerejeta.  I think that in fact it previously had the corrected figure of 33 bytes and I "fixed" it to say 32 bytes, thinking that it was a typo.  I checked the rest of the codebase and other references to this are correct, but the Plutus Core specification still talks about uncompressed 64 byte keys.  I will fix that and ask you to check it later.

kwxm (approved):
I agree that this is correct.  We're having a problem with CI at the moment but I'll merge it when we've worked out what's going on.

PR #4974, opened 2022-11-21
Add hlint support in shell and GHA

zeme-iohk:
The HLint GHA started failing with:
```
Run haskell/actions/hlint-run@v1
hlint
  Running hlint --hint=./.github/.hlint.yaml -j --json -- .
  /opt/hostedtoolcache/hlint/3.4/x64/hlint: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
  hlint completed with status code 127
Error: SyntaxError: Unexpected end of JSON input
```

This is outside our control.
In this PR, we replace the hlint coming from GH with the hlint coming from the nix shell.


.github/workflows/hlint.yml
```diff
@@ -12,11 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: haskell/actions/hlint-setup@v2
-        with:
-          version: '3.4'
-      - uses: haskell/actions/hlint-run@v1
-        with:
-          hlint-bin: "hlint --hint=./.github/.hlint.yaml"
-          fail-on: status
```
michaelpj, line 21: This is important! This is how it worked in the first place! I think maybe we should instead do something like
```
nix build .#hlint
...
hlint-bin: "./result/bin/hlint"
```
or something

michaelpj: We do want to use `hlint-run`. It's doesn't just run hlint, it turns the output into github checks output.

PR #4975, opened 2022-11-21
Update comment on cost model tests

kwxm:
PLT-603 is an issue to check that the order of [steps 8 and 9 ](https://github.com/input-output-hk/plutus/blob/master/plutus-core/cost-model/CostModelGeneration.md#step-8-test-the-haskell-versions-of-the-costing-functions) in `CostModelGeneration.md` is correct.  I think that this issue was prompted by a question (some time ago) from @zliu41.

I believe that the order is correct (run the tests _before_ updating the cost model file), but I got a bit confused myself so I've expanded a comment a bit to clarify what's going on, and I'm also appending a detailed comment from PLT-603 below.  I've also added a new issue at PLT-1250 to add  test that cost model serialisation/deserialisation is correct, which would fill in the gap in the testing mentioned below.  

> As explained [here](https://github.com/input-output-hk/plutus/blob/cfc26a48f9151bb0940c9e4bf04638161cfc3c3f/plutus-core/cost-model/test/TestCostModels.hs#L52), the tests in `plutus-core:cost-model-test` run the R modelling code which reads the benchmark results and fits costing functions to them; they then read the costing functions from the R process, convert them into the Haskell objects which represent costing functions, and compare the result of  evaluating those functions (on random inputs) with the result of feeding the same inputs (converted from costing integers to real numbers) to the predictions of the models in the R process. 
> 
> When the cost model is updated, it runs the R code, reads the modelling functions from the R process into Haskell objects (as above) and then serialises those objects into JSON in `builtinCostModel.json`.  Subsequently the evaluator reads that file (or receives some equivalent data from the node) and reconstructs the objects representing the costing functions.  We don’t check that those costing functions are the same as the ones we get by reading the R models directly; we’re trusting that serialising to JSON and then deserialising again gives us back exactly the same thing, which could in principle fail because of, for example, loss of precision during the serialisation/deserialisation process.  Maybe we should check this as well: the comment linked to above already mentions this.  Doing the tests this way is arguably better because it may reveal problems in the costing interface before the cost model file is updated, and we want to be sure that we don’t include an incorrect costing function in the JSON.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-core/cost-model/CostModelGeneration.md
+++ b/plutus-core/cost-model/CostModelGeneration.md
@@ -295,7 +295,7 @@ Now a CPU usage benchmark for the function will have to be added in
 will have to be added in [`models.R`](./data/models.R) to process the results of
 the benchmark (see Step 6 below).  The benchmark should aim to cover a wide
 range of inputs in order to get a good idea of the worst-case behaviour of the
-function: some experimentation may be needed to acheive this.
+function: experimentation may be needed to achieve this.
 
 Once the benchmark is in its final form, run it using `cabal run
 plutus-core:cost-model-budgeting-bench -- --csv <file>` as described in the
@@ -333,8 +333,8 @@ sometimes more complex code may be required to obtain a good model of the
 behaviour of the function. Ideally the R model should accurate over a wide range
 of inputs so that charges for "typical" inputs are reasonable but worst-case
 inputs which require large computation times incur large charges which penalise
-excessive computation. Some experimentation may be required to achieve this, and
-it may not always be possible to satisfy both goals simultaneously. In such
+excessive computation. Developing the model may involve some experimentation,
+and it may not always be possible to satisfy both goals simultaneously. In such
 cases it may be necessary to sacrifice some accuracy in order to guarantee
 security.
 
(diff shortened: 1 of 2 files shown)


zliu41:
I can no longer recall the exact problem I ran into. I guess I talked to you in person and didn't write it down. I don't remember it having to do with `builtinCostModel.json`. More like one step (step 8?) needed the updated `bench.csv`, but `bench.csv` wouldn't get updated till the next step (step 9?), and the error was like "bench.csv does not have the required data for the new builtin". Does that sound like something that could have happened?

kwxm:
> I can no longer recall the exact problem I ran into.

Neither can I! This has been stuck in the backlog for a long time.

> I don't remember it having to do with builtinCostModel.json. More like one step (step 8?) needed the updated bench.csv, but bench.csv wouldn't get updated till the next step (step 9?), and the error was like "bench.csv does not have the required data for the new builtin". Does that sound like something that could have happened?

I'll think about this some more. Maybe we should get some unsuspecting guinea-pig to try it and see if something goes wrong ...

zliu41 (approved):
We can merge this and consider it done for now. Next time if someone runs into a problem we can revisit it.

kwxm:
> More like one step (step 8?) needed the updated bench.csv, but bench.csv wouldn't get updated till the next step (step 9?), and the error was like "bench.csv does not have the required data for the new builtin". Does that sound like something that could have happened?

Ah, you were right about that.  Step 5 said to add a new benchmark but there was nothing to tell you to run it, so the numbers wouldn't be there when the tests tried to read them.  I've added some instructions there and swapped steps 6 and 7: now the order is

* add the benchmark
* run it
* update the R code
* update the Haskell code that reads stuff from R
* test that the Haskell and R models give similar results
* update the cost model.

kwxm:
I think I fixed the original problem, so I'll merge this.

PR #4977, opened 2022-11-24
Josfaj/plt 787 new tutorial

joseph-fajen:
This is a first draft of a revised English Auction tutorial that Ziyang and I are collaborating on. This originally appeared as part of a Plutus Pioneer Program tutorial. With the consent of Lars Brünjes, we have revised this example, changing some points of emphasis and scope, partly due to not relying on using the Plutus Playground. 

We still need to assess whether/how to mention ways that someone could try deploying and interacting with the auction smart contract in some environment (testnet, local node), referencing some Plutus Tools.


doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+In practice, you may consider using a higher-level library that abstracts away some of the details.
+For example, `plutus-apps <https://github.com/input-output-hk/plutus-apps>`_ provides a state machine library and a constraint library for writing Plutus Tx.
+Using these libraries, writing a validator in Plutus Tx becomes a matter of defining state transactions and the corresponding constraints, e.g., the condition ``refundsPreviousHighestBid`` can simply be written as ``Constraints.mustPayToPubKey bidder (singleton adaSymbol adaToken amt)``.
+
+Off-chain code
+-----------------------
+
+Since the main purpose of this tutorial is to introduce Plutus Tx and Plutus Core, we only walked through the on-chain code, responsible for validating transactions (in the sense of determining whether a transaction is allowed to spend an EUTXO).
+
+In addition to the on-chain code, one typically needs the accompanying off-chain code and services to perform tasks like building transactions, submitting transactions, deploying smart contracts, querying for available EUTXOs on the chain, etc.
```
koslambrou, line 340: I would also mention in a paragraph `cardano-api` which also provides (in a more low-level way) the capability to create, balancing and submit transactions with a local running node (with even some querying functions like getting the EUTXOs of a particular address).

koslambrou (commented):
The tutorial is the first one in the list of tutorial (before "Using Plutus Tx" and "Writing basic validator scripts"). Shouldn't it be after?

doc/read-the-docs-site/tutorials/AuctionValidator.hs
```diff
@@ -0,0 +1,194 @@
...
+    bAmount :: Integer
+  }
+
+PlutusTx.deriveShow ''Bid
+PlutusTx.unstableMakeIsData ''Bid
+
+instance PlutusTx.Eq Bid where
+  {-# INLINEABLE (==) #-}
+  bid == bid' =
+    bBidder bid PlutusTx.== bBidder bid'
```
michaelpj, line 44: I'm not sure why you're using all these operations qualified - would it not make more sense in a context like this to just import them unqualified?

doc/read-the-docs-site/tutorials/AuctionValidator.hs
```diff
@@ -0,0 +1,194 @@
...
+          -- The highest bidder gets the asset.
+          highestBidderGetsAsset
+        ]
+-- BLOCK3
+    sufficientBid :: Bid -> Bool
+    sufficientBid (Bid _ amt) =
+      maybe
+        (amt PlutusTx.>= apMinBid params)
+        (\(Bid _ amt') -> amt PlutusTx.> amt')
+        highestBid
```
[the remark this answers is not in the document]
michaelpj: Or perhaps even better since it avoids duplicating the logic:
```
amt >= fromMaybe (apMinBid params) highestBid
```

doc/read-the-docs-site/tutorials/AuctionValidator.hs
```diff
@@ -0,0 +1,194 @@
...
+    validBidTime = to (apEndTime params) `contains` txInfoValidRange txInfo
+-- BLOCK5
+    refundsPreviousHighestBid :: Bool
+    refundsPreviousHighestBid = case highestBid of
+      Nothing -> True
+      Just (Bid bidder amt) ->
+        case PlutusTx.filter
+          (\o -> txOutAddress o PlutusTx.== pubKeyHashAddress bidder)
+          (txInfoOutputs txInfo) of
+          [o] -> txOutValue o PlutusTx.== singleton adaSymbol adaToken amt
```
michaelpj, line 106: What if I was the previous highest bidder? Then there is likely another output going to my public key with my change from creating the current transaction. Worth a comment about the edge case even if we're not going to handle it. Probably good to draw attention to this sort of thing, approximately any time you make assertions that the list of outputs has some particular shape you need to be quite careful!

doc/read-the-docs-site/tutorials/AuctionValidator.hs
```diff
@@ -0,0 +1,194 @@
...
+              )
+
+    highestBidderGetsAsset :: Bool
+    highestBidderGetsAsset = case highestBid of
+      Nothing -> True
+      Just (Bid bidder _) ->
+        case PlutusTx.filter
+          ( \o ->
+              txOutAddress o PlutusTx.== pubKeyHashAddress bidder
+                PlutusTx.&& txOutValue o PlutusTx.== apAsset params
```
michaelpj, line 167: ditto

michaelpj (commented):
Just reviewed the code so far, will try to look at the text also later.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+
+The general logic of the auction takes into account the following parameters:
+
+* the owner of the token,
+* the token itself,
+* a minimal bid,
+* a higher bid, and
+* a deadline.
+
+At the outset, Alice creates a transaction for the English auction that has a EUTXO at the script output.
```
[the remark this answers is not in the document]
joseph-fajen: Good catch!

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+Bidder transactions
+----------------------
+
+In this example, Bob, the first bidder, wants to bid 100 Ada for Alice's NFT.
+In order to do this, Bob creates a transaction that has two inputs and one output.
+
+The two inputs are:
+
+* the auction EUTXO
+* Bob's bid of 100 Ada
```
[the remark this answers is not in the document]
joseph-fajen: Now that you mention it, I wonder if it is a good idea or not to ignore the fee requirement.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+
+Overview
+------------
+
+This tutorial describes the general logic, structure, and Plutus Tx code needed for designing and writing a Plutus smart contract that can be used for an English auction that is executed on the Cardano blockchain.
+
+Key aspects of this example are the EUTXO model, the concepts involved in designing the EUTXOs, the actions required of the parties involved with the auction transactions, and Plutus on-chain code examples.
+The specific on-chain code examples included here are accompanied by detailed explanations of what the Haskell code is doing.
+
+The EUTXO model
```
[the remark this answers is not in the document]
joseph-fajen: I like your suggestion. I'll update this with more specific references.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+
+Datum
+~~~~~~~~~
+
+The output is at the output script, but now the value and the datum have changed.
+Previously the datum was nothing but now it has the value of: (Bob, 100).
+
+Now there is not only the NFT in the EUTXO, but also the 100 Ada bid.
+
++----------------------+-------------------+
```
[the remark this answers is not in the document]
joseph-fajen: I was experimenting with the small tables to see if they would be an effective way to present the info. But I don't think they are terribly useful here.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+If a node receives a new transaction, it validates it before accepting it into its mempool and eventually into a block.
+For each input of the transaction, if that input happens to be a script address, the corresponding script is executed.
+If the script does not succeed, the transaction is invalid and does not get completed.
+
+Plutus Core
+~~~~~~~~~~~~~~
+
+The programming language this script is expressed in is called (Untyped) Plutus Core, but you never write Plutus Core by hand.
+Instead, you write Plutus Tx (a subset of Haskell) and that gets compiled down to Plutus Core.
+Eventually there may be other high-level languages such as Solidity, C or Python that can compile down to Plutus Core.
```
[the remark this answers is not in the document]
joseph-fajen: Are you saying that there are alternatives to Plutus Tx for writing the code that gets compiled to Plutus Core? What are those alternatives?

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+   The latter won't work, because everything being compiled (inside ``[||...||]``) must be known at compile time, but ``params`` is not: it is a formal parameter.
+   Instead, we compile the entire ``auctionUntypedValidator`` into UPLC, then use ``liftCode`` to lift ``params`` into a UPLC term, and apply the compiled ``auctionUntypedValidator`` to it at the UPLC level.
+   To do so, we need the ``Lift`` instance for ``AuctionParams``, derived via ``PlutusTx.makeLift``.
+
+Libraries for Writing Plutus Tx Scripts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This auction example shows a relatively low-level way of writing scripts using Plutus Tx.
+In practice, you may consider using a higher-level library that abstracts away some of the details.
+For example, `plutus-apps <https://github.com/input-output-hk/plutus-apps>`_ provides a state machine library and a constraint library for writing Plutus Tx.
```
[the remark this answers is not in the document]
joseph-fajen: Can you suggest an example that can be used in production? I would like to be able to point to something specific if possible.

doc/read-the-docs-site/tutorials/english-auction-example.rst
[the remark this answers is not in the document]
koslambrou: [Plutarch?](https://github.com/Plutonomicon/plutarch-plutus)

doc/read-the-docs-site/tutorials/english-auction-example.rst
joseph-fajen: Thanks for this suggestion. I'd like to get a better handle on how cardano-api is used to do these sorts of things. It would be great to be able to at least broadly describe a valid process of using cardano-api to create, balance and submit a transaction. Or perhaps point to another source that may already exist that describes it.

zliu41:
Converted to draft - I'll do a pass and make some revisions.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+| The auction EUTXO    | Bob               |
++----------------------+-------------------+
+| Bob's bid of 100 Ada | 100 Ada           |
++----------------------+-------------------+
+
+Redeemer
+~~~~~~~~~~~~
+
+As a redeemer, in order to unlock the original auction EUTXO, we use something called `Bid`.
+This is an algebraic data type. There will be other values as well.
```
michaelpj, line 94: I'm not sure what you're trying to convey here tbh?

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
++======================+===================+
+| The auction EUTXO    | Bob               |
++----------------------+-------------------+
+| Bob's bid of 100 Ada | 100 Ada           |
++----------------------+-------------------+
+
+Redeemer
+~~~~~~~~~~~~
+
+As a redeemer, in order to unlock the original auction EUTXO, we use something called `Bid`.
```
michaelpj, line 93:
```suggestion
As a redeemer, in order to unlock the original auction EUTXO, we use a value of the `Bid` type, which contains the information about who is bidding and how much they are bidding.
```

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+
+* `Architecting DApps on the EUTXO ledger <https://iohk.io/en/blog/posts/2021/11/16/architecting-dapps-on-the-eutxo-ledger/>`_
+* `The EUTXO Handbook <https://www.essentialcardano.io/article/the-eutxo-handbook>`_
+* `Concurrency and all that: Cardano smart contracts and the EUTXO model <https://iohk.io/en/blog/posts/2021/09/10/concurrency-and-all-that-cardano-smart-contracts-and-the-eutxo-model/>`_
+* Cardano's Extended UTXO accounting model---built to support multi-assets and smart contracts:
+
+   * `Part 1 <https://iohk.io/en/blog/posts/2021/03/11/cardanos-extended-utxo-accounting-model/>`_
+   * `Part 2 <https://iohk.io/en/blog/posts/2021/03/12/cardanos-extended-utxo-accounting-model-part-2/>`_
+
+Conceptualizing the English auction in the EUTXO framework
```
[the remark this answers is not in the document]
michaelpj: Er, probably make the sequence diagram correspond to the worked example later, of course.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+* a minimal bid,
+* a higher bid, and
+* a deadline.
+
+At the outset, Alice creates a transaction for the English auction that has a EUTXO at the script output.
+
+The value of the EUTXO is the NFT, and the datum initially is nothing.
+Later it will be the highest bidder and the highest bid. But at this stage, a bid has not yet taken place.
+
+   *On the Cardano blockchain you can’t have a EUTXO that contains only native tokens. A EUTXO always has to be accompanied by some Ada, but for the sake of simplicity we will temporarily ignore that requirement for now.*
```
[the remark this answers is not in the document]
michaelpj: You already use one below! I guess just you might want to use one here, then.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+However, because there is no bidder, the NFT doesn’t go to the highest bidder, but instead simply goes back to Alice.
+
+The logic in this case is slightly different.
+It will check that the NFT goes back to Alice.
+However, it doesn’t need to check the recipient because the transaction will be triggered by Alice and she can send the NFT wherever she wants.
+
+Diagram/graphic of how the smart contract works
+--------------------------------------------------
+
+*Assess what sort of diagrams would be most helpful.*
```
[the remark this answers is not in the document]
michaelpj: I think you can just inline a pretty easy sequence diagram, see my suggestion above, and I suggest doing it right at the beginning.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+* the auction EUTXO
+* Charlie's bid of 200 Ada
+
+The two outputs are:
+
+* the updated auction EUTXO
+* a EUTXO that returns Bob's bid of 100 Ada
+
+.. note::
+    To be more clear, technically, the auction EUTXO is not getting updated because nothing ever changes.
```
michaelpj, line 120:
```suggestion
    To be more clear, technically, the auction UTXO is not getting updated because UTXOs are immutable - they always refer to one specific output of one specific transaction.
```

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
++==========================+=========================================+
+| The auction EUTXO        | Newly created unspent transaction       |
++--------------------------+-----------------------------------------+
+| Charlie's bid of 200 Ada | EUTXO that returns Bob's bid of 100 Ada |
++--------------------------+-----------------------------------------+
+
+Bid redeemer logic
+~~~~~~~~~~~~~~~~~~~~~
+
+This time we again use the `Bid` redeemer. The script has to check that these conditions are satisfied:
```
michaelpj, line 135:
```suggestion
This time we again use the `Bid` redeemer containing the information about Charlie's bid. The script has to check that these conditions are satisfied:
```

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+Finally, we need to consider the scenario in which nobody bids in the auction.
+Alice creates the auction, but the auction receives no bids.
+In this case, there must be a mechanism for Alice to retrieve her NFT.
+
+To address this possibility, Alice creates a transaction with the `Close` redeemer.
+However, because there is no bidder, the NFT doesn’t go to the highest bidder, but instead simply goes back to Alice.
+
+The logic in this case is slightly different.
+It will check that the NFT goes back to Alice.
+However, it doesn’t need to check the recipient because the transaction will be triggered by Alice and she can send the NFT wherever she wants.
```
michaelpj, line 179: In the previous sentence you said it checks that the NFT goes back to Alice, then you say it doesn't check the recipient, which contradict each other. It _could_ not check that the NFT goes to Alice _if_ it checks that Alice signed the transaction, but you didn't say that.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+Eventually there may be other high-level languages such as Solidity, C or Python that can compile down to Plutus Core.
+
+The task of a script is to say yes or no regarding whether a transaction can consume an output.
+
+Next, we'll go over the Plutus Tx code for the auction validator.
+The full source code can be found in `AuctionValidator.hs <https://github.com/input-output-hk/plutus/blob/master/doc/read-the-docs-site/tutorials/AuctionValidator.hs>`_.
+
+Data Types
+~~~~~~~~~~~~~~~~~~~~~~~
+
```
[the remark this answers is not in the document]
michaelpj: Or alternatively give a link to the full source?

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+First, let's define the following data types and instances for the validator:
+
+.. literalinclude:: AuctionValidator.hs
+   :start-after: BLOCK1
+   :end-before: BLOCK2
+
+The purpose of ``makeLift`` and ``unstableMakeIsData`` will be explained later.
```
michaelpj, line 218: You could actually put them later in the program, then?

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+   * `Part 1 <https://iohk.io/en/blog/posts/2021/03/11/cardanos-extended-utxo-accounting-model/>`_
+   * `Part 2 <https://iohk.io/en/blog/posts/2021/03/12/cardanos-extended-utxo-accounting-model-part-2/>`_
+
+Conceptualizing the English auction in the EUTXO framework
+-------------------------------------------------------------
+
+On the Cardano blockchain, a transaction contains an arbitrary number of inputs and an arbitrary number of outputs.
+The effect of a transaction is to consume inputs and produce new outputs.
+
+In this English auction example, the item owner, Alice, wants to auction an NFT (Non-fungible token), a native token on Cardano that exists only once.
```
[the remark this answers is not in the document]
joseph-fajen: Good question. I had to look it up myself. There are very many types of auctions, but the "English auction" is what I'm most familiar with. It's basically an open public auction where the current high bid is known by all the bidders, and the bidding starts low and gets higher until the auctioneer says the time is up. We may be able to just call this a common public auction and leave it at that.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+An NFT can represent a piece of digital art or possibly a real-world asset.
+
+The general logic of the auction takes into account the following parameters:
```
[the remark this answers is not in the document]
joseph-fajen: It is intended to be pretty straight forward. I'll try to simplify/make it more clear. Maybe removing the word "general" would help. I'll review.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,344 @@
...
+Now we are ready to introduce our main validator function.
+The beginning of the function looks like the following:
+
+.. literalinclude:: AuctionValidator.hs
+   :start-after: BLOCK2
+   :end-before: BLOCK3
+
+Depending on whether this transaction is attempting to submit a new bid or request payout, the validator validates the corresponding set of conditions.
+
+The ``sufficientBid`` condition verifies that the bid amount is sufficient:
```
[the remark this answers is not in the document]
zliu41: See https://github.com/input-output-hk/plutus/pull/4977/files#r1036154887

doc/read-the-docs-site/tutorials/english-auction-example.rst
zliu41: The only option would be to put it at the beginning of the next block, but that feels awkward and distracting.

doc/read-the-docs-site/tutorials/english-auction-example.rst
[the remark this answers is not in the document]
koslambrou: For production, use can use `cardano-api`, `cardano-ledger` packages or the Contract API in plutus-apps (but not the State Machine).

The state machine auto generates Plutus on-chain code depending on the constraints used by the user. However, the on-chain code is REALLY large (with also large ExMem and ExCPU), so it won't run with mainnet's current config.

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,332 @@
...
+The frontend and backend of the off-chain part of smart contracts are not covered in this page.
+
+Before we get to the Plutus Tx code, let's briefly go over some basic concepts, including UTXO, datum, redeemer and script context.
+
+The EUTXO Model, Datum, Redeemer and Script Context
+--------------------------------
+
+UTXO (unspent transaction output) is the ledger model used by some blockchains, including bitcoin.
+A UTXO is produced by a transaction, is immutable, and can only be spent once by another transaction.
+In the classic UTXO model, a UTXO contains a wallet address and a value (e.g., some amount of Ada and/or other tokens), and is identified by the wallet address.
```
bezirg, line 28: I know that we use a lot the term "classic UTXO model", but I wonder if the term "original utxo model" is perhaps a better ift. For example we could say before in line 26:

UTXO (unspent transaction output) is the ledger model used by some blockchains, and was originally introduced in bitcoin.

Then we can refer it here as "the original utxo model". Otherwise the reader does not know what the "classic utxo model" refers to

bezirg: `and is identified by the wallet address.` is ambiguous. I don't know how much level of information you want to give to the user but you could say

```
In the original UTXO model, a UTXO contains a wallet address and a value (e.g., some amount of Ada and/or other tokens). Inside a transaction, a UTXO is uniquely identified by the wallet address. At the global blockchain level (a chain of blocks of transactions), the UTXO can be uniquely identified by the transaction id and wallet address.
```
You may leave out the "global blockchain level" sentence because then you have to explain what a transaction id is!

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,332 @@
...
+It can be spent by a transaction, if the transaction is signed by the private key of the wallet address.
+
+The EUTXO (Extended UTXO) model introduces a new kind of UTXO: script UTXO.
```
bezirg, line 31:
```suggestion
The EUTXO (Extended UTXO) model extends the original model with a new kind of UTXO: a script UTXO.
```

doc/read-the-docs-site/tutorials/english-auction-example.rst
bezirg: Actually what I am saying above is not fully correct, I think you need an extra transaction index, both at the local level and the global level, right?

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,332 @@
...
+A script UTXO contains a value, a script (usually a Plutus script) and a piece of data called *datum*, and is identified by the hash of the script.
+For a transaction to spend it, the transaction must provide a piece of input data to the script, referred to as the *redeemer*.
+The script is then run, and it must succeed in order for the transaction to be allowed to spend the UTXO.
+Besides the redeemer, the script also has access to the datum contained in the UTXO, as well as the details of the transaction trying to spend it (referred to as *script context*).
+
+Note that the only thing a Plutus script does is to determine whether a transaction can spend the script UTXO that contains the script. It is *not* responsible for such things as deciding whether it can spend a different UTXO, checking that the input value in a transaction equals the output value, or updating the state of the smart contract.
+Consider it a pure function that returns ``Bool``.
```
bezirg, line 38: what does `checking that the input value in a transaction equals the output value` exactly mean? 

I thought it was possible that a script can do that for the current transaction, but *cannot* do that for any other transaction?

doc/read-the-docs-site/tutorials/english-auction-example.rst
```diff
@@ -0,0 +1,332 @@
...
+Checking transaction validity is done by the ledger rules, and updating the state of a smart contract is done by constructing the transaction to produce a new script UTXO with an updated datum.
+
+The immutability of UTXOs leads to the extremely useful property of completely predictable transaction fees.
+The Plutus script in a transaction can be run off-chain to determine the fee before submitting the transaction onto the blockchain.
+When the transaction is submitted, if some UTXOs it tries to spend has already been spent, the transaction is immediately rejected without penalty.
+If all input UTXOs still exist, and the Plutus script is invoked, the on-chain behavior would be exactly identical to the off-chain behavior.
+This cannot be achieved if transaction inputs are mutable (e.g., in Ethereum's account-based model).
```
bezirg, line 45: Perhaps using past tense is better here?

```This could not be achieved if transaction inputs were mutable (e.g., using Ethereum's account-based model).```

File: doc/read-the-docs-site/tutorials/english-auction-example.rst [hunk 29]
```diff
@@ -0,0 +1,332 @@
...
+
+The ``refundsPreviousHighestBid`` condition checks that the transaction pays the previous highest bid to the previous bidder:
+
+.. literalinclude:: AuctionValidator.hs
+   :start-after: BLOCK5
+   :end-before: BLOCK6
+
+It uses ``PlutusTx.find`` to find the transaction output (a UTXO) that pays to the previous bidder the amount equivalent to the previous highest bid, and verifies that there is at least one such output.
+
+``singleton adaSymbol adaToken amt`` constructs a ``Value`` with ``amt`` Lovelaces.
```
bezirg, line 150: suggestion: `Lovelaces ( the subunit of Ada currency)`

bezirg:
Very well written and explained; the code is also easily readable!

doc/read-the-docs-site/tutorials/AuctionValidator.hs
michaelpj: I guess it depends a bit on the pedagogical aims here? This tutorial seems quite light on the details of how exactly the Haskell is turning into Plutus Core, what's going on with the plugin, etc. , instead focussing on the _logic_ of the application. And I think that logic is clearer without all the qualification, we can make that point more clearly in a tutorial focussed on that, perhaps?

doc/read-the-docs-site/tutorials/english-auction-example.rst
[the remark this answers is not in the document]
michaelpj: I think I just disagree. We're giving people a lot of new concepts here, we want them to be _really clear_ on what it is we're implementing. If they're at all uncertain about how the auction is meant to work then they're going to have a lot more trouble interpreting the following sections.

doc/read-the-docs-site/tutorials/english-auction-example.rst
joseph-fajen: I incorporated your comment but didn't include the sentence about "global blockchain level".

doc/read-the-docs-site/tutorials/english-auction-example.rst
joseph-fajen: I'll need to confirm the answer to this question @bezirg . Maybe I'll ask the question on plutus-core-dev channel.

doc/read-the-docs-site/tutorials/english-auction-example.rst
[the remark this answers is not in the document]
joseph-fajen: I'm preferring to create diagrams that use larger text and can include manually created lines and boxes. My goal is for the diagrams to be easier/simpler to follow. Although the first versions of the diagrams I'm including in the next draft may have too many words!

doc/read-the-docs-site/tutorials/english-auction-example.rst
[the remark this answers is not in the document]
joseph-fajen: I've added three manually created diagrams to reflect first bid, second bid and closing transaction. We've located them next to where the corresponding events are being discussed. We'll need another review to assess their effectiveness!

doc/read-the-docs-site/tutorials/english-auction-example.rst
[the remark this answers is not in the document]
joseph-fajen: We're opting to not use these tables and instead I've inserted 3 diagrams.

doc/read-the-docs-site/tutorials/english-auction-example.rst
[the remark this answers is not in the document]
joseph-fajen: Ziyang has added some nice descriptions of the concepts of contract parameters, datum, redeemer, and script context.

doc/read-the-docs-site/tutorials/english-auction-example.rst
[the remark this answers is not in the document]
michaelpj: Fair enough. I like maintainability and not having to check in images, but up to you. Also sequence diagrams are a well-understood kind of diagram.

doc/read-the-docs-site/tutorials/english-auction-example.rst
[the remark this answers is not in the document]
olgahryniuk: Just for consideration, this entry covers the definitions of a datum, a redeemer, and context and also explains what context can include: https://docs.cardano.org/plutus/plutus-scripts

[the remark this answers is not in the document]
olgahryniuk: Joseph, please also consider linking to the EUTXO handbook: https://ucarecdn.com/3da33f2f-73ac-4c9b-844b-f215dcce0628/EUTXOhandbook_for_EC.pdf

joseph-fajen:
@olgahryniuk from our Content Team offered to create a ticket so that Lee Rosney can design a more attractive version of the three diagrams. Once those are available, I'll replace the temporary (but technically accurate) diagrams we have in place right now.

doc/read-the-docs-site/simple example.rst
```diff
@@ -0,0 +1,324 @@
...
+An asset is identified by a (symbol, token) pair, where the symbol represents the policy that controls the minting and burning of tokens, and the token represents a particular kind of token manipulated by the policy.
+``(adaSymbol, adaToken)`` is the special identifier for Ada/Lovelace.
+
+The ``correctNewDatum`` condition verifies that the transaction produces a *continuing output* containing the correct datum (the new highest bid):
+
+.. literalinclude:: AuctionValidator.hs
+   :start-after: BLOCK6
+   :end-before: BLOCK7
+
+A "continuing output" is a transaction output that pays to the same script address from which we are currently spending.
```
koslambrou, line 165: I thought a continuing output is an output with the same address, value, datum and reference script as the original output, but with a different id. Am I wrong?

james-iohk: If all those stay constant why have a transaction :)
I think an output paying to the same script is accurate. Value and datum will often change, and reference script should probably be kept somewhere else for the long term.

michaelpj: I think you meant this to be a response to a different comment.

I take your point, but I think I disagree. I would say that it _is_ a tutorial and so it should live with the tutorials, or else we should throw away that organizational structure. But I don't like being in the situation where we have the organizational structure but we're using it inconsistently, that makes it very hard to know where to put things. _However_ this is your call I think, I'm just registering my thoughts.

Putting it under tutorials doesn't prevent us from linking to it specially, though.

zliu41: @michaelpj This page is more like "getting started", which is usually separate from "tutorial/user guide".

We don't call it "getting started" because there's no easy way for someone to _actually_ get started - we don't have a browser-based playground, or a simple way to set up a local environment, or things like that. But the bottom line is I want this to be the very first thing one reads when coming to the RTD site.

PR #4979, opened 2022-11-24
PLT-379: Add new conformance tests

kwxm:
This adds the Hachi examples from [here](https://github.com/HachiSecurity/plc-llvm/tree/main/compiler/test-data/untyped) and our golden tests from [here](https://github.com/input-output-hk/plutus/blob/master/plutus-core/untyped-plutus-core/test/Evaluation/Golden.hs).

The Hachi examples had quite a bit of overlap with our existing ones, so I tried to eliminate duplicates.   I've put them in the three existing directories with comments at the top of each example saying they came from Hachi. I also added an acknowledgment in README.md; maybe that's enough and we don't need the comments.

I've added the golden tests from `untyped-plutus-core` in `untyped-plutus-core-tests`.  The original tests construct UPLC ASTs programatically, so I had to extract the source using a separate executable (which I haven't committed, but I could in case we want to do this again).  I haven't removed the original tests because they're used for a [number of different things](https://github.com/input-output-hk/plutus/blob/79779bde26de9c45a486591cb49d929e8de8d62a/plutus-core/untyped-plutus-core/test/Evaluation/Golden.hs#L414) and it seemed a bit weird to remove the untyped tests.

I also reorganised the existing tests a little and renamed some of them so that the names were a bit more consistent.  I just did that manually and forgot to tell git, but I think it managed to work out what I'd done.  I don't think I lost anything.  I made the numbers a lot bigger in some of the integer tests to make sure that things that don't fit into 64 bits are handled correctly.

A few things occurred to me while I was doing this. We might want to open issues for some of those.

* The infrastructure allows property tests but I don't think we have any of those yet.  Maybe we should add some.  Things like the integer div/rem/quot/mod operations and the signature verification functions might be sensible candidates here. There are very extensive tests for signature verification [here](https://github.com/input-output-hk/plutus/blob/master/plutus-core/untyped-plutus-core/test/Evaluation/Builtins/SignatureVerification.hs), but those might be really slow if someone was trying to test an evaluator written in some other language.  Maybe we could cut them down a bit. 
* Maybe the tests for the string operations and `encodeUtf8/decodeUtf8` could be extended to deal with more complicated Unicode strings.  A single Unicode character may occupy 1,2,of 4 bytes in the UTF-8 encoding so we could try to make sure that those are encoded and decoded correctly and that eg `indexString` gets you the right character, and not just some random byte out of the encoded string.
* I don't think we have any tests involving literal `Data`, but I'm not sure if our own parser can handle that yet.
* The `builtin` directory contains a subdirectory called `constant` with tests for literal constants.  Maybe we could move that up a level.
* We could use `nofib-exe` to dump some of the nofib programs to give us some bigger examples.  If we're going to do that we should probably choose parameters that give us programs that don't take too long to run (some of them can take minutes or hours if you ask them to solve big problems)
* Should we have V1/V2 tests?  The behaviour of at least one of the builtins changes, .so we might want to check that.  Maybe there's an issue for that already; there are definitely some for making it easier to switch between V1 and V2 for other things (like in `uplc`).


kwxm:
I should add that the Haskell tests all pass and the Agda ones all pass except for a few that are turned off because they don't work yet.

kwxm:
I just noticed that at the bottom of the issue @thealmarty says that RV have some tests too.  I didn't see that before, so I haven't done anything about that.

michaelpj:
> The infrastructure allows property tests but I don't think we have any of those yet. Maybe we should add some.

A cheap one would be to use the "interesting terms" generator and evaluate them. It generates terms _and_ an expected result, so you can check that it evaluates to what you expect.

> We could use nofib-exe to dump some of the nofib programs to give us some bigger examples.

Hmmm, unsure. We could also use them as programmatically generated tests rather than literal tests, alongside any property tests. That would make the conformance tests depend on the plugin, which might be undesirable (we could switch it off with a flag if it was a problem, I guess?).

michaelpj:
> Should we have V1/V2 tests? The behaviour of at least one of the builtins changes, .so we might want to check that. Maybe there's an issue for that already

We definitely should. I can't remember if there's an issue yet. It's going to be a bit annoying to work into the infrastructure, I expect...

> I haven't removed the original tests because they're used for a [number of different things](https://github.com/input-output-hk/plutus/blob/79779bde26de9c45a486591cb49d929e8de8d62a/plutus-core/untyped-plutus-core/test/Evaluation/Golden.hs#L414) and it seemed a bit weird to remove the untyped tests.

That's a bit annoying. Ultimately if we have typechecking and typed evaluation tests then we *will* be covering all of that, so we could remove them then.

That said... if we're putting these tests in to the conformance test suite as literal examples, we probably will need to duplicate the inputs for the typed conformance suite, rather than relying on erasing the typed examples. So maybe that's an argument for just getting rid of the untyped tests now also. Not sure.

plutus-conformance/agda/Spec.hs
```diff
@@ -43,7 +43,11 @@ failingTests = [
     "test-cases/uplc/evaluation/builtin/mkNilPairData"
     , "test-cases/uplc/evaluation/builtin/mkNilData"
     -- this is because agda has the BuiltinVersion=V1; haskell defaults to latest BuilinVersion
-    , "test-cases/uplc/evaluation/builtin/consByteString"
+    , "test-cases/uplc/evaluation/builtin/consByteString/consByteString1"
+    , "test-cases/uplc/evaluation/builtin/consByteString/consByteString2"
+    -- This is because the V2 version of consByteString requires the first
```
michaelpj, line 48: This is a bit confusing, normally we put comments _above_ the thing they're commenting on?

plutus-conformance/test-cases/uplc/evaluation/untyped-plutus-core-tests/README.md
```diff
@@ -0,0 +1,2 @@
+The test cases here are all extracted from the tests in
```
michaelpj, line 1: I'm not sure it's interesting to future people that these tests came from there. I think it would be more helpful to try and categorize them like the other tests.

plutus-conformance/test-cases/uplc/evaluation/untyped-plutus-core-tests/iteAtHigherKindFullyApplied/iteAtHigherKindFullyApplied.uplc
```diff
@@ -0,0 +1,15 @@
+(program
```
michaelpj, line 1: I'm pretty sure some of these are only interesting for type reasons. e.g. this one says something about "higher kinds", which obviously aren't there in the _untyped_ tests. So I think we may want to delete some of these. Testing that if-then-else works at various types is good, though. Perhaps we just need to rename some of them if they're not redundant?

plutus-conformance/test-cases/uplc/evaluation/untyped-plutus-core-tests/iteAtString/iteAtString.uplc
```diff
@@ -0,0 +1 @@
+(program 1.0.0 (force (builtin ifThenElse)))
```
michaelpj, line 1: e.g. this is obviously a stupid test :D

plutus-conformance/test-cases/uplc/evaluation/untyped-plutus-core-tests/polyError/polyError.uplc
```diff
@@ -0,0 +1 @@
+(program 1.0.0 (delay (error )))
```
michaelpj, line 1: Probably also uninteresting. Or I guess it's interesting for a different reason: it shows that delay can block evaluation errors. So maybe just needs a rename.

michaelpj (commented):
I think this needs a review of the generated tests to check that they make sense in the untyped setting. Given that we were generating them from _typed_ examples, I think many of them aren't that interesting in the untyped setting, or are redundant, or confusingly named.

plutus-conformance/agda/Spec.hs
kwxm: Oh yeah, that's messed up.  I'll fix that.

plutus-conformance/test-cases/uplc/evaluation/untyped-plutus-core-tests/iteAtHigherKindFullyApplied/iteAtHigherKindFullyApplied.uplc
kwxm: Maybe I should just call them `test1`, `test2`, ...: that'd solve that problem.  It's getting to the point where it's difficult to work out which tests are redundant/duplicated because there are so many now.  I suppose I could defend that on the basis that it's better to have too many tests rather than too few.

plutus-conformance/test-cases/uplc/evaluation/untyped-plutus-core-tests/iteAtString/iteAtString.uplc
kwxm: I'm not so sure.  That would fail if (immediately) it was `(force (builtin modInteger))` or something, so it is testing something.  Maybe we need a test for the second case too!

plutus-conformance/test-cases/uplc/evaluation/untyped-plutus-core-tests/iteAtHigherKindFullyApplied/iteAtHigherKindFullyApplied.uplc
michaelpj: I agree that it's going to be annoying and tedious to sort them out, but I think we should do it anyway, otherwise it will only get worse over time.

plutus-conformance/test-cases/uplc/evaluation/untyped-plutus-core-tests/README.md
thealmarty: I think it's fine here in the README about our tests if we don't plan to move those. But I think the comments about the origins of the Hachi tests *in the tests* are more concerning because 
(1) I didn't track where the RV tests are coming from like that so it's inconsistent.
(2) The tests may be moved so the comments are not necessarily accurate. In fact RV has re-org their tests since I added them.
(3) The tests are kinda generic so it's not very important where they came from.

plutus-conformance/test-cases/uplc/evaluation/term/lam/lam-2/lam-2.uplc
```diff
@@ -0,0 +1,4 @@
+-- From  https://github.com/HachiSecurity/plc-llvm/tree/main/compiler/test-data/untyped
```
thealmarty, line 1: Not sure if we should add these comments in the tests. See my other comment.

kwxm: I'll remove all of those.

thealmarty:
> The Hachi examples had quite a bit of overlap with our existing ones, so I tried to eliminate duplicates. I've put them in the three existing directories with comments at the top of each example saying they came from Hachi. I also added an acknowledgment in README.md; maybe that's enough and we don't need the comments.
>
Thanks well done. And yeah I see you already removed the comments.
 
> I've added the golden tests from `untyped-plutus-core` in `untyped-plutus-core-tests`. The original tests construct UPLC ASTs programatically, so I had to extract the source using a separate executable (which I haven't committed, but I could in case we want to do this again). I haven't removed the original tests because they're used for a [number of different things](https://github.com/input-output-hk/plutus/blob/79779bde26de9c45a486591cb49d929e8de8d62a/plutus-core/untyped-plutus-core/test/Evaluation/Golden.hs#L414) and it seemed a bit weird to remove the untyped tests.
> 

Yeah like Michael said there's arguments for both removing and leaving them here. So whatever you feel more right. Either way, maybe we can leave a comment somewhere for when we add the typed tests? Perhaps in the haskell Spec file?

> A few things occurred to me while I was doing this. We might want to open issues for some of those.

Yeah there aren't many detailed and specific issues/plans on how to expand the conformance test suite further yet, these are good suggestions and I agree we should open issues for them.
> 
>     * The infrastructure allows property tests but I don't think we have any of those yet.  Maybe we should add some.  Things like the integer div/rem/quot/mod operations and the signature verification functions might be sensible candidates here. There are very extensive tests for signature verification [here](https://github.com/input-output-hk/plutus/blob/master/plutus-core/untyped-plutus-core/test/Evaluation/Builtins/SignatureVerification.hs), but those might be really slow if someone was trying to test an evaluator written in some other language.  Maybe we could cut them down a bit.
> 
👍🏾 
>     * Maybe the tests for the string operations and `encodeUtf8/decodeUtf8` could be extended to deal with more complicated Unicode strings.  A single Unicode character may occupy 1,2,of 4 bytes in the UTF-8 encoding so we could try to make sure that those are encoded and decoded correctly and that eg `indexString` gets you the right character, and not just some random byte out of the encoded string.
👍🏾 
> 
>     * I don't think we have any tests involving literal `Data`, but I'm not sure if our own parser can handle that yet.
> 
There an issue for that already and yes it's blocked atm.

>     * The `builtin` directory contains a subdirectory called `constant` with tests for literal constants.  Maybe we could move that up a level.
> 
>     * We could use `nofib-exe` to dump some of the nofib programs to give us some bigger examples.  If we're going to do that we should probably choose parameters that give us programs that don't take too long to run (some of them can take minutes or hours if you ask them to solve big problems)

Yeah I wonder about bigger examples. We have an issue for running some tests not continuously on the CI but every so often, maybe that's a good way to add some there. I'll put a comment there. And yeah adding some big but fast examples here would be good too. Worth a ticket.
 
>     * Should we have V1/V2 tests?  The behaviour of at least one of the builtins changes, .so we might want to check that.  Maybe there's an issue for that already; there are definitely some for making it easier to switch between V1 and V2 for other things (like in `uplc`).

There is an issue AND a spike for this already!

thealmarty (approved):
Awesome! Thanks!

plutus-conformance/test-cases/uplc/evaluation/untyped-plutus-core-tests/iteAtHigherKindFullyApplied/iteAtHigherKindFullyApplied.uplc
thealmarty: Hm yeah I think it's best to clean this out too, I'm not sure how though!

bezirg (approved):

kwxm:
As requested, I've removed the duplicate tests (or most of them).  This took hours!  Git also got very confused, but I think I've manged not to lose anything.  I've rearranged the directory structure a little: most of the builtin tests which existed before this are now in a directory called `builtin/semantics` and the de-duplicated tests from `plutus-core` are mostly in `builtins/interleaving` because they're testing that term arguments and `force`s (coming from type arguments in TPLC) are correctly interleaved.  This is quite a complex aspect of the builtin machinery which could be tricky to implement correctly so it's probably sensible to separate the tests.

kwxm:
I moved the tests for constants to their own directory and added some more.  They're really testing the parser.

kwxm:
I'm confused about the syntax now.  The test for `simplePair` was giving a parse error and I changed it from 

```
(program 0.0.0 (con pair (integer) (bool) (12345, True)))
```
to
```
(program 0.0.0 (con (pair integer bool) (12345, True)))
```
and to my surprise it accepted that.  Did we eventually decide that that was the syntax we were going to go with?  If so we should write it down somewhere.

Also, there are [tests](https://github.com/input-output-hk/plutus/tree/kwxm/conformance/new-tests/plutus-conformance/test-cases/uplc/evaluation/builtin/constant/data) for `data` which all give parse errors.  We seem to have concrete syntax for `data` now, but did we ever implement it in the parser?

kwxm:
Sorry @bezirg , I accidentally asked you to re-review this and I don't seem to be able to undo that: just ignore it!

thealmarty:
> and to my surprise it accepted that. Did we eventually decide that that was the syntax we were going to go with? If so we should write it down somewhere.
>

I think we just have to assume that (unless it's obviously a bug) whatever the parser does is what we decided. I think I did ask about whether we should keep a parser spec but the idea is that we don't want to commit to a spec.
 
> Also, there are [tests](https://github.com/input-output-hk/plutus/tree/kwxm/conformance/new-tests/plutus-conformance/test-cases/uplc/evaluation/builtin/constant/data) for `data` which all give parse errors. We seem to have concrete syntax for `data` now, but did we ever implement it in the parser?

No I don't think so.

plutus-core/plutus-core.cabal
```diff
@@ -532,6 +532,32 @@ test-suite untyped-plutus-core-test
     , tasty-hunit
     , text
 
+executable untyped-plutus-core-test-dump
```
thealmarty, line 535: what's this?

thealmarty:
Did you want to add the tickets as mentioned [here](https://github.com/input-output-hk/plutus/pull/4979#issuecomment-1327698644)? And what do you think about leaving a comment maybe [here](https://github.com/input-output-hk/plutus/blob/8593b510687433a75e44a9c3c9ac285b48bc70e6/plutus-core/untyped-plutus-core/test/Evaluation/Golden.hs#L417) that the tests should be removed once we move more to the conformance test suite?

plutus-core/plutus-core.cabal
kwxm: Oh, thanks for spotting that.  I added an executable to dump UPLC source code from the plutus-core tests (they just construct ASTs directly in Haskell) but accidentally committed the cabal entry for it. I've removed that again now.

kwxm:
> Did you want to add the tickets as mentioned [here](https://github.com/input-output-hk/plutus/pull/4979#issuecomment-1327698644)? And what do you think about leaving a comment maybe [here](https://github.com/input-output-hk/plutus/blob/8593b510687433a75e44a9c3c9ac285b48bc70e6/plutus-core/untyped-plutus-core/test/Evaluation/Golden.hs#L417) that the tests should be removed once we move more to the conformance test suite?

I already added a few unit tests involving large integers to the existing tests (or updated them to use nig numbers instead of small ones). I've also added an ticket (PLT-1303) to add property tests to the conformance suite, including more tests for integer operations.  I've added a separate ticket (PLT-1304) for `encodeUtf8` and `decodeUtf8` since what they do is very complicated and we currently only have very simple tests for them.  

As you suggested, I've also added a [comment](https://github.com/input-output-hk/plutus/blob/14252d922e8ae37e3d1269768a8beb8b4e85bdfd/plutus-core/untyped-plutus-core/test/Evaluation/Golden.hs#L417) to the plutus-core tests to suggest removing them once we've ported more of them to the conformance suite.

kwxm:
I also added a ticket at PLT-1305 to think about adding some larger examples, maybe from `nofib`.

michaelpj (approved):
Looks great! The only thing that occurs to me now is that it might be nice to have a systematic approach to the tests that give errors for the main term operations. I guess either:
1. Each of the directories for the term type should have the error cases as tests
2. We should have a dedicated directory for error case

1 is nice for checking that each term type has its error cases included, 2 is nice for checking overall that we haven't missed any. But no need to do that in this PR.

thealmarty:
>We should have a dedicated directory for error case

@michaelpj I did have a dedicated directory for error case but I think you thought it wasn't helpful before. Maybe let's stick with 1.

plutus-core/plutus-core.cabal
thealmarty: If we find that helpful maybe we can leave it but we'll need to add README and docs etc.

michaelpj:
> @michaelpj I did have a dedicated directory for error case but I think you thought it wasn't helpful before. Maybe let's stick with 1.

Clearly I am wildly inconsistent :)

PR #4982, opened 2022-11-28
Add Haddock to inliner

thealmarty:
The inliner was written before the style guide existed. I've fixed the line lengths and added more Haddock as per the style guide.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
@@ -1,4 +1,3 @@
--- editorconfig-checker-disable-file
 {-# LANGUAGE ConstraintKinds  #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs            #-}
@@ -121,10 +120,12 @@ inline. This is essentially the reason for the existence of the UPLC inlining pa
 -- See Note [Inlining approach and 'Secrets of the GHC Inliner']
 newtype InlineTerm tyname name uni fun a = Done (Dupable (Term tyname name uni fun a))
 
-newtype TermEnv tyname name uni fun a = TermEnv { _unTermEnv :: UniqueMap TermUnique (InlineTerm tyname name uni fun a) }
+newtype TermEnv tyname name uni fun a =
+    TermEnv { _unTermEnv :: UniqueMap TermUnique (InlineTerm tyname name uni fun a) }
     deriving newtype (Semigroup, Monoid)
 
-newtype TypeEnv tyname uni a = TypeEnv { _unTypeEnv :: UniqueMap TypeUnique (Dupable (Type tyname uni a)) }
+newtype TypeEnv tyname uni a =
+    TypeEnv { _unTypeEnv :: UniqueMap TypeUnique (Dupable (Type tyname uni a)) }
     deriving newtype (Semigroup, Monoid)
 
 data Subst tyname name uni fun a = Subst { _termEnv :: TermEnv tyname name uni fun a
@@ -161,8 +162,10 @@ data InlineInfo name fun a = InlineInfo
     }
 makeLenses ''InlineInfo
 
--- Using a concrete monad makes a very large difference to the performance of this module (determined from profiling)
-type InlineM tyname name uni fun a = ReaderT (InlineInfo name fun a) (StateT (Subst tyname name uni fun a) Quote)
+-- Using a concrete monad makes a very large difference to the performance of this module
+-- (determined from profiling)
+type InlineM tyname name uni fun a =
+    ReaderT (InlineInfo name fun a) (StateT (Subst tyname name uni fun a) Quote)
 
 lookupTerm
     :: (HasUnique name TermUnique)
@@ -228,7 +231,8 @@ inline hints ver t = let
 {- Note [Removing inlined bindings]
 We *do* remove bindings that we inline (since we only do unconditional inlining). We *could*
 leave this to the dead code pass, but it's helpful to do it here.
-Crucially, we have to do the same reasoning wrt strict bindings and purity (see Note [Inlining and purity]):
+Crucially, we have to do the same reasoning wrt strict bindings and purity
+(see Note [Inlining and purity]):
 we can only inline *pure* strict bindings, which is effectively the same as what we do in the dead
 code pass.
 
@@ -242,20 +246,22 @@ processTerm
     => Term tyname name uni fun a
     -> InlineM tyname name uni fun a (Term tyname name uni fun a)
 processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
-    handleTerm :: Term tyname name uni fun a -> InlineM tyname name uni fun a (Term tyname name uni fun a)
+    handleTerm ::
+        Term tyname name uni fun a
+        -> InlineM tyname name uni fun a (Term tyname name uni fun a)
     handleTerm = \case
         v@(Var _ n) -> fromMaybe v <$> substName n
         Let a NonRec bs t -> do
             -- Process bindings, eliminating those which will be inlined unconditionally,
             -- and accumulating the new substitutions
             -- See Note [Removing inlined bindings]
-            -- Note that we don't *remove* the bindings or scope the state, so the state will carry over
-            -- into "sibling" terms. This is fine because we have global uniqueness
+            -- Note that we don't *remove* the bindings or scope the state, so the state will carry
+            -- over into "sibling" terms. This is fine because we have global uniqueness
             -- (see Note [Inlining and global uniqueness]), if somewhat wasteful.
             bs' <- wither (processSingleBinding t) (toList bs)
             t' <- processTerm t
-            -- Use 'mkLet': we're using lists of bindings rather than NonEmpty since we might actually
-            -- have got rid of all of them!
+            -- Use 'mkLet': we're using lists of bindings rather than NonEmpty since we might
+            -- actually have got rid of all of them!
             pure $ mkLet a NonRec bs' t'
         -- We cannot currently soundly do beta for types (see SCP-2570), so we just recognize
         -- immediately instantiated type abstractions here directly.
@@ -269,7 +275,8 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
         t -> forMOf termSubterms t processTerm
     applyTypeSubstitution :: Type tyname uni a -> InlineM tyname name uni fun a (Type tyname uni a)
     applyTypeSubstitution t = gets isTypeSubstEmpty >>= \case
-        -- The type substitution is very often empty, and there are lots of types in the program, so this saves a lot of work (determined from profiling)
+        -- The type substitution is very often empty, and there are lots of types in the program,
+        -- so this saves a lot of work (determined from profiling)
         True -> pure t
         _    -> typeSubstTyNamesM substTyName t
     -- See Note [Renaming strategy]
@@ -279,7 +286,9 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
     substName :: name -> InlineM tyname name uni fun a (Maybe (Term tyname name uni fun a))
     substName name = gets (lookupTerm name) >>= traverse renameTerm
     -- See Note [Inlining approach and 'Secrets of the GHC Inliner']
-    renameTerm :: InlineTerm tyname name uni fun a -> InlineM tyname name uni fun a (Term tyname name uni fun a)
+    renameTerm ::
+        InlineTerm tyname name uni fun a
+        -> InlineM tyname name uni fun a (Term tyname name uni fun a)
     renameTerm = \case
         -- Already processed term, just rename and put it in, don't do any
         -- further optimization here.
@@ -337,7 +346,9 @@ maybeAddSubst body a s n rhs = do
         then extendAndDrop (Done $ dupable rhs')
         else pure $ Just rhs'
     where
-        extendAndDrop :: forall b . InlineTerm tyname name uni fun a -> InlineM tyname name uni fun a (Maybe b)
+        extendAndDrop ::
+            forall b . InlineTerm tyname name uni fun a
+            -> InlineM tyname name uni fun a (Maybe b)
         extendAndDrop t = modify' (extendTerm n t) >> pure Nothing
 
         checkPurity :: Term tyname name uni fun a -> InlineM tyname name uni fun a Bool
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
@@ -338,16 +338,17 @@ processSingleBinding body = \case
     b -> Just <$> forMOf bindingSubterms b processTerm
 
 -- | Check against the heuristics we have for inlining and either inline the term binding or not.
+-- The arguments to this function are the fields of the `TermBinding` being processed.
 -- Nothing means that we are inlining the term:
 --   * we have extended the substitution, and
 --   * we are removing the binding (hence we return Nothing).
 maybeAddSubst
     :: forall tyname name uni fun a. InliningConstraints tyname name uni fun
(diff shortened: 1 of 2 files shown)


plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -53,18 +52,15 @@ We differ from the paper a few ways, mostly leaving things out:
...
 inlining nested let-bindings, since we don't inline any.
 
-CallSiteInline: not worth it.
+CallSiteInline: we do it based on ambient arity, not context. (TODO in PLT-1146)
 
 Inlining recursive bindings: not worth it, complicated.
 
-Context-based inlining: we don't do CallSiteInline, so no point.
-
-Beta reduction: not worth it, but would be easy. We could do the inlining of the argument
```
thealmarty, line 66: I think this is outdated? We do it in [Beta.hs](https://github.com/input-output-hk/plutus/blob/5c2ccce330f5d1680956945bccb435ae6e02ece3/plutus-core/plutus-ir/src/PlutusIR/Transform/Beta.hs#L1)?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -53,18 +52,15 @@ We differ from the paper a few ways, mostly leaving things out:
 Functionality
 ------------
 
-PreInlineUncoditional: we don't do it, since we don't bother using usage information.
+PreInlineUnconditional: we don't do it, since we don't bother using usage information.
 We *could* do it, but it doesn't seem worth it. We also don't need to worry about
 inlining nested let-bindings, since we don't inline any.
 
-CallSiteInline: not worth it.
+CallSiteInline: we do it based on ambient arity, not context. (TODO in PLT-1146)
```
michaelpj, line 59: This distinction isn't right. We need both the _arity_ of the function to inline and the _context_ of the call site in order to do call site inlining of saturated functions.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -53,18 +52,15 @@ We differ from the paper a few ways, mostly leaving things out:
...
 
 Inlining recursive bindings: not worth it, complicated.
 
-Context-based inlining: we don't do CallSiteInline, so no point.
-
-Beta reduction: not worth it, but would be easy. We could do the inlining of the argument
-here and also detect dead immediately-applied-lambdas in the dead code pass.
+Context-based inlining: CallSiteInline is done based on ambient arity, so no point.
```
michaelpj, line 63: See previous comment.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
michaelpj: Yes, but say that - the point is that we don't do it here.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -117,16 +113,26 @@ turn into a delay/force pair and get simplified away, and then we have something
 inline. This is essentially the reason for the existence of the UPLC inlining pass.
 -}
 
--- 'SubstRng' in the paper, no 'Susp' case
+-- | Substitution range, 'SubstRng' in the paper.
```
michaelpj, line 116: To be clear, I don't think that "substitution range" is a good name, which is why I didn't use it! It's helpful to identify links to the paper, we _don't have to follow it_, in terminology or otherwise.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -117,16 +113,26 @@ turn into a delay/force pair and get simplified away, and then we have something
...
+-- No 'Susp' case because we don't do PreInlineConditionally.
 -- See Note [Inlining approach and 'Secrets of the GHC Inliner']
-newtype InlineTerm tyname name uni fun a = Done (Dupable (Term tyname name uni fun a))
+newtype InlineTerm tyname name uni fun a =
+    Done (Dupable (Term tyname name uni fun a)) --out expressions
 
-newtype TermEnv tyname name uni fun a = TermEnv { _unTermEnv :: UniqueMap TermUnique (InlineTerm tyname name uni fun a) }
+-- | Substitution, 'Subst' in the paper.
```
michaelpj, line 122:
```suggestion
-- | Term substitution, 'Subst' in the paper.
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -117,16 +113,26 @@ turn into a delay/force pair and get simplified away, and then we have something
...
+-- A map of unprocessed variable and its substitution range.
```
michaelpj, line 123: e.g. I find this use of "substitution range" confusing

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -152,48 +158,60 @@ type InliningConstraints tyname name uni fun =
     , PLC.ToBuiltinMeaning uni fun
     )
 
-
+-- | Useful info for inlining.
```
michaelpj, line 161: I think this is actually so brief it's not worth it. A comment on a thing called `foo` that says "the foo" _actually_ doesn't add anything.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -152,48 +158,60 @@ type InliningConstraints tyname name uni fun =
...
 
--- Using a concrete monad makes a very large difference to the performance of this module (determined from profiling)
-type InlineM tyname name uni fun a = ReaderT (InlineInfo name fun a) (StateT (Subst tyname name uni fun a) Quote)
+-- | The monad the inliner runs in.
+-- Using a concrete monad makes a very large difference to the performance of this module
+-- (determined from profiling)
+type InlineM tyname name uni fun a =
+    ReaderT (InlineInfo name fun a) (StateT (Subst tyname name uni fun a) Quote)
 
+-- | Look up the unprocessed variable from the substitution.
```
michaelpj, line 180:
```suggestion
-- | Look up the unprocessed variable in the substitution.
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -152,48 +158,60 @@ type InliningConstraints tyname name uni fun =
...
 lookupTerm
     :: (HasUnique name TermUnique)
-    => name
-    -> Subst tyname name uni fun a
+    => name -- ^ The name of the variable.
+    -> Subst tyname name uni fun a -- ^ The substitution.
     -> Maybe (InlineTerm tyname name uni fun a)
 lookupTerm n subst = lookupName n $ subst ^. termEnv . unTermEnv
 
+-- | Insert the unprocessed variable to the substitution.
```
michaelpj, line 188:
```suggestion
-- | Insert the unprocessed variable into the substitution.
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -152,48 +158,60 @@ type InliningConstraints tyname name uni fun =
...
-    => name
-    -> InlineTerm tyname name uni fun a
-    -> Subst tyname name uni fun a
+    => name -- ^ The name of the variable.
+    -> InlineTerm tyname name uni fun a -- ^ The substitution range.
+    -> Subst tyname name uni fun a -- ^ The substitution.
     -> Subst tyname name uni fun a
 extendTerm n clos subst = subst & termEnv . unTermEnv %~ insertByName n clos
 
+-- | Look up the unprocessed type variable from the substitution.
```
michaelpj, line 197:
```suggestion
-- | Look up the unprocessed type variable in the substitution.
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -152,48 +158,60 @@ type InliningConstraints tyname name uni fun =
...
     => tyname
     -> Subst tyname name uni fun a
     -> Maybe (Dupable (Type tyname uni a))
 lookupType tn subst = lookupName tn $ subst ^. typeEnv . unTypeEnv
 
+-- | Check if the type substitution is empty.
 isTypeSubstEmpty :: Subst tyname name uni fun a -> Bool
 isTypeSubstEmpty (Subst _ (TypeEnv tyEnv)) = isEmpty tyEnv
 
+-- | Insert the unprocessed type variable to the substitution.
```
michaelpj, line 209:
```suggestion
-- | Insert the unprocessed type variable into the substitution.
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
michaelpj: And TBH I would say "add a mapping for a type variable to the substitution". You can't put a variable _alone_ in there!

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -237,25 +256,28 @@ TODO: merge them or figure out a way to share more work, especially since there'
 This might mean reinventing GHC's OccAnal...
 -}
 
+-- | Run the inliner through a `Core.Type.Term`.
```
michaelpj, line 259:
```suggestion
-- | Run the inliner on a `Core.Type.Term`.
```

michaelpj: I think this comment is also borderline not adding anything.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -294,10 +319,11 @@ avoid any variable capture.
 We rename both terms and types as both may have binders in them.
 -}
 
+-- | Run the inliner through a single non-recursive let binding.
```
michaelpj, line 322:
```suggestion
-- | Run the inliner on a single non-recursive let binding.
```
ditto

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -309,16 +335,17 @@ processSingleBinding body = \case
     -- Just process all the subterms
     b -> Just <$> forMOf bindingSubterms b processTerm
 
--- NOTE:  Nothing means that we are inlining the term:
+-- Check against the heuristics we have for inlining and either inline the term binding or not.
```
michaelpj, line 338:
```suggestion
-- | Check against the heuristics we have for inlining and either inline the term binding or not.
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -309,16 +335,17 @@ processSingleBinding body = \case
...
-    => Term tyname name uni fun a
-    -> a
-    -> Strictness
-    -> name
-    -> Term tyname name uni fun a
+    => Term tyname name uni fun a -- ^ The body of the let binding that is to be inlined.
+    -> a -- ^ Its annotation.
+    -> Strictness -- ^ Its `Strictness`.
+    -> name -- ^ Its name.
+    -> Term tyname name uni fun a -- ^ Its RHS.
```
michaelpj, line 348: Do these argument comments add something?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -309,16 +335,17 @@ processSingleBinding body = \case
...
---   * we are removing the binding (hence we return Nothing)
+--   * we are removing the binding (hence we return Nothing).
 maybeAddSubst
     :: forall tyname name uni fun a. InliningConstraints tyname name uni fun
-    => Term tyname name uni fun a
-    -> a
-    -> Strictness
-    -> name
-    -> Term tyname name uni fun a
+    => Term tyname name uni fun a -- ^ The body of the let binding that is to be inlined.
```
michaelpj, line 344:
```suggestion
    => Term tyname name uni fun a -- ^ The body of the let term that is to be inlined.
```

michaelpj: There's a confusion here: we are processing one _binding_ from a given _let term_. The body is the body of the _let term_, the other bits are the pieces of the _binding_. The comments make it sound like they're all associated with the same structure, which they're not.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -447,10 +476,13 @@ For non-strict bindings, the effects already happened at the use site, so it's f
...
 
+-- | Check against the inlining heuristics for types and either inline it, returning Nothing, or
+-- just return the type without inlining.
+-- We only inline if (1) the type is used at most once OR (2) its a `trivialType`.
 maybeAddTySubst
     :: forall tyname name uni fun a . InliningConstraints tyname name uni fun
-    => tyname
-    -> Type tyname uni a
+    => tyname -- ^ The type name.
+    -> Type tyname uni a -- ^ Its type.
```
michaelpj, line 485: This is confusing, I wouldn't say that the type "belongs" to the type variable. Both the name and the type belong to a _binding_. Or you could say "a type variable and the value of the variable"

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -83,7 +84,8 @@ type InliningConstraints name uni fun =
 -- See Note [Differences from PIR inliner] 2
 data InlineInfo name a = InlineInfo { _usages :: Usages.Usages, _hints :: InlineHints name a }
 
--- Using a concrete monad makes a very large difference to the performance of this module (determined from profiling)
+-- Using a concrete monad makes a very large difference to the performance of this module
```
michaelpj, line 87:
```suggestion
-- | The monad the inliner runs in.
-- Using a concrete monad makes a very large difference to the performance of this module
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -152,48 +158,60 @@ type InliningConstraints tyname name uni fun =
...
+    -- ^ have we explicitly been told to inline.
     , _iiBuiltinVer    :: PLC.BuiltinVersion fun
+    -- ^ the builtin version.
     }
 makeLenses ''InlineInfo
 
--- Using a concrete monad makes a very large difference to the performance of this module (determined from profiling)
-type InlineM tyname name uni fun a = ReaderT (InlineInfo name fun a) (StateT (Subst tyname name uni fun a) Quote)
+-- | The monad the inliner runs in.
+-- Using a concrete monad makes a very large difference to the performance of this module
```
michaelpj, line 175: It's a bit subtle, but what I've tended to do is to leave some comments as non-haddock comments, if they're only interesting for people looking at the _definition_ of the name. i.e. do you want to see this when you hover over the name? Probably not. It's useful if you're looking at the definition, though! So I'd be tempted to do:
```
-- Using a concrete monad ...
-- | The monad the inliner runs in.
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
thealmarty: OK I'll mention we do that in Beta.hs, not here then.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
thealmarty: I was wondering why it wasn't used! I think it's nice to follow it though because we are following the paper for these things.

michaelpj (approved):

PR #4987, opened 2022-12-01
PLT-830: Initial Plutus Core debugger TUI

zliu41:
The TUI has two stages: browsing for files to debug, and debugging.

Usage: `cabal run plutus-core:debugger -- [FilePath]`

If `FilePath` is a directory, it enters the browsing stage. If it is a file, it enters the debugging stage directly. `Filepath` defaults to the current working directory.

In the file browsing stage the user is expected to select a UPLC file, but the parser is not currently integrated, so it accepts any file.

The debugging stage has 4 windows, showing, respectively:

- The UPLC program
  - This is the only one currently working. Press `s` to step through it line by line.
- The source program
- The value being returned in CEK machine's return phase
- The CEK machine state (stack, environment)


plutus-core/executables/debugger/FileBrowser.hs
```diff
@@ -0,0 +1,64 @@
...
+                            & dsAppStage .~ Debugging
+                            & dsUplcEditor .~ BE.editorText EditorUplc Nothing uplcText
+        _ -> pure ()
+    pure ()
+
+fileBrowserKeyBindings :: forall n. B.Widget n
+fileBrowserKeyBindings =
+    B.vBox
+        [ B.withAttr menuAttr . B.vBox $
+            [ B.txt "Move cursor                      (Up/Down)"
```
michaelpj, line 57: Where do these come from? Is this just a listing of the standard key bindings in the standard brick file browser widget?

michaelpj: Also I'm slightly surprised that upstream doesn't provide something like this, maybe you should make an issue?

plutus-core/executables/debugger/Debugger.hs
```diff
@@ -0,0 +1,157 @@
...
+                (BE.renderEditor (B.txt . Text.unlines))
+                (st ^. dsCekStateEditor)
+    ui =
+        B.vBox
+            [ BC.center uplcEditor B.<+> BC.center sourceEditor
+            , B.vLimit (st ^. dsVLimitBottomEditors) $
+                BC.center returnValueEditor B.<+> BC.center cekStateEditor
+            , footer
+            ]
+    drawTexts :: [Text] -> B.Widget ResourceName
```
michaelpj, line 66: A bit vague and in fact it's quite specific: this is drawing a _document_ (i.e. a list of lines) with support for highlighting. Could even be promoted to a top-level function taking the highlighting options I think, and it would be clearer. I could have done with a comment explaining the general approach here (treat the document as a list of widgets, one for each row, so we can individually highlight the selected row).

plutus-core/executables/debugger/Debugger.hs
```diff
@@ -0,0 +1,157 @@
...
+    drawTexts =
+        B.vBox
+            . toWidgets
+            -- This is needed for empty lines to be rendered correctly.
+            -- `Brick.Widgets.Core.txt` does the same via `fixEmpty`.
+            . fmap (\t -> if Text.null t then " " else t)
+
+    toWidgets :: [Text] -> [B.Widget ResourceName]
+    toWidgets rows = case st ^. dsUplcHighlight of
+        Nothing -> B.txt <$> rows
```
michaelpj, line 76: I wonder if this would be simpler if you did it the other way around:
```
case highlight of 
   Nothing -> B.txt (Text.unlines rows)
   Just (Highlight.... ) -> toWidgets $ fmap ....
```
Maybe not, then you're at risk of the row-combination behaviour being different between the two cases :thinking:

plutus-core/executables/debugger/Debugger.hs
```diff
@@ -0,0 +1,157 @@
...
+        Just (HighlightCode (B.Location (sRow, sCol)) eLoc) ->
+            let alg :: (Text, Int) -> B.Widget ResourceName
+                alg (row, rowNo)
+                    | rowNo < sRow || rowNo > eRow = B.txt row
```
michaelpj, line 80: Unsure if it's worth it, but there's a nice https://hackage.haskell.org/package/text-rope package for dealing with line-oriented text data like this, including efficient splitting at positions. We use it in `lsp`, might be handy here, although probably only really shows its necessity for really big texts.

plutus-core/executables/debugger/Debugger.hs
```diff
@@ -0,0 +1,157 @@
...
+                            -- right -> no highlight
+                            ((left, middle), right) = first (Text.splitAt s) (Text.splitAt e row)
+                         in B.hBox $
+                                [B.txt left | not (Text.null left)]
+                                    ++ [ B.withAttr highlightAttr $ B.txt middle
+                                       | not (Text.null middle)
+                                       ]
+                                    ++ [B.txt right | not (Text.null right)]
+                  where
+                    B.Location (eRow, eCol) = fromMaybe (B.Location (sRow, Text.length row)) eLoc
```
michaelpj, line 102: Took me a while to find this... have I mentioned I don't like `where`? :sweat_smile:

plutus-core/executables/debugger/Debugger.hs
```diff
@@ -0,0 +1,157 @@
...
+                                    else sCol - 1
+                            e =
+                                if rowNo < eRow
+                                    then Text.length row
+                                    else eCol
+                            -- left -> no highlight
+                            -- middle -> highlight
+                            -- right -> no highlight
+                            ((left, middle), right) = first (Text.splitAt s) (Text.splitAt e row)
+                         in B.hBox $
```
michaelpj, line 95: maybe worth a utility function of its own?
```
highlightedRow :: Text -> (Int, Int) -> Widget
```

plutus-core/executables/debugger/Debugger.hs
```diff
@@ -0,0 +1,157 @@
...
+                                    ++ [B.txt right | not (Text.null right)]
+                  where
+                    B.Location (eRow, eCol) = fromMaybe (B.Location (sRow, Text.length row)) eLoc
+             in fmap alg (zip rows [1 ..])
+
+-- | Handle events in the debugging stage
+handleDebuggerEvent :: B.BrickEvent ResourceName e -> B.EventM ResourceName DebuggerState ()
+handleDebuggerEvent bev@(B.VtyEvent ev) = do
+    focusRing <- gets (^. dsFocusRing)
+    let handleEditorEvent = case B.focusGetCurrent focusRing of
```
michaelpj, line 109: kind of surprised there isn't an abstraction for a collection of widgets managed by a focus ring such that events automatically get dispatched to the focussed widget, but whatever

plutus-core/executables/debugger/Debugger.hs
```diff
@@ -0,0 +1,157 @@
...
+                B.zoom dsSourceEditor $ BE.handleEditorEvent bev
+            Just EditorReturnValue ->
+                B.zoom dsReturnValueEditor $ BE.handleEditorEvent bev
+            Just EditorCekState ->
+                B.zoom dsCekStateEditor $ BE.handleEditorEvent bev
+            _ -> pure ()
+    case ev of
+        Vty.EvKey (Vty.KChar 's') [] -> modify' $ \st' ->
+            -- Stepping. Currently it highlights one line at a time.
+            let highlightNextLine = \case
```
michaelpj, line 122: belongs next to `HighlightCode`?

michaelpj (commented):
Looks nice! Could do with a bit more Haddock generally, I think?

plutus-core/executables/debugger/FileBrowser.hs
```diff
@@ -0,0 +1,64 @@
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+-- | Logic for the file browsing stage.
+module FileBrowser where
```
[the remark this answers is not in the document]
zliu41: Fair point. The file browser is quite simple since we just need to use `Brick.Widgets.FileBrowser`, but this is indeed unnecessary.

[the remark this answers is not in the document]
zliu41: Removed.

plutus-core/executables/debugger/FileBrowser.hs
zliu41: > I'm slightly surprised that upstream doesn't provide something like this

I guess there are lots of ways to show key bindings, and this is just one of them, so it doesn't make sense for the upstream to provide it.

plutus-core/executables/debugger/Debugger.hs
zliu41: `where` can be convenient when using pattern guards and some bindings are used by multiple conditions.

plutus-core/executables/debugger/Debugger.hs
zliu41: Yeah, I changed to something similar to what you suggested.

plutus-core/executables/debugger/Debugger.hs
zliu41: Probably not worth it, unless we run into performance issues.

plutus-core/executables/debugger/Debugger.hs
zliu41: This is temporary. Once we have the CEK machine integrated, `highlightNextLine` will be replaced by highlighting the next Term being computed/returned, and I think that logic should live in this module.

plutus-core/executables/debugger/Debugger.hs
zliu41: Yeah, it doesn't seem easy to generalize this, given how their code is structured.

plutus-core/executables/debugger/Draw.hs
```diff
@@ -0,0 +1,164 @@
...
+    ui =
+        B.vBox
+            [ BC.center uplcEditor B.<+> BC.center sourceEditor
+            , B.vLimit (st ^. dsVLimitBottomEditors) $
+                BC.center returnValueEditor B.<+> BC.center cekStateEditor
+            , footer
+            ]
+
+-- | Draw a document, a consecutive part of which may be highlighted.
+drawDocumentWithHighlight ::
```
michaelpj, line 69: nice

plutus-core/executables/debugger/Draw.hs
```diff
@@ -0,0 +1,164 @@
...
+                        B.Location (eLine, eCol) =
+                            fromMaybe
+                                (B.Location (sLine, Text.length line))
+                                eLoc
+                 in fmap alg (zip lns [1 ..])
+         in B.vBox
+                . toWidgets
+                -- This is needed for empty lines to be rendered correctly.
+                -- `Brick.Widgets.Core.txt` does the same via `fixEmpty`.
+                . fmap (\t -> if Text.null t then " " else t)
```
michaelpj, line 105: I note that nothing in `toWidgets` touches the `Text`s, only `highlightLine` does. So maybe this workaround could go in `highlightLine`? Indeed, it's not clear to me which bit will go wrong, making the workaround more localised might help with that. Is it just that we need

```
txt' t | T.null t = txt " "
txt' t = txt t
```
?

michaelpj (approved):
Anyone else want to take a look at this?

bezirg:
For a start, it looks good to me. The only thing is that I don't see the point of focusing in sub-windows, but maybe it will come into play later. Also, are we planning to step over (line by line for now) the plutustx/haskell source or the plc source code?

zliu41:
> The only thing is that I don't see the point of focusing in sub-windows, but maybe it will come into play later.

By focusing on a window one can perform operations like select, copy, scroll etc.

plutus-core/executables/debugger/Draw.hs
```diff
@@ -0,0 +1,164 @@
...
+            , B.vLimit (st ^. dsVLimitBottomEditors) $
+                BC.center returnValueEditor B.<+> BC.center cekStateEditor
+            , footer
+            ]
+
+-- | Draw a document, a consecutive part of which may be highlighted.
+drawDocumentWithHighlight ::
+    forall n.
+    -- | The part of the document to be highlighted.
+    Maybe HighlightSpan ->
```
thealmarty, line 72: Nit: I was told that we like to have `->` at the start of the line.

plutus-core/executables/debugger/Draw.hs
```diff
@@ -0,0 +1,164 @@
...
+    -- | The document to draw, one `Text` per line.
+    [Text] ->
+    B.Widget n
+drawDocumentWithHighlight = \case
```
thealmarty, line 76: Is there a way to make this less nested? This is a bit hard to keep track.

plutus-core/executables/debugger/Main.hs
```diff
@@ -0,0 +1,110 @@
+{-# LANGUAGE ApplicativeDo     #-}
+{-# LANGUAGE LambdaCase        #-}
```
thealmarty, line 2: remove?

plutus-core/executables/debugger/Main.hs
```diff
@@ -0,0 +1,110 @@
...
+import Brick.Main qualified as B
+import Brick.Util qualified as B
+import Brick.Widgets.Edit qualified as BE
+import Control.Monad.Extra
+import Data.Text.IO qualified as Text
+import Graphics.Vty qualified as Vty
+import Options.Applicative qualified as OA
+import System.Directory.Extra
+
+debuggerAttrMap :: DebuggerState -> B.AttrMap
```
thealmarty, line 33: dead argument?

plutus-core/executables/debugger/Main.hs
```diff
@@ -0,0 +1,110 @@
...
+    uplcText <- Text.readFile (optPath opts)
+
+    let app :: B.App DebuggerState e ResourceName
+        app =
+            B.App
+                { B.appDraw = drawDebugger
+                , B.appChooseCursor = B.showFirstCursor
+                , B.appHandleEvent = handleDebuggerEvent
+                , B.appStartEvent = pure ()
+                , B.appAttrMap = debuggerAttrMap
```
thealmarty, line 78: oh I see you need to match the signature here.

plutus-core/executables/debugger/Main.hs
thealmarty: oic.

plutus-core/executables/debugger/Types.hs
```diff
@@ -0,0 +1,42 @@
+{-# LANGUAGE TemplateHaskell #-}
+
```
thealmarty, line 2: Haddock?

plutus-core/executables/debugger/Main.hs
thealmarty: maybe add a comment?

thealmarty (approved):
Cool!

plutus-core/executables/debugger/Draw.hs
zliu41: I'd agree with it only if we have a formatter that does this for me automatically.

plutus-core/executables/debugger/Draw.hs
zliu41: You should read more pir/plc programs :joy:

plutus-core/executables/debugger/Main.hs
zliu41: Oh, the argument is not used, indeed.

plutus-core/executables/debugger/Draw.hs
zliu41: > txt' t | T.null t = txt " "
txt' t = txt t

Exactly. But we also don't want `txt'` everywhere. It is just when the text we are rendering has _multiple_ lines, then the empty lines need to be fixed. And moving the workaround to `highlightLine` will break this:
> | lineNo < sLine || lineNo > eLine = B.txt line

PR #4988, opened 2022-12-02
modernize tullia CI

dermetfan:


nix/cells/automation/pipelines.nix
```diff
@@ -27,13 +25,15 @@ in
...
-    preset.github-ci.sha = config.preset.github-ci.lib.readRevision
+    preset.github.ci.revision = config.preset.github.lib.readRevision
       inputs.cells.cloud.library.actions.ci.input
       null;
 
-    command.text = ''
-      nix build -L .#${lib.escapeShellArg system}.automation.ciJobs.required
-    '';
+    command.text = config.preset.github.status.lib.reportBulk {
+      bulk.text = "nix eval .#__std.init --apply __attrNames --json | nix-systems -i";
```
michaelpj, line 33: What does this do? What is `nix-systems`? Shouldn't we be splicing that in, how do we know it's going to be present on the system? Implicit runtime dependencies make us sad...

nix/cells/automation/pipelines.nix
```diff
@@ -27,13 +25,15 @@ in
...
+      each.text = ''nix build -L .#"$1".automation.ciJobs.required'';
```
michaelpj, line 34: What is `$1`? What does any of this do? Please write some comments!

michaelpj: I did look at the definition of `reportBulk`, but it's undocumented so that didn't help. You could document that instead!

nix/cells/automation/pipelines.nix
dermetfan: [`nix-systems`](https://github.com/input-output-hk/tullia/blob/ead1f515c251f0e060060ef0e2356a51d3dfe4b0/cells/tullia/apps.nix#L229) is added to the runtime dependencies by `preset.nix.enable = true`.

```
❯ nix run github:input-output-hk/tullia#nix-systems -- -h
Without arguments, prints which systems nix can build for on this machine.
With -l, considers only the system of local machine to be supported.
With -r, considers only systems of remote builders to be supported.
With -i, prints whether the systems read from the JSON list on stdin are supported.
With -I, prints whether the systems read from lines on stdin are supported.
With -O, prints supported systems line by line.
With -o, prints unsupported systems line by line.
```

nix/cells/automation/pipelines.nix
dermetfan: Does this help? https://github.com/input-output-hk/tullia/commit/0a98e231cdec070d3771eb27c73e1577a5b5895a

dermetfan:
I pushed a commit that only builds for systems that we have builders for.

But I wonder: How did that work with Hydra? Did it report green if it skipped builds for unsupported systems?

I'm undecided on what the best approach is. If the flake declares `.#hydraJobs.aarch64-linux`, doesn't that mean we _should_ report an `error` status because we can't build that? By silently skipping it we are dismissing the intent expressed in `.#hydraJobs`.

So I tend to think that tullia's behavior is correct here. If the flake authors _do_ want to silently skip systems that they declared jobs for, they should express that intent by filtering them out as I did in the latest commit. Or of course just remove them from `.#hydraJobs` to express that they don't want to build for these systems at all.

michaelpj:
Perhaps the key point is that we should not (or `std` should not?) assume that "set of systems we want to support building on" == "set of systems we want to run CI on".

michaelpj:
Anyway, I'm happy with this for now, if you think of a better fix, let us know!

PR #4990, opened 2022-12-04
Remove references to hydra.iohk.io from CONTRIBUTING.adoc

zeme-iohk:


CONTRIBUTING.adoc
```diff
@@ -65,16 +65,16 @@ caches have not been configured correctly.
...
 extra-experimental-features = nix-command flakes
 ----
 
 On NixOS, set the following NixOS options:
 ----
 nix = {
-  binaryCaches          = [ "https://hydra.iohk.io" "https://cache.iog.io" ];
-  binaryCachePublicKeys = [ "hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=" ];
+  binaryCaches          = [ "https://cache.iog.io" "https://cache.nixos.org"];
+  binaryCachePublicKeys = [ "hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=" "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=" ];
```
LGLO, line 77: I think you should also update trusted key here: remove one of hydra but add one for cache.iog.io.

cache.iog.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= is my best known value of it

michaelpj: I wasn't aware we had a signing key for `cache.iog.io`? We probably want to keep the hydra one indefinitely: that cache probably has a bunch of stuff that was built by hydra, I don't know if it has been re-signed by `cache.iog.io`?

michaelpj (approved):

CONTRIBUTING.adoc
LGLO: I'm sorry. Seems my comment was invalid.

CONTRIBUTING.adoc
```diff
@@ -65,16 +65,16 @@ caches have not been configured correctly.
...
 trusted-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
 extra-experimental-features = nix-command flakes
 ----
 
 On NixOS, set the following NixOS options:
 ----
 nix = {
-  binaryCaches          = [ "https://hydra.iohk.io" "https://cache.iog.io" ];
-  binaryCachePublicKeys = [ "hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=" ];
+  binaryCaches          = [ "https://cache.iog.io" "https://cache.nixos.org"];
```
michaelpj, line 76: I think you don't need to do this? the nixos options here are unioned, so the default setting of `cache.iog.io` will not be overwritten. Or am I forgetting?

CONTRIBUTING.adoc
michaelpj: Does what you wrote work?

CONTRIBUTING.adoc
zeme-iohk: You mean the default settings of `https://cache.nixos.org`? 
In general they are merged yes, so we don't need that. Though it doesn't hurt to keep.
I'll remove it.

michaelpj: As someone who knows how nixos works it looked wrong to me, I think it's better to write the idiomatic thing!

michaelpj (approved):

PR #4991, opened 2022-12-05
Optimize `maybeAddSubst` in the inliner.

thealmarty:
The same changes will happen in the UPLC inliner in another PR once this is approved.

I've made it so that we don't need to check the purity repeatedly, and avoided checking if it's immediately evaluated in more cases, which can be slow.

I understand that we don't care about performance very much here but there is almost no cost here (I think it's not harder to read?) so maybe it's worth doing? 

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
@@ -357,15 +357,48 @@ maybeAddSubst body a s n rhs = do
     hints <- view iiHints
     let hinted = shouldInline hints a n
 
-    preUnconditional <- preInlineUnconditional rhs'
-    if preUnconditional
+    if hinted -- if we've been told specifically, then do it right away
     then extendAndDrop (Done $ dupable rhs')
     else do
-        -- See Note [Inlining approach and 'Secrets of the GHC Inliner']
-        postUnconditional <- postInlineUnconditional rhs'
-        if hinted || postUnconditional
+        -- See Note [Inlining and purity]
+        termIsPure <- checkPurity rhs'
+        let preInlineUnconditional :: InlineM tyname name uni fun a Bool
+            preInlineUnconditional = do
+                usgs <- view iiUsages
+                -- 'inlining' terms used 0 times is a cheap way to remove dead code while we're here
+                let termUsedAtMostOnce = Usages.getUsageCount n usgs <= 1
+                if not termUsedAtMostOnce -- if the term is used more than once, we don't inline
+                then pure False
+                else do -- otherwise, check if it's nonstrict first
+                            -- effectSafe = case s of
+                                -- Strict    -> termIsPure || immediatelyEvaluated
+                                -- NonStrict -> True
+                    if s == NonStrict
+                    then pure True -- if it's nonstrict, we inline
+                    else if termIsPure
+                        then pure True
+                        else do
+                        -- This can in the worst case traverse a lot of the term, which could lead
+                        -- to us doing ~quadratic work as we process the program. However in
+                        -- practice most term types will make it give up, so it's not too bad.
+                        let immediatelyEvaluated = case firstEffectfulTerm body of
+                                Just (Var _ n') -> n == n'
+                                _               -> False
+                        pure immediatelyEvaluated
+
+        preUnconditional <- preInlineUnconditional
+        if preUnconditional
         then extendAndDrop (Done $ dupable rhs')
-        else pure $ Just rhs'
+        -- See Note [Inlining and purity]
+        -- This is the case where we don't know that the number of occurrences is exactly one,
+        -- so there's no point checking if the term is immediately evaluated.
+        else if not termIsPure then pure $ Just rhs'
+        else do
+        -- See Note [Inlining approach and 'Secrets of the GHC Inliner']
+            postUnconditional <- postInlineUnconditional rhs'
+            if postUnconditional
+            then extendAndDrop (Done $ dupable rhs')
+            else pure $ Just rhs'
     where
         extendAndDrop ::
             forall b . InlineTerm tyname name uni fun a
@@ -379,37 +412,13 @@ maybeAddSubst body a s n rhs = do
             let strictnessFun = \n' -> Map.findWithDefault NonStrict (n' ^. theUnique) strctMap
             pure $ isPure builtinVer strictnessFun t
 
-        preInlineUnconditional :: Term tyname name uni fun a -> InlineM tyname name uni fun a Bool
-        preInlineUnconditional t = do
-            usgs <- view iiUsages
-            -- 'inlining' terms used 0 times is a cheap way to remove dead code while we're here
-            let termUsedAtMostOnce = Usages.getUsageCount n usgs <= 1
-            -- See Note [Inlining and purity]
-            termIsPure <- checkPurity t
-            -- This can in the worst case traverse a lot of the term, which could lead to us
-            -- doing ~quadratic work as we process the program. However in practice most term
-            -- types will make it give up, so it's not too bad.
-            let immediatelyEvaluated = case firstEffectfulTerm body of
-                 Just (Var _ n') -> n == n'
-                 _               -> False
-                effectSafe = case s of
-                    Strict    -> termIsPure || immediatelyEvaluated
-                    NonStrict -> True
-
-            pure $ termUsedAtMostOnce && effectSafe
-
         -- | Should we inline? Should only inline things that won't duplicate work or code.
         -- See Note [Inlining approach and 'Secrets of the GHC Inliner']
         postInlineUnconditional ::  Term tyname name uni fun a -> InlineM tyname name uni fun a Bool
         postInlineUnconditional t = do
             -- See Note [Inlining criteria]
             let acceptable = costIsAcceptable t && sizeIsAcceptable t
-            -- See Note [Inlining and purity]
-            -- This is the case where we don't know that the number of occurences is exactly one,
-            -- so there's no point checking if the term is immediately evaluated.
-            termIsPure <- checkPurity t
-
-            pure $ acceptable && termIsPure
+            pure acceptable
 
 {- |
 Try to identify the first sub term which will be evaluated in the given term and
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
@@ -359,8 +359,10 @@ maybeAddSubst body a s n rhs = do
         termIsPure <- checkPurity rhs'
         preUnconditional <- liftM2 (&&) (termUsedAtMostOnce n) (effectSafe body s n termIsPure)
         -- similar to the paper, preUnconditional inlining checks that the binder is 'OnceSafe'.
-        -- I.e., it's used at most once AND it's `effectSafe` (it isn't doing any substantial work).
-        -- Although we actually also inline 'Dead' binders (i.e., remove dead code) here.
+        -- I.e., it's used at most once AND it neither duplicate code or work.
+        -- While we don't check for lambda etc like in the paper, `effectSafe` ensures that it
+        -- isn't doing any substantial work.
+        -- We actually also inline 'Dead' binders (i.e., remove dead code) here.
         if preUnconditional
         then extendAndDrop (Done $ dupable rhs')
         else do
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
@@ -64,11 +64,13 @@ Implementation
 --------------
 
 In-scope set: we don't bother keeping it, since we only ever substitute in things that
-don't have bound variables. This is largely because we don't do PreInlineUnconditional, which
-would inline big things that were only used once, including lambdas etc.
+don't have bound variables.
 
-Suspended substitutions for values: we don't do it, since you only need it for
-PreInlineUnconditional
+Suspended substitutions ('SuspEx') for values: we don't need it, because we simplify the RHS before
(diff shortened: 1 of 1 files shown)


plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -357,15 +357,48 @@ maybeAddSubst body a s n rhs = do
...
+        termIsPure <- checkPurity rhs'
+        let preInlineUnconditional :: InlineM tyname name uni fun a Bool
+            preInlineUnconditional = do
+                usgs <- view iiUsages
+                -- 'inlining' terms used 0 times is a cheap way to remove dead code while we're here
+                let termUsedAtMostOnce = Usages.getUsageCount n usgs <= 1
+                if not termUsedAtMostOnce -- if the term is used more than once, we don't inline
+                then pure False
+                else do -- otherwise, check if it's nonstrict first
+                            -- effectSafe = case s of
```
zliu41, line 373: remove

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -357,15 +357,48 @@ maybeAddSubst body a s n rhs = do
...
+                                -- Strict    -> termIsPure || immediatelyEvaluated
+                                -- NonStrict -> True
+                    if s == NonStrict
+                    then pure True -- if it's nonstrict, we inline
```
zliu41, line 377: I don't believe this is necessary. There is no `IO` in `InlineM`, and using `||` and `&&` should be able to short circuit and avoid the expensive computation. The current implementation does call `checkPurity` twice, though.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -357,15 +357,48 @@ maybeAddSubst body a s n rhs = do
...
-        postUnconditional <- postInlineUnconditional rhs'
-        if hinted || postUnconditional
+        -- See Note [Inlining and purity]
+        termIsPure <- checkPurity rhs'
+        let preInlineUnconditional :: InlineM tyname name uni fun a Bool
+            preInlineUnconditional = do
+                usgs <- view iiUsages
+                -- 'inlining' terms used 0 times is a cheap way to remove dead code while we're here
+                let termUsedAtMostOnce = Usages.getUsageCount n usgs <= 1
+                if not termUsedAtMostOnce -- if the term is used more than once, we don't inline
```
zliu41, line 370: Unnecessary comment - `if not termUsedAtMostOnce` is perfectly clear and the comment adds nothing.

zliu41 (commented):
Hmm, I didn't realize that we apply `preInlineUnconditional` and `postInlineUnconditional` to the same `rhs'` (it is not the case with GHC). Then it would be strange to call them "pre" and "post" - why not combine them into a single `inlineUnconditional`?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -357,15 +357,48 @@ maybeAddSubst body a s n rhs = do
...
-    if preUnconditional
+    if hinted -- if we've been told specifically, then do it right away
     then extendAndDrop (Done $ dupable rhs')
     else do
-        -- See Note [Inlining approach and 'Secrets of the GHC Inliner']
-        postUnconditional <- postInlineUnconditional rhs'
-        if hinted || postUnconditional
+        -- See Note [Inlining and purity]
+        termIsPure <- checkPurity rhs'
+        let preInlineUnconditional :: InlineM tyname name uni fun a Bool
```
michaelpj, line 365: I think it would be better to instead have 
```
preInlineUnconditional :: Term -> Bool -> InlineM ...
preInlineUnconditional t isPure = ...
```
Reasons:
1. It keeps the symmetry between the multiple inlining functions: you can just pass them all the purity information
2. This version puts _too much stuff_ in scope in `preInlineUnconditional`: it's now much less clear what stuff is relevant for it. Even the main argument now just comes in from an outer scope!
3. It just keeps the indentation down, which is nice.

If you do that I think you can keep almost exactly the code that was there before, just call `termIsPure` once and pass in the result to the functions.

thealmarty: Sure. While I'm here I'll just take them out of `maybeAddSubst`? I'll also rename them. (See my reply to Ziyang's comment)

zliu41:
> Um. I think this may mean I did something wrong when I implemented this?

Hmm, this could indeed be wrong. Section 6.1 of the GHC Inliner paper has an example that demonstrates it is wrong to run the `postInlineUnconditionally` check _before_ simplifying the RHS:

```haskell
-- from the paper
let
  a = ...big...
  b = a
in
...b...b...b...
```

What we do is the opposite: we run the `preInlineUnconditionally` check [_after_ simplifying the RHS](https://github.com/input-output-hk/plutus/blob/fb2e67c45d669f9802b1830c2725f681f8ef4348/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs#L354). But the counter example is similar:

```haskell
let
  a = ...big...
in let b = a
    in ...b...b...b...
```

After simplifying RHS, it becomes

```haskell
let a = ...big... in ...a...a...a...
```

and unless we update the usage count of `a`, we'd inline `a`.

michaelpj:
> Simplifying RHS before preInlineUnconditionally should be correct, I think. It could potentially cause the RHS to be simplified twice, but looks like we avoid doing so.

Right, I think I did think that through. It wasn't clear to me from the paper why you *wouldn't* simplify the RHS before `preInlineUnconditional`, so I just did it. I think perhaps it makes a difference for GHC: they can do more simplification in context, but they only want to process the binding RHS once, so delaying it until the call site is better if it's a single-occurrence term. But in our case I think we're just happy to leave any improved in-context simplification to later passes, so it's not a big deal. Worth a comment, though.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -116,7 +112,7 @@ inline. This is essentially the reason for the existence of the UPLC inlining pa
 -}
 
 -- | Substitution range, 'SubstRng' in the paper.
--- No 'Susp' case because we don't do PreInlineConditionally.
+-- IS THIS STILL RIGHT? No 'Susp' case because we don't do PreInlineConditionally.
```
michaelpj, line 115: I think it's wrong: we do do `PreInlineUnconditional`, but with the tweak from my comment. So we don't need `Susp`, still.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -357,15 +353,24 @@ maybeAddSubst body a s n rhs = do
...
+        -- See Note [Inlining and purity]
+        termIsPure <- checkPurity rhs'
+        preUnconditional <- preInlineUnconditional body s n termIsPure
+        if preUnconditional
         then extendAndDrop (Done $ dupable rhs')
-        else pure $ Just rhs'
+        -- See Note [Inlining and purity]
+        -- This is the case where we don't know that the number of occurrences is exactly one,
+        -- so there's no point checking if the term is immediately evaluated.
+        else if not termIsPure then pure $ Just rhs'
```
michaelpj, line 367: Why did you move this out of `postInlineUnconditional`? It seems to be part of it, I don't see why we would make the two cases asymmetrical like this?

thealmarty: I'm actually thinking about renaming these and separate them out even more (effectSafe, usedOnce etc). We don't follow the paper so closely - I'll add comments about their correspondence to the paper. What do you think?

michaelpj: Well, if you take that to its logical conclusion you're just talking about inlining `preInlineUnconditional` and `postInlineUnconditional`. All those functions _are_ is a collection of checks to decide what to do. If we start taking some of the checks out from the functions, then their existence doesn't make much sense any more.

michaelpj: I think maybe just try it and then take a look and see if you think it's actually clearer.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -420,6 +392,30 @@ maybeAddSubst body a s n rhs = do
             let acceptable = costIsAcceptable t && sizeIsAcceptable t
             pure acceptable
 
+-- according to the paper we could only do this when it's `OnceSafe` - the single occurrence is not
+-- inside a lambda, nor is a constructor argument?
```
thealmarty, line 396: is there a reason we don't check this? According to the paper we could duplicate work.

thealmarty: But `effectSafe` says to inline a `nonstrict` binding unconditionally?

thealmarty:
Once this is merge I'll open a PR to sync up the UPLC inliner.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
zliu41: I guess in PIR, inlining a lazy binding cannot duplicate work. e.g., suppose `x` is lazy, in `let x = <big> in ...x...x...`, `<big>` is evaluated twice whether `x` is inlined or not (unlike GHC). So the "Safe" part of "OnceSafe" does not apply.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -357,59 +353,71 @@ maybeAddSubst body a s n rhs = do
     hints <- view iiHints
     let hinted = shouldInline hints a n
 
-    preUnconditional <- preInlineUnconditional rhs'
-    if preUnconditional
+    if hinted -- if we've been told specifically, then do it right away
```
zliu41, line 356: Probably fine, but beware that this is not necessarily an improvement. This does mean we check `preInlineUnconditional` less, but it also means we check `hinted` more. And the inline hints can contain arbitrary logic with unknown complexity.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -357,59 +353,71 @@ maybeAddSubst body a s n rhs = do
...
     then extendAndDrop (Done $ dupable rhs')
     else do
-        -- See Note [Inlining approach and 'Secrets of the GHC Inliner']
-        postUnconditional <- postInlineUnconditional rhs'
-        if hinted || postUnconditional
+        termIsPure <- checkPurity rhs'
+        preUnconditional <- liftM2 (&&) (termUsedAtMostOnce n) (effectSafe body s n termIsPure)
+        -- similar to the paper, preUnconditional inlining checks that the binder is 'OnceSafe'.
```
zliu41, line 361: I don't think our "Safe" and their "Safe" mean the same thing.
Their "Safe" means "don't duplicate work".
Our "Safe" means "don't alter the semantics of the program".

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
```diff
@@ -357,59 +353,71 @@ maybeAddSubst body a s n rhs = do
...
+        -- I.e., it's used at most once AND it's `effectSafe` (it isn't doing any substantial work).
+        -- Although we actually also inline 'Dead' binders (i.e., remove dead code) here.
+        if preUnconditional
         then extendAndDrop (Done $ dupable rhs')
-        else pure $ Just rhs'
+        else do
+        -- See Note [Inlining approach and 'Secrets of the GHC Inliner'] and [Inlining and purity].
```
zliu41, line 367: Please align the indentation.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
thealmarty: > the inline hints can contain arbitrary logic with unknown complexity.

this should just be checking a bool? Can you explain a bit more?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
thealmarty: Not altering the semantics of the program is a given. Ideally we also want to not duplicate work.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
zliu41: `shouldInline hints a n` can have arbitrary cost.

zliu41 (approved):

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
michaelpj: I think it's a reasonable constraint on the caller that computing the inlining hints should not be expensive!

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline.hs
michaelpj: Right, but Ziyang is pointing out that we currently use the words differently. And I think the comment here _does_ confuse those senses, because it references the paper's use of "safe" right next to a different use of "safe".

thealmarty: Updated.

thealmarty:
I built and ran plc and pir tests locally and they passed. This may need a force merge.

thealmarty:
I rebased, since PR#5002 is merged - maybe the CI will work better now?

PR #4992, opened 2022-12-05
Remove references to Hydra in nix code

zeme-iohk:


nix/cells/automation/ciJobs.nix
```diff
@@ -1,8 +1,3 @@
-# TODO(std) replace all occurrences of hydra with cicero once hydra is gone.
-# Normally we'd name this file default.nix and put it into a hydraJobs folder.
-# However that causes all hydra job names to contain an extra string (the flake fragment name)
-# in the final attribute path. To make job names shorter we name this file hydraJobs.nix instead.
```
michaelpj, line 4: Well, should we move it to `ciJobs/default.nix`? Or does this still matter, in which case we still need the comment?

michaelpj (approved):

nix/cells/automation/ciJobs.nix
zeme-iohk: Good catch, I will create the default.nix file

zeme-iohk: Actually the comment is no longer relevant. We can keep as is

PR #4993, opened 2022-12-05
Remove pkg-config from the devshell packages

zeme-iohk:
I run `nix flake lock --update-input haskell-nix --update-input hackage-nix`
Recent version of haskell.nix fixes https://github.com/input-output-hk/haskell.nix/issues/1753

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -583,11 +583,11 @@
     "hackage-nix": {
       "flake": false,
       "locked": {
-        "lastModified": 1667178734,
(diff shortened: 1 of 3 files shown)


michaelpj:
Weirdly big diff to the lockfile but okay. I do actually want to see this one build so we'll have to wait until we figure out what's up with CI.

michaelpj (approved):
Merrily building GHC, let's see what happens.

michaelpj:
It got some way but then got killed, so I guess push a no-op commit to get it to try again.

PR #4995, opened 2022-12-06
Revert tullia version

zeme-iohk:
Let's see if this does it

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1124,16 +1124,16 @@
         "std": "std_2"
       },
       "locked": {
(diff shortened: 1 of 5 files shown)


michaelpj:
I'm pretty sure the benchmarking stuff will rely on the branch that Parthiv set it to, but I guess we can fix that later.

zeme-iohk:
Yeah the gh-comment branch... still not working tho. Maybe I needed to recreate the actions? I just did that, let's see

PR #5001, opened 2022-12-07
Benchmarking experiments

kwxm:
Nothing important has changed here.  I'm just experimenting with the benchmarks.


kwxm:
Where's my rocket?

Pacman99:
Benchmark invocation has been having issues, fix in #5002.
Also I think the rocket reaction was from the old workflow which Michael removed. It would have to be added at the start of the benchmark script to bring it back.

PR #5002, opened 2022-12-09
fix(ci/benchmark): export script config

Pacman99:
Some invocations have been failing because of the benchmark name having newlines ex: https://cicero.ci.iog.io/invocation/c8c2aaba-83f6-4e7b-ac0b-215c08bef970

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/nix/cells/automation/library/plutus-benchmark-runner.nix
+++ b/nix/cells/automation/library/plutus-benchmark-runner.nix
@@ -18,9 +18,9 @@ pkgs.writeShellApplication {
     pkgs.coreutils
   ];
   text = ''
-    PR_NUMBER=${PR_NUMBER} \
-    BENCHMARK_NAME=${BENCHMARK_NAME} \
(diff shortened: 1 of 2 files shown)


michaelpj:
I think the old workflow just took the content of the line so we didn't have this problem :thinking: Could we not just change the comment match regex?

Pacman99:
The commentMatch regex in the action should stay the same, we would still want benchmarks to be triggered even if there are new lines in the comment.

I just updated the PR to use `builtins.match` to find the benchmark name, so we can use regex for this too.

michaelpj (approved):

michaelpj:
Okay, for some reason this broke the CI action now it's on master. HOW??? Ah, because it fails hard if there's no match on the comment, and that gets run unconditionally on all comments. WTF.

PR #5005, opened 2022-12-09
Add Github Actions with Netlify to deploy haddock docs

sfoo-iohk:
Demo for deploying website on Netlify

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,33 @@
+name: Haddock
+on:
+  pull_request:
+  push:
+jobs:
+  build-haddock-site:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,33 @@
...
+        with: 
+          nix_conf: |
+            experimental-features = nix-command flakes
+            accept-flake-config = true 
+      - name: Build haddock site
+        run: |
+          nix build .#combined-plutus-haddock
+          mkdir dist
+          cp -RL ./result/share/doc/* ./dist/
+      - name: Deploy to Netlify
```
michaelpj, line 20: I think we can just deploy it to github pages more simply?

PR #5006, opened 2022-12-12
PLT-1276 Update plc-agda command line

mjaskelioff:
The purpose of this PR is to align more the interface of `plc-agda` to `plc`/`uplc`.
In particular:

* Removed unused modes of execution
* Change input option from `-f` /`--file` to `-i`/--`input` so that it agrees with `uplc`/`plc`.
* Add support for input files in flat format.
* A bit of clean up of `Main.lagda`.
* Prepare code for reusing `Common.hs` and `Parsers.hs`  from the `uplc` and `plc` executables. So far the files are copied. When [PLT-1312](https://input-output.atlassian.net/browse/PLT-1312) is done, These files can be erased and the executable library be imported instead.


plutus-metatheory/src/Main.lagda
```diff
@@ -63,17 +64,12 @@ instance
   IOMonad : Monad IO
   IOMonad = record { return = IO.return; _>>=_ = IO._>>=_ }
 
--- Bytestring stuff
+-- Parsing stuff
```
thealmarty, line 67:
```suggestion
-- For parsing
```

thealmarty: Just a nit, "stuff" is a bit too informal for my taste. Feel free to ignore!

plutus-metatheory/src/Main.lagda
```diff
@@ -379,54 +382,43 @@ alphaU plc1 plc2 | _ | _ = Bool.false
...
 
 data EvalOptions : Set where
-  EvalOpts : Input → EvalMode → EvalOptions
+  EvalOpts : Input → Format → EvalMode → EvalOptions
 
 {-# COMPILE GHC EvalOptions = data EvalOptions (EvalOpts) #-}
 
-data TCOptions : Set where
-  TCOpts : Input → TCOptions
+data TypecheckOptions : Set where
```
thealmarty, line 392: 👍🏾 on the renaming.

plutus-metatheory/src/Opts.hs
```diff
@@ -6,27 +8,13 @@ import Data.Text qualified as T
...
-  <> help "Read from file" )
+import Common
+import Parsers
 
-stdInput :: Parser Input
-stdInput = flag' StdInput
-  (  long "stdin"
-  <> help "Read from stdin" )
+-- The different evaluation modes of plc-agda
+data EvalMode = U | TL | TCK | TCEK deriving stock (Show, Read)
```
thealmarty, line 15: `L` and `CK` modes are removed?

mjaskelioff: Those modes are longer supported (there's no evaluator code for them).

thealmarty (approved):

effectfully (commented):
I've requested Kenneth's review, since I'm not too familiar with the executables.

kwxm:
What's happened here?  There seem to be lots of irrelevant changes from recent commits by other people.  Did you rebase or something?

plutus-metatheory/src/Parsers.hs
```diff
@@ -0,0 +1,154 @@
+{-# LANGUAGE LambdaCase #-}
```
kwxm, line 1: I think this file is identical to the one in `plutus-core/executables`, is that right?  That must mean that it's got lots of stuff that the metatheory doesn't need, but there's probably not much point in removing it since that'll make it difficult to maintain.  Can we re-use the `plutus-core` version directly, or would that require `plutus-core` to export something that it doesn't want to?

kwxm: ... and the same comment for `Common.hs`.

kwxm (approved):
Looks fine to me. It might be worth trying to remove the irrelevant commits, since they might mess up the git history. I'm not sure though: maybe if you merge it everything will be OK anyway.

plutus-metatheory/src/Parsers.hs
mjaskelioff: Yes, the files are copied from `plutus-core/executables`. In [PLT-1312](https://input-output.atlassian.net/browse/PLT-1312) I'll make the original files available as a library and remove the copied files.

PR #5007, opened 2022-12-13
PLT-456: Combine `verbose` and `debug` compiler options into `verbosity`

zliu41:
Once we have the debugger, the `debug` option will be confusing - it could be thought of as the option for producing Plutus Core with SrcSpan annotations.


michaelpj:
```
Warning, treated as error:
reference/writing-scripts/compiler-options-table.rst:6:Error parsing content block for the "list-table" directive: exactly one bullet list expected.
```
Looks like what you have there just doesn't parse at all? not sure why

zliu41:
Heh, indentation was messed up, not sure why. Fixed.

PR #5019, opened 2022-12-15
deleted cross-reference to technical report that doesn't yet contain …

joseph-fajen:
…cost model information

I've verified the accuracy of this page with Plutus Core team members. The only change is to remove a cross-reference from the top part of the page that points to the technical report that does not yet contain any information about the cost model.

### DIFF
--- a/doc/read-the-docs-site/reference/cardano/cost-model-parameters.rst
+++ b/doc/read-the-docs-site/reference/cardano/cost-model-parameters.rst
@@ -7,8 +7,6 @@ The cost model for Plutus Core scripts has a number of parameters.
 These are listed and briefly described below.
 All of these parameters are listed in the Cardano protocol parameters and can be individually adjusted.
 
-For more details on the meaning of the parameters, consult :cite:t:`plutus-report`.
-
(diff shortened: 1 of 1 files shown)


zliu41 (commented):
I just don't see the reason for this page to exist. And it seems that its creation involves manual steps, which means even if it isn't out-of-date today (I don't even know how to check that without a lot of manual work), it will be at some point. My recommendation is to remove it, and if someone complains and provides a legitimate reason, we can then implement and put up an auto-generated version.

michaelpj:
> My recommendation is to remove it, and if someone complains and provides a legitimate reason, we can then implement and put up an auto-generated version.

We had this conversation a few times, and Joseph says that Kevin wants it. I also think we should delete it.

michaelpj:
But removing the cross-reference is fine.

zliu41:
> We had this conversation a few times, and Joseph says that Kevin wants it.

I was suggesting that we remove it in spite of that. And bring and automated version if someone _really_ complains with a good use case.

If someone actually needs it, having a version up that was last updated in June and that we don't know whether it's accurate or not can be more problematic.

PR #5024, opened 2022-12-17
Update `getSourceSpan`

zliu41:
It should look at `ModBreaks` in the `Breakpoint` case, not the `HpcTick` case.


effectfully (commented):
No idea what's going on here.

bezirg:
I am not exactly sure what the problem is that this patch solves.

Is it that this function `getSourceSpan` used an hpc index to lookup the ghci-breakpoint-array? It looks indeed to be a problem; @zliu41 did you stumble on this problem using plutus's coverage tool or were you planning to reuse `getSourceSpan` for the debugger?

Also it is good to check that this patch does not break the coverage tool (i don't think so because coverage tool relies on dwarf and not ghci for its sourcespans)

zliu41:
@bezirg `getSourceSpan` will be used by the debugger. This bug makes no difference though, since the debugger uses neither HPC nor breakpoints, but I just noticed the bug.

How should I check whether or not it breaks the coverage tool? It shouldn't (unless the coverage tool is already broken), since using `tickId` on breakpoints makes zero sense.

bezirg (approved):

PR #5027, opened 2022-12-21
PLT-1334 ADR for inlining of saturated functions.

thealmarty:
This is part of PLT-1146, which has the objective of adding inlining of saturated function to the PIR and UPLC inliner. This ADR contains details of the implementation of determining whether a function is fully applied. It also lists the weaknesses of this approach and plans to mitigate them. 

Thank you @michaelpj and @zliu41 for your valuable inputs.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

[PLT-1146]: https://input-output.atlassian.net/browse/PLT-1146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
+++ b/doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
@@ -12,15 +12,15 @@ Accepted
 
 ## Context
 
-As part of our effort to increase script capacity, 
-we take from [Plutonomy](https://github.com/well-typed/plutonomy), 
-which gives 15-30% speedup on benchmarks. 
-See also [PLT-1044](https://input-output.atlassian.net/browse/PLT-1044).
+As part of our effort to increase script capacity,
+we take from [Plutonomy](https://github.com/well-typed/plutonomy),
+which gives 15-30% speedup on benchmarks.
+See [PLT-1044](https://input-output.atlassian.net/browse/PLT-1044).
 
-The goal is to integrate what we know already works there and 
-also add optimizations we can do before UPLC (Plutonomy only applies optimization to UPLC). 
-We currently have an inliner pass for PIR and UPLC. 
-One of the optimization we can add is inlining of saturated functions in both PIR and UPLC. 
+The goal is to integrate what we know already works there and
+also add optimizations we can do before UPLC (Plutonomy only applies optimization to UPLC).
+We currently have an inliner pass for PIR and UPLC.
+One of the optimization we can add is inlining of saturated functions in both PIR and UPLC.
 From our analysis, this is one of the optimizations that contributes to the speedup in Plutonomy.
 
 For example:
@@ -31,23 +31,24 @@ let f x y = 1
 in f a b 
 ```
 
-we should inline `f` on line 3 (even though it’s used more than once), because it’s saturated, 
-i.e. it has all its arguments, and we can reduce it. 
-To do this we need to do some arity analysis of functions, 
-so we know how many term/type arguments they have.
+we should inline `f` on line 3 (even though it’s used more than once), because it’s saturated,
+i.e. it has all its arguments, and we can reduce it.
+([PLT-1146](https://input-output.atlassian.net/browse/PLT-1146))
 
+To do this we need to do some arity analysis of functions,
+so we know how many term/type arguments they have.
 We also need to decide whether or not to inline based on the call site, which we don’t currently.
 
-This ADR records the details of the design decisions we made for implementing 
-inlining of saturated functions ([PLT-1146](https://input-output.atlassian.net/browse/PLT-1146)).
+This ADR records the details of the design decisions we made for the arity analysis of functions.
+The mechanics of whether or not to inline based on the call site will be covered in a separate ADR.
 
 ## Decision
 
-For a first round of implementation, we are adding a simplified pass. 
-We will refine the implementation over time. 
+For a first round of implementation, we are adding a simplified pass.
+We will refine the implementation over time.
 In each round, we benchmark and test and investigate the results.
 
-### Introduction
+### The implementation: Counting lambdas
 
 **Arity** is relevant when we talk about a fully applied function.
 A fully applied function has an arity of 0.
@@ -71,9 +72,9 @@ If _v_ is fully applied in the body, i.e., if
 1. the number of type lambdas equals the number of type arguments applied, and
 2. the number of term lambdas equals the number of term arguments applied,
 
-we inline _v_, i.e., replace its occurrence with _rhs_ (**if** other call site inlining
-conditions are satisfied. We currently inline too little, this will be improved later, see below).
-For the rest of the discuss we focus on the fully applied condition only.
+we inline _v_, i.e., replace its occurrence with _rhs_ (**if other call site inlining
+conditions are satisfied**. We currently inline too little, this will be improved later, see below).
+For the rest of the discussion here we focus on the fully applied condition only.
 
 E.g., for
 
@@ -84,44 +85,69 @@ let f = \x.\y.x+y in
 
 the _body_ `f 2 4` is a function application of `f`. Since `f` is fully applied, we inline `f`.
 
-### Other examples
+### Cases we miss
 
-The above case is straight forward. But there are other cases that makes it difficult to determine
-whether a function is fully applied. E.g.,
+The above case is straight forward. But there are other cases that
+may not give us the desired result if we just count the lambdas. Below are some examples.
+
+#### Example 1: function in body
 
 ```haskell
-let f = \x . 1
+let f = \x . x
      g = \y . f
-in g a b
+in g a
 ```
+
+`f` and `g` each has arity of 1. However, `g`'s _body_ includes `f` which also has arity of 1.
+Since we only count the number of lambdas, `g` is fully applied, and we inline.
+But `g a` reduces to `f`, which has arity of 1.
+So inlining `g` doesn't give the desired result of a fully reducible term in this case.
+
+#### Example 2: function as an argument
 
 ```haskell
 let id :: forall a . a -> a
-     id x = x
+    id x = x
 in id {int -> int} (\x -> x +1)
 ```
 
-What about function application?
+Here we have a type application for a function that takes one type variable.
+I.e., it's fully applied in terms of type.
+In terms of terms, `id` takes one argument, and is indeed applied to one.
+So `id` is fully applied! And we inline it.
 
-(\x.\y.x+y) 4
+However, reducing the _body_ `id {int -> int} (\x -> x +1)` leaves us `(\x -> x +1)`,
+a function of arity 1, not 0.
 
-With beta-reduction, it becomes
+In this case, inlining and reducing `id` reduces the amount of code,
+so even though it reduces to a not fully applied function, it may be desirable.
+Again, we need more sophisticated heuristics to improve the performance of this.
 
-\y.4+y and it has an arity of 1.
+#### Example 3: function application in _RHS_
 
+```haskell
+let f = (\x.\y.x+y) 4
+in f 5
+```
 
+With beta-reduction, `f` becomes `\y.4+y` and it has an arity of 1.
+The _body_ `f 5` has arity of 0 and thus is a fully applied function!
+However, because the _rhs_ `(\x.\y.x+y) 4` is a function application, not a lambda abstraction,
+we won't inline it with the current implementation.
 
-## Future plan
+## Future plans
 
-After PLT-1146 is done, we will work on improving this pass so that it will catch more cases.
+We have plans to improve this pass so that it will catch more cases that are desirable,
+and not inline undesirably.
 
 ### Add beta-reduction pass
 
-Beta-reduction of rhs is run first. This will catch more functions.
+If beta-reduction of _rhs_ is run before this pass, it can catch cases categorized by example 3.
 
 ### Add inline size heuristic
 
-
+If we do some analysis of the inline size, we can better decide whether it's desirable to inline.
+Plutonomy also [uses a similar heuristic](https://github.com/well-typed/plutonomy/blob/14b9bd46084db1b785b3a99d55f7f10d38165ee8/src/Plutonomy/Hereditary/Transform.hs#L266).
 
 ### Add more refined heuristic for call site inlining
 
@@ -132,7 +158,7 @@ We will optimize this further in PLT-1041.
 
 ### Further understanding of the optimization
 
-It's important to point out that it is not obvious why inlining saturated functions causes speedups.
+It is not obvious why inlining saturated functions causes speedups.
 We are doing this because we _observed_ the speedup from Plutonomy's implementation.
 It would be useful if we could understand what's causing the speedup,
-which will enable us to potentially improve the optimization further, or add other optimization.
+which will enable us to potentially improve the optimization further, or add other optimizations.
--- a/doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
+++ b/doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
@@ -1,0 +1,138 @@
+# ADR 5: Inlining fully saturated functions
+
+Date: 2022-12
+
+## Authors
+
+Marty Stumpf <marty.stumpf@iohk.io>
+
+## Status
+
+Accepted
+
+## Context
+
+As part of our effort to increase script capacity, 
+we take from [Plutonomy](https://github.com/well-typed/plutonomy), 
+which gives 15-30% speedup on benchmarks. 
+See also [PLT-1044](https://input-output.atlassian.net/browse/PLT-1044).
+
+The goal is to integrate what we know already works there and 
+also add optimizations we can do before UPLC (Plutonomy only applies optimization to UPLC). 
+We currently have an inliner pass for PIR and UPLC. 
+One of the optimization we can add is inlining of saturated functions in both PIR and UPLC. 
+From our analysis, this is one of the optimizations that contributes to the speedup in Plutonomy.
+
+For example:
+
+```haskell
+let f x y = 1
+    y = f q
+in f a b 
+```
+
+we should inline `f` on line 3 (even though it’s used more than once), because it’s saturated, 
+i.e. it has all its arguments, and we can reduce it. 
+To do this we need to do some arity analysis of functions, 
+so we know how many term/type arguments they have.
+
+We also need to decide whether or not to inline based on the call site, which we don’t currently.
+
+This ADR records the details of the design decisions we made for implementing 
+inlining of saturated functions ([PLT-1146](https://input-output.atlassian.net/browse/PLT-1146)).
+
+## Decision
+
+For a first round of implementation, we are adding a simplified pass. 
+We will refine the implementation over time. 
+In each round, we benchmark and test and investigate the results.
+
+### Introduction
+
+**Arity** is relevant when we talk about a fully applied function.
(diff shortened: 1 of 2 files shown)


doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,206 @@
...
+#### Example 3: function application in _RHS_
+
+```haskell
+let f = (\x.\y.x+y) 4
+in f 5
+```
+
+With beta-reduction, `f` becomes `\y.4+y` and it has an arity of 1.
+The _body_ `f 5` has an arity of 0 and thus is a fully applied function!
+We can reduce it to 9.
```
thealmarty, line 136: is this true?

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,206 @@
...
+```haskell
+let f = \x . x
+     g = \y . f
+in g a
+```
+
+`f` and `g` each has an arity of 1. However, `g`'s _body_ includes `f` which also has an arity of 1.
+Since we only count the number of lambdas, `g` is fully applied, and we inline.
+But `g a` reduces to `f`, which has an arity of 1.
+So inlining `g` doesn't give the desired result of a fully reducible term in this case.
```
zliu41, line 105: I don't think this example is correct. We _should_ inline `g`, because inlining `g` produces `let f = \x . x in f`, which is better than the original.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,206 @@
...
+See [PLT-1041](https://input-output.atlassian.net/browse/PLT-1041).
+At the moment we inline any lambda with a trivial body.
+This is too aggressive and has been observed to lead to size increases.
+We will optimize this further in PLT-1041.
+
+### Further understanding of the optimization
+
+It is not obvious why inlining saturated functions causes speedups.
+We are doing this because we _observed_ the speedup from Plutonomy's implementation.
+It would be useful if we could understand what's causing the speedup,
```
mjaskelioff, line 205: Without looking at the actual code, a general guess is that inlining a saturated function avoids creating a closure, whereas for an unsaturated function the closure would need to be created anyway.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,206 @@
...
+Here we have a type application for a function that takes one type variable.
+I.e., it's fully applied in terms of type.
+In terms of terms, `id` takes one argument, and is indeed applied to one.
+So `id` is fully applied! And we inline it.
+
+However, reducing the _body_ `id {int -> int} (\x -> x +1)` leaves us `(\x -> x +1)`,
+a function of arity 1, not 0.
+
+In this case, inlining and reducing `id` reduces the amount of code,
+so even though it reduces to a not fully applied function, it may be desirable.
```
zliu41, line 124: Then what's the problem? Ininling it is desirable, and we indeed inline it.

mjaskelioff (approved):

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,206 @@
...
+```haskell
+let f = (\x.\y.x+y) 4
+in f 5
+```
+
+With beta-reduction, `f` becomes `\y.4+y` and it has an arity of 1.
+The _body_ `f 5` has an arity of 0 and thus is a fully applied function!
+We can reduce it to 9.
+However, because the _rhs_ `(\x.\y.x+y) 4` is a function application, not a lambda abstraction,
+we won't inline it with the current implementation.
```
zliu41, line 138: It depends on the order of optimizations. If we run beta reduction first, `f` will become `\y . 4+y`. If beta reduction runs after inlining fully saturated functions, it can still be inlined in the next simplifier pass (by default we run 4 passes).

zliu41 (changes requested):
I feel like this doesn't need to be a standlone ADR, because arity is simply calculated by counting lambdas. There's not really any alternatives. Maybe combine it with the "separate ADR" in

> The mechanics of whether or not to inline based on the call site will be covered in a separate ADR.

?

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
zliu41: It simply reduces code size, as shown in the first example in this ADR. I don't think "what's causing the speedup" is not understood because the example already demonstrates it.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,206 @@
...
+## Status
+
+Accepted
+
+## Context
+
+As part of our effort to increase script capacity,
+we take from [Plutonomy](https://github.com/well-typed/plutonomy),
+which gives 15-30% speedup on benchmarks.
+See [PLT-1044](https://input-output.atlassian.net/browse/PLT-1044).
```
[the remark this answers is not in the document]
thealmarty: Hmm...I think it's useful for internal reference but yeah it's not nice to link a non-public link here. Not sure I want to delete this though!

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,206 @@
...
+For a first round of implementation, we are adding a simplified pass.
+We will refine the implementation over time.
+In each round, we benchmark and test and investigate the results.
+
+### The rationale: Counting lambdas
+
+**Arity** is relevant when we talk about a fully applied function.
+A fully applied function has an arity of 0.
+The arity of a function is the number of arguments it takes minus the number of arguments applied.
+E.g., `\x.\y.x+y` has an arity of 2. `(\x.\y.x+y) 1` has an arity of 1.
```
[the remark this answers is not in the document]
thealmarty: I suppose the question is whether the goal is to inline fully reducible terms or not. Which "arity" should I be talking about?

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,206 @@
...
+We pattern match the _rhs_ with `LamAbs` or `TyAbs` (lambda abstraction for terms or types),
+and count the number of lambdas.
+Then, in the _body_, we check for term or type application (`Apply` or `TyInst`)of _v_.
+
+If _v_ is fully applied in the body, i.e., if
+
+1. the number of type lambdas equals the number of type arguments applied, and
+2. the number of term lambdas equals the number of term arguments applied,
+
+we inline _v_, i.e., replace its occurrence with _rhs_ in the _body_ (**if other call site inlining
```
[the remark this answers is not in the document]
thealmarty: that's in the "if other call site inlining conditions are satisfied"....I'll move that sentence earlier.

thealmarty:
>I feel like this doesn't need to be a standlone ADR, because arity is simply calculated by counting lambdas. There's not really any alternatives. 

This is mainly to discuss/record the concerns/imperfection of this implementation. Although like you pointed out, maybe the examples aren't good enough to cause concerns, or there are no concerns. I'll keep revising this and maybe add the other part.

zliu41:
> Although like you pointed out, maybe the examples aren't good enough to cause concerns, or there are no concerns

Yeah, I don't see any concerns with simply counting lambdas.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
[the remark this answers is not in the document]
zliu41: Just need a clarification: by definition `\x. \y. body` has aridy=1 (since we don't have multilambdas yet), but here we consider its arity=2 for the purpose of determining whether it is fully applied. Or maybe there's a different way of saying it without using "arity" at all.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
thealmarty: I was gonna discuss the other call site inlining condition too and with `sizeIsAcceptable`, `g` will not be inlined anyway. (We reject `LamAbs`) Perhaps I will just work on the code as is planned (counting lambdas, `acceptable` condition) and see how it goes. I can come back with results in the ADR. Bottom line is that this needs more work.

zliu41: > with sizeIsAcceptable, g will not be inlined anyway.

`g` _will_ be inlined, because once `g` is fully applied, its body is just a single variable, `f`.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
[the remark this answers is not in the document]
thealmarty: Agreed, I'll remove mentions of arity, like I said in [the other comment](https://github.com/input-output-hk/plutus/pull/5027#discussion_r1060763054).

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+As part of our effort to increase script capacity,
+we take from [Plutonomy](https://github.com/well-typed/plutonomy),
+which gives 15-30% speedup on benchmarks.
+See [PLT-1044](https://input-output.atlassian.net/browse/PLT-1044).
+
+The goal is to integrate what we know already works there and
+also add optimizations we can do before UPLC (Plutonomy only applies optimization to UPLC).
+We currently have an inliner pass for PIR and UPLC.
+One of the optimization we can add is inlining of saturated functions in both PIR and UPLC.
+From our analysis, this is one of the optimizations that contributes to the speedup in Plutonomy.
```
michaelpj, line 24: Is this the totality of our justification? We discussed that it wasn't totally clear _why_ inlining saturated functions is good, did you have more thoughts on that? which also bears on what we consider to be saturated

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+For example:
+
+```haskell
+let f x y = 1
+    y = f q
+in f a b 
+```
+
+we should inline `f` on line 3 (even though it’s used more than once), because it’s saturated,
+i.e. it has all its arguments, and we can reduce it.
```
michaelpj, line 35: ... whereas the `f` on line 2 only has one of its arguments so is not saturated

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+
+To do this we need to count the number of type/term lambda abstractions,
+so we know how many term/type arguments they have.
+We also need to decide whether or not to inline based on the call site, which we don’t currently.
+
+## Decision
+
+For a first round of implementation, we are adding a simplified pass.
+We will refine the implementation over time.
+In each round, we benchmark and test and investigate the results.
```
michaelpj, line 46: I think I'm confused by the point of the ADR. What is the thing we are deciding? This feels like a design doc for the inliner that you're making into an ADR. Maybe not everything needs to be an ADR!

I could see a _decision_ about whether or not we should do inlining of saturated functions, but that's not what you're talking about.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+We pattern match the _rhs_ with `LamAbs` or `TyAbs` (lambda abstraction for terms or types),
+and count the number of lambdas.
+Then, in the _body_, we check for term or type application (`Apply` or `TyInst`) of _v_.
+
+If _v_ is fully applied in the body, i.e., if
+
+1. the number of type lambdas equals the number of type arguments applied, and
+2. the number of term lambdas equals the number of term arguments applied, and
```
michaelpj, line 65: ... in the right order!

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+so we know how many term/type arguments they have.
+We also need to decide whether or not to inline based on the call site, which we don’t currently.
+
+## Decision
+
+For a first round of implementation, we are adding a simplified pass.
+We will refine the implementation over time.
+In each round, we benchmark and test and investigate the results.
+
+### The heuristic: Counting lambdas and arguments
```
michaelpj, line 48: Why is this a heuristic? What's it an approximation to?

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+
+In this inline pass we check for a binding that is a lambda abstraction and is fully applied in the body.
```
michaelpj, line 50: Is this a pass? I thought this was just a conceptual heuristic. I'm confused.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+We pattern match the _rhs_ with `LamAbs` or `TyAbs` (lambda abstraction for terms or types),
+and count the number of lambdas.
+Then, in the _body_, we check for term or type application (`Apply` or `TyInst`) of _v_.
+
+If _v_ is fully applied in the body, i.e., if
+
+1. the number of type lambdas equals the number of type arguments applied, and
+2. the number of term lambdas equals the number of term arguments applied, and
+
+if other call site inlining conditions are satisfied, (we currently inline too little, this will be improved later, see below)
```
michaelpj, line 67: Are we going to check the conditions on `rhs` or on the body of the function in `rhs` after the lambdas have been stripped off? I thought we were going to do the latter? Seems important.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+
+#### Example 1: function in body
+
+```haskell
+let f = \x . x
+     g = \y . f
+in g a
+```
+
+`f` and `g` each has 1 lambda. However, `g`'s _body_ includes `f` which also has a lambda.
```
michaelpj, line 93: I'm confused, why is this example not "fully reducible/applied"? I don't think we have a clear concept of what that means? This is the stuff I'd really like us to be clear about: what _exactly_ do we mean by "saturated", "fully applied" etc. There are several options!

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+in id {int -> int} (\x -> x +1)
+```
+
+Here we have a type application for a function that takes one type variable.
+I.e., it's fully applied in terms of type.
+In terms of terms, `id` takes one argument, and is indeed applied to one.
+So `id` is fully applied! And we inline it.
+
+Reducing the _body_ `id {int -> int} (\x -> x +1)` leaves us `(\x -> x +1)`,
+a function with 1 lambda. Even though it's not fully applied, inlining and reducing `id` reduces the amount of code, as desired.
```
michaelpj, line 111: This is the first time you've mentioned that reducing code size is desirable.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+```haskell
+let f = (\x.\y.x+y) 4
+in f 5
+```
+
+With beta-reduction, `f` becomes `\y.4+y` and it has 1 lambda.
+The _body_ `f 5` is a fully applied function!
+We can reduce it to 9.
+However, because the _rhs_ `(\x.\y.x+y) 4` is a function application, not a lambda abstraction,
+we won't inline it with the current implementation. This is something we may be able to improve.
```
michaelpj, line 124: I think this isn't _that_ concerning, since we can probably rely on our other optimization passes to turn this into something we can deal with.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+
+### The implementation: Counting lambdas and applications
+
+We run `countLam` to the _rhs_ of the binding:
+
+```haskell
```
michaelpj, line 130: I'm not sure this conveys more than the english definition of what you're doing that you already gave? and it's more error-prone to have a big hunk of code in here.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+
+Another substantial component of this implementation is the actual inlining at the call site.
+Details of this to be worked out.
+
+## Future plans
+
+We have plans to improve this pass so that it will catch more cases that are desirable,
+and not inline undesirably.
+
+### Add beta-reduction pass
```
michaelpj, line 171: we already have one

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+
+If beta-reduction of _rhs_ is run before this pass, it can catch cases categorized by example 3.
+
+### Add inline size heuristic
+
+If we do some analysis of the inline size, we can better decide whether it's desirable to inline.
```
michaelpj, line 177: what's the "inline size"? I don't think we have a definition for that anywhere?

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
```diff
@@ -0,0 +1,185 @@
...
+Plutonomy also [uses a similar heuristic](https://github.com/well-typed/plutonomy/blob/14b9bd46084db1b785b3a99d55f7f10d38165ee8/src/Plutonomy/Hereditary/Transform.hs#L266).
+
+### Add more refined heuristic for call site inlining
```
michaelpj, line 180: I think we should just delete it, inlining of saturated functions is a superior version anyway.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
michaelpj: TBH I think most of this section would be fine as a long Note in `Inline.hs`. We already have many notes of comparable length.

michaelpj (commented):
I think this document is a bit confused about what it's trying to be. I'd also _really_ like to see discussion of what it means to be saturated and why/when inlining such things is good, since that's the core of the problem and also IMO the thing where we actually have to make difficult decisions.

doc/read-the-docs-site/adr/0005-inlining-saturated-functions.md
thealmarty: Yeah I'm thinking so too. I'm leaning toward not putting code in these docs.

thealmarty:
The content of this was added in as code comments, as discussed. Closing this.

PR #5028, opened 2022-12-22
ADR Term with type information

thealmarty:
Please see ADR5 ([PR#5027](https://github.com/input-output-hk/plutus/pull/5027)) . With the challenges described in ADR5, I'm proposing this. Thank you @mjaskelioff  for the idea!

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/doc/read-the-docs-site/adr/0006-term-with-type-information.md
+++ b/doc/read-the-docs-site/adr/0006-term-with-type-information.md
@@ -1,0 +1,96 @@
+# ADR 6: `Term` with type information
+
+Date: 2022-12
+
+## Authors
+
+Marty Stumpf <marty.stumpf@iohk.io>
+
+## Status
+
+Proposed
+
+## Context
+
+We are exploring various ways to increase script capacity and efficiency
+(see [PLT-1044](https://input-output.atlassian.net/browse/PLT-1044)).
+E.g., inlining saturated functions. See also ADR 5. It's a complex issue but type information can help.
+If we have the type information it can open up other ways we can improve efficiency.
+
+Currently, we don't have terms that are annotated with type information after they are typechecked.
+This ADR describes an implementation to add that without disrupting a large portion of the codebase.
+
+## Decision
+
+### Adding a term type with type information
+
+Currently, we have one type of terms for each PIR, PLC, and UPLC,
+and the terms don't carry type information from the type checker.
+E.g., our PIR [Term](https://github.com/input-output-hk/plutus/blob/06c5831ea5934beef2f4c4178a8c53771381f1bf/plutus-core/plutus-ir/src/PlutusIR/Core/Type.hs#L117) is:
+
+```haskell
+-- current PIR terms
+data Term tyname name uni fun a =
+                        -- Plutus Core (ish) forms, see note [Declarations in Plutus Core]
+                          Let a Recursivity (NonEmpty (Binding tyname name uni fun a)) (Term tyname name uni fun a)
+                        | Var a name
+                        | TyAbs a tyname (Kind a) (Term tyname name uni fun a)
+                        | LamAbs a name (Type tyname uni a) (Term tyname name uni fun a)
+                        | Apply a (Term tyname name uni fun a) (Term tyname name uni fun a)
+                        | Constant a (PLC.Some (PLC.ValueOf uni))
+                        | Builtin a fun
+                        | TyInst a (Term tyname name uni fun a) (Type tyname uni a)
+                        | Error a (Type tyname uni a)
+                        | IWrap a (Type tyname uni a) (Type tyname uni a) (Term tyname name uni fun a)
+                        | Unwrap a (Term tyname name uni fun a)
+                        deriving stock (Functor, Show, Generic)
+```
+
+We proposed to add a term type that is distinguished from the current type in that it carries [
+type](https://github.com/input-output-hk/plutus/blob/06c5831ea5934beef2f4c4178a8c53771381f1bf/plutus-core/plutus-core/src/PlutusCore/Core/Type.hs#L93)
+information.
+The proposed term types include a generic term type,
+a term type with type information, and a term type without type information (what we have currently).
+
+```haskell
+-- the new generic term: it has an additional parameter: tyinfo
+data GenericTerm tyinfo tyname name uni fun a =
+    Let tyinfo a Recursivity (NonEmpty (Binding tyname name uni fun a)) (Term tyname name uni fun a)
+    | Var tyinfo a name
+    | TyAbs tyinfo a tyname (Kind a) (Term tyname name uni fun a)
+    | LamAbs tyinfo a name (Type tyname uni a) (Term tyname name uni fun a)
+    | Apply tyinfo a (Term tyname name uni fun a) (Term tyname name uni fun a)
+    | Constant tyinfo a (PLC.Some (PLC.ValueOf uni))
+    | Builtin tyinfo a fun
+    | TyInst tyinfo a (Term tyname name uni fun a) (Type tyname uni a)
+    | Error tyinfo a (Type tyname uni a)
+    | IWrap tyinfo a (Type tyname uni a) (Type tyname uni a) (Term tyname name uni fun a)
+    | Unwrap tyinfo a (Term tyname name uni fun a)
+    deriving stock (Functor, Show, Generic)
+
+-- the new term type that is similar to our current implementation, with tyinfo set to ()
+type Term = GenericTerm () tyname name uni fun a
+
+-- the new term type that has type information
+type TermWithTypeInfo = GenericTerm (Type tyname uni ann) tyname name uni fun a
+```
+
+### Using the `PatternSynonyms` pragma to minimize code change
+
+Using the `PatternSynonyms` extension, we can now add these synonyms so that all existing
+code that pattern matches on the terms work:
+
+```haskell
+pattern Let = Let ()
+pattern Var = Var ()
+pattern TyAbs = TyAbs ()
+...
+```
+
+### Using the type checker to generate the type information
+
+Now that we have a type of terms with type information, the type checker can return a `TermWithTypeInfo` so that we can utilize these for our optimization passes.
+
+## Conclusion
+
+This proposal is still a significant change to our codebase. However, it's benefit may outweigh the cost. Script capacity and efficiency is important and having type information can expand and improve optimizations we can do.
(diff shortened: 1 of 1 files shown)


effectfully (commented):
Why can't we reuse the type of annotations?

zliu41:
Is this necessary, given that the type of a term can be calculated on the fly when needed?

thealmarty:
> Why can't we reuse the type of annotations?

If you mean the `a` in `Term tyname name uni fun a` then yes, that's possible, but then we don't get the line/source info, and that may not be ideal?

mjaskelioff:
> > Why can't we reuse the type of annotations?
> 
> If you mean the `a` in `Term tyname name uni fun a` then yes, that's possible, but then we don't get the line/source info, and that may not be ideal?

One could think of having an annotation pair of (source code info,typeinfo), but the annotation parameter is passed to Type, where the typeinfo doesn't make sense.

thealmarty:
> > > Why can't we reuse the type of annotations?
> > 
> > 
> > If you mean the `a` in `Term tyname name uni fun a` then yes, that's possible, but then we don't get the line/source info, and that may not be ideal?
> 
> One could think of having an annotation pair of (source code info,typeinfo), but the annotation parameter is passed to Type, where the typeinfo doesn't make sense.

Also I think changing the annotation to a pair would disrupt more of the code base than this?

doc/read-the-docs-site/adr/0006-term-with-type-information.md
```diff
@@ -0,0 +1,96 @@
...
+    | Constant tyinfo a (PLC.Some (PLC.ValueOf uni))
+    | Builtin tyinfo a fun
+    | TyInst tyinfo a (Term tyname name uni fun a) (Type tyname uni a)
+    | Error tyinfo a (Type tyname uni a)
+    | IWrap tyinfo a (Type tyname uni a) (Type tyname uni a) (Term tyname name uni fun a)
+    | Unwrap tyinfo a (Term tyname name uni fun a)
+    deriving stock (Functor, Show, Generic)
+
+-- the new term type that is similar to our current implementation, with tyinfo set to ()
+type Term = GenericTerm () tyname name uni fun a
```
effectfully, line 72: Should be `type Term = GenericTerm ()`.

doc/read-the-docs-site/adr/0006-term-with-type-information.md
```diff
@@ -0,0 +1,96 @@
...
+
+-- the new term type that has type information
+type TermWithTypeInfo = GenericTerm (Type tyname uni ann) tyname name uni fun a
```
effectfully, line 75: Here we do need the type parameters, but they need to be bound on the LHS. Plus there's that confusion of `ann` vs `a`. I'm much on the side of always using `ann` instead of `a` to convey the semantics.

effectfully:
It seems weird to me to have annotations and then add a new type variable whenever we need some special kind of annotations.

@mjaskelioff 

> One could think of having an annotation pair of (source code info,typeinfo), but the annotation parameter is passed to Type, where the typeinfo doesn't make sense.

True, but I'd rather generalize `ann` to make it aware of the context (type vs term) as it's what I wished we'd do to `tyname` vs `name`, so that we don't have two type parameters just for the names (particularly since we probably never instantiate them to different things? Not sure).

Or add a new type variable and call it `tyann`. `tyinfo` seems way too specific to make it right into the AST.

@zliu41 

> Is this necessary, given that the type of a term can be calculated on the fly when needed?

Repeated type checking is way too expensive compared to doing it once and then using the type info. QuviQ's generator suffer from this problem too.

Type checking was actually implemented via putting types into annotations initially, but then I changed it when I reimplemented the type checker (for separate reasons) and I believe that was a mistake on my part.

effectfully:
@thealmarty 

> Also I think changing the annotation to a pair would disrupt more of the code base than this?

I personally don't care at all about the scale of disruption as long as things get improved and there's little subtlety going on (to avoid potentially introducing bugs). YMMV

zliu41:
> Repeated type checking is way too expensive compared to doing it once and then using the type info

Ideally I'd like to see some evidence and/or discussion of this in the ADR.

GHC's `exprType` basically calculates the type of a `CoreExpr` on the fly. If GHC doesn't need to annotate `CoreExpr`s with types, we probably don't either.

effectfully:
> Ideally I'd like to see some evidence and/or discussion of this in the ADR.

That would be great indeed.

> GHC's `exprType` basically calculates the type of a `CoreExpr` on the fly. If GHC doesn't need to annotate `CoreExpr`s with types, we probably don't either.

GHC breaks out of that each time there's a `case` (see `Note [Why does Case have a 'Type' field?]` in [this](https://downloads.haskell.org/ghc/9.2.4/docs/html/libraries/ghc-9.2.4/src/GHC-Core.html) file) and pattern matching is very common in Haskell. Plus, GHC doesn't have a single huge expression like Plutus Core does, instead it has a sequence of bindings, so it never needs to process the entire program at once. And Plutus Core programs are notoriously known for having an absolutely ridiculous size (GHC can produce huge bindings too, though).

Here's another data point regarding GHC: see `Note [Kind caching]` in [this](https://hackage.haskell.org/package/base-4.17.0.0/docs/src/Data.Typeable.Internal.html) file (it's about `Typeable`, not internals of the compiler).

doc/read-the-docs-site/adr/0006-term-with-type-information.md
thealmarty: Totally agree about `ann`!

thealmarty:
I've added some of the discussion here to the ADR. I haven't changed any of the proposed implementation yet. I'd like to wait until people are back from the holiday to join in the discussion first.

thealmarty:
@effectfully good to know about Quviq's generator regarding this. Will adding infra for type info (however we choose to do it in the end) help with the generator too? Or is that too late?

zliu41:
> GHC breaks out of that each time there's a case (see Note [Why does Case have a 'Type' field?] in [this](https://downloads.haskell.org/ghc/9.2.4/docs/html/libraries/ghc-9.2.4/src/GHC-Core.html) file) and pattern matching is very common in Haskell. Plus, GHC doesn't have a single huge expression like Plutus Core does, instead it has a sequence of bindings, so it never needs to process the entire program at once. And Plutus Core programs are notoriously known for having an absolutely ridiculous size (GHC can produce huge bindings too, though).

I see. Both points are quite convincing.

I'm also in favor of using `ann`, because it is a kind of annotation.
75timi

thealmarty:
We may be able to use pattern synonyms to disrupt the code base less with a pair for the `ann` too. It's best if pattern matching still works with all our and external partners' tests etc.

```haskell
-- I don't know if we can do this
pattern Let _ _ .. () = Let _ _ .. ((),())
...
pattern Let _ _ .. ann = Let _ _ .. (ann, ())
...
```

effectfully:
> @effectfully good to know about Quviq's generator regarding this. Will adding infra for type info (however we choose to do it in the end) help with the generator too? Or is that too late?

Hard to tell, but probably. It'll mostly help with the shrinker, but then I'm not sure how much we care about the performance of the shrinker. Perhaps quite a bit, given how ridiculously slow shrinking tests are? Not sure. I don't think it's too late.

> We may be able to use pattern synonyms to disrupt the code base less with a pair for the `ann` too. It's best if pattern matching still works with all our and external partners' tests etc.

Instead of hardcoding pairs into the AST we would keep the current general `ann` and turn it into `(tyinfo, SrcLoc)` or whatever when we need that, so it'd only affect things that need type information.

bezirg:
I am also in favor of reusing/extending `ann`

michaelpj:
I'm kind of confused. What, specifically, do we want this for? ADR5 doesn't have any examples where we need type information.

Moreover, if we want this in PIR (which is where we do most optimization), then I would expect that:
1. We mostly care about the types of variables
2. The variables we care about are mostly let-bound

and let-bound variables have _type annotations_! So if you need to know the type of let-bound variables in a pass, just put the annotations in a map as you traverse the program.

So it would help to have an example where we a) need this b) for a non-variable or non-let-bound variable.

(As a historical note, I think long ago the PLC typechecker used to work exactly like this: it produced a program annotated with types. But it was never useful and so we dropped it as a simplification.)

mjaskelioff:
I suggested using type information to help deciding if a function is saturated or not. Would the type information be (cheaply) available in some other way?

michaelpj:
In PIR we have e.g.
```
let f : Int -> Int = \x . x +1
in ... f 1 ...
```
so yeah, we have the type annotation right there in the AST, and we can assume it's correct (otherwise the typechecker will fail).

I'm not sure if we want to use type information to check for saturation. But perhaps this is part of the discussion of what the _point_ of inlining saturated things is. I think it's mostly about inlining things that we _know_ will reduce, and that's a _syntactic property. Consider this:
```
\(f :: Int -> Int) .
let g :: Int -> Int = f
in g 1
```
Should we inline `g`? If you looked at the type you would say "yes it has one argument", but it won't actually reduce! We need to look at the _body_ of `g` to tell whether or not it will reduce.

michaelpj (commented):
I also think we should 100% use `ann` for this if we do it. Easy and cheap, it's also what we did in the past when we did this.

mjaskelioff:
I agree that the point of inlining saturated things should be discussed. Perhaps, more generally, the discussion could nbe about what should be inlined.

In your example above inlining `g` might be beneficial even if it doesn't reduce. For example, it might trigger dead code elimination for the `let g` and then you have a smaller term.

thealmarty:
Oh, bindings *do* have type info! But the body doesn't. 
```haskell
data Binding tyname name uni fun a = TermBind a Strictness (VarDecl tyname name uni a) (Term tyname name uni fun a)
...

data VarDecl tyname name uni ann = VarDecl
    { _varDeclAnn  :: ann
    , _varDeclName :: name
    , _varDeclType :: Type tyname uni ann
    } deriving stock (Functor, Show, Generic)
makeLenses ''VarDecl
```
So I think being able to get the type info for a term (body) is still useful?

Since we are not hard coding `ann`, there's not really anything that needs doing beside having an extra function that is like the type checker but returns a term with type annotations. It doesn't need an ADR so I'm closing this. Quite like this way of presenting the idea though! Seems to get more discussion/reactions than on slack.

michaelpj:
> So I think being able to get the type info for a term (body) is still useful?

My argument [above](https://github.com/input-output-hk/plutus/pull/5028#issuecomment-1370775099) is that it isn't. Which is why I really want actual examples where this is useful! Otherwise this discussion is very difficult!

thealmarty:
@michaelpj I'll come back to this after I'm done a first implementation of call site inlining - I think I can answer better then. Thanks for the comments!

PR #5032, opened 2023-01-03
PLT-1147: Compiler pass for floating bindings inwards

zliu41:
See Haddock in `PlutusIR.Transform.LetFloatIn` for details.

This PR implements float-in for non-recursive term bindings only. Cost savings are observed in all affected test cases except one.


plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
...
+    Term tyname name uni fun a ->
+    m (Term tyname name uni fun a)
+floatTerm ver = pure . fmap fst . go <=< PLC.rename
+  where
+    -- | Float bindings in the given `Term` inwards, and calculate the set of
+    -- variable `Unique`s in the result `Term`.
+    go ::
+        Term tyname name uni fun a ->
+        Term tyname name uni fun (a, Uniques)
+    go t = case t of
```
thealmarty, line 62: you can use the lambda case syntax here

zliu41: Technically yes, but since `t` is used below, this way is more convenient.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
...
+                us = Set.union (termUniqs body) (foldMap bindingUniqs bs)
+             in Let (a, us) Rec bs body
+        Var a n -> Var (a, Set.singleton (n ^. PLC.unique)) n
+        Constant{} -> (,mempty) <$> t
+        Builtin{} -> (,mempty) <$> t
+        Error{} -> (,mempty) <$> t
+
+    -- | Float bindings in the given `Binding` inwards, and calculate the set of
+    -- variable `Unique`s in the result `Binding`.
+    go' ::
```
thealmarty, line 102: we should avoid naming functions with `'` - it's in the style guide. Maybe `goTerm` and `goBinging` or something?

zliu41: Sure, although I think that particular guideline makes much more sense for exported functions than it does for private/inner functions.

thealmarty: I agree it's even more important for exported functions, but it does improve readability so it's good to do for everything, more consistent too!

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
...
+        Binding tyname name uni fun a ->
+        Binding tyname name uni fun (a, Uniques)
+    go' = \case
```
michaelpj, line 105: Again, this would be easy to do with a traversal if you weren't using the annotation field

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
...
+
+ Unlike the float-in in GHC, we /can/ float bindings into lambda abstractions,
+ because doing so cannot duplicate work. Therefore we simply float a binding
+ inwards to the deepest possible position.
+
+ We avoid moving a strict binding whose RHS is not a value, because doing so
+ may either change the semantics (if the RHS has effects) or duplicate work
+ (if it is floated into a lambda abstraction).
+
+ Floating inwards is done in a single pass. The subterms are processed bottom-up,
```
michaelpj, line 20: I actually think it might be nicer to do it in two. Annotating the AST with used variable sets is generically useful, and I think it would make the code clearer. We already do lots of passes, one more isn't so bad, it's only linearly more work.

michaelpj: Thinking about it some more, I can think of at least one reason not to do this. Floating in changes the set of used variables, so if you have something like this:
```
let x = 1
in \y .
let z = x
in \p .
z
```
you want to float in both let bindings the whole way. But if you did the free variable computation first then you could only do this:
```
\y .
let x = 1
in \p .
let z = x
in z
```
because you think that the `\p . ...` uses `x`, even though it doesn't any more.

But it would be good to explain this (https://www.michaelpj.com/blog/2022/04/24/on-commenting-code.html#road-not-travelled-comments).

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
+{-# LANGUAGE BangPatterns      #-}
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TupleSections     #-}
+{-# LANGUAGE ViewPatterns      #-}
+
+{- | Float bindings inwards. A binding can be floated if it is a lazy binding, or
```
michaelpj, line 7: I was expecting some discussion of the implementation approach, it's not _totally_ obvious. In particular, it's _very_ different from the float out pass. Did you look at that/talk to Nikos about it? Why are they different? Future readers will want to know (as do I!).

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
...
+        TermBind (_, usBind) _strictness var _rhs -> case body of
+            Apply (a, usBody) fun arg
+                | (var ^. PLC.unique) `Set.notMember` termUniqs fun ->
+                    -- `fun` does not mention `var`, so we can place the binding inside `arg`.
+                    Apply (a, Set.union usBind usBody) fun (go b arg)
+                | (var ^. PLC.unique) `Set.notMember` termUniqs arg ->
+                    -- `arg` does not mention `var`, so we can place the binding inside `fun`.
+                    Apply (a, Set.union usBind usBody) (go b fun) arg
+            LamAbs (a, usBody) n ty lamAbsBody ->
+                -- A binding can always be placed inside a `LamAbs`.
```
michaelpj, line 164: This is a BIG DEAL and needs an appropriately big signpost and probably a Note. The moment we float a let into a lambda we can repeat work, so we need to be very careful with this. Just unilaterally floating lets into lambdas seems pretty risky to me.

Consider
```
let x = expensive 
in fmap (\y -> y + x) [0...1000000]
```
vs
```
fmap (\y -> let x = expensive in y + x) [0...100000]
```

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
...
+                LamAbs (a, Set.union usBind usBody) n ty (go b lamAbsBody)
+            TyAbs (a, usBody) n k tyAbsBody ->
+                -- A binding can always be placed inside a `TyAbs`.
+                TyAbs (a, Set.union usBind usBody) n k (go b tyAbsBody)
+            TyInst (a, usBody) tyInstBody ty ->
+                -- A binding can always be placed inside a `TyInst`.
```
michaelpj, line 170:
```suggestion
                -- A term binding can always be placed inside the body a `TyInst` because the
                -- type cannot mention the bound variable 
```

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
...
+    Binding tyname name uni fun (a, Uniques) ->
+    Uniques
+bindingUniqs b = Set.union (snd (bindingAnn b)) $ case b of
+    TermBind _a _s var _rhs -> Set.singleton $ var ^. PLC.varDeclName . PLC.unique
+    _                       -> mempty
+
+noUniq :: Functor f => f a -> f (a, Uniques)
+noUniq = fmap (,mempty)
+
+{- | The algebra for floating a `Binding` inwards. Given a `Term` and a `Binding`, determine
```
michaelpj, line 126: I've noticed you like to call things "algebras". Is this _actually_ an algebra, though? I'm not sure that it is. What about "floatInBinding" - that's what it does, no?

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
...
+                LamAbs (a, Set.union usBind usBody) n ty (go b lamAbsBody)
+            TyAbs (a, usBody) n k tyAbsBody ->
+                -- A binding can always be placed inside a `TyAbs`.
+                TyAbs (a, Set.union usBind usBody) n k (go b tyAbsBody)
+            TyInst (a, usBody) tyInstBody ty ->
+                -- A binding can always be placed inside a `TyInst`.
+                TyInst (a, Set.union usBind usBody) (go b tyInstBody) ty
+            Let (a, usBody) r bs letBody
+                -- The binding can be placed inside a `Let`, if the right hand sides of the
+                -- bindings of the `Let` do not mention `var`.
```
michaelpj, line 174: What about pushing it into one of the bindings? i.e.
```
let x = 1
in 
let 
   y = x 
   z = 1
in ...
```
Would be nice to turn that into
```
let 
  y = let x = 1 in x
  z = 1
in ...
```

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
...
+
+ Floating inwards is done in a single pass. The subterms are processed bottom-up,
+ during which we keep track of the set of variable @Unique@s within each subterm.
+ The time complexity is linear wrt AST size (strictly speaking it is /O(n*log n)/
+ where the /log n/ comes from `Set` operations).
+
+ Currently we only move non-recursive term bindings. In the future we should also
+ implement float-in for other kinds of bindings.
+-}
+module PlutusIR.Transform.LetFloatIn (floatTerm) where
```
michaelpj, line 28: no tests? :(

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
michaelpj: Okay, it must be non-strict or pure in this case, but maybe it's still risky. Not sure. Worth a comment in any case, since this is a very important corner!

michaelpj:
Also needs a changelog entry!

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
zliu41: I didn't look into the details of the float-out pass. It is basically the opposite of float-in, so implementation-wise I imagine them to be totally different. @bezirg do you think there are parts of float-in and float-out that can be shared?

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
zliu41: I didn't know we have tests for individual optimizations. OK, I'll add some there.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
...
+    PLC.BuiltinVersion fun ->
+    Term tyname name uni fun a ->
+    m (Term tyname name uni fun a)
+floatTerm ver = pure . fmap fst . go <=< PLC.rename
+  where
+    -- | Float bindings in the given `Term` inwards, and calculate the set of
+    -- variable `Unique`s in the result `Term`.
+    go ::
+        Term tyname name uni fun a ->
+        Term tyname name uni fun (a, Uniques)
```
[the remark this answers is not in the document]
zliu41: I guess I don't have to use annotation. I did it just because `Uniques` conceptually _is_ a kind of annotation - we are annotating the Terms and Bindings with `Uniques` during the process.

I think using `(Uniques, Term tyname name uni fun a)` would be more or less equivalent. I don't see how it simplifies anything. `Writer`, on the other hand, would be inconvenient to use, because I'd be calling `runWriter` all the time.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
zliu41: It is an algebra. `foldr` is basically a catamorphism that turns an F-algebra into a fold.

To be precise, the `a -> b -> b` argument of `foldr` is only a part of the algebra. The full algebra is `(a -> b -> b, b)` (isomorphic to `ListF a b -> b`). But technically it is itself also an algebra.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
zliu41: Hmm, good point, I didn't think of this!

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,188 @@
...
+            IWrap (a, usBody) ty1 ty2 iwrapBody ->
+                -- A binding can always be placed inside an `IWrap`.
+                IWrap (a, Set.union usBind usBody) ty1 ty2 (go b iwrapBody)
+            Unwrap (a, usBody) unwrapBody ->
+                -- A binding can always be placed inside an `Unwrap`.
+                Unwrap (a, Set.union usBind usBody) (go b unwrapBody)
+            _ ->
+                let us = Set.union (termUniqs body) (bindingUniqs b)
+                 in Let (letAnn, us) NonRec (pure b) body
+        _ ->
```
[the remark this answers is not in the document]
zliu41: Yeah, I think this PR is like 90% of the work. But I still prefer multiple relatively small PRs to one big PR.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
zliu41: Well, I tried it (https://github.com/input-output-hk/plutus/commit/c88f5b088da588caaab684fd26c42e746bced05b), but it makes things worse. It's hard to tell exactly why since the pir/plc are quite big, but my guess is that once we turn `let a = ... in let b = a in body` into `let b = (let a = ... in a) in body`, the cost and size of `b` are no longer acceptable, which affects inlining.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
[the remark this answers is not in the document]
michaelpj: It means you're not constantly fetching things in and out of the annotation. You'd just be matching on a simple tuple returned from `go` instead of then also having to mangle the annotations back and forth constantly. I'd have to try it to be sure.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
michaelpj: Well, both of them have a common activities:
- Identify let bindings that can be moved
- Identify the correct place to put the let bindings in, dealing with e.g. lambdas and so on
- Put the bindings in

They're both passes that move let bindings to optimal scopes! 

If anything, I would have thought they would be very similar, with float in aiming to put them "as low as possible" and float out looking to put them "as high as possible". But you should definitely look!

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
michaelpj: I guess I'm not sure I find that actually informative about what it does.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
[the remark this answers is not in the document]
michaelpj: If you're planning to do it later then a comment would have stopped me from asking :)

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
michaelpj: A signpost explaining why it's fine would have stopped me getting the wrong end of the stick here :)

michaelpj:
Also on the float-out thing: it might be the case that the float-out pass is too complicated and we could redo it more like this pass to make it simpler. That would also be a win!

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
zliu41: > Identify let bindings that can be moved

This is basically "lazy or pure". The `isPure` function is already shared.

> Identify the correct place to put the let bindings in, dealing with e.g. lambdas and so on

The float-out pass has a `calcFreeVars`, which _may_ be reusable by float-in. The float-in pass just collects all `Unique`s, but it should probably collect `FreeVars` instead, which can be a much smaller set than `Unique`s. I'll investigate in this separately. Other than this, there's not much in common between float-out and float-in.

> Put the bindings in

That's just a matter of constructing a `Let` term. Not much to share either.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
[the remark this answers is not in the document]
zliu41: Actually I think using annotation is much more convenient. Because in `Term (ann, Uniques)`, each subterm is also annotated with `Uniques`, which is what's needed, since I need to keep descending. `(Term ann, Uniques)` won't work.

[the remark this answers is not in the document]
michaelpj: Yes, I figured out that you do genuinely need it for later processing. I still think it might make the annotating-part simpler but whatever.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
michaelpj: What about the edge cases I've mentioned in comments on this PR? There are lots of interesting cases specifically for float-in! And it's not that clear that the test for float-in are appropriate...

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
michaelpj: I mean, I think we can do the obviously sensible floating in behaviour here and worry about making the inliner better separately.

And there's really no reason not to add an indicative test case.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,210 @@
...
+        TermBind _a Strict _var rhs
+            -- A strict binding whose RHS is not pure cannot be moved.
+            | not (isPure ver (const NonStrict) rhs) ->
+                let us = Set.union (termUniqs body) (bindingUniqs b)
+                 in Let (letAnn, us) NonRec (pure b) body
+        TermBind (_, usBind) _strictness var _rhs -> case body of
+            Apply (a, usBody) fun arg
+                | (var ^. PLC.unique) `Set.notMember` termUniqs fun ->
+                    -- `fun` does not mention `var`, so we can place the binding inside `arg`.
+                    Apply (a, Set.union usBind usBody) fun (go b arg)
```
michaelpj, line 171: This is a little subtle and it took me a while to realise exactly why we need this union: it's so that we consider the variables used by the _right-hand side_ of the binding to be used in this branch of the application. I'd really like a test case that demonstrates that we handle this properly.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
michaelpj: Well, your code _does_ have to update the annotated sets as it pushes things in, so I'm not sure this version is simpler. Indeed, I'm not 100% confident it's all doing the right thing... if we had a pass for annotating free/used variables then we could also test that independently!

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
michaelpj: ... and the test cases for float-out include some comments that don't make any sense for float-in

plutus-benchmark/nofib/test/formulaBudget.budget.golden
```diff
@@ -1,3 +1,3 @@
-({ cpu: 37605791908
-| mem: 161972648
+({ cpu: 26593253908
```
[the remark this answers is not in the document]
kwxm: I wonder why this one is improved so much compared to the others...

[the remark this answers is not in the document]
zliu41: > I wonder why this one is improved so much compared to the others...

@kwxm This is my understanding: the reason float-in reduces costs is because if a binding is floated into an unevaluated branch, it reduces the number of lambda abstractions and applications (which are what bindings are compiled into) as well as delays (because the arguments of the applications are usually inside `delay`s) the CEK machine needs to perform. This particular test case probably has many bindings floated into unevaluated branches. Here are the number of CEK steps:

| step kind | without float-in | with float-in |
| ------------- | ------------- | ------------- |
| BLamAbs  | 594154  | 434552  |
| BApply  | 496702  | 337100 |
| BDelay  | 195463  | 35861 |

The rest of the step kinds are identical between the two.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
zliu41: @michaelpj I think I know why - it has nothing to do with inlining. It's because of floating bindings into lambdas. We only float lazy or strict pure bindings, so when floating into a lambda, the RHS itself is not evaluated more times. However, because a binding is compiled into `(\x -> body) (delay rhs)`, it incurs more machine steps: Delay, Apply and LamAbs. This is confirmed by counting the machine steps - only these three steps increased.

The `checkScriptContext` test cases have `fmap mkTxOut [1..i]`, so this all makes total sense now. We should avoid floating bindings into lambdas.

michaelpj: Right, so there's some overhead from the let-binding itself, that's a good point. But that seems like a problem which is _independent_ of what we're discussing in this thread, right? Which is about floating things into the RHS of let-bindings. So I'm not sure that explains why we get an increase, unless I'm missing something.

> Delay, Apply and LamAbs. This is confirmed by counting the machine steps - only these three steps increased.

Most of the program is applications and lambda abstractions! I'm not sure how much this tells us...

zliu41: > But that seems like a problem which is independent of what we're discussing in this thread, right? Which is about floating things into the RHS of let-bindings. So I'm not sure that explains why we get an increase, unless I'm missing something.

This is exactly why! The RHS of let-bindings contains lambdas, and by floating things into RHS, we float them into lambdas.

Another way to demonstrate this is

> fmap mkTxOut [1..i]

When `i=0`, floating things into RHS saves cost. When `i=1`, it breaks even. The larger `i` is, the worse it gets.

zliu41: > Most of the program is applications and lambda abstractions! I'm not sure how much this tells us...

Btw, it can also be observed that when floating things into RHS, `Delay`, `Apply` and `LamAbs` increase by the same number, and that number equals `i-1`. The numbers of all other machine steps remain the same.

michaelpj: Ah, so indeed the correct fix is to be more careful about floating into lambdas. That makes sense why you wouldn't have seen the problem before - indeed most lambdas probably appear on the RHS of bindings.

Well, I feel vindicated for claiming that the decision of whether or not to float into lambdas is complicated and needs a careful comment at least :)

michaelpj: I'm confused. I think some examples would really really help (as test cases!).

I actually think that what I said was wrong. It's not _lambdas_ that cause multiple evaluation, it's _applications_. The only way that something can get evaluated multiple times is by being bound to a variable and then having the variable used multiple times, i.e. by being in the _argument_ of an application.

e.g. 

```
let x = ...
in [(\y . ... y... y...) x]
```
x gets evaluated multiple times, despite not being in a lambda.

And of course the RHSs of bindings _are_ arguments of applications and indeed can lead to things being evaluated more often. If you pushed a let-binding into the RHS of the x binding above it would get evaluated twice.

So I think the strict rule is:
- Don't float into the arguments of applications
- Don't float into the RHS of bindings

Maybe we can relax that? I'm not sure. But this needs a careful note and some examples.

zliu41: Hmm, I have to think about it a bit more. I thought lambdas with a bigger body is obviously more expensive if applied multiple times, but I guess it's not that simple?

zliu41: Well, it's not about how expensive `expensive` is. The `expensive` itself will always be evaluated the same number of times (as long as `x` is lazy; if `x` is strict, we wouldn't move `let x = expensive`). It's about how many times `delay expensive` is evaluated. Each time `delay expensive` is evaluated, it incurs an additional LamAbs, Apply and Delay step.

So how many times is `delay expensive` evaluated in this example? The compiled code for the three cases are, respectively
```
(\x -> (\f -> f1 f2) (\y -> ..x..)) (delay expensive)
(\f -> f1 f2) ((\x -> (\y -> ..x..)) (delay expensive))
(\f -> f1 f2) (\y -> (\x -> ..x..) (delay expensive))
```

As far as I can tell, case 1 evaluates `delay expensive` once, while cases 2 and 3 evaluate it twice. So I think what you said previously:

> So I think the strict rule is:
Don't float into the arguments of applications
Don't float into the RHS of bindings

is correct.

zliu41: Wait, I was wrong - case 2 also evaluates `delay expensive` only once. I wrote and ran this test, and 1 and 2 have the same cost, and 3 has a higher cost. So, it should probably be "Don't float into lambdas within RHSs of bindings" and "Don't float into lambdas within arguments of applications"...

kwxm: I'm going to add a small note of caution here.  From my experiments with Plutonomy (see the notes attached to PLT-1044), it seemed that the float-in optimisation was most effective in combination with some of the other optimisations.  Rather than trying to make this particular transformation as good as possible at this stage, it might be sensible to implement something simple and then add other optimisations as separate steps in the pipeline.  In the long run that might amount to the same thing as having one big super-efficient inliner, but it could be conceptually simpler if we have a number of separate stages.  I'm just guessing though: I could easily be wrong!

Also, I think part of the value of this is understanding _why_ all this stuff works, and I think we're beginning to see that in some of the discussion here.  We should try to make a record of the thinking that's been going on...

zliu41: @michaelpj Ok, I've tried a number of alternatives on our existing tests, including
- no floating into any lambda abstractions
- no floating into any arguments
- no floating into lambda abstractions that are in RHSs of bindings
- no floating into lambda abstractions that are in RHSs of bindings or in arguments

And these are all _strictly worse_ than what this PR currently does (i.e., no floating into RHSs). Here's my explanation:
- Not floating into arguments at all is bad, because in an argument there may be unevaluated branches, and floating bindings into those branches saves cost.
- Not floating into lambdas at all is bad, for similar reasons. Also many lambdas may not be applied at all.
- Floating into lambdas inside RHSs can be bad, as demonstrated in Michael's example above (case 3).
- Theoretically, if floating into RHSs can be bad, then floating into arguments can also be bad, because `let x = RHS in body` is the same as `(\x -> body) RHS`. However, due to GHC's transformations we get the former much more often than the latter (even if we don't run our own beta-reduction pass).
- Floating into RHSs, even without floating into lambdas, can _also be bad_! The following is an example:

```haskell
let x=1+2
 in let y=x+x
     in y+y
```

Without floating, this is turned into
```
(\x -> (\y -> y+y) (delay $ x+x)) (delay 1+2)
```

and `delay ...` is evaluated twice. If `x` is floated into the RHS of `y`, then it is turned into

```
(\y -> y+y) (delay $ (\x -> x+x) (delay 1+2))
```

and `delay ...` is evaluated three times (note the second `delay` is inside the first `delay`)!

Based on this observation, I think what this PR currently does is what we should do. @kwxm Hopefully this is a good thread to refer to in the future!

michaelpj: > Theoretically, if floating into RHSs can be bad, then floating into arguments can also be bad, because let x = RHS in body is the same as (\x -> body) RHS. However, due to GHC's transformations we get the former much more often than the latter (even if we don't run our own beta-reduction pass)

I don't find this convincing. These two _are_ very similar, and your argument for floating into arguments "there might be unevaluated branches" is just as valid for let-RHSs.

> Without floating, this is turned into

To be clear, you're talking about non-strict let-bindings here (we should _really_ be explicit, it makes a big difference!). Strict let-bindings don't have the delays, and are quite different.

----

One thing that I think we're looking for the answer to the question of "when is a let binding work-free?". I think the answer is that _no let-binding is ever work-free_ because of the abstraction/application overhead of processing the binding. That's an important thing to note, IMO it means it is _never_ fully safe to float _any_ let-binding into a context where we can repeat work.

We might want to relax this in some cases, since the overhead is small, but I think it's useful to note that strictly they're never work-free.

----

> Floating into RHSs, even without floating into lambdas, can also be bad! The following is an example:

I think what you're pointing to again is about _delaying evaluation_ anything that delays evaluation, which is lambdas, type abstractions, _and the rhs of non-strict bindings_ (I was missing this one!) is problematic and can lead to repeated evaluation.

I would buy a heuristic that says not to push anything that's not work-free into anything that can lead to repeated evaluation. I think you want to carve out some special cases but I'm not sure that what you've written is principled enough.

I think there is a variant of this that is principled, though. Given a term, I think we can identify the set of child nodes that will be evaluated _strictly the same number of times_ as the overall term. We do that by recursing into only the bodies of let-terms and the _functions_ (not arguments!) of applications. Then it _is_ safe to move even non-work-free bindings within this subset of nodes, since it'll still be evaluated strictly the same number of times.

-----

I don't think this is complete, but I think it's getting there. Perhaps for the first version we should skip the equally-evaluated thing and just not float into things that can delay evaluation.

-----

> Hopefully this is a good thread to refer to in the future!

THIS MUST GO INTO A NOTE!! I'm sure you know that but I just really want to emphasize this. Looking back for an old PR comment thread is _not_ an adequate way of recording this sort of knowledge! We really want the fully-processed version of this discussion recorded alongside the pass.

zliu41: > I would buy a heuristic that says not to push anything that's not work-free into anything that can lead to repeated evaluation. I think you want to carve out some special cases but I'm not sure that what you've written is principled enough.

Unfortunately, if we avoid floating into any RHS or any argument whatsoever, then there's zero savings among our test cases.

Although a lazy binding or a strict pure binding, I think it is unlikely that floating them inwards leads to unbounded cost increase. The additional cost from one more LamAbs, Apply and Delay is likely small compared to whatever real work is being done.

So unless there's a better approach that guarantees no cost increase whatsoever while preserving at least some benefits, we'd have to make a decision between not floating-in at all vs. accepting some small cost increases.

michaelpj: > Although a lazy binding or a strict pure binding is not work-free, I think it is unlikely that floating them inwards leads to unbounded cost increase.

Yeah, I think this would also be reasonable. We just need an `isWorkFreeApartFromOverhead` predicate and a clear explanation. But it does mean that there probably will be _some_ cases where things get worse.

zliu41: > We just need an isWorkFreeApartFromOverhead predicate and a clear explanation

That's just `isPure`, I think, which means it's the same as what I'm currently doing, or are you thinking about something else?

michaelpj: `isPure` is a property of terms, `isWorkFreeWithoutOverhead` is a property of _bindings_, since the overhead comes from the binding. I just think it's good to have explicit names for the concepts we're using, even if they're ultimately implemented very similarly.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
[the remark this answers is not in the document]
zliu41: Added a TODO.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
zliu41: > we could also test that independently!

True. I just doubt that having a first pass makes the second pass much simpler. During the second pass the free variables will need to be maintained when we push a binding down the tree, and that doesn't sound much easier than calculating the free variables from scratch!

zliu41 (commented):
@michaelpj Updated.

- Re-did the test cases. See `plutus-core/plutus-ir/test/transform/letFloatIn`.
- Added two Notes which I believe address all the questions and concerns we've had so far. @Marty I think you also had some questions before about why float-in is helpful, so please take a look.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,329 @@
...
+as `force`. The number of `Force` steps are the same with or without floating in,
+which follows from the fact that `rhs_x` is always evaluated the same number of times
+however we do the floating.
+
+==== Floating a binding into the argument of an application ====
+
+Since floating a binding into the RHS of another binding can lead to increased costs,
+it follows that floating a binding into the argument of an application can as well,
+because `let x (nonstrict) = rhs in body` is the same as `(\x -> force body) (delay rhs)`,
+and `let x (strict) = rhs in body` is the same as `(\x -> body) rhs`.
```
michaelpj, line 130: Did I say that floating into arguments was bad? I did, but not in my last comment... I think it's actually _fine_, because of the strictness.

In `[ (\x. .. x .. x ..) t ]` t gets evaluated _once_, even though the resulting value is used multiple times. The problem only comes if we are in a context where we get used multiple times _and_ we've got delayed evaluation. So I continue to think (as I said here: https://github.com/input-output-hk/plutus/pull/5032#discussion_r1065703355) that what really matters is _which constructs delay evaluation_. And that's lambdas, type abstractions, and non-strict let-binding RHSs.

Which is to say: if we do the right thing there, then floating into arguments is fine.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,329 @@
...
+Moving a lazy binding obviously doesn't change the semantics of the program, since the RHS of
+a lazy binding is always evaluated whenever needed, and only evaluated when needed, regardless
+of where the binding is placed. Lazy bindings therefore can be floated inwards.
+
+Why, one might ask, does floating lazy bindings inwards save cost, if the RHS is always
+evaluated the same number of times? Because when we move a lazy binding `let x (nonstrict) = rhs`,
+what we are really moving around is `delay rhs` and lambda applications, and neither
+`delay rhs` nor lambda application is free - they incur additional `Delay`, `LamAbs`
+and `Apply` steps in the CEK machine. Thus cost can be saved by floating bindings into
+unevaluated branches.
```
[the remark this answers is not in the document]
michaelpj: Coming back to this... we don't move strict non-pure let bindings, so that suggests that we _only_ move bindings such that the difference comes from the overhead cost. So... _all_ the cost savings must be coming from these "trivial" reductions! Which is a) really surprising and b) makes the argument that we don't need to worry about repeating them more questionable.

Or have I got this wrong :thinking:

michaelpj (commented):
I think maybe we should have a call, I feel a bit like I'm going around in circles and I think maybe I'm saying contradictory things. I do think we have at least one disagreement: I'm _somewhat_ sure that the right thing to do is to focus on which things delay evaluation and then not float into those.

zliu41:
> I think maybe we should have a call

Yeah, let's do that given the number of back and forths we've had. I disagree that there is any fundamental difference between non-strict and strict bindings, but many of your comments seem to revolve around that argument.

Also let me address what seems to be one of your main concerns:

> So... all the cost savings must be coming from these "trivial" reductions! Which is a) really surprising and b) makes the argument that we don't need to worry about repeating them more questionable.

When floating a binding into an unevaluated branch, it's pure cost reduction. If this is done many times, then the cost saving can add up. Indeed there's one test case that demonstrates this.

On the other hand, when repeat them, every time we do so, there's almost certainly some "real work" to be done at the same time. Imagine a lambda that is applied 100 times. When we move the binding to the arguments, we are repeating the overhead 100 times. But each time the lambda is applied, there's almost some "real work" as well, and there's also 100x "real work". I know it sounds quite un-rigorous but that's the idea.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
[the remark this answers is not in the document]
L-as: let-bindings are much more expensive than they need to be AFAICT

zliu41:
@michaelpj Updated. This version of the implementation works as follows (this is also explained in the Note):

We recurses into the top-level `Term` using the following context type:

```haskell
data FloatInContext = FloatInContext
    { _ctxtInManyOccRhs :: Bool
    , _ctxtUsages       :: Usages.Usages
    }
```

`ctxtUsages` is the usage counts of variables, and `ctxtInManyOccRhs` is initially `False`.
`ctxtInManyOccRhs` is set to `True` if we descend into:

1. The RHS of a binding whose LHS is used more than once
2. The argument of an application, unless the function is a LamAbs whose bound variable
is used at most once, or a Builtin.

The value of `ctxtInManyOccRhs` is used as follows:

1. When `ctxtInManyOccRhs = False`, we avoid descending into the RHS of a non-strict binding
whose LHS is used more than once, and we descend in all other cases;
2. When `ctxtInManyOccRhs = True`, we additionally avoid descending into `LamAbs` or `TyAbs`.

Also rewrote the Note and updated the test cases.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,406 @@
...
+            TyInst (a, usBody) tyInstBody ty ->
+                -- A term binding can always be placed inside the body a `TyInst` because the
+                -- type cannot mention the bound variable
+                TyInst (a, Set.union usBind usBody) <$> go b tyInstBody <*> pure ty
+            Let (a, usBody) NonRec bs letBody
+                -- The binding can be placed inside a `Let`, if the right hand sides of the
+                -- bindings of the `Let` do not mention `var`.
+                | (var ^. PLC.unique) `Set.notMember` foldMap bindingUniqs bs ->
+                    Let (a, Set.union usBind usBody) NonRec bs <$> go b letBody
+                | (var ^. PLC.unique) `Set.notMember` termUniqs letBody -> do
```
michaelpj, line 329: redundant do?

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,406 @@
...
+        FloatInContext
+        ( Maybe
+            ( [Binding tyname name uni fun (a, Uniques)]
+            , Binding tyname name uni fun (a, Uniques)
+            , [Binding tyname name uni fun (a, Uniques)]
+            , Bool
+            -- \^ Whether the chosen binding occurs more than once.
+            )
+        )
+splitBindings u bs = case is of
```
michaelpj, line 381: This function merges two things:
1. Identifying the sole binding that uses the variable
2. Determining whether that binding is safe to float into

I think it would be clearer if we just did 1 and then we did 2 at the call site. We already handle the rest of the floating safety stuff in `go`.

Also as it is the Haddock is misleading.

plutus-core/plutus-ir/test/transform/letFloatIn/avoid-floating-into-RHS
```diff
@@ -0,0 +1,22 @@
+(let
+  (nonrec)
+  (termbind
+    (strict)
+    (vardecl addInteger (fun (con integer) (fun (con integer) (con integer))))
+    (builtin addInteger)
+  )
+  (termbind
```
michaelpj, line 8: We might as well simplify these, even if they're generated. I'm still reading them! You can just delete both of these bindings and put `(builtin addInteger)` everywhere.

plutus-core/plutus-ir/test/transform/letFloatIn/avoid-floating-into-RHS.golden
```diff
@@ -0,0 +1,25 @@
+(let
+  (nonrec)
+  (termbind
+    (strict)
+    (vardecl addInteger (fun (con integer) (fun (con integer) (con integer))))
+    (let
```
michaelpj, line 6: Unless you want to demonstrate _this_ floating, but I don't think that you do.

plutus-core/plutus-ir/test/transform/letFloatIn/avoid-floating-into-RHS
michaelpj: Same in the other tests

michaelpj (approved):
_Great_ tests!

plutus-core/plutus-ir/test/transform/letFloatIn/avoid-floating-into-RHS.golden
zliu41: Yeah, this was generated from PlutusTx, and can be simplified manually

plutus-benchmark/plutus-benchmark.cabal
```diff
@@ -41,7 +41,7 @@ common lang
 
   ghc-options:
     -fno-specialise -fno-spec-constr -fno-ignore-interface-pragmas
-    -fno-omit-interface-pragmas
+    -fno-omit-interface-pragmas -fforce-recomp
```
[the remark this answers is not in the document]
zliu41: I accidentally committed this

plutus-core/changelog.d/20230105_135411_unsafeFixIO_floatin.md
```diff
@@ -0,0 +1,42 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+
+### Added
+
+- A float-in compilation pass.
```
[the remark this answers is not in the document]
zliu41: Done.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
```diff
@@ -0,0 +1,406 @@
...
+        goBinding = \case
+            TermBind a s var rhs ->
+                let rhs' = go rhs
+                 in TermBind (a, termUniqs rhs') s (noUniq var) rhs'
+            other -> noUniq other
+
+termUniqs :: Term tyname name uni fun (a, Uniques) -> Uniques
+termUniqs = snd . termAnn
+
+-- | The result includes both the `Unique` of the bound var and the `Unique`s in the RHS.
```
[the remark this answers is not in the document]
zliu41: `Uniques` currently contains all variables, not just free variables, so for consistency reason I'm including the bound variables. Arguably we should calculate only the set of free variables since it could be much smaller.

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
zliu41: Good point!

plutus-core/plutus-ir/src/PlutusIR/Transform/LetFloatIn.hs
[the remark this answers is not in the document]
zliu41: Updated the comment and also added a comment on `type Uniques = ...`

PR #5033, opened 2023-01-03
PLT-787-follow-up-replacing first version of diagrams with upgraded version

joseph-fajen:
The only change with this follow-up for PLT-787 is that the three diagrams have been replaced with versions produced by a graphic designer from the content team.


zliu41 (approved):

michaelpj:
This is one of the reasons I like having generated diagrams. Every time we do this we bloat the repository size by a few hundred kilobytes.

joseph-fajen:
Interesting point. I'm most concerned with a good reader experience and ease of comprehension, so I like this more designed approach to diagrams generally. But if code generated diagrams are more effective for the reader, or if the extra kilobytes are too great a cost, we can reconsider our approach. Perhaps I can solicit feedback from a wider audience.

michaelpj:
At least we should be checking that the PNGs are optimized for size, which I think we are doing automatically.

PR #5035, opened 2023-01-05
PLT-1335 Organize inliner into a directory with separate modules

thealmarty:
We are going to add a few more optimizations to the inliner. There will be substantial code added so it’s not desirable to put it all in `Inline.hs`, which is already over 500 lines long. 

I started an inline directory, and a `CallSiteInline` module for call site inlining. I also started `Heuristics.hs`. 

The `CallSiteInline` module is incomplete here - this PR focuses on re-organizing the modules.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


zliu41:
Could you elaborate a bit more the reasoning behind the new module structure? It's not very intuitive to me because
- The inliner does two kinds of inlining: unconditional inlining, and callsite inlining. If there's a `CallSiteInline` module, then I'd expect to see an `UnconditionalInline` module. Is the latter too simple to be put in a separate module?
- The `CallSiteInline` module contains a `countLam` function. But that function is basically heuristics. It's no different than `effectSafe` or other things in the `Heuristics` module. So why not put `countLam` there also?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Heuristics.hs
```diff
@@ -0,0 +1,253 @@
...
+After that it gets more difficult. As soon as we're inlining things that are not variable-sized
+and are used more than once, we are at risk of doing more work or making things bigger.
+
+There are a few things we could do to do this in a more principled way, such as call-site inlining
+based on whether a function is fully applied.
+-}
+
+-- | Substitution range, 'SubstRng' in the paper but no 'Susp' case.
+-- See Note [Inlining approach and 'Secrets of the GHC Inliner']
+newtype InlineTerm tyname name uni fun a =
```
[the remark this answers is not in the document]
michaelpj: `Types.hs` is traditional

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,38 @@
...
+-}
+
+module PlutusIR.Transform.Inline.CallSiteInline where
+
+import PlutusIR
+import PlutusIR.Transform.Rename ()
+import PlutusPrelude
+
+-- | Count the number of type and term lambdas in the RHS of a binding
+countLam ::
```
michaelpj, line 18: Many of these functions are generic functions that don't necessarily uniquely relate to inlining, so if we're going to move them I think we might as well put them in some more generic utility module.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Heuristics.hs
```diff
@@ -0,0 +1,253 @@
...
+acceptable t =
+    -- See Note [Inlining criteria]
+    pure $ costIsAcceptable t && sizeIsAcceptable t
+
+{- |
+Try to identify the first sub term which will be evaluated in the given term and
+which could have an effect. 'Nothing' indicates that we don't know, this function
+is conservative.
+-}
+firstEffectfulTerm :: Term tyname name uni fun a -> Maybe (Term tyname name uni fun a)
```
michaelpj, line 160: This is a generic utility.

michaelpj (commented):
Personally I don't think `Inline.hs` is too big, but this seems okay. I also don't find `CallSiteInline.hs` compelling, I think we should just move the generic utility functions to a generic utility module, not sure where.

thealmarty:
>     * The inliner does two kinds of inlining: unconditional inlining, and callsite inlining. If there's a `CallSiteInline` module, then I'd expect to see an `UnconditionalInline` module. Is the latter too simple to be put in a separate module?

I think right now `Inline.hs` is `UnconditionalInine`. So yeah I can rename that. I like that.
> 
>     * The `CallSiteInline` module contains a `countLam` function. But that function is basically heuristics. It's no different than `effectSafe` or other things in the `Heuristics` module. So why not put `countLam` there also?

Yes, I thought that might be better too. But then I thought it's heuristics for callsite inlining so I left it there. With Michael's comment I now think maybe I just need a `Type.hs` module separated from the inline modules?

thealmarty:
@zliu41 @michaelpj how about this:

`UnconditionalInline.hs` for for current inlining function (what's in `Inline.hs` currently in this PR)
`Utility.hs` for the generic utility functions and types (what's in `Heuristics.hs` currently in this PR) 
`CallSiteInline.hs` for call site inlining functions including the count lambda function and anything I will add in the next PR that does call site inlining.

michaelpj:
> It will be too big. As I mentioned I'll have substantial notes in there too. I think it's cleaner and easier to find if we put all callsite inline stuff in a module.

I think this is an example where it _doesn't_ make so much sense to split things up. It's hard for us to review things where the motivation for why the thing makes sense isn't visible. At that point it's probably better to do them together.

thealmarty:
> I think this is an example where it _doesn't_ make so much sense to split things up. It's hard for us to review things where the motivation for why the thing makes sense isn't visible. At that point it's probably better to do them together.

No problem, I'll add the notes etc to this too. I'll ping again when it's ready for a re-review.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Heuristics.hs
thealmarty: Moved to `Purity` module. It's kinda related...

zliu41:
Yeah I agree with Michael. IMO implementing the inlining of fully saturated function first, and then re-organize the modules would be the correct order. Not only does it make reviewing easier, but it also makes sense to tackle the most difficult part of the task first.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
michaelpj: I think that's "too small" IMO. It would be a bit different if it was something that was a bit bigger or of obviously utility that we're definitely going to use elsewhere, but this is so tiny and specific I think it's not worth it.

thealmarty:
@michaelpj @zliu41 does this sound ok:

I've re-org the dir and I'll put some notes in the `CallSiteInline` module WITHOUT the `countLam` function. After this is merged, I'll fill in the rest of the `CallSiteInline` module in the next PR.

So this PR I want to focus on the new module structures and the notes, which I suspect will prompt some further discussions.

It's just that when I organize the modules everything will be read as new so I want that to be a separate PR for readability.

If it's really preferred I can just add the notes to `Inline` module and come back to this PR. I don't prefer it because 

- this makes `Inline` big and messy in the interim
- I've already done the org work here so it'll be less work for me to not need to go back and forth

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,123 @@
...
+We consider cases when `v` is a function/lambda abstraction(s). I.e.:
+
+let v :: type = \x1.\x2...\xn.VBody in body
+
+In the `body`, where `v` is *called*,
+if it was given `n` arguments, then it is fully applied in the `body`.
+We inline the *call site* of `v` in this case, i.e.,
+we replace `v x_1 ... x_n` in the `body` with `rhs`. E.g.
+
+let f :: TyLam (TODO)= \x.\y -> x*y
```
zliu41, line 29: Did you mean to put a TODO here?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,123 @@
...
+    y :: TyLam = f q
+in f a b
+
+becomes `let y = f q in a*b`. What about
```
zliu41, line 33: 1. You still need the `let f :: ...` otherwise `f` is undefined here
2. You may want to replace `a * b` with something smaller, say `a`. Because `a * b` and `f a b` have the same size, so it is unclear why we want to turn `f a b` into `a * b`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,123 @@
...
+
+let f = \x.\y -> x*y
+    y = f q
+in f a b + y
+
+Here we have a `f` that is not fully applied via `y`, so we cannot eliminate the let-binding of `f`.
```
zliu41, line 39: I don't know what "f is not fully applied via y" means. It seems to me `f` is fully applied here, just like the previous example.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,123 @@
...
+```haskell
+let f = \x . x
+     g = \y . f
+in g a
+```
+
+`f` and `g` each has 1 lambda. However, `g`'s _body_ includes `f` which also has a lambda.
+Since we only count the number of lambdas, `g` is fully applied, and we inline.
+`g a` reduces to `f`, which reduces the amount of code even though `f` has a lambda that
+is not applied.
```
zliu41, line 92: The "even though `f` has a lambda that is not applied" part doesn't make sense to me. `f` is not fully applied, and we don't inline `f`, and we shouldn't inline `f`. I don't see a problem here.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,123 @@
...
+in id {int -> int} (\x -> x +1)
+```
+
+Here we have a type application for a function that takes one type variable.
+I.e., it's fully applied in terms of type.
+In terms of terms, `id` takes one argument, and is indeed applied to one.
+So `id` is fully applied! And we inline it.
+
+Reducing the _body_ `id {int -> int} (\x -> x +1)` leaves us `(\x -> x +1)`,
+a function with 1 lambda. Even though it's not fully applied, inlining
```
zliu41, line 108: Similarly, the "even though" doesn't make sense to me. `id` is fully applied - it is applied to a type arg and a term arg. And we inline it. And we should inline it.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,123 @@
...
+#### Example 3: function application in _RHS_
+
+```haskell
+let f = (\x.\y.x+y) 4
+in f 5
+```
+
+With beta-reduction, `f` becomes `\y.4+y` and it has 1 lambda.
+The _body_ `f 5` is a fully applied function!
+We can reduce it to 9.
```
zliu41, line 120: The compiler can't reduce it to 9. The compiler can only reduce it to 4+5.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,123 @@
...
+However, because the _rhs_ `(\x.\y.x+y) 4` is a function application, not a lambda abstraction,
+we won't inline it with the current implementation. This is something we may be able to improve.
```
zliu41, line 122: We will inline it if we run beta-reduction first, before the inliner. Even if we run inliner first, as long as the simplifier runs two or more passes, it will still get inlined.

zliu41 (approved):
I think it's fine to merge this. The main Note will need some work. But we can worry about it once we are done with the implementation, at which point it probably will need to be updated anyway.

thealmarty:
> I think it's fine to merge this. The main Note will need some work. But we can worry about it once we are done with the implementation, at which point it probably will need to be updated anyway.

Thanks! Yes I was going to polish the note more before merging but I agree, I will be editing the note as I do the implementation anyway so yeah we don't need to worry about it much.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,115 @@
...
+Call site inlining machinery. For now there's only one part: inlining of fully applied functions.
+See ADR TODO for motivation. We inline fully applied functions if the cost and size are acceptable.
+See note [Inlining of fully applied functions].
+-}
+
+module PlutusIR.Transform.Inline.CallSiteInline where
+
+{- Note [Inlining of fully applied functions]
+
+We inline the call site when a function is fully applied AND the cost and size are acceptable.
```
michaelpj, line 16: cost and size of _what_? The function? the body of the function? be specific!

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,115 @@
...
+and count the number of lambdas.
+Then, in the _body_, we check for term or type application (`Apply` or `TyInst`) of _v_.
+
+If _v_ is fully applied in the body, i.e., if
+
+1. the number of type lambdas equals the number of type arguments applied, and
+2. the number of term lambdas equals the number of term arguments applied, and
+
+if other call site inlining conditions are satisfied,
+(we currently inline too little, this will be improved later, see below)
```
michaelpj, line 71: irrelevant to this note iMO

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,115 @@
...
+
+we inline _v_, i.e., replace its occurrence with _rhs_ in the _body_.
+For the rest of the discussion here we focus on the conditions 1 and 2 only.
+
+Below are some examples that involve a _body_ that is not fully reducible/applied but
```
michaelpj, line 76: What does this mean? I still want a specific discussion of the various concepts of reducible/saturated, I think it's _critical_.

michaelpj (commented):
Sorry, now you added the note so I have comments on the note :sweat_smile:

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,115 @@
...
+Consider `let v :: ty = rhs in body`, in which `body` calls `v`.
+
+We consider cases when `v` is a function/lambda abstraction(s). I.e.:
+
+let v :: type = \x1.\x2...\xn.VBody in body
+
+In the `body`, where `v` is *called*,
+if it was given `n` arguments, then it is fully applied in the `body`.
+We inline the *call site* of the fully applied `v` in this case, i.e.,
+we replace `v x_1 ... x_n` in the `body` with `rhs`. E.g.
```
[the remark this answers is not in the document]
thealmarty: Sorry yes that's what I meant.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: clarified in the bottom.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -0,0 +1,115 @@
...
+so we know how many term/type arguments they have.
+
+We pattern match the _rhs_ with `LamAbs` or `TyAbs` (lambda abstraction for terms or types),
+and count the number of lambdas.
+Then, in the _body_, we check for term or type application (`Apply` or `TyInst`) of _v_.
+
+If _v_ is fully applied in the body, i.e., if
+
+1. the number of type lambdas equals the number of type arguments applied, and
+2. the number of term lambdas equals the number of term arguments applied, and
```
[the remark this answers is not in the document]
thealmarty: I've been trying to avoid the mentioning of "arity". I'll come back to this after the implementation.

thealmarty:
Thanks @michaelpj I've made some improvements and as mentioned I'll be revising as I implement so you can review again with the implementation in the next PR.

PR #5037, opened 2023-01-11
Generalize CEK input-term annotations

bezirg:
I would like to have `ann` available in CEK's Term, because then I can share more between the canonical cek and the debuggable-steppable cek.

Originally, I put `ann` everywhere both in the `Term`s that come in the CEK and the `Term`s that come out of the CEK.
@effectfully pointed to me that the out-Terms should remain as-is, i.e. with `ann`==`()` . This implies that the dischargeCekValue/Env functions have to be constrained to `()` and that had me put some `void`s over there. I don't like these `void`s there but discharge should suppose to be lazy anyway.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetingDefaults.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetingDefaults.hs
@@ -74,10 +74,10 @@ defaultCekCostModel = CostModel defaultCekMachineCosts defaultBuiltinCostModel
 defaultCostModelParams :: Maybe CostModelParams
 defaultCostModelParams = extractCostModelParams defaultCekCostModel
 
-defaultCekParameters :: MachineParameters CekMachineCosts CekValue DefaultUni DefaultFun
+defaultCekParameters :: Typeable ann => MachineParameters CekMachineCosts CekValue DefaultUni DefaultFun ann
 defaultCekParameters = mkMachineParameters def defaultCekCostModel
 
-unitCekParameters :: MachineParameters CekMachineCosts CekValue DefaultUni DefaultFun
+unitCekParameters :: Typeable ann => MachineParameters CekMachineCosts CekValue DefaultUni DefaultFun ann
 unitCekParameters =
     mkMachineParameters def $
         CostModel unitCekMachineCosts unitCostBuiltinCostModel
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
@@ -182,15 +182,15 @@ but functions are not printable and hence we provide a dummy instance.
 -}
 
 -- See Note [Show instance for BuiltinRuntime].
-instance Show (BuiltinRuntime (CekValue uni fun)) where
+instance Show (BuiltinRuntime (CekValue uni fun ann)) where
     show _ = "<builtin_runtime>"
 
 -- 'Values' for the modified CEK machine.
-data CekValue uni fun =
+data CekValue uni fun ann =
     -- This bang gave us a 1-2% speed-up at the time of writing.
     VCon !(Some (ValueOf uni))
-  | VDelay !(Term NamedDeBruijn uni fun ()) !(CekValEnv uni fun)
-  | VLamAbs !NamedDeBruijn !(Term NamedDeBruijn uni fun ()) !(CekValEnv uni fun)
+  | VDelay !(Term NamedDeBruijn uni fun ann) !(CekValEnv uni fun ann)
+  | VLamAbs !NamedDeBruijn !(Term NamedDeBruijn uni fun ann) !(CekValEnv uni fun ann)
     -- | A partial builtin application, accumulating arguments for eventual full application.
     -- We don't need a 'CekValEnv' here unlike in the other constructors, because 'VBuiltin'
     -- values always store their corresponding 'Term's fully discharged, see the comments at
@@ -209,12 +209,12 @@ data CekValue uni fun =
       -- be returned in the result. The laziness is important, because the arguments are discharged
       -- values and discharging is expensive, so we don't want to do it unless we really have
       -- to. Making this field strict resulted in a 3-4.5% slowdown at the time of writing.
-      !(BuiltinRuntime (CekValue uni fun))
+      !(BuiltinRuntime (CekValue uni fun ann))
       -- ^ The partial application and its costing function.
       -- Check the docs of 'BuiltinRuntime' for details.
     deriving stock (Show)
 
-type CekValEnv uni fun = Env.RAList (CekValue uni fun)
+type CekValEnv uni fun ann = Env.RAList (CekValue uni fun ann)
 
 -- | The CEK machine is parameterized over a @spendBudget@ function. This makes the budgeting machinery extensible
 -- and allows us to separate budgeting logic from evaluation logic and avoid branching on the union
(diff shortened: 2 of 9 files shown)


plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetingDefaults.hs
```diff
@@ -74,10 +74,10 @@ defaultCekCostModel = CostModel defaultCekMachineCosts defaultBuiltinCostModel
 defaultCostModelParams :: Maybe CostModelParams
 defaultCostModelParams = extractCostModelParams defaultCekCostModel
 
-defaultCekParameters :: MachineParameters CekMachineCosts CekValue DefaultUni DefaultFun
+defaultCekParameters :: Typeable ann => MachineParameters CekMachineCosts CekValue DefaultUni DefaultFun ann
```
bezirg, line 77: Here I could constrain ann==`()` but leaving it generalized may help me reuse these for the debugger

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -490,28 +490,28 @@ dischargeCekValEnv valEnv = go 0
...
     VCon     val                         -> Constant () val
-    VDelay   body env                    -> dischargeCekValEnv env $ Delay () body
+    VDelay   body env                    -> dischargeCekValEnv env $ Delay () (void body)
     -- 'computeCek' turns @LamAbs _ name body@ into @VLamAbs name body env@ where @env@ is an
     -- argument of 'computeCek' and hence we need to start discharging outside of the reassembled
     -- lambda, otherwise @name@ could clash with the names that we have in @env@.
     VLamAbs (NamedDeBruijn n _ix) body env ->
         -- The index on the binder is meaningless, we put `0` by convention, see 'Binder'.
-        dischargeCekValEnv env $ LamAbs () (NamedDeBruijn n deBruijnInitIndex) body
+        dischargeCekValEnv env $ LamAbs () (NamedDeBruijn n deBruijnInitIndex) (void body)
```
bezirg, line 502: Enter the void

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -644,7 +644,7 @@ enterComputeCek = computeCek (toWordArray 0) where
         !unbudgetedSteps' <- stepAndMaybeSpend BBuiltin unbudgetedSteps
         meaning <- lookupBuiltin bn ?cekRuntime
         -- The @term@ is a 'Builtin', so it's fully discharged.
-        returnCek unbudgetedSteps' ctx (VBuiltin bn term meaning)
+        returnCek unbudgetedSteps' ctx (VBuiltin bn (void term) meaning)
```
bezirg, line 647: Also, I am not very sure about the implications of this `void`. Probably it is fine

bezirg:
I keep this as draft right now, because I have to write comments and also benchmark it

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -640,11 +640,11 @@ enterComputeCek = computeCek (toWordArray 0) where
...
     -- s ; ρ ▻ con c  ↦  s ◅ con c
     -- s ; ρ ▻ builtin bn  ↦  s ◅ builtin bn arity arity [] [] ρ
-    computeCek !unbudgetedSteps !ctx !_ term@(Builtin _ bn) = do
+    computeCek !unbudgetedSteps !ctx !_ (Builtin _ bn) = do
         !unbudgetedSteps' <- stepAndMaybeSpend BBuiltin unbudgetedSteps
         meaning <- lookupBuiltin bn ?cekRuntime
-        -- The @term@ is a 'Builtin', so it's fully discharged.
-        returnCek unbudgetedSteps' ctx (VBuiltin bn term meaning)
+        -- The term is a 'Builtin', so it's fully discharged.
+        returnCek unbudgetedSteps' ctx (VBuiltin bn (Builtin () bn) meaning)
```
michaelpj, line 647: why not generalise ann here also?

michaelpj (approved):
looks plausible, modulo benchmarks

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: That thing goes out from the machine (during discharging), which is why it has to be `()` there, if we agree we don't need to return annotated terms, which would be problematic, because we'd need to annotate each result of a builtin application and it's not particularly clear how to synthesize an annotation in that case.

michaelpj: We can put a `void` in before we throw it as we do in `discharge`, no? It's also potentially even performance relevant: this forces us to take apart and reconstruct a value instead of just passing through the program node.

effectfully: Hm, we could I guess, but we'd need to propagate annotations really far. So putting `ann` there would require us to also find an annotation in

```haskell
    applyEvaluate !unbudgetedSteps !ctx (VBuiltin fun term runtime) arg = do
        let argTerm = dischargeCekValue arg
            -- @term@ and @argTerm@ are fully discharged, and so @term'@ is, hence we can put it
            -- in a 'VBuiltin'.
            term' = Apply () term argTerm
```

which will require us to make `applyEvaluate` to take an annotation, which will require us to put an annotation into `FrameApplyFun`, which will require an annotation in `FrameApplyArg`.

All for removing them in the end anyway. So I'd rather not.

effectfully: ... and the same for

```haskell
    forceEvaluate !unbudgetedSteps !ctx (VBuiltin fun term runtime) = do
        -- @term@ is fully discharged, and so @term'@ is, hence we can put it in a 'VBuiltin'.
        let term' = Force () term
```

effectfully: It's not on the very-hot evaluation path, because that field of `VBuiltin` is lazy. Which is why all that `Apply ()` and `Force ()` is fine (with the former being particularly expensive, since it discharges the argument, but we don't care since the field is lazy).

Plus, `void term` would reconstruct the term anyway, it's not that `void` magically changes the pointer to the annotation or something.

michaelpj: Ah yes, good point about the laziness. I thought we were relying on laziness for the construction of the `Apply` stuff and so on, but yeah the `VBuiltin` field is also lazy. Unclear why it helps that it is, though, given that I'd expect the thing we put into it to be evaluated...

> Plus, void term would reconstruct the term anyway, it's not that void magically changes the pointer to the annotation or something.

I was unclear: I meant, leave the case of `computeCek` the same, and use `void` in `forceEvaluate`, because that code is only called on the error path.

effectfully: >  I thought we were relying on laziness for the construction of the `Apply` stuff and so on, but yeah the `VBuiltin` field is also lazy. 

It's the same field, `Apply ()` happens within this exact field of `VBuiltin`.

> I was unclear: I meant, leave the case of `computeCek` the same, and use `void` in `forceEvaluate`, because that code is only called on the error path.

Ah, I see.

bezirg:
No significant change when manual benchmarking

michaelpj (approved):

PR #5041, opened 2023-01-13
Bump stylish-haskell and use new import grouping feature

michaelpj:
- Bump stylish-haskell to 0.14.4
- Add grouping to stylish-haskell config
- Apply changes

EDIT: hmm, my PR body got lost. This bumps our `stylish-haskell` version so we can group imports in the way we've often discussed doing, but automatically. The rule mechanism is fairly flexible, so we can do various things quite easily. I think what I've put here is decent, namely:
- Group everything with "Plutus" in the first component
- Within that, group on the first component
- Everything else goes together

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Result.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Result.hs
@@ -16,9 +16,9 @@ module PlutusCore.Evaluation.Result
     , isEvaluationFailure
     ) where
 
-import PlutusPrelude
-
 import PlutusCore.Pretty
+
+import PlutusPrelude
 
 import Control.Lens
 import Control.Monad.Except
--- a/plutus-core/plutus-core/src/PlutusCore/Subst.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Subst.hs
@@ -24,12 +24,13 @@ module PlutusCore.Subst
     , tvTy
     ) where
 
-import PlutusPrelude
-
 import PlutusCore.Core
+
+import PlutusPrelude
 
 import Control.Lens
 import Control.Lens.Unsound qualified as Unsound
+
 import Data.Set as Set
 
 purely :: ((a -> Identity b) -> c -> Identity d) -> (a -> b) -> c -> d
--- a/plutus-core/plutus-core/src/PlutusCore/TypeCheck/Internal.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/TypeCheck/Internal.hs
@@ -27,15 +27,16 @@ import PlutusCore.Normalize.Internal (MonadNormalizeType)
 import PlutusCore.Normalize.Internal qualified as Norm
 import PlutusCore.Quote
 import PlutusCore.Rename
+
 import PlutusPrelude
 
 import Control.Lens
 import Control.Monad.Error.Lens
 import Control.Monad.Except
--- Using @transformers@ rather than @mtl@, because the former doesn't impose the 'Monad' constraint
--- on 'local'.
 import Control.Monad.Trans.Reader
+
 import Data.Array
+
 import Universe
 
 {- Note [Global uniqueness]
--- a/plutus-core/plutus-core/test/Names/Spec.hs
+++ b/plutus-core/plutus-core/test/Names/Spec.hs
@@ -7,21 +7,21 @@
 
 module Names.Spec where
 
-import PlutusCore.Test
-
 import PlutusCore
 import PlutusCore.DeBruijn
-import PlutusCore.Mark
-import PlutusCore.Pretty
-import PlutusCore.Rename.Internal
-
 import PlutusCore.Generators.Hedgehog
 import PlutusCore.Generators.Hedgehog.AST as AST
 import PlutusCore.Generators.Hedgehog.Interesting
+import PlutusCore.Mark
+import PlutusCore.Pretty
+import PlutusCore.Rename.Internal
+import PlutusCore.Test
 
 import GHC.Exts (fromString)
+
 import Hedgehog hiding (Var)
 import Hedgehog.Gen qualified as Gen
+
 import Test.Tasty
 import Test.Tasty.Hedgehog
 import Test.Tasty.HUnit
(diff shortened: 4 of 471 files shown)


thealmarty (approved):
So tidy! I like it! Maybe worth a change log?

michaelpj:
> Maybe worth a change log?

It's not user-facing?

plutus-core/plutus-core/src/PlutusCore/Subst.hs
```diff
@@ -24,10 +24,10 @@ module PlutusCore.Subst
     , tvTy
     ) where
 
-import PlutusPrelude
-
 import PlutusCore.Core
 
+import PlutusPrelude
+
 import Control.Lens
```
effectfully, line 31: Meh.

plutus-core/plutus-core/test/Names/Spec.hs
```diff
@@ -7,17 +7,15 @@
...
 import PlutusCore
 import PlutusCore.DeBruijn
-import PlutusCore.Mark
-import PlutusCore.Pretty
-import PlutusCore.Rename.Internal
-
 import PlutusCore.Generators.Hedgehog
 import PlutusCore.Generators.Hedgehog.AST as AST
 import PlutusCore.Generators.Hedgehog.Interesting
+import PlutusCore.Mark
```
effectfully, line 15: No.

plutus-core/plutus-ir/src/PlutusIR/Parser.hs
```diff
@@ -14,18 +14,20 @@ module PlutusIR.Parser
...
 import PlutusIR.MkPir qualified as PIR
+
 import PlutusPrelude
-import Prelude hiding (fail)
 
 import Control.Monad.Combinators.NonEmpty qualified as NE
 import Control.Monad.Except (MonadError)
 import Data.Text (Text)
-import PlutusCore (MonadQuote)
-import PlutusCore.Error (AsParserErrorBundle)
```
effectfully, line 28: It clearly does some good stuff too.

plutus-core/plutus-ir/src/PlutusIR/Parser.hs
```diff
@@ -14,18 +14,20 @@ module PlutusIR.Parser
...
+import Prelude hiding (fail)
 import Text.Megaparsec hiding (ParseError, State, many, parse, some)
```
effectfully, line 31: But then.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -5,10 +5,11 @@
 
 module PlutusIR.Purity (isPure, firstEffectfulTerm) where
 
+import PlutusCore.Builtin
+
 import PlutusIR
```
effectfully, line 10: It also breaks our policy of most "distant" modules going last.

plutus-core/plutus-ir/test/GeneratorSpec.hs
```diff
@@ -5,7 +5,6 @@ import GeneratorSpec.Builtin
 import GeneratorSpec.Substitution
 import GeneratorSpec.Terms
 import GeneratorSpec.Types
-
 import Test.QuickCheck
```
effectfully, line 8: Why?

plutus-core/plutus-core/src/PlutusCore/Evaluation/Result.hs
```diff
@@ -16,10 +16,10 @@ module PlutusCore.Evaluation.Result
     , isEvaluationFailure
     ) where
 
-import PlutusPrelude
-
 import PlutusCore.Pretty
 
+import PlutusPrelude
+
```
effectfully, line 22: This is pretty much the only thing it does in modules I've written, i.e. only makes things worse. Hard for me to vote in favor of adopting it.

plutus-core/plutus-core/src/PlutusCore/TypeCheck/Internal.hs
```diff
@@ -27,13 +27,12 @@ import PlutusCore.Normalize.Internal (MonadNormalizeType)
...
 import PlutusCore.Rename
+
 import PlutusPrelude
 
 import Control.Lens
 import Control.Monad.Error.Lens
 import Control.Monad.Except
--- Using @transformers@ rather than @mtl@, because the former doesn't impose the 'Monad' constraint
--- on 'local'.
 import Control.Monad.Trans.Reader
```
effectfully, line 36: And deletes comments??? OK, that's enough.

effectfully (changes requested):
image

effectfully:
Maybe it's configurable? If it wasn't removing newlines, removing comments, reordering import groups, randomly assigning modules without prefixes to new groups...

michaelpj:
> Maybe it's configurable?

It's pretty configurable. You just provide it some regexes to use to group things, and it groups them. You can see the config I put in the `.hlint.yaml` file. If you can articulate what you want it to do I can see if I can make it do it! What I've put currently is my understanding of what we "usually" do.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
michaelpj: If we have such a policy I don't know it :sweat_smile: that's partly why I want to try and codfiy it!

michaelpj:
Looking at your comments, I think you want a rule like: if there are any `X.Y.Z` modules, then make a subgroup for the `X.Y` prefix, otherwise group it under `X`?

So you want

```
PlutusCore.X
PlutusCore.Y

PlutusCore.Z.P
PlutusCore.Z.Q
```
?

effectfully:
> Looking at your comments, I think you want a rule like: if there are any X.Y.Z modules, then make a subgroup for the X.Y prefix, otherwise group it under X?

Ideally yes, but not merging existing groups would also be (while allowing individual imports to switch groups).

michaelpj:
I don't think we can configure it to respect existing groups, and indeed part of the point is I'd like to not have to think about import grouping at all.

michaelpj:
Sadly many formatters have comment problems :( looks like this one is old: https://github.com/haskell/stylish-haskell/issues/39

Issue #4797, opened 2022-07-29
Add let binding to UPLC?


L-as:
### Describe the feature you'd like

Currently, in Plutarch (and I assume same for PlutusTx), we're emulating let bindings `let x = y; z` as `(\x -> z) y`, but this has the cost of introducing then eliminating a lambda.
If we had a `Let` constructor or similar, it's possible that it would be faster to interpret common scripts, if they were made to use native let-bindings.

michaelpj:
The whole point of PLC is to be very low level and not include such things. It sounds like you're gradually re-inventing PIR, which exists exactly to add types, (recursive) let-bindings, and datatypes :)

L-as:
The point is to increase performance of scripts.

michaelpj:
Why do you think it would be faster? In both cases most of the work is evaluating the RHS and putting it into the environment.

L-as:
Smaller AST.

michaelpj:
Here's my more detailed analysis.

First, what happens when we evaluate the two versions?

`[ (\x . t) a ]`
- Compute `[ ... ]`
- Compute `\x . t`
- Return `\x.t`
- Compute `a`
- Return `a`
- Apply

 `let x = a in t`
- Compute `let ...`
- Compute `a`
- Return `a`
- Apply

So indeed `let` would let us avoid computing and returning the lambda, because it's as if we statically know that we're applying a lambda and not an arbitrary term that needs evaluating.

So what does this save us? One compute-return step per let-binding, which corresponds to:
1. A pattern match
2. A couple of function calls
3. One step of costing
4. Allocating a lambda abstraction value

Of these we have evidence that 4 makes some difference from https://github.com/input-output-hk/plutus/pull/5042, but we can also avoid that by doing something like the optimization in that PR. The others are _very_ cheap except for 3, which might matter.

The other factor is that it depends how _hot_ (i.e. frequently evaluated) let-bindings are. My belief is that they are currently not that hot, see below for reasoning.

----

We can use the sums-of-products work as a convenient testbed for this. Sums-of-products introduces a bunch of immediately-applied lambdas whenever we evaluate a pattern match. This could instead be let-terms. Moreover, I implemented two versions of sums-of-products, one where case expressions bind variables, and one where they just apply lambdas. The former corresponds to having let-terms.

[...]

PR #5042, opened 2023-01-16
Don't allocate VLamAbs and VDelay if we don't need to

michaelpj:
I was thinking about why the non-binding version of products and sums is fast. One thing it avoids is creating values for each of the lambdas in the case matcher, i.e. we evaluate something like this:
```
(\x. (\y. (\z . t) ) ) a b c
```
by just stripping off the lambdas (naughtily) and shoving the arguments into the environment.

It occurred to me that we could get the same for normal lambdas in some cases. For that we need to do two things:
1. Evaluate the argument before the function, so we have the argument ready to go on the stack (this means switching to right-to-left evaluation order!)
2. When we call `computeCek` on a lambda, before returning, check whether there's an argument ready to go on the stack. If there is, just go straight into evaluating the body of the lambda without constructing a `VLamAbs`.

This is pretty gross. We're duplicating some code, and basically doing something like inlining a manually-specialized version of `returnCek` in a couple of places. I don't think we can massage GHC into doing this for us, though.

What do we get in return? By itself, not much (about 1-2%). That shouldn't be surprising, since this only works for lambdas where a literal lambda appears in appication. But these are statically beta-reducible! So there shouldn't be many in a sensible program.

I need to try this out in combination with the non-binding products branch, since that effectively introduces this case much more often.

So overall we probably shouldn't do it, unless we decide to change evaluation order anyway because right-to-left fits better with multilambdas.

### DIFF
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
@@ -81,6 +81,7 @@ import Data.Semigroup (stimes)
 import Data.Text (Text)
 import Data.Word
 import Data.Word64Array.Word8 hiding (Index)
+import GHC.Magic (inline)
 import Prettyprinter
 import Universe
 
@@ -625,14 +626,10 @@ enterComputeCek = computeCek (toWordArray 0) where
         returnCek unbudgetedSteps' ctx (VCon val)
     computeCek !unbudgetedSteps !ctx !env (LamAbs _ name body) = do
         !unbudgetedSteps' <- stepAndMaybeSpend BLamAbs unbudgetedSteps
-        case ctx of
-            FrameAwaitFun arg ctx' -> applyLam unbudgetedSteps ctx' env body arg
-            _                      -> returnCek unbudgetedSteps' ctx (VLamAbs name body env)
+        (inline returnCek) unbudgetedSteps' ctx (VLamAbs name body env)
     computeCek !unbudgetedSteps !ctx !env (Delay _ body) = do
         !unbudgetedSteps' <- stepAndMaybeSpend BDelay unbudgetedSteps
-        case ctx of
-            FrameForce ctx' -> forceDelay unbudgetedSteps ctx' env body
-            _               -> returnCek unbudgetedSteps' ctx (VDelay body env)
+        (inline returnCek) unbudgetedSteps' ctx (VDelay body env)
     -- s ; ρ ▻ lam x L  ↦  s ◅ lam x (L , ρ)
     computeCek !unbudgetedSteps !ctx !env (Force _ body) = do
         !unbudgetedSteps' <- stepAndMaybeSpend BForce unbudgetedSteps
@@ -682,19 +679,6 @@ enterComputeCek = computeCek (toWordArray 0) where
     returnCek !unbudgetedSteps (FrameAwaitFun arg ctx) fun =
         applyEvaluate unbudgetedSteps ctx fun arg
 
-
-    -- | Evaluate a delay that has been forced.
-    --
-    -- This just exists to share a little code.
-    forceDelay
-        :: WordArray
-        -> Context uni fun
-        -> CekValEnv uni fun
-        -> Term NamedDeBruijn uni fun ()
-        -> CekM uni fun s (Term NamedDeBruijn uni fun ())
-    forceDelay !unbudgetedSteps !ctx env body = computeCek unbudgetedSteps ctx env body
-    {-# INLINE forceDelay #-}
-
     -- | @force@ a term and proceed.
     -- If v is a delay then compute the body of v;
     -- if v is a builtin application then check that it's expecting a type argument,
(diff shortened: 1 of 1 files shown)


michaelpj:
Might be interesting to @effectfully

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -625,18 +625,22 @@ enterComputeCek = computeCek (toWordArray 0) where
...
-        returnCek unbudgetedSteps' ctx (VLamAbs name body env)
+        case ctx of
+            FrameAwaitFun arg ctx' -> applyLam unbudgetedSteps ctx' env body arg
+            _                      -> returnCek unbudgetedSteps' ctx (VLamAbs name body env)
     computeCek !unbudgetedSteps !ctx !env (Delay _ body) = do
         !unbudgetedSteps' <- stepAndMaybeSpend BDelay unbudgetedSteps
-        returnCek unbudgetedSteps' ctx (VDelay body env)
+        case ctx of
+            FrameForce ctx' -> forceDelay unbudgetedSteps ctx' env body
+            _               -> returnCek unbudgetedSteps' ctx (VDelay body env)
```
effectfully, line 635: The only other possible alternatives are `NoFrame` and `FrameAwaitArg`, right? Can we then handle those here as well and throw on `FrameApplyFun`, thus cutting one pattern match away across the board for every `VDelay` object?

Does some similar logic apply to `VLamAbs`?

It's probably indeed quite some duplication, but maybe we can unfold `returnCek` a bit by making it receive `returnCek` and `computeCek` and whatever else as additional arguments, so that `returnCek` isn't perceived as recursive by GHC and then tie the knot ourselves? It'll probably break join-point generation, though.

effectfully (commented):
Great idea.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
michaelpj: Hmm, does `inline` work on recursive functions? I wonder if I could just call `inline` at the uses of `returnCek`...

michaelpj: It works! And looking at the Core it does exactly what you asked for. I'll benchmark it.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -524,8 +524,8 @@ Morally, this is a stack of frames, but we use the "intrusive list" representati
 we can match on context and the top frame in a single, strict pattern match.
 -}
 data Context uni fun
-    = FrameApplyFun !(CekValue uni fun) !(Context uni fun)                         -- ^ @[V _]@
-    | FrameApplyArg !(CekValEnv uni fun) !(Term NamedDeBruijn uni fun ()) !(Context uni fun) -- ^ @[_ N]@
+    = FrameAwaitFun !(CekValue uni fun) !(Context uni fun)                         -- ^ @[V _]@
+    | FrameAwaitArg !(CekValEnv uni fun) !(Term NamedDeBruijn uni fun ()) !(Context uni fun) -- ^ @[_ N]@
```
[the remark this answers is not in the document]
michaelpj: I can never remember which one is which otherwise!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
michaelpj: Weirdly, this version is slightly worse???

effectfully: > It works!

You kidding me???

OK, after the initial shock I think I've found the answer: not all functions from a mutually recursive family are non-inlinable, only the [loop-breaker](https://wiki.haskell.org/Inlining_and_Specialisation#What_is_a_loop-breaker.3F) is.

> Weirdly, this version is slightly worse???

Maybe benchmarking is being weird?

michaelpj: Not sure, I ran both again and it was still worse? Maybe the first time was a fluke?

L-as:
> That shouldn't be surprising, since this only works for lambdas where a literal lambda appears in appication.

Isn't this just a let-binding? You could add a dedicated constructor as I noted in #4797 .

michaelpj:
> Isn't this just a let-binding?

Sure, that's one place it can occur.

effectfully:
@L-as makes a great point. @michaelpj have we considered multi-lets instead of multi-lambdas + multi-applications for making sums-of-products fast?

michaelpj:
I commented on the issue.

michaelpj:
To my surprise, it seems that GHC is clever enough to do this with `-O2`, I think it's `spec-constr` that does it.

michaelpj:
No wait, ignore me.

kwxm:
@michaelpj I'm going to close this since it's been open for nine months.

michaelpj:
Yep, this is a useful experiment to refer to, but it doesn't have to be open for that.

PR #5043, opened 2023-01-16
GitHub action to add "status: needs triage" label to new issues

kwxm:
This does what it says.  The YAML code is based on the example [here](https://docs.github.com/en/actions/managing-issues-and-pull-requests/adding-labels-to-issues).

### DIFF
--- a/.github/workflows/add-triage-label.yml
+++ b/.github/workflows/add-triage-label.yml
@@ -1,0 +1,21 @@
+name: Add triage label
+on:
+  issues:
+    types:
+      - reopened
+      - opened
+jobs:
+  add_triage_label:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


kwxm:
I'll have to merge this without review in order to test it.

kwxm:
I tested this in https://github.com/input-output-hk/plutus/issues/5044 and it appeared to work.  A quick retrospective review might be worthwhile though.

michaelpj:
The main alternative that I'm aware of is that you can add labels to issue templates so that they are created with those labels. The advantage is that you don't need an action, but the disadvantage is that if someone clicks "open a blank issue" it won't work.

thealmarty (commented):
Thanks!

PR #5045, opened 2023-01-16
Add source for new PLC specifcation and make it the main version (PLT-393)

kwxm:
This moves the (slightly tidied-up) source code for the new PLC specification into `plutus-core` and makes it the main version.   The nix code for building the specification may need to be updated: I'm trying to find out about that, and I'll make this a draft until I work out what needs to be done.


kwxm:
Oops, I accidentally requested a review from @cleverca22 .  Apologies for that: please ignore it!

michaelpj:
I'm a bit confused by the diff, git seems to think there are just new files for the old spec in the new location, without the old ones having gone away?

I suggest putting the old one in a sibling directory to the new one for simplicity, `plutus-core-spec-old` or something.

kwxm:
> I'm a bit confused by the diff, git seems to think there are just new files for the old spec in the new location, without the old ones having gone away?

I see what you mean.  I used `git mv` to move the old files into the subdirectory and then added the new files in the top directory. There are some files with the same name though, and it seems to have got confused by that. I'll close this and start again.

PR #5048, opened 2023-01-17
Kwxm/promote new specification 2

kwxm:
Another attempt at promoting the new version of the specification to be the main one and adding its source files.


kwxm:
I maybe still need to tell nix about new latex packages ...

michaelpj:
It's kind of weird, it's complaining about not recognizing the control sequence `\section`, I think, which is very odd... Lovely latex errors.

kwxm:
> It's kind of weird, it's complaining about not recognizing the control sequence `\section`, I think, which is very odd... Lovely latex errors.

I can reproduce this if it type (in a standard shell) `latexmk` or `latexmk builtins-alonzo.tex` in `plutus-core-spec`.  If instead I type `latexmk plutus-core-specfication.tex` then it builds a `.dvi` file (without the index of notation at the end: `latexmk` maybe doesn't know that it needs to run `makeindex` to build that).  I think that nix is just trying to build the first tex file it sees and that doesn't contain a `\documentclass`, so it's failing.  Presumably nix needs to be told _what_ to build, but I can't see how to do that at the moment.

kwxm:
OK, I found out how to make `latexmk` build the index of notation properly: see https://tex.stackexchange.com/questions/105943/latexmk-and-nomencl

kwxm:
Success!

michaelpj:
Well done!

michaelpj (approved):

PR #5049, opened 2023-01-17
zeme iohk/update std blaggacao

blaggacao:
- fix: rm tullia and instrumentation as they contain warnings
- fix: std action setup
- feat: set up std action for mac & linux (missing secrets & cache, yet)


nix/cells/automation/ciJobs.nix
```diff
@@ -28,15 +28,12 @@ let
...
 
-  other-jobs = inputs.cells.plutus.devshells // inputs.cells.plutus.packages;
-
   jobs =
     # Drop these once we switch to 9.2.4 by default
     { ghc8107 = native-plutus-8107-jobs; } //
     { ghc924 = native-plutus-924-jobs; } //
     # Only cross-compile to windows from linux
-    lib.optionalAttrs (system == "x86_64-linux") { mingwW64 = windows-plutus-924-jobs; } //
-    other-jobs;
```
michaelpj, line 39: I would prefer not to have done this until we were sure we could rely on the new thing.

michaelpj: Building the shells on both hydra and GHA isn't so bad, and it would ensure that things keep working for now.

blaggacao: Yes and yes! I'm still experimenting with this and will continue today. I'm glad @zeme-iohk merged this checkpoint in his branch.

I guess I'm also finally dogfooding what you where saying (at one point) about build matrices and instantiation with different compiler version.

michaelpj: Oh sorry, I didn't notice this wasn't targeting master. Never mind!

blaggacao: No worries, more bits and pieces incoming. I'll revert this shortly once I've figured out more stuff ...

PR #5050, opened 2023-01-17
PLT-828: Use the position of the opening bracket in `appTerm` and `tyInstTerm`

zliu41:
It would make more sense to use the position of the opening bracket, rather than the position of the function (for `appTerm`) or the polymorphic term (for `tyInstTerm`).

Otherwise `[ fun arg ]` would have the same `SourcePos` as `fun`.

Also look at, e.g., the `typeLet` test case - there's no `(con integer 1)` at `typeLet:4:5`.


plutus-core/plutus-ir/src/PlutusIR/Parser.hs
```diff
@@ -94,7 +94,9 @@ letTerm
 letTerm = Let <$> wordPos "let" <*> recursivity <*> NE.some (try binding) <*> pTerm
 
 appTerm :: Parametric
-appTerm tm = inBrackets $ PIR.mkIterApp <$> getSourcePos <*> tm <*> some tm
+appTerm tm = do
+    pos <- getSourcePos
+    inBrackets $ PIR.mkIterApp <$> pure pos <*> tm <*> some tm
 
 tyInstTerm :: Parametric
```
michaelpj, line 101: What about here?

michaelpj (approved):
It looks like many other terms have the same problem, i.e. the position for `(delay x)` is the "delay", not the opening paren. I think we should try and be consistent if we're fixing this.

Also, there's a fundamental problem in that we're attaching a `SrcPos`, whereas I think we should be attaching a source _span_. The `SrcPos` is fine for reporting parse errors, but not for actual term positions.

michaelpj:
I guess what I mean is: I'm fine with this, but I'd prefer to properly overhaul the position reporting to actually do the right thing.

zliu41 (commented):
> I think we should be attaching a source span.

Exactly. That's why in the debugger demo I could only highlight the first token of a Term.

> the position for (delay x) is the "delay", not the opening paren.

Probably. The difference is that there's no "app" when displaying an `appTerm`. In the debugger, for a "delay" term, I think it's better to highlight the "delay" instead of the "(". But all of this will be moot once we move to `SrcSpan`.

zliu41:
I'll make one now.

Issue #5038, opened 2023-01-12
Zigzag encoding formula wrong for negative numbers in plutus-core-specification.pdf


christianschmitz:
The zigzag encoding formula for negative numbers seems wrong.

In file `./doc/plutus-core-spec/draft-new-specification/plutus-core-specification.pdf` (edition Nov 2022), on page 35, the example says `-1` should become `1` once encoded.

The encoding formula for negative numbers is defined as `2(1 - n) + 1` which encodes `-1` as `5`. I think this should be `-2n - 1`.

Minor: the decoding formula for uneven numbers becoming negative numbers is correct, but could be simplified as `-(n+1)/2`

michaelpj:
cc @kwxm

kwxm:
Yes, I think you're right: thanks for pointing that out!  We'll fix that next week.

kwxm:
@christianschmitz This should be fixed in https://github.com/input-output-hk/plutus/pull/5052.  I tried to add you as a reviewer, but I don't think I'm allowed to do that.

Note that we've recently moved the specification to https://github.com/input-output-hk/plutus/tree/master/doc/plutus-core-spec and added the latex source.

Thanks again for your input.

christianschmitz:
Thank you for correcting this, and thank you for including the latex source!

PR #5052, opened 2023-01-18
Kwxm/plc spec/fix zigzag

kwxm:
@christianschmitz pointed out in #5038 that the definition of the zigzag encoding was wrong for negative integers. This should fix it.


michaelpj:
Can we bin the PDF? We should be able to get it from hydra in just the same way that we do for the other one.

michaelpj (approved):

kwxm:
> Can we bin the PDF? We should be able to get it from hydra in just the same way that we do for the other one.

Can I have a link?  I'm never able to find these things.  Also, is the location likely to change any time soon?

michaelpj:
But there is no way we need a checked in PDF any more!

kwxm:
I still don't know where to find it!

michaelpj:
Click here: 
image

kwxm:
I don't understand.

kwxm:
I'm going to merge this.  We can look at removing the PDF in a separate PR.

PR #5055, opened 2023-01-18
wip: used new pinned version of std-action

blaggacao:


zeme-iohk:
I'm not sure I'm too happy with this.
This PR introduces two new cells `{sphinx + toolchain}` and two new blocks `{papers + tasks}`.
While std'izing plutus I had initially organised the code in a similar manner, but then I opted for a more "flat" hierarchy 
`plutus{library,packages,devshells} + automation{ciJobs}`.
I found it easier to maintain and simpler to reason about. 
A more complex directory structure didn't buy me much.
This new structure makes sense, it's just not what I had in mind, and it breaks the assumption stated in the nix/README, namely: you can find all derivations, one-per-file, in the `packages` block.

blaggacao:
Got ya. Actually, I didn't know your preferences. I tacitly assumed the status quo was simply what most resembled status quo ante.

Anyway, I did this mostly as an exercise for my own understanding, since for me,. as "newcomer", things are a bit at odds with that mentioned simplicity to maintain: the flat structure was hard to reason about; it didn't guide me through the semantic structure of this repository; and without semantics I had a hard time to infer the intent of each target.

I still need to cycle back on the original intent of this and restore the original intent to reproduce the builds of `ciJobs` with the build matrices. Then we you may cherry pick the concepts you like, I'll try to be disciplined when commiting.

Therefore, I'll put this in draft, for now.

PR #5060, opened 2023-01-20
Add comment about `applyCostModelParam` being costly.

### LINKED ISSUE #4962: Performance issues with `applyCostModelParams`
### Summary

We noticed some significant performance issues with `Cardano.Api.Fees.evaluateTransactionFee` and `Cardano.Api.TxBody.makeTransactionBody`. After profiling the code, we noticed that the reason was because the functions would convert `Cardano.Api.ProtocolParameters` to cardano-ledger's `PParams` every time we call one of the functions. However, the conversion between the two datatypes take s a significant amount of time because of a plutus-core function (applyCostModelParams ) when creating the cost model.

An issue was created on `cardano-node` repo https://github.com/input-output-hk/cardano-node/issues/4622 in order to bypass the conversion.

Run profiling on a code which uses any function from cardano-api which uses `toLedgerPParams` such as `makeTransactionBody` and `evaluateTransactionFee`.

### Actual Result

Slow

### Expected Result

Faster

No idea.

### System info

OS: NixOS
Version: master
commit hash: a56c96598b4b25c9e28215214d25189331087244

thealmarty:
Per @kwxm 's suggestion for #4962. Closes #4962.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs
@@ -264,7 +264,9 @@ applySplitCostModelParams prefix model params =
     in CostModel <$> applyParams (_machineCostModel model) machineparams
                  <*> applyParams (_builtinCostModel model) builtinparams
 
--- | Update a CostModel for the CEK machine with a given set of parameters,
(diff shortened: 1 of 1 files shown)


kwxm (approved):
I came here to do this, so you've saved me the trouble!  I think that should be a useful reminder if we run into this again.

PR #5062, opened 2023-01-22
[Builtins] Inline `SatInt` stuff and costing monoids

effectfully:
Just something I noticed in the Core, needs benchmarking, so don't look here yet.

### DIFF
--- a/plutus-benchmark/ed25519-throughput/Main.hs
+++ b/plutus-benchmark/ed25519-throughput/Main.hs
@@ -35,6 +35,7 @@ import Cardano.Crypto.Seed (mkSeedFromBytes)
 import Data.ByteString (ByteString)
 import Data.ByteString qualified as BS
 import Data.ByteString.Hash qualified as Hash
+import Data.SatInt
 import Flat qualified
 import Hedgehog.Internal.Gen qualified as G
 import Hedgehog.Internal.Range qualified as R
@@ -188,7 +189,7 @@ evaluate (UPLC.Program _ _ prog) =
       (_res, Cek.TallyingSt _ budget, _logs) ->
           let ExCPU cpu = exBudgetCPU budget
               ExMemory mem = exBudgetMemory budget
-          in (fromIntegral cpu, fromIntegral mem)
+          in (fromIntegral $ unSatInt cpu, fromIntegral $ unSatInt mem)
 
 -- | Evaluate a script and print out the serialised size and the CPU and memory
 -- usage, both as absolute values and percentages of the maxima specified in the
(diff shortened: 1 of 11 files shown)


plutus-core/plutus-core/src/Data/SatInt.hs
```diff
@@ -49,49 +49,83 @@ instance Read SatInt where
...
 instance Enum SatInt where
-  succ (SI x) = SI (succ x)
-  pred (SI x) = SI (pred x)
-  toEnum                = SI
-  fromEnum              = unSatInt
+  {-# INLINE succ #-}
+  succ = coerce (succ :: Int -> Int)
+
+  {-# INLINE pred #-}
+  pred = coerce (pred :: Int -> Int)
```
effectfully, line 122: @michaelpj we still have this bug.

plutus-core/plutus-core/src/Data/SatInt.hs
```diff
@@ -49,49 +49,83 @@ instance Read SatInt where
...
     | y == minBound = maxBound
     | otherwise     = SI (negate y)
+
+  {-# INLINE abs #-}
   abs x
-    | x >= 0        = x
-    | otherwise     = negate x
-  signum (SI x)     = SI (signum x)
+    | x >= 0    = x
+    | otherwise = negate x
```
effectfully, line 91: Why is it a custom definition? How does it handle near-maximum values? Needs a few tests at the very least.

michaelpj: Basically this stuff all got copied from `safeint`... which sadly means I don't have good justifications for a bunch of things :/

kwxm: I remember thinking at the time that we should dump everything from `safeInt` except for the few things we actually need.

kwxm:
^ Just experimenting to see if running the benchmarks here uses the same machine that we use manually.  Looks like they do.  I've cancelled the benchmarking job now.

plutus-core/plutus-core/src/Data/SatInt.hs
effectfully: I've checked it and I think the current behavior is as correct as the default one:

```
>>> (minBound :: SatInt, negate (minBound :: SatInt))
(-9223372036854775808,9223372036854775807)
>>> (minBound :: SatInt, 0 - minBound :: SatInt)
(-9223372036854775808,9223372036854775807)
 
>>> (maxBound :: SatInt, negate (maxBound :: SatInt))
(9223372036854775807,-9223372036854775807)
>>> (maxBound :: SatInt, 0 - maxBound :: SatInt)
(9223372036854775807,-9223372036854775807)
```

Maybe it's a custom definition to make it faster or clearer.

Would still be great to have such tests if they're not there.

michaelpj (approved):
Looks fine!

PR #5063, opened 2023-01-22
[Builtins] Make builtin application look lazy

effectfully:
Those `oneShot`s should no longer be needed, because we now have them elsewhere. Making the result appear lazy helps sharing at least in case of builtins returning booleans. Needs benchmarking, don't look here yet.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
@@ -45,7 +45,7 @@ import Data.DList (DList)
 import Data.Either.Extras
 import Data.String
 import Data.Text (Text)
-import GHC.Exts (inline, oneShot)
+import GHC.Exts (inline)
 import GHC.TypeLits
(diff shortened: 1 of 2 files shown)


effectfully:
^ +1.28%. I'm genuinely surprised that this is slower, especially given #5186. This does two things though, let's do them separately. First: remove `oneShot` and not introduce the call to `lazy`.

effectfully:
^ -0.33%, i.e. noise. This is good (because removes tricky code) and is as expected. We should merge that change, so I'll create a separate PR for marking `makeKnown` with `lazy`.

effectfully:
> I'll create a separate PR for marking makeKnown with lazy.

Did that: #5188. We have +0.3% there, i.e. also noise. Negative noise + positive noise = positive non-noise? The hell is going on.

effectfully:
Still a lot slower. How can it be? Whatever, but I'm going to triple-check.

effectfully:
OK, no improvement, so I'm closing the PR.

PR #5080, opened 2023-01-24
[Costing] Drop 'uni' and 'ann' from 'MachineParameters

effectfully:
Just a tiny one. Now that `CkValue` and `CekValue` take a different number of arguments, it's even more pronounced that we shouldn't insist a particular kind on the type of values. Which is something we only do in `MachineParameters` as far as I'm aware, hence this PR fixing that.

### DIFF
--- a/plutus-core/cost-model/budgeting-bench/Benchmarks/Nops.hs
+++ b/plutus-core/cost-model/budgeting-bench/Benchmarks/Nops.hs
@@ -124,7 +124,7 @@ nopCostModel =
                   (ModelSixArgumentsConstantCost 600)
     }
 
-nopCostParameters :: MachineParameters CekMachineCosts CekValue DefaultUni NopFun ()
+nopCostParameters :: MachineParameters CekMachineCosts NopFun (CekValue DefaultUni NopFun ())
 nopCostParameters =
     mkMachineParameters def $
         CostModel defaultCekMachineCosts nopCostModel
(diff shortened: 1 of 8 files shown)


michaelpj (approved):
Yeah, this seems better

michaelpj:
image

Uh oh.

effectfully:
> Uh oh.

Inlining shenanigans. We have `defaultCekParameters` and `unitCekParameters`, both call `mkMachineParameters`, which gets inlined in both the cases, giving us 40k lines of very optimizable Core => GHC gets tired. We've had this before. I'll try to remove inlining for `unitCekParameters` at least and maybe increase ticks on top of that if it's not enough. We can't mess with `defaultCekParameters`, because it's being used for benchmarking currently. We'll need to fix that and make sure that `defaultCekParameters` isn't used anywhere where performance is important (including costing calibration), but this is for another PR (and potentially another person to tackle).

effectfully:
@michaelpj and now we're back to

```
plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetingDefaults.hs:31:18: error:
    Module ‘GHC.Exts’ does not export ‘noinline’
   |
31 | import GHC.Exts (noinline)
   |      
```

What's up with that?

bezirg (approved):

michaelpj:
I don't know, but `GHC.Exts` is a special module, so could be something to do with the exact `ghc` we're getting. Maybe import it from `GHC.Magic`?

PR #5087, opened 2023-01-26
Fix nix to build on mac

mjaskelioff:
Building plutus core fails on mac, because it doesn't find the R library.

```
Error: cabal: Could not resolve dependencies:
[__0] trying: plutus-core-1.1.1.0 (user goal)
[__1] rejecting: plutus-core:-with-inline-r (constraint from config file,
command line flag, or user target requires opposite flag selection)
[__1] trying: plutus-core:+with-inline-r
[__2] next goal: inline-r (dependency of plutus-core +with-inline-r)
[__2] rejecting: inline-r-1.0.0.0.0.0.0.1 (conflict: pkg-config package
libR>=3.0, not found in the pkg-config database)
[__2] rejecting: inline-r-1.0.0, inline-r-0.10.5, inline-r-0.10.4,
inline-r-0.10.3, inline-r-0.10.2, inline-r-0.10.1, inline-r-0.10,
inline-r-0.9.2, inline-r-0.9.1, inline-r-0.9.0.2, inline-r-0.9.0.1,
inline-r-0.9.0.0, inline-r-0.8.0.1, inline-r-0.8.0.0, inline-r-0.7.3.0,
inline-r-0.7.2.0, inline-r-0.7.1.2, inline-r-0.7.1.1, inline-r-0.7.1.0,
inline-r-0.7.0.0 (conflict: plutus-core +with-inline-r => inline-r>1.0.0)
[__2] fail (backjumping, conflict set: inline-r, plutus-core,
plutus-core:with-inline-r)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: plutus-core,
plutus-core:with-inline-r, inline-r
```

This fix is thanks to @angerman.

Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested

### DIFF
--- a/nix/cells/plutus/library/make-plutus-project.nix
+++ b/nix/cells/plutus/library/make-plutus-project.nix
@@ -21,6 +21,13 @@ let
     shell = {
       # We don't currently use this.
       withHoogle = false;
+
+
+      # We would expect R to be pulled in automatically as it's a dependency of
+      # plutus-core, but it appears it is not, so we need to be explicit about
+      # the dependency on R here.  Adding it as a buildInput will ensure it's
+      # added to the pkg-config env var.
+      buildInputs = [ pkgs.R ];
     };
 
     inputMap = { "https://input-output-hk.github.io/cardano-haskell-packages" = inputs.CHaP; };


bezirg:
Did u test it locally and it works?

bezirg (approved):

mjaskelioff:
Yes, I tested it. Yesterday it wasn't working because I was been stupid, and was changing branches *inside* the nix shell.

michaelpj:
This seems wrong, this should work automatically. cc @zeme-iohk, didn't we already deal with something like this that was due to a `haskell.nix` issue? Did we ever update to a `haskell.nix` with the fix?

bezirg:
@michaelpj Maybe the problem is that `inline-r` is behind a cabal flag? Which is by default off, but only enabled by the `cabal.project`

zeme-iohk:
We are running a recent haskell.nix, but I can try and update it, and see if `cabal build` works without the `buildInputs = [ pkgs.R ];`

michaelpj:
It definitely should work.

@bezirg I don't think it can be that, otherwise it would be broken on linux too! And anyway, haskell.nix should see the correct flag setting.

PR #5093, opened 2023-01-31
Label more PIR terms pure

zliu41:
This makes the inliner inline more things, and ensures the (upcoming) float-in pass won't affect the inliner negatively.


plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -84,6 +85,9 @@ isPure ver varStrictness = go
             TyAbs {} -> True
             Constant {} -> True
             IWrap _ _ _ t -> go t
+            Let _ _ bs t -> all isPureBinding bs && go t
+            Apply _ (LamAbs _ _ _ body) arg -> go body && go arg
+            TyInst _ (TyAbs _ _ _ body) _ -> go body
```
michaelpj, line 90: I'm unsure about these cases. I'm tempted to say we should just leave it up to `Beta` to turn such things into let-bindings, and leave them alone here. Generally there's a question of how smart this function should be. Unsure though.

michaelpj (approved):
I think this is worth a specific test case for the inliner. 
```
let x = let y = 1 in y
in x
```
or something

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
zliu41: In most of our test cases we either don't run simplifiers or run 4 iterations (which include both beta and inliner), so removing the `Apply` and `TyInst` cases doesn't affect anything (other than `single.golden`).

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -84,6 +85,9 @@ isPure ver varStrictness = go
             TyAbs {} -> True
             Constant {} -> True
             IWrap _ _ _ t -> go t
+            Let _ _ bs t -> all isPureBinding bs && go t
```
effectfully, line 88: Wait, a possibly recursive `Let` can't loop?

zliu41: Good question, I think it can
```
let a (strict) = b
     b (strict) = a
in ...
```

So a group of recursive let-bindings need to be all-nonstrict to be considered pure, right?

michaelpj: Great point.

I suggest a dumb approach: just don't consider recursive let bindings to be pure. They're rare anyway. Worth a Note regardless!

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
michaelpj: So why include them?

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -68,7 +68,8 @@ must be *conservative* (i.e. if you don't know, it's non-strict).
 
 -- | Will evaluating this term have side effects (looping or error)?
 -- This is slightly wider than the definition of a value, as
--- it includes things that can't be returned from the machine (as they'd be ill-scoped).
+-- it includes applications that are known to be pure, as well as
+-- things that can't be returned from the machine (as they'd be ill-scoped).
```
effectfully, line 72: @michaelpj I find this:

> as well as things that can't be returned from the machine (as they'd be ill-scoped).

very confusing. So for one, we don't have any machine for PIR, but I assume this talks about the UPLC subset of PIR. What's an example of something that can't be returned from the machine? A standalone variable I guess, since the comment talks about scoping, but it can't be returned not because it would be out of scope, but because the machine takes care of looking up variables and so we simply never run into handling variables when returning from the machine. In fact it is perfectly possible to return a free variable from the machine if you hide it behind a never-evaluated lambda.

I think we need a more formal definition of what "pure" means. Ideally, explicitly put into the context of our evaluation semantics (e.g. a variable wouldn't be pure if any of our evaluators was lazy). Maybe we can ask one of our formal methods guys to help out?

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
zliu41: Removed.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
zliu41: > just don't consider recursive let bindings to be pure.

That's fine by me. There's also recursive datatype bindings which I guess can also complicates things.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
michaelpj: > I find this ... very confusing

Me too. If I wrote that then that was not very helpful of past me :sweat_smile: 

I'd be very happy to have a clearer definition. I think I mean something like "when this term is evaluated, regardless of the context, will it have side effects or loop?" (note that we don't completely disregard the context, as you can tell it which variables are strict or not, as that tells you whether evaluating a variable would do anything). I _think_ that's fairly well-defined, but maybe I'm wrong. And it is true that it relies on an implicit model of PIR evaluation that it would be better to make explicit.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
michaelpj: I bet they're all at the top level though.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
effectfully: I've created PLT-1448 for that.

PR #5095, opened 2023-01-31
Moved Script Evaluation Test to GHA

zeme-iohk:
- Added `./.github/workflows/script-evaluation-test.yml` 
- Some renames and "cosmetics" to the other workflow files
- Partly tested locally using `act` and a fork of the repo.
- Must be pushed to master for proper e2e test


.github/workflows/script-evaluation-test.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Run
+        # Run the test cases sequentially. This ensures we don't need to simultaneously store
+        # multiple `ScriptEvaluationEvents`, which are large, in memory. Each test case
+        # contains many script evaluation events, and those are run in parallel based on
+        # the number of available processors.
+        run: |
+          export EVENT_DUMP_DIR="$HOME/mainnet-script-dump-downloaded"
+          nix develop --no-warn-dirty --accept-flake-config --command cabal update 
+          nix develop --no-warn-dirty --accept-flake-config --command \
```
zeme-iohk, line 34: We need two calls to nix develop unfortunately, this won't work:
```
nix develop --no-warn-dirty --accept-flake-config --command cabal update \ 
  cabal update && cabal v2-run ... 
``` 
As 
`nix develop --command` won't allow this syntax -> it breaks at `&&`

.github/workflows/script-evaluation-test.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Download and Unzip Dump Files
+        run: |
+          export LOCAL_DIR="$HOME/mainnet-script-dump-downloaded" 
+          nix develop --no-warn-dirty --accept-flake-config --command \
+            bash ./scripts/s3-sync-unzip.sh s3://plutus/mainnet-script-dump/ \*.event.bz2
+        env: 
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: us-east-1
+          AWS_ENDPOINT_URL: https://s3.devx.iog.io
```
zeme-iohk, line 24: Can someone confirm that these two are correct? `AWS_ENDPOINT_URL` and `AWS_DEFAULT_REGION`
Maybe @Pacman99

.github/workflows/script-evaluation-test.yml
```diff
@@ -0,0 +1,37 @@
+# A nightly job which downloads script evaluation dumps from S3 and runs a regression test.
+name: Script Evaluation Test
+on:
+  schedule:
+    - cron: 30 3 * * * # 3:30am every day
```
zeme-iohk, line 5: I chose at random, happy to change

.github/workflows/script-evaluation-test.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3.3.0
+
+      - name: Download and Unzip Dump Files
+        run: |
+          export LOCAL_DIR="$HOME/mainnet-script-dump-downloaded" 
+          nix develop --no-warn-dirty --accept-flake-config --command \
+            bash ./scripts/s3-sync-unzip.sh s3://plutus/mainnet-script-dump/ \*.event.bz2
```
michaelpj, line 19: Since we're running on a persistent machine, is there a way we can save this somewhere persistent and only sync the remainder?

.github/workflows/script-evaluation-test.yml
michaelpj: Alternatively, what about
```
nix run .#x86_64-linux.plutus.library.plutus-project-924.hsPkgs.plutus-ledger-api.components.exes.evaluation-test -- --num-threads=1
```

build/buildkite/script-evaluation-test.yml
```diff
@@ -1,25 +0,0 @@
...
-    # Run the test cases sequentially. This ensures we don't need to simultaneously store
-    # multiple `ScriptEvaluationEvents`, which are large, in memory. Each test case
-    # contains many script evaluation events, and those are run in parallel based on
-    # the number of available processors.
-    command: |
-      nix-shell --run 'cabal update && EVENT_DUMP_DIR=$HOME/mainnet-script-dump-downloaded cabal v2-run plutus-ledger-api:evaluation-test -- --num-threads=1'
-#     TODO(std) check that this works 
-#     nix develop --command 'cabal update && EVENT_DUMP_DIR=$HOME/mainnet-script-dump-downloaded cabal v2-run plutus-ledger-api:evaluation-test -- --num-threads=1'
-    concurrency: 1
-    concurrency_group: "plutus-script-evaluation"
```
michaelpj, line 23: Is this important? In fact, shouldn't we have a concurrency group for all things that use the benchmarking machine? or is that handled by having the benchmarking machine only accept one job at once?

michaelpj (approved):

build/buildkite/script-evaluation-test.yml
zeme-iohk: yeah I think it's important, we can also cancel previous jobs actually. see my update.

.github/workflows/script-evaluation-test.yml
zeme-iohk: I think the file `$HOME/mainnet-script-dump-downloaded` is going to persist across runs, no?

michaelpj: Oh I see. I also don't know if that's true. Maybe we should check with @Pacman99 .

I also don't know if that script does things incrementally, probably a question for @zliu41

.github/workflows/benchmark.yml
```diff
@@ -3,6 +3,10 @@ on:
   issue_comment:
     types: [created]
 
+concurrency:
+  group: plutus-benchmark
+  cancel-in-progress: true
```
michaelpj, line 8: That seems wrong. If people make benchmark requests on multiple PRs, we want them to queue up and all get executed.

michaelpj: Another reason why it's worth writing explanatory comments for stuff like this!

.github/workflows/script-evaluation-test.yml
```diff
@@ -4,6 +4,10 @@ on:
   schedule:
     - cron: 30 3 * * * # 3:30am every day 
 
+concurrency:
+  group: script-evaluation-test
+  cancel-in-progress: true
```
michaelpj: This seems okay... we shouldn't ever end up with multiple of these in flight? Worth a comment.

.github/workflows/script-evaluation-test.yml
```diff
@@ -30,8 +34,7 @@ jobs:
         # the number of available processors.
         run: |
           export EVENT_DUMP_DIR="$HOME/mainnet-script-dump-downloaded"
-          nix develop --no-warn-dirty --accept-flake-config --command cabal update 
-          nix develop --no-warn-dirty --accept-flake-config --command \
-            cabal v2-run plutus-ledger-api:evaluation-test -- --num-threads=1
+          Alternatively, what about
+          nix run .#x86_64-linux.plutus.library.plutus-project-924.hsPkgs.plutus-ledger-api.components.exes.evaluation-test -- --num-threads=1
```
michaelpj, line 36: We should test this, I'm not 100% sure whether it will work.

michaelpj (approved):

PR #5099, opened 2023-02-01
Sums-of-products mk II

michaelpj:
Don't look here yet.


michaelpj:
:boom: :facepalm: :sob:

michaelpj:
`applyEvaluate` wasn't getting inlined now because it appears multiple times. I added an `INLINE` pragma. That took us a bit closer, but not the whole way there, and I don't know why :thinking:

michaelpj:
The Core now looks identical apart from new stuff :|

PR #5103, opened 2023-02-02
Make two fake terms and two fake frames

michaelpj:


michaelpj:
Now with three frames to match the SOP work.

michaelpj:
Maybe I did something bad, let's check.

PR #5104, opened 2023-02-02
PLT-1462: Add parser for data constant

zliu41:
Currently the `PrettyConst Data` instance prints the serialised `Data`, so I added a corresponding parser.

We should decide on a better syntax for `Data`, such as the one being used in the existing conformance tests. But before we get there, we still need a parser that matches `PrettyConst Data`. The debugger, for example, needs to parse the UPLC in order to get `SourcePos` annotations.


effectfully (approved):

michaelpj:
We used to have a typeclass for builtin types that insisted we had both a prettyprinter and a parser, maybe we should bring that back :thinking:

PR #5105, opened 2023-02-03
Use `SrcSpan` instead of `SourcePos` for UPLC in debugger

zliu41:
Previously `mkUplcSpan` wasn't doing the right thing in the `Returning` case. It used the returned value to determine the length of the highlight. If, for example, the return value is a `lam`, and the original term that evaluated to that value is a `force`, it'd only highlight the first 3 letters of `force` (because `length "lam" == 3`).

This PR fixes it by using `SrcSpan` instead of `SourcePos` in `DAnn`. This is what we need to do anyway, once we make the UPLC parser return `SrcSpan`-annotated terms instead of `SourcePos`-annotated ones.

### DIFF
--- a/plutus-core/executables/debugger/Event.hs
+++ b/plutus-core/executables/debugger/Event.hs
@@ -8,7 +8,6 @@
 module Event where
 
 import Annotation
-import PlutusCore qualified as PLC
 import PlutusCore.Pretty qualified as PLC
 import Types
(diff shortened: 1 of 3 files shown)


michaelpj (approved):
This is fine, although at this point I wonder whether it would be less effort to just make the parser return `SrcSpan`s.

bezirg (approved):

PR #5109, opened 2023-02-05
PLT-1147: Mark more UPLC terms pure

zliu41:
so that the UPLC inliner inlines more things.

This makes the UPLC inliner more resilient to other optimizations, particularly float-in.


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -326,14 +326,19 @@ sizeIsAcceptable = \case
...
+            Var {}                        -> True
             -- These are syntactically values that won't reduce further
-            LamAbs {}   -> True
-            Constant {} -> True
-            Delay {}    -> True
-
+            LamAbs {}                     -> True
+            Constant {}                   -> True
+            Delay _ body                  -> delayIsPure || go delayIsPure body
+            Apply _ (LamAbs _ _ body) arg -> go delayIsPure arg && go delayIsPure body
```
michaelpj, line 337: Okay, so we're not doing it in PIR, but we are doing it here? Why? Presumably because we don't have let-bindings here so we need it in order to spot things like that. But that definitely needs a comment. And TBH that seems to me like an argument for doing it in PIR too - just consistency.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -326,14 +326,19 @@ sizeIsAcceptable = \case
   Delay _ t  -> sizeIsAcceptable t
 
 isPure :: Term name uni fun a -> Bool
-isPure = go
+isPure = go True
     where
-        go = \case
-            Var {}      -> True
+        go delayIsPure = \case
```
michaelpj, line 331: The `delayIsPure` thing at _least_ needs a note. In fact I think it's unsound. Consider:

```
(\x . force x) (delay impure)
```

Your analysis considers the `force x` to be pure but it isn't.

I'm going to be annoying again and say that you'd have noticed this if you had a) an argument for why it was sound, b) some tests illustrating the new behaviour you were trying to create :)

I'd be more convinced by a case for `force (delay x)`, even though that will definitely get optimized away, since it mirrors the case for `(\x . b) t`.

michaelpj (changes requested):

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
michaelpj: This is actually an instance where I think it would have been helpful to do both in one PR, so we can spot whether they're consistent. Keeping the PIR and UPLC inliners in sync is a pain, best to update them together.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
zliu41: Updated and added some tests.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -326,14 +326,50 @@ sizeIsAcceptable = \case
...
-            LamAbs {}   -> True
-            Constant {} -> True
-            Delay {}    -> True
-
+            LamAbs {}                     -> True
+            Constant {}                   -> True
+            Delay _ body                  -> delayAndVarIsPure || go delayAndVarIsPure body
+            -- This case is not needed in PIR's `isPure`, because PIR's beta-reduction pass
+            -- turns terms like this into `Let` bindings.
+            Apply _ (LamAbs _ _ body) arg -> go delayAndVarIsPure arg && go delayAndVarIsPure body
```
effectfully, line 340: It's hard for me to fully comprehend what's going on with the whole purity thing, especially with this new addition, but shouldn't it be `go True arg` instead of `go delayAndVarIsPure arg`?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -326,14 +326,50 @@ sizeIsAcceptable = \case
...
+            Force _ body                  -> go False body
             -- See Note [Differences from PIR inliner] 5
-            _           -> False
+            Builtin{}                     -> True
```
effectfully, line 343: Hm, I thought in PIR we check if the specific builtin is underapplied, in which case it's pure if all of its arguments are pure and type/term argument interleaving is correct. Do I misremember or do we not want to handle this case here for some reason?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
zliu41: Yeah you are right!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
zliu41: `Note [Differences from PIR inliner]` explains why we don't do that. I actually tried it, and it didn't improve any of the existing test cases while adding a lot of lines of code.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -326,14 +326,50 @@ sizeIsAcceptable = \case
...
+
+Once we descend into the body of a `Force` node, however, `Var` and `Delay` nodes can
+no longer be considered unconditionally pure, because the following terms are impure:
+
+```
+force (delay impure)
+force (force (delay (delay impure)))
+force x  -- because `x` may expand into an impure `Term`
+force (force (delay x))
+```
```
michaelpj, line 360: Great

plutus-core/untyped-plutus-core/test/Transform/inlinePure3.plc.golden
```diff
@@ -0,0 +1 @@
+(lam b_1 [ (lam x_4 (lam y_5 [ x_4 x_4 ])) (con integer 1) ])
```
michaelpj, line 1: should something have been inlined here?

plutus-core/untyped-plutus-core/test/Transform/inlinePure4.plc.golden
```diff
@@ -0,0 +1,9 @@
+(lam
```
michaelpj, line 1: ditto

michaelpj (approved):
Do you want to add similar tests for the PIR inliner or do it later?

plutus-core/untyped-plutus-core/test/Transform/inlinePure3.plc.golden
zliu41: No, and this is deliberate. It isn't pre-inlined because `x_4` is used twice. And it isn't post-inlined because `sizeIsAcceptable (Constant {}) = False`.

zliu41 (commented):
> Do you want to add similar tests for the PIR inliner or do it later?

Later. I want to get the float-in PR going first.

plutus-core/untyped-plutus-core/test/Transform/inlinePure3.plc.golden
michaelpj: Maybe you should include a comment, then? Otherwise it's not clear what the intended behaviour is. TBH they could all have comments.

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
```diff
@@ -46,12 +46,22 @@ mkInlinePurityTest termToInline = runQuote $ do
...
+--
+-- Note that this relies on @forceDelayCancel@ to cancel the @force@ and the @delay@,
+-- otherwise the inliner would treat the term as impure.
 inlinePure2 :: Term Name PLC.DefaultUni PLC.DefaultFun ()
 inlinePure2 = mkInlinePurityTest $ Force () . Delay () . Var () <$> freshName "a"
 
+-- | @[(\x -> \y -> [x x]) (con integer 1)]@ is pure.
+--
+-- Note that the @(con integer 1)@ won't get inlined: it isn't pre-inlined because
+-- @x@ occurs twice, and it isn't post-inlined because @costIsAcceptable Constant{} = False@.
```
michaelpj, line 63: should be "sizeIsAcceptable". The cost is fine, the size is not

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
```diff
@@ -36,6 +36,75 @@ basicInline = runQuote $ do
...
+        Apply
+            ()
+            (LamAbs () x $ LamAbs () y $ Apply () (Var () x) (Var () x))
+            (mkConstant @Integer () 1)
+
+-- | @force ([(\x -> delay (\y -> [x x])) (delay ([error (con integer 1)]))])@ is pure.
+--
+-- The @error@ is not evaluated when evaluating this term.
+inlinePure4 :: Term Name PLC.DefaultUni PLC.DefaultFun ()
+inlinePure4 = mkInlinePurityTest $ do
```
michaelpj, line 79: I will say that I find it much harder to review these tests than ones written with literal PIR. It's much easier to look at the diffs between two PIR programs than between this bit of haskell code and the PIR program...

I made several erroneous comments that I had to delete.

(cc @effectfully )

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -326,14 +326,50 @@ sizeIsAcceptable = \case
...
-            LamAbs {}   -> True
-            Constant {} -> True
-            Delay {}    -> True
-
+            LamAbs {}                     -> True
+            Constant {}                   -> True
+            Delay _ body                  -> delayAndVarIsPure || go delayAndVarIsPure body
+            -- This case is not needed in PIR's `isPure`, because PIR's beta-reduction pass
+            -- turns terms like this into `Let` bindings.
+            Apply _ (LamAbs _ _ body) arg -> go True arg && go delayAndVarIsPure body
```
michaelpj, line 340: This case is actually quite subtle, I had to stare at it for a while to work out why it's okay. Worth a comment, and does one of the tests witness that it does the right thing?

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
effectfully: > I will say that I find it much harder to review these tests than ones written with literal PIR. It's much easier to look at the diffs between two PIR programs than between this bit of haskell code and the PIR program...

> I made several erroneous comments that I had to delete.

I think the solution is trivial here: print the unoptimized PIR code alongside the optimized one. NOT write PIR by hand. Only a very few extremely evil people deserve this kind of punishment.

PR #5110, opened 2023-02-06
Updated S3 location in script-evaluation-test.yml

zeme-iohk:


michaelpj:
Okay, we need a comment here. Ideally one that explains the intended structure of the bucket so we don't get this wrong in the future. @zliu41 could you draft something?

zliu41:
I'll add a comment. This is temporary; once the other job is restored we should change it back. The correct S3 path should always be `s3://plutus/mainnet-script-dump/`.

zliu41:
On a related note, there's a problem with script dump job: if we upgrade the cardano-ledger version and the new version has an incompatible LedgerState, we'll need to re-create the dump from scratch since the job won't be able to deserialize the checkpoint file (which is basically LedgerState). That is why I moved the old event files - I was going to re-create them.

We have a ticket for this: PLT-497. I heard there's a ledger state transition function that converts ledger state from on era to the next. We should talk to Jared about how we can use it to solve the problem.

PR #5111, opened 2023-02-06
[WIP] Improve benchmarking.yml GHA

zeme-iohk:
- Since it's now all running on github, there's no need to have the benchmark script post the comment with the results. We can just do it in a step following the step that runs the script.
- The rocket reaction is fine, but it's nice to see how the job is doing. The buildkite job used to post a commit status, which was okay, but doesn't actually work well if you have multiple jobs running. So instead we post a comment with a link to the triggered action run.


.github/workflows/benchmark.yml
```diff
@@ -90,9 +92,61 @@ jobs:
       - name: Run
         run: |
           nix develop --no-warn-dirty --accept-flake-config --command bash ./scripts/ci-plutus-benchmark.sh 
-        if: success()
```
zeme-iohk, line 93: turns out that `if: success()` is implied in each step

.github/workflows/benchmark.yml
```diff
@@ -90,9 +92,61 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ github.token }}
           BENCHMARK_NAME: ${{ steps.extract-benchmark.outputs.benchmark }}
           PR_NUMBER: ${{ github.event.issue.number }}
           PR_BRANCH: ${{ steps.extract-branch.outputs.head_ref }}
+
+      - name: Publish Link To Action Run
```
michaelpj, line 101: Can't we do this more simply with something like this: https://stackoverflow.com/questions/59073850/github-actions-get-url-of-test-build

michaelpj: Also, I was thinking we should do this _before_ running the benchmark! The idea is that you get a link so you can see how it's progressing... which is not so useful once it's finished :)

michaelpj (commented):
Another idea that I had: at the moment you don't get any feedback until all previous jobs have finished, because it's the same job that posts the reaction that also runs the benchmark.

I think we could fix this by having two jobs in the workflow:
- `process_comment`: gets all the stuff out of the comment, puts it in job outputs; posts a comment to the workflow run; runs on any machine (so can run in parallel with others!)
- `benchmark`: depends on the previous job; gets the information out; runs the benchmark; runs on the self-hosted runner

.github/workflows/benchmark.yml
zeme-iohk: > Also, I was thinking we should do this before running the benchmark! The idea is that you get a link so you can see how it's progressing... which is not so useful once it's finished :)

Yes good catch, this should go before the `Run` step 😁 

> Can't we do this more simply with something like this: https://stackoverflow.com/questions/59073850/github-actions-get-url-of-test-build

Yes, something like that (e.g. https://github.com/input-output-hk/plutus/actions/runs/4102559646) is almost perfect, it was my starting point, but one still needs to click the `Benchmark` button to actually access the logs... I wanted to avoid that extra click.
I won't waste too much time on it though, testing this stuff locally is a nightmare

zeme-iohk:
Click [here](https://github.com/input-output-hk/plutus/actions/runs/4102717722/jobs/[object Object]) to check the status of your benchmark.

zeme-iohk:
Click [here](https://github.com/input-output-hk/plutus/actions/runs/4102717722/jobs/11132943649) to check the status of your benchmark.

zeme-iohk:
Click [here](https://github.com/input-output-hk/plutus/actions/runs/4102717722/jobs/11132943649) to check the status of your benchmark.

zeme-iohk:
Click [here](https://github.com/input-output-hk/plutus/actions/runs/4102717722/jobs/11132943649) to check the status of your benchmark.

zeme-iohk:
Click [here](https://github.com/input-output-hk/plutus/actions/runs/4102717722/jobs/11132943649) to check the status of your benchmark.

zeme-iohk:
Click [here](https://github.com/input-output-hk/plutus/actions/runs/4102717722/jobs/11132943649) to check the status of your benchmark.

zeme-iohk:
Click [here](https://github.com/input-output-hk/plutus/actions/runs/4102717722/jobs/11132943649) to check the status of your benchmark.

zeme-iohk:
Comparing benchmark results of 'nofib' on '' (base) and '' (PR)

<details>
<summary>Results table</summary>

| Script                                   |  Old       |  New       |   Change   |
| :------                                  |  :------:  |  :------:  |  :------:  |
</details>

michaelpj:
Another improvement: we could use `upload-artifact` to upload the two run logs and the comparison table as artifacts. Maybe not super necessary, but perhaps useful for debugging. And I guess it would let you put the final comment posting in its own job too, although I don't think that obviously wins us anything.

zeme-iohk:
I've added support for workflow_dispatch, so that we can test this without having to push to master. Can we merge this and then think about further improvements?

.github/workflows/benchmark.yml
```diff
@@ -12,35 +18,83 @@ jobs:
   
     if: | 
       startsWith(github.event.comment.body, '/benchmark') && 
-      github.event.issue.pull_request 
+      github.event.issue.pull_request || 
+      github.event_name == 'workflow_dispatch'
```
michaelpj, line 22: does this associate the way you expect? Can we bracket it please, I hate trying to remember precedence.

.github/workflows/benchmark.yml
```diff
@@ -12,35 +18,83 @@ jobs:
...
+                repo: context.issue.repo,
+                comment_id: context.payload.comment.id,
+                content: "rocket"
+              });  
+            }
+
+            if (core.getInput("event") === "issue_comment") {
+              await reactWithRocket();
+            } else {
+              core.debug("Workflow dispacted from GitHub GUI, not reacting with Rocket.")
```
michaelpj, line 70: This should be impossible given the `if` guard in the step, right? So maybe it should fail?

michaelpj (approved):
Just a nit, otherwise yes let's merge and improve later.

.github/workflows/benchmark.yml
zeme-iohk: Good catch, I'll remove the `if:`

PR #5112, opened 2023-02-06
Unrestricted force

michaelpj:
Don't look here yet. This is an experiment I've wanted to do for a
while, just looking to get some benchmark numbers.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


L-as:
Rather than doing this as a post-hoc optimisation, why not omit emitting unnecessary delays and forces? That seems like a simpler thing to do.

L-as:
You could in PIR make one constructor for both type abstractions and lambdas of arbitrary (type) parameters, which would seemingly allow you to emit more optimised UPLC.

michaelpj:
> Rather than doing this as a post-hoc optimisation, why not omit emitting unnecessary delays and forces? That seems like a simpler thing to do.

I disagree with this as a compiler design principle. IME (and I think this is the standard wisdom) it's almost always better to make code-gen as dumb as possible and then optimize away the bad stuff. That makes your code gen simpler (which is important since it's almost always pretty complicated anyway), and your optimizations will benefit you in other places.

michaelpj:
Pretty modest tbh, slightly disappointing.

PR #5114, opened 2023-02-06
PLT-1445: Add more tests to the inliner pass.

thealmarty:
The added tests target the upcoming implementation of call site inlining of fully applied functions. Ideally we should see improvements for them once the implementation is merged.

Some tests also target testing of unconditional inlining of functions to ensure there is no regression when the implementation is done.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-core/plutus-ir/test/transform/inline/letApp
```diff
@@ -0,0 +1,47 @@
+(let
```
thealmarty, line 1: This seems to be more lengthy than needed? Will look into this more. This is the PlutuxTx code:
```haskell
monoId :: CompiledCode Integer
{-# NOINLINE monoId #-}
monoId = plc (Proxy @"monoId") (
    let appNum :: Integer
        {-# NOINLINE appNum #-}
        appNum = 4
        funApp :: Integer -> Integer
        {-# NOINLINE funApp #-}
        funApp = (\x y -> x + y) appNum
    in funApp 5
    )
```

plutus-core/plutus-ir/test/transform/inline/letFunConst
```diff
@@ -0,0 +1,17 @@
+(let
```
thealmarty, line 1: The inliner already inlines `constFun`?

thealmarty:
Ignore "letApp" for now. The other tests show that the current inliner already inlines the functions?

plutus-core/plutus-ir/test/transform/inline/letApp
thealmarty: Updated to a simplified program. See investigation [here](https://gist.github.com/thealmarty/e967958f8d4cfc3657e76fb0cf561425).

plutus-core/plutus-ir/test/TransformSpec.hs
```diff
@@ -139,6 +139,13 @@ inline =
     , "single"
     , "immediateVar"
     , "immediateApp"
+    -- these tests are all let bindings of functions, for testing call site inlining of fully
+    -- applied functions.
+    ,"letFunConst" -- const fun fully applied
```
michaelpj, line 144: Note that I think we do allow comments in PIR files? So you can comment in the input file what you're expecting. I don't think we make much use of that at the moment, but maybe we should.

plutus-core/plutus-ir/test/transform/inline/letFunInFun.golden
```diff
@@ -0,0 +1 @@
+[ (lam eta (con integer) (lam x (con integer) x)) (con integer 3) ]
```
michaelpj, line 1: again, single use things get inlined

plutus-core/plutus-ir/test/TransformSpec.hs
```diff
@@ -139,6 +139,13 @@ inline =
...
+    , "letFunInFun" -- fully applied fn inside another let: let f = id, g = \x.f in g 3
+    , "letIdFunForall" -- id function fully applied
+    , "letApp" -- function application in rhs
+    , "letOverApp" -- over-application of a let
```
michaelpj, line 148: You mean overapplication of a function, right? It's not the let term that's being applied?

plutus-core/plutus-ir/test/transform/inline/letOverApp
```diff
@@ -0,0 +1,14 @@
...
+    (nonrec)
+    (termbind
+        (strict)
+        (vardecl idFun (fun (con integer) (con integer)))
+        (lam x (con integer) x)
+    )
+    (termbind
+        (strict)
+        (vardecl funApp (fun (con integer) (fun (con integer) (con integer))))
+        (lam eta (con integer) (lam eta (con integer) [ idFun eta ]))
```
michaelpj, line 11: GHC has eta-expanded this for you, so it's not even over-applied.

michaelpj: Same comment about single-occurrence inlining.

michaelpj (dismissed):

plutus-core/plutus-ir/test/transform/inline/letFunConst
```diff
@@ -0,0 +1,17 @@
...
+      (vardecl True Bool) (vardecl False Bool)
+    )
+  )
+  (termbind
+    (strict)
+    (vardecl constFun (fun (con integer) (fun Bool (con integer))))
+    (lam x (con integer) (lam y Bool x))
+  )
+  [ [ constFun (con integer 3) ] False ]
+)
```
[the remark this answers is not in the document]
thealmarty: True but the point is not to simplify things?

plutus-core/plutus-ir/test/transform/inline/letApp
```diff
@@ -0,0 +1,27 @@
...
+        eta
+        (con integer)
+        [
+          [ AddInteger appNum ]
+          eta
+        ]
+      )
+    )
+    [ funApp (con integer 5) ]
+)
```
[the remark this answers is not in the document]
thealmarty: I actually think we should move away from crafting the tests by hand, because we are not expecting users to do that and there are other factors affecting the output PIR code. See also [here](https://input-output.atlassian.net/browse/PLT-1489).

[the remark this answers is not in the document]
michaelpj: I said this in our meeting but I _strongly_ disagree with this. The point of these test cases is to test that the inliner is doing what we expect it to do. They absolutely _should_ be small and synthetic so that they're easy to inspect. They are _not_ supposed to be realistic examples, we have other tests for that.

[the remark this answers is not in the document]
thealmarty: These are all generated by the compiler and they are small enough to read I think? Eg, I'd be more comfortable if we have tests including `Bool` instead of only `Integer` even though `Bool` is longer. In fact, to actually test non-single occurrences they'll need to be longer still......maybe need to stick with `Integer` for those. I'll see. It's true we want them tractable!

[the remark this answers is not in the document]
michaelpj: I think we should try to test one thing at a time, as much as possible. If we don't think that having `Bool` in there will affect the behaviour of the inliner, then it's just noise that makes this test case harder to understand, IMO. 

I tend to think of test cases like this as witnessing a specific piece of behaviour, and so you want them to be as small and direct as possible while still showing the behaviour that you want.

plutus-core/plutus-ir/test/TransformSpec.hs
thealmarty: Will do.

plutus-core/plutus-ir/test/transform/inline/letApp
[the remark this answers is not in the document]
michaelpj: Right, but I'm suggesting that we want to be precise about what _statements_ we're verifying. I think it's more helpful to verify lots of small specific statements than fewer larger complex statements. Adding random additional complexity that we don't particularly expect to have an effect just muddles things.

If you want to have testcases that _specifically_ check whether using other datatypes has any effect, then sure, we can do that. But don't mix it up with the tests for other things, IMO.

Also obviously we can't test all combinations of "doing X other thing has no effect", we have to be judicious. I would only write such a test if I thought there was a reason such a thing _could_ interfere, but e.g. I'd written code to guard against it and I wanted to verify that it worked.

thealmarty:
So these tests ended up showing the current inliner does inline functions that have single occurrence. I will tidy it up and merge this and open another one for actually targeting call-site inlining.

michaelpj:
> So these tests ended up showing the current inliner does inline functions that have single occurrences. I will tidy it up and merge this and open another one for actually targeting call-site inlining of fully applied functions.

Why not fix these examples so that they don't get inlined? i.e. by adding multiple occurrences? I'm not sure they add much to our test suite if they're just "more examples of inlining single-occurrence things".

thealmarty:
> Why not fix these examples so that they don't get inlined? i.e. by adding multiple occurrences? I'm not sure they add much to our test suite if they're just "more examples of inlining single-occurrence things".

I can do that too. I thought it's good to include these too because they are different from what we have. But yeah I can just update these.

thealmarty:
Sorry I just don't understand the sudden change of desire here to have fewer tests. I don't think you'd agree that we just need one test for a single occurrence thing. These are different let bindings. I don't think having more tests can hurt? I just want a bit more clarification on what we want here.

michaelpj:
I think we need to have a team discussion about testing philosophy :)

Maybe it would help if we imagine that we had test coverage metrics. Adding more tests that doesn't increase your test coverage has limited value, because you're just testing the same code paths. Ideally you want more tests _that cover more of the behaviour of your code_. More tests in and of themselves are not valuable IMO.

> I don't think you'd agree that we just need one test for a single occurrence thing.

No, probably I would want:
- A simple test for a single occurrence binding
- A test where the RHS is very complex so wouldn't get inlined unless it's single-occurrence
- A test with multiple occurrences, to test that we don't just inline everything
- A test with a single occurrence but it's strict and impure, so we shouldn't inline it
- A test where there's a single occurrence, but it's a recursive binding
i.e. all the interactions and behaviours that I wrote code for

(the fact that the tests are not that clear and done in that way is of course my fault in this case :) but we can do better in future!)

Just having lots of tests that are slightly different versions of the first one is much less interesting.

To be clear, it's also helpful to have some specific tests for cases you specifically care about. E.g. with saturated inlining, we know that we care about inlining `const`, so that's useful to have as a test case. That's probably more of a regression test than anything else, though.

plutus-core/plutus-ir/test/TransformSpec.hs
thealmarty: So I just found out that the parser can't seem to handle comments. I don't know why atm, I checked the code and it looks like it should. :( Should I digress yet again to fix the parser? It seems we want to document the tests a bit more.

thealmarty: nvm, I think I found the problem, will push a fix.

thealmarty:
> I think we need to have a team discussion about testing philosophy :)
> 
I agree! I thought we were pretty much on the same page before but I was wrong!

> Maybe it would help if we imagine that we had test coverage metrics. Adding more tests that doesn't increase your test coverage has limited value, because you're just testing the same code paths. Ideally you want more tests _that cover more of the behaviour of your code_. More tests in and of themselves are not valuable IMO.
> 

I agree it's best to seek out tests that are more meaningfully testing specific behaviour, but I don't think it's generally agreed that test cases should be limited to a smaller amount. 

For a specific behaviour that we have in mind, we definitely need a test for it. But part of the point of testing is to *be surprised*! Some things may look like it's the same code path but it may be a corner case that we didn't expect. In short, I am more inclined to say "we don't know, let's test it and see". I think this is also reflected in the fact that test generators exist - people want to generate so many test cases that they don't want to write it by hand! 

> > I don't think you'd agree that we just need one test for a single occurrence thing.
> 
> No, probably I would want:
> 
>     * A simple test for a single occurrence binding
> 
>     * A test where the RHS is very complex so wouldn't get inlined unless it's single-occurrence
> 
>     * A test with multiple occurrences, to test that we don't just inline everything
> 
>     * A test with a single occurrence but it's strict and impure, so we shouldn't inline it
> 
>     * A test where there's a single occurrence, but it's a recursive binding
>       i.e. all the interactions and behaviours that I wrote code for
> 

These are all good examples that we should add, I agree.
> 
> (the fact that the tests are not that clear and done in that way is of course my fault in this case :) but we can do better in future!)
> 

We don't say the word "fault" here :) but yes let's improve the tests in this way! 

I've fixed the parser so we can add comments to the input files now. We can add the purposes of the tests there. I like adding it to the line where we include the tests too, so it's clear from there all at once. As we add more implementation it would be great to see the test outputs improve. Frankly with this PR I'm pretty sold on test-driven development now!

> Just having lots of tests that are slightly different versions of the first one is much less interesting.
> 
I agree. E.g., we don't need to add test cases of the functions applied to different integers, but I guess I just feel these ones are different/interesting enough.

> To be clear, it's also helpful to have some specific tests for cases you specifically care about. E.g. with saturated inlining, we know that we care about inlining `const`, so that's useful to have as a test case. That's probably more of a regression test than anything else, though.

I'm not completely sure what you're referring to here. The planned implementation for call site inlining right now depends on what `const`'s body is. Otherwise I don't think `const` is treated differently?

michaelpj:
> But part of the point of testing is to be surprised! Some things may look like it's the same code path but it may be a corner case that we didn't expect.

So I guess I think that that is usually covered by having bigger realistic examples, which we do have in our other test suites. Partly my objection is that it's _hard_ to find surprising examples by just writing random larger test cases, _especially_ when they are actually just minor variations on existing ones. That's why it's better to get them from a) test generation (which really can try lots of examples), b) examples produced from real-world-ish usage (which will uncover stuff that matters in practice), c) regression tests (i.e. preserved and condensed examples from b).

> I'm not completely sure what you're referring to here. The planned implementation for call site inlining right now depends on what const's body is. Otherwise I don't think const is treated differently?

Well, `const`'s body is always a variable, that's the point :) My point was: one of the things we were hoping to get out of it was to have it inline `const`, where currently we have a hack for that. So we could reasonably have a test for inlining const, since it's a specific goal. Then I was arguing that that's more like a "regression test", since it's a condensed version of something we saw in the real world that we want to make sure continues working.

thealmarty:
OK yes these tests (including the `const` one) are good regression tests, since I'll be changing the current inliner.

plutus-core/plutus-ir/test/transform/inline/letFunConst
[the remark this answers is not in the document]
thealmarty: It's nice to see `datatypebind` not messing up things? Will add a comment.

plutus-core/plutus-ir/test/transform/inline/letFunConstBool
```diff
@@ -1,3 +1,15 @@
+{-
```
michaelpj, line 1: I love these so much!

plutus-core/plutus-ir/test/transform/inline/letFunInFunAllMulti
```diff
@@ -0,0 +1,28 @@
...
+      (strict)
+      (vardecl idFun (fun (con integer) (con integer)))
+      (lam x (con integer) x)
+    )
+    (termbind
+      (strict)
+      (vardecl g (fun (con integer) (fun (con integer) (con integer))))
+      (lam y (con integer) idFun)
+    )
+    [ g (con integer 3) ]
```
michaelpj, line 27: Where's k?

michaelpj: In fact a couple of these don't match?

plutus-core/plutus-ir/test/transform/inline/letFunInFunMulti
```diff
@@ -0,0 +1,28 @@
...
+      (strict)
+      (vardecl idFun (fun (con integer) (con integer)))
+      (lam x (con integer) x)
+    )
+    (termbind
+      (strict)
+      (vardecl g (fun (con integer) (fun (con integer) (con integer))))
+      (lam y (con integer) idFun)
+    )
+    [ g (con integer 3) ]
```
michaelpj, line 27: Doesn't match commetn

plutus-core/plutus-ir/test/transform/inline/letFunInFunMultiFullyApplied
```diff
@@ -0,0 +1,29 @@
...
+      (strict)
+      (vardecl idFun (fun (con integer) (con integer)))
+      (lam x (con integer) x)
+    )
+    (termbind
+      (strict)
+      (vardecl g (fun (con integer) (fun (con integer) (con integer))))
+      (lam y (con integer) idFun)
+    )
+    [ g idFun (con integer 3) ]
```
michaelpj, line 28: doesn't match

plutus-core/plutus-ir/test/transform/inline/letFunConstBool
michaelpj: Many of the golden files are missing

thealmarty: Sorry I was still working on them. It's a draft. I'll ping you when ready.

plutus-core/plutus-ir/test/transform/inline/letAppMulti
```diff
@@ -0,0 +1,40 @@
+{- 
+Multiple occurrences of a let binding that is a function application (funApp). It should not get 
+inlined by the unconditional inlining optimization. However, the optimization that inlines fully
+applied function should inline funApp.
```
zliu41, line 4: I don't think so. If you inline `[ funApp appNum ]`, it becomes `[ [ (builtin addInteger) appNum ] appNum ]`, which is bigger, so it makes no sense.

plutus-core/plutus-ir/test/transform/inline/letFunConstMulti
```diff
@@ -0,0 +1,21 @@
+{-
+Multiple occurrences of the const function:
+The const function should not be inlined by the unconditional inlining optimization.
+The first constFun should be inlined the the optimization that inlines fully applied functions.
```
zliu41, line 4: type - the the

plutus-core/plutus-ir/test/transform/inline/letFunConstMulti
```diff
@@ -0,0 +1,21 @@
...
+The second constFun is not fully applied so it should not be inlined.
+ 
+let 
+  constFun :: Integer -> Integer -> Integer
+  constFun x y = x
+in constFun 3 (constFun 3)
```
zliu41, line 10: This doesn't typecheck - `constFun 3` is not an integer

plutus-core/plutus-ir/test/transform/inline/letFunInFunMultiFullyApplied.golden
```diff
@@ -0,0 +1,9 @@
+(let
+  (nonrec)
+  (termbind
+    (strict)
+    (vardecl idFun (fun (con integer) (con integer)))
+    (lam x (con integer) x)
+  )
+  [ [ (lam y (fun (con integer) (con integer)) idFun) idFun ] (con integer 1) ]
```
zliu41, line 8: Neither of the two occurrences of `idFun` is fully applied, so I don't think it will be inlined.

plutus-core/plutus-ir/test/transform/inline/letOverAppMulti
```diff
@@ -0,0 +1,57 @@
...
+k being inlined verifies that an over-applied function still gets inline.
+
+let 
+    idFun :: Integer -> Integer
+    idFun y = y
+    funApp :: (Integer -> Integer) -> (Integer -> Integer)
+    funApp x = idFun
+    k :: (Integer -> Integer) -> (Integer -> Integer)
+    k = funApp
+in k k 6
```
zliu41, line 15: This doesn't typecheck.

plutus-core/plutus-ir/test/transform/inline/letOverAppMultiId.golden
```diff
@@ -0,0 +1,9 @@
+(let
+  (nonrec)
+  (termbind
+    (strict)
+    (vardecl idFun (fun (con integer) (con integer)))
+    (lam y (con integer) y)
+  )
+  [ [ (lam x (con integer) idFun) idFun ] (con integer 6) ]
```
zliu41, line 8: Neither of the two `idFun`s is fully saturated.

zliu41 (changes requested):

plutus-core/plutus-ir/test/transform/inline/letOverAppMulti
```diff
@@ -0,0 +1,57 @@
...
+it is fully applied.
+k being inlined verifies that an over-applied function still gets inline.
+
+let 
+    idFun :: Integer -> Integer
+    idFun y = y
+    funApp :: (Integer -> Integer) -> (Integer -> Integer)
+    funApp x = idFun
+    k :: (Integer -> Integer) -> (Integer -> Integer)
+    k = funApp
```
michaelpj, line 14: I think it would generally make your life easier to just add dead bindings that reference the binding you don't want to be single-occurrence inlined.

plutus-core/plutus-ir/test/transform/inline/letOverAppMultiId.golden
michaelpj: Again, this would only work if we were running much more of the simplifier, which we shouldn't in the inliner tests anyway.

plutus-core/plutus-ir/test/transform/inline/letAppMulti
thealmarty: Looking at the [code](https://github.com/input-output-hk/plutus/blob/ed44f958a37f126b2aaa33cd3569691d1c32bbb7/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs#L211) it seems builtins are cost and size acceptable? Also, I think it makes sense to accept builtins, especially seeing [this](https://gist.github.com/thealmarty/a4facd1ed3b25302eb60497c1f10a696).

plutus-core/plutus-ir/test/transform/inline/letFunConstMulti
thealmarty: Oops, I just found another advantage of writing in PlutusTx! ;)

Polymorphic `const` sounds interesting, maybe I should add that?

plutus-core/plutus-ir/test/transform/inline/letFunInFunMultiFullyApplied.golden
thealmarty: Sorry I don't see why they are not fully applied. First application is
```[ (lam y (fun (con integer) (con integer)) idFun) idFun ]``` in Haskell ```\y.idFun idFun``` so it simplifies to `idFun`.

Second one is what the first simplifies to (i.e. `idFun`) applies to `(con integer 1)`.

File: plutus-core/plutus-ir/test/transform/inline/letOverAppMultiId.golden (hunk 18 above)
Comment (thealmarty) [reply plutus-core/plutus-ir/test/transform/inline/letOverAppMultiId.golden:8]: I just realized this is the same as the one above. Deleting.

File: plutus-core/plutus-ir/test/transform/inline/letAppMulti (hunk 13 above)
Comment (thealmarty) [reply plutus-core/plutus-ir/test/transform/inline/letAppMulti]: Although applications are not cost and size acceptable, so it should not get inlined, as desired.

**Comment by thealmarty** [comment]
I don't know why some of the comments went away after I pushed some updated code. Anyway, I think they are all addressed except [this](https://github.com/input-output-hk/plutus/pull/5114/files#r1105286068).

File: plutus-core/plutus-ir/test/transform/inline/letAppMulti [hunk 20]
```diff
@@ -0,0 +1,42 @@
...
+let 
+  appNum :: Integer
+  appNum = 4
+  in 
+    let funApp :: Integer -> Integer
+        funApp x = addInteger appNum x
+    in
+      let k :: Integer -> Integer
+      k = funApp
+in funApp appNum
```
zliu41, line 17: This still won't (and shouldn't) get inlined, because replacing `funApp appNum` with `addInteger appNum x` _increases_ the size of the program.

zliu41: Oh, I just realized that that is exactly what the test is for, so disregard the comment above.

plutus-core/plutus-ir/test/transform/inline/letFunInFunMultiFullyApplied.golden
zliu41: > In Haskell: \y.idFun idFun so it simplifies to idFun.

Like @michaelpj said, in order to simplify `(\y -> idFun) idFun` to `idFun`, you need to run beta reduction. And if you do run beta reduction to simplify `(\y -> idFun) idFun` to `idFun`, then `idFun` will now have a single occurrences, and will be unconditionally inlined. So this cannot test inlining fully saturated functions.

thealmarty: hm...I'll think about whether I want to delete this or somehow make it not single occurrence then. Any preference? Maybe I can just delete this?

zliu41: You can just change `(con integer 1)` to `[ idFun (con integer 1) ]`. Then this `idFun` occurrence will be inlined, the other two won't.

michaelpj: I think it helps to focus on the fact that these are tests _just_ for the inliner. That's the only thing that we run! So in particular:
- You can't assume other simplification will happen, that's not what we're doing
- Dead code won't run

The second bit means that you can easily make things multiple-occurrence by just adding references to them in dead code, e.g.
```
let x = ..
      y = x
in ... x ...
```

thealmarty: Yes, I know that's what you mean about dead code, but remember? The unconditional inliner has deadcode removal in it! See [nameUsedAtMostOnce](https://github.com/input-output-hk/plutus/blob/48c326a1c155be44b67a366241eb29c80e3beed2/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs#L161).

thealmarty:
Pushed the final fix. Please approve so I can merge?

plutus-core/plutus-ir/test/transform/inline/letOverAppMulti
michaelpj: Goddamn it, so it does. Maybe we should stop doing that :joy:

thealmarty: Perhaps! But it's also so easy to do right there!

thealmarty:
@zliu41 @michaelpj please approve so I can merge this. Thanks!

zliu41 (approved):

PR #5116, opened 2023-02-07
[Test] [Builtin] Add 'ForallFortyTwo'

effectfully:
Testing that builtins with one type argument and no term arguments are OK.


plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
```diff
@@ -258,7 +288,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
     toBuiltinMeaning _ver IdRank2 =
         makeBuiltinMeaning
             (Prelude.id
-                :: afa ~ Opaque val (TyForallRep a (TyVarRep f `TyAppRep` TyVarRep a))
+                :: afa ~ Opaque val (TyForallRep @GHC.Type a (TyVarRep f `TyAppRep` TyVarRep a))
```
effectfully, line 291: This is because I added `PolyKinds` to the list of extensions.

michaelpj (approved):

PR #5118, opened 2023-02-08
PLT-1375: Highlight source code while tracing through Plutus Core

bezirg:
This allows highlighting of the source Tx file using the SourceSpans annotations.

Also it allows a bit more input choices for the debugger executable. I tried to reuse parts of the execlib where possible;
and for that i did a small change to execlib that should not break the other executables.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-core/executables/debugger/Event.hs
+++ b/plutus-core/executables/debugger/Event.hs
@@ -1,8 +1,6 @@
 {-# LANGUAGE FlexibleInstances     #-}
-{-# LANGUAGE LambdaCase            #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings     #-}
-{-# LANGUAGE TupleSections         #-}
 
 -- | Handler of debugger events.
 module Event where
@@ -10,7 +8,6 @@ module Event where
 import PlutusCore.Annotation
 import PlutusCore.Pretty qualified as PLC
 import Types
-import UntypedPlutusCore qualified as UPLC
 import UntypedPlutusCore.Evaluation.Machine.Cek.Debug.Driver qualified as D
 import UntypedPlutusCore.Evaluation.Machine.Cek.Debug.Internal
 
@@ -20,6 +17,8 @@ import Brick.Types qualified as B
 import Brick.Widgets.Edit qualified as BE
 import Control.Concurrent.MVar
 import Control.Monad.State
+import Data.Coerce
+import Data.Set as S
 import Graphics.Vty qualified as Vty
 import Lens.Micro
 import Prettyprinter
@@ -74,11 +73,8 @@ handleDebuggerEvent driverMailbox bev@(B.VtyEvent ev) = do
             pure ()
         _ -> handleEditorEvent
 handleDebuggerEvent _driverMailbox (B.AppEvent (UpdateClientEvent cekState)) = do
-    let uplcHighlight = do
-            uplcSpan <- uplcAnn <$> case cekState of
-                Computing _ _ _ t -> Just (UPLC.termAnn t)
-                Returning _ ctx _ -> contextAnn ctx
-                _                 -> Nothing
+    let uplcHighlight :: Maybe HighlightSpan = do
+            uplcSpan <- uplcAnn <$> cekStateAnn cekState
             pure HighlightSpan
                 { _hcSLoc = B.Location (srcSpanSLine uplcSpan, srcSpanSCol uplcSpan),
                   -- The ending column of a `SrcSpan` is usually one more than the column of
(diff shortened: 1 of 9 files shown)


plutus-core/executables/debugger/Event.hs
```diff
@@ -74,38 +73,46 @@ handleDebuggerEvent driverMailbox bev@(B.VtyEvent ev) = do
...
+    let sourceHighlight :: Maybe HighlightSpan = do
+            txSpans <- txAnn <$> cekStateAnn cekState
+            -- FIXME: use some/all spans for highlighting, not just the first one
+            firstTxSpan <- S.lookupMin $ coerce txSpans
+            -- TODO: the HS_FILE supplied from the command line gets highlighted
+            -- The highlighting will not make sense or even break if the user provides
+            -- wrong HS_FILE or if the UPLC originated from multiple HS modules.
+            pure HighlightSpan
+                { _hcSLoc = B.Location (srcSpanSLine firstTxSpan, srcSpanSCol firstTxSpan),
+                  _hcELoc = Just $ B.Location (srcSpanELine firstTxSpan, srcSpanECol firstTxSpan)
```
zliu41, line 91: This should be `srcSpanECol firstTxSpan - 1`
GHC's `SrcSpan`'s ending column is one larger than the column of the last character.
[It is adjusted when the SrcSpan is pretty-printed](https://gitlab.haskell.org/ghc/ghc/-/blob/7eac2468a726f217dd97c5e2884f6b552e8ef11d/compiler/GHC/Types/SrcLoc.hs#L728).

zliu41 (approved):
Nice!

plutus-core/executables/debugger/Event.hs
```diff
@@ -74,38 +73,46 @@ handleDebuggerEvent driverMailbox bev@(B.VtyEvent ev) = do
...
-    modify' $ \st -> case cekState of
-        Computing{} ->
-            st & dsUplcHighlight .~ uplcHighlight
+    let sourceHighlight :: Maybe HighlightSpan = do
+            txSpans <- txAnn <$> cekStateAnn cekState
+            -- FIXME: use some/all spans for highlighting, not just the first one
+            firstTxSpan <- S.lookupMin $ coerce txSpans
+            -- TODO: the HS_FILE supplied from the command line gets highlighted
+            -- The highlighting will not make sense or even break if the user provides
+            -- wrong HS_FILE or if the UPLC originated from multiple HS modules.
```
michaelpj, line 88: `SrcSpan` contains a `FilePath`. Why aren't we just loading up the file from the `FilePath` when we highlight a `SrcSpan`?

plutus-core/executables/debugger/Main.hs
```diff
@@ -207,20 +181,65 @@ driverThread driverMailbox brickMailbox prog = do
...
-    UPLC.Program UPLC.NamedDeBruijn DefaultUni DefaultFun () ->
-    IO (UPLC.Program UPLC.Name DefaultUni DefaultFun ())
-unDeBruijnProgram p = do
-    either (fail . show) pure
-        . PLC.runQuote
-        . runExceptT @UPLC.FreeVariableError
-        $ traverseOf UPLC.progTerm UPLC.unDeBruijnTerm p
+-- | Read uplc code in a given format
+--
+--  Adaptation of `Common.getProgram`
```
michaelpj, line 192: I _really_ want to do that thing where we add "text envelopes" for serialized programs that tell us the type. This stuff is such a pain otherwise.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Zip.hs
```diff
@@ -0,0 +1,70 @@
...
+           (t ~ Term name uni fun, Eq (t ()), MonadFail m)
+         => (ann1 -> ann2 -> ann3)
+         -> t ann1
+         -> t ann2
+         -> m (t ann3)
+tzipWith f term1 term2 = do
+    -- first establish that terms are equal modulo annotations
+    -- slower this way but avoids some equality boilerplate for now
+    when (void term1 /= void term2) $
+       fail "Terms do not match."
```
michaelpj, line 42: Why does this help? Doesn't the catch-all pattern match below account for this regardless?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Zip.hs
```diff
@@ -0,0 +1,70 @@
...
+
+import Control.Monad (when)
+import PlutusPrelude
+import UntypedPlutusCore.Core.Instance.Eq ()
+import UntypedPlutusCore.Core.Type
+
+-- | Zip two programs using a combinator function for annotations.
+-- Fail if the input programs are not "equal" modulo annotations.
+pzipWith :: forall p name uni fun ann1 ann2 ann3 m.
+           (p ~ Program name uni fun, (Eq (Term name uni fun ())), MonadFail m)
```
michaelpj, line 18: Can we just never use `MonadFail`. IMO even `MonadError String` is better.

michaelpj (approved):

plutus-core/executables/debugger/Event.hs
bezirg: Thanks!

plutus-core/executables/debugger/Event.hs
bezirg: I discussed this before with Ziyang, and we leave it for another PR since it would not some filepath resolution checks. Currently we ignore these annotated filepaths and only use the filepath supplied at commandline.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Zip.hs
bezirg: I thnk there are some checks missing from this zip function, right? 
Checks missing:
- Alpha equivalence check in case t is `Term Name`.
- Eq uni and Eq fun check for builtins.

So I thought to stay on the safe side.

Btw, I should add a comment f input terms t1 and t2 is `Term Name` , the output is "left-biased" and will return the names of t1.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Zip.hs
bezirg: good call

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Zip.hs
michaelpj: I see. I'd be tempted to do an equality check on the builtins/constants/names, but I gues this is also okay? But write something down!

PR #5119, opened 2023-02-08
PLT-1402: Use SrcSpan in UPLC parser

zliu41:
instead of `SourcePos`.

PIR and TPLC parsers will be migrated separately.

Now the debugger can highlight an entire UPLC term, rather than just the first token.


plutus-core/plutus-core/src/PlutusCore/Annotation.hs
```diff
@@ -76,6 +76,8 @@ data SrcSpan = SrcSpan
     , srcSpanSCol  :: Int
     , srcSpanELine :: Int
     , srcSpanECol  :: Int
+    -- ^ The ending column is one more than the column of the last character (same as
```
michaelpj, line 79: What happens if the last character is the last token in the line? Does that mean that this will be the index of a non-existent character?

plutus-core/plutus-core/src/PlutusCore/Parser/ParserCommon.hs
```diff
@@ -99,6 +119,24 @@ wordPos ::
     T.Text -> Parser SourcePos
 wordPos w = lexeme $ try $ getSourcePos <* symbol w
 
+-- | Given a `Parser` that captures the ending position, supply the starting position
+-- to make it a `SrcSpan`.
+withSpan :: (SrcSpan -> a -> b) -> Parser (a, SourcePos) -> Parser b
```
michaelpj, line 124: So, this seemed a bit clumsy to me. I thought about it a bit, and indeed, `ParsecT` has an instance of `MonadFix`. So we can write this with `RecursiveDo`:
```
withSpan :: (SrcSpan -> Parser a) -> Parser a
withSpan' f = mdo
  start <- getSourcePos
  res <- f spn
  end <- getSourcePos
  let spn = toSrcSpan start end
  pure res

lamTerm :: Parser PTerm
lamTerm =
    withSpan
        (\sp -> inParens (UPLC.LamAbs sp <$> (symbol "lam" *> name) <*> term))
```
Note that it's using the normal `inParens` etc. I think this is very nice and clean and expresses exactly what it is that we want to do, while being minimally invasive to everything else (all `Parser`s always return things with `SrcSpan`s in them).

And yes, this works, I tried it. So long as the argument to `withSpan` is lazy, which it always is.

michaelpj (commented):
I had what I think is a pretty great idea.

plutus-core/plutus-core/src/PlutusCore/Parser/ParserCommon.hs
zliu41: But `type Parser = ParsecT ParserError T.Text (StateT ParserState Quote)`, and `Quote` doesn't have a `MonadFix` instance

Also the normal `inParens` consumes trailing whitespaces, so I assume you meant the version of `inParens` that doesn't.

zliu41: Heh, Ok, I didn't know `StateT` has `MonadFix` instances. I've never used `MonadFix` in my life :stuck_out_tongue:

plutus-core/plutus-core/src/PlutusCore/Annotation.hs
zliu41: Normally there's at least a line break character following the last character.
But technically the last character could be the line break, so I'll update the comment

plutus-core/plutus-core/src/PlutusCore/Parser/ParserCommon.hs
michaelpj: This is also the first time I've used it! But it does seem like a perfect usecase!

plutus-core/plutus-core/src/PlutusCore/Annotation.hs
zliu41: Updated the comment.

plutus-core/plutus-core/src/PlutusCore/Parser/ParserCommon.hs
zliu41: Updated. This is indeed a great use case of `mdo`!

michaelpj (approved):

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Parser.hs
```diff
@@ -31,34 +34,39 @@ import PlutusCore.Parser hiding (parseProgram, parseTerm, program)
 -- Parsers for UPLC terms
 
 -- | A parsable UPLC term.
-type PTerm = UPLC.Term PLC.Name PLC.DefaultUni PLC.DefaultFun SourcePos
+type PTerm = UPLC.Term PLC.Name PLC.DefaultUni PLC.DefaultFun SrcSpan
 
 conTerm :: Parser PTerm
-conTerm = inParens $ UPLC.Constant <$> wordPos "con" <*> constant
+conTerm = lexemeWithSpan $ \sp ->
```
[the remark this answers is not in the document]
zliu41: This is a case where it may make sense to just call the two versions `withSpan` and `withSpan'`, and make it clear what they do in the Haddock. I think the one that consumes whitespaces is more useful, so I'd call that one `withSpan`.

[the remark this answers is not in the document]
zliu41: I mean, I don't really have a good, short name for `withSpanAndNoTrailingWhitespace`.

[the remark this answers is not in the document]
michaelpj: Yeah me neither. Fine with giving them non-descriptive names, the "lexeme" bit just feels actively misleading.

PR #5121, opened 2023-02-09
Fix benchmark.yml

zeme-iohk:
I'm giving up on `workflow_dispatch`: you *still* need to push to master for changes to `benchmark.yml` to take effect.
`workflow_dispatch` lets you choose which branch to run on, but it will still use the `benchmark.yml` from master.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,13 +1,8 @@
+
 name: Benchmark
 on:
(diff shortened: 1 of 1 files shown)


michaelpj:
`workflow_dispatch` would still be useful for running the benchmarks on e.g. an already merged branch, I guess? but we've never found ourselves wanting that so simpler is probably better

PR #5126, opened 2023-02-09
[WIP] Add check-changelog.yml workflow

zeme-iohk:


.github/workflows/check-changelog.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          files: '**/changelog.d/**'
+
+      - name: Enforce New File or DISABLE_CHANGELOG_CHECK
+        uses: actions/github-script@v6
+        with:
+          script: | 
+            function shouldCheckChangelog() {
+              const prDescription = '${{ github.event.pull_request.body }}';
+              return !prDescription.includes("DISABLE_CHANGELOG_CHECK");
```
michaelpj, line 23: Maybe we should use a label for this? Also makes it easier to see elsewhere.

zeme-iohk: I'm not sure what labels are in this context

zeme-iohk: Or so we'd look for a label named "Requires Changelog"? or "Does not Require Changelog"?

michaelpj: I think we want to label the exceptional case, so we'd have a "No Changelog Required" label and the workflow would check for that when deciding whether to complain. We would also need to rerun the workflow if the labels change.

.gitignore
```diff
@@ -89,6 +89,7 @@ plutus-playground-server/playground.yaml
 plutus-pab/plutus-pab.yaml
 
 # Misc
+act-event.json
```
zeme-iohk, line 92: We need this for testing GH actions locally

zeme-iohk:
I've tested manually by adding & removing the label, and by committing and deleting a file in plutus-core/changelog.d

.github/workflows/pre-commit-check.yml
```diff
@@ -1,6 +1,9 @@
 name: Pre Commit Check
 on:
   pull_request:
+    types:
+      - opened
+      - synchronize
```
michaelpj, line 6: Do we need this?

michaelpj (approved):
LGTM.

michaelpj:
Oh, probably needs to be documented in the changelog section of CONTRIBUTING

.github/workflows/pre-commit-check.yml
zeme-iohk: I should have added a comment: with this, the "Pre Commit Check" will *not* be run when we label/unlabel the PR, which I trust is what we want

zeme-iohk:
> Oh, probably needs to be documented in the changelog section of CONTRIBUTING

Done

PR #5134, opened 2023-02-13
PLT-1402: Use `SrcSpan` in PIR and TPLC parsers

zliu41:
Now all 3 parsers use `SrcSpan`.

### DIFF
--- a/plutus-core/executables/src/PlutusCore/Executable/Common.hs
+++ b/plutus-core/executables/src/PlutusCore/Executable/Common.hs
@@ -35,7 +35,7 @@ import PlutusCore.StdLib.Data.Unit qualified as StdLib
 import UntypedPlutusCore qualified as UPLC
 import UntypedPlutusCore.Check.Uniques qualified as UPLC (checkProgram)
 import UntypedPlutusCore.Evaluation.Machine.Cek qualified as Cek
-import UntypedPlutusCore.Parser qualified as UPLC (parse, program, spanToPos)
+import UntypedPlutusCore.Parser qualified as UPLC (parse, program)
 
 import PlutusIR.Core.Type qualified as PIR
 import PlutusIR.Parser qualified as PIR (parse, program)
@@ -86,7 +86,7 @@ class ProgramLike p where
     -- | Parse a program.  The first argument (normally the file path) describes
     -- the input stream, the second is the program text.
     parseNamedProgram ::
-        String -> T.Text -> Either ParserErrorBundle (p PLC.SourcePos)
+        String -> T.Text -> Either ParserErrorBundle (p PLC.SrcSpan)
 
     -- | Check a program for unique names.
     -- Throws a @UniqueError@ when not all names are unique.
@@ -128,8 +128,7 @@ instance ProgramLike PlcProg where
 
 -- | Instance for UPLC program.
 instance ProgramLike UplcProg where
-    parseNamedProgram inputName =
-        fmap (fmap UPLC.spanToPos) . PLC.runQuoteT . UPLC.parse UPLC.program inputName
+    parseNamedProgram inputName = PLC.runQuoteT . UPLC.parse UPLC.program inputName
     checkUniques = UPLC.checkProgram (const True)
     serialiseProgramFlat nameType p =
         case nameType of
@@ -343,11 +342,11 @@ getInput StdInput         = T.getContents
 
 -- | For PLC and UPLC source programs. Read and parse and check the program for @UniqueError@'s.
 parseInput ::
-    (ProgramLike p, PLC.Rename (p PLC.SourcePos)) =>
+    (ProgramLike p, PLC.Rename (p PLC.SrcSpan)) =>
     -- | The source program
     Input ->
     -- | The output is either a UPLC or PLC program with annotation
-    IO (p PLC.SourcePos)
+    IO (p PLC.SrcSpan)
 parseInput inp = do
     contents <- getInput inp
     -- parse the program
@@ -363,7 +362,7 @@ parseInput inp = do
             let checked = through PlutusCore.Executable.Common.checkUniques renamed
             case checked of
                 -- pretty print the error
-                Left (err :: PLC.UniqueError PLC.SourcePos) ->
+                Left (err :: PLC.UniqueError PLC.SrcSpan) ->
                     error $ PP.render $ pretty err
                 Right _ -> pure p
 
@@ -430,18 +429,18 @@ loadUplcASTfromFlat flatMode inp = do
 getProgram ::
     ( ProgramLike p
     , Functor p
-    , PLC.Rename (p PLC.SourcePos)
+    , PLC.Rename (p PLC.SrcSpan)
     ) =>
     Format ->
     Input ->
-    IO (p PLC.SourcePos)
+    IO (p PLC.SrcSpan)
 getProgram fmt inp =
     case fmt of
         Textual -> parseInput inp
         Flat flatMode -> do
             prog <- loadASTfromFlat flatMode inp
             -- No source locations in Flat, so we have to make them up.
-            return $ PLC.topSourcePos <$ prog
+            return $ PLC.SrcSpan "top" 1 1 1 2 <$ prog
 
 ---------------- Serialise a program using Flat and write it to a given output ----------------
 
@@ -733,7 +732,7 @@ runPrintBuiltinSignatures = do
 
 runPrint :: PrintOptions -> IO ()
 runPrint (PrintOptions iospec mode) = do
-    parsed <- (parseInput (inputSpec iospec) :: IO (PlcProg PLC.SourcePos))
+    parsed <- (parseInput (inputSpec iospec) :: IO (PlcProg PLC.SrcSpan))
     let printed = show $ getPrintMethod mode parsed
     case outputSpec iospec of
         FileOutput path -> writeFile path printed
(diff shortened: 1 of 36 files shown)


plutus-core/plutus-core/src/PlutusCore/Annotation.hs
```diff
@@ -98,8 +99,10 @@ instance Show SrcSpan where
 instance Pretty SrcSpan where
     pretty = viaShow
 
+instance NFData SrcSpan
```
michaelpj, line 102: add to `deriving anyclass` on line 85

plutus-core/plutus-core/src/PlutusCore/Annotation.hs
```diff
@@ -98,8 +99,10 @@ instance Show SrcSpan where
...
+
 newtype SrcSpans = SrcSpans {unSrcSpans :: Set SrcSpan}
-    deriving newtype (Eq, Ord, Semigroup, Monoid, MonoFoldable)
+    deriving newtype (Eq, Ord, Semigroup, Monoid, MonoFoldable, NFData)
```
michaelpj, line 105: I'm actually not sure whether it's correct to newtype derive `NFData`? I guess it must be since it's representationally equal to the underlying type.

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
```diff
@@ -74,7 +74,7 @@ conList uniA = inBrackets $ constantOf uniA `sepBy` symbol ","
 
 -- | Parser for pairs.
 conPair :: DefaultUni (Esc a) -> DefaultUni (Esc b) -> Parser (a, b)
-conPair uniA uniB = inParens $ do
+conPair uniA uniB = inParensSpc $ do
```
michaelpj, line 77: Why is this different?

plutus-core/plutus-core/src/PlutusCore/Parser/ParserCommon.hs
```diff
@@ -144,25 +128,20 @@ toSrcSpan start end =
...
     y <- Lex.decimal
     void $ char '.'
     Version sp x y <$> Lex.decimal
 
 name :: Parser Name
-name = lexeme $ try $ do
+name = lexeme name'
+
+-- | Like `name` but does not consume trailing whitespaces.
+name' :: Parser Name
```
michaelpj, line 143: I wonder if this should be unticked and we should audit the cases where the other one is needed?

plutus-core/plutus-ir/src/PlutusIR/Parser.hs
```diff
@@ -29,109 +28,113 @@ import PlutusCore.Error (AsParserErrorBundle)
 import Text.Megaparsec hiding (ParseError, State, many, parse, some)
 
 -- | A parsable PIR pTerm.
-type PTerm = PIR.Term TyName Name PLC.DefaultUni PLC.DefaultFun SourcePos
+type PTerm = PIR.Term TyName Name PLC.DefaultUni PLC.DefaultFun SrcSpan
 
 recursivity :: Parser Recursivity
-recursivity = inParens $ (wordPos "rec" >> return Rec) <|> (wordPos "nonrec" >> return NonRec)
+recursivity = inParensSpc $ (symbol "rec" $> Rec) <|> (symbol "nonrec" $> NonRec)
```
michaelpj, line 34: I think we need a note about whitespace handling. Initially I thought all uses of `inParensSpc` must be wrong, because that meant that either a) the whitespace would get included in the span or b) we weren't including the parens in the span, which we should be. Now I see that it can work, but only if we are parsing something that doesn't have a span annotation.

TBH, I think I'd find it clearer if we just used an explicit `*> whitespace` instead. Or a careful note on `inParensSpc`.

Generally I think it's not totally clear what you're supposed to do about whitespace. I think the answer is "all parsers should consume trailing whitespace, but it should not be included in the source span". `withSpan` does this automatically, which is good. Perhaps we should just use `withSpan` for everything, and just ignore the span if we don't have a place for it?

plutus-core/plutus-ir/src/PlutusIR/Parser.hs
```diff
@@ -29,109 +28,113 @@ import PlutusCore.Error (AsParserErrorBundle)
...
-program :: Parser (Program TyName Name PLC.DefaultUni PLC.DefaultFun SourcePos)
-program = whitespace >> do
-    prog <- inParens $ do
-        p <- wordPos "program"
-        option () $ void version
-        Program p <$> pTerm
+program :: Parser (Program TyName Name PLC.DefaultUni PLC.DefaultFun SrcSpan)
+program = do
+    whitespace
+    prog <- withSpan $ \sp ->
```
michaelpj, line 134: For consistency, I wonder if we should pull this bit out as the program parser and then have a "top-level" parser that does the "leading whitespace and consumes all input" check.

plutus-core/testlib/PlutusCore/Test.hs
```diff
@@ -251,6 +254,12 @@ goldenUEvalProfile
     => String -> [a] -> TestNested
 goldenUEvalProfile name values = nestedGoldenVsDocM name $ pretty . view _2 <$> (rethrow $ runUPlcProfile values)
 
+initialSrcSpan :: FilePath -> SrcSpan
```
michaelpj, line 257: Haddock! The meaning of these is unclear, and where to use them. Should they be defined next to `SrcSpan`? Why not?

plutus-core/untyped-plutus-core/test/Generators.hs
```diff
@@ -80,6 +84,25 @@ propParser = testPropertyNamed "Parser" "parser" $ property $ do
             :: T.Text -> Either ParserErrorBundle (Program Name DefaultUni DefaultFun SrcSpan)
         parseProg = runQuoteT . parseProgram
 
+-- | The `SrcSpan` of a parsed `Term` should not including trailing whitespaces.
+propTermSrcSpan :: TestTree
+propTermSrcSpan = testPropertyNamed
+    "parser captures ending positions correctly"
+    "propTermSrcSpan"
+    . property
+    $ do
```
michaelpj, line 93: You could write this test generically as a test that takes a program generator, a pretty-printer, and a parser. Not sure if it's worth it.

michaelpj (approved):
The main thing is that I think the whitespace handling strategy is now confusing. I think we should pick a consistent approach and document it.

plutus-core/plutus-core/src/PlutusCore/Annotation.hs
zliu41: Yes, I think so, and I've never heard of anyone recommending against doing that.

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
zliu41: `inParensSpc` is the new `inParens`. Previously, `inParens` consumes whitespaces, while `inParens'` doesn't.

plutus-core/plutus-core/src/PlutusCore/Parser/ParserCommon.hs
zliu41: I think the other one (which consumes whitespaces) is more common. 
- If `name` is in the middle, we do need to consume whitespaces (e.g., `LamAbs ann name body`)
- If `name` is at the end, then we don't consume whitespaces (e.g., `Var ann name`).

We can remove the other one and add `<* whitespace` whenever whitespaces need to be consumed, but I don't think it reduces the chance of forgetting to consume whitespaces.

And yes, this is ugly either way.

plutus-core/plutus-ir/src/PlutusIR/Parser.hs
zliu41: > I think I'd find it clearer if we just used an explicit *> whitespace

I'm fine with that, but like I said above, I don't think it reduces the chance of doing the wrong thing. One could easily forget to add `*> whitespace`.

> Perhaps we should just use withSpan for everything, and just ignore the span if we don't have a place for it?

This would make it cleaner, but it goes against the recommendation, which is that `getSourcePos` is expensive, and should not be used more than needed.

The cleanest solution is to use a lexer like you've suggested before.

plutus-core/plutus-ir/src/PlutusIR/Parser.hs
zliu41: Sure.

plutus-core/testlib/PlutusCore/Test.hs
zliu41: This is just for testing.

plutus-core/untyped-plutus-core/test/Generators.hs
zliu41: Good point.

plutus-core/plutus-ir/src/PlutusIR/Parser.hs
zliu41: I added helper functions `leadingWhitespace` and `trailingWhitespace`, and replaced `inParensSpc` with `trailingWhitespace . inParens`.

plutus-core/untyped-plutus-core/test/Generators.hs
zliu41: It'd also need to take a `termAnn` function as the parameter since it is different for PIR/TPLC/UPLC. So it's probably not worth it - the amount of reused code is limited.

bezirg (approved):

plutus-core/executables/src/PlutusCore/Executable/Common.hs
```diff
@@ -430,18 +429,18 @@ loadUplcASTfromFlat flatMode inp = do
...
-    IO (p PLC.SourcePos)
+    IO (p PLC.SrcSpan)
 getProgram fmt inp =
     case fmt of
         Textual -> parseInput inp
         Flat flatMode -> do
             prog <- loadASTfromFlat flatMode inp
             -- No source locations in Flat, so we have to make them up.
-            return $ PLC.topSourcePos <$ prog
+            return $ PLC.SrcSpan "top" 1 1 1 2 <$ prog
```
bezirg, line 443: I would prefer if we kept this as a constant named topSrcSpan like we did before with topSourcePos

PR #5143, opened 2023-02-15
[Ledger API] Expose 'mkTermToEvaluate'

### LINKED ISSUE #5133: [plutus-ledger-api] Expose mkTermToEvaluate or fromSerialisedScript

As the Hydra team, we want to profile our fully applied plutus-tx scripts.

For this, we have cobbled together a way to get the actual scripts from transactions we create in [this PR](https://github.com/input-output-hk/hydra/pull/721/files#diff-d5c861d76b1d4c0f37308f18a620ccb49518112fed2bdb2e3f1a8923edde8384R180-R184).

The current approach uses a ledger function to collect the scripts and the `plutus-ledger-api` `mkTermToEvaluate` function to apply datum, redeemer and context arguments to the validator function. This function was exposed in the `plutus` version we use (`plutus-ledger-api-1.0.0.0`) via module `Plutus.ApiCommon`.

This function is not exposed anymore on master.

We would consider vendoring the contents of `mkTermToEvaluate`, but that would require `fromSerialisedScript` which is also not exposed.

If we do it wrong in the first place, any suggestions other than exposing and using these functions are welcome!

effectfully:
People find this helper useful, hence we should expose it.

Fixes #5133.

### DIFF
--- a/plutus-ledger-api/src/PlutusLedgerApi/Common.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/Common.hs
@@ -51,6 +51,7 @@ module PlutusLedgerApi.Common
     , EvaluationContext (..)
     , mkDynEvaluationContext
     , toMachineParameters
+    , mkTermToEvaluate
     ) where
 
 import PlutusCore.Data as PlutusCore (Data (..))
--- a/plutus-ledger-api/src/PlutusLedgerApi/Common.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/Common.hs
@@ -51,6 +51,9 @@ module PlutusLedgerApi.Common
     , EvaluationContext (..)
     , mkDynEvaluationContext
     , toMachineParameters
+    -- While not strictly used by the ledger, this is useful for people trying to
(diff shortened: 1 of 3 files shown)


ch1bo (approved):
Thanks. Are the export lists the "interface" which you agree on with us users? i.e. How can we ensure this is not unexported by someone else who would not remember the request in #5133?

michaelpj:
Great question. The primary user of `plutus-ledger-api` is the ledger. So I think we should probably add comments for things that we export that are not directly for the use of the ledger.

plutus-ledger-api/src/PlutusLedgerApi/Common.hs
```diff
@@ -51,6 +51,7 @@ module PlutusLedgerApi.Common
     , EvaluationContext (..)
     , mkDynEvaluationContext
     , toMachineParameters
+    , mkTermToEvaluate
```
michaelpj, line 54:
```suggestion
    -- While not strictly used by the ledger, this is useful for people trying to
    -- reconstruct the term evaluated by the ledger from the arguments, e.g.
    -- for profiling purposes.
    , mkTermToEvaluate
```

PR #5145, opened 2023-02-15
PLT-1546: Split out PIR into a sublib

michaelpj:
I felt inspired last night so I made some progress on PLT-178. This is the easiest and least controversial bit: splitting out PIR.

I also took the liberty to split out `satint` and to organize the cabal file a bit.


michaelpj:
Hmm, I believe the problem here is that Haddock and Hackage still don't work that well with re-exported modules. I'm not entirely sure what to do here. In the long run I don't see how we can reconcile:
1. Having sub-libs for each of our "languages"
2. Not requiring the use of public sub-libraries downstream

Point 2 requires us to have a meta-lib that bundles all the other ones, which is easily accomplished with `reexported-moudules`, and not easily accomplished otherwise. So I'm somewhat inclined to double down, and maybe just drop the hacky Haddock linking in the RTD site for now, which never got much use anyway.

michaelpj:
Main alternative that I can see: use the main library for UPLC and have TPLC and PIR as public sub-libraries that are _not_ re-exported and have to be depended on explicitly.

effectfully (approved):
Seems good, although I don't really know anything about public sublibraries.

PR #5149, opened 2023-02-16
Add Slack Notifications for script-evaluation-test Workflow

zeme-iohk:


.github/workflows/script-evaluation-test.yml
```diff
@@ -18,21 +18,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3.3.0
 
-      - name: Notify Slack
```
zeme-iohk, line 21: Ignore this, it was a dead-end attempt.

.github/workflows/slack-message-broker.yml
```diff
@@ -0,0 +1,49 @@
...
+            const emojy = { success: "✅", failure: "❌", cancelled: "✋" }[status];
+            const conclusion = "${{ github.event.workflow_job.conclusion }}";
+            const action = "${{ github.event.action }}";
+            const message = `${emojy} \`${name}\` *${action}* | *${status}* | *${conclusion}* 👉🏻 <${url}|view logs>`;
+            core.setOutput("message", message); 
+
+      - name: Notify Slack
+        uses: slackapi/slack-github-action@v1.23.0
+        with:
+          channel-id: my-private-channel
```
zeme-iohk, line 34: @michaelpj @zliu41 
Again, I must push to master before I can test this for real.
It works locally btw.
If you approve, I'll run `script-evaluation-test` and `benchmark` and I'll open a new PR changing `my-private-channel` to `plutus-core-dev` (unless you want to create a new channel to recevi these kinds of notifications?)

.github/workflows/slack-message-broker.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  slack-broker:
+    runs-on: [ubuntu-latest]
+
+    steps:
+
+      - name: Prepare Slack Message
+        uses: actions/github-script@v6
+        id: prepare-slack-message
+        if: contains(fromJSON('["script-evaluation-test", "benchmark"]'), github.event.workflow_job.name)
```
michaelpj: I think I'd mildly prefer it if the direction of causality went the other way around? Can we explicitly trigger this workflow at the end of the other workflows?

.github/workflows/slack-message-broker.yml
michaelpj: If we're going to post in `plutus-core-dev` then it needs to be for _exceptional_ events, i.e. failures. I'm unconvinced of the utility of notifying for successful runs: if you want to know about those, you can just go to the actions page on github.

michaelpj: Or at least, that's my interpretation of the point of such messages. I guess we could have a channel that gets logs for all runs, but I would never look at it (seems strictly worse than going on github).

zeme-iohk: THB this was originally meant just for the script-evaluation-test, which runs daily, and for that it would be nice to know when it has succeeded/failed, but also that is has started at all.
We can actually remove Benchmark from the list here

.github/workflows/slack-message-broker.yml
zeme-iohk: The line `if: contains(fromJSON('["script-evaluation-test", "benchmark"]'), github.event.workflow_job.name)` is invalid. I've changed this in the latest commit

.github/workflows/slack-message-broker.yml
michaelpj: So I guess my feeling is that for any workflow which doesn't report its status via a commit status or similar, it would be nice to have a failure notification in the channel. Because those are the things we're going to miss. For the benchmarking job I guess it's less of a problem now that it posts a comment with a link to the workflow run, but maybe it would still be useful to see failures.

michaelpj:
Okay, so now I'm confused: what _does_ trigger this workflow? :thinking: just everything?

zeme-iohk:
> Okay, so now I'm confused: what _does_ trigger this workflow? 🤔 just everything?

```
on:
  workflow_run:
    workflows: [Script Evaluation Test, Benchmark]
    types: [completed, requested, in_progress]
```

This workflow is triggered whenever any of these [Script Evaluation Test, Benchmark] workflows are either [completed, requested, in_progress].

> So I guess my feeling is that for any workflow which doesn't report its status via a commit status or similar, it would be nice to have a failure notification in the channel. 

Sure, that's even better. 
I'll add `Build and Deploy to Github Pages` (that the only other relevant workflow) to the list, and change the code so that only failures are reported.

michaelpj:
Ah right, I missed that. Do we need to document this anywhere? At least it would be good to have a comment in the workflow explaining what it does, how to trigger it, how to add more jobs etc.

PR #5150, opened 2023-02-16
PLT-1552: Better support for Plutus Core language versions

michaelpj:
See changelogs for pointers to some specific issues. At the moment nothing relies on the new version, and I'll add some tests to `plutus-ledger-api` for properly rejecting wrongly versioned programs etc. once there's actually a difference.

I went a bit beyond the brief in `plutus-ledger-api`, but once I was there I thought it was worth making it consistent with the new terminology in CIP-35.


plutus-core/plutus-core/src/PlutusCore/Version.hs
```diff
@@ -48,6 +52,25 @@ instance Ord Version where
   compare (Version major1 minor1 patch1) (Version major2 minor2 patch2) =
     compare major1 major2 <> compare minor1 minor2 <> compare patch1 patch2
 
--- | The default version of Plutus Core supported by this library.
-defaultVersion :: Version
-defaultVersion = Version 1 0 0
+-- | The first version of Plutus Core supported by this library.
+firstVersion :: Version
+firstVersion = plcVersion110
```
zliu41, line 57: Should be 100

plutus-core/plutus-core/src/PlutusCore/Version.hs
```diff
@@ -48,6 +52,25 @@ instance Ord Version where
...
+
+-- | Plutus Core version 1.1.0
+plcVersion110 :: Version
+plcVersion110 = Version 1 1 0
+
+-- | The latest version of Plutus Core supported by this library.
+latestVersion :: Version
+latestVersion = plcVersion110
+
+instance Bounded Version where
```
zliu41, line 71: What is this useful for? Not all versions between `minBound` and `maxBound` are valid.

plutus-core/changelog.d/20230216_173134_michael.peyton-jones_plc_versions.md
```diff
@@ -0,0 +1,9 @@
+### Added
+
+- New language version 1.1
+- PIR programs now have a version, which is optionally parsed (defaulting to the latest version)
+
+### Changed
+
+- `defaultVersion` renamed to `latestVersion`
+- `applyProgram` now merges annotations and takes the latest program version
```
zliu41, line 9: This may not be the right thing to do if the fun and the arg have different major versions. Can we require that they have the same major version?

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
```diff
@@ -79,49 +95,72 @@ builtinsIntroducedIn = Map.fromList [
...
-  -- Chang is protocolversion=8.0
-  ((PlutusV2, changPV), Set.fromList [
+  ((PlutusV2, valentinePV), Set.fromList [
           VerifyEcdsaSecp256k1Signature, VerifySchnorrSecp256k1Signature
           ])
   ]
 
-{-| Query the protocol version that a specific ledger Plutus version was first introduced in.
+{-| A map indicating which Plutus Core versions were introduced in which
+'ProtocolVersion' and 'PlutusLedgerLanguage'. Each version should appear at most once.
```
zliu41, line 107: Then perhaps the version should be the map key?

bezirg: We could do something better here for this map indeed.

I think a first step  is to define its keys in terms of  `languageIntroducedIn` to avoid mistakes on PVs?

bezirg: Another idea is to turn this into a 3-assoc. list with type [(PlutusLedgerLanguage,ProtocolVersion,Version)] and either on compile time or runtime have a test that the x,y,z parts are unique:

```
plcVersionsIntroducedIn = [ (PlutusV1, alonzoPv, plcVersion100)  , ...]
(x,y,z) = unzip3 plcVersionsIntroducedIn
test that isUnique x , y , z 
isUnique l =  length l == nubOrd l (or some other more efficient impl.)
```

bezirg: We could also delve into some custom `IsList`  uniqueness checker that blows up with a runtime test? I am not sure how that would look like

bezirg (changes requested):
I don't like the new renaming; other than that, it looks good to me

plutus-core/plutus-ir/src/PlutusIR/Parser.hs
```diff
@@ -128,14 +129,15 @@ pTerm = leadingWhitespace go
...
 program :: Parser (Program TyName Name PLC.DefaultUni PLC.DefaultFun SrcSpan)
 program = leadingWhitespace go
   where
     go = do
         prog <- withSpan $ \sp ->
-            inParens $ Program sp <$> (symbol "program" *> option () (void version) *> pTerm)
+            -- PIR programs do have a version, but we don't require it: if it's missing,
+            -- we assume the latest version
+            inParens $
+              Program sp <$> (symbol "program" *> option PLC.latestVersion version) <*> pTerm
```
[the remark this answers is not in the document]
michaelpj: Yeah, I guess we could make it mandatory. Mostly I thought it would be annoying to change all the test files, but I can do that.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore.hs
```diff
@@ -23,15 +23,15 @@ import PlutusCore.Name as Export
...
     Program name uni fun a ->
     Program name uni fun a ->
     Program name uni fun a
-applyProgram (Program a1 _ t1) (Program a2 _ t2) =
-    Program (a1 <> a2) PLC.defaultVersion (Apply (termAnn t1 <> termAnn t2) t1 t2)
+applyProgram (Program a1 v1 t1) (Program a2 v2 t2) =
+    Program (a1 <> a2) (max v1 v2) (Apply (termAnn t1 <> termAnn t2) t1 t2)
 
-{- | DON'T USE, we'll be getting rid of `defaultVersion`.
+{- | DON'T USE, we'll be getting rid of `latestVersion`.
```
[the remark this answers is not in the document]
michaelpj: I'm not sure. I think `latestVersion` has a bit more of a meaning than `defaultVersion`, so maybe it's okay. I do think it's the case that it shouldn't be used except at the "edges", i.e. we shouldn't provide functions that sneakily use `latestVersion` under the hood. So this function should maybe still die.

plutus-core/plutus-core/src/PlutusCore/Version.hs
michaelpj: I also think it's a bit suspicious now that I think about it. I think we want to say something like "you can enumerate the known versions of PLC". I think better versions might be:
1. `KnownVersion` newtype with `Enum` and `Bounded` instances
2. A `knownVersions :: Set Version` value

thoughts?

plutus-core/changelog.d/20230216_173134_michael.peyton-jones_plc_versions.md
michaelpj: And otherwise do what? Give up? I think so long as it's documented this is fairly reasonable behaviour? (Note to self: document it)

plutus-ledger-api/src/PlutusLedgerApi/Common/ProtocolVersions.hs
```diff
@@ -53,6 +54,16 @@ alonzoPV = ProtocolVersion 5 0
...
+
+-- | This is a placeholder for when we don't yet know what protocol
+-- version will be used for something. It's a very high protocol
+-- version that should never appear in reality.
+--
+-- We should not assign names to future protocol versions until it's
+-- confirmed that they are correct, otherwise we could accidentally
+-- associate something with the wrong protocol version.
+futurePV :: ProtocolVersion
+futurePV = ProtocolVersion maxBound 0
```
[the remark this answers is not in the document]
michaelpj: I guess because this isn't really a "top bound" for protocol versions, it's a fake protocol version. Same thing as with `Version`, really. Perhaps again we just want a `knownProtocolVersions :: Set ProtocolVersion` value.

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
```diff
@@ -79,49 +95,72 @@ builtinsIntroducedIn = Map.fromList [
...
     -- OPTIMIZE: could be done faster using takeWhile
-    ledgerVersionToSet :: LedgerPlutusVersion -> Set.Set LedgerPlutusVersion
+    ledgerVersionToSet :: PlutusLedgerLanguage -> Set.Set PlutusLedgerLanguage
     ledgerVersionToSet lv
-        | languageIntroducedIn lv <= searchPv = Set.singleton lv
+        | ledgerLanguageIntroducedIn lv <= searchPv = Set.singleton lv
         | otherwise = mempty
 
-{-| Which builtin functions are available in the given 'ProtocolVersion'?
+{-| Which Plutus language versions are available in the given 'PlutusLedgerLanguage'
```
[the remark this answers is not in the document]
michaelpj: Nobody commented on my CIP changes :) Have a read: https://cips.cardano.org/cips/cip35/

It's the least confusing thing I could come up with.

plutus-ledger-api/testlib/PlutusLedgerApi/Test/Examples.hs
```diff
@@ -19,7 +20,7 @@ It seems better therefore to avoid depending on Plutus Tx in any "core" projects
 
 -- | Creates a script which has N arguments, and always succeeds.
 alwaysSucceedingNAryFunction :: Natural -> SerialisedScript
-alwaysSucceedingNAryFunction n = serialiseUPLC $ UPLC.Program () PLC.defaultVersion (body n)
+alwaysSucceedingNAryFunction n = serialiseUPLC $ UPLC.Program () PLC.plcVersion100 (body n)
```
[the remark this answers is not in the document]
michaelpj: Why not just be specific: this is a specific program that we're testing, no?

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
```diff
@@ -79,49 +95,72 @@ builtinsIntroducedIn = Map.fromList [
...
+'ProtocolVersion' and 'PlutusLedgerLanguage'. Each version should appear at most once.
+
+This __must__ be updated when new versions are added.
+See Note [New builtins/language versions and protocol versions]
+-}
+plcVersionsIntroducedIn :: Map.Map (PlutusLedgerLanguage, ProtocolVersion) (Set.Set Version)
+plcVersionsIntroducedIn = Map.fromList [
+  ((PlutusV1, alonzoPV), Set.fromList [ plcVersion100 ]),
+  ((PlutusV3, futurePV), Set.fromList [ plcVersion110 ])
+  ]
```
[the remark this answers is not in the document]
michaelpj: TBH it was just me blindly mirroring the stuff for builtins. I think keeping them similar is somewhat helpful?

plutus-core/plutus-core/src/PlutusCore/Version.hs
michaelpj: I've added `knownVersions`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore.hs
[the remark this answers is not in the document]
michaelpj: I killed `mkDefaultProgram`

plutus-ledger-api/src/PlutusLedgerApi/Common/ProtocolVersions.hs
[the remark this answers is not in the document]
michaelpj: I put in `knownPVs`

plutus-core/changelog.d/20230216_173134_michael.peyton-jones_plc_versions.md
zliu41: Why not. A quick-and-dirty `error "not allowed"` is better than allowing it and potentially leading to undefined behavior.

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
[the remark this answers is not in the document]
zliu41: "plutus core version" is good and clear. I'm indifferent to "ledger language version" vs. "plutus ledger version". Not a huge fan of either, but it is at least good to mention "ledger" - better than "plutus version" for sure!

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
zliu41: I mean we could simply reverse the key and the value for this map, and lookup by plutus core version.

plutus-core/changelog.d/20230216_173134_michael.peyton-jones_plc_versions.md
michaelpj: I'd rather make it return a `Maybe`. I think this could easily get used in production code paths. I'll try that.

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
michaelpj: It's the same thing as for `builtinsIntroducedIn`. I'd rather keep them consistent and fix them together, if we can!

michaelpj:
@bezirg have you taken a look at CIP-35? What do you think of the naming?

plutus-core/executables/uplc/Main.hs
```diff
@@ -197,7 +198,7 @@ runApply (ApplyOptions inputfiles ifmt outp ofmt mode) = do
   let appliedScript =
         case void <$> scripts of
           []          -> errorWithoutStackTrace "No input files"
-          progAndargs -> foldl1 UPLC.applyProgram progAndargs
+          progAndargs -> foldl1 (fromJust .* UPLC.applyProgram) progAndargs
```
bezirg, line 201: I didn't know about this operator.

What I do instead of this operator is:

```
(.*) f g = (f .) . g
(.**) f g = ( (f .) .) . g
...etc
```

plutus-ledger-api/changelog.d/20230216_173032_michael.peyton-jones_plc_versions.md
```diff
@@ -0,0 +1,7 @@
+### Added
+
+- Support for multiple Plutus Core language versions.
+
+### Changed
+
+- The naming around "Plutus langauge versions" changed to talk about "Plutus ledger languages" following CIP-35.
```
bezirg, line 7: Typo: language

plutus-core/changelog.d/20230216_173134_michael.peyton-jones_plc_versions.md
```diff
@@ -0,0 +1,9 @@
+### Added
+
+- New language version 1.1
+- PIR programs now have a version, which is optionally parsed (defaulting to the latest version)
+
+### Changed
+
+- `defaultVersion` renamed to `latestVersion`
+- `applyProgram` now merges annotations and requires matchin program versions
```
bezirg, line 9: typo: matching

plutus-tx/src/PlutusTx/Code.hs
```diff
@@ -55,11 +55,14 @@ data CompiledCodeIn uni fun a =
...
+-- | Apply a compiled function to a compiled argument. Will fail if the versions don't match.
 applyCode
     :: (PLC.Closed uni, uni `PLC.Everywhere` Flat, Flat fun)
-    => CompiledCodeIn uni fun (a -> b) -> CompiledCodeIn uni fun a -> CompiledCodeIn uni fun b
-applyCode fun arg = DeserializedCode (UPLC.applyProgram (getPlc fun) (getPlc arg)) (PIR.applyProgram <$> getPir fun <*> getPir arg) (getCovIdx fun <> getCovIdx arg)
+    => CompiledCodeIn uni fun (a -> b) -> CompiledCodeIn uni fun a -> Maybe (CompiledCodeIn uni fun b)
+applyCode fun arg = do
+  uplc <- UPLC.applyProgram (getPlc fun) (getPlc arg)
+  pir <- PIR.applyProgram <$> getPir fun <*> getPir arg
+  pure $ DeserializedCode uplc pir (getCovIdx fun <> getCovIdx arg)
```
bezirg, line 65: applyCode is often used with liftCode, so we should try to keep the same interface with liftCode / safeLiftCode.

So I suggest that applyCode remains with the old type signature:

```
applyCode :: code (a -> b) -> code a -> code b
applyCode = unsafely $ safeApplyCode
```

and we introduce the `safeApplyCode`:

```
safeApplyCode
    :: (PLC.Closed uni, uni `PLC.Everywhere` Flat, Flat fun, MonadError e m)
    => CompiledCodeIn uni fun (a -> b) -> CompiledCodeIn uni fun a -> m (CompiledCodeIn uni fun b)
safeApplyCode fun arg = do
  uplc <- maybe throwSth pure $ UPLC.applyProgram (getPlc fun) (getPlc arg)
  pir <- maybe throwSth pure $ PIR.applyProgram <$> getPir fun <*> getPir arg
  pure $ DeserializedCode uplc pir (getCovIdx fun <> getCovIdx arg)
```

bezirg:
> @bezirg have you taken a look at CIP-35? What do you think of the naming?

I did yesterday. What CIP-35 says is correct/solid; it could be explained a bit better, but I should have provided some feedback on it earlier. Sorry for that

bezirg (approved):

plutus-tx/src/PlutusTx/Code.hs
michaelpj: Oh wait, you mean just `applyCode`. Yes, that probably is worth having the extra function, I'll add it.

zliu41: I agree with @bezirg on keeping the same interface, though I like `unsafeFoo/foo` slightly better than `foo/safeFoo`. The former is much more common in Haskell.

zliu41 (approved):
Looks good!

PR #5152, opened 2023-02-17
Version 1.2.0.0

zliu41:
We haven't made a release for a while. We've split PIR into a separate library, updated parser, among other things, and started enforcing changelogs. So it is time for a release.

Since there's no major change to the API, I'm bumping the second component of the major version.

The release does not include `word-array` or `prettyprinter-configurable` since they have not been modified.


zliu41:
Closing this; we need to update the version bound for cardano-crypto-class, and also we should release from an earlier commit (the point where we started enforcing changelogs)

PR #5159, opened 2023-02-22
Allow newer `cardano-crypto-class`. Bump up `plutus-core-1.1.1.1`

lehins:
Not much has changed in `cardano-crypto-class` since the last release. A fix in a function name required a major bump, but it does not affect plutus-core.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-core/plutus-core.cabal
+++ b/plutus-core/plutus-core.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               plutus-core
-version:            1.1.1.0
+version:            1.1.1.1
 license:            Apache-2.0
 license-files:
   LICENSE
@@ -245,7 +245,7 @@ library
     , bimap
     , bytestring
     , cardano-crypto
-    , cardano-crypto-class        ^>=2.0.0.1
+    , cardano-crypto-class        >=2.0.0.1 && <2.2
     , cassava
     , cborg
     , composition-prelude         >=1.1.0.1


plutus-core/plutus-core.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               plutus-core
-version:            1.1.1.0
+version:            1.1.1.1
```
zliu41, line 3: Please change this back. We cannot make a patch release from master. Also, we have a script for this and don't change the version number manually.

plutus-core/plutus-core.cabal
```diff
@@ -245,7 +245,7 @@ library
     , bimap
     , bytestring
     , cardano-crypto
-    , cardano-crypto-class        ^>=2.0.0.1
+    , cardano-crypto-class        >=2.0.0.1 && <2.2
```
zliu41, line 248: Did you verify that 2.1 works with plutus? If you don't change the index-state in cabal.project then I think the CI still uses 2.0.0.1.

zliu41 (changes requested):

plutus-core/plutus-core.cabal
lehins: I did verify that it works in another repo (by using allow-newer)

plutus-core/plutus-core.cabal
lehins: > We cannot make a patch release from master.

Why not?

> we have a script for this and don't change the version number manually.

Cool. I'll let you take care of it then

zliu41: Our release process suggets tagging major/minor releases (x.y.z.0) from master, while patch releases (x.y.z.w) by backporting to x.y.z.w-1.

zliu41: There has been a lot of changes since 1.1.1.0, some of which are breaking, so if we make a release from master it should be at least 1.2.0.0.

lehins: > There has been a lot of changes since 1.1.1.0,

That makes total sense. I am not sure about your dev/version/release process, but it might be a good idea to bump the version as soon as breaking change is introduced, this way it is always possible to release from master  and it is clear that current version is way ahead of the latest released one.

Feel free to close this PR. Revision takes care of this issue for me! Thank you

zliu41: Thanks for the suggestion. We are still experimenting on the release process.

By the way, we have a CI job that enforces adding to changelogs, and that's pretty useful - if the change logs are complete then at least it's easier to tell whether there are breaking changes when it's time to do a release.

zliu41: Ok, closing the PR as you suggested.

PR #5160, opened 2023-02-22
PLT-1446: Add function to track the type and term lambda abstraction order of a term in inliner

thealmarty:
This will be used in the call-site inlining of fully applied functions. The notes in `CallSiteInline.hs` is not finalized. It will be further polished/finalized along with the implementation work, which will be in a follow up PR.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -127,3 +140,36 @@ counted out already so we should not immediately encounter those in `VBody`.
...
+    deriving stock (Eq, Show)
+
+instance Pretty TermOrType where
+  pretty = viaShow
+
+-- | Counts the type and term lambdas in the RHS of a binding and returns an ordered list
+countLam ::
+    Term tyname name uni fun a -- ^ the RHS of the let binding
+    -> TermOrTypeOrder
+countLam = countLamInner []
```
zliu41, line 159: This is a left fold, which is inefficient: you need to append to the list (`lamStack <> [MkTerm])`, and the first element of the list won't be available until the whole function is done.

A right fold is both shorter and more efficient:

```haskell
countLam = \case
  LamAbs _ _ _ body -> MkTerm : countLam body
  TyAbs _ _ _ body -> MkType : countLam body
  _ -> countLam body

File: plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs [hunk 2]
```diff
@@ -127,3 +140,36 @@ counted out already so we should not immediately encounter those in `VBody`.
...
+    where
+      countLamInner ::
+        TermOrTypeOrder
+        -> Term tyname name uni fun a -- ^ The rhs term that is being counted.
+        -> TermOrTypeOrder
+      countLamInner lamStack (LamAbs _a _n _ty body) =
+        -- If the term is a term lambda abstraction, add it to the list, and
```
zliu41, line 166: This is quite obvious, I don't think a comment is needed.

plutus-core/plutus-ir/test/TransformSpec.hs
```diff
@@ -209,6 +211,27 @@ inline =
             , "letOverAppMulti" -- multiple occurrences of an over-application of a function
             ]
 
+countLamTest :: TestNested
+countLamTest = testNested "countLamTest" $
+        map
+            (goldenPir (countLam . runQuote . PLC.rename) pTerm)
+            [ "var" -- from inline tests, testing let terms
```
zliu41, line 218: A bit too many tests for something as simple as `countLam`, since you've already written them it's fine to keep them.

michaelpj:
A bit of name bikeshedding: I really think we should use some naming related to "arity". "Arity" is a very well-established word for "how many arguments a function takes". What we've got here is a generalization of arity to include arguments of different types. In fact, I think it wouldn't be that confusing to just call it "arity".

So my vote would go for:
```

data Arity = Arity [TermOrType]

computeArity :: Term -> Arity
```

Are you planning to do the analysis pass that computes `termArities :: Term -> Map Name Arity` later?

thealmarty:
> A bit of name bikeshedding: I really think we should use some naming related to "arity". "Arity" is a very well-established word for "how many arguments a function takes". What we've got here is a generalization of arity to include arguments of different types. In fact, I think it wouldn't be that confusing to just call it "arity".
> 
> So my vote would go for:
> 
> ```
> 
> data Arity = Arity [TermOrType]
> 
> computeArity :: Term -> Arity
> ```

Sure! Nicer to read like this! I'll open a PR just for this change so it's not mixed in the next PR.
> 
> Are you planning to do the analysis pass that computes `termArities :: Term -> Map Name Arity` later?

Yes, in the next PR where I will implement the actual inlining.

PR #5168, opened 2023-02-25
PLT-1539: Add `ToList` class and rewrite some functions in `PlutusTx.Foldable`

zliu41:
With `ToList`, different types can provide different `toList` implementations which can be much cheaper than the default implementation.

Several `Foldable` functions related to `Bool` are rewritten to use `toList`. This makes them lazy and cheaper. Even when the entire list needs to traversed, it is still cheaper (e.g., see the `anyExpensive` test case).

However, it is more expensive than before if the list is empty. I think it is likely because when the list is empty, some overheads in the current implementation (e.g., from the `Monoid` dictionary) don't come into play. But as soon as the list has 1 element the new implementation is cheaper, so seems worth it.


effectfully (commented):
You're really cheating there. You give a fast implementation for lists (which is only fast due to you being lucky enough that `||` is getting inlined) and then insist that any `ToList` container needs to go through `toList`, but the _only_ container that doesn't require processing of the whole spine in its `ToList` instance is the type of lists itself! (Well, I guess list-based stuff like `NonEmpty` would fall into this category too)

So in effect you have an implementation of `any` that only properly short-circuits on lists and then you test it only on lists.

`any` on something like a tree would traverse the entire spine of the tree. Although this version is still much better than what we used to have, since at least you only run the predicate until it succeeds, while the previous one would run on every single element.

I don't understand why we need the `foldMap`-based `Foldable` class, if we have `ToList`. Those two are equal in power and `foldMap` is hardly useful in Plutus where

1. there's much less reason to use right folds in the first place, since you're not getting any short-circuiting with them due to the language being strict
2. even if you do need a right fold, you probably still don't want to carry that stupid `Monoid` dictionary around. Although it seems to me that the reason why we carry it around in the first place is that our implementation for `foldMap` is inefficient.

Given 2, could you please create a separate PR replaing the current `FoldMap []` instance with

```haskell

instance Foldable [] where
    {-# INLINABLE foldMap #-}
    foldMap :: forall m a. Monoid m => (a -> m) -> [a] -> m
    foldMap f = go where
        go :: [a] -> m
        go []     = mempty
        go (x:xs) = f x <> go xs
```

? Just to see how much that affects the budget. It's obviously not going to be anywhere near the numbers we're seeing in this PR, because it won't make anything short-circuit, but I'm curious how much performance we're leaving on the table by not optimizing things.

effectfully:
Also, could you please generate "readable" output in addition to / instead of the S-expressions abomination? I don't feel like passing a reverse Turing test each time I need to look into the output of the compiler.

effectfully:
> It's obviously not going to be anywhere near the numbers we're seeing in this PR, because it won't make anything short-circuit

Actually, I take that back. If `foldMap` gets inlined like it should as well as the function it gets supplied like it should too, you will get short-circuiting and the same or nearly the same results.

zliu41:
> could you please create a separate PR replaing the current FoldMap [] instance with

This does lead to improvements across the board, sometimes by a large amount, which surprised me. What a low hanging fruit! However, it alone isn't enough to make it short circuit, see https://github.com/input-output-hk/plutus/commit/3ab30055a007a72fa146136e5182a5204bc7a5e4

> If foldMap gets inlined like it should as well as the function it gets supplied like it should too, you will get short-circuiting and the same or nearly the same results.

How do you make it happen though? Presumably not by changing `INLINEABLE` to `INLINE` and running the GHC inliner? Is there a different way of writing `foldMap` to make it more inlineable?

> could you please generate "readable" output in addition to / instead of the S-expressions abomination?

Interesting. I don't think I've seen a readable PIR before, and it is indeed much more readable. We should use readable PIRs at least for `Budget.Spec`.

effectfully:
> However, it alone isn't enough to make it short circuit, see https://github.com/input-output-hk/plutus/commit/3ab30055a007a72fa146136e5182a5204bc7a5e4

Bad GHC, bad!

> How do you make it happen though? Presumably not by changing `INLINEABLE` to `INLINE` and running the GHC inliner? Is there a different way of writing `foldMap` to make it more inlineable?

`foldMap` is inlined just fine, it's the `Monoid` stuff that is getting in the way. I've tried various stuff to convince GHC to inline the monoid stuff properly, but for some reason it just refuses to do that, which is kinda puzzling to me. 

Anyway, we should try is to remove `foldMap` from `Foldable` and base the class on `foldr` instead. Fighting inlining is so much easier when you don't need to deal with constraints. Plus, we can probably benefit from deforestation more if we use `foldr`, although I don't know if our `foldr` is suitable for that.

zliu41:
> it's the Monoid stuff that is getting in the way.

Right. The gist of the problem is, in pseudo-code:

```haskell
let (<>) :: Bool -> Bool -> Bool
    (<>) a b = if a then delay True else delay b
    go x xs = (1 < x) <> go xs
```

If `<>` could be inlined by GHC, then `go xs` would be hidden behind the `delay`, and we'd have short-circuiting.

> base the class on foldr instead

Even if that works, the inconsistency between `foldr` and `foldMap` is still a problem. A few thoughts on fixing `foldMap`:

- Mark `<>` `INLINE` and run GHC inliner (presumably you've tried that)
- Can we modify the `foldMap` implementation to make it lazier, say by using one of the tricks in the Augustsson post you shared.

effectfully:
> Even if that works, the inconsistency between `foldr` and `foldMap` is still a problem. 

What inconsistency? `foldr` (possibly) inlining well and `foldMap` not inlining well? You won't get any consistency with inlining, forget about it right from the start.

In general, people did end up caring about efficiency of smart contracts => people needs tools for controlling efficiency => Haskell doesn't really provide those (apart from those that we can't make use of like unboxing) => we at least shouldn't use features that are really hard to control. I like type classes as much as everyone else, but shoving so many of them right into the Plutus Tx Prelude isn't gonna give the user efficient code no matter how much we try to optimize them.

Best we can do in my opinion is using monomorphic functions as much as possible and only relying on inlining when absolutely necessary (like with `||` and `&&`).

> run GHC inliner (presumably you've tried that)

I thought it would run without any interventions? I didn't run it myself purposefully.

zliu41:
> Best we can do in my opinion is using monomorphic functions as much as possible

That's the idea of `ToList` - so that everything becomes list :slightly_smiling_face: We absolutely want `any` to short-circuit, but the overhead of converting things that are not lists to lists is indeed unsatisfactory. 

Alternatively we can put `any` etc., all in single-method classes, so every type can implement its monomorphic version. But that's too ugly.

Again the bottom line is we must short-circuit. I'll try `foldr` and if it does the job then we'll go with that.

zliu41:
```haskell
{-# INLINE any #-}
any p = foldr go False
  where
    {-# INLINE go #-}
    go a acc = if p a then True else acc
```

This _does_ short-circuit, but only if we run the GHC inliner upfront (`sm_inline = True`) so that `go` gets inlined. Although it works in this instance, in general this approach is questionable. I think turning on the GHC inliner is a good thing, but in general, running the GHC inliner once upfront, on the entire `CoreExpr`, may not be enough, and we may need to keep calling `GHC.simplExpr` on subexpressions (I also tried doing the similar thing with `foldMap` but it didn't work). Also, turning on the GHC inliner may increase the cost in some cases but I think that's acceptable.

michaelpj:
I'm sceptical of this for the reasons that Roman mentions. In general, I'm not sure we can get good performance here in a principled way without providing a lot of manually specialized functions that we check behave the way that we want. And I think _anything_ that relies on proper inlining is a fool's errand.

So things that might actually help:
1. Write a specializer.
2. Make stuff less typeclass-polymorphic, probably at least kill `foldMap`, or don't implement everything in terms of it. 

> This does short-circuit, but only if we run the GHC inliner upfront (sm_inline = True) so that any and go get inlined. 

Something I've been meaning to do for a while, now we have the "must inline" annotation that the plugin can supply: if we process a binding that is annotated in Haskell with `INLINE`, then we should put our own "must inline" annotation on it. Then you wouldn't need to rely on the GHC inliner like this and you could get a bit better control over inlining.

---------

(Note: not a good idea, leaving in because I already wrote it)

A completely different idea. What about having an explicit lazy list type?

```haskell
data Delayed a = Delayed (() -> a)

-- must be inlined otherwise it won't work
{-# INLINE delay #-} 
delay :: a -> Delayed a
delay x = \_ -> x

force :: Delayed a -> a
force (Delayed f) = f ()

data LazyList a = Nil | a <: Delayed (LazyList a)

listToLazyList :: [a] -> LazyList a
listToLazyList [] = Nil
-- conversion of the tail is delayed
listToLazyList (x:xs) = x <: delay (listToLazyList xs)

any :: (a -> Bool) -> [a] -> Bool
any f xs = any' f (listToLazyList xs)

any' :: (a -> Bool) -> LazyList a -> Bool
any' f Nil = False
any' f (x:xs) = if f x then True else any f (force xs)
```

Okay, so I thought this would be nice and let us make the delaying part of the data structure, but it's basically the same as writing a short-circuiting `any` directly, and you can't write a short-circuiting `foldMap` this way without the monoid function inlining, so it's not that great.

PR #5171, opened 2023-02-26
Use unique-less readable PIR for Budget.Spec

zliu41:
The golden PIR files for `Budget.Spec` often need to be inspected to understand why a change to the compiler leads to different costs. Using the readable version makes this much easier.

Adding the uniques would be helpful for disambiguation (although it can also be distracting), but it makes it harder to diff two PIRs since the uniques can be totally different. So it doesn't print the uniques.

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Core/Instance/Pretty/Readable.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Core/Instance/Pretty/Readable.hs
@@ -12,6 +12,7 @@
 {-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
 module PlutusIR.Core.Instance.Pretty.Readable
   ( prettyPirReadable
+  , prettyPirReadableNoUnique
   , PrettyPir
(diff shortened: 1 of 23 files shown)


effectfully (approved):
MUCH better!

michaelpj (approved):
I wonder if we should use this throughout?

zliu41:
> I wonder if we should use this throughout?

No objection!

kwxm:
> So it doesn't print the uniques.

Doesn't that introduce a risk that two different things may end up looking the same?  Presumably that's unlikely but possible.x

zliu41:
> Doesn't that introduce a risk that two different things may end up looking the same? Presumably that's unlikely but possible.

The whole PIR programs look the same but are actually different? That seems very unlikely. If they look the same and also have the same cost then I don't think we need to care. If they look the same but have different costs, then we can debug further by adding the uniques to the output.

zliu41:
And we didn't print the uniques anyway - this is just a more readable version.

PR #5173, opened 2023-02-26
PLT-1539: Add Note [Haskell.Monad]

zliu41:


plutus-tx/src/PlutusTx/Foldable.hs
```diff
@@ -225,6 +228,20 @@ find p = coerce . foldMap (\ x -> First (if p x then Just x else Nothing))
...
+{- Note [Haskell.Monad]
+There is no `PlutusTx.Monad`. Since do-notations use `Haskell.Monad` (without
+`RebindableSyntax` anyway), it won't be very useful to have `PlutusTx.Monad`.
+
+`Haskell.Monad` instances for `Maybe`, `Either` and `[]` (and likely many other types)
+all work fine since the unfoldings of their `>>=` methods don't contain anything the
+plugin can't handle. This is of course not true in general, e.g., the plugin won't be
+able to handle a (hypothetical) `Haskell.Monad` instance whose `>>=` definition contains
+`GHC.Num.(+)` (until we make the plugin support functions like `GHC.Num.(+)`) or is
+not inlinable.
```
effectfully, line 245: This doesn't make any sense to me. We have this entirely separate hierarchy of type classes and then stop short of `Monad`, because of some concerns regarding `do`, which are easily resolvable with [`QualifiedDo`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/qualified_do.html). I'm OK with using Haskell's type classes when they're magical like `Coercible`, but having our own `Applicative` and using Haskell's `Monad` is just extra weird.

plutus-tx/src/PlutusTx/Foldable.hs
```diff
@@ -225,6 +228,20 @@ find p = coerce . foldMap (\ x -> First (if p x then Just x else Nothing))
 -- | Plutus Tx version of 'Data.Foldable.mapM_'.
 {-# INLINABLE mapM_ #-}
 mapM_ :: (Foldable t, Haskell.Monad m) => (a -> m b) -> t a -> m ()
+-- See Note [Haskell.Monad]
 mapM_ f = foldr c (Haskell.return ())
   where c x k = f x Haskell.>> k
         {-# INLINE c #-}
```
effectfully, line 234: What's even the point of having this function? It's a bad version of `traverse_`.

plutus-tx/src/PlutusTx/Foldable.hs
```diff
@@ -157,6 +159,7 @@ traverse_ f = foldr c (pure ())
 
 -- | Plutus Tx version of 'Data.Foldable.sequence_'.
 sequence_ :: (Foldable t, Haskell.Monad m) => t (m a) -> m ()
+-- See Note [Haskell.Monad]
 sequence_ = foldr c (Haskell.return ())
   where c m k = m Haskell.>> k
         {-# INLINE c #-}
```
effectfully, line 165: Bad version of `sequenceA_`.

plutus-tx/src/PlutusTx/Foldable.hs
```diff
@@ -139,12 +139,14 @@ product = getProduct #. foldMap Product
 
 -- | Plutus Tx version of 'Data.Foldable.foldrM'.
 foldrM :: (Foldable t, Haskell.Monad m) => (a -> b -> m b) -> b -> t a -> m b
+-- See Note [Haskell.Monad]
 foldrM f z0 xs = foldl c Haskell.return xs z0
   where c k x z = f x z Haskell.>>= k
         {-# INLINE c #-}
```
effectfully, line 145: The existence of this function very much suggests that monads aren't only useful because of do-notation.

plutus-tx/src/PlutusTx/Foldable.hs
```diff
@@ -139,12 +139,14 @@ product = getProduct #. foldMap Product
...
 
 -- | Plutus Tx version of 'Data.Foldable.foldlM'.
 foldlM :: (Foldable t, Haskell.Monad m) => (b -> a -> m b) -> b -> t a -> m b
+-- See Note [Haskell.Monad]
 foldlM f z0 xs = foldr c Haskell.return xs z0
   where c x k z = f z x Haskell.>>= k
         {-# INLINE c #-}
```
effectfully, line 152: Ditto. Ironically though, this function is less inappropriate in Plutus than it is in Haskell (lazy left folds are stupid).

effectfully:
(I know you proposed that on slack and I was going to respond there, but got caught up in other stuff)

plutus-tx/src/PlutusTx/Foldable.hs
zliu41: > which are easily resolvable with [QualifiedDo](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/qualified_do.html)

This is technically easy but I find it user-unfriendly, having to use different `do`s in different places. Going forward we should start reducing our own typeclass hierarchy rather than adding to it (which is what PLT-847 is about). Our `Applicative` and a few others can probably go away right now without affecting anything.

Plutus Tx needs to be kept simple and avoid advanced Haskell features. I have never used `QualifiedDo` myself.

effectfully: I wouldn't mind removing all the definitions that rely on `Haskell.Monad` at all. Particularly since I find them either completely useless (like `sequence_` and `mapM_`) or very rarely useful (like `foldrM` and `foldlM`), so the Plutus Tx Prelude could live without them.

> Plutus Tx needs to be kept simple and avoid advanced Haskell features.

I wouldn't call superficial syntactic extension an advanced feature, but YMMV.

> I have never used `QualifiedDo` myself.

Me neither, but I don't use TH in the middle of a definition too, while in Plutus it's perfectly normal. In my opinion, if you have an entirely separate hierarchy of type classes, you can afford using an entirely separate do-notation as well.

plutus-tx/src/PlutusTx/Foldable.hs
michaelpj: I agree, this is quite suspect still. It's surprising that this works.

plutus-tx/src/PlutusTx/Foldable.hs
```diff
@@ -225,6 +228,20 @@ find p = coerce . foldMap (\ x -> First (if p x then Just x else Nothing))
...
 {-# INLINABLE mapM_ #-}
 mapM_ :: (Foldable t, Haskell.Monad m) => (a -> m b) -> t a -> m ()
+-- See Note [Haskell.Monad]
 mapM_ f = foldr c (Haskell.return ())
   where c x k = f x Haskell.>> k
         {-# INLINE c #-}
+
+{- Note [Haskell.Monad]
+There is no `PlutusTx.Monad`. Since do-notations use `Haskell.Monad` (without
+`RebindableSyntax` anyway), it won't be very useful to have `PlutusTx.Monad`.
```
michaelpj, line 238: Maybe we should just suggest that people use `RebindableSyntax` if they want to use do notation? It's pretty sketchy what we're doing right now.

plutus-tx/src/PlutusTx/Foldable.hs
michaelpj: Yeah, I do think the current situation is pretty suspect and I think I would somewhat prefer either:
1. Remove all the functions we provide that use Haskell`s `Monad`, note that do-notation works somewhat by coincidence.
2. Provide our own `Monad` and related functions, encourage people to use `RebindableSyntax`/`QualifiedDo` or similar.

plutus-tx/src/PlutusTx/Foldable.hs
michaelpj: I think this note just makes us sound like we know what we're doing too much :D I think the truth is probably more like "we didn't want to bother with dealing with `QualifiedDo` etc. (and it's only in GHC 9), and the status quo seemed to work so we didn't look too hard".

PR #5176, opened 2023-02-27
Use tee in ci-plutus-benchmark.sh

zeme-iohk:


scripts/ci-plutus-benchmark.sh
```diff
@@ -48,7 +48,7 @@ echo "[ci-plutus-benchmark]: Clearing caches with cabal clean ..."
 cabal clean
 
 echo "[ci-plutus-benchmark]: Running benchmark for PR branch at $PR_BRANCH_REF ..."
-cabal bench $BENCHMARK_NAME >bench-PR.log 2>&1
```
michaelpj, line 51: hmm, do we not want to redirect stderr still?

PR #5179, opened 2023-02-27
PLT-1539: Use `foldr` for `Foldable`

zliu41:
Suggested by @effectfully in https://github.com/input-output-hk/plutus/pull/5168#issuecomment-1445268943

This doesn't make `any` and friends short-circuit, but does reduce the cost (sometimes dramatically) across the board, even more so than [this suggestion for improving foldMap](https://github.com/input-output-hk/plutus/pull/5168#pullrequestreview-1314554431).


michaelpj:
Seems good to me. I doubt anyone is defining their own `Foldable` instances, and if they are it's an easy fix.

plutus-tx-plugin/test/Budget/allCheap.plc.golden
```diff
@@ -2,66 +2,33 @@ letrec data (List :: * -> *) a | Nil_match where
...
+                  /\dead ->
+                    let !acc : Bool = go xs in Bool_match
+                                                 (ifThenElse
+                                                    {Bool}
+                                                    (lessThanEqualsInteger 1 x)
+                                                    False True) {all dead. Bool}
+                                                 (/\dead -> acc) (/\dead ->
+                                                                    False)
+                                                 {all dead. dead})
+               {all dead. dead}
```
effectfully, line 23: So everything works just fine except GHC makes a thunk for `go xs` and pulls it out of `ifThenElse`. And we interpret it a strict binding.

plutus-tx/src/PlutusTx/Foldable.hs
```diff
@@ -37,75 +38,74 @@ module PlutusTx.Foldable (
...
 instance Foldable [] where
-    {-# INLINABLE foldMap #-}
-    foldMap _ []     = mempty
-    foldMap f (x:xs) = f x <> foldMap f xs
+    {-# INLINABLE foldr #-}
+    foldr f z = go
+      where
+        go = \case
+          []     -> z
+          x : xs -> f x (go xs)
```
effectfully, line 70: What about `oneShot (f x) (go xs)` then? I know we probably shouldn't try to do fancy things to get efficient compilation, because it's way too unreliable, but maybe a few general policies here and there would be fine?

plutus-tx/src/PlutusTx/Foldable.hs
```diff
@@ -200,12 +200,16 @@ or = sum
 -- | Plutus Tx version of 'Data.Foldable.any'.
 {-# INLINABLE any #-}
 any :: Foldable t => (a -> Bool) -> t a -> Bool
-any p = getSum #. foldMap (Sum #. p)
+any p = foldr f False
+  where
+    f a acc = p a || acc
```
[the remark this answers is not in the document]
zliu41: I've tried inlining `f` and `||` manually, as well as `oneShot`, `inline` and `{-# INLINE #-}`, but without telling the GHC to run the inliner (`sm_inline = False`), nothing I've tried managed to make GHC turn

```haskell
(\a acc -> if p a then True else acc) x (go xs)
```

into
```haskell
if p x then True else go xs
```

We do run GHC's pre-inliner, but apparently that doesn't do it.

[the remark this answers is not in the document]
michaelpj: We need to propagate the `INLINE`, then it might work.

plutus-tx-plugin/test/Budget/allCheap.plc.golden
```diff
@@ -2,66 +2,33 @@ letrec data (List :: * -> *) a | Nil_match where
...
-                                                      (lessThanEqualsInteger
-                                                         1 v) False True)
-     ((let a = List integer in \(c : integer -> a -> a) (n : a) ->
-                                 c
-                                   1 (c
-                                        2 (c
-                                             3 (c
+in
+letrec !go : List integer -> Bool
+         = \(ds : List integer) ->
```
michaelpj, line 10: prettyprinting improvement: always put the binding on the next line so we can indent it just 2 instead of 7 for `letrec`

plutus-tx-plugin/test/Budget/allCheap.plc.golden
```diff
@@ -2,66 +2,33 @@ letrec data (List :: * -> *) a | Nil_match where
...
+             Nil_match
+               {integer} ds {all dead. Bool} (/\dead -> True)
+               (\(x : integer) (xs : List integer) ->
+                  /\dead ->
+                    let !acc : Bool = go xs in Bool_match
```
michaelpj, line 15: why didn't we break at the `in` here?

plutus-tx-plugin/test/Budget/allCheap.plc.golden
```diff
@@ -2,66 +2,33 @@ letrec data (List :: * -> *) a | Nil_match where
...
+                                                    (lessThanEqualsInteger 1 x)
+                                                    False True) {all dead. Bool}
+                                                 (/\dead -> acc) (/\dead ->
+                                                                    False)
+                                                 {all dead. dead})
+               {all dead. dead}
+in
+let !eta : List integer
+      = (let a = List integer in \(c : integer -> a -> a) (n : a) ->
+                                   c
```
michaelpj, line 27: this is hilarious but I'm not sure what we could do better

zliu41:
I'll capture these prettyprinter improvements in a ticket.

I also noticed

```
data (Monoid :: * -> *) a | Monoid_match where
    CConsMonoid : (\a -> a -> a -> a) a -> a -> Monoid a
```

That seems messed up.

michaelpj:
I just went ahead and did it: https://github.com/input-output-hk/plutus/pull/5184

Not sure what's up with the `CConsMonoid` thing though!

PR #5182, opened 2023-02-27
PLT-1605: Add some benchmarks for comparing the script context as data

michaelpj:
There are some concerns that no longer encoding the script context as `Data` removes the ability to compare chunks of it for equality using `equalsData`. I wanted to check how bad that was, hence this benchmark.

It looks pretty bad, ~90x worse! That's not entirely surprising, in that in this case a) we really are able to offload a lot of work onto Haskell, and b) it just is more complex to implement equality this way rather than on a homogenous structured data type.

I want to see how this changes on the SOP branch also.

(We can't even blame it on all of our boolean checks being non-short-circuiting, since this tests the pessimal case of two things that are equal.)

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


kwxm:
> It looks pretty bad, ~90x worse!

Where does the number 90 come from?  I tried subtracting the overhead from both of the other results and dividing, but that gave me 65c CPU usage.

kwxm (approved):
Looks fine apart from a typo or two.  I have a methodology question though.  The point of these is to see what the overhead is, but doing it this way means that you have to generate some golden files and look at their contents.  Another way to do this would be to write an executable that constructs the terms, feeds them to the evaluator, and reports the overheads on the terminal (and I have some code for the BLS primitives that works like that, although it's testing dozens of cases to see how cost varies with size).  Have you done it this way just because it's cheap to implement since you can easily fit it into the existing infrastructure, or is there some deeper point of principle involved?

michaelpj:
>  I have a methodology question though. The point of these is to see what the overhead is

No, the point is to see how much the _actual computation is_. I just realised that it's hard to see that without accounting for the overhead.

> Have you done it this way just because it's cheap to implement because you can easily fit it into the existing infrastructure

Exactly. I'm lazy. This way I can just spit out the budgets, it's fairly easy to compute the things you care about, and I don't particularly anticipate looking at this a lot in future.

michaelpj:
> Where does the number 90 come from? I tried subtracting the overhead from both of the other results and dividing, but that gave me 65x CPU usage.

It comes from me doing arithmetic in my head badly :)

kwxm:
> No, the point is to see how much the _actual computation is_. I just realised that it's hard to see that without accounting for the overhead.

Oh yes, that's just me using the wrong word!

PR #5183, opened 2023-02-28
PLT-1607: Propagate INLINE pragmas to PIR

michaelpj:
This is easy to do, and gives the user a bit more control over inlining.

As a bonus, it looks like GHC adds inline pragmas to some things like `DFun`s (dictionary selectors), which is also to our advantage.


zliu41:
Looks good! Please rebase and see if it makes a difference on `anyCheap` etc.

michaelpj:
No change to `anyCheap`, but I don't think we were expecting anything?

zliu41:
> No change to anyCheap, but I don't think we were expecting anything?

Currently there's only `INLINALBE` pragma on `foldr` and `any`. It may make a difference if you change them to `INLINE`, but I'm fine if you don't want to do it in this PR.

michaelpj:
> It may make a difference if you change them to INLINE, but I'm fine if you don't want to do it in this PR.

Yeah, I'll leave it for now.

PR #5184, opened 2023-02-28
Improve readable prettyprinting of PIR

michaelpj:
This fixes some things I didn't like about the readable PIR prettyprinting, notably:
- Bindings in lets were always indented at least as far as the `let`/`letrec`. If we break immediately we can consistently indent them at most 2 spaces.
- The body of a let was not broken vertically when the rest of the let was, which seemed inconsistent to me.
- Arguments in applications and bindings were laid out with `fillSep`, which led to weird behaviour especially when one of the later items was itself substantial and needed to break. If we just do the boring thing and use `vsep` then we get IMO simpler and more readable output.
- When we split a variable declaration, the second part went at the same indentation level, which looked odd.

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Core/Instance/Pretty/Readable.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Core/Instance/Pretty/Readable.hs
@@ -85,19 +85,18 @@ instance (PrettyConstraints configName tyname name uni, Pretty fun)
           compoundDocM juxtFixity $ \ prettyIn ->
             let ppArg (Left a)  = braces $ prettyIn ToTheRight botFixity a
                 ppArg (Right t) = prettyIn ToTheRight juxtFixity t
-            -- Using `align` here and gathering the arguments together helps to lay out
-            -- function applications compactly like:
+            -- Lay out function applications like:
             --
-            -- foo a b c
-            --     d e f
+            -- foo
+            --   a
+            --   b
+            --   c
             --
             -- or
             --
-            -- foo
-            --  veryLongArg
-            --  a b c
+            -- foo a b c
             --
-            in prettyIn ToTheLeft juxtFixity fun <?> align (fillSep (map ppArg args))
+            in prettyIn ToTheLeft juxtFixity fun <?> vsep (map ppArg args)
         Apply{}    -> error "The impossible happened. This should be covered by the `viewApp` case above."
         TyInst{}   -> error "The impossible happened. This should be covered by the `viewApp` case above."
         Var _ name -> prettyM name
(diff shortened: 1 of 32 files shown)


plutus-core/plutus-ir/test/prettyprintingReadable/listMatch.golden
```diff
@@ -1,8 +1,13 @@
...
-     (/\a_1 -> \(x_5 : a_1) -> x_5) (\(head_6 : all a_1. a_1 -> a_1) (tail_7
-                                      : List_0 (all a_1. a_1 -> a_1)) ->
-                                       head_6)
\ No newline at end of file
+letrec
+  data (List_0 :: * -> *) a_1 | match_List_2 where
+    Nil_3 : List_0 a_1
+    Cons_4 : a_1 -> List_0 a_1 -> List_0 a_1
+in
+match_List_2
```
zliu41, line 6: Is it better to add indentation here? That would make it more clear where the body of the binding ends (especially if there are multiple bindings, e.g. `let ... in let ... in let ... in`)

zliu41 (approved):
Much better.

plutus-core/plutus-ir/test/prettyprintingReadable/listMatch.golden
michaelpj: I think the whole point is to avoid increasing the indentation when there are multiple let bindings! The existing prettyprinter does this already. I think in other cases where it's ambiguous we should get parentheses? :thinking:

zliu41: Ok, I'm fine either way. Since we have `LetMerge` the indentation shouldn't be terrible, but not having indentation isn't bad either.

plutus-core/plutus-ir/src/PlutusIR/Core/Instance/Pretty/Readable.hs
```diff
@@ -106,19 +105,18 @@ instance (PrettyConstraints configName tyname name uni, Pretty fun)
...
-            --  \ (x : t) (y : t')
-            --    (z : t'') -> body
-            -- or
-            --  \ (x : t) (y : t')
-            --    (z : t'') ->
-            --    bigStartOfBody
+            --  \(x : t)
+            --   (y : t')
+            --   (z : t'') ->
+            --    body
```
effectfully, line 115: Yay, I never liked the original alignment.

plutus-core/plutus-ir/src/PlutusIR/Core/Instance/Pretty/Readable.hs
```diff
@@ -135,14 +133,17 @@ instance (PrettyConstraints configName tyname name uni, Pretty fun)
...
+                    -- so that the bindings can be simply indented by 2 spaces, keeping the indent low
+                    prettyLet r binds = vsep [ nest 2 ("let" <> prec r <> line <> vcatHard (prettyBot <$> binds)), "in"]
                 -- Lay out let-bindings in a layout-sensitive way
                 --
-                -- let !x : t = a
+                -- let
+                --   !x : t = a
+                --   !y : t = b
                 -- in
-                -- let !y : t = b
```
effectfully, line 142: Even yayer!

effectfully (approved):
So much better!

PR #5186, opened 2023-02-28
[Builtins] [Costing] Add 'lazy' to each clause

effectfully:
Just checking, don't look here.


effectfully:
I'll include these changes in a future PR where there's interplay between these changes and other ones, so closing the PR.

PR #5156, opened 2023-02-21
Kwxm/expose optimisers (PLT-1145)

kwxm:
[**PLEASE DON'T REVIEW THIS YET:** I'm trying to fix something]

This does what it says in the changelog:

* Adds  `optimise` options for the `pir`, `plc`, and `uplc` commands.
* Adds a `compile` option for the `pir` command which allows a PIR file to be compiled to PLC or UPLC.
* Generalises some of the other things in `pir` to accept both flat and textual input.

The `optimise` options just do the easiest things that were already available (in particular, for PIR optimisation is kind of mixed up with translation to PLC so it was tricky to expose everything that it does).  There's another ticket to add a configurable optimisation pipeline so when/if that gets done I may come back and extend the `optimise` commands here.

PR #5189, opened 2023-03-01
Kwxm/expose optimisers 3 (PLT-1145)

kwxm:
In my [previous attempt](https://github.com/input-output-hk/plutus/pull/5156) at this I disappeared down a dead end about de Bruijn indices.  Here's a new version.

This does what it says in the changelog:

  * Adds `optimise` options for the `pir`, `plc`, and `uplc` commands.
  * Adds a compile option for the `pir` command which allows a PIR file to be compiled to PLC or UPLC.
  * Generalises some of the other things in `pir` to accept both flat and textual input.
  * I also split the `Common` file up into several bits because I was getting lost trying to navigate through it.
  
The `optimise` options just do the easiest things that were already available (in particular, for PIR optimisation is kind of mixed up with translation to PLC so it was tricky to expose everything that it does and what's implemented here doesn't really _optimise_ things). There's another ticket to add a configurable optimisation pipeline so when/if that gets done I may come back and extend the `optimise` commands here.


plutus-core/plutus-core/src/PlutusCore/Subst.hs
```diff
@@ -161,6 +164,56 @@ termSubstClosedTerm varFor new = go where
          t                    -> t & over termSubterms go
     goUnder var term = if var == varFor then term else go term
 
+-- Mapping names
```
kwxm, line 167: This is so that the `plc` and `uplc` commands can perform the same conversions. It wasn't too hard to implement so I thought it wouldn't do any harm to have it since we already `flat` instances for the different types of names.  It's a different story for PIR though.

plutus-core/executables/pir/Main.hs
```diff
@@ -1,117 +1,242 @@
...
-
-compile :: COpts -> PirProg () -> Either PIRError PLCTerm
-compile opts (PIR.Program _ pirT) = do
+type PirError a = PIR.Error PLC.DefaultUni PLC.DefaultFun a
+type UnitProvenance = PIR.Provenance ()
+
+
+{- Note [De Bruijn indices and PIR]
+   The `plc` and `uplc` commands both support ASTs whose "names" are de Bruijn
+   inidces.  These aren't supported for PIR because PIR has `Let` blocks of
```
bezirg, line 43: type `indices`

plutus-core/executables/pir/Main.hs
```diff
@@ -1,117 +1,242 @@
...
+   possibly mutually recursive definitions and it's not clear whether de Bruijn
+   indices (or levels) really make sense in the presence of mutual recursion
+   since scopes aren't properly nested in that case: if we process the AST in
+   the normal way then a declaration may have to refer to another variable which
+   hasn't yet been declared.  It may be possbile to overcome this by allowing
```
bezirg, line 48: typo `possible`

plutus-core/executables/src/PlutusCore/Executable/Parsers.hs
```diff
@@ -4,10 +4,16 @@
...
-import PlutusCore.Executable.Common
+import PlutusCore.Executable.Types
 
 import Options.Applicative
 
+-- | Invert a switch: return False if the switch is supplied on the command
+-- line, True otherwise.  This is used for command-line options to turn things
+-- off.
+invertedSwitch :: Mod FlagFields Bool -> Parser Bool
+invertedSwitch = fmap not . switch
```
bezirg, line 15: alternative implementation: `flag False True`

plutus-core/executables/pir/Main.hs
```diff
@@ -1,117 +1,242 @@
...
+pOptimise = invertedSwitch
+            (  long "dont-optimise"
+            <> long "dont-optimize"
+            <> help ("Turn off optimisations")
+            )
+
+pJustTest :: Parser Bool
+pJustTest = switch ( long "test"
+                   <> help "Just report success or failure, don't produce an output file"
+                   )
```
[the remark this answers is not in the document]
bezirg: I take it back, it is not the same as dry-run

plutus-core/executables/pir/Main.hs
```diff
@@ -1,117 +1,242 @@
...
+    let plcCompilerOpts =
+            if optimise
+            then PLC.defaultCompilationOpts
+            else PLC.defaultCompilationOpts & PLC.coSimplifyOpts . UPLC.soMaxSimplifierIterations .~ 0
+    in flip runReader plcCompilerOpts $ runQuoteT $ PLC.compileProgram plcProg
+
+loadPirAndCompile :: CompileOptions -> IO ()
+loadPirAndCompile (CompileOptions language optimise test inp ifmt outp ofmt mode)  = do
+    pirProg <- readProgram (pirFormatToFormat ifmt) inp
+    when test $ putStrLn "!!! Compiling"
```
bezirg, line 174: Perhaps stderr is a better place for this and then you can leave it always enabled (removing the `when test`)

plutus-core/executables/src/PlutusCore/Executable/AstIO.hs
```diff
@@ -0,0 +1,192 @@
...
+{- | Convert an untyped program to one where the 'name' type is
+ textual names with de Bruijn indices.
+-}
+toNamedDeBruijnUPLC ::
+    UplcProg ann ->
+    IO (UPLC.Program UPLC.NamedDeBruijn PLC.DefaultUni PLC.DefaultFun ann)
+toNamedDeBruijnUPLC prog =
+    case runExcept @PLC.FreeVariableError $ traverseOf UPLC.progTerm UPLC.deBruijnTerm prog of
+        Left e  -> error $ show e
+        Right p -> return p
```
bezirg, line 55: the todebruijnuplc can reuse the toNamedDeBruijnUplc as:

```
toDeBurijnUPLC prog = UPLC.programMapNames unNameDeBruijn <$> toNamedDeBruijnUplc
```

plutus-core/executables/src/PlutusCore/Executable/AstIO.hs
```diff
@@ -0,0 +1,192 @@
...
+ Bruijn indices.
+-}
+toNamedDeBruijnPLC ::
+    PlcProg ann ->
+    IO (PLC.Program PLC.NamedTyDeBruijn PLC.NamedDeBruijn PLC.DefaultUni PLC.DefaultFun ann)
+toNamedDeBruijnPLC prog =
+    case runExcept @PLC.FreeVariableError $ traverseOf PLC.progTerm PLC.deBruijnTerm prog of
+        Left e  -> error $ show e
+        Right p -> return p
+
```
bezirg, line 74: the same for those two

plutus-core/executables/src/PlutusCore/Executable/AstIO.hs
```diff
@@ -0,0 +1,192 @@
...
+        case nameType of
+          Named         -> pure $ BSL.fromStrict $ flat p
+          DeBruijn      -> BSL.fromStrict . flat <$> toDeBruijnUPLC p
+          NamedDeBruijn -> BSL.fromStrict . flat <$> toNamedDeBruijnUPLC p
+
+
+-- Deserialising ASTs from Flat
+
+fakeTyNameDeBruijn :: PLC.DeBruijn -> PLC.NamedTyDeBruijn
+fakeTyNameDeBruijn = PLC.NamedTyDeBruijn . fakeNameDeBruijn
```
bezirg, line 114: I would move this to DeBruijn/Internal and reexport from Debruijn.hs

plutus-core/executables/src/PlutusCore/Executable/AstIO.hs
```diff
@@ -0,0 +1,192 @@
...
+    fromDeBruijn :: UplcProgramNdB ann -> IO (UplcProg ann)
+    fromDeBruijn prog = do
+      case PLC.runQuote $
+           runExceptT @UPLC.FreeVariableError $ traverseOf UPLC.progTerm UPLC.unDeBruijnTerm prog of
+        Left e  -> error $ show e
+        Right p -> return p
+    handleResult =
+        \case
+            Left e  -> error $ "Flat deserialisation failure: " ++ show e
+            Right r -> return r
```
bezirg, line 191: `handleResult` is duplicated 2 times (3 in total)

plutus-core/plutus-core/src/PlutusCore/Subst.hs
```diff
@@ -161,6 +164,56 @@ termSubstClosedTerm varFor new = go where
...
+-- termMapNames requires two function arguments: one (called g) to map type names
+-- and another (called f) to map variable names.
+termMapNames
+    :: forall tyname tyname' name name' uni fun ann
+    .  (tyname -> tyname')
+    -> (name -> name')
+    -> Term tyname name uni fun ann
+    -> Term tyname' name' uni fun ann
+termMapNames g f = go
+    where
```
bezirg, line 195: Nit: same for this g

bezirg: Another idea that would perhaps fit better is that `termMapNames` accepts one function only (name -> name') and leaves tynames untouched.

Then have another function `termMapBothNames = typeMapNames . termMapNames`. This ofcourse is slower, but is more consistent. Right now termMapNames also maps types

bezirg (approved):

plutus-core/executables/pir/Main.hs
kwxm: Thanks.  GitHub suggestions (the +/ thing above the comment window) are good for typos!

plutus-core/executables/src/PlutusCore/Executable/Parsers.hs
kwxm: Ah, maybe that would be more compact.  I'll try that.

plutus-core/plutus-core/src/PlutusCore/Subst.hs
kwxm: Yes, that might be better. All of this would be a lot easier if TPLC just had one name parameter that it used for both types and terms.

kwxm: Oh wait, that doesn't quite work.  `typeMapNames` just maps a function over a type, so we'd need another full traversal over terms that applies `typeMapNames` to the types in the term.  I think I'll leave that as it is for now.

plutus-core/executables/pir/Main.hs
kwxm: I'm not sure about that.  It's probably safer to keep the noise down, but I was trying to preserve the previous behaviour as well.

plutus-core/executables/src/PlutusCore/Executable/AstIO.hs
kwxm: Thanks!  When I started to tidy that up I realised that there was a lot of monadic stuff in there that we didn't actually need.

kwxm:
There was a slightly tricky merge conflict with #5150, but I think I got away with it.

plutus-core/executables/pir/Main.hs
kwxm: > It should definitely go to stderr, but it looks like we pervasively log to stdout. We should perhaps change that separately.

I've created a ticket for that at PLT-5273.

plutus-core/changelog.d/20230220_085551_kenneth.mackenzie_expose_optimisers.md
```diff
@@ -0,0 +1,11 @@
...
+- `optimise` options for the `pir`, `plc`, and `uplc` commands.
+- A `compile` option for the `pir` command which allows a PIR file to be
+  compiled to PLC or UPLC.
+- Functions for mapping over names and typenames in the PLC AST.
+
+### Changed
+- Some of the `pir` commands have been extended to allow both `flat` and textual
+  input.
+- Some of the code under the `executables` directory has been moved into new
+  files `AST.hs` and `Types.hs` to make it easier to navigate the code.
```
bezirg, line 11:
```suggestion
- Some of the code under the `executables` directory has been moved into new
  files `AST.hs` and `Types.hs` to make it easier to navigate the code.
```

I don't think this point is changelog-worthy

plutus-core/executables/src/PlutusCore/Executable/Common.hs
```diff
@@ -366,80 +300,22 @@ parseInput inp = do
...
     Input ->
     IO (p PLC.SrcSpan)
-getProgram fmt inp =
+readProgram fmt inp =
     case fmt of
         Textual -> snd <$> parseInput inp
         Flat flatMode -> do
             prog <- loadASTfromFlat @p @() flatMode inp
             return $ topSrcSpan <$ prog
+        where
```
bezirg, line 318: Is this an empty `where`?

bezirg (approved):

PR #5190, opened 2023-03-02
PLT-1611: Fix Interval for discrete intervals

michaelpj:
`Interval` was not in a good place. It violated a bunch of laws. The first thing this PR does is add a bunch more tests:
- Generic tests for the typeclasses that `Interval` has instances of.
- Some specific tests that test that finite intervals behave the same as sets of consecutive numbers.

The latter gives us a "semantic" test that lets us check whether our operations on intervals are doing the right thing with respect to a model - in this case the model is the set of values that fall into the inteval.

Then we need to actually fix `Interval`. The most straightforward solution I could see is just to restrict most of its operations to intervals on enumerable types. That lets us turn non-inclusive bounds into inclusive bounds, which simplifies things a lot. Without that information, we can't easily even tell if e.g. an interval is empty or not.

Arguably we should just have never had non-inclusive bounds in the first place and it's unnecessary generality. However, I think changing that now would be more disruptive than it is worth.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
@@ -44,6 +44,7 @@ import PlutusTx qualified
 import PlutusTx.Lift (makeLift)
 import PlutusTx.Prelude
 
+-- See Note [Enumerable Intervals]
 {- | An interval of @a@s.
 
 The interval may be either closed or open at either end, meaning
@@ -55,7 +56,7 @@ The 'Haskell.Eq'uality of two intervals is specified as the canonical, structura
 the equality of the elements of their two underlying sets; the same holds for 'Haskell.Ord'.
 -}
 data Interval a = Interval { ivFrom :: LowerBound a, ivTo :: UpperBound a }
-    deriving stock (Haskell.Eq, Haskell.Ord, Haskell.Show, Generic)
+    deriving stock (Haskell.Show, Generic)
     deriving anyclass (NFData)
 
 instance Functor Interval where
@@ -66,7 +67,7 @@ instance Pretty a => Pretty (Interval a) where
 
 -- | A set extended with a positive and negative infinity.
 data Extended a = NegInf | Finite a | PosInf
-    deriving stock (Haskell.Eq, Haskell.Ord, Haskell.Show, Generic)
+    deriving stock (Haskell.Show, Generic)
     deriving anyclass (NFData)
 
 instance Functor Extended where
@@ -79,13 +80,29 @@ instance Pretty a => Pretty (Extended a) where
     pretty PosInf     = pretty "+∞"
     pretty (Finite a) = pretty a
 
+-- See Note [Enumerable Intervals]
 -- | Whether a bound is inclusive or not.
 type Closure = Bool
 
 -- | The upper bound of an interval.
 data UpperBound a = UpperBound (Extended a) Closure
-    deriving stock (Haskell.Eq, Haskell.Ord, Haskell.Show, Generic)
+    deriving stock (Haskell.Show, Generic)
     deriving anyclass (NFData)
+
+-- | For an enumerable type, turn an upper bound into a single inclusive
+-- bounding value.
+--
+-- Since the type is enumerable, non-inclusive bounds are equivalent
+-- to inclusive bounds on the predecessor.
+--
+-- See Note [Enumerable Intervals]
+strictUpperValue :: Enum a => UpperBound a -> Extended a
+-- already inclusive
+strictUpperValue (UpperBound v True)           = v
+-- take pred
+strictUpperValue (UpperBound (Finite x) False) = Finite $ pred x
+-- an infinity: inclusive/non-inclusive makes no difference
+strictUpperValue (UpperBound v False)          = v
 
 instance Functor UpperBound where
   fmap f (UpperBound e c) = UpperBound (f <$> e) c
@@ -98,8 +115,23 @@ instance Pretty a => Pretty (UpperBound a) where
 
 -- | The lower bound of an interval.
 data LowerBound a = LowerBound (Extended a) Closure
-    deriving stock (Haskell.Eq, Haskell.Ord, Haskell.Show, Generic)
+    deriving stock (Haskell.Show, Generic)
     deriving anyclass (NFData)
+
+-- | For an enumerable type, turn an lower bound into a single inclusive
+-- bounding value.
+--
+-- Since the type is enumerable, non-inclusive bounds are equivalent
+-- to inclusive bounds on the successor.
+--
+-- See Note [Enumerable Intervals]
+strictLowerValue :: Enum a => LowerBound a -> Extended a
+-- already inclusive
+strictLowerValue (LowerBound v True)           = v
+-- take succ
+strictLowerValue (LowerBound (Finite x) False) = Finite $ succ x
+-- an infinity: inclusive/non-inclusive makes no difference
+strictLowerValue (LowerBound v False)          = v
 
 instance Functor LowerBound where
   fmap f (LowerBound e c) = LowerBound (f <$> e) c
@@ -127,6 +159,9 @@ instance Eq a => Eq (Extended a) where
     Finite l == Finite r = l == r
     _        == _        = False
 
+instance Eq a => Haskell.Eq (Extended a) where
+    (==) = (==)
+
 instance Ord a => Ord (Extended a) where
     {-# INLINABLE compare #-}
     NegInf   `compare` NegInf   = EQ
@@ -137,31 +172,40 @@ instance Ord a => Ord (Extended a) where
     PosInf   `compare` _        = GT
     Finite l `compare` Finite r = l `compare` r
 
-instance Eq a => Eq (UpperBound a) where
+instance Ord a => Haskell.Ord (Extended a) where
+    compare = compare
+
+-- See Note [Enumerable Intervals]
+instance (Enum a, Eq a) => Eq (UpperBound a) where
     {-# INLINABLE (==) #-}
-    UpperBound v1 in1 == UpperBound v2 in2 = v1 == v2 && in1 == in2
+    b1 == b2 = strictUpperValue b1 == strictUpperValue b2
 
-instance Ord a => Ord (UpperBound a) where
+instance (Enum a, Eq a) => Haskell.Eq (UpperBound a) where
+    (==) = (==)
+
+-- See Note [Enumerable Intervals]
+instance (Enum a, Ord a) => Ord (UpperBound a) where
     {-# INLINABLE compare #-}
-    UpperBound v1 in1 `compare` UpperBound v2 in2 = case v1 `compare` v2 of
-        LT -> LT
-        GT -> GT
-        -- A closed upper bound is bigger than an open upper bound. This corresponds
-        -- to the normal order on Bool.
-        EQ -> in1 `compare` in2
+    b1 `compare` b2 = strictUpperValue b1 `compare` strictUpperValue b2
 
-instance Eq a => Eq (LowerBound a) where
+instance (Enum a, Ord a) => Haskell.Ord (UpperBound a) where
+    compare = compare
+
+-- See Note [Enumerable Intervals]
+instance (Enum a, Eq a) => Eq (LowerBound a) where
     {-# INLINABLE (==) #-}
-    LowerBound v1 in1 == LowerBound v2 in2 = v1 == v2 && in1 == in2
+    b1 == b2 = strictLowerValue b1 == strictLowerValue b2
 
-instance Ord a => Ord (LowerBound a) where
+instance (Enum a, Eq a) => Haskell.Eq (LowerBound a) where
+    (==) = (==)
+
+-- See Note [Enumerable Intervals]
+instance (Enum a, Ord a) => Ord (LowerBound a) where
     {-# INLINABLE compare #-}
-    LowerBound v1 in1 `compare` LowerBound v2 in2 = case v1 `compare` v2 of
-        LT -> LT
-        GT -> GT
-        -- An open lower bound is bigger than a closed lower bound. This corresponds
-        -- to the *reverse* of the normal order on Bool.
-        EQ -> in2 `compare` in1
+    b1 `compare` b2 = strictLowerValue b1 `compare` strictLowerValue b2
+
+instance (Enum a, Ord a) => Haskell.Ord (LowerBound a) where
+    compare = compare
 
 {-# INLINABLE strictUpperBound #-}
 {- | Construct a strict upper bound from a value.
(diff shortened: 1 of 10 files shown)


michaelpj:
I think moving `Interval` to `plutus-tx` is also good, but I thought it would make this PR hard to review, so I'll do it afterwards.

zliu41:
> I think moving Interval to plutus-tx is also good

I disagree - plutus-ledger-api should not depend on plutus-tx, and we were planning to reverse the dependency at some point.

michaelpj:
> I disagree - plutus-ledger-api should not depend on plutus-tx, and we were planning to reverse the dependency at some point.

I'm confused about this. Did we say that? I'm not sure it's really viable given how much basic stuff is defined in `plutus-tx`. But maybe it is.

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
```diff
@@ -137,31 +172,40 @@ instance Ord a => Ord (Extended a) where
...
-        GT -> GT
-        -- A closed upper bound is bigger than an open upper bound. This corresponds
-        -- to the normal order on Bool.
-        EQ -> in1 `compare` in2
-
-instance Eq a => Eq (LowerBound a) where
+    b1 `compare` b2 = strictUpperValue b1 `compare` strictUpperValue b2
+
+instance (Enum a, Ord a) => Haskell.Ord (UpperBound a) where
+    compare = compare
```
zliu41, line 192: nit: `compare = PlutusTx.compare` for better readibility

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
```diff
@@ -79,14 +80,30 @@ instance Pretty a => Pretty (Extended a) where
...
+--
+-- Since the type is enumerable, non-inclusive bounds are equivalent
+-- to inclusive bounds on the predecessor.
+--
+-- See Note [Enumerable Intervals]
+strictUpperValue :: Enum a => UpperBound a -> Extended a
+-- already inclusive
+strictUpperValue (UpperBound v True)           = v
+-- take pred
+strictUpperValue (UpperBound (Finite x) False) = Finite $ pred x
```
zliu41, line 103: `pred` is partial - undefined on `minBound`. So it should return Maybe, but that's probably more costly, so I'm fine with documenting the partiality.

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
```diff
@@ -267,46 +324,66 @@ overlaps l r = not $ isEmpty (l `intersection` r)
...
 
 {-# INLINABLE after #-}
 -- | Check if a value is later than the end of an 'Interval'.
-after :: Ord a => a -> Interval a -> Bool
+after :: (Enum a , Ord a) => a -> Interval a -> Bool
 after h (Interval _ t) = upperBound h > t
+
+{- Note [Enumerable Intervals]
+The 'Interval' type is set up to handle open intervals, where we have non-inclusive
+bounds. These are only meaningful for _dense_ orders (forall a b . exists c . a < b < c),
```
zliu41, line 373: a < c < b

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
```diff
@@ -79,14 +80,30 @@ instance Pretty a => Pretty (Extended a) where
...
     deriving anyclass (NFData)
 
+-- | For an enumerable type, turn an upper bound into a single inclusive
+-- bounding value.
+--
+-- Since the type is enumerable, non-inclusive bounds are equivalent
+-- to inclusive bounds on the predecessor.
+--
+-- See Note [Enumerable Intervals]
+strictUpperValue :: Enum a => UpperBound a -> Extended a
```
zliu41, line 99: Haskell has `Haskell.Enum Double`, we don't, so I guess this is ok. But generally speaking I'm not sure whether `Enum` necessarily implies discreteness.

zliu41 (commented):
Our `Interval` is way too complicated... In Plutus V3 we should consider supporting closed intervals only, simply define `Interval a` as `Empty | NonEmpty a a`, hide the data constructors and provide smart constructors.

If we really want open intervals, we can still use this definition of `Interval a`, e.g., `Interval (Closed a)` vs. `Interval (Open a)`. I don't think we need to support half-open intervals, or checking whether an open interval contains a closed interval - that's too complicated and I doubt anyone needs it.

michaelpj:
> Our Interval is way too complicated... In Plutus V3 we should consider supporting closed intervals only

I think that would be reasonable, but we have to support the old version anyway, and it's more copies of things to maintain. In this case I think it would only incur a copy of the `TxInfo` type, though, so maybe it's fine.

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
michaelpj: > Haskell.Enum Double

Wow. Um, okay. Then I don't think we have a typeclass that expresses what we want. It's really something like `next (x) = inf { y | y > x }` or something.

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
michaelpj: We don't have fixed-size integers, so that's less of a problem. In fact I think we don't have any partial `succ` or `pred` definitions. Still a bit suspect I agree.

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
michaelpj: I updated the comments to try and express this issue.

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
bezirg: > We don't have fixed-size integers, so that's less of a problem. In fact I think we don't have any partial `succ` or `pred` definitions. Still a bit suspect I agree.

We have partial `Enum Bool`:  `succ True` and `pred False` ?

michaelpj: You are absolutely right, and so is the instance for `()`.

bezirg: Isn't this equivalent and simpler? 

```
strictUpperValue (UpperBound (Finite x) False) = Finite $ pred x
strictUpperValue v = v
```

bezirg: Perhaps a better function name: `discretizeUpperBound` or `normalizeUpperBound`?

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
```diff
@@ -98,9 +117,24 @@ instance Pretty a => Pretty (UpperBound a) where
...
+-- to inclusive bounds on the successor.
+--
+-- See Note [Enumerable Intervals]
+strictLowerValue :: Enum a => LowerBound a -> Extended a
+-- already inclusive
+strictLowerValue (LowerBound v True)           = v
+-- take succ
+strictLowerValue (LowerBound (Finite x) False) = Finite $ succ x
+-- an infinity: inclusive/non-inclusive makes no difference
+strictLowerValue (LowerBound v False)          = v
```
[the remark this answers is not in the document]
bezirg: Perhaps a better function name: discretizeLowerBound or normalizeLowerBound?

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
bezirg: Also we would have a problem with `Enum ()` and `Enum Ordering` in PlutusTx

plutus-tx/src/PlutusTx/Enum.hs
```diff
@@ -12,9 +12,15 @@ import PlutusTx.Trace
...
+  -- | The successor of a value.  For numeric types, 'succ' adds 1.
+  --
+  -- For types that implement 'Ord', @succ x@ should be the least element
+  -- that is greater than @x@.
   succ :: a -> a
-  -- | the predecessor of a value.  For numeric types, 'pred' subtracts 1.
+  -- | The predecessor of a value.  For numeric types, 'pred' subtracts 1.
+  --
+  -- For types that implement 'Ord', @pred x@ should be the greatest element
+  -- that is less than @x@.
```
bezirg, line 23: ... and `error` if no such element exists.

plutus-ledger-api/test/Spec/Interval.hs
```diff
@@ -6,15 +6,46 @@
...
+genUpperBound finite g = Interval.UpperBound <$> genExtended finite g <*> Gen.bool
+
+genInterval :: Bool -> Hedgehog.Gen a -> Hedgehog.Gen (Interval.Interval a)
+genInterval finite g = Interval.Interval <$> genLowerBound finite g <*> genUpperBound finite g
+
+genFiniteDiscreteInterval :: Hedgehog.Gen (Interval.Interval Integer)
+genFiniteDiscreteInterval = genInterval True (Gen.integral (toInteger <$> Range.linear @Int 0 100))
+
+genDiscreteInterval :: Hedgehog.Gen (Interval.Interval Integer)
+genDiscreteInterval = genInterval False (Gen.integral (toInteger <$> Range.linear @Int 0 100))
```
bezirg, line 45: How about generating other staff than Integer, e.g. Bool?

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
```diff
@@ -267,46 +326,67 @@ overlaps l r = not $ isEmpty (l `intersection` r)
...
 isEmpty :: (Enum a, Ord a) => Interval a -> Bool
-isEmpty (Interval (LowerBound v1 in1) (UpperBound v2 in2)) = case v1 `compare` v2 of
-    LT -> if openInterval then checkEnds v1 v2 else False
+isEmpty (Interval lb ub) = case strictLowerValue lb `compare` strictUpperValue ub of
+    -- We have at least two possible values, the lower bound and the upper bound
+    LT -> False
+    -- We have one possible value, the lower bound/upper bound
+    EQ -> False
+    -- We have no possible values
     GT -> True
```
bezirg, line 361: I think this is equivalent to:

```
isEmpty (Interval lb ub) = strictLowerValue lb > strictUpperValue ub
```

, but it's ok for me to leave it as it is for clarity

bezirg (changes requested):

bezirg (commented):
-

bezirg (commented):
I like the direction of this patch. I just don't agree with the Enum constraint.

As @zliu41 correctly pointed out we rely on non-partial succ,pred and this breaks for some datatypes. So we are fixing one issue but creating another one.

Could we instead monomorphize this Internal module to only work with Integers , while keeping the interface to/from the ledger the same (i.e. makeIsData*) ?

Then we will be:

constraining this Interval implementation in plutus v1&v2 only to the correct Integers
and for v3 we scrap this Interval coming up with a better (non-backwards compatible) Interval definition.

michaelpj:
> Could we instead monomorphize this Internal module to only work with Integers

We'd have to monomorphize it to `POSIXTime`, since that's the actual thing the ledger has an interval over.

If we had GADTs we could just restrict these functions to only work over closed intervals, but we don't.

I think maybe the compromise position is what's here: there's a risk of partiality if you have an interval with non-inclusive endpoints over a type with a partial `Enum` instance. That's not _too_ bad: who makes the interval `(False, True]` anyway?

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
michaelpj: > Isn't this equivalent and simpler?

It's equivalent but I think I prefer my version because the fall-through case is okay for two distinct reasons.

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
michaelpj: Yeah, I just had to reason this one out a few times, so I thought I'd be super explicit!

plutus-ledger-api/test/Spec/Interval.hs
michaelpj: I added some tests for closed boolean intervals, open ones obviously don't work.

michaelpj:
Okay, I think this should be good to go if we can agree that this is the least bad way forwards. The only other one that seems remotely plausible is monomorphizing to `POSIXTime`, which seems kind of sad.

bezirg:
> Okay, I think this should be good to go if we can agree that this is the least bad way forwards. The only other one that seems remotely plausible is monomorphizing to `POSIXTime`, which seems kind of sad.

How about using the subclass of Enum, `Integral` as a constraint? This satisfies both posixtime and integer

zliu41 (approved):
> I think that would be reasonable, but we have to support the old version anyway, and it's more copies of things to maintain. In this case I think it would only incur a copy of the TxInfo type, though, so maybe it's fine.

A closed-only interval type for Plutus V3 could also be cheaper than the current interval type, given that it's much simpler.

> How about using the subclass of Enum, Integral as a constraint? This satisfies both posixtime and integer

That would be slightly better but not very much of an improvement because (1) we don't currently have an `Integral` class, and will have to make one specifically for this; (2) although we currently don't have those, there could technically be finite `Integral` types, whose `succ` and `pred` would either over/underflow or crash. So I'm OK with this current version.

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
```diff
@@ -195,25 +239,38 @@ The resulting bound includes all values that are equal or smaller than the input
...
 
-instance Eq a => Eq (Interval a) where
+-- See Note [Enumerable Intervals]
+instance (Enum a, Ord a) => Eq (Interval a) where
     {-# INLINABLE (==) #-}
-    l == r = ivFrom l == ivFrom r && ivTo l == ivTo r
+    -- Degenerate case: both the intervals are empty.
+    -- There can be many empty intervals, so we check for this case
+    -- explicitly
+    iv1 == iv2 | isEmpty iv1 && isEmpty iv2 = True
```
uhbif19, line 275: Why are empty Intervals equal? They do behave differently, for example if you call `hull` on them

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
```diff
@@ -267,46 +324,66 @@ overlaps l r = not $ isEmpty (l `intersection` r)
...
 @member s a@.
 -}
-contains :: Ord a => Interval a -> Interval a -> Bool
+contains :: (Enum a, Ord a) => Interval a -> Interval a -> Bool
+-- Everything contains the empty interval
+contains _ i2 | isEmpty i2 = True
+-- Nothing is contained in the empty interval (except the empty interval,
+-- which is already handled)
+contains i1 _ | isEmpty i1 = False
+-- Otherwise we check the endpoints. This doesn't work for empty intervals,
```
uhbif19, line 353: Why this does not work for empty intervals? I do not know reasons why containment relation on intervals should coincide to their points subset relation.

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
```diff
@@ -137,31 +172,40 @@ instance Ord a => Ord (Extended a) where
...
     {-# INLINABLE compare #-}
-    UpperBound v1 in1 `compare` UpperBound v2 in2 = case v1 `compare` v2 of
-        LT -> LT
-        GT -> GT
-        -- A closed upper bound is bigger than an open upper bound. This corresponds
-        -- to the normal order on Bool.
-        EQ -> in1 `compare` in2
-
-instance Eq a => Eq (LowerBound a) where
+    b1 `compare` b2 = strictUpperValue b1 `compare` strictUpperValue b2
```
uhbif19: Why? Why upper bound should be Ord at all?

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
```diff
@@ -267,46 +324,66 @@ overlaps l r = not $ isEmpty (l `intersection` r)
 {-# INLINABLE intersection #-}
 -- | 'intersection a b' is the largest interval that is contained in 'a' and in
 --   'b', if it exists.
-intersection :: Ord a => Interval a -> Interval a -> Interval a
+intersection :: (Enum a, Ord a) => Interval a -> Interval a -> Interval a
 intersection (Interval l1 h1) (Interval l2 h2) = Interval (max l1 l2) (min h1 h2)
 
 {-# INLINABLE hull #-}
 -- | 'hull a b' is the smallest interval containing 'a' and 'b'.
```
uhbif19, line 338: What does mean "containing"? This does not work with current `contains` definition.

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
```diff
@@ -267,46 +324,66 @@ overlaps l r = not $ isEmpty (l `intersection` r)
...
-hull :: Ord a => Interval a -> Interval a -> Interval a
+hull :: (Enum a, Ord a) => Interval a -> Interval a -> Interval a
 hull (Interval l1 h1) (Interval l2 h2) = Interval (min l1 l2) (max h1 h2)
 
 {-# INLINABLE contains #-}
 {- | @a `contains` b@ is true if the 'Interval' @b@ is entirely contained in
```
uhbif19, line 343: Why this is not only true, but necessary for containment?

plutus-ledger-api/src/PlutusLedgerApi/V1/Interval.hs
uhbif19: @michaelpj .

uhbif19:
> I don't think we need to support half-open intervals, or checking whether an open interval contains a closed interval - that's too complicated and I doubt anyone needs it.

Now we have published example of using half-open intervals.

https://github.com/mlabs-haskell/hydra-auction/pull/325/files#diff-af2bca17155b0096b1175f870c829fb2051dfd87558433e8feb70e05b9dc9e64L31

I think almost any "stages" intervals implementation requires half-open intervals.

> Okay, I think this should be good to go if we can agree that this is the least bad way forwards. The only other one that seems remotely plausible is monomorphizing to POSIXTime, which seems kind of sad.

You can just create your own class, with such laws written in docs, and export it for cases of extending for new types.

PR #5195, opened 2023-03-07
PLT-1539: Remove Plutus Tx functions with Haskell.Monad constraint

zliu41:
I'm fine with removing these to make things look less weird (especially `sequence_` and `mapM_`, which shouldn't be used anyway).

However, I still don't think we need our own `Monad` class. Rather, I'm in favor of removing our `Applicative` and `Functor` classes (and perhaps more) and use the Haskell ones, because.. why not? We don't have any instance of those classes whose definitions differ from the respective Haskell instances.

Some other classes, like `PlutusTx.AdditiveSemigroup`, exist for a good reason (for now): `Integer`'s `PlutusTx.+` differ from its `GHC.Num.+`, and the latter cannot be used ("for now" because they can be made to work in the future). I don't think this applies to `Monad`, `Applicative` and `Functor`.

### DIFF
--- a/plutus-tx/changelog.d/20230307_151002_unsafeFixIO_haskell_monad.md
+++ b/plutus-tx/changelog.d/20230307_151002_unsafeFixIO_haskell_monad.md
@@ -1,0 +1,3 @@
+### Removed
+
+- Removed Plutus Tx library functions with the `Haskell.Monad` constraint.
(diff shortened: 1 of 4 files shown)


effectfully (approved):

effectfully:
> We don't have any instance of those classes whose definitions differ from the respective Haskell instances.

We may not have the unfoldings for them, though?

plutus-tx/changelog.d/20230307_151002_unsafeFixIO_haskell_monad.md
```diff
@@ -0,0 +1,3 @@
+### Removed
+
+- Removed Plutus Tx library functions with the `Haskell.Monad` constraint.
```
michaelpj, line 3: Maybe include some justification here?

michaelpj (approved):

zliu41:
> We may not have the unfoldings for those, though?

The common ones to have unfoldings, I think. If there is indeed one that doesn't, newtyping it would be a more lightweight solution IMO.

PR #4914, opened 2022-10-19
[Builtins] [Evaluation] Drop lookups

effectfully:
Don't look here yet.

PR #5200, opened 2023-03-09
Benchmark the production code rather than some arbitrary thing

effectfully:
Hopefully fixes the issue described [here](https://github.com/input-output-hk/plutus/pull/4914#issuecomment-1387710846)

### DIFF
--- a/plutus-benchmark/validation/BenchCek.hs
+++ b/plutus-benchmark/validation/BenchCek.hs
@@ -17,7 +17,7 @@ import UntypedPlutusCore as UPLC
      `cabal bench -- plutus-benchmark:validation --benchmark-options crowdfunding`.
 -}
 main :: IO ()
-main = evaluate getEvalCtx *> benchWith mkCekBM
+main = evaluate (force getEvalCtx) *> benchWith mkCekBM
  where
    mkCekBM file program =
        -- don't count the undebruijn . unflat cost
--- a/plutus-benchmark/validation/BenchCek.hs
+++ b/plutus-benchmark/validation/BenchCek.hs
@@ -3,6 +3,7 @@ module Main where
 
 import Common
 import Control.DeepSeq (force)
+import Control.Exception
 import Criterion
 import PlutusBenchmark.Common
 import UntypedPlutusCore as UPLC
@@ -16,11 +17,10 @@ import UntypedPlutusCore as UPLC
      `cabal bench -- plutus-benchmark:validation --benchmark-options crowdfunding`.
 -}
 main :: IO ()
-main = benchWith mkCekBM
+main = evaluate getEvalCtx *> benchWith mkCekBM
  where
    mkCekBM file program =
        -- don't count the undebruijn . unflat cost
        -- `force` to try to ensure that deserialiation is not included in benchmarking time.
        let !nterm = force (toNamedDeBruijnTerm $ UPLC._progTerm $ unsafeUnflat file program)
-       in whnf unsafeEvaluateCekNoEmit' nterm
-
+       in whnf evaluateCekLikeInProd nterm
--- a/plutus-benchmark/validation/Common.hs
+++ b/plutus-benchmark/validation/Common.hs
@@ -3,10 +3,13 @@
 module Common (
     benchWith
     , unsafeUnflat
-    , unsafeEvaluateCekNoEmit'
+    , getEvalCtx
+    , evaluateCekLikeInProd
     , peelDataArguments
     , Term
     ) where
+
+import PlutusPrelude
 
 import PlutusBenchmark.Common (getConfig, getDataDir)
 import PlutusBenchmark.NaturalSort
@@ -15,7 +18,10 @@ import PlutusCore qualified as PLC
 import PlutusCore.Builtin qualified as PLC
 import PlutusCore.Data qualified as PLC
 import PlutusCore.Evaluation.Machine.ExBudgetingDefaults qualified as PLC
-import PlutusCore.Evaluation.Machine.Exception
+import PlutusCore.Evaluation.Result
+import PlutusLedgerApi.Common (LedgerPlutusVersion (PlutusV3), evaluateTerm)
+import PlutusLedgerApi.Common.Versions (languageIntroducedIn)
+import PlutusLedgerApi.V3 (EvaluationContext, ParamName, VerboseMode (..), mkEvaluationContext)
 import UntypedPlutusCore qualified as UPLC
 import UntypedPlutusCore.Evaluation.Machine.Cek qualified as UPLC
 
@@ -24,6 +30,8 @@ import Criterion.Main.Options (Mode, parseWith)
 import Criterion.Types (Config (..))
 import Options.Applicative
 
+import Control.Monad.Trans.Except
+import Control.Monad.Trans.Writer.Strict
 import Data.ByteString qualified as BS
 import Data.List (isPrefixOf)
 import Flat
@@ -128,13 +136,30 @@ benchWith act = do
         env (BS.readFile $ dir </> file) $ \scriptBS ->
             bench (dropExtension file) $ act file scriptBS
 
-unsafeEvaluateCekNoEmit' :: UPLC.Term PLC.NamedDeBruijn PLC.DefaultUni PLC.DefaultFun () -> PLC.EvaluationResult  (UPLC.Term PLC.NamedDeBruijn PLC.DefaultUni PLC.DefaultFun ())
-unsafeEvaluateCekNoEmit' =
-       (\(e, _, _) -> unsafeExtractEvaluationResult e) .
-            UPLC.runCekDeBruijn
-                PLC.defaultCekParameters
-                UPLC.restrictingEnormous
-                UPLC.noEmitter
+getEvalCtx
+    :: Either
+            (UPLC.CekEvaluationException UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun)
+            EvaluationContext
+getEvalCtx = do
+    costParams <-
+        maybe
+            (Left evaluationFailure)
+            (Right . take (length $ enumerate @ParamName) . toList)
+            PLC.defaultCostModelParams
+    either (const $ Left evaluationFailure) (Right . fst) . runExcept . runWriterT $
+        mkEvaluationContext costParams
+{-# NOINLINE getEvalCtx #-}
+
+evaluateCekLikeInProd
+    :: UPLC.Term PLC.NamedDeBruijn PLC.DefaultUni PLC.DefaultFun ()
+    -> Either
+            (UPLC.CekEvaluationException UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun)
+            (UPLC.Term UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun ())
+evaluateCekLikeInProd term = do
+    evalCtx <- getEvalCtx
+    let (getRes, _, _) =
+            evaluateTerm UPLC.restrictingEnormous (languageIntroducedIn PlutusV3) Quiet evalCtx term
+    getRes
 
 type Term = UPLC.Term UPLC.DeBruijn UPLC.DefaultUni UPLC.DefaultFun ()
 
--- a/plutus-benchmark/validation/Common.hs
+++ b/plutus-benchmark/validation/Common.hs
@@ -150,6 +150,7 @@ getEvalCtx = do
         mkEvaluationContext costParams
 {-# NOINLINE getEvalCtx #-}
 
+-- | Evaluate a term as it would be evaluated using the on-chain evaluator.
 evaluateCekLikeInProd
     :: UPLC.Term PLC.NamedDeBruijn PLC.DefaultUni PLC.DefaultFun ()
     -> Either
@@ -158,7 +159,8 @@ evaluateCekLikeInProd
 evaluateCekLikeInProd term = do
     evalCtx <- getEvalCtx
     let (getRes, _, _) =
-            evaluateTerm UPLC.restrictingEnormous (languageIntroducedIn PlutusV3) Quiet evalCtx term
+            -- The validation benchmarks were all created from PlutusV1 scripts
+            evaluateTerm UPLC.restrictingEnormous (languageIntroducedIn PlutusV1) Quiet evalCtx term
     getRes
 
 type Term = UPLC.Term UPLC.DeBruijn UPLC.DefaultUni UPLC.DefaultFun ()
--- a/plutus-benchmark/validation/Common.hs
+++ b/plutus-benchmark/validation/Common.hs
@@ -19,7 +19,7 @@ import PlutusCore.Builtin qualified as PLC
 import PlutusCore.Data qualified as PLC
 import PlutusCore.Evaluation.Machine.ExBudgetingDefaults qualified as PLC
 import PlutusCore.Evaluation.Result
-import PlutusLedgerApi.Common (LedgerPlutusVersion (PlutusV3), evaluateTerm)
+import PlutusLedgerApi.Common (LedgerPlutusVersion (PlutusV1), evaluateTerm)
 import PlutusLedgerApi.Common.Versions (languageIntroducedIn)
 import PlutusLedgerApi.V3 (EvaluationContext, ParamName, VerboseMode (..), mkEvaluationContext)
 import UntypedPlutusCore qualified as UPLC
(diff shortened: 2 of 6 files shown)


effectfully:
Not running the benchmarks yet, because the machine is busy apparently.

effectfully:
OK, in addition to the issues referenced in the description of the PR, we have more problems:

1. benchmarking on `master` is pessimistic, because `unsafeEvaluateCekNoEmit'` returns an `EvaluationResult` rather than an `Either` and the former is strict, meaning forcing it will result in discharging being performed. Not sure if it actually matters in our specific benchmarks (if they all compute to, say, a constant, then there's basically no difference)
2. the actual production evaluator appears to be 4-5% slower than whatever we benchmark on `master` currently. Not good, but could be much worse

So why the 4-5% difference? I've checked the Core, this is what we have on `master`:

```haskell
-- RHS size: {terms: 7, types: 6, coercions: 63, joins: 0/0}
unsafeEvaluateCekNoEmit'2
  :: forall {s}.
     State# s
     -> (# State# s,
           ExBudgetInfo RestrictingSt DefaultUni DefaultFun s #)
unsafeEvaluateCekNoEmit'2
  = \ (@s_adFA) (eta_adFB :: State# s_adFA) ->
      $wrestricting
        (unsafeEvaluateCekNoEmit'3 `cast` <Co:63>)
        9223372036854775807#
        9223372036854775807#
        eta_adFB

-- RHS size: {terms: 26, types: 114, coercions: 98, joins: 0/0}
unsafeEvaluateCekNoEmit'
  :: Term NamedDeBruijn DefaultUni DefaultFun ()
     -> EvaluationResult (Term NamedDeBruijn DefaultUni DefaultFun ())
unsafeEvaluateCekNoEmit'
  = \ (x_abKi :: Term NamedDeBruijn DefaultUni DefaultFun ()) ->
      case runCekDeBruijn
             (unsafeEvaluateCekNoEmit'3 `cast` <Co:63>)
             defaultCekParameters1
             (unsafeEvaluateCekNoEmit'2 `cast` <Co:16>)
             (noEmitter1 `cast` <Co:19>)
             x_abKi
      of
      { (e_a9Z4, ds_dbMx, ds1_dbMy) ->
      case e_a9Z4 of {
        Left ds2_abKw ->
          case ds2_abKw of { ErrorWithCause evalErr_abKB cause_abKC ->
          case evalErr_abKB of {
            InternalEvaluationError err_abMs ->
              unsafeEvaluateCekNoEmit'1 err_abMs cause_abKC;
            UserEvaluationError ds3_adLv -> EvaluationFailure
          }
          };
        Right term1_abMv -> $WEvaluationSuccess term1_abMv
      }
      }
```

and this is what we have in this branch:

```haskell
-- RHS size: {terms: 7, types: 6, coercions: 63, joins: 0/0}
evaluateCekLikeInProd1
  :: forall {s}.
     State# s
     -> (# State# s,
           ExBudgetInfo RestrictingSt DefaultUni DefaultFun s #)
evaluateCekLikeInProd1
  = \ (@s_ahLF) (eta_ahLG :: State# s_ahLF) ->
      $wrestricting
        (evaluateCekLikeInProd2 `cast` <Co:63>)
        9223372036854775807#
        9223372036854775807#
        eta_ahLG

-- RHS size: {terms: 16, types: 76, coercions: 99, joins: 0/0}
evaluateCekLikeInProd
  :: Term NamedDeBruijn DefaultUni DefaultFun ()
     -> Either
          (CekEvaluationException NamedDeBruijn DefaultUni DefaultFun)
          (Term NamedDeBruijn DefaultUni DefaultFun ())
evaluateCekLikeInProd
  = \ (term_acKg :: Term NamedDeBruijn DefaultUni DefaultFun ()) ->
      case getEvalCtx of {
        Left l_ahIy -> Left l_ahIy;
        Right r_ahIA ->
          case runCekDeBruijn
                 (evaluateCekLikeInProd17 `cast` <Co:63>)
                 (r_ahIA `cast` <Co:1>)
                 (evaluateCekLikeInProd1 `cast` <Co:16>)
                 (noEmitter1 `cast` <Co:19>)
                 term_acKg
          of
          { (getRes_adJt, ds_dfRE, ds1_dfRF) ->
          getRes_adJt
          }
      }
```

I can't pinpoint what makes this evaluator slower than the one we have on `master`. The only differences I'm able to observe are:

1. `master`'s evaluator shares pretty-printing, `Typeable` etc instance across the budgeter and the CEK machine runner, while evaluator in this branch doesn't. I can't imagine this making a noticeable difference,  but it's an odd detail
2. builtins are instantiated differently. This can easily make the difference, because builtins are heavily optimized and one failed `INLINE` can affect evaluation times enormously. In the evaluator used on `master` all `BuiltinRuntime`s are pulled to the top-level and turned into CAFs (last time I checked), while the production evaluator encapsulated all the thunks under a lambda that later gets instantiated. Such a huge difference can have huge impact on evaluation times, but it would be very hard to tell what exactly causes it

Overall, we should just finish off this PR, i.e. document it and test that the evaluator used for benchmarking doesn't perform discharging (like we test for the production evaluator in `plutus-ledger-api`), and create a ticket for investigating the performance gap further.

plutus-benchmark/validation/BenchCek.hs
```diff
@@ -16,11 +17,10 @@ import UntypedPlutusCore as UPLC
      `cabal bench -- plutus-benchmark:validation --benchmark-options crowdfunding`.
 -}
 main :: IO ()
-main = benchWith mkCekBM
+main = evaluate getEvalCtx *> benchWith mkCekBM
```
michaelpj, line 20: Does this definitely ensure that this is shared across the runs? It seems like it would be simpler to use https://hackage.haskell.org/package/criterion-1.6.0.0/docs/Criterion-Types.html#v:env

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
```diff
@@ -132,6 +133,25 @@ mkDynEvaluationContext ver newCMP =
 assertWellFormedCostModelParams :: MonadError CostModelApplyError m => Plutus.CostModelParams -> m ()
 assertWellFormedCostModelParams = void . Plutus.applyCostModelParams Plutus.defaultCekCostModel
 
+evaluateTerm
```
michaelpj, line 136:
```suggestion
-- | Evaluate a fully-applied term using the CEK machine. Useful for mimicing the behaviour of the on-chain evaluator.
evaluateTerm
```

plutus-benchmark/validation/Common.hs
```diff
@@ -128,13 +136,30 @@ benchWith act = do
...
-       (\(e, _, _) -> unsafeExtractEvaluationResult e) .
-            UPLC.runCekDeBruijn
-                PLC.defaultCekParameters
-                UPLC.restrictingEnormous
-                UPLC.noEmitter
+getEvalCtx
+    :: Either
+            (UPLC.CekEvaluationException UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun)
+            EvaluationContext
+getEvalCtx = do
```
michaelpj, line 143: I think it would be legit to just do `mkDynEvaluationContext <version> PLC.defaultCostModelParams`. Since we're trying to not include this in the benchmark run.

This is also fine, though.

plutus-benchmark/validation/Common.hs
```diff
@@ -128,13 +136,30 @@ benchWith act = do
...
+
+evaluateCekLikeInProd
+    :: UPLC.Term PLC.NamedDeBruijn PLC.DefaultUni PLC.DefaultFun ()
+    -> Either
+            (UPLC.CekEvaluationException UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun)
+            (UPLC.Term UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun ())
+evaluateCekLikeInProd term = do
+    evalCtx <- getEvalCtx
+    let (getRes, _, _) =
+            evaluateTerm UPLC.restrictingEnormous (languageIntroducedIn PlutusV3) Quiet evalCtx term
```
michaelpj, line 161:
```suggestion
            -- The validation benchmarks were all created from PlutusV1 scripts
            evaluateTerm UPLC.restrictingEnormous (languageIntroducedIn PlutusV1) Quiet evalCtx term
```

plutus-benchmark/validation/Common.hs
```diff
@@ -128,13 +136,30 @@ benchWith act = do
...
+    costParams <-
+        maybe
+            (Left evaluationFailure)
+            (Right . take (length $ enumerate @ParamName) . toList)
+            PLC.defaultCostModelParams
+    either (const $ Left evaluationFailure) (Right . fst) . runExcept . runWriterT $
+        mkEvaluationContext costParams
+{-# NOINLINE getEvalCtx #-}
+
+evaluateCekLikeInProd
```
michaelpj, line 153:
```suggestion
-- | Evaluate a term as it would be evaluated using the on-chain evaluator.
evaluateCekLikeInProd
```

plutus-benchmark/validation/Common.hs
```diff
@@ -128,13 +136,30 @@ benchWith act = do
...
-                PLC.defaultCekParameters
-                UPLC.restrictingEnormous
-                UPLC.noEmitter
+getEvalCtx
+    :: Either
+            (UPLC.CekEvaluationException UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun)
+            EvaluationContext
+getEvalCtx = do
+    costParams <-
+        maybe
```
michaelpj, line 145: @bezirg do we not have a function somewhere for going from `CostModelParams` to the list of integers?

michaelpj: We probably should even just so we can test that they round-trip.

plutus-benchmark/validation/BenchCek.hs
michaelpj: Ah, or perhaps it wasn't clear: the creation of the evaluation context has non-trivial cost, but _is_ shared by the ledger between script evaluations. So it shouldn't be part of the benchmarked code.

effectfully: Lol, yes, of course it should be `evaluate (force getEvalCtx)`, there's that lazy `Either`.

plutus-benchmark/validation/Common.hs
effectfully: > Since we're trying to not include this in the benchmark run.

What "this"? 

In any case, I wanted to stay as close to the production evaluator as possible and not do any "reasoning". `getEvalCtx` is a CAF that is forced before benchmarking starts and is used across all benchmarks, hence I don't see how it would distort the results (assuming I've fixed the forcing to go to NF rather than WHNF, but even the latter can't distort the results in a meaningful manner).

effectfully:
Hm, it's +6.39% now. Let's try again.

effectfully:
-5.07%. That's a pretty big gap in benchmarking results TBH.

effectfully:
Comments addressed, @michaelpj merge if you're OK with what's in here.

plutus-benchmark/validation/BenchCek.hs
michaelpj: > Did you read the description? The function seems completely irrelevant.

Yes, and I've used it before. I thought the point of it was to clearly compute a value (it lets you use IO so you can do `evaluate`) before the benchmarking begins, and feed it into the individual benchmarks as a function argument. We could even share it across all the benchmarking runs, which would save us some time.

> getEvalCtx is a CAF and doesn't get inlined (it has a NOINLINE pragma and I've verified that it's not getting inlined by looking at the Core anyway). I can't imagine it not being reused, why would GHC ever recompute a non-functional monomorphic top-level value?

This seems like far more complex reasoning than using a function argument and `env`? That guarantees it'll only be computed once.

plutus-benchmark/validation/Common.hs
michaelpj: > What "this"?

This function goes through the dance of computing the integers-only version of the cost model, and then feeding it back into the function that converts it into the original form. That's good, insofar as it mirrors the production version, except that this happens in the shared "compute the evaluation context" block, and so shouldn't be included in the benchmarks anyway. So it's fine to use the faster version that just creates the evaluation context without jumping through the hoops.

michaelpj:
I'm happy modulo the thing about how we ensure that the evaluation context is shared. I just think we could establish it more simply.

plutus-benchmark/validation/BenchCek.hs
effectfully: > I thought the point of it was to clearly compute a value (it lets you use IO so you can do evaluate) before the benchmarking begins, and feed it into the individual benchmarks as a function argument.

effectfully: > I thought the point of it was to clearly compute a value (it lets you use IO so you can do evaluate) before the benchmarking begins, and feed it into the individual benchmarks as a function argument.

But it's not what this function is for!

> Motivation. In earlier versions of criterion, all benchmark inputs were always created when a program started running. By deferring the creation of an environment when its associated benchmarks need the its, we avoid two problems that this strategy caused:
>
> - Memory pressure distorted the results of unrelated benchmarks. If one benchmark needed e.g. a gigabyte-sized input, it would force the garbage collector to do extra work when running some other benchmark that had no use for that input. Since the data created by an environment is only available when it is in scope, it should be garbage collected before other benchmarks are run.
> - The time cost of generating all needed inputs could be significant in cases where no inputs (or just a few) were really needed. This occurred often, for instance when just one out of a large suite of benchmarks was run, or when a user would list the collection of benchmarks without running any.

OK, forget about CAF etc. We only care about `getEvalCtx` being forced before benchmarking starts. `evaluate` does exactly that. Once it's forced, it can't be unforced and whatever they do with `env` isn't more reliable than that.

And `getEvalCtx` isn't an input to the benchmarks to even consider `env`. It's a value that the function being benchmarked uses under the hood, you can't feed the value to the function, it's already inside!

If you want to change `evaluateCekLikeInProd` to accept an `EvaluationContext` as an input, I don't mind that, in that case it would make more sense to call `env`, although as I said whatever lazy magic it does, we don't need or care about it, it's completely irrelevant for our use case, so even in that case it would make to force `getEvalCtx` manually.

michaelpj: > If you want to change evaluateCekLikeInProd to accept an EvaluationContext as an input

Yes sorry, that's what I was suggesting. `evaluateCekLikeInProd` just calls `getEvalCtx` and then calls `evaluateTerm`. Note that that's actually _less_ like the production version, which really does take it as a parameter and it's cached externally. This way we're relying on some complicated logic that we can force it here and then apparently recompute it inside a function and have it not repeat work. It seems much clearer to just... pass it as an argument!

effectfully: OK then, sounds reasonable.

Feel like doing it yourself? I'm trying to give away as much non-costing-related work as possible.

michaelpj: Sure.

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
bezirg: @michaelpj @effectfully What do you mean by "mimicking the behaviour of the on-chain evaluator"? This function `evaluateTerm` is what the actual on-chain node calls through the `plutus-ledger-api` !?

michaelpj: But it's not _the actual function_ that the ledger calls. So we are still trying to make sure we mimic it's behaviour, and it could go wrong if there was something important in the bit we're not calling.

bezirg:
The only difference I can see is that:

- before, in every benchmark run, the runCekDebruijn was called with the CAF `defaultCekParameters :: MachineParameters`
- now, in every benchmark run, the runCekDeBruijn evaluates `(toMachineParameters pv ectx)`

I don't know if this is the culprit of the 5%.

effectfully:
> now, in every benchmark run, the runCekDeBruijn evaluates `(toMachineParameters pv ectx)`

It's basically zero-cost:

```haskell
newtype EvaluationContext = EvaluationContext
    { machineParameters :: DefaultMachineParameters
    }

toMachineParameters :: ProtocolVersion -> EvaluationContext -> DefaultMachineParameters
toMachineParameters _ = machineParameters
```

> before, in every benchmark run, the runCekDebruijn was called with the CAF `defaultCekParameters :: MachineParameters`

Now we have a different CAF: `getEvalCtx`.

PR #5205, opened 2023-03-10
Add workflow for longitudinal benchmarking

zeme-iohk:
This PR introduces a new file `.github/workflows/new-benchmark.yml`, which describes a new workflow for longitudinal benchmarking. 
It uses [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark).

In short, whenever the benchmarks are run, the results are stored and added to the history. 
The charts will be available at the plutus GitHub Pages URL: `https://input-output-hk.github.io/plutus/dev/bench/`
The suffix `/dev/bench` is configurable.
When a regression is detected (new value is larger than % of old value, configurable via the action's `alert-threshold` field) the workflow job fails and a comment is posted.

I have tested this on my plutus fork.
This is what it looks like with two data points: https://zeme-iohk.github.io/plutus/dev/bench/
This is what it looks like when a regression is detected: https://github.com/zeme-iohk/plutus/commit/83e504b719061e39e018a47b4647420c6a5f6bf1#commitcomment-103842408

This is a more complete example: https://benchmark-action.github.io/github-action-benchmark/dev/bench/

What's left to do:
- Decide when to run the benchmarks (push to master? specific PR label?)
- Which benchmarks to run
- Format the output of the benchmarks according to `github-action-benchmark`

One limitation is that the `alert-threshold` value is shared across all benchmarks.


.github/workflows/new-benchmark.yml
```diff
@@ -0,0 +1,55 @@
...
+          output-file-path: output.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # Workflow will fail when an alert happens
+          fail-on-alert: true
+          # Push and deploy GitHub pages branch automatically
+          auto-push: true
+          # Enable alert commit comment
+          comment-on-alert: true
+          # Mention @rhysd in the commit comment
+          alert-comment-cc-users: '@zeme-iohk'
```
zeme-iohk, line 55: Who would like to be notified here?

.github/workflows/new-benchmark.yml
```diff
@@ -0,0 +1,56 @@
...
+          alert-threshold: '110%'
```
zeme-iohk, line 56: Percentage value like "110%". It is a ratio indicating how worse the current benchmark result is. For example, if we now get 110 ns/iter and previously got 100 ns/iter, it gets 110% worse.

michaelpj:
We need to document this also. I'd like to see some notes on your investigation and what you decided? In particular, where's the data being stored? It's not obvious - can we get it out later? Is it safe?

How does this interact with the stuff that's already getting pushed to github pages?

zeme-iohk:
The choice of benchmarking tool was very straightforward, as I could find no valid alternatives to `github-action-benchmark`.

`github-action-benchmark` does exactly what we need and more: 
- Is language agnostic, we just have to format the benchmark data accordingly
- Is a GitHub action that requires minimal setup 
- Automatically pushes to GitHub pages <- no need for 3rd party hosting solution 
- Automatically charts the benchmarking data
- Can detect performance regressions and send notifications  
- Provides a feature to download the entire benchmarking data history

I've also looked at these: 
https://www.tweag.io/blog/2018-12-12-benchgraph/
https://github.com/piotrmurach/rspec-benchmark
https://github.com/wg/wrk
And many more.
But they are not what we are looking for: those are tools for benchmarking code, not for (benchmark) data visualisation.

Again the only limitation I could find with `github-action-benchmark` is the `alert-threshold`, which is shared across all benchmarks.
This can be worked around either by:
- Creating a separate job for each benchmark (which allows us to specify a separate `alert-threshold`)
- Writing a custom script to detect regressions according to our own logic, and to manually send notifications.
  - This does not interfere with the data viz. features, which we still get "for free" 

> In particular, where's the data being stored? It's not obvious - can we get it out later? Is it safe?
> How does this interact with the stuff that's already getting pushed to github pages?

The data is being stored at `<github-pages-url>/dev/bench/data.js`
The entire history can be downloaded as json by clicking the blue `download data as json` button on the bottom left of the page. E.g. https://benchmark-action.github.io/github-action-benchmark/dev/bench/

The frontend code is stored in the `/dev/bench` directory in the gh-pages branch.
That path is configurable.
It's isolated from the stuff that's already getting pushed to github pages.

michaelpj:
Maybe worth writing up your research somewhere? It might be useful for other people or teams in the future. Could be an ADR?

.github/workflows/new-benchmark.yml
```diff
@@ -0,0 +1,55 @@
...
+
+jobs:
+  new-benchmark:
+    name: Performance regression check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.3.0
+
+      - name: Run benchmark
+        run: |
```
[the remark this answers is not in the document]
michaelpj: I don't think we want to run `ci-benchmark-compare` at all. We're not running a comparison, we're just emitting the results for a single commit, the comparison is done in the visualisation etc.

Maybe the thing to do is to use criterion's JSON output and then munge that with `jq`? http://www.serpentine.com/criterion/tutorial.html#writing-out-data

.github/workflows/new-benchmark.yml
```diff
@@ -0,0 +1,56 @@
...
+  contents: write
+
+jobs:
+  new-benchmark:
+    name: Performance regression check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.3.0
+
+      - name: Run benchmark
```
michaelpj, line 21: I think it's worthwhile doing this in its own job so it alone can run on the benchmarking machine, and it's isolated.

.github/workflows/new-benchmark.yml
```diff
@@ -0,0 +1,56 @@
...
+          
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1.16.2
+        with:
+          name: My Project Go Benchmark
+          tool: 'customSmallerIsBetter'
+          output-file-path: output.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # Workflow will fail when an alert happens
+          fail-on-alert: true
```
michaelpj, line 49: I don't think we want this? This doesn't help us much: those are (presumably) true results that we want to push and record. We do want to be notified of them but the workflow failing isn't helpful.

.github/workflows/new-benchmark.yml
michaelpj: @input-output-hk/plutus-core ?

.github/workflows/new-benchmark.yml
michaelpj: Put the explanation in the code :) I think we probably want something lower than this, maybe 5%?

.github/workflows/new-benchmark.yml
michaelpj: Also then we can run this as a matrix job if we want to do multiple benchmarks in the future.

.github/workflows/new-benchmark.yml
```diff
@@ -0,0 +1,56 @@
+name: New Benchmark
```
michaelpj, line 1: Longitudinal benchmark?

.github/workflows/new-benchmark.yml
```diff
@@ -0,0 +1,56 @@
...
+
+on:
+  push:
+    branches:
+      - master
```
michaelpj, line 6: I think we might want to set up some ignore paths or something. This is going to be somewhat expensive (30mins for the validation benchmark, per commit), so maybe we should try and skip some commits that don't change anything?

.github/workflows/new-benchmark.yml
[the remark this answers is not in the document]
michaelpj: Maybe we should see if we can help get https://github.com/benchmark-action/github-action-benchmark/pull/130 over the line?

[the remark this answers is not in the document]
zeme-iohk: > I don't think we want to run ci-benchmark-compare at all. 

Yes, my mistake. 
Here we just run the actual benchmark to produce the expected `output.json`

.github/workflows/new-benchmark.yml
zeme-iohk: 👍

.github/workflows/new-benchmark.yml
```diff
@@ -1,4 +1,4 @@
-name: New Benchmark
+name: Longitudinal Benchmarking
```
michaelpj, line 1: change the file name too?

.github/workflows/longitudinal-benchmark.yml
```diff
@@ -0,0 +1,75 @@
...
+        run: | 
+          for bench in $BENCHMARKS; do 
+            2>&1 cabal run "$bench" | tee "$bench-output.txt"
+          done 
+          
+          read -r -d '' PYTHON_SCRIPT <<- END_SCRIPT
+              import json 
+              result = []
+              for benchmark in "$BENCHMARKS".split():
+                  with open(f"{benchmark}-output.txt", "r") as file:
```
michaelpj, line 41: You decided not to use criterion's JSON output? seems like it might have been a bit easier, but not a big deal I guess.

.github/workflows/longitudinal-benchmark.yml
```diff
@@ -0,0 +1,75 @@
...
+                                  "value": float(mean)
+                              })
+              with open("output.json", "w") as file: 
+                json.dump(result, file)
+          END_SCRIPT
+          
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1.16.2
+        with:
+          name: My Project Go Benchmark
```
michaelpj, line 62:
```suggestion
          name: `Plutus Benchmarks`
```

michaelpj (approved):

.github/workflows/longitudinal-benchmark.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Run benchmarks
+        env: 
+          BENCHMARKS: "validation validation-decode"
+        run: | 
+          for bench in $BENCHMARKS; do 
+            2>&1 cabal run "$bench" | tee "$bench-output.txt"
+          done 
+          
+          read -r -d '' PYTHON_SCRIPT <<- END_SCRIPT
```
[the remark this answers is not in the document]
zeme-iohk: Sure

Issue #5204, opened 2023-03-09
PlutusCore.Default.Builtins takes ~8 seconds to reload on a powerful machine


L-as:
### Summary

Doing `:r` in GHCi on this module takes around ~8 seconds on a powerful Hetzner Cloud machine.
I assume this is due to the heavy type class magic on `makeBuiltinMeaning`.

### Steps to reproduce the behavior

1. Open file
2. Save
3. Reload

### Actual Result

~8 seconds to reload.

### Expected Result

~1 second at most.

### Describe the approach you would take to fix this

Removing the magic from `makeBuiltinMeaning` would likely fix the issue. A solution might be to make the arity explicit through e.g. `NP` from `sop-core` rather than inferring the arity, removing implicit conversions, etc.
Do note I do not fully understand the module, so I might be wrong.

### System info

```
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         40 bits physical, 48 bits virtual
  Byte Order:            Little Endian
CPU(s):                  16
  On-line CPU(s) list:   0-15
Vendor ID:               AuthenticAMD
  Model name:            AMD EPYC Processor
    CPU family:          23
    Model:               49
    Thread(s) per core:  1
    Core(s) per socket:  16
    Socket(s):           1
    Stepping:            0
    BogoMIPS:            4890.80
 ```
 (snippet from `lscpu`)

effectfully:
Just checked, reloads in 4 seconds on my machine. 

> Removing the magic from `makeBuiltinMeaning` would likely fix the issue. 

Yes. Specifically, it's probably [elaboration](https://github.com/input-output-hk/plutus/blob/4497b2f7d1fe120424ffe405feab7c045b32fe19/plutus-core/plutus-core/src/PlutusCore/Builtin/Elaborate.hs) stuff. Even more specifically, it's likely the logic of merging list of referenced variables using [`Merge`](https://github.com/input-output-hk/plutus/blob/4497b2f7d1fe120424ffe405feab7c045b32fe19/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs#L281) as type families are notoriously inefficient. Not sure what to do about it, maybe making the user specify type variables themselves would work, somehow...

Anyways, I don't think we should mark it as `bug`, it's not really such a high priority issue.

I'll look into it some time, thank you for bringing it up.

L-as:
[...]

PR #5208, opened 2023-03-11
[Builtins] Optimize 'ToBinds'

effectfully:
Prompted by  #5204. This version of `ToBinds` should be much more efficient than the previous one, but it doesn't seem to matter at all. Still, why be slow if you can be fast with roughly the same complexity.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
@@ -20,8 +20,8 @@ module PlutusCore.Builtin.KnownTypeAst
     , TypeHole
     , KnownBuiltinTypeAst
     , KnownTypeAst (..)
+    , Insert
     , Delete
-    , Merge
     ) where
 
 import PlutusCore.Builtin.Emitter
@@ -179,9 +179,9 @@ class KnownTypeAst uni x where
     type ToHoles x = ToHoles (ElaborateBuiltin x)
 
     -- | Collect all unique variables (a variable consists of a textual name, a unique and a kind)
-    -- in an @a@.
-    type ToBinds x :: [GADT.Some TyNameRep]
-    type ToBinds x = ToBinds (ElaborateBuiltin x)
+    -- in an accumulator and return the accumulator once a leaf is reached.
+    type ToBinds (acc :: [GADT.Some TyNameRep]) x :: [GADT.Some TyNameRep]
+    type ToBinds acc x = ToBinds acc (ElaborateBuiltin x)
 
     -- | The type representing @a@ used on the PLC side.
     toTypeAst :: proxy x -> Type TyName uni ()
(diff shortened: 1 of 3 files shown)


plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
```diff
@@ -271,13 +271,15 @@ instance
...
+    Insert x '[]      = '[x]
+    Insert x (x : xs) = x ': xs
+    Insert x (y : xs) = y ': Insert x xs
+
 -- | Delete all @x@s from a list.
 type family Delete x xs :: [a] where
     Delete _ '[]       = '[]
-    Delete x (x ': xs) = Delete x xs
+    Delete x (x ': xs) = xs
     Delete x (y ': xs) = y ': Delete x xs
```
effectfully, line 285: Now that we do this kind of logic in `Insert`, we don't need to do it here.

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
```diff
@@ -229,29 +229,29 @@ toTyNameAst _ =
...
-    type ToBinds (TyVarRep name) = '[ 'GADT.Some name ]
+    type ToBinds acc (TyVarRep name) = Insert ('GADT.Some name) acc
     toTypeAst _ = TyVar () . toTyNameAst $ Proxy @('TyNameRep text uniq)
     {-# INLINE toTypeAst #-}
 
 instance (KnownTypeAst uni fun, KnownTypeAst uni arg) => KnownTypeAst uni (TyAppRep fun arg) where
     type IsBuiltin (TyAppRep fun arg) = IsBuiltin fun && IsBuiltin arg
     type ToHoles (TyAppRep fun arg) = '[RepHole fun, RepHole arg]
-    type ToBinds (TyAppRep fun arg) = Merge (ToBinds fun) (ToBinds arg)
+    type ToBinds acc (TyAppRep fun arg) = ToBinds (ToBinds acc fun) arg
```
effectfully, line 254: Merging lists each time we stumble upon a type application or a `->` was quite inefficient. Now we just pass an accumulator around and only touch it when there's a variable or a binder, which happens fairly rarely (the latter never happens for the default builtins).

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
michaelpj: Worth a comment? Or updating the Haddock? "Delete the first instance of `x` from a list (which is okay since we only ever put things in once"?

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
```diff
@@ -180,8 +180,8 @@ class KnownTypeAst uni x where
 
     -- | Collect all unique variables (a variable consists of a textual name, a unique and a kind)
     -- in an @a@.
-    type ToBinds x :: [GADT.Some TyNameRep]
-    type ToBinds x = ToBinds (ElaborateBuiltin x)
+    type ToBinds (acc :: [GADT.Some TyNameRep]) x :: [GADT.Some TyNameRep]
```
michaelpj, line 183: document the accumulator? maybe it's obvious

michaelpj (approved):

PR #5217, opened 2023-03-17
PLT-1041 and PLT-1146 Implement inlining of fully applied functions.

thealmarty:
THIS ISN'T FINISHED I'll need to tidy it up more and fix some tests that it breaks. But it is working to some degree!

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-core/plutus-ir/test/datatypes/listMatchEval.golden
```diff
@@ -1 +1,13 @@
-(delay (lam x_0 x_0))
\ No newline at end of file
+(lam
```
thealmarty, line 1: ?????????

plutus-core/plutus-ir/test/transform/inline/letFunConstMulti.golden
```diff
@@ -5,5 +5,11 @@
     (vardecl constFun (fun (con integer) (fun (con integer) (con integer))))
     (lam x (con integer) (lam y (con integer) x))
   )
-  [ constFun [ [ constFun (con integer 3) ] (con integer 5) ] ]
+  [
+    constFun
+    [
+      [ (lam x (con integer) (lam y (con integer) x)) (con integer 3) ]
```
thealmarty, line 11: only the second `constFun` is inlined, as desired.

plutus-core/plutus-ir/test/transform/inline/letFunInFunMultiFullyApplied.golden
```diff
@@ -7,6 +7,6 @@
   )
   [
     [ (lam y (fun (con integer) (con integer)) idFun) idFun ]
-    [ idFun (con integer 1) ]
+    [ (lam x (con integer) x) (con integer 1) ]
```
thealmarty, line 10: only the fully applied `idFun` is inlined, as desired.

plutus-core/plutus-ir/test/transform/inline/letOverAppMulti.golden
```diff
@@ -17,6 +17,12 @@
       )
       (lam x (fun (con integer) (con integer)) idFun)
     )
-    [ [ k [ k idFun ] ] (con integer 6) ]
+    [
+      [
+        (lam x (fun (con integer) (con integer)) idFun)
+        [ (lam x (fun (con integer) (con integer)) idFun) idFun ]
```
thealmarty, line 23: Both instances of `k` got inlined to `funApp`, as desired.

thealmarty:
Summary so far:
- renamed `UnconditionalInline.hs` to `Inline.hs` because the call site inlining code is embedded in the unconditional `inline` function from before.
- checked that the inline tests make sense:
  - `letAppMulti` and others stay unchanged as desired
  - `letFunConstMulti.golden` etc have desired changes
-  `letIdFunForallMulti` and `letIdFunForallMultiNotSat` currently shows that type application may be a problem. Need to investigate further.

plutus-core/plutus-ir/test/transform/inline/letIdFunForallMulti.golden
```diff
@@ -4,7 +4,10 @@
     (strict) (vardecl idFun (all a (type) (fun a a))) (abs a (type) (lam x a x))
   )
   [
-    [ (builtin addInteger) [ { idFun (con integer) } (con integer 3) ] ]
-    [ { idFun (con integer) } (con integer 3) ]
+    [
```
thealmarty, line 7: both `idFun` are inlined, as desired.

plutus-core/plutus-ir/test/transform/inline/letIdFunForallMultiNotSat.golden
```diff
@@ -3,5 +3,8 @@
   (termbind
     (strict) (vardecl idFun (all a (type) (fun a a))) (abs a (type) (lam x a x))
   )
-  [ { idFun (fun (con integer) (con integer)) } { idFun (con integer) } ]
+  [
+    { (abs a (type) (lam x a x)) (fun (con integer) (con integer)) }
```
thealmarty, line 7: only the first `idFun` is inlined, as desired.

thealmarty:
OK this is ready for review. Instead of looking at the diff I advise looking at the new versions of `Inline.hs`, `CallSiteInline.hs`, and `Utils.hs`. 

- global uniqueness still needs to be re-established. But I prefer to do it in a follow up PR, if it doesn't break too many things. This one is large enough. My current plan is to rename the binders to fresh ones when I substitute in so we shouldn't need to rename the whole program. `ExternalConstraints` will probably be the only constraints that's needed. I'll rename it etc so yeah I prefer to do all that in another PR.
- curiously when the inlining of saturated functions with type lambdas was broken we got **more** improvement! [This](https://github.com/input-output-hk/plutus/pull/5217/commits/db3e9413f399f7d27899e88695f713eb00a1e527) is the improvement when saturated functions with type lambdas were NOT inline. Should we consider only inlining functions with term lambdas only? I'm not sure why inlining functions with type lambdas also would be worse.

zliu41 (commented):
The names of the test cases are hard to parse, e.g., what is `letFunInFunMultiFullyApplied`? Consider renaming them so that it is easier to answer questions like these: (1) which test cases involve type abstractions? (2) which test cases verify that `costIsAcceptable` and `sizeIsAcceptable` are respected? (3) which test cases involve strict vs. lazy bindings, or nonrec vs. rec bindings?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -53,21 +70,7 @@ in
...
-let f = \x.\y -> x
-in
-  let z = f q
-  in a + z c
-
-We cannot eliminate the let-binding of `f` because it is not dead (it's called in `z`).
-We don't inline `z` because it's not a lambda abstraction, it's an application.
+our dead code elimination pass is able to further reduce the code to just `a`.
 
 To find out whether a function is fully applied,
```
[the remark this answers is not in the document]
michaelpj: And the next paragraph is wrong.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
 
 We currently reuse the heuristics 'Utils.sizeIsAcceptable' and 'Utils.costIsAcceptable'
-that are used in 'UnconditionalInline'. For
+that are used in unconditional inlining. For
 
 let v = \x1.\x2...\xn.VBody in body
 
 we check `VBody` with the above "acceptable" functions.
-Note that all `LamAbs` and `TyAbs` should have been
+Note that all `LamAbs` and `TyAbs` are not acceptable in terms of size, but they should have been
```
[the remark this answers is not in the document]
michaelpj: And I'm not really sure what it means.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+-- it's `Utils.acceptable`.
 computeArity ::
-    Term tyname name uni fun a
-    -> Arity
+    Term tyname name uni fun ann
+    -> (Arity, Term tyname name uni fun ann)
 computeArity = \case
-    LamAbs _ _ _ body -> MkTerm : computeArity body
-    TyAbs _ _ _ body  -> MkType : computeArity body
+    LamAbs _ _ _ body -> (first ((:) MkTerm) (computeArity body))
```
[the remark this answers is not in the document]
michaelpj: Or just 
```
let (args', body') = computeArity body in (MkTerm:args', body')
```
Nothing wrong with a let binding and some pattern-matching.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
-    _                 -> []
+    tm                -> ([],tm)
+
+-- | A term or type argument.
+data Args tyname name uni fun ann =
+  MkTermArg (Term tyname name uni fun ann)
+  | MkTypeArg (Type tyname uni ann)
+
+-- | A list of type or term argument(s) that are being applied.
+type ArgOrder tyname name uni fun ann = [Args tyname name uni fun ann]
```
michaelpj, line 134: And _this_ is 'Args'!

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+
+-- | A pair of argument and the annotation of the term being applied to,
```
michaelpj, line 136: I have a feeling we could do this more simply but I'll need to think about it.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+-- so a term that was traversed can be built back with `mkApps`.
+type ArgOrderWithAnn tyname name uni fun ann =
```
michaelpj, line 138: Also the naming confused me again: it's not with _an_ ann, each one has an ann. So following my previous naming suggestions this would be `ArgsWithAnns`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -16,86 +16,153 @@ import PlutusIR.Purity (firstEffectfulTerm, isPure)
...
+e.g. consider the term @let id = \x -> x in id@: the variable @id@ has syntactic
+arity @[]@, but does in fact need an argument before it does any work.
+-}
+type Arity = [TermOrType]
+
+-- | An ordered list of the term and type arguments applied to a function.
+type AppOrder = [TermOrType]
+
+-- | Info attached to a let-binding needed for call site inlining.
+data CalledVarInfo tyname name uni fun ann =
```
michaelpj, line 78: I think the doc and naming seem a bit confusing. We don't know that the variables are "called". Why not just `VarInfo`?

Or, indeed, we could follow the paper and call it the "in-scope set". Or at least let's refer to the fact that this is the "in-scope set" from the paper.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+    (MkType, MkTypeArg _) -> isFullyApplied (init lamOrder) (init argsOrder)
+    _                     -> False
+
+-- | Inline fully applied functions iff the body of the function is `acceptable`.
+inlineSat :: forall tyname name uni fun ann. InliningConstraints tyname name uni fun
+    => Term tyname name uni fun ann -- ^ The `body` of the `Let` term.
+    -> InlineM tyname name uni fun ann (Term tyname name uni fun ann)
+inlineSat appOrTyInstTm =
+    case appOrTyInstTm of
+      -- If the term is a term or type application, check it's applying to a var that we may inline
```
michaelpj, line 179: why do we need this check? Shouldn't everything work gracefully in the case of 0 arguments?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+  case (last lamOrder, last argsOrder) of
+    (MkTerm, MkTermArg _) -> isFullyApplied (init lamOrder) (init argsOrder)
+    (MkType, MkTypeArg _) -> isFullyApplied (init lamOrder) (init argsOrder)
+    _                     -> False
+
+-- | Inline fully applied functions iff the body of the function is `acceptable`.
+inlineSat :: forall tyname name uni fun ann. InliningConstraints tyname name uni fun
+    => Term tyname name uni fun ann -- ^ The `body` of the `Let` term.
+    -> InlineM tyname name uni fun ann (Term tyname name uni fun ann)
+inlineSat appOrTyInstTm =
```
michaelpj, line 177:
```suggestion
inlineSat t = do
  let (f, args) = collectArgs t
  varInfoMap <- gets ...
  case f of 
     Var n | Just varInfo <- lookupVarInfo n varInfoMap = ...
```
could even put the `isAcceptable` checks into the guards, but maybe that's too much.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+            case argsAppliedTo of
+              -- if it is a `Var` that is being applied to, check to see if it's fully applied
+              Var _ann name -> do
+                maybeVarInfo <- gets (lookupCalled name)
+                case maybeVarInfo of
+                  -- the variable maybe a *recursive* let binding, in which case we don't process,
+                  -- and it won't be in the map. ATM recursive bindings aren't inlined.
+                  Nothing -> pure tm
+                  Just varInfo -> do
+                    isAcceptable <- acceptable (calledVarBody varInfo)
```
michaelpj, line 198: Hmm, why is `acceptable` monadic? It shouldn't be, looking at the definition.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+isFullyApplied [] [] = True
+isFullyApplied lamOrder argsOrder =
+  -- start comparing from the end because there may be over-application
+  case (last lamOrder, last argsOrder) of
+    (MkTerm, MkTermArg _) -> isFullyApplied (init lamOrder) (init argsOrder)
+    (MkType, MkTypeArg _) -> isFullyApplied (init lamOrder) (init argsOrder)
+    _                     -> False
+
+-- | Inline fully applied functions iff the body of the function is `acceptable`.
+inlineSat :: forall tyname name uni fun ann. InliningConstraints tyname name uni fun
```
michaelpj, line 174:
```suggestion
tryInlineSat :: forall tyname name uni fun ann. InliningConstraints tyname name uni fun
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -170,16 +167,24 @@ This might mean reinventing GHC's OccAnal...
...
     handleTerm ::
-        Term tyname name uni fun a
-        -> InlineM tyname name uni fun a (Term tyname name uni fun a)
+        Term tyname name uni fun ann
+        -> InlineM tyname name uni fun ann (Term tyname name uni fun ann)
     handleTerm = \case
         v@(Var _ n) -> fromMaybe v <$> substName n
-        Let a NonRec bs t -> do
+        -- we need to check at an `Apply` note if there is a saturated function to be inlined.
+        appTerm@(Apply _varAnn _fun _arg) -> do
```
michaelpj, line 180: IMO both these cases should just go away in favour of what I wrote below.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -16,86 +16,153 @@ import PlutusIR.Purity (firstEffectfulTerm, isPure)
...
+  MkCalledVarInfo {
+    calledVarDef    :: Term tyname name uni fun ann -- ^ its definition
```
michaelpj, line 80: Is this processed or not? that's very important!

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -190,27 +195,37 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
             t' <- processTerm t
             -- Use 'mkLet': we're using lists of bindings rather than NonEmpty since we might
             -- actually have got rid of all of them!
-            pure $ mkLet a NonRec bs' t'
+            pure $ mkLet ann NonRec bs' t'
         -- This includes recursive let terms, we don't even consider inlining them at the moment
         t -> forMOf termSubterms t processTerm
```
michaelpj, line 200:
```suggestion
        t -> do
             t' <- tryInlineSat t
             forMOf termSubterms t' processTerm
```
Writing it like this exposes an important subtlety that we have to think about: do we want to continue processing the term after we've inlined the head of an application? i.e. in which order do we want to run these steps? It won't affect our inlining decision here, but it will affect how much work we do. This also depends on whether or not we have processed the RHS of the binding that we put into the `VarInfo`. See p11 of the paper for why it's important.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -223,17 +238,44 @@ We rename both terms and types as both may have binders in them.
...
-    TermBind a s v@(VarDecl _ n _) rhs -> do
-        maybeRhs' <- maybeAddSubst body a s n rhs
-        pure $ TermBind a s v <$> maybeRhs'
-    TypeBind a v@(TyVarDecl _ n _) rhs -> do
+    :: forall tyname name uni fun ann. InliningConstraints tyname name uni fun
+    => Term tyname name uni fun ann -- ^ The body of the let binding.
+    -> Binding tyname name uni fun ann -- ^ The binding.
+    -> InlineM tyname name uni fun ann (Maybe (Binding tyname name uni fun ann))
+processSingleBinding body =
+    -- when we encounter a binding that is a function, we add it to the global map
```
michaelpj, line 246: Why not just do it for everything?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -223,17 +238,44 @@ We rename both terms and types as both may have binders in them.
...
+    -- `Utils.CalledVarEnv`. When we check the body of the let binding we look in this map for
+    -- call site inlining.
+    let addVarToMap letBody v ann s n rhs = do
+            -- we want to do unconditional inline if possible
+            maybeRhs' <- maybeAddSubst letBody ann s n rhs
```
michaelpj, line 251: okay, so it's processed.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -223,17 +238,44 @@ We rename both terms and types as both may have binders in them.
...
+    in
+    \case
+    -- The let binding is a function type here.
+    -- If it's not unconditionally inlined, we add it to the `CalledVarEnv`
+    -- and consider whether we want to inline at the call site.
+    TermBind ann s v@(VarDecl _ n (TyFun _ _tyArg _tyBody)) rhs -> addVarToMap body v ann s n rhs
+    -- The let binding is a type lambda abstraction.
+    TermBind ann s v@(VarDecl _ n _) rhs@(TyAbs {}) -> addVarToMap body v ann s n rhs
+    -- for binding that aren't functions, maybe unconditional inline
+    TermBind ann s v@(VarDecl _ n _) rhs -> do
```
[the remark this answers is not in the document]
michaelpj: YES!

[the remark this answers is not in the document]
michaelpj: Plus then you can inline `addVarToMap`, which will simplify things a lot.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
michaelpj: I would not expect this function to also modify the substitution! It sounds like it's just adding the `VarInfo` but it isn't.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -223,17 +238,44 @@ We rename both terms and types as both may have binders in them.
...
+                -- inline and remove binding when the conditions for unconditional inlining are met.
+                Nothing -> pure $ TermBind ann s v <$> maybeRhs'
+                Just rhsProcess -> do
+                    let
+                        varLamOrder = fst $ computeArity rhs
+                        bodyToCheck = snd $ computeArity rhs
+                    -- add the function to `CalledVarEnv`, because we may want to inline this.
+                    -- We don't remove the binding because we decide *at the call site*
+                    -- whether we want to inline, and it may be called more than once.
+                    void $ modify' $ extendCalled n (MkCalledVarInfo rhs varLamOrder bodyToCheck)
```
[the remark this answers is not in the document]
michaelpj: I think conceptually we want to do that check at the call site. I think it's pretty likely that in future our acceptability check might be context-sensitive.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
michaelpj: Okay, so the RHS in the `VarInfo` is processed. Then we should first of all process the subterms of the application and _then_ consider saturated inlining, since otherwise we will re-process the RHS.

plutus-core/plutus-ir/test/transform/computeArityTest/absapp.golden
```diff
@@ -1 +1 @@
-[]
\ No newline at end of file
+([], { (abs a (type) (lam x a x)) (con integer) })
```
michaelpj, line 1: I'm unsure why this test is interesting - did you mean to test something different?

michaelpj: Oh I see, it's just a copy of our set of random examples. I'd really prefer to have some specific interesting tests for this function, these don't seem that useful.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+  -> ArgOrderWithAnn tyname name uni fun ann
+  -> Term tyname name uni fun ann
+mkApps f ((MkTermArg tmArg, ann) : args) = mkApps (Apply ann f tmArg) args
+mkApps f ((MkTypeArg tyArg, ann) : args) = mkApps (TyInst ann f tyArg) args
+mkApps f []                              = f
+
+-- | Given the arity of a function, and the list of arguments applied to it, return whether it is
+-- fully applied or not.
+isFullyApplied :: Arity -> ArgOrder tyname name uni fun ann -> Bool
+isFullyApplied [] (_argsOrder:_as) = True -- over-application
```
michaelpj, line 163:
```suggestion
isFullyApplied [] (_arg:_args) = True -- over-application
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+isFullyApplied (_arity:_) [] = False -- under-application
+isFullyApplied [] [] = True
+isFullyApplied lamOrder argsOrder =
+  -- start comparing from the end because there may be over-application
```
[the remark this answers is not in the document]
michaelpj: I'm also confused. I don't see why this doesn't just compare the two things at the heads of the list and then recurse.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+  case (last lamOrder, last argsOrder) of
```
michaelpj, line 168: `last` is linear. What does that mean for the overall runtime of this function?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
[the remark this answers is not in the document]
michaelpj: Do we have any test for over/under-application? Not sure that we do.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -16,86 +16,153 @@ import PlutusIR.Purity (firstEffectfulTerm, isPure)
...
+Note that this is the syntactic arity, i.e. it just corresponds to the number of
+syntactic lambda and type abstractions on the outside of the term. It is thus
+an under-approximation of how many arguments the term may need.
+e.g. consider the term @let id = \x -> x in id@: the variable @id@ has syntactic
+arity @[]@, but does in fact need an argument before it does any work.
+-}
+type Arity = [TermOrType]
+
+-- | An ordered list of the term and type arguments applied to a function.
+type AppOrder = [TermOrType]
```
michaelpj, line 75: Dead?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -16,86 +16,153 @@ import PlutusIR.Purity (firstEffectfulTerm, isPure)
...
+-- | Info attached to a let-binding needed for call site inlining.
+data CalledVarInfo tyname name uni fun ann =
+  MkCalledVarInfo {
+    calledVarDef    :: Term tyname name uni fun ann -- ^ its definition
+    , arity         :: Arity -- ^ its sequence of term and type lambdas
+    , calledVarBody :: Term tyname name uni fun ann
+    -- ^ the body of the function, for checking `acceptable` or not
+  }
+-- | Is the next argument a term or a type?
+data TermOrType =
```
michaelpj, line 86: Naming suggestion. At the moment you have two lists of things that get kind of described as "arg lists": the arity, and the arguments at the call site. I think the usual CS terminology is to call the first thing "parameters" and the second thing "arguments". So how about:
```
data ParamKind
  TermParam | TypeParam
```

michaelpj (changes requested):
So I think the big high-level conceptual point is that you're making life a lot harder for yourself by not making things a tiny bit more general. You go to a lot of effort to avoid having things with "0 arity". But if we can just handle that case gracefully, then there is no problem. And indeed we can: things with 0 arity will always be fully-applied, but then they face exactly the same `isAcceptable` check as they would otherwise, and the argument for why it's a good idea is just as valid. Plus, we get to say that we have `VarInfo` for all in-scope variables (we could even call it.. the in-scope set ;) ). So by making things a tiny bit more general we save ourselves a lot of work.

michaelpj:
> renamed UnconditionalInline.hs to Inline.hs because the call site inlining code is embedded in the unconditional inline function from before.

TBH, I was never entirely convinced that the split would be workable since are going to always want a single main function for the inliner :shrug: I don't think we should have `Inline/Inline.hs`, though. Sometimes we do `Foo/Internal.hs`. Could do `Driver.hs` as it's the "driver" for the inliner...

You could have put the `preInlineUnconditional` functions into `UnconditionalInline.hs` but you inlined those in your last PR :)

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -16,86 +16,153 @@ import PlutusIR.Purity (firstEffectfulTerm, isPure)
...
+
 -- | Substitution for both terms and types.
 -- Similar to 'Subst' in the paper, but the paper only has terms.
-data Subst tyname name uni fun a = Subst { _termEnv :: TermEnv tyname name uni fun a
-                                         , _typeEnv :: TypeEnv tyname uni a
-                                         }
+data Subst tyname name uni fun ann =
+    Subst { _termEnv       :: TermEnv tyname name uni fun ann
+           , _typeEnv      :: TypeEnv tyname uni ann
+           , _calledVarEnv :: CalledVarEnv tyname name uni fun ann
```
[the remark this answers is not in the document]
thealmarty: To reduce the number of layers, instead of having one field for substitution, I renaming `Subst`  to `InlinerContext` and `termEnv` to `termSub` and `typeEnv` to `typeSub`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -6,30 +6,47 @@
...
 
 (1) What do we mean by "fully applied"?
 
+A function is fully applied when it has been applied to all arguments as indicated by its
+syntactic arity.
+
 Consider `let v = rhs in body`, in which `body` calls `v`.
 
-We consider cases when `v` is a function/lambda abstraction(s). I.e.:
+We consider cases when `v` is a function. I.e., it has type/term lambda abstraction(s). I.e.:
```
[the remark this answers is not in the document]
thealmarty: Yes but that's not very interesting. Did you want me to talk about that non-functions have arity 0?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
 
 If _v_ is fully applied in the body, i.e., if the sequence of type and term lambda abstractions is
```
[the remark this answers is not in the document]
thealmarty: Yeah sorry this was talking about the implementation but yeah I see that it's a bit repetitive.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+that are used in unconditional inlining. For
 
 let v = \x1.\x2...\xn.VBody in body
 
 we check `VBody` with the above "acceptable" functions.
-Note that all `LamAbs` and `TyAbs` should have been
+Note that all `LamAbs` and `TyAbs` are not acceptable in terms of size, but they should have been
 counted out already so we should not immediately encounter those in `VBody`.
 Also, we currently reject `Constant` (has acceptable cost but not acceptable size).
 We may want to check their sizes instead of just rejecting them.
```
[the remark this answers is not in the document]
thealmarty: I thought it makes sense to discuss some of that here because the reader may wonder why we chose the heuristics - I'm saying we haven't much yet, we'll fine-tune it. Maybe I should make that more clear.

plutus-core/plutus-ir/test/transform/inline/letOverAppMulti.golden
thealmarty: @michaelpj this is the test for over-application.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+isFullyApplied :: Arity -> ArgOrder tyname name uni fun ann -> Bool
+isFullyApplied [] (_argsOrder:_as) = True -- over-application
+isFullyApplied (_arity:_) [] = False -- under-application
+isFullyApplied [] [] = True
+isFullyApplied lamOrder argsOrder =
+  -- start comparing from the end because there may be over-application
+  case (last lamOrder, last argsOrder) of
+    (MkTerm, MkTermArg _) -> isFullyApplied (init lamOrder) (init argsOrder)
+    (MkType, MkTypeArg _) -> isFullyApplied (init lamOrder) (init argsOrder)
+    _                     -> False
```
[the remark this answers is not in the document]
thealmarty: Not sure. How much/far do we allow things to be polymorphic? Need to look at the type checker. For this instance, having a `False` here should be fine though?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -170,16 +167,24 @@ This might mean reinventing GHC's OccAnal...
...
-        -> InlineM tyname name uni fun a (Term tyname name uni fun a)
+        Term tyname name uni fun ann
+        -> InlineM tyname name uni fun ann (Term tyname name uni fun ann)
     handleTerm = \case
         v@(Var _ n) -> fromMaybe v <$> substName n
-        Let a NonRec bs t -> do
+        -- we need to check at an `Apply` note if there is a saturated function to be inlined.
+        appTerm@(Apply _varAnn _fun _arg) -> do
+            inlinedSat <- inlineSat appTerm
+            forMOf termSubterms inlinedSat processTerm
```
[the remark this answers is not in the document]
thealmarty: I meant for `inlineSat` to not do that actually. Thanks.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -79,97 +82,124 @@ Then, in the _body_, we track the term and type application (`Apply` or `TyInst`
...
+inlineSat :: forall tyname name uni fun ann. InliningConstraints tyname name uni fun
+    => Term tyname name uni fun ann -- ^ The `body` of the `Let` term.
+    -> InlineM tyname name uni fun ann (Term tyname name uni fun ann)
+inlineSat appOrTyInstTm =
+    case appOrTyInstTm of
+      -- If the term is a term or type application, check it's applying to a var that we may inline
+      Apply _varAnn _fun _arg  -> go appOrTyInstTm
+      TyInst _varAnn _fun _arg -> go appOrTyInstTm
+      -- otherwise, check all subterms
+      _                        -> forMOf termSubterms appOrTyInstTm inlineSat
```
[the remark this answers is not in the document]
thealmarty: I'm not completely sure I got everything you said, but yes, I meant for `inlineSat` to not do the subterms, they are done in the `inline` function.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: good point. Fixed.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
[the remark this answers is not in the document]
thealmarty: I can reverse the lists and do the regular thing. I thought if the arg list is long this may be wasteful but as Michael pointed out this is not fast either.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: see above.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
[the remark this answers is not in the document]
zliu41: > As for collectArgs, it returns an ArgOrder with its head being the outer most application

I don't think this is true. Given `fun a b c d e f`, your `collectArgs` returns `(fun, [a, b, c, d, e, f])`, as far as I can tell. So you should be comparing the heads of both lists. If `fun` is the function you wrote above, then `a` should be compared against `MkType`, and `f` should be ignored.

And if I was wrong and `collectArgs` actually returns `(fun, [f, e, d, c, b, a])`, then it is still incorrect because you should compare the last element of one list against the first element of the other list.

thealmarty:
I'm still working on addressing the comments, I'll re-request review when I'm ready.

thealmarty:
OK CI seems to be working and thus we can revert [this commit](https://github.com/input-output-hk/plutus/commit/27c62375479fa028032a571596b52ee3ea6f2d01) safely afterwards.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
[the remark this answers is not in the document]
thealmarty: OK I changed to comparing the heads and it didn't change the test results. I don't know why it *was* working - maybe the tests aren't covering it properly...

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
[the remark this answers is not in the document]
thealmarty: OK yes I can add a check for arity 0.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: yes, it should indeed.

plutus-core/plutus-ir/test/transform/computeArityTest/absapp.golden
thealmarty: Yeah this was to show type applications won't be counted. Maybe a bit overkill.

plutus-core/plutus-core.cabal
```diff
@@ -447,7 +447,7 @@ library plutus-ir
     PlutusIR.Transform.Beta
     PlutusIR.Transform.DeadCode
     PlutusIR.Transform.Inline.CallSiteInline
-    PlutusIR.Transform.Inline.UnconditionalInline
+    PlutusIR.Transform.Inline.Inline
```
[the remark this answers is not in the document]
thealmarty: I'll fix that in another PR. I'll probably rename this to `Driver.hs`, but then I'll also need to move the unconditional inlining specific functions to a `Unconditional.hs` module. Does that sound good? @zliu41

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -16,86 +16,152 @@ import PlutusIR.Purity (firstEffectfulTerm, isPure)
...
+data ParamKind =
+    TermParam | TypeParam
+    deriving stock (Eq, Show)
+
+instance Pretty ParamKind where
+  pretty = viaShow
+
+-- | Inliner context for both unconditional inlining and call site inlining.
+-- It includes substitution for both terms and types, which is similar to 'Subst' in the paper.
+-- It also includes the in-scope set for call site inlining.
```
michaelpj, line 92: :+1:

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -6,30 +6,47 @@
...
 let v = \x1.\x2...\xn.VBody in body
 
-In the `body`, where `v` is *called*,
-if it was given `n` arguments, then it is _fully applied_ in the `body`.
+(x1,x2...xn) is the syntactic arity of a term. That is, a record of the arguments that the
+term expects before it may do some work. Since we have both type and lambda
+abstractions, this is not a simple argument count, but rather a list of values
+indicating whether the next argument should be a term or a type.
+
+Note that this is the syntactic arity, i.e. it just corresponds to the number of
```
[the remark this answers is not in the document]
michaelpj: Sure, but it has to fit into the overall prose!

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
michaelpj: Well, you fixed it but it was quadratic. Really worth noticing such things. Any time you use a function whose runtime is at least linear in the size of the structure while traversing the structure an alarm bell should go off.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -53,123 +69,122 @@ in
...
+-- fully applied or not.
+isFullyApplied :: Arity -> ArgOrder tyname name uni fun ann -> Bool
+isFullyApplied [] (_arg:_args) = True -- over-application
+isFullyApplied (_lam:_lams) [] = False -- under-application
+isFullyApplied [] [] = True
+isFullyApplied (hdLams:tlLams) (hdArg:tlArg) =
+  case (hdLams, hdArg) of
+    (TermParam, MkTermArg _) -> isFullyApplied tlLams tlArg
+    (TypeParam, MkTypeArg _) -> isFullyApplied tlLams tlArg
+    _                        -> False
```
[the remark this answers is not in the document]
thealmarty: Good point. I wonder if it would help in some cases to only check the number of arguments...

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -16,86 +16,152 @@ import PlutusIR.Purity (firstEffectfulTerm, isPure)
...
 
--- | Substitution for both terms and types.
--- Similar to 'Subst' in the paper, but the paper only has terms.
-data Subst tyname name uni fun a = Subst { _termEnv :: TermEnv tyname name uni fun a
-                                         , _typeEnv :: TypeEnv tyname uni a
-                                         }
+-- For call site inlining:
+
+-- | A mapping including all let-bindings that are functions.
+newtype InScopeSet tyname name uni fun ann =
```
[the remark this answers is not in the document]
thealmarty: OK sure. I guess it shouldn't blow up or anything.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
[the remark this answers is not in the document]
thealmarty: As Michael suggested I'm storing all non recursive variables in the map.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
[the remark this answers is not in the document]
thealmarty: Maybe this feels less repetitive now?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -223,39 +233,55 @@ We rename both terms and types as both may have binders in them.
...
-    TypeBind a v@(TyVarDecl _ n _) rhs -> do
+    (TermBind ann s v@(VarDecl _ n _) rhs) -> do
+        -- we want to do unconditional inline if possible
+        maybeRhs' <- maybeAddSubst body ann s n rhs
+        case maybeRhs' of
+            -- inline and remove binding when the conditions for unconditional inlining are met.
+            Nothing -> pure Nothing
+            Just rhsProcess -> do
+                let (varArity, bodyToCheck) = computeArity rhs
+                -- if the variable is not a function, don't add it to the map
```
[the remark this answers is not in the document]
thealmarty: OK sure. Even though they're not used atm (1) it's tidier to have the whole non-rec in scope set (2) we may have use for this later.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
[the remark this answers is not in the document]
thealmarty: Added a comment.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: I think what I have now is also fine?

thealmarty:
> The names of the test cases are hard to parse, e.g., what is `letFunInFunMultiFullyApplied`? Consider renaming them so that it is easier to answer questions like these: (1) which test cases involve type abstractions? (2) which test cases verify that `costIsAcceptable` and `sizeIsAcceptable` are respected? (3) which test cases involve strict vs. nonstrict bindings, pure vs. impure bindings, or nonrec vs. rec bindings?

It's kinda already like that, e.g., `Multi` in the name meaning it involves multiple occurrences, `OverApp` is testing over applications etc. I will see if I can improve it more.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -223,39 +234,53 @@ We rename both terms and types as both may have binders in them.
...
-        maybeRhs' <- maybeAddSubst body a s n rhs
-        pure $ TermBind a s v <$> maybeRhs'
-    TypeBind a v@(TyVarDecl _ n _) rhs -> do
+    (TermBind ann s v@(VarDecl _ n _) rhs) -> do
+        -- we want to do unconditional inline if possible
+        maybeRhs' <- maybeAddSubst body ann s n rhs
+        case maybeRhs' of
+            -- this binding is going to be unconditionally inlined
+            Nothing -> pure Nothing
+            Just rhsProcess -> do
```
zliu41, line 248: weird name - it sounds like an action to be taken on `rhs`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -190,27 +187,41 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
-    -- See Note [Inlining approach and 'Secrets of the GHC Inliner']
-    -- Already processed term, just rename and put it in, don't do any further optimization here.
-    renameTerm ::
-        InlineTerm tyname name uni fun a
-        -> InlineM tyname name uni fun a (Term tyname name uni fun a)
-    renameTerm (Done t) = liftDupable t
+        t -> do
+            -- process all subterms
+            processedT <- forMOf termSubterms t processTerm
+            -- consider call site inlining for each node that have been unconditionally inlined
```
zliu41, line 195: I think you mean "have _not_ been unconditionally inlined"

zliu41 (commented):
I think you are considering 0-arity bindings to be always fully applied. If so, then you can't just check `acceptable`, but you must also check for purity (non-0-arity bindings are always pure). Please also add a test case involving non-pure 0-arity bindings.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -190,27 +188,39 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
-    substName :: name -> InlineM tyname name uni fun a (Maybe (Term tyname name uni fun a))
-    substName name = gets (lookupTerm name) >>= traverse renameTerm
-    -- See Note [Inlining approach and 'Secrets of the GHC Inliner']
-    -- Already processed term, just rename and put it in, don't do any further optimization here.
-    renameTerm ::
-        InlineTerm tyname name uni fun a
-        -> InlineM tyname name uni fun a (Term tyname name uni fun a)
-    renameTerm (Done t) = liftDupable t
+        t -> do
+            processedT <- forMOf termSubterms t processTerm
```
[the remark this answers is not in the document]
michaelpj: Also it's not generally quadratic! We only recurse into subterms once: here. So it's traversing everything once in the main loop.

What _is_ slightly risky is that `inlineSat` does a traversal looking downwards to see if an application is saturated. And because we process subterms first, we will do this for each prefix of the application first. So the runtime _is_ quadratic in the length of the largest saturated applications, but we just have to live with that, I think. Absolutely worth a comment, though.

e.g. what happens in this program?
```
let x = \a1 ... an . 1
in x 1 2 ... n
```

Also worth noting that this _won't_ happen in the context-based version (why?), which is a disadvantage of our approach.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -53,123 +69,124 @@ in
...
+-- | Given the arity of a function, and the list of arguments applied to it, return whether it is
+-- fully applied or not.
+isFullyApplied :: Arity -> Args tyname name uni fun ann -> Bool
+isFullyApplied [] _ = True -- The function needs no more arguments
+isFullyApplied (_lam:_lams) [] = False -- under-application
+isFullyApplied (hdLams:tlLams) (hdArg:tlArg) =
+  case (hdLams, hdArg) of
+    (TermParam, MkTermArg _) -> isFullyApplied tlLams tlArg
+    (TypeParam, MkTypeArg _) -> isFullyApplied tlLams tlArg
+    _                        ->
```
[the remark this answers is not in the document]
michaelpj: I asked the same here: https://github.com/input-output-hk/plutus/pull/5217#discussion_r1151772400

I think it's marginally better to leave things alone in this case: something bad is going on, and we can just decide to do nothing rather than try and make an argument about how we're not making things worse. But the comment should make that clear.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -53,123 +69,124 @@ in
...
+mkApps :: Term tyname name uni fun ann
+  -> ArgsWithAnns tyname name uni fun ann
+  -> Term tyname name uni fun ann
+mkApps f ((MkTermArg tmArg, ann) : args) = mkApps (Apply ann f tmArg) args
+mkApps f ((MkTypeArg tyArg, ann) : args) = mkApps (TyInst ann f tyArg) args
+mkApps f []                              = f
+
+-- | Given the arity of a function, and the list of arguments applied to it, return whether it is
+-- fully applied or not.
+isFullyApplied :: Arity -> Args tyname name uni fun ann -> Bool
```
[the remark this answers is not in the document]
thealmarty: Yeah frankly I should, the current test case that tests this also magically failed and not as I apply comment suggestions. I'm not happy about that.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
thealmarty: hm...I meant that have gone through the unconditional inliner. I'll revise that.

plutus-core/plutus-core.cabal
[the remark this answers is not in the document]
thealmarty: any thoughts @michaelpj @zliu41? I also wonder if I should put `Deadcode.hs` and `Beta.hs` in the `inline` folder too.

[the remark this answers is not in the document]
michaelpj: Um, why? They're not inlining?

thealmarty:
@zliu41 @michaelpj regarding always inlining 0-arity bindings: since we do unconditional inlining before `considerInlineSat`, the nodes that are going to `considerInlineSat` occur multiple times. So they are either `MultiSafe` or `MultiUnsafe`, depending on whether or not it may occur inside lambdas. I'm not sure whether just checking "purity" and "acceptable" is sufficient? I mean, right now it probably is because "acceptable" is very conservative. Maybe we can do that and add a comment.
And yes I will add tests as well.

plutus-core/plutus-core.cabal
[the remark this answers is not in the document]
thealmarty: Makes sense. Is it ok if I take them out of the comment then? Maybe I'll add your 2. in the comment actually.

thealmarty:
Need to investigate why a few of the tests now don't get inlined when non-pure binding are not inlined.

thealmarty:
The latest commit shows that NOT checking purity does NOT make the inliner inline the non-pure binding. Need to investigate further. I'll also look into our purity function more - it doesn't seem to do what we think it does atm.

thealmarty:
Wait nvm, applications are not acceptable, so it won't inline either way. Need a better test. But the other tests still shouldn't fail.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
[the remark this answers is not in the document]
thealmarty: Tests added.

thealmarty:
OK, other than not being able to prove that we **do** inline non-pure variables when we **don't** check purity, it's done. I can spend more time figuring it out in detail or we can merge this - there is currently no undesired behaviour.

thealmarty:
Actually nvm, variables were never added to the map, that's why. I can add them though.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -53,123 +69,133 @@ in
...
+            -- if the body of `Var` is `acceptable` and
+            -- it is fully applied (over-application is allowed) then
+            if isAcceptable && isFullyApplied (arity varInfo) (map fst args)
+            then do
+              -- check if term is pure. Because non-functions or binding with arity 0 are always
+              -- fully applied, we only want to inline them if they are also pure.
+              -- For strict bindings we can't accidentally make any effects happen less often than
+              -- it would have before, but we can make it happen more often.
+              -- We could potentially do this safely in non-conservative mode.
+              isTermPure <- checkPurity v
```
michaelpj, line 195: This is checking the wrong thing? We want to check if the RHS of the binding we're about to inline is pure, i.e. `varDef varInfo`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -53,123 +69,133 @@ in
...
+          -- and we don't process it. ATM recursive bindings aren't inlined.
+          Nothing -> pure tm
+          Just varInfo -> do
+            let isAcceptable = acceptable (varBody varInfo)
+            -- if the body of `Var` is `acceptable` and
+            -- it is fully applied (over-application is allowed) then
+            if isAcceptable && isFullyApplied (arity varInfo) (map fst args)
+            then do
+              -- check if term is pure. Because non-functions or binding with arity 0 are always
+              -- fully applied, we only want to inline them if they are also pure.
```
michaelpj, line 191: I don't think that's the reason. In fact, the justification is independent of the artity. We just don't want to inline impure things because that can duplicate effects. 

And as previously discussed, it depends on the strictness of the binding too, which you need to track in the `VarInfo`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -273,24 +306,24 @@ maybeAddSubst body a s n rhs = do
             -- See Note [Inlining approach and 'Secrets of the GHC Inliner'] and [Inlining and
             -- purity]. This is the case where we don't know that the number of occurrences is
             -- exactly one, so there's no point checking if the term is immediately evaluated.
-            postUnconditional <- fmap ((&&) termIsPure) (acceptable rhs')
+            let postUnconditional = isTermPure && acceptable rhs'
```
michaelpj, line 309: Unrelated to this PR, but I think this line is wrong? Even if something occurs multiple times, we don't need to care if it's pure for _non-strict_ bindings, because the effects would have occurred at each use site regardless.

michaelpj (commented):
I think you just have a small bug.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
thealmarty: Yeah this is unrelated to this PR but I can fix it here I guess. I'll add a `wontDuplicateEffect`, in which I check the strictness first. If it's non-strict, then it's `True` and go on to checking `purity`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
michaelpj: I'm confused. Yes, we check purity as a proxy for duplicating effects often, but there's nothing wrong with doing it here?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
michaelpj: Well, the criterion is "is this _binding_ pure, i.e. will evaluating it have effects", so maybe you just want `isTermBindingPure :: Strictness -> Term -> Bool`  or whatever.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: I'm just saying that I'll say that I'm checking `wontDuplicateEffect` here, which is what I really want, which includes checking purity AND strictness. Am I making sense? (same as the comment below)

thealmarty: OK I just saw your other comment, so I'll do `isTermBindingPure` here as well.

thealmarty:
@michaelpj please approve so I can merge.

I've fixed the bug but the non-pure tests still don't fail when we don't checking if the binding is pure. But this is actually right: the binding got unconditionally inlined! (By postUnconditional) And so it is removed. And therefore, the call site inlining won't have seen it because we try to do unconditional inlining first.

michaelpj (approved):

plutus-core/plutus-ir/test/transform/inline/letNonPureMultiStrict
```diff
@@ -0,0 +1,21 @@
+{- 
+This test checks that impure bindings are not inlined by call site inlining.
+y is strict in this case.
+
+let y = trace "hello" 1
+    z = y
+    x = y -- so it looks small enough to inline
```
michaelpj, line 7: since y is a strict variable this is in fact pure.  You want to make the binding for y non-strict in this case.

Issue #4808, opened 2022-08-11
Pretty-printing uniques explicitly


effectfully:
### Summary

Currently the unique of a variable is not pretty-printed with the default strategy, which I believe is in fact a bug, because parsing a pretty-printed program may get you back a _different_ program. Not up to alpha, an actually different program. A program with a free variable is almost guaranteed to trigger this, however it doesn't need to be free for that, consider pretty-printing and parsing back `\x_0 x_1 -> x_0` -- you'll get `\x_0 x_1 -> x_1` back.

Now we could just pretty-print the uniques by default, but this has its own problems: it practically changes the names, because syntactically a pretty-printed unique is indistinguishable from a normal part of a name (i.e. `x_0` was either the `x` name with the pretty-printed `0` unique or the `x_0` name without a pretty-printed unique). Randomly changing names confuse people, invalidate in-memory environments and cause other kinds of disruption.

Thanks to [Runtime Verification Inc.](https://github.com/runtimeverification) for reaching out to us regarding these issues.

### Steps to reproduce the behavior

1. pretty-print the AST of `\x_0 x_1 -> x_0` using the default strategy
2. parse it back

### Actual Result

got a different term back

### Expected Result

the resulting term is alpha-equal to the original one

### Describe the approach you would take to fix this

We can solve all these problems by always pretty-printing the uniques and not making them a syntactically legal part of the textual name of the variable. For example by pretty-printing the `x` variable having the `0` unique as `x!0` or `x#0` or whatever (suggestions are welcome).

### System info

Plutus: 1203fa3f63db6f949af4bba889a0d0691921d2af

michaelpj:
> We can solve all these problems by always pretty-printing the uniques and not making them a syntactically legal part of the textual name of the variable.

And by parsing this format as well, obviously.

bezirg:
mpj wrote on SCP-817:

PR #5218, opened 2023-03-20
Allow symbols in PLC names

zliu41:
Doesn't seem harmful, and improves readability.


effectfully (commented):
I have very mixed feelings about this:

1. can it clash with the readable pretty-printer? Maybe the latter omits spaces sometimes
2. `$f`, `$c` etc stuff is noisy and not helpful, it would be ideal if we parsed it and turned it into something readable
3. I think it's an improvement overall, because I didn't like throwing away useful information, in particular turning `==` into `bad_name` etc

I feel like the best solution would be to allow arbitrary ASCII symbols in identifies, but quote those identifiers that are not composed or alphanumerics, `_` and `'`. We could use `` ` `` in the beginning of an identifier containing symbols to express "everything up until the next space is a single identifier". And yeah, I would like us to parse `$f`, `$c` etc.

michaelpj:
I do like getting rid of `bad_name`.

Main question: does the parser handle these? I don't think we should allow non-parseable identifiers.

> And yeah, I would like us to parse $f, $c etc.

I'm not sure what you mean by that? Do you mean "use our domain knowledge of GHC to reverse-engineer the meaning of those names and emit a more human readable version"? e.g. "$c becomes '_specialized'" or something? I'm not at all sure about that - I think it's a lot simpler to just reflect what we get from GHC; then at least knowledge about GHC transfers over, rather than it being a mix of GHC and our (probably partial) attempts to back-translate it.

zliu41:
> can it clash with the readable pretty-printer? Maybe the latter omits spaces sometimes

There's no need to have space between identifier names and non-identifier names (e.g., brackets). If it omits spaces between two things that both can be identifier names, then it is the pretty printer that should be fixed.

> $f, $c etc stuff is noisy and not helpful, it would be ideal if we parsed it and turned it into something readable

Perhaps, but on the other hand these are standard in GHC Core, so will look more familiar to people who are used to it.

> I think it's an improvement overall, because I didn't like throwing away useful information, in particular turning == into bad_name etc

Not only that, but currently, `$fInteger_$c>`, `$fInteger_$c<` etc., all become `fInteger_c` which is quite annoying.

kwxm:
I like this because it would allow me to have built-in function names with dots in them, which I was asking about the other day!  I think there's no danger of breaking anything because the syntax of Plutus Core, for all its faults, is very minimal and doesn't make any use of the symbols that this PR makes valid in names.  However, one argument against it might be that if someone like me was to go ahead and  define builtins with strange characters in their names (which I think can be done simply by producing a hand-written `Show` instance for the DefaultFun type) then we could have builtins whose literal names are significantly different from the name of the constructor that represents them internally, which could lead to confusion.  Perhaps I should just be forbidden from doing that.  I suppose this might also be bad for anyone who might have implemented tools that deal with PLC with alternative syntax which might make use of the new symbols we're allowing here.

On the whole, I think I'm in favour of it because of the GCH compatibility problem that it solves.

plutus-core/plutus-core/src/PlutusCore/Parser/ParserCommon.hs
```diff
@@ -114,7 +114,19 @@ inBraces :: Parser a -> Parser a
 inBraces = between (symbol "{") (char '}')
 
 isIdentifierChar :: Char -> Bool
-isIdentifierChar c = isAlphaNum c || c == '_' || c == '\''
+isIdentifierChar c = isAlphaNum c || elem c identifierSymbols
```
kwxm, line 117: I could complain that this should be called `isIdentifierNonStartingChar` or something, but that would be horrible.

kwxm (approved):

effectfully:
> If we do that then we need to make sure there's a space after an identifier name, not a closing bracket or anything else. 

Sorry, I should've said "a delimiter" instead of "a space", of course.

> Also, to me that is a very unusual way of doing identifier names and I'm not aware of any other language that does that.

Some languages have quoted identifiers, except those languages usually mark both the sides of a quoted identifier with a symbol. In PureScript the symbol is `"` (see [this](https://tek.brick.do/pure-script-got-field-names-right-haskell-got-them-wrong-gWrg5VqN9Qyk)) and in SQL it can be `` ` `` (see [this](https://stackoverflow.com/questions/73348354/can-a-backticked-identifier-have-a-backtick-in-it)) or `"` (see [this](https://www.mssqltips.com/sqlservertip/6314/sql-server-set-quotedidentifier-and-set-ansipadding-proper-usage-examples/)).

As for marking only one side, LISP uses a single `'` to denote quotation, albeit a very different kind of quotation.

So I thought we could combine the two strategies. But perhaps they don't combine well indeed and it's worth to mark an identifier at both of its sides to quote it.

zliu41:
> in SQL it can be `

@effectfully I like this and I updated the PR to do so. The additional backticks are not quite aesthetically pleasing, and it adds some complexity to the parser, but it is a more robust way of allowing symbols in names.

plutus-core/changelog.d/20230319_184206_unsafeFixIO_symbols.md
```diff
@@ -0,0 +1,3 @@
+### Changed
+
+- Allow PLC names quoted in backticks. Quoted names may have symbolic characters.
```
michaelpj, line 3:
```suggestion
- The PLC, UPLC, and PIR parsers accept names quoted in backticks. Quoted names may have symbolic characters.
```

plutus-core/plutus-core/src/PlutusCore/Parser/ParserCommon.hs
```diff
@@ -155,7 +167,17 @@ version = trailingWhitespace $ do
...
+        void $ lookAhead (satisfy isIdentifierStartingChar)
+        str <- takeWhileP (Just "identifier-unquoted") isIdentifierChar
+        Name str <$> intern str
+    parseQuoted = do
+        void $ char backTick
+        void $ lookAhead (satisfy isQuotedIdentifierChar)
+        strUnquoted <- takeWhileP (Just "identifier-quoted") isQuotedIdentifierChar
+        void $ char backTick
+        let str = T.cons backTick (T.snoc strUnquoted backTick)
+        Name str <$> intern str
```
michaelpj, line 182: I think I slightly disagree with this approach. This approach says that the parsed name is the string _including_ the backticks. That seems wrong to me: the backticks are part of  the _syntax_ of names, that allow us to give an _actual name_ which has special characters in it. The backticks themselves are not part of the name.

If we view them as syntax then we do things slightly differently:
- The name we parse here is without the backticks
- When we prettyprint a name, we need to decide what syntax to use depending on whether it can be printed unquoted or not

zliu41: Scala is a language I know that supports quoted names. I thought `foo` and `` `foo` `` are different names, but I was wrong - they are the same name. So I agree with you.

zliu41: @michaelpj Updated. The only slight problem is that `(all *_0 (type) *_0)` becomes `` (all `*`_0 (type) `*`_0) `` (see [e1b585e](https://github.com/input-output-hk/plutus/pull/5218/commits/e1b585eb9bbae8440b4e0321a72912d0a690cdf7)). The `*` comes from `dummyTyname`. Can we use something else there, say `a`? That would be less confusing anyway since `*` usually refers to kinds.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Names.hs
```diff
@@ -17,8 +17,9 @@ support unicode identifiers as well.
 
 replacements :: [(T.Text, T.Text)]
 replacements = [
-    -- this helps with module prefixes
-    (".", "_")
+    ("-", "minus")
```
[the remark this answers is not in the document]
zliu41: I've removed `replacements`. I don't think we need it any more.

plutus-core/plutus-core/src/PlutusCore/Parser/ParserCommon.hs
michaelpj: > We can still use a for that. I think GHC invents type variables on similar situations, rather than using *

GHC makes _fresh_ type variables for each such thing, which is fine. We'd have to do something similar, which complicates things.

`any` seems okay.

> I think the behavior of Parser Name has always been wrong. If the name being parsed contains the unique. It would parse name_unique into the name field.

It's kind of dumb, but this is sort of fine when we use that syntax to print uniques: then it effectively produces globally unique names, which parses fine! Not sure if anyone relies on that, though :thinking:

zliu41: > GHC makes fresh type variables for each such thing, which is fine.

We can reserve a special unique for `dummyName`, but I'll just go with `any`.

> It's kind of dumb, but this is sort of fine when we use that syntax to print uniques: then it effectively produces globally unique names, which parses fine! Not sure if anyone relies on that, though thinking

:facepalm: 

Ok, then I think the easiest solution to restore old behavior is to quote around `name_unique`, i.e., `` `foo_i0` `` instead of `` `foo`_i0 ``

michaelpj: I do agree it's super ugly, and this now just makes the ugliness very apparent... not that that helps :shrug:

effectfully: We have an [issue](https://github.com/input-output-hk/plutus/issues/4808) about pretty-printing of uniques. If anybody feels like taking over, that would be highly appreciated.

zliu41: Updated based on comments above.

plutus-core/plutus-core/src/PlutusCore/DeBruijn/Internal.hs
```diff
@@ -119,11 +119,10 @@ newtype TyDeBruijn = TyDeBruijn DeBruijn
 instance Wrapped TyDeBruijn
 
 instance HasPrettyConfigName config => PrettyBy config NamedDeBruijn where
-    prettyBy config (NamedDeBruijn txt0 (Index ix))
-        | showsUnique = pretty txt <> "_i" <> pretty ix
-        | otherwise   = pretty txt
+    prettyBy config (NamedDeBruijn txt (Index ix))
+        | showsUnique = pretty . toPrintedName $ txt <> "_i" <> render (pretty ix)
```
michaelpj, line 123: Given how weird this is, I think it's worth recording our discussion in a note so the reasons for doing this weird thing are recorded.

michaelpj:
Has conflicts.

effectfully (approved):
Didn't review in detail, but seems good to me.

kwxm (approved):
What's the status of this?  I just noticed that it was on my list of things to review: I did review it at the time but I think I forgot to click the `approve` button.  It seems harmless and easy to change if we later decide that we don't like something, so it looks safe to merge (although it's probably got a bit out of sync by now).

plutus-core/untyped-plutus-core/test/Evaluation/Golden/iteAtIntegerAtInteger.type.golden
```diff
@@ -1,6 +1,6 @@
 (Left Type mismatch at () in term
   '{ (builtin ifThenElse) (con integer) }'.
 Expected type
-  '(all *_0 (type) *_0)',
+  '(all `*`_0 (type) `*`_0)',
```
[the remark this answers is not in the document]
zliu41: https://input-output.atlassian.net/browse/PLT-6059

zliu41 (commented):
> What's the status of this?

Oh, it completely slipped my mind. I think it's ready to merge.

PR #5219, opened 2023-03-20
[prototype] Support `base`'s classes and methods in Plutus Tx

zliu41:
This PR requires a patched GHC, so it won't be mergeable for a while. But it is reviewable and I'm looking to get some early feedback.

This PR adds support for Haskell's `Eq`, `Ord` and `Num`. Other classes can be supported in a similar way. The benefits of doing so include:
- One would be able to write regular Haskell rather than studying and using PlutusTx's version of `base`, and having to make sure things are imported from the right places.
- It opens up the possibilities of using third party libraries.
- Things like matching on integer literals and list range syntax will work (the latter doesn't currently work because we'd need to support `enumFromTo`, but is easily fixable)
- This approach, with the help of the GHC patch, makes it easier to generate good error messages for unsupported features, e.g., we will be able to report `Unsupported: (==) @Int` rather than `Unsupported: I#`.

User migration will be easy because one can freely mix Haskell's and PlutusTx's typeclasses.

I've updated the modules in `plutus-benchmark` to use Haskell's typeclasses whenever possible. There are some cost increases, and as far as I can tell, they are all due to the fact that Haskell's `Eq` has two methods. This can be fixed separately in multiple ways, e.g., with a better inliner, or with some tricks in the plugin itself.

The main changes are in `PlutusTx.Compiler.Expr` and `PlutusTx.Compiler.Dictionary.*`. Other changes are mostly boilerplate.

The [GHC patch](https://gitlab.haskell.org/ghc/ghc/-/commit/3cbf2dcdfa232c8d94303be2fc389081716393f9) is a fairly simple one: it adds a field in `SimplMode` and a GHC flag to turn off builtin rules, which stops GHC from turning `(==) @Integer` into `$fEqInteger_$c==`.


bezirg:
Great work, is this GHC patch going to be mainlined? If so is there a MR to track it?

cabal.project
```diff
@@ -32,6 +32,7 @@ packages: doc/read-the-docs-site
           prettyprinter-configurable
           word-array
           stubs/plutus-ghc-stub
+with-compiler: /home/zliu41/ghc/_build/stage1/bin/ghc
```
bezirg, line 35: Maybe we can integrate the GHC patch with Nix inside a local .nix file?

plutus-ledger-api/src/PlutusLedgerApi/V1/Contexts.hs
```diff
@@ -132,7 +132,11 @@ data ScriptContext = ScriptContext
     { scriptContextTxInfo  :: TxInfo -- ^ information about the transaction the currently-executing script is included in
     , scriptContextPurpose :: ScriptPurpose -- ^ the purpose of the currently-executing script
     }
-    deriving stock (Generic, Haskell.Eq, Haskell.Show)
+    deriving stock (Generic, Haskell.Show)
+
+instance Haskell.Eq ScriptContext where
+    {-# INLINABLE (==) #-}
+    (==) = (==)
```
bezirg, line 139: suggestion: use some module qualification to make it look less like a loop

plutus-tx-plugin/src/PlutusTx/Plugin.hs
```diff
@@ -161,6 +161,7 @@ mkSimplPass flags logger =
             , GHC.sm_uf_opts = GHC.defaultUnfoldingOpts
             , GHC.sm_dflags = flags
             , GHC.sm_rules = False
+            , GHC.sm_builtin_rules = False
```
bezirg, line 164: Nice! Add a comment here or a note on why we have this set to False (default for ghc being True)

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -848,6 +857,280 @@ compileExpr e = withContextM 2 (sdToTxt $ "Compiling expr:" GHC.<+> GHC.ppr e) $
...
+            f targs (cls, v, args)
+        else Nothing
+isKnownPred _ _ _ = Nothing
+
+-- | Like `knownApps` but for class dictionaries.
+knownPreds ::
+    CompilingDefault uni fun m ann =>
+    Map (Maybe String, String)
+        ([GHC.Type] -> (GHC.Class, GHC.Var, [GHC.CoreExpr]) -> Maybe (m (PIRTerm uni fun)))
+knownPreds =
```
michaelpj, line 1046: Hm. This feels very complicated. I don't suppose there's any way that we can match the ids of the method functions themselves? That would make things a _ton_ easier: we would just substitute exactly those for our functions and we'd be done, i.e.

```
MkEqDict @Integer integerEq integerNEq -- spot integerEq and replace _there_
```

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -848,6 +857,280 @@ compileExpr e = withContextM 2 (sdToTxt $ "Compiling expr:" GHC.<+> GHC.ppr e) $
...
+    CompilingDefault uni fun m ann =>
+    Map (Maybe String, String) ([GHC.CoreExpr] -> Maybe (m (PIRTerm uni fun)))
+knownApps =
+    Map.fromListWithKey (\n -> error ("knownApps: key defined more than once: " <> show n))
+        . fmap (first splitThName)
+        $ [ ( '(GHC.Classes.==)
+          , \case
+                [GHC.Type ty, _eqDict]
+                    | ty `GHC.eqType` GHC.integerTy ->
+                        -- Compile `(==) @Integer` via `Builtins.equalsInteger`.
```
michaelpj, line 882: Isn't this pretty redundant with the class-dictionary substitution you do below? You're already going to replace the dictionary for `Eq @Integer`. And if we have `(==) @Integer` then we _do_ have a concrete, known dictionary and we _should_ be able to optimize that away. So I'm unconvinced that we need this for the typeclass functions.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -848,6 +857,280 @@ compileExpr e = withContextM 2 (sdToTxt $ "Compiling expr:" GHC.<+> GHC.ppr e) $
...
+                    | ty `GHC.eqType` GHC.integerTy ->
+                        Just $
+                            compileExpr . GHC.Var =<< lookupId =<< thNameToGhcNameOrFail 'Builtins.greaterThanEqualsInteger
+                [GHC.Type ty, _ordDict, GHC.Lit (GHC.LitNumber _ i), GHC.Lit (GHC.LitNumber _ j)]
+                    | ty `GHC.eqType` GHC.integerTy -> Just $ do
+                        res <- lookupDataCon =<< thNameToGhcNameOrFail (if i >= j then 'True else 'False)
+                        compileDataConRef res
+                _ -> Nothing
+            )
+          , ( '(GHC.Num.+)
```
michaelpj, line 952: These ones I guess we do need since we don't have `Num` :/

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -848,6 +857,280 @@ compileExpr e = withContextM 2 (sdToTxt $ "Compiling expr:" GHC.<+> GHC.ppr e) $
...
+                        Just $
+                            compileExpr . GHC.Var =<< lookupId =<< thNameToGhcNameOrFail 'BI.multiplyInteger
+                [GHC.Type ty, _numDict, GHC.Lit (GHC.LitNumber numTy i), GHC.Lit (GHC.LitNumber _numTy j)]
+                    | ty `GHC.eqType` GHC.integerTy ->
+                        Just . compileExpr . GHC.Lit $ GHC.LitNumber numTy (i * j)
+                _ -> Nothing
+            )
+          , ( 'GHC.Num.fromInteger
+            , \case
+                -- Compile `fromInteger @Integer` via `id`.
```
michaelpj, line 987: Does this actually come up much? Also does the base version not work given that it's presumably implemented as `id`?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -848,6 +857,280 @@ compileExpr e = withContextM 2 (sdToTxt $ "Compiling expr:" GHC.<+> GHC.ppr e) $
...
+                [GHC.Type ty, _numDict]
+                    | ty `GHC.eqType` GHC.integerTy -> Just $ do
+                        idId <- lookupId =<< thNameToGhcNameOrFail 'id
+                        compileExpr $ GHC.mkCoreApps (GHC.Var idId) [GHC.Type GHC.integerTy]
+                [GHC.Type ty, _numDict, arg]
+                    | ty `GHC.eqType` GHC.integerTy -> Just $ compileExpr arg
+                _ -> Nothing
+            )
+          , ( 'GHC.Num.negate
```
michaelpj, line 996: Ditto, I would expect the base version to work here.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -848,6 +857,280 @@ compileExpr e = withContextM 2 (sdToTxt $ "Compiling expr:" GHC.<+> GHC.ppr e) $
...
+                        compileExpr $
+                            GHC.mkCoreApps
+                                (GHC.Var subtractIntegerId)
+                                [GHC.Lit $ GHC.LitNumber GHC.LitNumInteger 0]
+                [GHC.Type ty, _numDict, GHC.Lit (GHC.LitNumber numTy i)]
+                    | ty `GHC.eqType` GHC.integerTy ->
+                        Just . compileExpr . GHC.Lit $ GHC.LitNumber numTy (-i)
+                _ -> Nothing
+            )
+          , ( 'GHC.Real.toInteger
```
michaelpj, line 1011: ditto

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -848,6 +857,280 @@ compileExpr e = withContextM 2 (sdToTxt $ "Compiling expr:" GHC.<+> GHC.ppr e) $
...
+        Nothing -> throwPlain $ UnsupportedError "Empty class"
+        Just args ->
+            -- Here we cannot apply `compileExpr` to
+            -- `GHC.mkCoreApps (GHC.Var v) (otherArgs <> selArgs)` directly, because
+            -- it will loop. We therefore separately compile the last argument.
+            PIR.Apply annMayInline
+                <$> compileExpr (GHC.mkCoreApps (GHC.Var v) (NE.init args))
+                <*> compileExpr (NE.last args)
+
+splitGhcName :: GHC.Name -> (Maybe String, String)
```
michaelpj, line 1125: I'm just uncovinced you need this at all if you use a function and the appropriate `Eq` instances.

plutus-tx-plugin/src/PlutusTx/Compiler/Types.hs
```diff
@@ -174,13 +177,69 @@ stableModuleCmp m1 m2 =
     -- See Note [Stable name comparisons]
     (GHC.moduleUnit m1 `GHC.stableUnitCmp` GHC.moduleUnit m2)
 
+class Monad m => MonadCoreM m where
```
bezirg, line 180: You forgot the`lookupThing` method.
Wouldn't it be better to add some orphan instances of the `MonadThings` class? Then only the `thNameToGhcName` method would be missing, but then you could create a subclass:

```
class MonadThings m => MonadCoreM m where
    thNameToGhcName = ...
```

https://hackage.haskell.org/package/ghc-9.2.7/docs/GHC-Types-TyThing.html#t:MonadThings

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -848,6 +857,280 @@ compileExpr e = withContextM 2 (sdToTxt $ "Compiling expr:" GHC.<+> GHC.ppr e) $
...
+    -- | The class whose dictionary we are compiling
+    GHC.Class ->
+    -- | Dictionary data constructor
+    GHC.Var ->
+    -- | Arguments to the dictionary data constructor
+    [GHC.CoreExpr] ->
+    -- | A list of dictionary method names where alternative definitions exist
+    [TH.Name] ->
+    m (PIRTerm uni fun)
+compileDict cls v args0 txSelNames = do
```
michaelpj, line 1096: This is all super fiddly and I don't feel very good about it. I'd have to stare at it a while to have ideas though.

plutus-tx-plugin/test/Budget/allCheap.plc.golden
```diff
@@ -23,7 +30,103 @@ letrec
                  !acc : Bool = go xs
                in
                Bool_match
-                 (ifThenElse {Bool} (lessThanEqualsInteger 1 x) False True)
+                 ((let
```
michaelpj, line 33: wow this is terrible D:

plutus-tx-plugin/test/Budget/allEmptyList.plc.golden
```diff
@@ -23,7 +30,103 @@ letrec
...
+                                    True
+                                    False)
+                                 {all dead. integer}
+                                 (/\dead -> x)
+                                 (/\dead -> y)
+                                 {all dead. dead})
+                    in
+                    Ord_match
+                      {integer}
+                      v
```
michaelpj, line 117: A context-based inlining heuristic would help here: inline if the thing you are inlining is a constructor application and the variable is the scrutinee of a pattern match!

michaelpj: Also need case-of-known-constructor.

plutus-tx-plugin/test/Budget/show.plc.golden
```diff
@@ -270,6 +270,82 @@ letrec
                    {all dead. dead})
             {all dead. dead}
 in
+let
```
michaelpj, line 273: I'm unsure why this one changed so much?

plutus-tx-plugin/test/Plugin/Errors/literalCaseBs.plc.golden
```diff
@@ -1 +1,3 @@
-Error: Unsupported feature: Use of == from the Haskell Eq typeclass
\ No newline at end of file
+Error: Reference to a name which is not a local, a builtin, or an external INLINABLE function: Variable PlutusTx.Builtins.Internal.$fEqBuiltinByteString
```
michaelpj, line 1: I guess we should do the trick for `==` too, but we also probably want to keep the old fallback error which was helpful.

plutus-tx-plugin/test/Plugin/Errors/literalCaseOther.plc.golden
```diff
@@ -1 +1,2 @@
-Error: Unsupported feature: Use of fromString on type other than builtin strings or bytestrings: Plugin.Errors.Spec.AType
\ No newline at end of file
+Error: Reference to a name which is not a local, a builtin, or an external INLINABLE function: Variable $c/=
```
michaelpj, line 1: ditto

michaelpj (commented):
The implementation seems very fiddly at the moment, I'd be keen to try and reduce that. I also think we probably shouldn't actually merge this unless we can optimize away the worse codegen for basic equality and comparison operations. But it's relatively clear how to do that at least in the monomorphic case, which is most important.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -848,6 +857,280 @@ compileExpr e = withContextM 2 (sdToTxt $ "Compiling expr:" GHC.<+> GHC.ppr e) $
...
+    GHC.Var ->
+    [GHC.CoreExpr] ->
+    Maybe (m (PIRTerm uni fun))
+isKnownApp fun args = Map.lookup (splitGhcName (GHC.varName fun)) knownApps >>= ($ args)
+
+-- | A list of applications that should be compiled in specific ways.
+knownApps ::
+    CompilingDefault uni fun m ann =>
+    Map (Maybe String, String) ([GHC.CoreExpr] -> Maybe (m (PIRTerm uni fun)))
+knownApps =
```
bezirg, line 875: I think this function could be code-deduplicated slightly

bezirg:
Here I paste an old comment of mine when debugging an older patched GHC (not this one):

> Problems encountered:
> 
> 1. breaks on banged ! data constructors iff they are defined and `compile`d  in the same module. BangPatterns on expression code works fine however.
> 
> 2. Even if we try to avoid relying on ghc `base` code by using our own prelude, ghc will emit core that implicitly calls some `base` functions. `Implicitly` means that even if we never explicitly mentioned `base` code in Plutus, ghc will emit `base` code. This happens in at least two cases: a) literal patterns b) (list) stream fusion code.
> 
> this is also a problem that should be fixed by the patched-ghc team, by having extensible fields also for `base` modules.

PR #5229, opened 2023-03-23
PLT-5287: Define a runCekM-like function `giveCekReqs` for the debugger

bezirg:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Debug/Internal.hs
```diff
@@ -345,6 +347,9 @@ throwingDischarged
     -> CekM uni fun s x
 throwingDischarged l t = throwingWithCause l t . Just . dischargeCekValue
 
+-- | Duplicate of 'Cek.Internal.runCekM'.
+--
+-- Exporting and reusing that function resulted in +20% perf loss for Cek, see #5223.
 runCekM
```
bezirg, line 353: I could also edit the existing `runCekM` to make use of `giveCekReqs` but I thought to keep it the same as the original cek internal module.

zliu41: Does `{-# INLINE runCekM #-}` help? It could be due to inlining vs. not, but could also be for other reasons.

zliu41 (approved):
Looks good. The fewer implicit parameters the better!

bezirg:
After discussing with Michael, I will try to improve this PR a bit to make it less leaky

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Debug/Driver.hs
```diff
@@ -87,10 +88,10 @@ type Driving m uni fun ann bps =
     )
 
 -- | Entrypoint of the driver
-runDriver :: forall uni fun ann bps m.
+runDriverT :: forall uni fun ann bps m.
             (Breakpointable ann bps, MonadFree (DebugF uni fun ann bps) m)
           => NTerm uni fun ann -> m ()
-runDriver = void . runReaderT driver . initState
+runDriverT = void . runReaderT driver . initState
```
michaelpj, line 94: This is a kind of weird name. It sounds like it should be a function that gets you out of the `DriverT` monad transformer... but that's definitely not what's happening here?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -58,6 +58,7 @@ module UntypedPlutusCore.Evaluation.Machine.Cek.Internal
     , Slippage
     , defaultSlippage
     , NTerm
+    , runCekM
```
michaelpj, line 61: is this used somewhere? I guess it must be.

michaelpj (approved):
MUCH better IMO.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
bezirg: It is used by the steppable cek to provide the same `runCekDeBruijn` run-to-completion interface as the original cek

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Debug/Driver.hs
bezirg: It is indeed. It does not actually run the whole `Driving` monad stack (i.e. reader,free); it only runs the `Reader CekState` which is the state of the driver.
So it is very similar to `runReaderT`. If you have a better name, let me know!

PR #5230, opened 2023-03-23
Set dependent-sum lower bound in plutus-core.cabal

zliu41:


effectfully (approved):

plutus-core/plutus-core.cabal
```diff
@@ -62,7 +62,7 @@ common lang
   ghc-options:
     -Wall -Wnoncanonical-monad-instances -Wincomplete-uni-patterns
     -Wincomplete-record-updates -Wredundant-constraints -Widentities
-    -Wunused-packages -Wmissing-deriving-strategies
+    -Wmissing-deriving-strategies
```
michaelpj, line 65: This is too big a hammer. I think we can do better than this.

michaelpj:
Here's my take: https://github.com/input-output-hk/plutus/pull/5233

PR #5233, opened 2023-03-24
Fix dependent-sum issue through bounding some

michaelpj:
- Bound some to get a newer dependent-sum
- Bump hackage and CHaP

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -14,12 +14,12 @@ repository cardano-haskell-packages
 -- update either of these.
 
 -- repeating hackage index-state to work around haskell.nix parsing limitation
-index-state: 2022-11-15T00:00:00Z
+index-state: 2023-03-24T00:00:00Z
 index-state:
   -- Bump this if you need newer packages from Hackage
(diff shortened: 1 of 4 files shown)


michaelpj:
This is based on Andrea's idea from https://github.com/input-output-hk/cardano-haskell-packages/pull/167#issuecomment-1482060744

I also bumped Hackage so we could actually get a newer `some` which would manifest the issue if it wasn't fixed.

zliu41:
This could be a problem if a downstream project needs a newer `some`.
Then it would need to `allow-newer`, but then that would make cabal pick an older `dependent-sum` and fail to build which is confusing.
How about sticking an unused import in some module so that we do depend on `dependent-sum` directly?

michaelpj:
>  How about sticking an unused import in some module so that we do depend on dependent-sum directly?

Oh yes, that's a good idea. I'll do that.

zliu41 (approved):

PR #5239, opened 2023-03-28
[Builtins] Optimize costing by being lazier

effectfully:
This makes costing lazier to improve both performance and expressiveness.

The `Costing.hs` module with tests isn't ready for review yet, everything else is ready for review.

### DIFF
--- a/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
+++ b/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
@@ -481,7 +481,10 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
             @(Data -> [Integer])
             (\_ -> unsafePerformIO $ do
                 performMinorGC
-                threadDelay 50000  -- Giving GC enough time to finish up.
+                -- Somehow this allows the remaining finalizers to complete, while 'yeild' doesn't.
+                -- No idea what's going on. Not that we care much though, since even if we lose
+                -- several dozens of remaining elements it would a bug for that to affect anything.
+                threadDelay 1
                 finalize
                 results <- readIORef resultsVar
                 pure $ reverse results)
--- a/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
+++ b/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
@@ -23,6 +23,7 @@ import PlutusCore.Builtin
 import PlutusCore.Data
 import PlutusCore.Evaluation.Machine.BuiltinCostModel
 import PlutusCore.Evaluation.Machine.ExBudget
+import PlutusCore.Evaluation.Machine.ExBudgetStream
 import PlutusCore.Evaluation.Machine.Exception
 import PlutusCore.Pretty
 
--- a/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
+++ b/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
@@ -29,7 +29,6 @@ import PlutusCore.Pretty
 
 import PlutusCore.StdLib.Data.ScottList qualified as Plc
 
-import Control.Concurrent (threadDelay)
 import Control.Concurrent.MVar
 import Control.Exception
 import Control.Monad
@@ -474,7 +473,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
                     -- and don't update @pendingGcVar@. I.e. this entire testing machinery can
                     -- affect how budgets are retained, but the impact of the 'MVar' business is
                     -- negligible, since @pendingGcVar@ is filled immediately after it's emptied.
-                    Nothing        -> void $ atomicModifyIORef'_ numsOfGcedVar (0 :)
+                    Nothing -> pure ()
                     -- If @pendingGcVar@ is not empty, then we cons its content to the resulting
                     -- list and put @0@ as the new content of the variable.
                     Just pendingGc -> do
@@ -491,14 +490,18 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
                 -- We need to trigger GC occasionally (otherwise it can easily take more than 100k
                 -- elements before GC is triggered and the number can go much higher depending on
                 -- the RTS options), so we picked 100 elements as a cutoff number. Doing GC less
-                -- often makes tests slower (for lower cutoff numbers -- by a lot), doing GC more
-                -- often requires us to generate longer streams in tests in order to observe
-                -- meaningful results making tests slower.
+                -- often makes tests slower, doing GC more often requires us to generate longer
+                -- streams in tests in order to observe meaningful results making tests slower.
                 when (pendingGc' >= 100) performMinorGC
 
             -- Call @regBudget@ over each element of the stream of budgets.
-            regBudgets (ExBudgetLast budget)         = do
+            regBudgets (ExBudgetLast budget) = do
                 regBudget budget
+                -- Run @finalize@ one final time before returning the last budget.
+                finalize
+                -- Make all outstanding finalizers inert, so that we don't mix up budgets GCed
+                -- during spending with budgets GCed right after spending finishes.
+                _ <- takeMVar pendingGcVar
                 pure $ ExBudgetLast budget
             regBudgets (ExBudgetCons budget budgets) = do
                 regBudget budget
@@ -508,24 +511,12 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
                 budgets' <- unsafeInterleaveIO $ regBudgets budgets
                 pure $ ExBudgetCons budget budgets'
 
-            -- Just a random model that keeps the original costs coming from the 'ExMemoryUsage'
-            -- instance.
+            -- Just a random model that keeps the costs coming from the 'ExMemoryUsage' instance.
             linear1 = ModelOneArgumentLinearCost $ ModelLinearSize 1 1
             model   = CostingFun linear1 linear1
         pure $ makeBuiltinMeaning
             @(Data -> [Integer])
-            (\_ -> unsafePerformIO $ do
-                -- Pick up remaining budgets.
-                performMinorGC
-                -- Somehow this allows the remaining finalizers to complete, while 'yeild' doesn't.
-                -- No idea what's going on. Not that we care much though, since even if we lose
-                -- several dozens of remaining elements it would be a bug for that to affect
-                -- anything.
-                threadDelay 1
-                -- Run @finalize@ one final time.
-                finalize
-                numsOfGced <- readIORef numsOfGcedVar
-                pure $ reverse numsOfGced)
+            (\_ -> unsafePerformIO $ reverse <$> readIORef numsOfGcedVar)
             (\_ -> unsafePerformIO . regBudgets . runCostingFunOneArgument model)
 
 instance Default (BuiltinVersion ExtensionFun) where
--- a/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
+++ b/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
@@ -20,25 +20,33 @@ module PlutusCore.Examples.Builtins where
 
 import PlutusCore
 import PlutusCore.Builtin
+import PlutusCore.Data
+import PlutusCore.Evaluation.Machine.BuiltinCostModel
 import PlutusCore.Evaluation.Machine.ExBudget
 import PlutusCore.Evaluation.Machine.Exception
 import PlutusCore.Pretty
 
 import PlutusCore.StdLib.Data.ScottList qualified as Plc
 
+import Control.Concurrent (threadDelay)
+import Control.Concurrent.MVar
 import Control.Exception
+import Control.Monad
 import Data.Default.Class
 import Data.Either
-import Data.Hashable (Hashable)
 import Data.Kind qualified as GHC (Type)
 import Data.Proxy
 import Data.Some.GADT qualified as GADT
 import Data.Tuple
 import Data.Void
 import GHC.Generics
+import GHC.IORef
 import GHC.Ix
 import GHC.TypeLits
 import Prettyprinter
+import System.IO.Unsafe (unsafeInterleaveIO, unsafePerformIO)
+import System.Mem (performMinorGC)
+import System.Mem.Weak (addFinalizer)
 
 instance (Bounded a, Bounded b) => Bounded (Either a b) where
     minBound = Left  minBound
@@ -128,6 +136,7 @@ data ExtensionFun
     | SwapEls  -- For checking that nesting polymorphic built-in types and instantiating them with
                -- a mix of monomorphic types and type variables works correctly.
     | ExtensionVersion -- Reflect the version of the Extension
+    | TrackCosts
     deriving stock (Show, Eq, Ord, Enum, Bounded, Ix, Generic)
     deriving anyclass (Hashable)
 
@@ -214,12 +223,12 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
     toBuiltinMeaning _ver Factorial =
         makeBuiltinMeaning
             (\(n :: Integer) -> product [1..n])
-            mempty  -- Whatever.
+            (\_ _ -> ExBudgetLast mempty)  -- Whatever.
 
     toBuiltinMeaning _ver ForallFortyTwo =
         makeBuiltinMeaning
             forallFortyTwo
-            (\_ -> ExBudget 1 0)
+            (\_ -> ExBudgetLast mempty)
       where
         forallFortyTwo :: MetaForall ('TyNameRep @GHC.Type "a" 0) Integer
         forallFortyTwo = MetaForall 42
@@ -227,27 +236,27 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
     toBuiltinMeaning _ver SumInteger =
         makeBuiltinMeaning
             (sum :: [Integer] -> Integer)
-            mempty  -- Whatever.
+            (\_ _ -> ExBudgetLast mempty)  -- Whatever.
 
     toBuiltinMeaning _ver Const =
         makeBuiltinMeaning
             const
-            (\_ _ _ -> ExBudget 1 0)
+            (\_ _ _ -> ExBudgetLast mempty)
 
     toBuiltinMeaning _ver Id =
         makeBuiltinMeaning
             Prelude.id
-            (\_ _ -> ExBudget 1 0)
+            (\_ _ -> ExBudgetLast mempty)
 
     toBuiltinMeaning _ver IdAssumeBool =
         makeBuiltinMeaning
             (Prelude.id :: Opaque val Bool -> Opaque val Bool)
-            (\_ _ -> ExBudget 1 0)
+            (\_ _ -> ExBudgetLast mempty)
 
     toBuiltinMeaning _ver IdAssumeCheckBool =
         makeBuiltinMeaning
             idAssumeCheckBoolPlc
-            mempty  -- Whatever.
+            (\_ _ -> ExBudgetLast mempty)  -- Whatever.
       where
         idAssumeCheckBoolPlc :: Opaque val Bool -> EvaluationResult Bool
         idAssumeCheckBoolPlc val =
@@ -258,7 +267,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
     toBuiltinMeaning _ver IdSomeConstantBool =
         makeBuiltinMeaning
             idSomeConstantBoolPlc
-            mempty  -- Whatever.
+            (\_ _ -> ExBudgetLast mempty)  -- Whatever.
       where
         idSomeConstantBoolPlc :: SomeConstant uni Bool -> EvaluationResult Bool
         idSomeConstantBoolPlc = \case
@@ -268,7 +277,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
     toBuiltinMeaning _ver IdIntegerAsBool =
         makeBuiltinMeaning
             idIntegerAsBool
-            mempty  -- Whatever.
+            (\_ _ -> ExBudgetLast mempty)  -- Whatever.
       where
         idIntegerAsBool :: SomeConstant uni Integer -> EvaluationResult (SomeConstant uni Integer)
         idIntegerAsBool = \case
@@ -278,19 +287,19 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
     toBuiltinMeaning _ver IdFInteger =
         makeBuiltinMeaning
             (Prelude.id :: fi ~ Opaque val (f `TyAppRep` Integer) => fi -> fi)
-            (\_ _ -> ExBudget 1 0)
+            (\_ _ -> ExBudgetLast mempty)
 
     toBuiltinMeaning _ver IdList =
         makeBuiltinMeaning
             (Prelude.id :: la ~ Opaque val (PlcListRep a) => la -> la)
-            (\_ _ -> ExBudget 1 0)
+            (\_ _ -> ExBudgetLast mempty)
 
     toBuiltinMeaning _ver IdRank2 =
         makeBuiltinMeaning
             (Prelude.id
                 :: afa ~ Opaque val (TyForallRep @GHC.Type a (TyVarRep f `TyAppRep` TyVarRep a))
                 => afa -> afa)
-            (\_ _ -> ExBudget 1 0)
+            (\_ _ -> ExBudgetLast mempty)
 
     toBuiltinMeaning _ver ScottToMetaUnit =
         makeBuiltinMeaning
@@ -302,36 +311,36 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
                 -- (unlike in the case of @IdRank2@ where 'TyAppRep' preserves the Rep context).
                 :: oa ~ Opaque val (TyVarRep a)
                 => Opaque val (TyForallRep a (oa -> oa)) -> ())
-            (\_ _ -> ExBudget 1 0)
+            (\_ _ -> ExBudgetLast mempty)
 
     toBuiltinMeaning _ver FailingSucc =
         makeBuiltinMeaning
             @(Integer -> Integer)
             (\_ -> throw BuiltinErrorCall)
-            (\_ _ -> ExBudget 1 0)
+            (\_ _ -> ExBudgetLast mempty)
 
     toBuiltinMeaning _ver ExpensiveSucc =
         makeBuiltinMeaning
             @(Integer -> Integer)
             (\_ -> throw BuiltinErrorCall)
-            (\_ _ -> unExRestrictingBudget enormousBudget)
+            (\_ _ -> ExBudgetLast $ unExRestrictingBudget enormousBudget)
 
     toBuiltinMeaning _ver FailingPlus =
         makeBuiltinMeaning
             @(Integer -> Integer -> Integer)
             (\_ _ -> throw BuiltinErrorCall)
-            (\_ _ _ -> ExBudget 1 0)
+            (\_ _ _ -> ExBudgetLast mempty)
 
     toBuiltinMeaning _ver ExpensivePlus =
         makeBuiltinMeaning
             @(Integer -> Integer -> Integer)
             (\_ _ -> throw BuiltinErrorCall)
-            (\_ _ _ -> unExRestrictingBudget enormousBudget)
+            (\_ _ _ -> ExBudgetLast $ unExRestrictingBudget enormousBudget)
 
     toBuiltinMeaning _ver IsConstant =
         makeBuiltinMeaning
             isConstantPlc
-            mempty  -- Whatever.
+            (\_ _ -> ExBudgetLast mempty)  -- Whatever.
       where
         -- The type signature is just for clarity, it's not required.
         isConstantPlc :: Opaque val a -> Bool
@@ -340,7 +349,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
     toBuiltinMeaning _ver UnsafeCoerce =
         makeBuiltinMeaning
             unsafeCoercePlc
-            (\_ _ -> ExBudget 1 0)
+            (\_ _ -> ExBudgetLast mempty)
       where
         -- The type signature is just for clarity, it's not required.
         unsafeCoercePlc :: Opaque val a -> Opaque val b
@@ -349,7 +358,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
     toBuiltinMeaning _ver UnsafeCoerceEl =
         makeBuiltinMeaning
             unsafeCoerceElPlc
-            (\_ _ -> ExBudget 1 0)
+            (\_ _ -> ExBudgetLast mempty)
       where
         unsafeCoerceElPlc
             :: SomeConstant DefaultUni [a]
@@ -361,19 +370,23 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
     toBuiltinMeaning _ver Undefined =
         makeBuiltinMeaning
             undefined
-            (\_ -> ExBudget 1 0)
+            (\_ -> ExBudgetLast mempty)
 
     toBuiltinMeaning _ver Absurd =
         makeBuiltinMeaning
             absurd
-            (\_ _ -> ExBudget 1 0)
+            (\_ _ -> ExBudgetLast mempty)
 
     toBuiltinMeaning _ver ErrorPrime =
         makeBuiltinMeaning
             EvaluationFailure
-            (\_ -> ExBudget 1 0)
+            (\_ -> ExBudgetLast mempty)
 
-    toBuiltinMeaning _ver Comma = makeBuiltinMeaning commaPlc mempty where
+    toBuiltinMeaning _ver Comma =
+        makeBuiltinMeaning
+            commaPlc
+            (\_ _ _ -> ExBudgetLast mempty)
+      where
         commaPlc
             :: SomeConstant uni a
             -> SomeConstant uni b
@@ -381,7 +394,11 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
         commaPlc (SomeConstant (Some (ValueOf uniA x))) (SomeConstant (Some (ValueOf uniB y))) =
             fromValueOf (DefaultUniPair uniA uniB) (x, y)
 
-    toBuiltinMeaning _ver BiconstPair = makeBuiltinMeaning biconstPairPlc mempty where
+    toBuiltinMeaning _ver BiconstPair =
+        makeBuiltinMeaning
+            biconstPairPlc
+            (\_ _ _ _ -> ExBudgetLast mempty)
+      where
         biconstPairPlc
             :: SomeConstant uni a
             -> SomeConstant uni b
@@ -396,7 +413,11 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
                 Just Refl <- pure $ uniB `geq` uniB'
                 pure $ fromValueOf uniPairAB (x, y)
 
-    toBuiltinMeaning _ver Swap = makeBuiltinMeaning swapPlc mempty where
+    toBuiltinMeaning _ver Swap =
+        makeBuiltinMeaning
+            swapPlc
+            (\_ _ -> ExBudgetLast mempty)
+      where
         swapPlc
             :: SomeConstant uni (a, b)
             -> EvaluationResult (SomeConstant uni (b, a))
@@ -404,7 +425,11 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
             DefaultUniPair uniA uniB <- pure uniPairAB
             pure $ fromValueOf (DefaultUniPair uniB uniA) (snd p, fst p)
 
-    toBuiltinMeaning _ver SwapEls = makeBuiltinMeaning swapElsPlc mempty where
+    toBuiltinMeaning _ver SwapEls =
+        makeBuiltinMeaning
+            swapElsPlc
+            (\_ _ -> ExBudgetLast mempty)
+      where
         -- The type reads as @[(a, Bool)] -> [(Bool, a)]@.
         swapElsPlc
             :: SomeConstant uni [SomeConstant uni (a, Bool)]
@@ -418,12 +443,48 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
     -- See Note [Versioned builtins]
     toBuiltinMeaning ver ExtensionVersion =
         makeBuiltinMeaning
-        @(() -> EvaluationResult Integer)
-        (\(_ :: ()) -> EvaluationSuccess $ case ver of
-                ExtensionFunV0 -> 0
-                ExtensionFunV1 -> 1
-        )
-        mempty  -- Whatever
+            @(() -> EvaluationResult Integer)
+            (\(_ :: ()) -> EvaluationSuccess $ case ver of
+                    ExtensionFunV0 -> 0
+                    ExtensionFunV1 -> 1)
+            (\_ _ -> ExBudgetLast mempty)  -- Whatever
+
+    toBuiltinMeaning _ TrackCosts = unsafePerformIO $ do
+        counterVar <- newMVar 0
+        resultsVar <- newIORef []
+        let finalize =
+                tryTakeMVar counterVar >>= \case
+                    Nothing      -> void $ atomicModifyIORef'_ resultsVar (0 :)
+                    Just counter -> do
+                        _ <- atomicModifyIORef'_ resultsVar (counter :)
+                        putMVar counterVar 0
+
+            regBudget budget = do
+                counter <- takeMVar counterVar
+                let counter' = succ counter
+                putMVar counterVar counter'
+                addFinalizer budget finalize
+                when (counter' >= 100) performMinorGC
+
+            regBudgets (ExBudgetLast budget)         = do
+                regBudget budget
+                pure $ ExBudgetLast budget
+            regBudgets (ExBudgetCons budget budgets) = do
+                regBudget budget
+                budgets' <- unsafeInterleaveIO $ regBudgets budgets
+                pure $ ExBudgetCons budget budgets'
+
+            linear1 = ModelOneArgumentLinearCost $ ModelLinearSize 1 1
+            model   = CostingFun linear1 linear1
+        pure $ makeBuiltinMeaning
+            @(Data -> [Integer])
+            (\_ -> unsafePerformIO $ do
+                performMinorGC
+                threadDelay 50000  -- Giving GC enough time to finish up.
+                finalize
+                results <- readIORef resultsVar
+                pure $ reverse results)
+            (\_ -> unsafePerformIO . regBudgets . runCostingFunOneArgument model)
 
 instance Default (BuiltinVersion ExtensionFun) where
     def = ExtensionFunV1
--- a/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
+++ b/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
@@ -137,7 +137,8 @@ data ExtensionFun
     | SwapEls  -- For checking that nesting polymorphic built-in types and instantiating them with
                -- a mix of monomorphic types and type variables works correctly.
     | ExtensionVersion -- Reflect the version of the Extension
-    | TrackCosts
+    | TrackCosts  -- For checking that each cost is released and can be picked up by GC once it's
+                  -- accounted for in the evaluator.
     deriving stock (Show, Eq, Ord, Enum, Bounded, Ix, Generic)
     deriving anyclass (Hashable)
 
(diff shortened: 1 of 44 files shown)


effectfully:
Eek.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
```diff
@@ -262,17 +302,138 @@ instance ExMemoryUsage a => ExMemoryUsage [a] where
...
+-- of a function passed as an accumulator.
+flattenCostRoseGo :: Int# -> [CostRose] -> (Int# -> CostStream) -> CostStream
+flattenCostRoseGo i0 []                                    k = k i0
+-- I cannot imagine this tilda making a difference in any even remotely realistic scenario, but
+-- better safe than sorry.
+flattenCostRoseGo i0 (~(CostRose cost1 forest1) : forest2) k =
+    CostCons (coerce $ I# i0) $
+        case coerce cost1 of
+            I# i1 ->
+                flattenCostRoseGo i1 forest1 $ \i2 ->
```
zliu41, line 429: As far as I can tell this is amortized O(1) for producing each element, but could take linear time for some elements (e.g., the last one). Is this optimal?

plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
```diff
@@ -418,12 +443,48 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
...
+                budgets' <- unsafeInterleaveIO $ regBudgets budgets
+                pure $ ExBudgetCons budget budgets'
+
+            linear1 = ModelOneArgumentLinearCost $ ModelLinearSize 1 1
+            model   = CostingFun linear1 linear1
+        pure $ makeBuiltinMeaning
+            @(Data -> [Integer])
+            (\_ -> unsafePerformIO $ do
+                performMinorGC
+                threadDelay 50000  -- Giving GC enough time to finish up.
```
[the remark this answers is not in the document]
effectfully: Right, silly me. However that `threadDelay 50000` does affect the results: without it some finalizers are run too late and we don't get some of the trailing elements of the stream registered. It's like 50 elements on my machine, but still.

I just tried `threadDelay 1`, it helped just as much. So I also tried `yield` and it didn't help with the loss of trailing elements. I don't have the slightest idea of what's going on, but it's not too important anyway. I'll make `threadDelay 1` just to please the gods of GHC, so that they don't run finalizers prematurely (and maybe don't insert unwanted join points into my code so much).

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -146,21 +146,27 @@ runCostingFunOneArgument
...
     :: ModelOneArgument
-    -> ExMemory
-    -> CostingInteger
-runOneArgumentModel (ModelOneArgumentConstantCost c) = lazy $ \_ -> c
-runOneArgumentModel (ModelOneArgumentLinearCost (ModelLinearSize intercept slope)) = \(ExMemory s) ->
-    s * slope + intercept
+    -> CostStream
+    -> CostStream
+runOneArgumentModel (ModelOneArgumentConstantCost c) =
+    lazy $ \_ -> CostLast c
```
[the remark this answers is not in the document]
effectfully: [Yes](https://github.com/input-output-hk/plutus/blob/65794ecffa70443c3474b11f8628991cce6445a8/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs#L104), but it needs an additional comment. Will add it, thanks.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
```diff
@@ -262,17 +302,138 @@ instance ExMemoryUsage a => ExMemoryUsage [a] where
...
+                    flattenCostRoseGo i2 forest2 k
+
+-- | Flatten the tree structure of 'CostRose' into a linear 'CostStream' lazily. This function
+-- is productive regardless of the recursion pattern of the given argument.
+flattenCostRose :: CostRose -> CostStream
+flattenCostRose (CostRose cost [])     = CostLast cost
```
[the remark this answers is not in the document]
effectfully: Yes, I'll write a Note about that. Basically, we special-case handling of single-element streams all over the place for performance reasons.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
effectfully: Hm, I'm not seeing any linear time in there, could you explain? Linear memory in the _depth_ of tree -- sure, but I don't think we can do better than that and it's conservative too (it's just that previously the linearity-of-memory-consumption was in the `ExMemoryUsage Data` instance rather than here).

So `flattenCostRoseGo` always immediately emits an element when the forest is non-empty. Then it grows the continuation (constant-time too) and recurses into the new node. If the node is non-empty too, then we get another element in constant time, grow the continuation in constant time and recurse again. If the next node contains an empty forest, then we don't emit an element and jump to the continuation, which in constant time to reduces to a `flattenCostRoseGo` call (i.e. without forcing the entirety of the so-far-accumulated continuation, just by reducing the outermost lambda), which again emits a cost immediately.

Thus, it seems to me that every cost is emitted in constant time, but maybe I'm misunderstanding something?

zliu41:
```haskell
flattenCostRose (CostRose 1 [r1]) where
  r1 = CostRose 2 [r2]
  r2 = CostRose 3 [r3]
  r3 = CostRose 4 []
```

In this case it will reduce to

```haskell
CostCons 1 $ CostCons 2 $ CostCons 3 $ <continuation that produces CostLast 4 in 4 steps>
```

The continuation is essentially
```haskell
\x1 -> flattenCostRoseGo x1 [] $
  \x2 -> flattenCostRoseGo x2 [] $
    \x3 -> flattenCostRoseGo x3 [] $
      \x4 -> CostLast x4
```

So it seems producing `4` takes linear time.

effectfully: OK, I was indeed misunderstanding, thanks a lot for the clarification and for catching. I'll think if it's a problem (releasing a linear amount of memory is a linear operation anyway, so maybe we can wait for the final cost as well, but not sure) and what to do about it.

effectfully: > So it seems producing 4 takes linear time.

(linear in the depth of the tree, just like memory consumption)

zliu41: If it's amortized O(1) then it's not a problem. I think the following is O(1) worst case, but I wouldn't be surprised if it's slower:
```haskell
flattenCostRose :: CostRose -> CostStream
flattenCostRose (CostRose cost []) = CostLast cost
flattenCostRose (CostRose cost (r:rs)) = CostCons cost (go r rs)

go :: CostRose -> [CostRose] -> CostStream
go (CostRose cost []) [] = CostLast cost
go (CostRose cost []) (r:rs) = CostCons cost (go r rs)
go (CostRose cost (r:rs)) rs' = CostCons cost (go r (rs ++ rs'))
```

effectfully: Your code doesn't look much different to me, but perhaps I'm misunderstanding again. E.g.

```
go [CostRose cost1 [CostRose cost2 [CostRose cost3 []]]] rs' ~>

CostRose cost1 $ go [CostRose cost2 [CostRose cost3 []]] ([] ++ rs') ~>

CostRose cost1 $ CostRose cost2 $ go [CostRose cost3 []] ([] ++ ([] ++ rs'))
```

It seems like for long chain of one-element forest we're going to end up with `[] ++ ([] ++ ([] ++ <...>))`, which is effectively the same, except we also pay for all the overhead of list concatenation (it's lazy, so costs are amortized, but they're still there).

> If it's amortized O(1) then it's not a problem. 

Yeah, and I feel like special-casing single-element forests could help us level the cost of retrieving the next element, but at the expense of introducing an additional cost for every single node, which wouldn't pay off, I think. Still, I'm going to play with it.

effectfully: How about this:

```haskell
flattenCostRoseGo :: Int# -> CostRose -> [CostRose] -> (Int# -> CostStream) -> CostStream
flattenCostRoseGo i0 rose1 forest2 k =
    CostCons (coerce $ I# i0) $
        let !(CostRose cost1 forest1) = rose1
            !(I# i1) = coerce cost1
        in case forest1 of
            [] -> case forest2 of
                []                -> k i1
                rose2' : forest2' -> flattenCostRoseGo i1 rose2' forest2' k
            rose1' : forest1' -> case forest2 of
                []                -> flattenCostRoseGo i1 rose1' forest1' k
                rose2' : forest2' ->
                    flattenCostRoseGo i1 rose1' forest1' $ \i2 ->
                        flattenCostRoseGo i2 rose2' forest2' k

flattenCostRose :: CostRose -> CostStream
flattenCostRose (CostRose cost [])              = CostLast cost
flattenCostRose (CostRose cost (rose : forest)) =
    case coerce cost of
        I# i -> flattenCostRoseGo i rose forest $ \iz -> coerce CostLast $ I# iz
{-# INLINE flattenCostRose #-}
```

We essentially need to prove by induction that both `flattenCostRoseGo` and its continuation argument always produce an element in constant time. For `flattenCostRoseGo` it's entirely obvious and for the continuation we have several cases:

1. the initial case with `CostLast` -- obviously constant-time
2. the case when either of `forest1` or `forest2` is empty (but not both at the same time) -- we pass the continuation unchanged and hence don't change any of its behavior
3. the case when both `forest1` and `forest2` are non-empty -- we do grow the continuation, but by induction hypothesis `flattenCostRoseGo` produces an element in constant time and so is the new continuation 

I.e. we break the previous chains of empty forest handling by matching on `forest2` _before_ recursing in order not to grow the continuation with a fancy form of `id`. And we also don't seem to do any extra work overall, since that matching is useful (now that `flattenCostRoseGo` receives both a node and a remaining forest), albeit slightly premature (we may run out of budget before getting to accounting for costs from `forest2`, but we don't care much about optimizing the unhappy path, since it's when the smart contracts fails anyway).

zliu41: > It seems like for long chain of one-element forest we're going to end up with [] ++ ([] ++ ([] ++ <...>)), which is effectively the same

Yeah, I think you are right.

> What do you think?

Yes I think this is indeed O(1) worst case. However, I don't quite see the benefit of CPS, because in my non-CPS version, you can also match on the `rs` and avoid concatenation if it is empty.

plutus-core/untyped-plutus-core/test/Evaluation/Machines/Budget/IdNat/3.plc.golden
```diff
@@ -6,8 +6,8 @@
 | BStep BDelay causes ({cpu: 851000 | mem: 3700})
 | BStep BForce causes ({cpu: 667000 | mem: 2900})
 | BStep BBuiltin causes ({cpu: 69000 | mem: 300})
-| BBuiltinApp Id causes ({cpu: 3 | mem: 0})
+| BBuiltinApp Id causes ({cpu: 0 | mem: 0})
 | BStartup causes ({cpu: 100 | mem: 100})})
-| budget: ({cpu: 9844103
+| budget: ({cpu: 9844100
```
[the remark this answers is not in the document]
effectfully: No, it changes the cost of an "example" builtin (specifically, the `Id` one). We used to have inconsistent costs there, I just made them all `mempty`.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
effectfully: > However, I don't quite see the benefit of CPS, because in my non-CPS version, you can also match on the `rs` and avoid concatenation if it is empty.

Yours is different. In mine we use the matching not only to check that the list is not empty, but also to retrieve its head to use it later, so in any case the matching is useful. In your case it's only useful when the list is empty, provided I got the code right:

```haskell
flattenCostRoseGo :: CostRose -> [CostRose] -> CostStream
flattenCostRoseGo (CostRose cost1 forest1) forest2 =
    case forest1 of
        [] -> case forest2 of
            []                -> CostLast cost1
            rose2' : forest2' -> CostCons cost1 $ flattenCostRoseGo rose2' forest2'
        rose1' : forest1' ->
            CostCons cost1 $ case forest1' of
                [] -> flattenCostRoseGo rose1' forest2
                _  -> flattenCostRoseGo rose1' $ forest1' ++ forest2

flattenCostRose :: CostRose -> CostStream
flattenCostRose (CostRose cost [])              = CostLast cost
flattenCostRose (CostRose cost (rose : forest)) = CostCons cost $ flattenCostRoseGo rose forest
{-# INLINE flattenCostRose #-}
```

That is one overhead, the other one is `++`. But maybe both of these overheads are negligible compared to the overhead of passing a continuation around.

I'll run the validation benchmarks with your version. If the results are the same, I'll pick it as it's much simpler than the continuation one.

Although I also wanted to try a different definition of `CostRose`.

effectfully:
-3.51% on average! Damn, didn't expect @zliu41's suggestion to perform that well. Gonna rerun the benchmarks to make sure it's not some kind of a glitch.

effectfully:
-2.39% on average now. That's a bit more reasonable, although still a noticeable improvement. Let's try something entirely different now.

effectfully:
Eek, forgot an `undefined` in the code. Rerunning.

effectfully:
-2.87% with `CostRose` being a sum type. Let's run Ziyang's version once again.

effectfully:
-3.77% on average, i.e. even more than the first time, but at the same time with more consistent results. Whatever then, I'm going to use @zliu41's version for the time being as it's the simplest one among the safest versions and I'll create a Jira ticket for investigating this further some time later.

plutus-core/satint/src/Data/SatInt.hs
```diff
@@ -25,255 +25,61 @@ import Data.Csv
...
-                   go_dn x | isTrue# (x <# y') = SI (I# x) `c` n
-                           | otherwise = SI (I# x) `c` go_dn (x +# delta)
-               in SI (I# x1) `c` go_dn x2
-
--- The following code is copied/adapted from GHC.Real.
-
-instance Real SatInt where
-  toRational (SI x) = toInteger x % 1
-
-instance Integral SatInt where
```
zliu41, line 228: Not sure removing this is a good idea. We perhaps don't need, for example, integer division on `SatInt` at the moment, but if we do in the future, it's easy to overlook the possible overflow and use ``unSatInt x `div` unSatInt y``. The instance should also be included if we ever want to publish it as a separate library.

zliu41 (approved):
This looks mergeable to me (other than the build errors), since it achieves the desired effect. To make reviewing easier we can open separate PRs for further improvements.

plutus-core/satint/src/Data/SatInt.hs
effectfully: I would've kept `Integral`, if only not its `Enum` constraint which brings in a ton of complexity, which I don't want to even think about.  I mean, we got `pred` and `succ` wrong, what guarantees do we have about the way more complicated stuff?

But maybe I should rename `toSatInt` to `unsafeToSatInt`? `unSatInt` is safe, it's only unsafe to embed an arbitrary, potentially underflown/overflown, `Int`.

zliu41: > But maybe I should rename toSatInt to unsafeToSatInt? unSatInt is safe, it's only unsafe to embed an arbitrary, potentially underflown/overflown, Int.

Great idea, I like it. Also `toSatInt` should have fewer usages than `unSatInt` so we don't litter the code with `unsafe`.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -249,61 +257,85 @@ runCostingFunTwoArguments
...
+    (ModelTwoArgumentsAddedSizes (ModelAddedSizes intercept slope)) =
+        lazy $ \costs1 costs2 ->
+            scaleLinearly intercept slope $ addCostStream costs1 costs2
 runTwoArgumentModel
-    (ModelTwoArgumentsAddedSizes (ModelAddedSizes intercept slope)) = \(ExMemory size1) (ExMemory size2) ->
-        (size1 + size2) * slope + intercept -- TODO is this even correct? If not, adjust the other implementations too.
+    (ModelTwoArgumentsSubtractedSizes (ModelSubtractedSizes intercept slope minSize)) =
+        lazy $ \costs1 costs2 -> do
+            let !size1 = sumCostStream costs1
+                !size2 = sumCostStream costs2
```
effectfully, line 284: If we wanted to, we could've opted for convertnig `costs*` to `size*` automatically via a type class, but I thought it wouldn't be worth it.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -323,19 +355,27 @@ instance Default ModelThreeArguments where
...
-    size3 * slope + intercept
+    -> CostStream
+    -> CostStream
+    -> CostStream
+    -> CostStream
+runThreeArgumentModel (ModelThreeArgumentsConstantCost c) = lazy $ \_ _ _ -> CostLast c
+runThreeArgumentModel
+    (ModelThreeArgumentsAddedSizes (ModelAddedSizes intercept slope)) =
+        lazy $ \costs1 costs2 costs3 ->
+            scaleLinearly intercept slope . addCostStream costs1 $ addCostStream costs2 costs3
```
effectfully, line 366: This is perhaps not as fast as if we had `addCostStream3` or some n-ary version with type-level tracking of the number of streams for unfolding purposes. Not sure if we care, though.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -0,0 +1,213 @@
...
+import Data.SatInt
+import Data.Text qualified as T
+import Data.Word
+import GHC.Exts (Int (I#))
+import GHC.Integer
+import GHC.Integer.Logarithms
+import GHC.Prim
+import Universe
+
+{- Note [ExMemoryUsage instances for non-constants]
```
effectfully, line 30: An old Note in a new file.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek.hs
```diff
@@ -35,6 +35,7 @@ module UntypedPlutusCore.Evaluation.Machine.Cek
     , RestrictingSt (..)
     , CekMachineCosts
     -- ** Costing modes
+    , monoidalBudgeting
```
effectfully, line 38: Used in for running `TrackCosts`.

effectfully:
The `Costing.hs` module with tests isn't ready for review yet, everything else is ready for review.

michaelpj (approved):
LGTM, haven't reviewed the big test module.

plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
```diff
@@ -418,12 +444,80 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
...
+    --
+    -- We track how many budgets get GCed by attaching a finalizer (see "System.Mem.Weak") to each
+    -- of them.
+    toBuiltinMeaning _ TrackCosts = unsafePerformIO $ do
+        -- A variable for storing the number of elements from the stream of budgets pending GC.
+        pendingGcVar <- newMVar 0
+        -- A variable for storing all the final numbers from @pendingGcVar@ appearing there right
+        -- before another GC is triggered. We store the results in reverse order for fast consing
+        -- and then 'reverse' them in the denotation of the builtin.
+        numsOfGcedVar <- newIORef []
```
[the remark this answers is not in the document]
bezirg: you could also use  atomic-primops CAS for the pendingGc MVar. I assume it is not worth it, since it is just a test

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostStream.hs
```diff
@@ -0,0 +1,128 @@
...
+{-# INLINE mapCostStream #-}
+
+-- See Note [Global local functions].
+addCostStreamGo :: CostStream -> CostStream -> CostStream
+addCostStreamGo (CostLast costL)        costsR = CostCons costL costsR
+addCostStreamGo (CostCons costL costsL) costsR = CostCons costL $ addCostStreamGo costsR costsL
+
+-- | Add two streams by interleaving their elements (as opposed to draining out one of the streams
+-- before starting to take elements from the other one). No particular reason to prefer
+-- interleaving over draining out one of the streams first.
```
bezirg, line 96: You could see it as some kind of "fairness"

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -594,6 +595,17 @@ lookupVarName varName@(NamedDeBruijn _ varIx) varEnv =
...
+spendBudgetStreamCek
+    :: GivenCekReqs uni fun ann s
+    => ExBudgetCategory fun
+    -> ExBudgetStream
+    -> CekM uni fun s ()
+spendBudgetStreamCek exCat = go where
+    go (ExBudgetLast budget)         = spendBudgetCek exCat budget
+    go (ExBudgetCons budget budgets) = spendBudgetCek exCat budget *> go budgets
+{-# INLINE spendBudgetStreamCek #-}
+
```
bezirg, line 608: Can you export this from Cek/Internal so I can use it in Debug/Internal? Btw, if you are going to rebase onto master, the Debug/Internal module is renamed to SteppableCek/Internal

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -146,21 +148,29 @@ runCostingFunOneArgument
...
+            zipCostStream
+                (runCpu mem1)
+                (runMem mem1)
 {-# INLINE runCostingFunOneArgument #-}
 
+-- | Take an intercept, a slope and a stream and scale each element of the stream by the slope and
+-- cons the intercept to the stream afterwards.
+scaleLinearly :: CostingInteger -> CostingInteger -> CostStream -> CostStream
+scaleLinearly intercept slope = addCostStream (CostLast intercept) . mapCostStream (slope *)
+{-# INLINE scaleLinearly #-}
```
bezirg, line 164: i love this

bezirg (approved):

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
```diff
@@ -1,40 +1,25 @@
...
-import GHC.Prim
+import Data.Semigroup
+import GHC.Generics
 import Language.Haskell.TH.Syntax (Lift)
 import NoThunks.Class
-import Universe
+import Text.Pretty
+import Text.PrettyBy
 
 {-
```
kwxm, line 24: The warning below should probably be moved to `ExMemoryUsage.hs`.

plutus-core/satint/src/Data/SatInt.hs
kwxm: I approve of getting rid of the division operations, since they let you take something saturated and convert it into something unsaturated without warning (ie, you could overflow and then convert back to a reasonable-looking result).

kwxm (commented):
Looks good.

plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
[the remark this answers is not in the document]
effectfully: I've changed the logic of the builtin and it now works better and tells us more without any `threadDelay` garbage. Documented all of it.

plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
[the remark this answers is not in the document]
effectfully: @michaelpj 

> I'm sure there is a good reason why we're using an MVar and an IORef for the different bits

I usually think along the lines of "Need an `MVar` => use an `MVar`". Same applies to `TVar`, `TMVar`, `PVar` and whatever `*Var` there exist. If you don't need any of those, just use an `IORef`.

I can change both to `MVar` if you prefer.

@bezirg 

> you could also use atomic-primops CAS for the pendingGc MVar.

Never heard of it! Do you know how it improves upon `MVar`?

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetStream.hs
```diff
@@ -0,0 +1,74 @@
...
+
+import Data.Coerce
+
+-- | A lazy stream of 'ExBudget's. Basically @NonEmpty ExBudget@, except the elements are
+-- stored strictly.
+--
+-- The semantics of a stream are those of the 'fold' of its elements. I.e. a stream that is a
+-- reordered version of another stream is considered equal to that stream.
+--
+-- An 'ExBudgetStream' is what one gets by zipping two 'CostStream's (one for CPU, one for memory),
```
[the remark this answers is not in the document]
effectfully: I think `zipWith` counts as zipping? Its Haddock says

> [zipWith](https://hackage.haskell.org/package/base-4.18.0.0/docs/GHC-OldList.html#v:zipWith) generalises [zip](https://hackage.haskell.org/package/base-4.18.0.0/docs/GHC-OldList.html#v:zip) by zipping with the function given as the first argument, instead of a tupling function.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -0,0 +1,235 @@
...
+-}
+
+-- | A lazy tree of costs. Convenient for calculating the costs of values of built-in types, because
+-- they may have arbitrary branching (in particular a 'Data' object can contain a list of 'Data'
+-- objects inside of it).
+--
+-- 'CostRose' gets collapsed to a lazy linear structure down the pipeline, so that we can
+-- stream the costs to the outside where, say, the CEK machine picks them up one by one and handles
+-- somehow (in particular, subtracts from the remaining budget).
+data CostRose = CostRose {-# UNPACK #-} !CostingInteger ![CostRose]
```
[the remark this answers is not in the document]
effectfully: Good point. Maybe we should use `Ex` for this kind of thing. So far I've been calling these things costs and `ExBudget`s budgets. In a sense that "cost" is a precursor to "budget" and it takes two costs to get a budget.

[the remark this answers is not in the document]
effectfully: Perhaps "metric" would be a better name indeed. We can bikeshed it to death after the PR is merged, I guess.

plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
```diff
@@ -214,40 +224,40 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
...
         makeBuiltinMeaning
             (\(n :: Integer) -> product [1..n])
-            mempty  -- Whatever.
+            (\_ _ -> ExBudgetLast mempty)  -- Whatever.
 
     toBuiltinMeaning _ver ForallFortyTwo =
         makeBuiltinMeaning
             forallFortyTwo
-            (\_ -> ExBudget 1 0)
+            (\_ -> ExBudgetLast mempty)
```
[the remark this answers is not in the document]
effectfully: It's just examples, so I don't think we should. I'll add "whatever" here and in other places just to be clear. Maybe I'll even define `whatever` as a top-level value!

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -0,0 +1,235 @@
...
+        nodeMem = CostRose 4 []
+        {-# INLINE nodeMem #-}
+
+        -- Add two 'CostRose's. We don't make this into a 'Semigroup' instance, because there exist
+        -- different ways to add two 'CostRose's (e.g. we could optimize the case when one of the
+        -- roses contains only one element or we can make the function lazy in the second argument).
+        -- Here we chose the version that is most efficient when the first argument is @nodeMem@ (we
+        -- didn't do any benchmarking though, so it may not be the most efficient one) -- we don't
+        -- have any other cases.
+        combine (CostRose cost1 forest1) (CostRose cost2 forest2) =
```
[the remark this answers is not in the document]
effectfully: It doesn't, but I felt like it would less clear if it wasn't this general and `nodeMem` was hardcoded in there. It does indeed get inlined and optimized properly.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs
effectfully: Oh right, silly me. Thanks for catching!

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Costing.hs
```diff
@@ -0,0 +1,425 @@
...
+    NonEmpty $ cost : concatMap (getNonEmpty . fromCostRose) costs
+
+instance Arbitrary CostRose where
+    arbitrary = arbitrary >>= genCostRose
+
+    shrink (CostRose cost costs) = do
+        (costs', cost') <- shrink (costs, cost)
+        pure $ CostRose cost' costs'
+
+test_multiSplitDistributionAt :: Int -> TestTree
```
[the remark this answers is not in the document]
effectfully: Yeah, I wasn't sure how to handle that either. It is a diagnostic, but I wasn't sure it was worth introducing a separate pattern in addition to `test_*`.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostStream.hs
effectfully: I don't think I can claim fairness, since we use this function to interleave three streams too and that results in twice more elements drawn from the first stream compared to the other two. Plus I just don't feel like we care about fairness there: if favoring one stream over another can result in an issue, then one could just use the same argument twice and make fairness irrelevant.

kwxm (approved):

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -0,0 +1,235 @@
...
+class ExMemoryUsage a where
+    -- Inlining the implementations of this method gave us a 1-2% speedup.
+    memoryUsage :: a -> CostRose
+
+instance (ExMemoryUsage a, ExMemoryUsage b) => ExMemoryUsage (a, b) where
+    memoryUsage (a, b) = CostRose 1 [memoryUsage a, memoryUsage b]
+    {-# INLINE memoryUsage #-}
+
+instance ExMemoryUsage (SomeTypeIn uni) where
+  memoryUsage _ = CostRose 1 []
```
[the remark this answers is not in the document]
effectfully: > Things like `CostRose 1 []` occur quite a lot.  I was going to suggest using some kind of alias for things like this (define `constantCostRose n = CostRose n []` or something), but on second thoughts that would probably just be confusing. 

I've added 

```haskell
singletonRose :: CostingInteger -> CostRose
singletonRose cost = CostRose cost []
{-# INLINE singletonRose #-}
```

and defined those instances in terms of it. I think it's better this way: it's slightly clearer and it doesn't hardcode a particular representation of `CostRose`, which is important, because there exist other sensible ones (e.g. [this](https://github.com/input-output-hk/plutus/pull/5239/commits/803c8d9fdaf66cfe70067f03feb2f59f9494bea0) commit).

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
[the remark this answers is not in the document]
effectfully: (I think I like "unit" most as it's as abstract as it should be)

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
effectfully: Yeah, we shouldn't make the slowest thing the easiest.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Costing.hs
```diff
@@ -0,0 +1,425 @@
...
+    go (CostLast cost)       = [cost]
+    go (CostCons cost costs) = cost : go costs
+
+toExBudgetList :: ExBudgetStream -> NonEmptyList ExBudget
+toExBudgetList = NonEmpty . go where
+    go (ExBudgetLast budget)         = [budget]
+    go (ExBudgetCons budget budgets) = budget : go budgets
+
+-- >>> magnitudes
+-- [(0,10),(11,100),(101,1000),(1001,10000),(10001,100000),(100001,1000000),(1000001,10000000),(10000001,100000000),(100000001,1000000000),(1000000001,10000000000),(10000000001,100000000000),(100000000001,1000000000000),(1000000000001,10000000000000),(10000000000001,100000000000000),(100000000000001,1000000000000000),(1000000000000001,10000000000000000),(10000000000000001,100000000000000000),(100000000000000001,1000000000000000000),(1000000000000000001,9223372036854775807)]
```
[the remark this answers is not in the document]
effectfully: > If you want to generate numbers between 1 and b whose lengths are uniformly distributed then you could generate uniform numbers in the range `[0,log b)` (say log_10, but it doesn't matter) and then exponentiate those, so something like

This would only generate nice round numbers, right? I don't want to discriminate against the ugly ones!

> You could also bias the numbers towards the endpoints by taking a normal distribution, chopping it in half, and swapping the halves around (there are 10,000 samples here):

That's clever. For now I simply skew generation of numbers towards smaller ones, so that when we generate streams of those we get a pretty nice distribution like this:

```
        +++ OK, passed 10000 tests:
         7.54% 1000000000001 - 10000000000000
         7.31% 100000000001 - 1000000000000
         7.06% 10000000000001 - 100000000000000
         6.99% 100000000000001 - 1000000000000000
         6.45% 10000000001 - 100000000000
         6.16% 1000000001 - 10000000000
         6.13% 1000000000000001 - 10000000000000000
         5.35% 10000000000000001 - 100000000000000000
         5.22% 100000001 - 1000000000
         4.77% 1000000000000000001 - 9223372036854775807
         4.77% 10000001 - 100000000
         4.76% 0 - 10
         4.35% 1000001 - 10000000
         4.21% 11 - 100
         4.13% 100001 - 1000000
         3.96% 100000000000000001 - 1000000000000000000
         3.75% 101 - 1000
         3.57% 10001 - 100000
         3.52% 1001 - 10000
```

I do think we can improve it, but it's perhaps not a priority right now.

effectfully:
I've addressed all the comments that we can't address later (a few not-too-important ones are left as future work) and documented everything, including the `Costing.hs` module with tests. If you haven't already reviewed the latter, please do it.

Otherwise should be ready for merging.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Costing.hs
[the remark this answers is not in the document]
kwxm: > This would only generate nice round numbers, right? I don't want to discriminate against the ugly ones!

No, I don't think so.  Generate floats/doubles in the range [0, log maxBound) and then call `exp` and take the floor.  I think that'll give you numbers like you're currently getting but more smoothly distributed.  I'll try that in your branch and see how the numbers compare with the ones above.

kwxm:
Manual benchmark results for this branch.  Looks good.
```
Script                              Strict         Lazy        Change
----------------------------------------------------------------------------------
auction_1-1                       179.6 μs	 176.3 μs      -1.8%
auction_1-2                       743.8 μs	 721.6 μs      -3.0%
auction_1-3                       733.8 μs	 712.8 μs      -2.9%
auction_1-4                       234.8 μs	 228.7 μs      -2.6%
auction_2-1                       182.9 μs	 179.5 μs      -1.9%
auction_2-2                       744.6 μs	 721.4 μs      -3.1%
auction_2-3                       988.7 μs	 947.7 μs      -4.1%
auction_2-4                       740.8 μs	 714.6 μs      -3.5%
auction_2-5                       236.9 μs	 228.9 μs      -3.4%
crowdfunding-success-1            215.9 μs	 211.8 μs      -1.9%
crowdfunding-success-2            215.7 μs	 212.0 μs      -1.7%
crowdfunding-success-3            216.1 μs	 212.8 μs      -1.5%
currency-1                        277.6 μs	 267.5 μs      -3.6%
escrow-redeem_1-1                 383.0 μs	 372.3 μs      -2.8%
escrow-redeem_1-2                 385.3 μs	 375.3 μs      -2.6%
escrow-redeem_2-1                 450.5 μs	 435.2 μs      -3.4%
escrow-redeem_2-2                 445.5 μs	 434.2 μs      -2.5%
escrow-redeem_2-3                 447.1 μs	 434.0 μs      -2.9%
escrow-refund-1                   159.2 μs	 156.7 μs      -1.6%
future-increase-margin-1          277.6 μs	 266.8 μs      -3.9%
future-increase-margin-2          601.3 μs	 581.3 μs      -3.3%
future-increase-margin-3          608.2 μs	 584.5 μs      -3.9%
future-increase-margin-4          556.6 μs	 542.6 μs      -2.5%
future-increase-margin-5          948.0 μs	 922.5 μs      -2.7%
future-pay-out-1                  278.3 μs	 266.6 μs      -4.2%
future-pay-out-2                  603.4 μs	 577.9 μs      -4.2%
future-pay-out-3                  610.7 μs	 581.9 μs      -4.7%
future-pay-out-4                  948.6 μs	 914.9 μs      -3.6%
future-settle-early-1             277.6 μs	 266.2 μs      -4.1%
future-settle-early-2             605.9 μs	 577.6 μs      -4.7%
future-settle-early-3             606.5 μs	 583.0 μs      -3.9%
future-settle-early-4             701.8 μs	 684.1 μs      -2.5%
game-sm-success_1-1               433.2 μs	 425.3 μs      -1.8%
game-sm-success_1-2               203.8 μs	 197.2 μs      -3.2%
game-sm-success_1-3               735.0 μs	 715.6 μs      -2.6%
game-sm-success_1-4               236.5 μs	 230.9 μs      -2.4%
game-sm-success_2-1               430.5 μs	 425.2 μs      -1.2%
game-sm-success_2-2               203.8 μs	 197.9 μs      -2.9%
game-sm-success_2-3               731.7 μs	 716.7 μs      -2.1%
game-sm-success_2-4               239.0 μs	 231.4 μs      -3.2%
game-sm-success_2-5               726.0 μs	 713.8 μs      -1.7%
game-sm-success_2-6               236.4 μs	 230.6 μs      -2.5%
multisig-sm-1                     447.8 μs	 430.1 μs      -4.0%
multisig-sm-2                     438.2 μs	 420.3 μs      -4.1%
multisig-sm-3                     444.6 μs	 426.9 μs      -4.0%
multisig-sm-4                     449.8 μs	 434.1 μs      -3.5%
multisig-sm-5                     643.1 μs	 624.1 μs      -3.0%
multisig-sm-6                     446.7 μs	 430.3 μs      -3.7%
multisig-sm-7                     434.3 μs	 420.2 μs      -3.2%
multisig-sm-8                     443.5 μs	 426.6 μs      -3.8%
multisig-sm-9                     447.7 μs	 430.8 μs      -3.8%
multisig-sm-10                    641.8 μs	 621.0 μs      -3.2%
ping-pong-1                       372.4 μs	 360.2 μs      -3.3%
ping-pong-2                       371.9 μs	 359.9 μs      -3.2%
ping-pong_2-1                     218.4 μs	 208.7 μs      -4.4%
prism-1                           171.8 μs	 165.6 μs      -3.6%
prism-2                           467.0 μs	 444.8 μs      -4.8%
prism-3                           398.5 μs	 384.7 μs      -3.5%
pubkey-1                          145.2 μs	 140.1 μs      -3.5%
stablecoin_1-1                    1.043 ms	 1.015 ms      -2.7%
stablecoin_1-2                    199.6 μs	 193.3 μs      -3.2%
stablecoin_1-3                    1.199 ms	 1.165 ms      -2.8%
stablecoin_1-4                    211.7 μs	 206.1 μs      -2.6%
stablecoin_1-5                    1.525 ms	 1.484 ms      -2.7%
stablecoin_1-6                    261.5 μs	 254.1 μs      -2.8%
stablecoin_2-1                    1.045 ms	 1.013 ms      -3.1%
stablecoin_2-2                    198.9 μs	 193.8 μs      -2.6%
stablecoin_2-3                    1.194 ms	 1.170 ms      -2.0%
stablecoin_2-4                    211.2 μs	 205.9 μs      -2.5%
token-account-1                   202.2 μs	 196.3 μs      -2.9%
token-account-2                   371.4 μs	 356.2 μs      -4.1%
uniswap-1                         466.2 μs	 441.5 μs      -5.3%
uniswap-2                         238.7 μs	 228.5 μs      -4.3%
uniswap-3                         2.005 ms	 1.933 ms      -3.6%
uniswap-4                         343.1 μs	 332.1 μs      -3.2%
uniswap-5                         1.297 ms	 1.250 ms      -3.6%
uniswap-6                         329.5 μs	 319.4 μs      -3.1%
vesting-1                         388.9 μs	 383.7 μs      -1.3%
```

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Costing.hs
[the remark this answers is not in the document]
kwxm: Hmmm.  Maybe not.  I tried
```
    arbitrary = (unsafeToSatInt . floor . exp) <$> (choose (0::Double, log $ fromSatInt maxBound))
```
and got 
```
     distribution of the generated CostStream values:                         OK (0.20s)
        +++ OK, passed 10000 tests:
        53.09% 1000000000000000001 - 9223372036854775807
        12.47% 100000000000000001 - 1000000000000000000
         6.19% 10000000000000001 - 100000000000000000
         4.70% 1000000000000001 - 10000000000000000
         3.34% 100000000000001 - 1000000000000000
         2.81% 10000000000001 - 100000000000000
         2.24% 1000000000001 - 10000000000000
         1.73% 10000000001 - 100000000000
         1.57% 1000000001 - 10000000000
         1.50% 100000000001 - 1000000000000
         1.38% 1000001 - 10000000
         1.35% 100000001 - 1000000000
         1.33% 10000001 - 100000000
         1.19% 10001 - 100000
         1.12% 1001 - 10000
         1.11% 0 - 10
         1.01% 100001 - 1000000
         0.95% 101 - 1000
         0.92% 11 - 100
```
It could be the case that `choose` isn't producing uniformly distributed numbers though.  I'll have to go and remind myself how QuickCheck generators work.

effectfully:
-3.12% on average. Looks fine indeed. Plus I believe we can squeeze out more, given that we didn't implement a bunch of optimizations discussed above.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Costing.hs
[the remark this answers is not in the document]
effectfully: >  Generate floats/doubles

Ah right, I missed that those were doubles rather than integers.

> It could be the case that choose isn't producing uniformly distributed numbers though. I'll have to go and remind myself how QuickCheck generators work.

No, I think it's the expected result. Note that the current generator of `SatInt`s is skewed towards generating small values, so that the generator for `CostStream` is less skewed towards bigger sums:

```
-- | Generate asymptotically bigger 'SatInt's with exponentially lower chance. This is in order to
-- make the generator of 'CostStream' produce streams whose sums are more or less evenly distributed
-- across 'magnitudes'.
instance Arbitrary SatInt where
    arbitrary = frequency . zip freqs . reverse $ map chooseSatInt magnitudes where
        freqs = map floor $ iterate (* 1.3) (2 :: Double)
```

michaelpj (approved):

plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
```diff
@@ -190,6 +199,18 @@ data BuiltinErrorCall = BuiltinErrorCall
     deriving stock (Show, Eq)
     deriving anyclass (Exception)
 
+-- | For the most part we don't care about costing functions of example builtins, hence this class
```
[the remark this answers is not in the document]
effectfully: I highly prefer the semantic clarity of `whatever`. `default` may sound like we actually use some sensible default, which we don't. That type class is just several lines of code, I think it's worth it.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Costing.hs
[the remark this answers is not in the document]
effectfully: > Strictly this is more of a diagnostic than a test (which is OK!). I suppose that you could try to have some kind of coverage check that tests that every case occurs more often than some minimum, but maybe that would risk random failures.

I'm going to claim this isn't too important to spend a lot of time on it, so I'm resolving the conversation.

PR #5244, opened 2023-04-03
Fix bug in format-benchmark-output.py

zeme-iohk:


michaelpj (approved):
Seems fragile but fine.

PR #5246, opened 2023-04-04
PLT-5498: Fix safeEncodeBits overflow

bezirg:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


michaelpj:
We *could* try and avoid the error statically by using some type-level naturals instead. GHC even provides [type-level exponentiation](https://hackage.haskell.org/package/base-4.16.0.0/docs/GHC-TypeLits.html#t:-94-), so we can check the whole property, I think.

But this is also fine!

bezirg:
> We _could_ try and avoid the error statically by using some type-level naturals instead. GHC even provides [type-level exponentiation](https://hackage.haskell.org/package/base-4.16.0.0/docs/GHC-TypeLits.html#t:-94-), so we can check the whole property, I think.
> 
> But this is also fine!

That's cool, but let's leave it for another PR.

PR #5249, opened 2023-04-07
PLT-507: Case of known constructor for PIR

zliu41:


plutus-core/plutus-ir/src/PlutusIR/Core.hs
```diff
@@ -1,5 +1,35 @@
...
+    mkTermApps,
+) where
+
+import Data.List (foldl')
 
 import PlutusIR.Core.Instance ()
 import PlutusIR.Core.Plated as Export
 import PlutusIR.Core.Type as Export
+
+data Expr tyname name uni fun ann
```
michaelpj, line 14: I don't love this naming. I would expect "expression" to be synonymous with "term", this isn't that. What was wrong with "arg"? It _is_ representing arguments?

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+    Map name [name] ->
+    Term tyname name uni fun a ->
+    m (Term tyname name uni fun a)
+go ctxt = go'
+  where
+    go' = \case
+        t@Apply{} -> do
+            let (fun, args) = collectArgs t
+            case fun of
+                Var _ n
```
michaelpj, line 79:
```suggestion
        t
            | (Var _ destr, args) <- collectArgs t
```
more consistent with below

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                    | Just cons <- Map.lookup n ctxt
```
michaelpj, line 80:
```suggestion
                    -- it appears in our map, so it's a destructor
                    | Just cons <- Map.lookup n ctxt
```

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                    , ((TermExpr scrut, _) : (TypeExpr _resTy, _) : rest) <-
+                        dropWhile (isTyArg . fst) args
```
michaelpj, line 82:
```suggestion
                        -- The datatype may have some type arguments, we
                        -- aren't interested in them, so we drop them. We can
                        -- do this easily because we know that they are all type
                        -- arguments, and then we have a term argument for the
                        -- scrutinee
                        dropWhile (isTyArg . fst) args
```

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                    , (Var _ con, conArgs) <- collectArgs scrut
```
michaelpj, line 83:
```suggestion
                    -- The scrutinee is itself an application
                    , (Var _ con, conArgs) <- collectArgs scrut
```

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                    , Just i <- List.findIndex (== con) cons
```
michaelpj, line 84:
```suggestion
                    -- ... of one of the constructors from the same datatype as the destructor
                    , Just i <- List.findIndex (== con) cons
```

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                    , Just (TermExpr branch, _) <- rest List.!? i
```
michaelpj, line 85:
```suggestion
                    -- ... and there is a  branch for that constructor in the destructor application
                    , Just (TermExpr branch, _) <- rest List.!? i
```

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                    , -- This condition ensures the destructor is fully-applied
```
michaelpj, line 86: Good, I was just thinking we needed this!

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                      -- (which should always be the case).
```
michaelpj, line 87:
```suggestion
                      -- (which should always be the case in programs that come from Plutus Tx, but not necessarily in arbitrary PIR programs).
```

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                      length rest >= length cons -> do
```
michaelpj, line 88:
```suggestion
                      -- it's okay to have more arguments: the result of the destructor might itself be
                      -- a function which is being applied!
                      length rest >= length cons -> do
```

michaelpj: Haven't looked yet, but I'd like to see a test case for this.

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                        Map.singleton destr (_varDeclName <$> cons)
+                    _ -> mempty
+                )
+                (t ^.. termSubtermsDeep . termBindings)
+    go ctxt t
+
+go ::
+    forall m tyname name uni fun a.
+    ( Ord name
+    , PLC.MonadQuote m
```
michaelpj, line 67: unused? This whole thing seems to be non-monadic in fact?

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                                branch
+                                -- The arguments to the selected branch consists of the arguments
+                                -- to the constructor (without the leading type arguments - e.g.,
+                                -- if the scrutinee is `Just {integer} 1`, we only need the `1`),
+                                -- and the remaining arguments in `rest` (because the destructor
+                                -- may be over-applied).
+                                ((dropWhile (isTyArg . fst) conArgs) <> drop (length cons) rest)
+                _ ->
+                    -- This is more efficient than `forOf termSubterms t go'`.
+                    -- The latter is quadratic in the length of `args`.
```
michaelpj, line 100: I'm not actually too worried about that so long as we note it. We have the same problem for the saturated inliner, we can just solve both together. I think I'm mildly inclined to have the simpler code and just write a Note for future fixing.

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                        <*> traverse
+                            ( bitraverse
+                                ( \case
+                                    TermExpr term -> TermExpr <$> go' term
+                                    expr          -> pure expr
+                                )
+                                pure
+                            )
+                            args
+        t -> forOf termSubterms t go'
```
michaelpj, line 112: Why not use the `transformOf` pattern? I think it's because you want to control the recursion yourself? I guess we could use it if we decide we don't mind about that.

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+    , PLC.MonadQuote m
+    ) =>
+    Term tyname name uni fun a ->
+    m (Term tyname name uni fun a)
+knownCon t0 = do
+    t <- PLC.rename t0
+    let ctxt =
+            foldMap
+                ( \case
+                    DatatypeBind _ (Datatype _ _ _ destr cons) ->
```
michaelpj, line 57: If you wanted to be really clear, you could also record the number of type arguments, so you can be sure you're discarding the correct number when you discard them later. At the moment you're relying on noticing where it changes from type to term arguments, which I think _is_ fine, but counting the number of type arguments might be even more solid.

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,114 @@
...
+                      length rest >= length cons -> do
+                        pure $
+                            mkTermApps
+                                branch
+                                -- The arguments to the selected branch consists of the arguments
+                                -- to the constructor (without the leading type arguments - e.g.,
+                                -- if the scrutinee is `Just {integer} 1`, we only need the `1`),
+                                -- and the remaining arguments in `rest` (because the destructor
+                                -- may be over-applied).
+                                ((dropWhile (isTyArg . fst) conArgs) <> drop (length cons) rest)
```
michaelpj, line 97: You could consider writing
```
-- returns Nothing if the index is out of bounds
splitAtPedantic :: Int -> [a] -> Maybe ([a], [a])
```
Then you could do
```
 | Just (_, trailingArgs) <- splitAtPedantic (length cons) rest
```
which would get you the rest of the arguments here without having to do another drop

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
michaelpj: Although if you _were_ processing prefixes recursively then you wouldn't need this: you could assert that there are exactly the right number of arguments for the case branches, and also not have to worry about trailing arguments below.

plutus-core/plutus-ir/test/transform/knownCon/list
```diff
@@ -0,0 +1,36 @@
+(let
```
michaelpj, line 1: Marty is leading the way in us adding comments into test cases explaining what they're testing, let's keep doing that :)

plutus-core/plutus-ir/test/transform/knownCon/maybe-unsaturated
```diff
@@ -0,0 +1,25 @@
...
+      (all dead (type) (con integer))
+    }
+    (lam
+      a
+      (con integer)
+      (abs dead (type) [ [ (builtin addInteger) (con integer 2) ] a ])
+    )
+  ]
+)
+-- `Maybe_match` is unsaturated. This should never happen in practice.
```
michaelpj, line 25: put it at the top?

also I want to see an over-saturated case! where the match returns a function

```
(case (Just addInteger) of 
   Just f -> f
   Nothing -> subtractInteger) 1 2
```

plutus-core/plutus-ir/test/transform/knownCon/applicative
```diff
@@ -0,0 +1,128 @@
...
+          )
+        )
+      ]
+      (abs dead (type) { Nothing (con integer) })
+    ]
+    (all dead (type) dead)
+  }
+)
+-- Adapted from `Budget.Spec.applicative`. The last `Maybe_match` scrutinizes
+-- a constructor application.
```
michaelpj, line 128: at the top please

Do we need this test given that we have the `applicative` test elsewhere?

michaelpj (commented):
Interesting but not totally surprising that this doesn't have an effect on budgets by itself.

Needed a lot more explanation of what was happening, IMO, it's not all that obvious. Suggested explanations in comments.

plutus-core/plutus-ir/src/PlutusIR/Core.hs
zliu41: It's just "either Term or Type", which can be arguments, but doesn't have to be. And I don't like the name `TermOrType`.

GHC's `Expr` can be either term or type, and this just follows that naming.

plutus-core/plutus-ir/test/transform/knownCon/applicative
zliu41: This is not the applicative budget test. That test has no know constructors. This is modified so that one of the scrutinee is a known constructor.

plutus-core/plutus-ir/test/transform/knownCon/list
```diff
@@ -0,0 +1,36 @@
...
+  (rec)
+  (datatypebind
+    (datatype
+      (tyvardecl List (fun (type) (type)))
+      (tyvardecl a (type))
+      Nil_match
```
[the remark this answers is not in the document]
zliu41: Yes, time to allow symbolic names.

plutus-core/plutus-ir/test/transform/knownCon/maybe-unsaturated
zliu41: Most if not all test cases are over-saturated, the additional argument being `(all dead (type) dead)`.

plutus-core/plutus-ir/src/PlutusIR/Core.hs
michaelpj: It's a datatype for a _specific use_, namely representing arguments. What's wrong with that?

> GHC's Expr can be either term or type, and this just follows that naming.

I don't understand this argument. GHC's `Expr` is our `Term`. Adding something that isn't the main term type called "Expr" is just confusing IMO.

plutus-core/plutus-ir/test/transform/knownCon/maybe-unsaturated
michaelpj: Aha, good point! Worth calling out explicitly.

plutus-core/plutus-ir/test/transform/knownCon/applicative
michaelpj: I guess my question is: what is this adding to the test suite?

plutus-core/plutus-ir/src/PlutusIR/Core.hs
zliu41: > It's a datatype for a specific use, namely representing arguments. What's wrong with that?

Nothing terribly wrong, but in such cases my preference is to name the datatype properly, and use a synonym `type Arg = <proper name>`

> I don't understand this argument. GHC's Expr is our Term. Adding something that isn't the main term type called "Expr" is just confusing IMO.

My point is GHC's `Expr` has a `Type` constructor so it can be either a term or a type. But if "expression" and "term" are often used interchangeably in the literature then that is indeed not a good name.

plutus-core/plutus-ir/test/transform/knownCon/applicative
zliu41: A more complex test case. Other test cases are nothing but destructor applications. This one has `Lam`, `TermBind` etc.

plutus-core/plutus-ir/src/PlutusIR/Core.hs
michaelpj: > Nothing terribly wrong, but in such cases my preference is to name the datatype properly, and use a synonym type Arg = <proper name>

Well, it's not clear what the "proper" name for this is? It's a datatype for either a term or a type. The only place that really comes up is a genericised version of "arguments"; so it seems like the sensible names are either `Arg` or `TermOrType`.

> But if "expression" and "term" are often used interchangeably in the literature then that is indeed not a good name.

Seems like they're pretty interchangeable, or used ambiguously: https://cs.stackexchange.com/a/64702

plutus-core/plutus-ir/test/transform/knownCon/applicative
michaelpj: Fair enough.

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
zliu41: Exactly

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
zliu41: Fair enough. Updated.

plutus-core/plutus-ir/src/PlutusIR/Transform/KnownCon.hs
```diff
@@ -0,0 +1,96 @@
...
+    pure $ transformOf termSubterms (go ctxt) t
+
+go ::
+    forall tyname name uni fun a.
+    (Ord name) =>
+    -- | A map from destructor to (constructors, number of type variables)
+    Map name ([name], Int) ->
+    Term tyname name uni fun a ->
+    Term tyname name uni fun a
+go ctxt t
```
michaelpj, line 73: Beautiful!

michaelpj (approved):

PR #5250, opened 2023-04-11
Flat: fix sizing of lists

michaelpj:
What I had before was wrong and was only counting the size of the
elements.

I reran the seed from https://ci.zw3rk.com/build/1924310/nixlog/3 and it
works now.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


kwxm (approved):
This looks fine to me.  However, `Flat.Encoder.Strict` purportedly has a [sizeListWith](https://hackage.haskell.org/package/flat-0.6/docs/Flat-Encoder-Strict.html#v:sizeListWith) function that does what we want.  I tried that and got 
> Module ‘Flat.Encoder.Strict’ does not export ‘sizeListWith’

That seems to have appeared in version 0.5, but we're using `< 0.5`.  I guess we're sticking with that for stability, right?

kwxm:
> I didn't notice that function

You even gave it exactly the same name!  At least the implementation's a bit different (and it's more obvious what yours is doing).

PR #5251, opened 2023-04-11
PLT-5439 ensure global uniqueness

thealmarty:
- rename before substituting in in call site inlining
- added tests that shows uniques, but it's really hard to read. Should I do property tests like [propRenameCheck](https://github.com/input-output-hk/plutus/blob/e9cbc3e84c0f17967cd92d9d26321ffd8ec08e23/plutus-core/plutus-core/test/Check/Spec.hs#L26)? Not sure how good that test is though!

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-core/changelog.d/20230411_100505_marty.stumpf_plt_5439_ensure_global_uniqueness.md
```diff
@@ -0,0 +1,7 @@
+### Added
+
+- Unit tests for global uniqueness for the PIR inliner.
```
michaelpj, line 3: I don't think we need to provide changelogs for out testsuite, generally.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -192,8 +192,10 @@ considerInlineSat tm = do
               -- it would have before, but we can make it happen more often.
               -- We could potentially do this safely in non-conservative mode.
               isBindingPure <- isTermBindingPure (varStrictness varInfo) (varDef varInfo)
-              if isBindingPure then
-                pure $ mkApps (varDef varInfo) args
+              if isBindingPure then do
+                -- rename the term before substituting in
+                renamed <- renameTerm (Done (Dupable (varDef varInfo)))
```
michaelpj, line 197: We should put the `varDef` in as `Done (dupable term)`, that's kind of the point of those types. Just exactly the same way as we do for other things that we put into the substitutions.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -192,8 +192,10 @@ considerInlineSat tm = do
               -- it would have before, but we can make it happen more often.
               -- We could potentially do this safely in non-conservative mode.
               isBindingPure <- isTermBindingPure (varStrictness varInfo) (varDef varInfo)
```
michaelpj, line 194: Incidentally, this function has an extra unnecessary level of nesting. You could do:
```
let isAcceptable = ...
isBindingPure <-
if isAcceptable && isFullyApplied ... && isBindingpure
then ...
else ...
``

File: plutus-core/plutus-ir/test/TransformSpec.hs [hunk 4]
```diff
@@ -222,6 +223,39 @@ inline =
             , "letNonPureMultiStrict"
             ]
 
+-- | Tests to ensure global uniqueness.
```
michaelpj, line 226: I don't think this is a good way to do this. It's adding a huge amount of duplicated tests that we have to check by eye. Why not add a programmatic check for global uniqueness to the existing tests? I don't think we have a function for that directly, but we could write one.

Such a check would benefit us in every other place where we want to be sure we have global uniqueness, for example.

(or is there a faster way? @effectfully ?)

effectfully: > Why not add a programmatic check for global uniqueness to the existing tests? I don't think we have a function for that directly, but we could write one.

I thought it was `PlutusCore.Check.Uniques.checkProgram` that you implemented? The one used in `[propRenameCheck](https://github.com/input-output-hk/plutus/blob/e9cbc3e84c0f17967cd92d9d26321ffd8ec08e23/plutus-core/plutus-core/test/Check/Spec.hs#L26)` referenced by Marty. I'm very confused otherwise!

michaelpj: Aha, yes, it does look like it will do that, I had forgotten.

thealmarty: Yeah. Thanks Roman. I'm wondering if we can trust that. It's not such a simple property that we can be sure it does the right thing it seems? What do you two think? Either way I feel unit tests can't hurt though. I was going to set up nicer unit tests today.

plutus-core/changelog.d/20230411_100505_marty.stumpf_plt_5439_ensure_global_uniqueness.md
thealmarty: Can we have some more general rules for change logs? It feels a bit random atm.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: Since this is monadic and thus sequential, I thought doesn't make sense to always calculate `isBindingPure`?

plutus-core/plutus-ir/test/TransformSpec.hs
effectfully: > I'm wondering if we can trust that. It's not such a simple property that we can be sure it does the right thing it seems? What do you two think?

It was Michael who implemented the uniqueness check, so I'm not quite sure, but I'd expect 

```haskell
Uniques.checkProgram (\case { FreeVariable{} -> False; IncoherentUsage {} -> False; _ -> True})
```

taken from `propRenameCheck` to work just as well in your case? I don't think you can reuse `propRenameCheck` directly, if that's what you're asking, but you should be able to reuse the line from it that I just quoted.

Well except maybe in your case you also want to throw an error upon encountering a free variable, I don't know, haven't looked into the inliner yet (and won't do it until much later unfortunately).

thealmarty: Thanks. As discussed I'll make a PIR version of `checkProgram` and add it to the existing `inline` tests.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: I guess because this it not impure it's still gonna be lazy. I will un-nest it.

plutus-core/plutus-ir/test/TransformSpec.hs
```diff
@@ -222,6 +238,11 @@ inline =
             , "letNonPureMultiStrict"
             ]
 
+-- | Check whether a term is globally unique.
+checkUniques :: (Ord a, MonadError (UniqueError a) m) => Term TyName Name uni fun a -> m ()
+checkUniques =
+    Uniques.checkTerm (\case { FreeVariable{} -> False; IncoherentUsage {} -> False; _ -> True})
```
michaelpj, line 244: I think we should also be checking for `MultiplyDefined`, since I think locally but not globally unique programs will trigger that. I think `propRenameCheck` should probably do that too.

michaelpj (changes requested):

plutus-core/plutus-ir/test/TransformSpec.hs
thealmarty: Right now, we only throw when the error is `MultiplyDefined`. For checking uniqueness, this is right? Do we want to check everything/more things in this check? I guess why not? We want to make sure the inliner doesn't mess things up in other ways too.

michaelpj: Argh, my bad, I had the meaning of the predicate backwards!

thealmarty: OK I turned on throwing the free variables error and got really strange results. See [here](https://gist.githubusercontent.com/thealmarty/bc16f24a4e45e60db06d0273d3224c8a/raw/19ff15342f381b344ff4bc12d0f802b3c9ea2e28/freeVariableInliner.md). Maybe we can investigate in another ticket. For this PR I'll only throw `MultiplyDefined`.

michaelpj: Probably because the uniques checker is going to be totally wrong because of https://github.com/input-output-hk/plutus/pull/5251#discussion_r1165314284

zliu41 (commented):
I don't understand the purpose of this PR.
Why do we need to verify global uniqueness _after_ inlining is done? I thought the compiler passes are allowed to not preserve global uniqueness, and any pass that requires global uniqueness should do renaming at the beginning?

plutus-core/plutus-ir/test/TransformSpec.hs
thealmarty: ah good point! I'm working on that now.

thealmarty: so do you think I should throw the other errors?

thealmarty:
> I don't understand the purpose of this PR. Why do we need to verify global uniqueness _after_ inlining is done? I thought the compiler passes are allowed to not preserve global uniqueness, and any pass that requires global uniqueness should do renaming at the beginning?

Renaming only works if we can be sure that we didn't mess up name capture in the first place. We can be more certain of that if there is no name shadowing after inlining. If you can think of a more fitting test, lmk.

plutus-core/plutus-ir/test/TransformSpec.hs
michaelpj: No, it's correct as it is. We want to check for global uniqueness, that's what it does.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Definitions.hs
```diff
@@ -0,0 +1,60 @@
...
+        HasUnique tyname TypeUnique,
+        MonadState (UniqueInfos ann) m,
+        MonadWriter [UniqueError ann] m)
+    => Term tyname name uni fun ann
+    -> m ()
+termDefs =
+    let -- Add bindings to definition maps.
+        -- addBindingDef :: Binding tyname name uni fun ann -
+        addBindingDef (TermBind _tmAnn _tmStrict (VarDecl varAnn n varTy) tm) =
+            addDef n varAnn TermScope >> typeDefs varTy >> tm
```
michaelpj, line 34: This is where recursion schemes don't work. Recursion schemes work for a self-recursive datatype. Here you have a term which is getting totally ignored because the recursion schemes don't "see" the recursion over here. You are going to need to rewrite it to not use recursion schemes: they just don't work for PIR because of the mutually recursive datatypes.

zliu41: Recursion schemes can work for mutually recursive datatypes but I also think the machinery is too heavy just for testing global uniqueness.

zliu41 (commented):
> We can verify that it's done this by checking whether we still have it after we've run the pass.

That's probably true, but I'm not yet 100% sure. I can imagine a counter example where during the inlining, global uniqueness is violated, but as things get inlined further, some bindings are dropped and global uniqueness holds again. Anyway, definitely worth a note.

michaelpj:
> I can imagine a counter example where during the inlining, global uniqueness is violated, but as things get inlined further, some bindings are dropped and global uniqueness holds again. Anyway, definitely worth a note.

For sure, that's true. This test doesn't guarantee that we don't make mistakes, but it _would_ have caught the mistake we did in fact make, so it seems worth doing.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Definitions.hs
michaelpj: Like I said elsewhere, these days I'm much more in favour of using our sub-term traversals rather than using `recursion-schemes`, I'd happily purge them from our codebase entirely at this point.

plutus-core/plutus-ir/test/TransformSpec.hs
thealmarty: Yes, sure. We have another ticket for adding other checks to the inliner I believe.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Definitions.hs
thealmarty: I'll change that out to sub-term traversals like in the inliner. Since this is not a trivial function, it probably should have some tests, like [multiplyDefined](https://github.com/input-output-hk/plutus/blob/a667f429516f4246e74f175e0f9407e57a4048ca/plutus-core/plutus-core/test/Check/Spec.hs#L55) but for PIR. I'll add that. I'll re-request review when I'm ready.

thealmarty:
- The tests also verifies that the unconditional inlining preserves global uniqueness, no tests were included before.
- I will a note about motivation.
- I will add a test for name capture.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Definitions.hs
michaelpj: I highly recommend updating the PLC one too if  you do that. They should be defined extremely similarly, so it should be simple enough to replace the PLC one (just delete the cases for PIR terms).

plutus-core/plutus-core/test/Check/Spec.hs
```diff
@@ -1,4 +1,3 @@
--- editorconfig-checker-disable-file
```
thealmarty, line 1: I just fix the line length in this module, no actual changes.

plutus-core/plutus-ir/test/Check/Spec.hs
```diff
@@ -0,0 +1,87 @@
...
+            let n = Name "yo" u
+            let term =
+                    Apply Ignore
+                    (LamAbs (Tag 1) n (TyVar Ignore ty) (Var Ignore n))
+                    (LamAbs (Tag 2) n (TyVar Ignore ty) (Var Ignore n))
+            checkTermUniques term
+        assertion = checked @?= Left (MultiplyDefined u (Tag 1) (Tag 2))
+    in testCase "multiplyDefined" assertion
+
+shadowedInLet :: TestTree
```
thealmarty, line 64: Other tests are the same as the PLC ones. I added this one to test `let`.

thealmarty:
> * [x]  PIR `checkTerm` etc using subterm traversals.
> 
>     * [x]  same for PLC.
> 
>     * [x]  same for UPLC.
> 
>     * [x]  add tests for `checkTerm` like [here](https://github.com/input-output-hk/plutus/blob/a667f429516f4246e74f175e0f9407e57a4048ca/plutus-core/plutus-core/test/Check/Spec.hs#L55).
> 
>     * [x]  add a test for name capture.

All done. Ready for review.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Definitions.hs
```diff
@@ -0,0 +1,108 @@
...
+-- | Given a PIR term, add all of its term and type definitions and usages, including its subterms
+-- and subtypes, to a global map.
+termDefs
+    :: (Ord ann,
+        HasUnique name TermUnique,
+        HasUnique tyname TypeUnique,
+        MonadState (UniqueInfos ann) m,
+        MonadWriter [UniqueError ann] m)
+    => Term tyname name uni fun ann
+    -> m (Term tyname name uni fun ann)
```
michaelpj, line 72: again you don't need to return  a term

plutus-core/plutus-ir/src/PlutusIR/Analysis/Definitions.hs
```diff
@@ -0,0 +1,108 @@
...
+            void $ termDefs rhs
+            forMOf termSubtypes rhs typeDefs
+        Var ann n         -> do
+            addUsage n ann TermScope
+            pure tm
+        LamAbs ann n ty t -> do
+            addDef n ann TermScope
+            void $ typeDefs ty
+            termDefs t
+        IWrap _ pat arg t -> do
```
michaelpj, line 86: handled by fallthrough

plutus-core/plutus-ir/src/PlutusIR/Analysis/Definitions.hs
```diff
@@ -0,0 +1,108 @@
...
+            void $ typeDefs pat
+            void $ typeDefs arg
+            termDefs t
+        TyAbs ann tn _ t  -> do
+            addDef tn ann TypeScope
+            void $ termDefs t
+            forMOf termSubtypes tm typeDefs
+        TyInst _ t ty     -> do
```
michaelpj, line 94: ditto

plutus-core/plutus-ir/test/transform/inline/nameCapture
```diff
@@ -0,0 +1,54 @@
+{- This test ensures that the inliner doesn't capture variable names incorrectly. E.g.,
+
+let a = 10
+    b = a -- to avoid the unconditional inlining inlining a away
+    c = b + a 
+in
+    let x = a 
+    in
+        let a = 7
```
michaelpj, line 9: This gets inlined, so I don't think the test shows what you want.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Analysis/Definitions.hs
```diff
@@ -10,25 +9,28 @@ module UntypedPlutusCore.Analysis.Definitions
...
         HasUnique name TermUnique,
         MonadState (UniqueInfos ann) m,
         MonadWriter [UniqueError ann] m)
     => Term name uni fun ann
-    -> m ()
-termDefs = cata $ \case
-    VarF ann n      -> addUsage n ann TermScope
-    LamAbsF ann n t -> addDef n ann TermScope >> t
-    x               -> sequence_ x
+    -> m (Term name uni fun ann)
```
michaelpj, line 25: ditto

michaelpj:
(Also, I would probably have split out the refactoring of the unique checker into a separate PR, but whatever)

michaelpj:
Also most of this is nitpicking, I think it basically looks good.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Definitions.hs
```diff
@@ -0,0 +1,108 @@
...
+import PlutusCore.Analysis.Definitions hiding (runTermDefs, termDefs)
+
+-- | Add bindings and their subterms and subtypes to definition maps.
+addBindingDef :: (Ord ann,
+    HasUnique name TermUnique,
+    HasUnique tyname TypeUnique,
+    MonadState (UniqueInfos ann) m,
+    MonadWriter [UniqueError ann] m)
+    => Binding tyname name uni fun ann -> m ()
+addBindingDef bd = case bd of
```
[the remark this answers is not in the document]
thealmarty: I was considering using those but I wasn't sure how to get the annotations of them too.

thealmarty:
> (Also, I would probably have split out the refactoring of the unique checker into a separate PR, but whatever)

If tests aren't required in the same PR then I would have split it out - I don't feel super comfortable pushing without the tests but for the sake of smaller PR maybe it makes sense?

plutus-core/plutus-ir/test/transform/inline/nameCapture.golden
```diff
@@ -0,0 +1,10 @@
+(let
+  (nonrec)
+  (termbind (strict) (vardecl a (con integer)) (con integer 10))
+  (termbind (strict) (vardecl c (con integer)) [ [ (builtin addInteger) a ] a ])
+  (let
+    (nonrec)
+    (termbind (strict) (vardecl a (con integer)) (con integer 7))
+    [ [ (builtin addInteger) [ [ (builtin addInteger) a ] a ] ] a ]
```
thealmarty, line 8: Oh noes!!!

thealmarty: The latest 2 commits show that the uniques are different.

thealmarty: `rename` should give fresh [names](https://github.com/input-output-hk/plutus/blob/760d54810e05e02c7c538d5e82b7a86a0fc9b3ec/plutus-core/plutus-core/src/PlutusCore/Rename.hs#L38) too it seems? I'm looking into it.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Definitions.hs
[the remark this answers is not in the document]
michaelpj: Ah right, you do need the annotation indeed. Then it's not so useful. There's one that traverses the `TyName`s, you could add one that traverses the `Name`s and then glue them together, potentially.

plutus-core/plutus-ir/test/transform/nameCapture/nameCapture.golden
```diff
@@ -0,0 +1 @@
...-34:32 (SomeTypeIn DefaultUniInteger)}) (Apply nameCapture:35:9-35:34 (Apply nameCapture:35:9-35:34 (Builtin nameCapture:35:10-35:29 AddInteger) (Var nameCapture:30:9-30:9 (Name {_nameText = \"a\", _nameUnique = Unique {unUnique = 4}}))) (Var nameCapture:35:33-35:33 (Name {_nameText = \"a\", _nameUnique = Unique {unUnique = 4}})))]) (Let nameCapture:44:9-52:13 NonRec (TermBind nameCapture:46:13-50:13 Strict (VarDecl {_varDeclAnn = nameCapture:48:17-48:41, _varDeclName = Name {_nameText = \"a\", _nameUnique = Unique {unUnique = 8}}, _varDeclType = TyBuiltin nameCapture:48:28-48:40 (SomeTypeIn DefaultUniInteger)}) (Constant nameCapture:49:17-49:31 (Some (ValueOf DefaultUniInteger 7))) :| []) (Apply nameCapture:51:13-51:63 (Apply nameCapture:51:13-51:63 (Builtin nameCapture:51:14-51:33 AddInteger) (Apply nameCapture:51:35-51:60 (Apply nameCapture:51:35-51:60 (Builtin nameCapture:51:36-51:55 AddInteger) (Var nameCapture:42:13-42:13 (Name {_nameText = \"a\", _nameUnique = Unique {unUnique = 4}}))) (Var nameCapture:51:59-51:59 (Name {_nameText = \"a\", _nameUnique = Unique {unUnique = 8}})))) (Var nameCapture:51:62-51:62 (Name {_nameText = \"a\", _nameUnique = Unique {unUnique = 8}}))))"
```
michaelpj, line 1: you can print the syntax properly showing uniques. can't remember how off the top of my head, but it's there

thealmarty: I tried multiple variants of [prettyPlcClassicDef](https://github.com/input-output-hk/plutus/blob/6485b75dd79e3a14ad27d805b6a84c7b3cb2665b/plutus-core/plutus-core/src/PlutusCore/Pretty/Plc.hs#L118) but they don't give much improvements. Or are you talking about other ways to configure the pretty printing?

plutus-core/plutus-ir/src/PlutusIR/Analysis/Definitions.hs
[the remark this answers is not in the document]
thealmarty: `termSubtermsDeep` and its type counter part do get bindings so I don't need to traverse over the subterms/types for the bindings. So it's actually pretty simple now. See what you think of the latest and please approve if you are fine with that.

plutus-core/plutus-ir/test/transform/nameCapture/nameCapture.golden
thealmarty: maybe @effectfully knows the best way?

thealmarty:
@michaelpj if there's nothing else please approve so I can merge. Thanks.

effectfully:
Sorry folks, I'll look at the pretty-printing stuff today.

plutus-core/plutus-ir/test/transform/nameCapture/nameCapture.golden
effectfully: I've updated this file to have readable syntax and removed the other one with the same name which didn't seem to be referenced from anywhere. Basically, if you need readable PIR, use `prettyPirReadable` from `PlutusIR.Core.Instance.Pretty.Readable`. Pushed.

thealmarty:
I'm now wondering if everything should be rendered that way?

plutus-core/plutus-core/src/PlutusCore/Check/Uniques.hs
```diff
@@ -40,14 +40,14 @@ checkTerm p t = do
...
-        HasUnique tyname TypeUnique,
-        AsUniqueError e ann,
-        MonadError e m)
-    => (UniqueError ann -> Bool)
-    -> Type tyname uni ann
-    -> m ()
-checkType p t = do
-    (_, errs) <- runTypeDefs t
-    for_ errs $ \e -> when (p e) $ throwing _UniqueError e
+-- checkType
```
michaelpj, line 43: commented out code

plutus-core/plutus-ir/test/TransformSpec.hs
```diff
@@ -246,6 +264,30 @@ inline =
...
+                        renamed <- PLC.rename pir
+                        Inline.inline mempty def renamed
+                -- Make sure the inlined term is globally unique.
+                _ <- checkUniques pirInlined
+                pure . render $ prettyPirReadable pirInlined
+    in
+    testNested "nameCapture" $
+        map
+            (goldenPirMUnique goldenInlineUnique pTerm)
+            [ "nameCapture"]
```
michaelpj, line 289: Why not just add this to the other list? You've duplicated the whole `goldenInlineUnique` bit just to put this in another test tree? You could just add `++ [ goldenPirMUnique goldenInlineUnique pterm "nameCapture" ]` to the end of the thing above, no?

michaelpj (approved):

plutus-core/plutus-ir/test/TransformSpec.hs
thealmarty: it doesn't work because if I do that it won't show the uniques. The two `goldenInlineUnique` are actually not the same, the one for the name capture print things out different. I'll change the name.

PR #5253, opened 2023-04-13
Fix saturated inline heuristic

michaelpj:
I was looking at this and I realised it's subtly wrong in a way we won't
notice yet, but will be a problem once we start inlining things that are
larger.

Specifically this example:
```
let x = \y. f y in x z
```
Even if we relaxed our size limits, we would currently reject this for
saturated inlinig because we would look at the `f x` and say its cost
was potentially too high. But we were always going to pay that at each
use site anyway! The extra cost is that of evaluating the definition,
i.e. the lambda.

So we have to check the size/cost of two different things. I removed
`acceptable` entirely because I think it's kind of a footgun.

We won't notice this today because anything that we might want to inline
that has non-acceptable cost will also look too big.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
@@ -165,36 +165,40 @@ isFullyApplied (hdLams:tlLams) (hdArg:tlArg) =
       -- inlining, and it won't make it any worse, so we could consider accepting this.
       False
 
--- | Consider whether to inline a single node. We inline fully applied functions iff the body of
--- the function is `acceptable`.
+-- | Consider whether to inline a saturated application.
 considerInlineSat :: forall tyname name uni fun ann. InliningConstraints tyname name uni fun
     => Term tyname name uni fun ann -- ^ The `body` of the `Let` term.
     -> InlineM tyname name uni fun ann (Term tyname name uni fun ann)
 considerInlineSat tm = do
     -- collect all the arguments of the term being applied to
-    let (fun, args) = collectArgs tm
-    case fun of
+    case collectArgs tm of
       -- if it is a `Var` that is being applied to, check to see if it's fully applied
-      Var _ann name -> do
+      (Var _ann name, args) -> do
         maybeVarInfo <- gets (lookupVarInfo name)
         case maybeVarInfo of
-          -- the variable maybe a *recursive* let binding, in which case it won't be in the map,
-          -- and we don't process it. ATM recursive bindings aren't inlined.
-          Nothing -> pure tm
           Just varInfo -> do
-            let isAcceptable = acceptable (varBody varInfo)
-            -- if the body of `Var` is `acceptable` and
-            -- it is fully applied (over-application is allowed) then
-            if isAcceptable && isFullyApplied (arity varInfo) (map fst args)
-            then do
-              -- check if binding is pure to avoid duplicated effects.
-              -- For strict bindings we can't accidentally make any effects happen less often than
-              -- it would have before, but we can make it happen more often.
-              -- We could potentially do this safely in non-conservative mode.
-              isBindingPure <- isTermBindingPure (varStrictness varInfo) (varDef varInfo)
-              if isBindingPure then
-                pure $ mkApps (varDef varInfo) args
-              else pure tm -- term is not pure, don't inline by default.
-            else pure tm
-      -- if the term being applied is not a `Var`, don't inline
+            let body = varBody varInfo
+                def = varDef varInfo
+                fullyApplied = isFullyApplied (arity varInfo) (map fst args)
+            -- It is the body that we will be left with in the program after we have
+            -- reduced the saturated application, so the size increase we will be left
+            -- with comes from the body, and that is what we need to check is okay
+                bodySizeOk = sizeIsAcceptable body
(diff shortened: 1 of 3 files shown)


michaelpj:
Sorry, this conflicts with your PR @thealmarty , but once I'd spotted this I just wrote down the fix. Annoyingly I don't think we can write tests without actually relaxing the size heuristic, which we should probably do elsewhere.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -165,36 +165,40 @@ isFullyApplied (hdLams:tlLams) (hdArg:tlArg) =
...
+                fullyApplied = isFullyApplied (arity varInfo) (map fst args)
+            -- It is the body that we will be left with in the program after we have
+            -- reduced the saturated application, so the size increase we will be left
+            -- with comes from the body, and that is what we need to check is okay
+                bodySizeOk = sizeIsAcceptable body
+            -- The definition itself will be inlined, so we need to check that the cost
+            -- of that is acceptable. Note that we do _not_ check the cost of the _body_.
+            -- We would have paid that regardless.
+            -- Consider e.g. `let y = \x. f x`. We pay the cost of the `f x` at every call
+            -- site regardless. The work that is being duplicated is the work for the labmda.
```
thealmarty, line 191:
```suggestion
            -- site regardless. The work that is being duplicated is the work for the lambda.
            -- We don't need to check the size of the definition because it will be reduced to the body, whose size we already checked.
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -165,36 +165,40 @@ isFullyApplied (hdLams:tlLams) (hdArg:tlArg) =
...
+            -- check if binding is pure to avoid duplicated effects.
+            -- For strict bindings we can't accidentally make any effects happen less often than
+            -- it would have before, but we can make it happen more often.
+            -- We could potentially do this safely in non-conservative mode.
+            defPure <- isTermBindingPure (varStrictness varInfo) def
+            pure $
+              if fullyApplied && bodySizeOk && defCostOk && defPure
+              then mkApps def args
+              else tm
+          -- We should have variable info for everything, but if we don't just give up
```
thealmarty, line 202: This was the comment I had in the nothing case:
```suggestion
          -- The variable maybe a *recursive* let binding, in which case it won't be in the map,
          -- and we don't process it. ATM recursive bindings aren't inlined.
```

thealmarty: Wait nvm, ignore my comment.

thealmarty: Actually I do think that we may see *recursive* bindings here right?

thealmarty (approved):
Thanks yeah this is more correct.

zliu41:
It seems handling zero-arity functions really complicates things without any benefit - it requires checking `costIsAcceptable` and purity (which can be omitted otherwise), and doesn't inline any more things not inlined uncondiontally.

michaelpj:
> It seems handling zero-arity functions really complicates things without any benefit - it requires checking costIsAcceptable and purity (which can be omitted otherwise), 

I don't agree with this. Think of it this way, we can either say:
1. When you inline a binding, you always have to check that the cost is acceptable and that it doesn't move effects; vs
2. When you inline a binding, you have to check that the cost is acceptable and that you don't move effects, except in this special case where we know that it's a function, which relies on arity analysis.

Personally I find 1 simpler: I like universal statements :) There's an interesting duality here: would you rather have a simpler check on a restricted subset of cases, or a more complex check on a larger set of cases? I think I generally prefer the more complex check on the larger set of cases if it's not too ad-hoc, often it feels enlightening.

I also think that it makes it easier to spot the issue here. If we weren't inlining 0-arity things, then the mistake would have been to check the cost at all. Which seems a bit weird: shouldn't we need to check the cost of something? Whereas here it's clear what we did wrong: we checked the cost of the wrong thing, because we have a consistent rule that we always apply.

> and doesn't inline any more things not inlined uncondiontally.

I think this is currently true, but it might not be in future. The heuristics could come apart in future, e.g. imagine if we had different size thresholds for saturated inlining. What we're doing now is future-proof because it's general, IMO.

That said, it's not a huge deal, probably largely personal preference.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: I meant recursive *variables*, in the subterms.

michaelpj: I was wrong! The original comment was right, we _don't_ put recursive bindings into the variable info map!

thealmarty:
Yeah this PR doesn't add extra checks because of zero-arity functions being included. I do find that conceptually it's a bit confusing to include 0-arity functions in the inlining mechanics of saturated functions but I think it's not too bad.

I'll leave this as is unless there's a strong opinion to do otherwise.

PR #5255, opened 2023-04-14
SOPs for plutus-ir and plutus-tx

michaelpj:
This is the second batch of SOPs work. There will be one more batch with a few standalonwe Plutus Tx bits, like the SOP-based lifting.

Notable things:
- `PlutusIR.Compiler.Datatype` has all the logic for compilation. I've attempted to keep the Notes updated and useful. Note that we support _both_ ways of compiling datatypes, for backwards compatibility.
- `PlutusIR.Compiler` has to work on `Program`s now, because we care about the version so we can ensure that we don't try to use the wrong compilation strategy on old versions.
- Various Plutus Tx stuff now has to be version-aware, which is somewhat painful. The `Lift` related stuff is particularly not great, but I'm not sure how to make it better.
- There are plugin options to control target version and compilation style. I didn't try to make it pick a sensible default based on the version: if you pick and old version you also need to pick the old compilation style.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


doc/read-the-docs-site/tutorials/AuctionValidator.hs
```diff
@@ -193,5 +193,5 @@ auctionValidatorScript ::
   CompiledCode (BuiltinData -> BuiltinData -> BuiltinData -> ())
 auctionValidatorScript params =
   $$(PlutusTx.compile [||auctionUntypedValidator||])
-    `PlutusTx.unsafeApplyCode` PlutusTx.liftCode params
+    `PlutusTx.unsafeApplyCode` PlutusTx.liftCodeDef params
```
zliu41, line 196: Should we do this in the tutorial? This won't work if someone actually tries to run this on testnet, does it? Because plutus core 1.1.0 hasn't been allowed yet.

zliu41 (commented):
Does it ever make sense to choose different encodings (SOP vs Scott) for different datatypes in the same script?

plutus-tx-plugin/test/Budget/toFromData.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 13616457
-| mem: 43192})
\ No newline at end of file
+({cpu: 12995457
```
zliu41, line 1: I expected a larger improvement than this for this test case. SOP is much closer to Data than Scott is, and I thought to/from Data would become way cheaper.

plutus-benchmark/lists/test/Sum/left-fold-built-in.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 160559654
```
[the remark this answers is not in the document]
michaelpj: Yes, that is a little odd. Especially given that this is the one doing a fold over a builtin list :thinking:

plutus-tx-plugin/src/PlutusTx/Options.hs
```diff
@@ -32,10 +37,11 @@ import PyF (fmt)
 
 import Text.Read (readMaybe)
 import Type.Reflection
-import UntypedPlutusCore qualified as UPLC
 
 data PluginOptions = PluginOptions
-    { _posDoTypecheck                    :: Bool
+    { _posPlcTargetVersion               :: PLC.Version
```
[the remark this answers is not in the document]
michaelpj: Yep, forgot to insert the check that uses it :roll_eyes: ]

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -189,10 +176,89 @@ forall (t_1 :: *) .. (t_n :: *) .
 
 The term for the destructor is:
 /\(t_1 :: *) .. (t_n :: *) .
-  \(x :: (T t_1 .. t_n)) ->
+  \(x :: (T t_1 .. t_n)) .
     x
+-}
+
+{- Note [SOP encoding of datatypes]
```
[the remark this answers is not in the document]
michaelpj: The point is to have a note for each that just explains carefully how it works, since they're both quite fiddly. I'll add a third note that links them.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -189,10 +176,89 @@ forall (t_1 :: *) .. (t_n :: *) .
...
+
+For example, consider 'Maybe a'. Then:
+- The type corresponding to the datatype is:
+  Maybe = \(a :: *) -> sop [] [a]
+- The type of the constructors are:
+  Just : forall (a :: *) . a -> Maybe a
+  Nothing : forall (a :: *) . Maybe a
+- The terms for the constructors are:
+  Just = \(arg1 : a). constr (Maybe a) a
+  Nothing = constr (Maybe a)
```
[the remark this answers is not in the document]
michaelpj: Yep

michaelpj:
> Does it ever make sense to choose different encodings (SOP vs Scott) for different datatypes in the same script?

I don't think so. I think the only reason to choose Scott encoding is if you are targeting a language version that doesn't support SOPs.

doc/read-the-docs-site/tutorials/AuctionValidator.hs
michaelpj: Well, they won't be able to use these versions of the libraries either! But yes, I guess we can include an explicit version here.

effectfully:
Some `nofib` programs got worse, but those slowdowns become speedups with my [patch](https://github.com/input-output-hk/plutus/pull/5266) (and all other `nofib` programs become faster compared to this branch as well, see [this](https://github.com/input-output-hk/plutus/pull/5266#issuecomment-1516974380)).

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -258,28 +334,56 @@ For a (self-)recursive datatype we have to change three things:
...
 -- | Make the "Scott-encoded" type for a 'Datatype', with type variables free.
+-- This is exactly the type of an eliminator function for the datatype.
+--
 -- @mkScottTy Maybe = forall out_Maybe. out_Maybe -> (a -> out_Maybe) -> out_Maybe@
-mkScottTy :: ann -> Datatype TyName Name uni ann -> TyName -> Type TyName uni ann
-mkScottTy ann (Datatype _ _ _ _ constrs) res =
+mkScottTy :: MonadQuote m => ann -> Datatype TyName Name uni ann -> m (Type TyName uni ann)
+mkScottTy ann d@(Datatype _ _ _ _ constrs) = do
+   res <- resultTypeName d
+   -- FIXME: normalize datacons' types also here
```
effectfully, line 345: Why?

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -258,28 +334,56 @@ For a (self-)recursive datatype we have to change three things:
...
+    mkDatatypeType List <pattern functor of List>
+        = fix List . <pattern functor of List>
+        = fix List . \(a :: *) -> forall (r :: *) . r -> (a -> List a -> r) -> r
+@
+
+SOPs:
+@
+    mkDatatypeType List <pattern functor of List>
+        = fix List . <pattern functor of List>
+        = fix List . \(a :: *) -> sop [] [a, List a]
```
effectfully, line 379: Great comment.

Not a fan of it suggesting that `fix` is a binder though. I normally write it as `fix \(List :: * -> *) (a :: *) -> sop [] [a, List a]` for example.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -258,28 +334,56 @@ For a (self-)recursive datatype we have to change three things:
...
-mkDatatypeType r pf (Datatype ann tn tvs _ _) = case r of
-    NonRec -> PlainType <$> (PLC.mkIterTyLam <$> pure tvs <*> pure pf)
+      -- c_1 -> .. -> c_n -> res
+      PIR.mkIterTyFun ann caseTys (TyVar ann res)
+
+mkDatatypeSOPTy :: ann -> Datatype TyName Name uni ann -> Type TyName uni ann
+mkDatatypeSOPTy ann (Datatype _ _ _ _ constrs) = TySOP ann (fmap constructorArgTypes constrs)
+
+-- | Make the "pattern functor" of a 'Datatype'. This is just the normal type, but with the
+-- type variable for the type itself free and its type variables free.
```
effectfully, line 357: Not quite! The pattern functor for, say, `list` is

```
listF = \(list :: * -> *) (a :: *) -> all (r :: *). r -> (a -> list a -> r) -> r
```

with `list` defined in terms of it as

```
list = \(a :: *) -> ifix listF a
```

There's no free variables in there anywhere. If we strip lambdas off of `listF`, we'll get "the body of the pattern functor". Well, according to the terminology I've been using anyway, maybe I've been butchering it, not sure.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -258,28 +334,56 @@ For a (self-)recursive datatype we have to change three things:
...
+--
+-- Scott: @mkDatatypePatternFunctor List = forall (r :: *) . r -> (a -> List a -> r) -> r@
+-- SOPs: @mkDatatypePatternFunctor List = sop [] [a, List a]@
+mkDatatypePatternFunctor :: MonadQuote m => DatatypeCompilationOpts -> ann -> Datatype TyName Name uni ann -> m (Type TyName uni ann)
```
effectfully, line 361: So this is `mkPatternFunctorBody` (I don't think you need to mention that it's a `Datatype` thing, since it's implied by `PatternFunctor`).

effectfully (approved):
Got tired reviewing `Datatype.hs`. I may or may not return to it later (it's probably hard to get something wrong there without triggering any kind of test failure).

plutus-benchmark/nofib/test/formulaBudget.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 26010640908
-| mem: 111558948})
\ No newline at end of file
+({cpu: 11717589908
+| mem: 49415248})
```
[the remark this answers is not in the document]
michaelpj: Well, the wall-clock improvement was 33%, and this is a 55% reduction. So it was a pretty dramatic reduction in the real time too. But I agree that we might be understating the budget. Not sure what to do about it, maybe we can rerun the  benchmarks for the CEK machine steps?

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -342,6 +444,40 @@ mkConstructor dty d@(Datatype ann _ tvs _ constrs) index = do
...
+
+            pure $ Constr ann unrolled index (fmap (PIR.mkVar ann) argsAndTypes)
+          ScottEncoding -> do
+            resultType <- resultTypeName d
+
+            -- case arguments and their types
+            casesAndTypes <- do
+                  -- these types appear *outside* the scope of the abstraction for the datatype, so we need to use the concrete datatype here
+                  -- see note [Abstract data types]
+                  -- FIXME: normalize datacons' types also here
```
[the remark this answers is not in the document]
michaelpj: This is a confusing comment, but I _think_ this is talking about normalization in the GHC world, in particular getting rid of type families. We have to do that in a few places. Not 100% sure though, -1 point to past me.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -189,10 +186,89 @@ forall (t_1 :: *) .. (t_n :: *) .
...
+
+For example, consider 'Maybe a'. Then:
+- The type corresponding to the datatype is:
+  Maybe = \(a :: *) -> sop [] [a]
+- The type of the constructors are:
+  Just : forall (a :: *) . a -> Maybe a
+  Nothing : forall (a :: *) . Maybe a
+- The terms for the constructors are:
+  Just = \(arg1 : a). constr (Maybe a) 0 a
+  Nothing = constr 1 (Maybe a)
```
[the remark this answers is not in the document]
michaelpj: I got it wrong indeed. I made it more explicit so I don't get confused!

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -189,10 +186,89 @@ forall (t_1 :: *) .. (t_n :: *) .
...
+  match_Maybe : forall (a :: *) . Maybe a -> forall (out_Maybe :: *) . out_maybe -> (a -> out_maybe) -> out_maybe
+- The term for the destructor is:
+  match_Maybe = /\(a :: *) -> \(x : Maybe a) -> /\(out_Maybe :: *) -> \(case_Just :: a -> r) (case_Nothing :: r) -> case out_Maybe x case_Just case_Nothing
+
+-- General case
+
+Consider a datatype as defined in Plutus IR:
+(datatype
+  (tyvardecl T (type))
+  (tyvardecl a_1 t_1)
```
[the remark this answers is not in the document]
michaelpj: It does actually matter because as it is I'm getting the constructor tags wrong!

plutus-benchmark/nofib/test/formulaBudget.budget.golden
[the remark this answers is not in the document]
effectfully: > Well, the wall-clock improvement was 33%, and this is a 55% reduction. So it was a pretty dramatic reduction in the real time too. 

Sure, but the difference is quite substantial, which is my point.

michaelpj:
I want to know though :D

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -189,10 +186,89 @@ forall (t_1 :: *) .. (t_n :: *) .
...
+/\(t_1 :: *) .. (t_n :: *) .
+  \(a_1 : t_c_j_1) .. (a_c_i_(c_i_k) : t_c_i_(c_i_k)) .
+    constr (T t_1 .. t_n) i (T t_1 .. t_n) a_1 .. a_c_i_(c_i_j)
+
+The type of the destructor is:
+forall (t_1 :: *) .. (t_n :: *) .
+  (T t_1 .. t_n) ->
+    forall out_T :: * .
+      (t_c_1_1 -> .. -> t_c_1_(c_1_k) -> out_T) -> .. -> (t_c_j_1 -> .. -> t_c_j_(c_j_k) -> out_T) ->
+        out_T
```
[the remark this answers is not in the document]
michaelpj: fair enough lol. I'm unsure whether these are worth it, but I wrote them already so :shrug:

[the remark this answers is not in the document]
effectfully: > I'm unsure whether these are worth it

I do write comments like that sometimes and I think they're helpful. I'm just a lazy ass :smile_cat:

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -342,6 +444,40 @@ mkConstructor dty d@(Datatype ann _ tvs _ constrs) index = do
...
+    constrBody <- case _dcoStyle opts of
+          SumsOfProducts -> do
+            -- We have to be a bit careful annotating the type of the constr. It is inside the 'wrap' so it
+            -- needs to be one level "unrolled".
+
+            -- The pattern functor with a hole in it
+            pf <- mkDatatypePatternFunctor opts ann d
+            -- ... and with the hole filled in with the datatype type, by making
+            -- a type lambda and immediately applying it
+            let unrolled = TyApp ann (PIR.mkIterTyLam [tv] pf) (getType dty)
```
michaelpj, line 456: @effectfully I think this might be wrong, or at least suspicious. I think this is exactly the type of the term "inside" a `wrap`. I don't know if we have a way to get that with the current recursive types machinery or whether it would be useful to add one?

michaelpj: No, it's fine. Roughly, inside a `wrap` we expect the term to have this complex type which is the pattern functor applied to the ifixed version of the pattern functor i.e. "unrolled" one time. That's exactly the type we need here, but we need to actually write it down on a term, which we don't normally, and it's annoying. And I'm suspicious I might do it wrong, hence why I'm wondering if the machinery might do it for us.

doc/read-the-docs-site/tutorials/AuctionValidator.hs
```diff
@@ -193,5 +196,5 @@ auctionValidatorScript ::
   CompiledCode (BuiltinData -> BuiltinData -> BuiltinData -> ())
 auctionValidatorScript params =
   $$(PlutusTx.compile [||auctionUntypedValidator||])
-    `PlutusTx.unsafeApplyCode` PlutusTx.liftCode params
+    `PlutusTx.unsafeApplyCode` PlutusTx.liftCode plcVersion100 params
```
bezirg, line 199: Idea: Could our plutus ghc plugin be used also to fill in the target-version in the `liftCode`?

The idea would be that:
a) `liftCode` would behave as the current `liftCodeDef` if it was not preprocessed by our plutus ghc plugin.
b) The plutus plugin would transform every  call to `liftCode` it finds to `liftCodeVer passedTargetVersion`.

It would be from one hand more user-friendly, but on the hand more 'magical' thus more error-prone.

bezirg: As an added extra, it could also be controlled behind a boolean plugin flag "addVersionsToLiftCodes"

bezirg (approved):

doc/read-the-docs-site/tutorials/AuctionValidator.hs
michaelpj: I'm generally anti magic :) I like the fact that at the moment our plugin pretty much only touches stuff that's clearly signalled with a special TH function, it makes it very clear where the cleverness is happening.

The more magical approach that I could be persuaded to consider would be something like promoting `Version` to the type level and having a `KnownVersion` class. Then it could be hidden that way.

plutus-tx-plugin/src/PlutusTx/Plugin.hs
```diff
@@ -394,49 +397,49 @@ runCompiler moduleName opts expr = do
...
+                 & set (PIR.ccOpts . PIR.coDoSimplifierUnwrapCancel)       (opts ^. posDoSimplifierUnwrapCancel)
+                 & set (PIR.ccOpts . PIR.coDoSimplifierBeta)               (opts ^. posDoSimplifierBeta)
+                 & set (PIR.ccOpts . PIR.coDoSimplifierInline)             (opts ^. posDoSimplifierInline)
                  & set (PIR.ccOpts . PIR.coInlineHints)                    hints
-                 & set (PIR.ccOpts . PIR.coRelaxedFloatin) (_posRelaxedFloatin opts)
+                 & set (PIR.ccOpts . PIR.coRelaxedFloatin) (opts ^. posRelaxedFloatin)
+                 -- We could make this configurable with an option, but:
+                 -- 1. The only other choice you can make is new version + Scott encoding, and there's really no reason to pick that
+                 -- 2. This is consistent with what we do in Lift
+                 & set (PIR.ccOpts . PIR.coDatatypes . PIR.dcoStyle) (if plcVersion < PLC.plcVersion110 then PIR.ScottEncoding else PIR.SumsOfProducts)
```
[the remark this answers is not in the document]
michaelpj: Yes it is. I'm not sure what to do about it. They're both clients of the compiler, and they make a choice of a compiler option for the user... `validateOpts` is doing the inverse, though.

plutus-benchmark/script-contexts/test/checkScriptContextEqualityOverhead-20.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 37927100
-| mem: 165000})
\ No newline at end of file
+({cpu: 45770100
+| mem: 199100})
```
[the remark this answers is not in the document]
michaelpj: These ones are weird. There's nothing going on in the UPLC except swapping the SOP and scott-encoded constructors. This one in particular is measuring overhead from constructing a big value. So maybe constructors are slightly more expensive? I think that might be true, some cost is moved:
```
-- pays for three applications, the constr node, and three var nodes
(\x y z . constr 0 x y z) 1 2 3
-- pays for three applications, and a lambda node
(\x y z . \case_A case_B -> case_A x y z
```
Of course, this is pretty minor and the scott-encoded version has to pay for the variable evaluations every time you deconstruct instead. But there is a way in which the SOP version can be more expensive if you never deconstruct stuff!

michaelpj:
I fixed almost all the regressions. There were some missing cases in the purity analysis for UPLC, so we weren't inlining as much. Fixed!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -356,10 +356,17 @@ isPure = go True
             -- This case is not needed in PIR's `isPure`, because PIR's beta-reduction pass
             -- turns terms like this into `Let` bindings.
             Apply _ (LamAbs _ _ body) arg -> go True arg && go delayAndVarIsPure body
+            -- Applications can do work
```
thealmarty, line 359: Maybe we could add an `Apply _ (Builtin ...` case?

michaelpj: That's explicitly not handled, there's a note on the `Builtin` case that points to the note about differences to the PIR inliner.

michaelpj:
Okay, I think this is good to go.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -121,78 +122,153 @@ We also need to think about the types. In general, we need:
...
-For example, consider 'Maybe a'. Then:
+For example, consider 'data Maybe a = Nothing | Just a'. Then:
 - The type corresponding to the datatype is:
-  Maybe = \(a :: *) -> forall out_Maybe . out_maybe -> (a -> out_Maybe) -> out_Maybe
+  Maybe = \(a :: *) -> forall out_Maybe . out_Maybe -> (a -> out_Maybe) -> out_Maybe
 - The type of the constructors are:
   Just : forall (a :: *) . a -> Maybe a
   Nothing : forall (a :: *) . Maybe a
 - The terms for the constructors are:
   Just = \(arg1 : a). /\ (out_Maybe :: *) . \(case_Nothing : out_Maybe) (case_Just : a -> out_Maybe) . case_Just arg1
```
effectfully, line 132: `Just` should bind `a` with a `/\`.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -121,78 +122,153 @@ We also need to think about the types. In general, we need:
...
-  Nothing = /\ (out_Maybe :: *) . \(case_Nothing : out_Maybe) (caseJust : a -> out_Maybe) . caseNothing
+  Nothing = /\ (out_Maybe :: *) . \(case_Nothing : out_Maybe) (case_Just : a -> out_Maybe) . case_Nothing
```
effectfully, line 133: And `Nothing` too.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -121,78 +122,153 @@ We also need to think about the types. In general, we need:
...
+- The type of the constructors are:
+  Just : forall (a :: *) . a -> Maybe a
+  Nothing : forall (a :: *) . Maybe a
+- The terms for the constructors are:
+  Just = \(arg1 : a). constr (Maybe a) 1 a
+  Nothing = constr 0 (Maybe a)
+- The type of the destructor is:
+  match_Maybe : forall (a :: *) . Maybe a -> forall (out_Maybe :: *) . out_Maybe -> (a -> out_Maybe) -> out_Maybe
+- The term for the destructor is:
+  match_Maybe = /\(a :: *) -> \(x : Maybe a) -> /\(out_Maybe :: *) -> \(case_Nothing :: out_Maybe) (case_Just :: a -> out_Maybe) -> case out_Maybe x case_Nothing case_Just
```
effectfully, line 218: Is that really the term and not `(\(x' : a)  -> case_Just x'` instead of just `case_Just`?

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -121,78 +122,153 @@ We also need to think about the types. In general, we need:
...
+-- General case
+
+Consider a datatype as defined in Plutus IR:
+(datatype
+  (tyvardecl T (type))
+  (tyvardecl a_0 t_0)
+  ..
+  (tyvardecl a_n t_n)
+  match_T
+  (vardecl c_0 (fun t_c_0_0 .. (fun t_c_0_(c_1_k)) [T t0 .. tn]))
```
effectfully, line 229: `1`

effectfully (commented):
Another chunk, only reviewed the docs so far and not completely, because there's plenty of duplication and I think it would make sense for me to review the duplicated parts once comments are addressed.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
effectfully: > @effectfully I think this might be wrong, or at least suspicious. I think this is exactly the type of the term "inside" a `wrap`. I don't know if we have a way to get that with the current recursive types machinery or whether it would be useful to add one?

I've spent hours staring at that single line. I'm going to take look at it from a different perspective.

And I'm still kinda astonished that it's possible to make `ifix` interoperate with sums-of-products in such an extremely concise and non-intrusive way.

effectfully (commented):
Another chunk.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -292,45 +397,43 @@ mkDatatypeValueType ann (Datatype _ tn tvs _ _)  = PIR.mkIterTyApp ann (PIR.mkTy
...
+            wrap <pattern functor of List> /\(out_List :: *) .
+                \(case_Nil : out_List) (case_Cons : a -> List a -> out_List) . case_Cons arg1 arg2
+@
+
+SOPs:
+@
+    mkConstructor <definition of List> List Cons
+        = /\(a :: *) . \(arg1 : a) (arg2 : <definition of List> a) .
+            wrap <pattern functor of List> /\(out_List :: *) .
+                constr ((\(List :: * -> *) . <pattern functor of List>) <definition of List>) arg1 arg2
```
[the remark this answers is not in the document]
effectfully: No, ignore this comment, I forgot the syntax lol.

effectfully, line 430: Those two occurrences of `<pattern functor of List>` make it seem like they are the same thing, but they're not:

1. the latter is a pattern functor body, not an entire pattern functor
2. the former is what I call "packed" in the compilation machinery, i.e. 1-ary one after all the encoding tricks are done, and the latter is the body of an original "unpacked" n-ary pattern functor

effectfully (commented):
OK, I think I'm now done reviewing it. Phew!

Great job!

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
michaelpj: Yes, it doesn't have to be eta-expanded

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
michaelpj: I'm leaving this in because I don't understand it :/

PR #5260, opened 2023-04-17
[Costing] 'newtype' 'intercept' and 'slope'

effectfully:
To make sure we don't confuse the two.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
@@ -9,6 +9,8 @@
 {-# LANGUAGE StrictData            #-}
 module PlutusCore.Evaluation.Machine.CostingFun.Core
     ( CostingFun(..)
+    , Intercept(..)
+    , Slope(..)
     , ModelAddedSizes(..)
     , ModelSubtractedSizes(..)
     , ModelConstantOrLinear(..)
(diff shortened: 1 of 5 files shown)


effectfully:
Needs a changelog entry, otherwise ready for review.

michaelpj (approved):

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -177,51 +192,51 @@ runOneArgumentModel (ModelOneArgumentLinearCost (ModelLinearSize intercept slope
 
 -- | s * (x + y) + I
 data ModelAddedSizes = ModelAddedSizes
-    { modelAddedSizesIntercept :: CostingInteger
-    , modelAddedSizesSlope     :: CostingInteger
+    { modelAddedSizesIntercept :: Intercept
```
kwxm, line 195: Maybe we should make all of these model types have a sub-record that describes a linear function via named fields or something, but let's think about that some other time.  This solve the problem for the time being.

kwxm (approved):
Yep, looks fine.

kwxm:
So it's _not_ just me that's getting these "Unexpected end-of-file" errors.

effectfully:
Oh, what's that? Repushing.

effectfully:
@michaelpj any idea what's going on with CI?

michaelpj:
Weird issue that other people have been seeing too, I'm just restarting the jobs for now.

PR #5264, opened 2023-04-18
PLT-5438: Relax `sizeIsAcceptable`

zliu41:
See Note [Size criteria for inlining].


michaelpj (commented):
I haven't looked through the implementation, but I think there's a much simpler way of doing what you want to do.

When we do saturated inlining, and when we do inlining in general, we _have the original node that we're replacing_. For saturated inlining that's the application. We can just compute the size of the original term, and the size of the thing we're inlining, and compare them.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -311,38 +315,118 @@ costIsAcceptable = \case
...
+          St
+            { _stRemainingSize = 1 + 2 * length params
+            , _stUnseenParams = params
+            , _stSeenParams = mempty
+            }
+        )
+        . go
+  where
+    go :: Term tyname name uni fun ann -> State (St name) Bool
+    go term = ifM (use stRemainingSize <&> (< 0)) (pure False) $ do
```
michaelpj, line 355: I think we should use a standalone size metric, for example the one defined in `Size.hs`. Then we can just compare it, rather than this complicated mix of computing it and traversing.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -311,38 +315,118 @@ costIsAcceptable = \case
...
+RHS is usually the definition of the variable. But when we are considering inlining
+fully-saturated applications, LHS can have parameters.
+
+For example, given `x = \a b c d -> f y`, normally LHS is `x`. But when we are considering
+inlining a call site of `x` where `x` is fully applied, LHS becomes `x a b c d`, and
+RHS becomes `f y`.
+
+In order to avoid increasing program sizes (which is much more important for Plutus Core than
+it is for general-purpose programming languages), we should inline only if size(RHS) is
+no bigger than size(LHS).
```
michaelpj, line 415: Do we also want to add an non-conservative mode that relaxes this a bit? I think we probably do. Maybe not here, I guess.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -311,38 +315,118 @@ costIsAcceptable = \case
...
+
+size(LHS) depends on the number of term parameters on the LHS. For `x = f y`, size(LHS) = 1.
+Each additional parameter increases size(LHS) by 2 (one for the `Apply` node, and one for
+the parameter itself). For `x a b c d = f y`, size(LHS) = 9.
+
+We ignore type parameters since types eventually get erased.
```
michaelpj, line 421: This isn't right: they get turned into `Delay`s, which do have size. Why not just say "we don't increase the size of the PIR program"? That's safe and locally correct.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -76,9 +79,10 @@ type Arity = [ParamKind]
 data VarInfo tyname name uni fun ann =
   MkVarInfo {
     varStrictness :: Strictness
-    ,varDef       :: Term tyname name uni fun ann
+    , varDef      :: Term tyname name uni fun ann
     -- ^ its definition that has been unconditionally inlined.
-    , arity       :: Arity -- ^ its arity, storing to avoid repeated calculations.
+    , varArity    :: Arity -- ^ its arity, storing to avoid repeated calculations.
+    , varParams   :: [name]
```
michaelpj, line 85: could consider rolling this into `Arity`: artiy is already a sequence of term/type parameters, including the names in there wouldn't be crazy.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -311,38 +315,118 @@ costIsAcceptable = \case
...
+size(LHS) depends on the number of term parameters on the LHS. For `x = f y`, size(LHS) = 1.
+Each additional parameter increases size(LHS) by 2 (one for the `Apply` node, and one for
+the parameter itself). For `x a b c d = f y`, size(LHS) = 9.
+
+We ignore type parameters since types eventually get erased.
+
+To measure size(RHS), we have to be mindful of multiple occurrences of parameters.
+For example, in `x a b c = b a a`, we cannot consider size(RHS) to be no bigger than size(LHS)
+just because it is not bigger in the binding. Otherwise we will turn `x <large> b c` into
+`let a = <large> in b a a`, which can be worse than not inlining.
```
michaelpj, line 426: Specifically because we won't get rid of the overhead of the binding/application? I had to think quite a lot to work out what this example was saying, worth elaborating.

So in general we _can't_ assume that a saturated application will reduce away entirely unless the arguments look good-to-inline. That actually seems like an attack on our whole argument for doing saturated inlining. So indeed your approach here seems critical. I think the note about saturated inlining needs updating, this is very important!

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -311,38 +315,118 @@ costIsAcceptable = \case
...
+
+We therefore only consider size(RHS) to be acceptable for inlining, if no parameter in the LHS
+occurs more than once in the RHS, or, to put it another way, if every parameter on the LHS can
+be unconditionally inlined after beta-reduction. This is not the optimal strategy but
+it balances simplicity and usefulness.
```
michaelpj, line 431: Okay, so I modify my suggestion from before. We can combine two, simple checks:
- Compare `size (original term)` and `size (body)`. The difference must be below some threshold (later relaxed/configurable)
- Check that the number of occurrences of the bound variables is 1. We already have a usages map for checking this, so no need to do something clever again. This is replicating part of our other inlining heuristics, so it _should_ use the same code.

michaelpj (commented):
Okay, I modified my commentary a bit. I still think we can do this more simply.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
zliu41: > Compare size (original term) and size (body).

Hmm, that's right. Somehow I overlooked this option.

zliu41:
> We can just compute the size of the original term, and the size of the thing we're inlining, and compare them.

@michaelpj after thinking a bit more about this, I don't like this idea, because of the difficulty of handling constants.

We don't want to inline something, if doing so brings in a large constant: e.g., we don't want to turn `f 1 2` into `<large_constant>`.

But we do want to turn `f 1 <large_constant>` into `<large_constant>`.

But then we don't want to turn `f 1 <large_constant>` into `<another_large_constant>`, unless we know `size(<large_constant>) >= size(<another_large_constant>)`.

To do all these correctly we'd need some mechanism to compute and compare the sizes of constants, which is messy. What this PR does seems more sensible.

michaelpj:
> To do all these correctly we'd need some mechanism to compute and compare the sizes of constants, which is messy. What this PR does seems more sensible.

Why not have a size metric that takes into account constant sizes? If that's what we _actually_ care about then I'm quite pro measuring that.

I also think that the approach here is potentially not that good. If we care about constant folding, then we do want to inline constants sometimes. Consider:
```
let 
  x = 1 + 2
  y = 3 +4
  z = x + y
in z
```
We'd like to reduce that all the way, but if we just constant fold we'll get
```
let 
  x = 3
  y = 7
  z = x + y
in z
```
We need to inline those constants to get further. So I think sensible inlining of constants should be a goal. If anything, I think very large constants are likely pretty unusual, so even a size metric that disregards them might be fine.

zliu41:
> Why not have a size metric that takes into account constant sizes?

I assume we measure constant sizes by turning them into Flat and take the length?

In any case we'd need to come up with a sensible size metric, because we would be combining the number of AST nodes and sizes of constants into a single metric, but they are two very different things.

zliu41:
Closed in favor of #5308

PR #5265, opened 2023-04-19
[Evaluation] Statically unroll 'itraverseCounter_'

effectfully:
Results in a statically unrolled loop:

```haskell
    let! { (# ipv10_X1x, ipv11_X1y #) ~ <- $wf_svLt 0# (ipv8_itGb `cast` <Co:8>) } in
    let! { (# ipv12_X1B, ipv13_X1C #) ~ <- $wf_svLt 1# ipv10_X1x } in
    let! { (# ipv14_X1F, ipv15_X1G #) ~ <- $wf_svLt 2# ipv12_X1B } in
    let! { (# ipv16_X1U, ipv17_X1V #) ~ <- $wf_svLt 3# ipv14_X1F } in
    let! { (# ipv18_X1X, ipv19_X1Y #) ~ <- $wf_svLt 4# ipv16_X1U } in
    let! { (# ipv20_X20, ipv21_X21 #) ~ <- $wf_svLt 5# ipv18_X1X } in
    let! { (# ipv22_X23, ipv23_X24 #) ~ <- $wf_svLt 6# ipv20_X20 } in
    let! { (# ipv24_X29, ipv25_X2a #) ~ <- $wf_svLt 7# ipv22_X23 } in
    let! { (# ipv26_X2f, ipv27_X2g #) ~ <- $wf_svLt 8# ipv24_X29 } in
    let! { (# ipv28_X2l, ipv29_X2m #) ~ <- $wf_svLt 9# ipv26_X2f } in
    let! { UnsafeRefl v4_itGA ~ <- unsafeEqualityProof } in
    let! { UnsafeRefl v5_itGz ~ <- unsafeEqualityProof } in
    let! { __DEFAULT ~ wild19_itGD
    <- setByteArray#
         (ipv5_itD3 `cast` <Co:11>)
         0#
         10#
         0#
         (ipv28_X2l `cast` <Co:10>) } in
    let! { UnsafeRefl v6_itGF ~ <- unsafeEqualityProof } in
    (# wild19_itGD, () #) `cast` <Co:15> } in
```

Not sure if it's worth it though, let's ask benchmarks.

### DIFF
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
@@ -1,16 +1,23 @@
-{-# LANGUAGE BangPatterns     #-}
-{-# LANGUAGE DataKinds        #-}
-{-# LANGUAGE KindSignatures   #-}
-{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE AllowAmbiguousTypes      #-}
+{-# LANGUAGE BangPatterns             #-}
+{-# LANGUAGE DataKinds                #-}
+{-# LANGUAGE FlexibleInstances        #-}
+{-# LANGUAGE KindSignatures           #-}
+{-# LANGUAGE MultiParamTypeClasses    #-}
+{-# LANGUAGE StandaloneKindSignatures #-}
+{-# LANGUAGE TypeApplications         #-}
+{-# LANGUAGE TypeFamilies             #-}
+{-# LANGUAGE TypeOperators            #-}
+{-# LANGUAGE UndecidableInstances     #-}
 module UntypedPlutusCore.Evaluation.Machine.Cek.StepCounter where
 
 import Control.Monad.Primitive
 import Data.Coerce (coerce)
+import Data.Kind
 import Data.Primitive qualified as P
 import Data.Proxy
 import Data.Word
-import GHC.TypeLits (Nat)
-import GHC.TypeNats (KnownNat, natVal)
+import GHC.TypeNats (KnownNat, Nat, natVal, type (-))
 
 -- See Note [Step counter data structure]
 -- You might think that since we can store whatever we like in here we might as well
@@ -73,18 +80,39 @@ modifyCounter i f c = do
   pure modified
 {-# INLINE modifyCounter #-}
 
+-- | The type of natural numbers in Peano form.
+data Peano
+  = Z
+  | S Peano
+
+type NatToPeano :: Nat -> Peano
+type family NatToPeano n where
+    NatToPeano 0 = 'Z
+    NatToPeano n = 'S (NatToPeano (n - 1))
+
+type UpwardsM :: (Type -> Type) -> Peano -> Constraint
+class Applicative f => UpwardsM f n where
+  -- | @upwardsM i k@ means @k i *> k (i + 1) *> ... *> k (i + n - 1)@.
+  -- We use this function in order to statically unroll a loop in 'itraverseCounter_' through
+  -- instance resolution. This makes @validation@ benchmarks a couple of percent faster.
+  upwardsM :: Int -> (Int -> f ()) -> f ()
+
+instance Applicative f => UpwardsM f 'Z where
+  upwardsM _ _ = pure ()
+  {-# INLINE upwardsM #-}
+
+instance UpwardsM f n => UpwardsM f ('S n) where
+  upwardsM !i k = k i *> upwardsM @f @n (i + 1) k
+  {-# INLINE upwardsM #-}
+
 -- | Traverse the counters with an effectful function.
 itraverseCounter_
   :: forall n m
-  . (KnownNat n, PrimMonad m)
+  . (UpwardsM m (NatToPeano n), PrimMonad m)
   => (Int -> Word8 -> m ())
   -> StepCounter n (PrimState m)
   -> m ()
 itraverseCounter_ f (StepCounter arr) = do
-  -- The implementation of this function is very performance-sensitive. I believe
-  -- it may be possible to do better, but it's time-consuming to experiment more
-  -- and unclear what improves things.
-
   -- The safety of this operation is a little subtle. The frozen array is only
   -- safe to use if the underlying mutable array is not mutated 'afterwards'.
   -- In our case it likely _will_ be mutated afterwards... but not until we
@@ -92,22 +120,12 @@ itraverseCounter_ f (StepCounter arr) = do
   -- the whole thing runs in 'm': future accesses to the mutable array can't
   -- happen until this whole function is finished.
   arr' <- P.unsafeFreezePrimArray arr
-  let
-    sz = fromIntegral $ natVal (Proxy @n)
-    go !i
-      | i < sz = do
-          f i (P.indexPrimArray arr' i)
-          go (i+1)
-      | otherwise = pure ()
-  go 0
--- I also tried INLINABLE + SPECIALIZE, which just resulted in it getting inlined, so might
--- as well just go straight for that
+  upwardsM @_ @(NatToPeano n) 0 $ \i -> f i $ P.indexPrimArray arr' i
 {-# INLINE itraverseCounter_ #-}
-
 
 -- | Traverse the counters with an effectful function.
 iforCounter_
-  :: (KnownNat n, PrimMonad m)
+  :: (UpwardsM m (NatToPeano n), PrimMonad m)
   => StepCounter n (PrimState m)
   -> (Int -> Word8 -> m ())
   -> m ()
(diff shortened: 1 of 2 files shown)


effectfully:
The first run is -1.13% on average, then second one is -1.12%.

Looks like a modest improvement, but really hard to tell with the benchmarks not being reliable for such little things at all.

effectfully:
^ +0.95% on average out of nowhere. Let's try again.

effectfully:
+0.31% on average.

OK, this looks like noise that isn't worth spending time on.

effectfully:
-2.2% to -3.2% with a bunch of readings around -2.5%, so let's call it -2.5%.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
```diff
@@ -73,41 +80,52 @@ modifyCounter i f c = do
...
+  -- We use this function in order to statically unroll a loop in 'itraverseCounter_' through
+  -- instance resolution. This makes @validation@ benchmarks a couple of percent faster.
+  upwardsM :: Int -> (Int -> f ()) -> f ()
+
+instance Applicative f => UpwardsM f 'Z where
+  upwardsM _ _ = pure ()
+  {-# INLINE upwardsM #-}
+
+instance UpwardsM f n => UpwardsM f ('S n) where
+  upwardsM !i k = k i *> upwardsM @f @n (i + 1) k
```
bezirg, line 105: Weren't you saying somewhere that we should always use  `(const+i)` instead of `(i+const)`? Not that it changes anything here...

bezirg, line 105: It feels a bit weird when thinking about the "types" going downwards while in the meantime the values going upwards...

Can't you use them going on the same direction, for example by calling `(i-1)` instead? Probably you would have to change to `(<*)` also.

On another note, assuming that NatToPeano is necessary, can't you still keep the KnownNat around so that in the method's definition you call to `natVal (Proxy @n)` instead of doing the arithmetic manually?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
```diff
@@ -73,41 +80,52 @@ modifyCounter i f c = do
   pure modified
 {-# INLINE modifyCounter #-}
 
+-- | The type of natural numbers in Peano form.
+data Peano
+  = Z
+  | S Peano
```
bezirg, line 86: Can you explain why this is needed? Is it because KnownNat + CmpNat  type family would require OverlappingInstances?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
```diff
@@ -73,41 +80,52 @@ modifyCounter i f c = do
...
+  upwardsM :: Int -> (Int -> f ()) -> f ()
+
+instance Applicative f => UpwardsM f 'Z where
+  upwardsM _ _ = pure ()
+  {-# INLINE upwardsM #-}
+
+instance UpwardsM f n => UpwardsM f ('S n) where
+  upwardsM !i k = k i *> upwardsM @f @n (i + 1) k
+  {-# INLINE upwardsM #-}
+
```
bezirg, line 107: How reliable is the use of INLINE pragma?

Let's say that we cannot guarantee that this will be unrolled (fully);
have you looked into the perf difference of old way vs the not-inlined method way? Could the not-inlined method way be actually much slower than the old way?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
```diff
@@ -73,41 +80,52 @@ modifyCounter i f c = do
...
 -- | Traverse the counters with an effectful function.
 itraverseCounter_
   :: forall n m
-  . (KnownNat n, PrimMonad m)
+  . (UpwardsM m (NatToPeano n), PrimMonad m)
   => (Int -> Word8 -> m ())
   -> StepCounter n (PrimState m)
   -> m ()
 itraverseCounter_ f (StepCounter arr) = do
```
bezirg, line 115: Seems that no other code is using itraverseCounter. How about we get rid of itraversecounter and switch to iforcounter directly?

bezirg (approved):

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
bezirg: Feel safe to ignore

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
bezirg: Maybe the rationale is `constant folding`?

effectfully: > Weren't you saying somewhere that we should always use `(const+i)` instead of `(i+const)`?

That's in UPLC, I don't think it matters in Haskell, but I'll be happy to get enlightened.

effectfully: > It feels a bit weird when thinking about the "types" going downwards while in the meantime the values going upwards...

I did exactly what the original implementation did, that's it. I don't think it matters much, but we can discuss it during tech discussion if anybody hates this version.

> Probably you would have to change to `(<*)` also.

That wouldn't change ordering of effects or the result.

> On another note, assuming that NatToPeano is necessary, can't you still keep the KnownNat around so that in the method's definition you call to `natVal (Proxy @n)` instead of doing the arithmetic manually?

Perhaps, but evidently constant folding works perfectly well here (see the GHC Core in the PR description), so who cares?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
```diff
@@ -73,41 +80,52 @@ modifyCounter i f c = do
...
+
+instance Applicative f => UpwardsM f 'Z where
+  upwardsM _ _ = pure ()
+  {-# INLINE upwardsM #-}
+
+instance UpwardsM f n => UpwardsM f ('S n) where
+  upwardsM !i k = k i *> upwardsM @f @n (i + 1) k
+  {-# INLINE upwardsM #-}
+
 -- | Traverse the counters with an effectful function.
```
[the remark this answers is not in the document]
effectfully: Thanks a lot for double-checking!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
effectfully: > How reliable is the use of INLINE pragma?

In my experience, for such small definitions it is reliable.

> have you looked into the perf difference of old way vs the not-inlined method way? Could the not-inlined method way be actually much slower than the old way?

 I haven't but I'd expect it to be noticeably slower indeed.

I don't think we need to worry about it, there's a huge tower of abstractions that we optimize away like this. The code here is very straightforward compared to some of the other stuff that we do. There's no way around looking into GHC Core occasionally and relying on longitudinal benchmarks anyway.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
effectfully: > Is it because KnownNat + CmpNat type family would require OverlappingInstances?

Yes. Or interleaving a type family and a type class (like I do [here](https://github.com/effectfully-ou/sketches/tree/master/avoid-overlapping-recursive)). Or maybe something else. The version in the PR is the simplest I'm aware of.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/StepCounter.hs
effectfully: Meh, it's normal for `for` to get defined in terms of `traverse` and having an extra function is fine. I'd keep it the way it is.

effectfully:
I'm going to claim that I've addressed the comments (by responding to them), so let's merge this bad boy.

PR #5266, opened 2023-04-20
[Evaluation] Use 'ArgsR' in the CEK machine

effectfully:
This should be faster, but let's see what benchmarks say.


effectfully:
Here's a comparison to the PR (the left `Change` column) that this one (the right `Change` column) builds upon: 

<details>
<summary>Results table</summary>

| Script               |   Change   |   Change   |
| :------              |  :------:  |  :------:  |
| clausify/formula1    |   -33.8%   |   -34.6%   |
| clausify/formula2    |   -33.1%   |   -33.9%   |
| clausify/formula3    |   -33.0%   |   -34.1%   |
| clausify/formula4    |   -29.7%   |   -30.4%   |
| clausify/formula5    |   -34.8%   |   -35.4%   |
| knights/4x4          |   -14.8%   |   -16.9%   |
| knights/6x6          |   -14.6%   |   -15.5%   |
| knights/8x8          |   -14.3%   |   -15.6%   |
| primetest/05digits   |    +0.3%   |    -3.8%   |
| primetest/08digits   |    +1.6%   |    -2.8%   |
| primetest/10digits   |    +1.8%   |    -3.2%   |
| primetest/20digits   |    +2.6%   |    -2.4%   |
| primetest/30digits   |    +2.5%   |    -2.5%   |
| primetest/40digits   |    +2.8%   |    -2.1%   |
| primetest/50digits   |    +2.9%   |    -1.4%   |
| queens4x4/bt         |   -14.1%   |   -16.5%   |
| queens4x4/bm         |   -13.6%   |   -17.4%   |
| queens4x4/bjbt1      |   -13.9%   |   -16.2%   |
| queens4x4/bjbt2      |   -13.3%   |   -15.5%   |
| queens4x4/fc         |   -13.1%   |   -14.5%   |
| queens5x5/bt         |   -13.5%   |   -16.1%   |
| queens5x5/bm         |   -14.4%   |   -18.9%   |
| queens5x5/bjbt1      |   -13.2%   |   -17.3%   |
| queens5x5/bjbt2      |   -13.2%   |   -16.6%   |
| queens5x5/fc         |   -13.4%   |   -15.0%   |
</details>

Everything's faster and slowdowns become speedups!

michaelpj:
Numbers look pretty good, I'll take a look at the code.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -556,12 +558,12 @@ data Context uni fun ann
     -- ^ @[V _]@
     | FrameApplyArg !(CekValEnv uni fun ann) !(NTerm uni fun ann) !(Context uni fun ann)
     -- ^ @[_ N]@
-    | FrameApplyValues ![CekValue uni fun ann] !(Context uni fun ann)
-    -- ^ @[_ V0...Vn]@
+    | FrameApplyArgVal !(CekValue uni fun ann) !(Context uni fun ann)
```
michaelpj, line 561: I think we should benchmark this change separately. I _think_ I tried this before and it was worse. A priori it adds another traversal of the arguments that we don't need.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -648,6 +650,20 @@ evalBuiltinApp fun term runtime = case runtime of
     _ -> pure $ VBuiltin fun term runtime
 {-# INLINE evalBuiltinApp #-}
 
+data ArgsR uni fun ann
```
michaelpj, line 653: So where is the speedup coming from? We can't unpack `CekValue` or `ArgsR`, so the only thing I can see is that it's spine-strict. Is that what you're expecting makes the difference?

If so, is there an advantage to having this fully-specialized one, rather than just defining a standalone spine-strict list type?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -648,6 +650,20 @@ evalBuiltinApp fun term runtime = case runtime of
...
+    = NilArgsR
+    | ConsArgsR !(CekValue uni fun ann) !(ArgsR uni fun ann)
+    deriving stock (Show)
+
+toListArgsR :: ArgsR uni fun ann -> [CekValue uni fun ann]
```
michaelpj, line 658: This isn't on the critical path, so I'd probably just have `ArgsR` be a normal strict list that we remember is in reverse order and stick a `reverse` in the places we materialize it back into a normal-order list. Unsure though.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
michaelpj: Sorry, it's fully strict, not just spine-strict.

michaelpj: And indeed, there are many strict lists defined elsewhere we could take. Unsure if it's worth it.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -648,6 +650,20 @@ evalBuiltinApp fun term runtime = case runtime of
...
+toListArgsR = go [] where
+    go acc NilArgsR             = acc
+    go acc (ConsArgsR arg args) = go (arg : acc) args
+
+appendArgsR :: ArgsR uni fun ann -> Context uni fun ann -> Context uni fun ann
```
michaelpj, line 663: I found the name of this a little confusing. Not sure what else to call it.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: > A priori it adds another traversal of the arguments that we don't need.

It's on par with using `DList`s, see [this](https://www.joachim-breitner.de/blog/620-Constructing_a_list_in_a_Monad) and [this](https://www.joachim-breitner.de/blog/684-Constructing_a_list_in_a_monad_revisited). Those `DList`s don't come for free!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: > And indeed, there are many strict lists defined elsewhere we could take. 

I actually couldn't find any. There's `strict-list`, but I'm skeptical about it's alignment (it claims reversing things all the time is faster, but I failed to find any evidence for that) and it doesn't have enough `INLINE` pragmas.

> We can't unpack `CekValue` or `ArgsR`, so the only thing I can see is that it's spine-strict. 

I think Kmett told us that monomorphizing data definitions is beneficial even when nothing can be unpacked? I don't actually know and would like to learn. Defining `ArgsR` manually is extremely cheap though, just a few lines of very local code and it has the advantage of clearly conveying the fact that the args are in reverse order.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: Well, I did manage to resist the temptation to call it [`chips`](https://github.com/pigworker/SSGEP-DataData/blob/26bc46552f188765d3e98a49270027390155e46c/STLC.lagda#L265-L273), don't make me question my hard-to-make decisions.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: I should've defined it as

```haskell
data ArgsR uni fun ann
    = NilArgsR
    | SnocArgsR !(ArgsR uni fun ann) !(CekValue uni fun ann)
```

BTW.

michaelpj: He did? I don't remember that... but you might be right.

> Defining ArgsR manually is extremely cheap though, just a few lines of very local code and it has the advantage of clearly conveying the fact that the args are in reverse order.

IDK, having custom datatypes makes the reader load more concepts into their brains. It feels easier for me to just see "this is a strict list". Not sure.

effectfully: > He did? I don't remember that...

Or I made it up, which is very possible. I wonder where I can learn about this.

> IDK, having custom datatypes makes the reader load more concepts into their brains. It feels easier for me to just see "this is a strict list". Not sure.

I don't have any strong feelings about that anyway.

michaelpj:
I'll look at this after the other PR is mergd.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -743,19 +759,17 @@ enterComputeCek = computeCek
     returnCek (FrameApplyFun fun ctx) arg =
         applyEvaluate ctx fun arg
     -- s , [_ V1 .. Vn] ◅ lam x (M,ρ)  ↦  s , [_ V2 .. Vn]; ρ [ x  ↦  V1 ] ▻ M
-    returnCek (FrameApplyValues args ctx) fun = case args of
-        (arg:rest) -> applyEvaluate (FrameApplyValues rest ctx) fun arg
```
effectfully, line 734: Note that you kinda process `args` twice here yourself. One matching per argument to get to the frame + one matching to get to the actual argument (at which point you reconstruct the frame).

michaelpj: That's a good point: we still allocate N `FrameApplyValues` constructors overall!

effectfully:
@michaelpj given #5288, should we close this one?

michaelpj:
Yeah sure.

PR #5269, opened 2023-04-24
PLT-5288: Add a generic builtin evaluation pass

michaelpj:
Turned out to be pretty straightforward. The only trickiness was around logging: if just go naively we elminate everything that looks like `trace <constant string> x`, which is most calls! So I added a flag to be conservative about this and linked it to the existing conservative optimization flag.

I also pulled out the argument-splitting bit from the call-site inliner into a "application context" module, which seems to match it better IMO.

Pretty modest improvements though.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-tx-plugin/test/Budget/show.plc.golden
```diff
@@ -227,6 +227,7 @@ let
                     {all dead. dead})
                {all dead. dead})
           {all dead. dead}
+  !w : bytestring = #5468697320697320616e206578616d706c65
```
michaelpj, line 230: this is kind of cool though

plutus-core/plutus-ir/src/PlutusIR/Compiler/Types.hs
```diff
@@ -49,19 +49,21 @@ instance PLC.HasTypeCheckConfig (PirTCConfig uni fun) uni fun where
...
-    , _coDoSimplifierBeta         :: Bool
-    , _coDoSimplifierInline       :: Bool
-    , _coInlineHints              :: InlineHints PLC.Name (Provenance a)
-    , _coProfile                  :: Bool
-    , _coRelaxedFloatin           :: Bool
+    , _coDoSimplifierUnwrapCancel     :: Bool
+    , _coDoSimplifierCaseReduce       :: Bool
+    , _coDoSimplifierBeta             :: Bool
+    , _coDoSimplifierInline           :: Bool
+    , _coDoSimplifierEvaluateBuiltins :: Bool
```
zliu41, line 62: If the only problematic one is `trace`, can we instead have a flag that controls the evaluation of only `trace`?

In fact I would like a more general flag that controls whether the passes are allowed to change logging. Because CSE and other passes may also change logging.

We can also consider adding optimization levels, where changes to logging is only allowed with `-O2` or above.

michaelpj: `trace` isn't the only one. All builtins _can_ emit log messages. And we can't know which ones the users care about really.

> In fact I would like a more general flag that controls whether the passes are allowed to change logging. 

That's what this flag was supposed to be. I'm not sure why I gave it the unnecessarily specific name. It _does_ currently control whether the pass will evaluate _any_ builtin with logging!

> We can also consider adding optimization levels, where changes to logging is only allowed with -O2 or above.

I think the current segmentation makes more sense for us. For normal compilers the tradeoff with higher optimization levels is usually "better code but slower compilation". We don't really care about slower compilation at the moment, so we'd always take the max level if we had them, so it's not so useful. But "does this mess with the effects" _does_ seem like something people will care about.

michaelpj: I renamed the option to "preserveLogging`, does that seem more okay?

plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
```diff
@@ -0,0 +1,72 @@
...
+      :: BuiltinRuntime (Term tyname name uni fun ())
+      -> AppContext tyname name uni fun a
+      -> Maybe (Term tyname name uni fun ())
+    eval (BuiltinResult _ getX) AppContextEnd =
+        case getX of
+            MakeKnownSuccess v           -> Just v
+            -- Evaluates successfully, but does logging. If we're being conservative
+            -- then we should leave these in, so we don't remove people's logging!
+            -- Otherwise `trace "hello" x` is a prime candidate for evaluation!
+            MakeKnownSuccessWithLogs _ v -> if conservative then Nothing else Just v
```
effectfully, line 43: You could take the result and annotate it with a `trace` call replicating the logging behavior of the builtin application.

plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
```diff
@@ -0,0 +1,72 @@
...
+    processTerm :: Term tyname name uni fun a -> Term tyname name uni fun a
+    -- See Note [Context splitting in a recursive pass]
+    processTerm t@(splitApplication -> (Builtin x bn, argCtx)) =
+      let runtime = toBuiltinRuntime costModel (toBuiltinMeaning ver bn)
+      in case eval runtime argCtx of
+           -- Builtin evaluation gives us a fresh term with no annotation.
+           -- Use the annotation of the builtin node, arbitrarily. This is slightly
+           -- suboptimal, e.g. in `ifThenElse True x y`, we will get back `x`, but
+           -- with the annotation that was on the `ifThenElse` node. But we can't
+           -- easily do better.
```
effectfully, line 69: Annoyingly, it's where we could actually return `x` with its original annotation, the builtins machinery does support that. What it doesn't support is turning the result of a builtin application when it's a constant into an annotated term, because where would we get an annotation from in this case? Compute it from the result somehow? I guess we could take the annotation of the builtin like you do, but it's kinda weird to hardcode that right into the evaluator. I think I could do it though if you want, for `()` it would all be the same, so on the production path we wouldn't care anyway.

Create a ticket for me, if you want me to look into it and I'll do it once I'm back from vacation the next week.

plutus-tx-plugin/test/Budget/ifThenElse1.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 796063
-| mem: 2302})
```
thealmarty, line 2: 👏🏾

effectfully (approved):
Very nice indeed!

plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
michaelpj: That's... kind of brilliant. We would then re-evaluate those trace calls every time, which would be kind of annoying.

Also, actually this pass doesn't know about `trace`, it just requires you to be able to evaluate builtins, so that would be a bit less general.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Types.hs
zliu41: >  For normal compilers the tradeoff with higher optimization levels is usually "better code but slower compilation". 

It's a separate discussion, orthogonal to this PR, but high optimization levels can also mean "risky optimizations", e.g., gcc's loop unrolling which may increase program size.

For us, we can use "risky optimization" imply both "may change logging" (evaluating builtins, CSE) and "may increase cost/size" (float-in, float-out). Having a higher optimization level would be a convenient way to say "just try your best to optimize!"

plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
michaelpj: I'm kind of confused then. I though the types implied that I get a term with unit annotation no matter what I do :thinking: Maybe I'm doing it wrong.

I also think the current behaviour is pretty reasonable and I'm not bothered by it.

plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
```diff
@@ -0,0 +1,46 @@
...
+fillAppContext t = \case
+  AppContextEnd                -> t
+  TermAppContext arg ann ctx   -> fillAppContext (Apply ann t arg) ctx
+  TypeAppContext tyArg ann ctx -> fillAppContext (TyInst ann t tyArg) ctx
+
+{- Note [Context splitting in a recursive pass]
+When writing a recursive pass that processes the whole program, you must be
+a bit cautious when using a context split. The context split may traverse
+part of the program, which will _also_ be traversed by the main recursive
+traversal. This can lead to "accidentally quadratic" behaviour.
```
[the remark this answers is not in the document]
michaelpj: "Accidentally quadratic" is a [meme](https://accidentallyquadratic.tumblr.com/), but perhaps that's me being too cute.

> but for these passes we don't care about that right?

Right, but I wanted to have a note that I can refer to about the general issue, so we have to think about whether or not it's okay (and when it is we can use the note as a justification).

plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
```diff
@@ -0,0 +1,46 @@
...
+import PlutusIR.Core
+
+-- | A context for an iterated term/type application, with the hole at the head of the
+-- application.
+data AppContext tyname name uni fun ann =
+  TermAppContext (Term tyname name uni fun ann) ann (AppContext tyname name uni fun ann)
+  | TypeAppContext (Type tyname uni ann) ann (AppContext tyname name uni fun ann)
+  | AppContextEnd
+
+-- | Takes a term and views it as a head plus an 'AppContext'.
```
thealmarty, line 17: Maybe add an example?
```suggestion
-{  Takes a term and views it as a head plus an 'AppContext'. E.g.,
`f (g x)`, which is 
`Apply ann f (Apply ann1 g x)` becomes (x, application context):
`(x, TermAppContext x ann1 (TermAppContext (Apply ann1 g x) ann AppContextEnd))`
```

plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
effectfully: > We would then re-evaluate those trace calls every time, which would be kind of annoying.

Sorry, didn't get it.

> Also, actually this pass doesn't know about `trace`, it just requires you to be able to evaluate builtins, so that would be a bit less general.

Right, that's unfortunate.

plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
michaelpj: I see, so we could in principle refactor it to, say, take an annotation to use for fresh terms (constants), and to otherwise preserve annotations on terms?

plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
michaelpj: I mean, the optimization will trigger on them every time:
```
trace "hello" x
--> (great, evaluate it) 
x
--> (oops, need to put a trace call in to replicate the logging)
trace "hello" x
--> (great evaluate it)
...
```
etc.

plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
effectfully: Yep, exactly.

plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
effectfully: Ah, I see. Yeah, if this function knew about `trace`, it could just not touch it. Or you could insert a special annotation and postprocess the term after the optimization is complete. Oh, wait, you can do that right now. If the builtin application emitted a log, you just add it to the annotation of the result and then once you have `trace` available somewhere down the pipeline, you turn it into the actual call to `trace`. Rather stupid design, though.

plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
```diff
@@ -0,0 +1,46 @@
...
+  -> (Term tyname name uni fun ann, AppContext tyname name uni fun ann)
+splitApplication tm
+  = go tm AppContextEnd
+  where
+    go (Apply ann f arg) ctx    = go f (TermAppContext arg ann ctx)
+    go (TyInst ann f tyArg) ctx = go f (TypeAppContext tyArg ann ctx)
+    go t ctx                    = (t, ctx)
+
+-- | Fills in the hole in an 'AppContext'.
+fillAppContext
```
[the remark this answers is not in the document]
michaelpj: I do think this is an example of something that's simple enough that it's not worth an independent test.

michaelpj:
Thanks to merging COKC, the `ifThenElse` examples now reduce away almost entirely, which is cool. But still not that impactful.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Types.hs
```diff
@@ -79,26 +82,30 @@ defaultCompilationOpts = CompilationOpts
   , _coDoSimplifierKnownCon = True
   , _coDoSimplifierBeta = True
   , _coDoSimplifierInline = True
+  , _coDoSimplifierEvaluateBuiltins = True
   , _coInlineHints = mempty
   , _coProfile = False
   , _coRelaxedFloatin = True
+  , _coPreserveLogging = False
```
bezirg, line 89: Isn't it a bit extreme to be False by default?

plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
```diff
@@ -0,0 +1,70 @@
+{-# LANGUAGE LambdaCase #-}
+-- | Datatypes representing 'contexts with holes' in Plutus IR terms.
+--
```
bezirg, line 3: Smells like a Zipper!

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -120,34 +121,32 @@ computeArity = \case
...
-      -- inlining, and it won't make it any worse, so we could consider accepting this.
-      False
+isFullyApplied (_lam:_lams) AppContextEnd = False -- under-application
+isFullyApplied (TermParam:tlLams) (TermAppContext _ _ ctx) = isFullyApplied tlLams ctx
+isFullyApplied (TypeParam:tlLams) (TypeAppContext _ _ ctx) = isFullyApplied tlLams ctx
+isFullyApplied _ _ =
+    -- wrong argument type, i.e., we have an ill-typed term here. It's not what we define as fully
+    -- applied. Although if the term was ill-typed before, it will be ill-typed after the
+    -- inlining, and it won't make it any worse, so we could consider accepting this.
+    False
```
bezirg, line 133: Can't you just do instead `arity == lengthContext`?

bezirg (approved):

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
michaelpj: Yes, but this is a bit more explicit about the weird case. We discussed this a bit before. Maybe it would be simpler and we should just use that.

plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
michaelpj: Same idea, yes. I think ultimately we may just want a full one-hole-context type that handles the whole term structure, but that's quite complex.

PR #4191, opened 2021-11-05
Short circuit ops for booleans (over QQ)

anton-k:
Fixes #4114

The problem is that plutus booleans don't use short circuit on evaluation.
There is ugly fix to rewrite all expressions with `if-then-else`. 
But code becomes rather clumsy. We can fix that with meta TH.

Proposed fix defines two qq-functions `and_if` and `or_if`:

```haskell
[and_if| expr1, expr2, expr3]
```

for arbitrary number of arguments is expanded to 

```haskell
if expr1
  then if expr2
    then if expr3
      then True
      else False
    else False
  else False 
```

and 

```haskell
[or_if| expr1, expr2, expr3]
```

is expanded to

```haskell
if expr1
  then True
  else if expr2
    then True
    else if expr3
      then True
      else False
```

PR #5270, opened 2023-04-26
PLT-4267: ADR: Plutus Tx's Evaluation Strategy, Compiler and Standard Library

zliu41:


doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+
+```haskell
+let x = nonstrict rhs in body
+```
+
+Because of the NOINLINE pragma, `nonstrict` is guaranteed to be visible to the Plutus Tx compiler.
+Whenever the Plutus Tx compiler sees `let x = nonstrict rhs`, it first compiles `rhs` normally; then it creates a nonstrict binding in PIR: `let ~x = rhs_compiled`.
+
+What if `x` gets inlined by GHC, i.e., GHC turns `let x = nonstrict rhs in ...x...` into `...(nonstrict rhs)...`?
+That's not a problem, because if we see a `nonstrict rhs` that isn't the RHS of a binding, we create a new, non-strict binding from it, i.e., it becomes, in PIR,
```
michaelpj, line 80: It's a shame we can't just use this strategy unconditionally, i.e. it would be nice if we could turn
```
let x = nonstrict rhs in body
```
into
```
let ~fresh = rhs
in let !x = fresh
in body
```
but that doesn't actually work. Anyway, we can discuss the specific details elsewhere. In fact, I think probably this ADR shouldn't cover them: the _decision_ is to handle this through explicit signalling that indicates a non-strict binding, and how we implement that isn't so relevant.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+    go = \case
+      []     -> z
+      x : xs -> f x (go xs)
+```
+
+Since Plutus Tx is strict by default, this `foldr` definition cannot short-circuit, because `go xs` will be evaluated before evaluating `f`.
+It is not sufficient to create a non-strict binding for `go xs` (i.e., `let y = go xs in f x y`) because GHC will unconditionally inline `y`.
+
+The problem can be solved by using `nonstrict`, replacing `f x (go xs)` with `f x (nonstrict (go xs))`.
+Now `go xs` is evaluated non-strictly, and `foldr` can short-circuit.
```
michaelpj, line 105: I think we discussed this but this is not true! It's no good because your non-strict variable then gets used _as a function argument to f_, which means it gets evaluated. You have to _also_ inline `f`, and to allow that you have to first inline `foldr` too.

Any time we have a higher-order function, the use of that function _will_ be strict and we can't easily make it lazy.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+For that reason, it may be a good idea to provide both a strict version and a non-strict version of `foldr`.
+
+Some strict languages have special syntaxes to mark a function lazy in a particular argument, e.g., Scala has the `=>` syntax.
+The following function is strict in the first argument and lazy in the second argument:
+
+```scala
+def foo(a: Int, b: => Int) { ... }
+```
+
+`nonstrict` can be used to achieve a similar effect, except that `nonstrict` is used at the call sites as opposed to function definitions.
```
michaelpj, line 116: Again, I think this is wrong.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+Interestingly, the `Foldable.forall` and `Foldable.exists` functions in cats are implemented by [converting the foldable structure to list first](https://github.com/typelevel/cats/blob/896f00cd689af20676556d1b57791d7ade5cfa04/core/src/main/scala/cats/Foldable.scala#L677-L686), rather than using `foldRight`.
+
+We can potentially introduce something similar to `Eval` in the Plutus Tx standard library.
+The advantage of `Eval` is that it can be provided entirely in a library, without special compiler support. However, this is also what leads to its disadvantage: `Eval` is more heavyweight to use, since it changes the types of many functions (such as `foldr`), while `nonstrict` has the same type as the identity function, which is possible because it is a magical function that the compiler treats specially.
+
+The takeaway is this: the standard libraries usually do the simple things, e.g., providing a monomorphic, non-short-circuiting version of `foldr` for lists, as well as functions like `all`, `any`, also monomorphically for lists.
+The polymorphic versions of those functions, if exist, are usually provided by third-party libraries like cats.
```
michaelpj, line 134: I think Scala isn't a good example here: the standard library is sort of half-heartedly functional, one of the distinguishing features of `cats` is to try and be much more serious about Haskell-inspired functional programming.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+
+For Plutus Tx, even with `nonstrict`, it is still useful to provide monomorphic versions of certain functions in the standard library, especially since they can be noticeably cheaper than the polymorphic versions.
+This will be discussed in the next section.
+
+### Monomorphizing Plutus Tx Standard Library
+
+Strict languages do not lend themselves to code reuse, as there is a high cost to use abstractions such as higher-order functions and parametric polymorphism.
```
michaelpj, line 141: Why is parametric polymorphism expensive? Do you mean ad-hoc polymorphism (i.e. typeclasses)? If so that's not really about strict languages: GHC would have exactly the same overheads as we do (and often does!), but also goes to some lengths to eliminate dictionary passing.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+For Plutus Tx, there's another reason why monomorphization of the standard library is beneficial: multiple method typeclasses are expensive.
+
+Single method typeclasses are cheap because they have simpler representations in GHC Core - a newtype that can be compiled away.
+On the other hand, multiple method typeclasses exist in GHC Core as actual dictionaries and need to be compiled into product types in PIR.
+
+This makes monomorphization the only choice in many cases. Consider `foldr` and `foldMap`.
+They can be implemented in terms of each other, so technically, only one of them needs to be in the `Foldable` class.
+But certain types have more efficient direct implementations of `foldr`; the same for `foldMap`.
+No matter which of them we move out of `Foldable`, it won't work optimally for certain types.
+Keeping both in `Foldable` will cause `Foldable` to be a multiple method typeclass, which is expensive in and of itself, as explained above.
```
michaelpj, line 153: I'm going to be annoying and say: we should actually do budget benchmarks of the two versions and see. It's not obvious to me how much difference it makes!

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+
+In Plutus Tx, like most other strict languages, only a few built-in constructs are non-strict (such as if-then-else).
+It is cumbersome, if not impossible, to create user-defined non-strict functions.
+The `nonstrict` function described before does not help because it cannot be used in defining a function.
+It would be useful to allow user-defined non-strict functions, so that we can define functions like `&&` and `||` that short-circuit properly.
+Some languages like Scala have special syntaxes for defining non-strict functions; we can't change Haskell's syntax, so we need to find a different solution.
+
+The idea is to make the Plutus Tx compiler perform GHC-like inlining for functions with the INLINE pragma.
+
+"GHC-like inlining", or "inlining in the GHC way", for lack of a better term, refers to inlining in a way similar to how GHC does it, i.e., replacing a variable with its RHS, then performing beta substitution.
```
michaelpj, line 177: I'm confused by this. I don't think GHC's inliner actually does the substitution? I think the simplifier will do it, but I find it confusing to associate this with GHC's inlining somehow?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+whereas inlining `(&&)` in the GHC way yields
+
+```haskell
+if expr1 == False then False else expr2
+```
+
+Normally, it is incorrect for the Plutus Tx compiler to inline `(&&)` in the GHC way, because in the `expr1 && expr2` example, before inlining, it evaluates `expr2` strictly, but after GHC-like inlining, `expr2` is evaluated non-strictly.
+But because of the `INLINE` pragma, the Plutus Tx compiler should inline `(&&)` in the GHC way.
+And this is desirable because we want `(&&)` to be non-strict.
+Functions that should not be inlined in the GHC way should use the `INLINABLE` pragma.
```
michaelpj, line 203: Why shouldn't INLINE mean exactly what it means today? i.e. I'm not sure why this alternative meaning of inlining should get to claim the pragma, indeed I think that's surprising. We could use an alternative GHC annotation, perhaps?

michaelpj (commented):
I'm broadly on board. As a process note, I feel like this ADR has too many details (which probably belong in Notes in the code), rather than focussing on the decision at hand.

I'd also like to see some discussion of backwards compatibility. At least for switching to strict let-bindings by default. We'll probably need a plugin option to support people who want the old behaviour.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+The takeaway is this: the standard libraries usually do the simple things, e.g., providing a monomorphic, non-short-circuiting version of `foldr` for lists, as well as functions like `all`, `any`, also monomorphically for lists.
+The polymorphic versions of those functions, if exist, are usually provided by third-party libraries like cats.
+
+For Plutus Tx, even with `nonstrict`, it is still useful to provide monomorphic versions of certain functions in the standard library, especially since they can be noticeably cheaper than the polymorphic versions.
+This will be discussed in the next section.
+
+### Monomorphizing Plutus Tx Standard Library
+
+Strict languages do not lend themselves to code reuse, as there is a high cost to use abstractions such as higher-order functions and parametric polymorphism.
+Polymorhpic functions taking `Foldable` or similar constraints, implemented in terms of higher-order functions like `foldr`, are rare in strict languages.
```
bezirg, line 142: typo `Polymorhpic`

bezirg (approved):
I like this proposal a lot.  We just have to make sure that all these great information and examples are properly catalogued in readthedocs. 

One thing: since this is a fairly big semantic change, what happens to existing written PlutusTx scripts? Are we going to direct the plutustx users to use an older plutustx compiler version to retain the old behaviour or are we going to expose some extra plutus-tx-plugin flags to "drive" this semantic change?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+What is unclear is whether Plutus Tx is strict or non-strict.
+Indeed, it is sometimes strict and sometimes non-strict, and it is fairly random and difficult to predict:
+
+- Let-bindings in Plutus Tx are by default non-strict, but function applications are strict.
+  That is, `let x = rhs in body` isn't equivalent to `(\x -> body) rhs`.
+  This is distinct from most other languages, regardless of their evaluation strategies, in which these two expressions are equivalent.
+
+- The conditional expression `if b then t else f` is always non-strict in `t` and `f`, like in most strict languages.
+  But what about `foo b t f = if b then t else f`? In a strict language, `foo` would be strict, and in a non-strict language, it would be non-strict.
+  But in Plutus Tx, due to how its compiler is currently implemented, whether `t` and `f` are evaluated strictly or not depends on whether GHC inlines `foo`.
```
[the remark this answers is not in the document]
bezirg: @michaelpj you mean when forcing the inlining in a strict language?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
bezirg: Oh yeah, that's right, it contradicts the gdrive

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
bezirg: @michaelpj  you mean this https://downloads.haskell.org/ghc/latest/docs/users_guide/extending_ghc.html?highlight=ann%20pragma#source-annotations  ? That's a nice idea, and better than giving special meanings to INLINE/INLINEABLE pragmas

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
zliu41: Oh, right. Then we may need a second magical function, say `nonstrictArg`. Not ideal but it should work. I'll update the PR.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
zliu41: > Why shouldn't INLINE mean exactly what it means today?

INLINE should mean exactly what _GHC_ means by INLINE. Which is different from what our inliner do. See the example above. We want the second result, not the first.

thealmarty: hm....it does seem a bit confusing. So, if users want to define non-strict functions, they use the `INLINE` pragma. Otherwise, they use the `INLINEABLE` pragma, and it will be inlined with our PIR inliner?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+As said before, one reason why the semantics of Plutus Tx is difficult to predict is because we run GHC's pre-inliner, which may perform unconditionally inlining.
+Since GHC 9.0, GHC provides [an option](https://hackage.haskell.org/package/ghc-9.6.1/docs/GHC-Core-Opt-Simplify-Env.html#v:sm_pre_inline) to turn it off when running the simplifier.
+By turning it off, GHC (hopefully) won't inline anything at all, which means it won't change the strictness of Plutus Tx programs.
+
+At this time we can't simply turn the pre-inliner off, because doing so breaks the Plutus Tx compiler.
+For details, see [this note](https://github.com/input-output-hk/plutus/blob/dc57fe1b8497835ef2a7794c017c5a50c0b7e647/plutus-tx-plugin/src/PlutusTx/Plugin.hs#L123).
+The problem described in the note should, however, be resolved in a different way, as explained in the note.
+Once done, we can go ahead and turn it off.
+
+### Performing GHC-like Inlining for Functions with the INLINE Pragma
```
thealmarty, line 167: Can this section be moved right under the "Using `nonstrict` to Make Arguments Non-strict" section? And perhaps even rename this section to "Using the INLINE pragma to define non-strict functions" or something like that?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+Ziyang Liu <ziyang.liu@iohk.io>
+
+## Status
+
+Proposed
+
+## Background
+
+Plutus Tx is a language for writing Plutus scripts.
+It is a subset of Haskell and can be compiled to Untyped Plutus Core (hereafter referred to as PLC) by the Plutus Tx compiler, a GHC plugin.
```
thealmarty, line 16: We usually shorten it to `UPLC`?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+
+In a strict-by-default language, it would make sense that let-bindings be strict by default, since that's what almost all strict languages do; plus, non-strict bindings should be used cautiously since they can easily lead to unbounded amount of repeated evaluation.
+But if make let-bindings strict by default, then the question is, how does one make a non-strict let-binding?
+
+Haskell supports lazy patterns with the `~` operator, and it can be used on let-bindings, e.g., `let ~x = rhs`. So one possibility is to make non-strict let-bindings this way.
+However, lazy pattern has no effect on simple pattern bindings (i.e., the pattern is a single variable), and it makes no difference in GHC Core, which means the Plutus Tx compiler can't actually see the `~`.
+It is technically possible to make `~` work but it would require adding a GHC parser plugin to capture `~` in GHC Core, among other things, adding a great deal of complexity.
+
+This is the reason why let-bindings have been non-strict by default in Plutus Tx, and creating strict let-bindings requires using bang patterns.
+
```
thealmarty, line 59: Along the line of my other comments, this can be a section named "Using `nonstrict` to create nonstrict let bindings". Overall, I think these sections should be grouped together because they are proposed solutions to letting users define nonstrict things. The other sections at the same level, e.g., "Monomorphizing Plutus Tx Standard Library" is an optimization that we can consider doing, possibly with some more benchmarking.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
thealmarty: I agree. I'm not sure how GHC draws the line exactly. To avoid confusion maybe we can just talk about the two ways of inlining explicitly:

(1) inlining AND performing the substitution
(2) only the inlining

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
zliu41: Yes, but AFAIK OCaml's `foldr` in its stdlib is also strict. It seems to make sense for a strict language to provide a strict `foldr`.

Also I'm not aware of a language that is perfectly analogous to Plutus Tx such that we can just copy what it does. E.g., for OCaml the choice is strict vs. lazy, but for Plutus Tx it is strict vs non-strict, which are totally different dynamics.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
zliu41: Yes, I meant ad-hoc polymorphism. I think it is more of a problem for strict languages. The example I had in mind was: it is difficult to implement a short-circuiting `any :: Foldable t => (a -> Bool) -> t a -> Bool`. So the cost of ad-hoc polymorphism in this case is the lack of short-circuiting.

I agree the dictionary thing is not unique to strict languages.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,205 @@
...
+```haskell
+if expr1 == False then False else expr2
+```
+
+Normally, it is incorrect for the Plutus Tx compiler to inline `(&&)` in the GHC way, because in the `expr1 && expr2` example, before inlining, it evaluates `expr2` strictly, but after GHC-like inlining, `expr2` is evaluated non-strictly.
+But because of the `INLINE` pragma, the Plutus Tx compiler should inline `(&&)` in the GHC way.
+And this is desirable because we want `(&&)` to be non-strict.
+Functions that should not be inlined in the GHC way should use the `INLINABLE` pragma.
+
+This has recently been partially done by @michaelpj [#5183](https://github.com/input-output-hk/plutus/pull/5183), which propagates `INLINE` pragmas but doesn't yet perform GHC-like inlining.
```
thealmarty, line 205: I think it would be nice to have a conclusion summarizing the take away and explaining the next steps.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
zliu41: We already have proof that multiple method classes are expensive than single-method ones, unless I misunderstood what you meant by "the two versions"?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
zliu41: Sounds good.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
zliu41: This is mainly a terminology issue. but look at the example below: by "GHC-like inlining", I mean getting the second result, which is what we'll get if we run GHC's inliner. Our own inliner will only give us the first result. Please suggest better terminology.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
thealmarty: It's not a big deal, I'm just wondering if we should be more consistent.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
michaelpj: No, it still doesn't work. None of these approaches will work in general. _Function application is strict_. There is no way around that, without inlining the function, which requires inlining `foldr`. The problem is not the beta optimization, the problem is _function application_. Seriously, work through the example. I'll pay attention if you can show it works :p

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
michaelpj: Short-circuting is, in general, a pretty specific problem. Most uses of ad-hoc polymorphism don't particularly have anything to do with short-circuiting. It's pretty much only folding over lists with boolean operations where people expect this. So I think the framing is too harsh (and also I didn't get what it was saying).

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
michaelpj: I don't doubt they're more expensive, but how much?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
michaelpj: That's not what GHC means by `INLINE`. GHC means the same thing that we do. The fact that GHC's simplifier then inlines things is just a weird byproduct. `INLINE` should mean "inline this", nothing else.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
michaelpj: I think what you're seeing there is due to the fact that GHC's inliner is just a part of the simplifier. So it's getting both inlined and simplified. I think it's confusing to call this "GHC-style inlining" since it isn't IMO. It's inlining-followed-by-semantically-incorrect-for-PIR-beta-reduction :joy: 

What I think this most looks like is macro-expansion. Indeed, that's what I've suggested in the past as a way to get short-circuiting behaviour: use macros (i.e. TH). You can [very easily](https://github.com/input-output-hk/plutus/pull/4191#issuecomment-962004503) write a TH macro that expands inline the way you want it. And it's much less misleading, since it also clearly only works when you have the argument expressions to hand. Whereas your `(&&)` can be passed e.g. to a higher-order-function where it will _not_ behave lazily with no indication of that. 

As you can tell I'm pretty unconvinced by this proposal. I think it abuses the meaning of `INLINE`, and it's fragile and unreliable.

michaelpj:
To be clear, at a high level I am pro the overall direction of:
- Go for strict-by-default
- Provide user-controllable mechanisms for getting non-strict let bindings and if possible some kinds of non-strict functions

I just have issues with the proposed mechanisms.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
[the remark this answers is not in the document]
michaelpj: Yes, you're right, sorry. The difference is that GHC can inline _and then beta-reduce_ the application, and it's the _beta-reduction_ that is different in GHC.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
michaelpj: And yes, I think source annotations could be a good tool here.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
michaelpj: In fact, I think this behaviour of `nonstrict` is _actually pointless_. There is no difference between `... t ...` and `let ~x = t in ... x ...`. In both cases we are going to evaluate `t` exactly when we hit the location where `t` is originally. So standalone `nonstrict` is pointless.

TBH I still don't love this approach. I don't like that we're putting something on the RHS of the binding to affect the LHS, and that it doesn't really have a standalone meaning or work anywhere else. I wish we could use an annotation or something, but source annotations can't be used on nested bindings :(

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,209 @@
...
+Indeed, it is sometimes strict and sometimes non-strict, and it is fairly random and difficult to predict:
+
+- Let-bindings in Plutus Tx are by default non-strict, but function applications are strict.
+  That is, `let x = rhs in body` isn't equivalent to `(\x -> body) rhs`.
+  This is distinct from most other languages, regardless of their evaluation strategies, in which these two expressions are equivalent.
+
+- The conditional expression `if b then t else f` is always non-strict in `t` and `f`, like in most strict languages.
+  But what about `foo b t f = if b then t else f`? In a strict language, `foo` would be strict, and in a non-strict language, it would be non-strict.
+  But in Plutus Tx, due to how its compiler is currently implemented, whether `t` and `f` are evaluated strictly or not depends on whether GHC inlines `foo`.
+  Although we turn off GHC's inliner in the Plutus Tx compiler, GHC may still decide to unconditionally inline `foo` if it is used only once.
```
effectfully, line 34: It stops doing so if one uses `-fmax-simplifier-iterations=0`.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,209 @@
...
+  In general, it is not easy to predict whether GHC will inline something or not.
```
effectfully, line 35: It's easy if you turn off all GHC optimization. RV didn't want any inlining whatsoever and turning off all GHC optimization with `-fmax-simplifier-iterations=0` worked for them.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,209 @@
...
+
+The lack of consistency makes it difficult to reason about the efficiency of Plutus Tx programs, and difficult to write useful libraries for Plutus Tx that guarantee to have the desirable behaviors (e.g., short-circuiting).
+
+This ADR proposes a few changes to the Plutus Tx language and the compiler to make the semantics more predictable.
+It also proposes changes to the Plutus Tx standard library to make it more performant.
+
+## Proposed Changes
+
+Plutus Tx should be a strict-by-default language.
+Non-strict-by-default languages are rarely used in practice since non-strict evaluation leads to repeated evaluations.
```
effectfully, line 45: I think there's a lot of terminology confusion here.

1. strict vs non-strict isn't about evaluation per se, that distinction lies within the realms of _denotational_ semantics. E.g. `seq` is strict in both of its arguments, which doesn't tell us much about how things _evaluate_
2. call-by-value is an evaluation strategy than a strict language can adopt. Similarly, call-by-name or call-by-need are evaluation strategies that a non-strict language can adopt. At this level we're talking about the _operational_ semantics of a language. For `seq` we'd say that it's supposed to evaluate its second argument after evaluating the first one. However if GHC devs were to make `seq` evaluate both of its arguments in parallel, that would still comply with the denotational specification of the function
3. eager vs lazy is fairly vague terms for describing the specific _implementation_ of the evaluation strategy of a language

As such, non-strictness doesn't lead to any repeated evaluation as it isn't about evaluation at all. Laziness can lead to repeated evaluation if the call-by-name evaluation strategy is chosen, but that's horrible enough for it to be very uncommon (Agda used to be call-by-name for an example, not sure how want they've gone in making it call-by-need instead). In practice lazy languages implement call-by-need (Haskell, R) and that doesn't lead to repeated evaluation.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,209 @@
...
+Unless we implement lazy evaluation for PLC in the future, strictness should be the default for Plutus Tx.
```
effectfully, line 46: Fully agree.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,209 @@
...
+Given this, the following changes are proposed.
+
+### Making Let-bindings Strict By Default
+
+In a strict-by-default language, it would make sense that let-bindings be strict by default, since that's what almost all strict languages do; plus, non-strict bindings should be used cautiously since they can easily lead to unbounded amount of repeated evaluation.
```
effectfully, line 51: Unless non-strict bindings are backed by a sensible evaluation strategy. Which isn't our case, yes.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,209 @@
...
+But if make let-bindings strict by default, then the question is, how does one make a non-strict let-binding?
+
+Haskell supports lazy patterns with the `~` operator, and it can be used on let-bindings, e.g., `let ~x = rhs`. So one possibility is to make non-strict let-bindings this way.
+However, lazy pattern has no effect on simple pattern bindings (i.e., the pattern is a single variable), and it makes no difference in GHC Core, which means the Plutus Tx compiler can't actually see the `~`.
+It is technically possible to make `~` work but it would require adding a GHC parser plugin to capture `~` in GHC Core, among other things, adding a great deal of complexity.
```
effectfully, line 56: Can't we just enable the `Strict` pragma by default and then use `let ~x = rhs`? Quoting the [docs](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/strict.html#extension-Strict):

> Informally the `Strict` language extension switches functions, data types, and bindings to be strict by default, allowing optional laziness by adding `~` in front of a variable. This essentially reverses the present situation where laziness is default and strictness can be optionally had by adding `!` in front of a variable.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,209 @@
...
+Non-strict-by-default languages are rarely used in practice since non-strict evaluation leads to repeated evaluations.
+Unless we implement lazy evaluation for PLC in the future, strictness should be the default for Plutus Tx.
+Given this, the following changes are proposed.
+
+### Making Let-bindings Strict By Default
+
+In a strict-by-default language, it would make sense that let-bindings be strict by default, since that's what almost all strict languages do; plus, non-strict bindings should be used cautiously since they can easily lead to unbounded amount of repeated evaluation.
+But if make let-bindings strict by default, then the question is, how does one make a non-strict let-binding?
+
+Haskell supports lazy patterns with the `~` operator, and it can be used on let-bindings, e.g., `let ~x = rhs`. So one possibility is to make non-strict let-bindings this way.
```
effectfully, line 54: I'd either use "irrefutable patterns" or "lazy matching". "lazy patterns" sounds weird to me, but maybe it's just me.

effectfully (commented):
Reviewed only the first half for now.

I think I've reached my ultimate form and now nitpicking the hell out of every sentence, sorry about that.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
zliu41: Here's how it works: `f x (go xs)` normally becomes `let !acc = go_xs_compiled in f x acc` as a result of beta reduction.
But with `f x (nonstrict (go xs))`, it will annotate `go_xs_compiled` with `NONSTRICT`, and because of that annotation, beta-reduction, will create a nonstrict binding from it, i.e., `let ~acc = go_xs_compiled in f x acc`. Does that make sense?

michaelpj: Okay, but now you have to evaluate the call to `f` (e.g. `&&` in the case of `all`)... _which will evaluate acc_. So you have gained nothing.

zliu41: Hmm, right. Looking at [anyCheap.plc.golden](https://github.com/input-output-hk/plutus/blob/master/plutus-tx-plugin/test/Budget/anyCheap.plc.golden#L23) - it will make a difference if `acc` is nonstrict. But we do have `foldr` inlined here, which I guess is your point.

michaelpj: Right, in that example we _have_ inlined `foldr` and _also_ `f`. Then, yes, we have a chance. But that's a pretty heavy requirement. And crucially, that's not a property of  the _argument_ to the function, but a property of the function itself (is it going to get inlined?). Making non-strict function application happen is already hard, and I think can only really be done by inlining the function, and we certainly can't do it by just doing something to the _argument_.

thealmarty:
> I like this proposal a lot. We just have to make sure that all these great information and examples are properly catalogued in readthedocs.
> 
> One thing: since this is a fairly big semantic change, what happens to existing written PlutusTx scripts? Are we going to direct the plutustx users to use an older plutustx compiler version to retain the old behaviour or are we going to expose some extra plutus-tx-plugin flags to "control" this semantic change?

@bezirg has a good point. We should probably discuss the implications for existing scripts in the ADR?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
zliu41: Good question, I didn't consider `StrictData`. But how can we enable `StrictData` by default? It can probably be done in a DynFlags plugin, but that means using another plugin.

zliu41 (commented):
> and if possible some kinds of non-strict functions

@michaelpj looks like this non-strict functions thing isn't going to work, unless, like you said, we add `INLINE` to a lot of functions but that can lead to explosion of script sizes. The only other option I can see is to provide some kind of non-strict application support in the evaluator, but that comes with all kinds of other issues.

And this means there's no good way to implement `any :: Foldable t => (a -> Bool) -> t a -> Bool` such that it reliably short circuits.

As such I'll delete the sections about non-strict functions. With regard to the stdlib, I think we can provide a non-short-circuiting `foldr`, and then implement monomorphic `any` etc.  for lists and other concrete types, without using `foldr` - like some other languages do. Does that make sense?

michaelpj:
> With regard to the stdlib, I think we can provide a non-short-circuiting foldr, and then implement monomorphic any etc. for lists and other concrete types, without using foldr - like some other languages do. Does that make sense?

Yeah, I think that's the only option, really.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
michaelpj: Not `StrictData`, `Strict`. `Strict` is super weird and almost nobody uses it. I guess we could try it, but it would be invasive, and require changing the compilation flags of modules the plugin doesn't run on, perhaps :thinking: We could just say "we recommend using `Strict`" quite clearly and use it in our examples and libraries? Worth an experiment at least.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
zliu41: See the updated draft, where this statement is removed.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
zliu41: I fee like the discussion on the cost of multiple-method typeclasses is independent of the evaluation strategy discussion, so I removed it in the new draft.

zliu41 (commented):
@michaelpj @effectfully I did a major rewrite based on the comments. Please take another look.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
effectfully: ... and apparently I was wrong anyway.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
effectfully: Nope, not at all.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,245 @@
...
+
+This ADR proposes a few changes to the Plutus Tx language and the compiler to make the semantics more predictable.
+It also proposes changes to the Plutus Tx standard library to make it more performant.
+
+## Proposed Changes
+
+Plutus Tx should be a strict-by-default language, mainly because it cannot currently use call-by-need (lazy) evaluation.
+If it was non-strict by default, call-by-name evaluation would be the default evaluation strategy.
+Call-by-name is rarely used in practice since it leads to repeated evaluations.
+Unless we implement call-by-need evaluation for PLC in the future, strictness should be the default for Plutus Tx.
```
effectfully, line 48: :+1:

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Making Let-bindings Strict By Default
+
+In a strict language, it would make sense that let-bindings be strict by default, since that's what almost all strict languages do.
+Plus, non-strict bindings should be used cautiously in a language that adopts call-by-name (which is the only possible way to evaluate non-strict bindings in Plutus Tx), since they can easily lead to an unbounded amount of repeated evaluation.
+But if make let-bindings strict by default, then the question is, how does one make a non-strict let-binding?
+
+Haskell supports irrefutable patterns with the `~` operator, and it can be used on let-bindings, e.g., `let ~x = rhs`. So one possibility is to make non-strict let-bindings this way.
+However, irrefutable patterns have no effect on simple pattern bindings (i.e., the pattern is a single variable), and it makes no difference in GHC Core, which means the Plutus Tx compiler can't actually see the `~`.
+It is technically possible to make `~` work but it would require adding a GHC parser plugin to capture `~` in GHC Core, among other things, adding a great deal of complexity.
```
effectfully, line 59: This doesn't consider the option of enabling `Strict`?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,245 @@
...
+
+We have a few options:
+
+1. Do nothing, and encourage people to use the `Strict` GHC extension, which makes     let-bindings strict by default.
+   This is not technically "strict by default", since the `Strict` extension is off by default, but it is still ergonomically better than using bang patterns on every binding.
+
+   One argument against this is that a module may contain both on-chain and off-chain code, some modules contain off-chain code only, and the same piece of code may be both on-chain and off-chain.
+   For off-chain code one typically doesn't want to enable `Strict`.
+
+   Another disadvantage of using `Strict` is that it may cause GHC to do additional optimizations in ways that can break the Plutus Tx compiler.
```
effectfully, line 71: Kinda embarrassing, even though not our fault in any way.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,245 @@
...
+  ```haskell
+  nonstrict :: forall a. a -> a
+  nonstrict = id
+  {-# NOINLINE nonstrict #-}
+  ```
+
+  Then, to create a non-strict let-binding, we write
+
+  ```haskell
+  let x = nonstrict rhs in body
```
effectfully, line 89: We could reuse `GHC.Magic.lazy` if we wanted to.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,245 @@
...
+The advantages of making let-bindings strict by default, and using `nonstrict` for non-strict bindings are:
+
+- It makes it more convenient, rather than less convenient, to create strict bindings than non-strict bindings, which is what a strict language should do.
+- It works for both on-chain code and off-chain code. A let-binding is strict by default when used in on-chain code, and non-strict by default otherwise.
+
+The disadvantages are:
+
+- It is a magic function that needs special support in the Plutus Tx compiler.
+- This is a critical one: it is susceptible to GHC inlining bindings unconditionally.
+  If GHC decides to turn `let x = rhs in ...x...` into `...rhs...`, then `x` effectively becomes non-strict, and there's nothing the Plutus Tx compiler can do to recover `x`.
```
effectfully, line 104: Can we automatically create a binding for anything `nonstrict` if it's not bound already? With a bit of CSE it looks like we could recover inlined bindings.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,245 @@
...
+type Delay = (->) ()
+
+delay :: a -> Delay a
+delay a = \_ -> a
+
+force :: Delay a -> a
+force a = a ()
+
+class Foldable_Delay t where
+  foldr_delay :: (a -> Delay b -> Delay b) -> Delay b -> t a -> Delay b
```
effectfully, line 201: Looks like it can work with just

```haskell
  foldr_delay :: (a -> Delay b -> b) -> Delay b -> t a -> b
```

?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,245 @@
...
+      if p a then delay True
+      else acc
+```
+
+However, we will not be doing this in the standard library, because even though this version of `any` does short-circuit, it is not as efficient as the monomorphic version due to the additional cost from the `delay` and `force` machinery.
+If this version is indeed preferred for some reason, it can be provided in a third party library.
+
+### Turning off GHC's pre-inliner
+
+As said before, one reason why the semantics of Plutus Tx is difficult to predict is because we run GHC's pre-inliner, which may perform unconditionally inlining.
```
effectfully, line 225:
```suggestion
As said before, one reason why the semantics of Plutus Tx is difficult to predict is because we run GHC's pre-inliner, which may perform unconditional inlining.
```

effectfully (approved):
Great job, a very clear write-up.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
michaelpj: The problem is the opposite way: GHC can inline things that should be _strict_, and that changes the behaviour.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,250 @@
...
+- It makes it more convenient, rather than less convenient, to create strict bindings than non-strict bindings, which is what a strict language should do.
+- It works for both on-chain code and off-chain code. A let-binding is strict by default when used in on-chain code, and non-strict by default otherwise.
+
+The disadvantages are:
+
+- It is a magic function that needs special support in the Plutus Tx compiler.
+- This is a critical one: it is susceptible to GHC inlining bindings unconditionally.
+  If GHC decides to turn `let x = rhs in ...x...` into `...rhs...`, then `x` effectively becomes non-strict, and there's nothing the Plutus Tx compiler can do to recover `x`.
+
+If we can modify the Plutus Tx compiler such that we can turn off GHC's pre-inliner (which will be discussed later), then option 3 becomes the most preferrable.
```
michaelpj, line 106: I think this points to an under-appreciated benefit of option 1: if we can persuade GHC that the source language is strict, then we should have to worry a _lot less_ in general about GHC mis-optimizing things, because GHC should have a compatible view of the world to us.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,250 @@
...
+
+### Defining non-strict functions
+
+In Plutus Tx, like most other strict languages, only a few built-in constructs are non-strict (such as if-then-else).
+It is often cumbersome, if not impossible, to create user-defined non-strict functions.
+
+It would be useful to allow user-defined non-strict functions, so that we can define functions like `&&` and `||` that short-circuit properly.
+Some languages like Scala have special syntaxes for defining non-strict functions; we can't change Haskell's syntax, so we need to find a different solution.
+
+The idea is to make the Plutus Tx compiler perform inlining and beta reduction for functions with INLINE pragmas, i.e., the same as how GHC treats INLINE pragmas.
```
michaelpj, line 117: I'm sorry, I still hate it :grimacing: I _really_ don't want to use `INLINE` for this. I'd be more happy to use a custom source annotation as previously discussed, although I still very much don't love it.

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
```diff
@@ -0,0 +1,250 @@
...
+      if p a then delay True
+      else acc
+```
+
+However, we will not be doing this in the standard library, because even though this version of `any` does short-circuit, it is not as efficient as the monomorphic version due to the additional cost from the `delay` and `force` machinery.
+If this version is indeed preferred for some reason, it can be provided in a third party library.
+
+### Turning off GHC's pre-inliner
+
+As said before, one reason why the semantics of Plutus Tx is difficult to predict is because we run GHC's pre-inliner, which may perform unconditional inlining.
```
michaelpj, line 225: Of course, if we get GHC to agree with us that the language is strict, then we shouldn't have to worry about this so much, no?

doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md
effectfully: Oh, right, silly me. Thanks.

kwxm (commented):
I failed to review this at the time (April 2023!) and now I have no idea what the status is: it seems that there's a [version of this document](https://github.com/IntersectMBO/plutus/blob/983c2f54ae85376142657502075d0adbbe796365/doc/read-the-docs-site/adr/0005-plutus-tx-compiler-and-library.md) merged already and I don't know how these changes realte to that.

Anyway, this has been on my list of outstanding GitHub review requests for a long time and I'm making this comment to get rid of it.

effectfully:
@zliu41 do we still want this?

zliu41:
This is too old so I'll close it. I'll make a new ticket to describe what I think we should investigate next regarding the Plinth compiler and stdlib.

zliu41:
I opened https://github.com/IntersectMBO/plutus-private/issues/1298

PR #5281, opened 2023-05-01
Add some budget tests for short-circuiting functions

zliu41:
Currently nothing short circuits.


plutus-tx-plugin/test/Budget/andCheap.plc.golden
```diff
@@ -0,0 +1,52 @@
...
+          c
+            False
+            (c
+               True
+               (c
+                  True
+                  (c
+                     True
+                     (c True (c True (c True (c True (c True (c True n))))))))))
+        (\(ds : Bool) (ds : List Bool) -> Cons {Bool} ds ds)
```
effectfully, line 49: This won't parse back correctly. Also it'd be great to make it simply `Cons {Bool}`. I wonder if we should have an eta-contraction pass specifically for builtins (it's not safe to eta-contract things in the general case). We're wasting quite a bit here because of those lambdas.

effectfully (approved):
Great to have these.

plutus-tx-plugin/test/Budget/andCheap.plc.golden
michaelpj: Ideally we'd inline it also. Another case of type-lets getting in the way!

> I wonder if we should have an eta-contraction pass specifically for builtins

There aren't any builtins here?

I think we could do this kind of eta-contraction fairly easily using our arity analysis, which currently implies that a value won't have any effects until it receives all its arguments. So in this instance we could deduce that `Cons {Bool}` is pure, which is what we need. Indeed I guess this is a generic enhancement to our purity analysis if we have arity information available.

plutus-tx-plugin/test/Budget/Spec.hs
```diff
@@ -141,11 +165,51 @@ compiledFilter = $$(compile [||
       let ls = [1,2,3,4,5,6,7,8,9,10] :: [Integer]
        in PlutusTx.filter PlutusTx.even ls ||])
 
-compiledElem :: CompiledCode Bool
-compiledElem = $$(compile [||
+compiledAndCheap :: CompiledCode Bool
```
michaelpj, line 168: TBH these could do with comments. What are they about? It's just a big list of random vairants - what are they showing?

plutus-tx-plugin/test/Budget/andCheap.budget.golden
```diff
@@ -0,0 +1,2 @@
+({cpu: 10994100
```
michaelpj, line 1: Am I reading this right and `andCheap` is more expensive than `andExpensive`? lol

michaelpj (approved):

plutus-tx-plugin/test/Budget/andCheap.plc.golden
effectfully: > Ideally we'd inline it also. 

Can it be less efficient? I think in theory the current alignment allows us to partially compute `Cons {Bool}` only once (where by "in theory" I mean "if we perform eta-contraction")? And it would be less efficient in terms of program size.

> There aren't any builtins here?

I confused `Cons` and `MkCons` :facepalm: 

> I think we could do this kind of eta-contraction fairly easily using our arity analysis, which currently implies that a value won't have any effects until it receives all its arguments. So in this instance we could deduce that `Cons {Bool}` is pure, which is what we need. Indeed I guess this is a generic enhancement to our purity analysis if we have arity information available.

That does sound very sensible.

plutus-tx-plugin/test/Budget/andCheap.budget.golden
zliu41: I guess the `True` and `False` cases don't have the exact same cost, but not sure.

plutus-tx-plugin/test/Budget/andCheap.plc.golden
zliu41: > This won't parse back correctly.

I don't think readable PIRs are supposed to be parseable, are they?

> eta-contraction pass

We have a ticket: https://input-output.atlassian.net/browse/PLT-1204

effectfully: > I don't think readable PIRs are supposed to be parseable, are they?

Right, we're not parsing readable syntax. I guess I should've said that the printed program doesn't convey the meaning of the original one.

PR #5286, opened 2023-05-03
Make it build with ghc-9.6

erikd:


cabal.project
```diff
@@ -54,3 +54,35 @@ package nothunks
...
+    , *:ghc-prim
+    , *:mtl
+    , *:template-haskell
+    , *:transformers
+    , hedgehog:*
+    , microlens-mtl:*
+    , protolude:*
+    , th-compat:*
+
+-- Version on Hackages does not compile with `ghc-9.6`
```
michaelpj, line 69: These can also go inside the `if impl` block, in fact!

cabal.project
```diff
@@ -54,3 +54,35 @@ package nothunks
...
+-- https://github.com/haskell-foundation/foundation/issues/576
+source-repository-package
```
michaelpj, line 71: Honestly, maybe we should just try and purge our `foundation` dep, it's a pain :thinking:

michaelpj: Bah, it comes in via `cardano-crypto`, which we need for some old crypto functions. Maybe we can vendor them...

doc/read-the-docs-site/plutus-doc.cabal
```diff
@@ -70,7 +70,7 @@ executable doc-doctests
     , base               >=4.9      && <5
     , bytestring
     , containers
-    , flat               <0.5
+    , flat               ^>=0.6
```
michaelpj, line 73: We need to be careful with this, we have to check carefully that upstream hasn't broken anything. Is this change necessary for 9.6?

erikd: The onl version does not compile with `9.6` (probably due to `base` bounds).

michaelpj: If we can get away with `allow-newer`ing the base bounds then I think I'd mildly prefer this, since this is a sensitive dependency and this will change the bound for all GHC versions atm.

erikd: Would it not make sense to do one or both of the following?

* Write tests for `flat` and submit them upstream.
* Write tests for `flat` in the Plutus repo.

erikd: I tried removing the version bump for `flat` and added `allow-newer: flat:*`. The result was:
```
src/Flat/Decoder/Strict.hs:240:19: error: [GHC-31891]
    • Illegal term-level use of the type constructor or class ‘TA.Array’
    • imported qualified from ‘Data.Text.Array’ at src/Flat/Decoder/Strict.hs:45:1-53
    • Perhaps use variable ‘array’ (line 239)
    • In the first argument of ‘T.Text’, namely ‘(TA.Array array)’
      In the first argument of ‘return’, namely
        ‘(T.Text (TA.Array array) 0 (lengthInBytes `div` 2))’
      In a stmt of a 'do' block:
        return (T.Text (TA.Array array) 0 (lengthInBytes `div` 2))
    |
240 |   return (T.Text (TA.Array array) 0 (lengthInBytes `div` 2))
    |                   ^^^^^^^^
```

cabal.project
erikd: Done!

cabal.project
erikd: I am chasing the `foundation` problem via other channels.

michaelpj:
Now it appears to fail to build with 8.10 and fail to even plan with 9.2

erikd:
@michaelpj I rebased this against `master`, then made sure it builds here with `ghc-8.10`, `ghc-9.2` and `ghc-9.6`. 

I think the failure to create a build was due to the line:
```
    build-depends: ghc >=9.6 && <9.8
```
Would be happy to hear any solutions to that issue.

doc/read-the-docs-site/plutus-doc.cabal
bezirg: Ok I am on it.

michaelpj:
I think we discussed this a bit on Slack, but:
- `plutus-tx-plugin` depends on `ghc`, and is finicky.
- So we're only supporting a single version of GHC and it's non-trivial to switch which one.
- That version is 9.2, so the bounds on `ghc` should remain the same.
- We prevent `plutus-tx-plugin` and its downstream packages from building on other GHCs with `buildable: False` conditionals. That prevents cabal from trying to put it in the plan for other GHC versions and getting confused.
- Those buildable conditionals need to be updated to exclude GHCs newer than 9.2.
- That should leave us with all those packages building only on 9.2, which is what we want for now.

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinarySlab.hs
```diff
@@ -16,7 +17,7 @@ module Data.RandomAccessList.SkewBinarySlab
 import Data.Bits (unsafeShiftR)
 import Data.Vector.NonEmpty qualified as NEV
 import Data.Word
-import GHC.Exts
+import GHC.Exts hiding (One, Many)
```
michaelpj, line 20: Hydra doesn't seem to like this on GHC 9.2, not sure why, linear types are definitely in 9.2. Simplest solution is probably just to import it qualified.

erikd: Fixed.

cabal.project
```diff
@@ -53,3 +53,33 @@ package nothunks
...
+    , *:template-haskell
+    , *:transformers
+    , constraints-extras:*
+    , dependent-map:*
+    , hedgehog:*
+    , protolude:*
+    , th-compat:*
+
+  constraints:
+    , microlens >= 0.2.0.0
```
michaelpj, line 71: Please add a comment explaining these (I would expect us cabal to pick the compatible versions of these libraries automatically) and when we can remove them.

cabal.project
```diff
@@ -53,3 +53,33 @@ package nothunks
...
+    , *:base
+    , *:ghc-prim
+    , *:mtl
+    , *:template-haskell
+    , *:transformers
+    , constraints-extras:*
+    , dependent-map:*
+    , hedgehog:*
+    , protolude:*
+    , th-compat:*
```
michaelpj, line 68: I'm assuming we have upstream issues for everything that's jailbroken by this?

doc/read-the-docs-site/plutus-doc.cabal
michaelpj: The test infrastructure this needs got broken a while ago so it might be a little while until we can test this. I am tentatively in favour of merging since I think it's very likely not broken.

plutus-core/plutus-core/src/PlutusCore/Core/Type.hs
```diff
@@ -118,8 +118,13 @@ data Term tyname name uni fun ann
     | Constant ann (Some (ValueOf uni)) -- ^ constants
     | Builtin ann fun -- ^ builtin functions
     | Error ann (Type tyname uni ann) -- ^ fail with error
-    deriving stock (Show, Functor, Generic)
-    deriving anyclass (NFData)
+    deriving stock (Functor, Generic)
+
+deriving stock instance (Show tyname, Show name, GShow uni, Everywhere uni Show, Show fun, Show ann, Closed uni)
```
michaelpj, line 123: :cry:

plutus-core/plutus-core/stdlib/PlutusCore/StdLib/Meta/Data/Tuple.hs
```diff
@@ -28,11 +27,11 @@ import Control.Lens.Indexed (ifor, itraverse)
 import Data.Traversable
 
 -- | A Plutus Core (Scott-encoded) tuple.
-data Tuple term uni ann where
-    Tuple :: TermLike term TyName Name uni fun =>
```
michaelpj, line 32: This seems suspiciously like a semantic change? cc @effectfully

plutus-core/untyped-plutus-core/test/Evaluation/FreeVars.hs
```diff
@@ -1,6 +1,7 @@
 -- editorconfig-checker-disable-file
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies     #-}
+{-# LANGUAGE TypeOperators    #-}
```
michaelpj, line 4: You know, maybe we should just give in and use some `default-extensions` already.

prettyprinter-configurable/prettyprinter-configurable.cabal
```diff
@@ -82,7 +82,7 @@ test-suite prettyprinter-configurable-test
     , base                        >=4.9 && <5
     , megaparsec
     , parser-combinators
-    , prettyprinter-configurable  ^>=1.6
+    , prettyprinter-configurable
```
michaelpj, line 85: This shouldn't be changed.

michaelpj (commented):
Looks pretty much there. We need to update the GHC versions section of CONTRIBUTING, but I can do that when I add 9.6 CI.

plutus-core/plutus-core/stdlib/PlutusCore/StdLib/Meta/Data/Tuple.hs
effectfully: It was proposed by me to fix some inference issue with the newer version of GHC (I wonder if it's related to the disgusting abomination that quick look impredicativity is). It is a semantic change, but it doesn't seem to affect anything in our codebase and I'm almost sure it ended up there accidentally. `git blame` points to Mario.

plutus-core/plutus-core/src/PlutusCore/Core/Type.hs
effectfully: Yeah, it really sucks. But perhaps we could introduce some aliases in order not to write this whole thing our every time. Maybe some generic one that works for every constraint (we'll only need to special-case things like `GShow`). Not sure if it's worth it, though.

plutus-core/plutus-core/stdlib/PlutusCore/StdLib/Meta/Data/Tuple.hs
erikd: @effectfully is the one who suggested that change.

prettyprinter-configurable/prettyprinter-configurable.cabal
erikd: Cabal says that is a redundant constraint because, the test suite will always use the library defined in the cabal file.

michaelpj: Ah I see, I missed where it was, sorry!

plutus-core/plutus-core/stdlib/PlutusCore/StdLib/Meta/Data/Tuple.hs
michaelpj: :+1:

michaelpj:
I pushed a commit attempting to make the 9.6 CI work, and also merged master to get a haskell.nix update.

michaelpj:
Sigh: https://github.com/input-output-hk/haskell.nix/issues/1943

doc/read-the-docs-site/plutus-doc.cabal
bezirg: The tests are finished locally. 
Both plutus-master and plutus-master+flat0.6.patch pass the current mainnet script tests

michaelpj:
```
src/Codec/CBOR/Magic.hs:260:20: error: [GHC-83865]
    • Couldn't match expected type ‘Word64#’ with actual type ‘Word#’
    • In the first argument of ‘W64#’, namely ‘(toWord w#)’
      In the expression: W64# (toWord w#)
      In an equation for ‘w64’: w64 w# = W64# (toWord w#)
    |
260 |     w64 w# = W64# (toWord w#)
    |                    ^^^^^^^^^
```

maybe https://github.com/well-typed/cborg/issues/311 ?

Looks like we need to poke our WT colleagues or something.

michaelpj:
Non-darwin fails because of `cardano-crypto-class` using `-Werror` :| I think they fixed that on master but we need a release.

```

src/Cardano/Crypto/Hash/Short.hs:23:35: error: [GHC-58520] [-Wtype-equality-requires-operators, Werror=type-equality-requires-operators]
    The use of ‘~’ without TypeOperators
    will become an error in a future GHC release.
    Suggested fix: Perhaps you intended to use TypeOperators
   |
23 | instance (KnownNat n, CmpNat n 33 ~ 'LT) => HashAlgorithm (Blake2bPrefix n) where
   |                                   ^
```

michaelpj:
- There was a cardano-crypto-class release that should have 9.6 support, bumped our lower bound
- I've enabled the 9.6 CI on linux only for now, to get around the cborg issue. That'll be a problem in the long run but apparently WT are aware of it and we can probably just ignore it for now.
- I simplified the `allow-newer`s a lot
- I added a s-r-p for `inline-r`, since it actually needs some code changes to build :(

I still seem to be getting some warnings locally, I'll fix those next, and then I think this might be ready.

michaelpj:
An actual test failure! And an interesting one.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/MachineParameters.hs
```diff
@@ -42,7 +42,12 @@ data MachineParameters machinecosts fun val =
     , builtinsRuntime :: BuiltinsRuntime fun val
     }
     deriving stock Generic
-    deriving anyclass (NFData, NoThunks)
+    deriving anyclass (NFData)
+
+-- For some reason the generic instance gives incorrect nothunk errors,
```
michaelpj, line 47: FYI @effectfully

angerman:
force-pushed to re-excite ci.

erikd:
I think all that is missing is https://github.com/input-output-hk/plutus/issues/5309

PR #5288, opened 2023-05-03
Experiments with CEK machine args

michaelpj:
Don't look here.


michaelpj:
Seems fine, I think it's noise on the benchmarking machine taking it between "fine" and "quite good", so I think worth it. Also it'll enable doing some of https://github.com/input-output-hk/plutus/pull/5042.

effectfully:
> Seems fine, I think it's noise on the benchmarking machine taking it between "fine" and "quite good", so I think worth it.

Are you telling me that the benchmarking results can be this shaky:

```
| queens4x4/bt        |    -5.0%   |    -0.8%   |
| queens4x4/bm        |    -5.5%   |    -2.1%   |
| queens4x4/bjbt1     |    -6.1%   |    -3.8%   |
| queens4x4/bjbt2     |    -5.1%   |    -0.9%   |
| queens4x4/fc        |    -4.3%   |    -0.6%   |
| queens5x5/bt        |    -5.8%   |    +0.1%   |
| queens5x5/bm        |    -7.5%   |    -1.3%   |
| queens5x5/bjbt1     |    -6.3%   |    -2.2%   |
| queens5x5/bjbt2     |    -6.2%   |    -1.8%   |
| queens5x5/fc        |    -5.5%   |    -1.4%   |
```

? At this point why not just use astrology instead of our benchmarking machine?

BTW, can merge-vs-rebase affect the results? You seem to have done a `git merge` there.

michaelpj:
Hmm no, that's much closer to the previous run and not close to the old run. Maybe I have genuinely made the patch worse somehow, argh.

michaelpj:
It really looks like the same code as your previous PR, though!

michaelpj:
I have no idea what's up. It's at least not worse, and I do think that it's nicer having multiple frames rather than one, so I'm inclined to take it regardless, just frustrating that the speedup seems to have vanished!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/SteppableCek/Internal.hs
```diff
@@ -171,27 +178,28 @@ returnCek NoFrame val = do
...
         (next : todo') -> computeCek (FrameConstr ann env i todo' done' ctx) env next
-        _              -> returnCek ctx $ VConstr i (toList done')
+        _              -> returnCek ctx $ VConstr i done'
 -- s , case _ (C0 ... CN, ρ) ◅ constr i V1 .. Vm  ↦  s , [_ V1 ... Vm] ; ρ ▻ Ci
 returnCek (FrameCases ann env cs ctx) e = case e of
     (VConstr i args) -> case cs ^? wix i of
-        Just t  -> computeCek (FrameApplyValues ann args ctx) env t
+        Just t  ->
+              let ctx' = transferArgStack ann args ctx
+              in computeCek ctx' env t
```
effectfully, line 202: This `let` is the only difference between this PR and my original one that I'm able to spot. I can't imagine it making this much difference, but perhaps still is worth trying to inline it or to mark as `INLINE`.

effectfully (commented):
I know it's getting really annoying, but could you please create a separate PR with this same diff (modulo the comment below) on top of the most recent `master`? I _think_ I've seen cases where that would affect benchmarking results.

I'd just really suspicious that we were consistently getting pretty big speedups (-3.5% on average) and now we consistently get almost negligible ones (-0.5% on average in the worst case).

effectfully:
Ah, and yeah, I'm in favor of merging it eventually even if doesn't give us any substantial speedup as I think it still makes the code a bit better.

michaelpj:
I rebased it recently but I'll just try again.

michaelpj:
:| 

Now it's better. WTF. Maybe it was that tiny inlining? Maybe it's just chance?

michaelpj:
So variance of over 3% on reruns. Great.

PR #5290, opened 2023-05-03
Monomorphize short-circuiting functions in PlutusTx.Foldable (PLT-4229)

zliu41:
This enables them to short-circuit properly.


plutus-benchmark/nofib/test/Spec.hs
```diff
@@ -46,7 +46,7 @@ testClausify = testGroup "clausify"
                , testCase "formula3" $ mkClausifyTest Clausify.F3
                , testCase "formula4" $ mkClausifyTest Clausify.F4
                , testCase "formula5" $ mkClausifyTest Clausify.F5
-               , Tx.fitsInto "formula1 (size)" (Clausify.mkClausifyCode Clausify.F1) 4781
+               , Tx.fitsInto "formula1 (size)" (Clausify.mkClausifyCode Clausify.F1) 4004
```
effectfully, line 49: Should we turn this one and the others into a golden test or something?

plutus-tx-plugin/test/Budget/notElemCheap.pir-readable.golden
```diff
@@ -41,9 +38,4 @@ let
         (\(ds : integer) (ds : List integer) -> Cons {integer} ds ds)
         (Nil {integer})
 in
-Bool_match
-  (go x)
-  {all dead. Bool}
-  (/\dead -> False)
-  (/\dead -> True)
-  {all dead. dead}
```
effectfully, line 49: Was it GHC who optimized this for us or is our own optimizer now that clever as well?

plutus-tx/src/PlutusTx/List.hs
```diff
@@ -49,6 +65,57 @@ map f l = case l of
...
+
+{-# INLINABLE notElem #-}
+-- | The negation of `elem`.
+notElem :: Eq a => a -> [a] -> Bool
+notElem a = not . elem a
+
+{-# INLINABLE find #-}
+-- | Returns the leftmost element matching the predicate, or `Nothing` if there's no such element.
+find :: (a -> Bool) -> [a] -> Maybe a
+find f = go
```
effectfully, line 113: Optimize `findIndex` as well?

effectfully (approved):
LGTM

plutus-tx/src/PlutusTx/List.hs
effectfully: > `findIndex` is already optimized

But it traverses the entire list, because `findIndices` that it calls does so?

zliu41: I don't think so. `findIndices` is implemented directly, not in terms of `foldr`.

zliu41: Ah, sorry, I had a brain fog. Yes `findIndex` is un-optimized.

effectfully: Sure, but it still traverses the entire list, it's the point of the function to return all the indices, for which it needs to traverse the whole list! `findIndex` then throws all of them but the first one away, but Plutus is strict, hence the list of indices still has to materialize first.

I really hope it's not one of these times when I'm saying utter nonsense with a straight face :smiley:

effectfully: Eek, didn't see your message, ignore my last one.

michaelpj (approved):

PR #5291, opened 2023-05-03
[Textual] Propagate 'PrettyReadable' throughout everything

effectfully:
Pretty-printing intra-universe application requires access to the current context for readable pretty-printing in order for us not to generate redundant parens. This PR adds support for that.


plutus-core/plutus-core/src/PlutusCore/Core/Instance/Pretty/Readable.hs
```diff
@@ -58,7 +58,7 @@ instance
         TyForall _ name kind body ->
             typeBinderDocM $ \prettyBinding prettyBody ->
                 "all" <+> prettyBinding name kind <> "." <+> prettyBody body
-        TyBuiltin _ builtin       -> unitDocM $ pretty builtin
+        TyBuiltin _ builtin       -> prettyM builtin
```
effectfully, line 61: The gist of the change. Previously we'd just call `pretty` purely and now we call the monadic `prettyM` having access to the current context.

plutus-core/plutus-core/test/Pretty/Golden/Readable/StdLib/Data/Data/caseData.golden
```diff
@@ -1 +1 @@
-\(d : data) -> /\r -> \(fConstr : integer -> (list data) -> r) -> \(fMap : (list (pair data data)) -> r) -> \(fList : (list data) -> r) -> \(fI : integer -> r) -> \(fB : bytestring -> r) -> chooseData {unit -> r} d (\(u : unit) -> (/\a -> /\b -> /\c -> \(f : a -> b -> c) -> \(p : pair a b) -> f (fstPair {a} {b} p) (sndPair {a} {b} p)) {integer} {(list data)} {r} fConstr (unConstrData d)) (\(u : unit) -> fMap (unMapData d)) (\(u : unit) -> fList (unListData d)) (\(u : unit) -> fI (unIData d)) (\(u : unit) -> fB (unBData d)) ()
```
effectfully, line 1: Note the red (i.e. removed) parens.

plutus-tx-plugin/src/PlutusTx/Plugin.hs
```diff
@@ -215,8 +215,11 @@ mkPluginPass opts = GHC.CoreDoPluginPass "Core to PLC" $ \ guts -> do
 type PluginM uni fun = ReaderT PluginCtx (ExceptT (CompileError uni fun Ann) GHC.CoreM)
 
 -- | Runs the plugin monad in a given context; throws a Ghc.Exception when compilation fails.
-runPluginM :: (PLC.Pretty (PLC.SomeTypeIn uni), PLC.Closed uni, PLC.Everywhere uni PLC.PrettyConst, PP.Pretty fun)
-           => PluginCtx -> PluginM uni fun a -> GHC.CoreM a
+runPluginM
+    :: ( PLC.Pretty (PLC.SomeTypeIn uni), PLC.PrettyReadable (PLC.SomeTypeIn uni)
+       , PLC.Closed uni, PLC.Everywhere uni PLC.PrettyConst, PP.Pretty fun
```
effectfully, line 220: It really looks like we should `PrettyBuiltins uni fun` or something, given the amount of repetition.

plutus-core/plutus-ir/src/PlutusIR/Error.hs
```diff
@@ -57,7 +57,7 @@ instance PLC.AsFreeVariableError (Error uni fun a) where
 ------------------
 
 type PrettyUni uni ann =
-    ( PP.Pretty (PLC.SomeTypeIn uni)
+    ( PP.Pretty (PLC.SomeTypeIn uni), PLC.PrettyReadable (PLC.SomeTypeIn uni)
     , PLC.Closed uni, uni `PLC.Everywhere` PLC.PrettyConst
```
effectfully, line 61: Or maybe we should at least make of this part more often.

plutus-core/plutus-core/test/Pretty/Golden/Readable/StdLib/Data/Data/caseData.golden
effectfully: I'm not going to pollute the changelog with such a tiny change.

plutus-tx-plugin/src/PlutusTx/Plugin.hs
effectfully: I decided to bite the bullet and perform the refactoring, but I'll do it in a separate PR.

plutus-core/plutus-core/src/PlutusCore/Error.hs
```diff
@@ -218,6 +219,7 @@ instance
 
 instance
         ( Pretty (SomeTypeIn uni)
```
michaelpj, line 221: suspcious?

michaelpj (approved):
Looks fine, just I would have expected some of the other constraints to be wrong/redundant now?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -195,11 +195,11 @@ instance HasRenderContext config => PrettyBy config (DefaultUni a) where
...
 
+instance HasRenderContext config => PrettyBy config (SomeTypeIn DefaultUni) where
+    prettyBy config (SomeTypeIn uni) = prettyBy config uni
+
 -- | This always pretty-prints parens around type applications (e.g. @(list bool)@) and
 -- doesn't pretty-print them otherwise (e.g. @integer@).
--- This is so we can have a single instance that is safe to use with both the classic and the
--- readable pretty-printers, even though for the latter it may result in redundant parens being
--- shown. We are planning to change the classic syntax to remove this silliness.
 instance Pretty (DefaultUni a) where
```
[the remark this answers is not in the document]
effectfully: Well, `Pretty` is for classic pretty-printing and `PrettyBy` is for readable one. We render built-in functions the same way in both, so we use `Pretty` for both, and we used to render built-in types the same way, but now they're just a tiny bit different. If we were to unify the two, then I think we would have to make the classic pretty-printer dependent on the readable one, so that we can call an equivalent of

```haskell
prettyBy $ RenderContext ToTheRight juxtFixity
```

(the current implementation of `Pretty (DefaultUni a)`) in the classic pretty-printer. But then both classic and readable pretty-printers carry a `configName` and we'd have to basically convert the config for the classic pretty-printer into the config of the readable one. Which doesn't seem too hard, but I don't feel comfortable entangling the two pretty-printers that much.

... on the other hand if we made the constraint `PrettyBy RenderContext` in both the cases and

1. created a `RenderContext` on the fly in the classic pretty-printer
2. extracted the `RenderContext` from the current config in the readable pretty-printer 

then it all would actually work, it seems.

Thanks for nagging me about it! I'm going to try out that latter idea.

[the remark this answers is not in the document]
michaelpj: I think that's kind of what I expected. In my head it's something like:
```
fooWithArg arg = ....
foo = fooWithArg def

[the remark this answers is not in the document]
Comment (effectfully) [reply plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs:203]: Yeah, I'm just being stupid.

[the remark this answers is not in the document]
Comment (effectfully) [reply plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs:204]: The `Pretty` instances have to stay, because we do occasionally need them when fitting universes stuff into an function with a `Pretty a` constraint (in particular, that happens in the QuviQ's unification machinery). But all duplication is gone, which is good.

**Comment by effectfully** [comment]
Ready for rereview, but given the existing approval, I'm enabling auto-merge to keep me going.

File: plutus-core/plutus-core/src/PlutusCore/Pretty/Extra.hs [hunk 7]
```diff
@@ -1,19 +1,52 @@
...
+    lmap = coerce (withReader :: (config' -> config) -> Reader config a -> Reader config' a)
+    {-# INLINE lmap #-}
+
+    rmap
+        :: forall config a b.
+           (a -> b)
+        -> InContextM config a
+        -> InContextM config b
+    rmap = coerce (mapReader :: (a -> b) -> Reader config a -> Reader config b)
+    {-# INLINE rmap #-}
```
mjaskelioff, line 41: Is this instance used anywhere?

mjaskelioff (approved):
I'm not familiar with the pretty printing machinery but everything looks sensible to me.

plutus-core/plutus-core/src/PlutusCore/Pretty/Extra.hs
effectfully: [Yes](https://github.com/input-output-hk/plutus/blob/baa58bc6a21d840d44737f3850e406e247d173f9/plutus-core/plutus-core/src/PlutusCore/Core/Instance/Pretty/Readable.hs#L61).

effectfully:
image

effectfully:
Oh come on. @michaelpj CI is bullying me.

image

PR #5300, opened 2023-05-04
Make PIR available in the result of liftCode

zliu41:
This enables us to see the PIR of many test cases, which is useful for debugging.

### DIFF
--- a/plutus-tx/src/PlutusTx/Lift.hs
+++ b/plutus-tx/src/PlutusTx/Lift.hs
@@ -24,7 +24,6 @@ import PlutusTx.Lift.Class qualified as Lift
 import PlutusTx.Lift.Instances ()
 import PlutusTx.Lift.TH (LiftError (..), makeLift, makeTypeable)
 
-import Data.Bifunctor
 import PlutusIR
 import PlutusIR qualified as PIR
 import PlutusIR.Compiler
@@ -42,11 +41,12 @@ import PlutusCore.Version qualified as PLC
 
 import UntypedPlutusCore qualified as UPLC
 
+import Control.Arrow ((***))
 import Control.Exception
 import Control.Lens hiding (lifted)
 import Control.Monad.Except hiding (lift)
 import Control.Monad.Reader hiding (lift)
-
+import Data.Bifunctor
 import Data.Default.Class
 import Data.Proxy
 import Data.Typeable qualified as GHC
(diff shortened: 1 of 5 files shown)


plutus-tx/src/PlutusTx/Lift.hs
```diff
@@ -85,10 +85,10 @@ safeLift v x = do
           -- that takes all the compilation options and everything.
           & set (ccOpts . coDatatypes . dcoStyle) (if v >= PLC.plcVersion110 then SumsOfProducts else ScottEncoding)
         ucOpts = PLC.defaultCompilationOpts & PLC.coSimplifyOpts . UPLC.soMaxSimplifierIterations .~ 0
-    plc <- flip runReaderT ccConfig $ compileProgram (Program () v lifted)
+    plc <- flip runReaderT ccConfig $ compileProgram (Program () v pir)
     uplc <- flip runReaderT ucOpts $ PLC.compileProgram plc
     (UPLC.Program _ _ db) <- traverseOf UPLC.progTerm UPLC.deBruijnTerm uplc
```
effectfully, line 90: What's `db`?

plutus-tx/src/PlutusTx/Lift.hs
```diff
@@ -101,8 +101,8 @@ safeLiftProgram
        , PrettyPrintable uni fun
        , Default (PLC.CostingPart uni fun)
        )
-    => PLC.Version -> a -> m (UPLC.Program UPLC.NamedDeBruijn uni fun ())
-safeLiftProgram v x = UPLC.Program () v <$> safeLift v x
+    => PLC.Version -> a -> m (PIR.Program PLC.TyName PLC.Name uni fun (), UPLC.Program UPLC.NamedDeBruijn uni fun ())
+safeLiftProgram v x = (PIR.Program () v *** UPLC.Program () v) <$> safeLift v x
```
effectfully, line 105: I think `bimap` is a bit more standard, but I'm obviously nitpicking.

effectfully (approved):

plutus-tx/src/PlutusTx/Lift.hs
zliu41: I think it means De Bruijn term.

michaelpj:
I don't really like this, I think it makes the interface much less usable. We could just have these functions return PIR programs (as the underlying `Lift` class does), and only compile them to UPLC when we construct `CompiledCode`?

zliu41:
> We could just have these functions return PIR programs (as the underlying Lift class does)

That would require bigger changes to the test code, although the test code doesn't necessarily represent how usable the interface is. How about providing separate functions returning PIR and UPLC?

effectfully:
> I don't really like this, I think it makes the interface much less usable. We could just have these functions return PIR programs (as the underlying `Lift` class does), and only compile them to `UPLC` when we construct `CompiledCode`?

Given this, I think I should clarify that my approvals for pull requests related to Plutus Tx, especially the compiler, shouldn't be treated too seriously. I don't really know anything about the Plutus Tx compiler.

PR #5307, opened 2023-05-08
Fix processSingleBinding

zliu41:
Fixed a bug in `processSingleBinding`, and added a test case that would otherwise fail.


michaelpj (approved):
Great, thanks.

PR #5308, opened 2023-05-09
Better size measure + beta reduction for inlining fully saturated applications

zliu41:
Consider

```haskell
let f = \x y -> f_body
 in ...f arg1 arg2...
```
Previously, we call `sizeIsAcceptable` on `f_body`, and if it returns True, and if other inlining criteria are met, then we replace `f` with RHS, i.e., `let f = \x y -> f_body in ...(\x y -> f_body) arg1 arg2...`.

In this new approach, as discussed in #5264, we compare the size (in terms of number of AST nodes) of `f arg1 arg2` against the size of `f_body [x := arg1, y := arg2]`.

If the latter is no bigger, and if `arg1` and `arg2` are pure, then we directly replace `f arg1 arg2` with `f_body [x := arg1, y := arg2]` (i.e., inlining + beta).

The reason that we perform beta is because (1) otherwise, we'd be relying on the inliner to inline `arg1` and `arg2`, but there is no guarantee that the inliner will do so, especially if either `arg1` or `arg2` (or both) occurs more than once; (2) we need to obtain 
`f_body [x := arg1, y := arg2]` anyway in order to compare the sizes, so we are not doing additional work.

The requirement that `arg1` and `arg2` be pure is because otherwise it isn't effect-safe to perform the beta reduction.

Note that with this change, we no longer need to care about whether or not an application is fully saturated - the same method is applicable to non-saturated applications. But I'll leave this generalization to a future PR.


plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -109,25 +110,25 @@ an under-approximation of how many arguments the term may need.
...
+type Arity tyname name = [Param tyname name]
 
 -- | Info attached to a let-binding needed for call site inlining.
-data VarInfo tyname name uni fun ann =
-  MkVarInfo {
-    varStrictness :: Strictness
-    ,varDef       :: InlineTerm tyname name uni fun ann
+data VarInfo tyname name uni fun ann = MkVarInfo
+    { varStrictness :: Strictness
+    , varDef        :: Term tyname name uni fun ann
```
thealmarty, line 118: This needs to be an `InlineTerm`, see [Dupable](https://github.com/input-output-hk/plutus/blob/2d0a953ffcdce0975595c3c9422ff1202d7ae63e/plutus-core/plutus-core/src/PlutusCore/Rename/Internal.hs#L46) and [this comment](https://github.com/input-output-hk/plutus/pull/5251#discussion_r1163900145).

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -12,11 +12,15 @@ See note [Inlining of fully applied functions].
...
+import PlutusIR.Analysis.Size
 import PlutusIR.Contexts
 import PlutusIR.Core
 import PlutusIR.Transform.Inline.Utils
+import PlutusIR.Transform.Substitute
+
+import Control.Monad.Extra
+import Control.Monad.State
+
 {- Note [Inlining of fully applied functions]
```
thealmarty, line 24:
```suggestion
{- Note [Inlining and beta reduction of fully applied functions]
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -45,10 +49,10 @@ an under-approximation of how many arguments the term may need.
 e.g. consider the term @let id = \x -> x in id@: the variable @id@ has syntactic
 arity @[]@, but does in fact need an argument before it does any work.
 
-In the `body`, where `v` is *called*,
+In `body`, where `v` is called,
 if it was given the `n` term or type arguments in the correct order, then it is *fully applied*.
-If `VBody` is acceptable in size and cost, we inline the call site of the fully applied `v` in this
-case, i.e., we replace `v` in the `body` with `rhs`. E.g.
+If the changes in size and cost caused by the inlining is acceptable, and the inlining is
+effect-safe, we inline the call site of the fully applied `v` in, and perform beta reduction. E.g.
```
thealmarty, line 55:
```suggestion
effect-safe, we inline the call site of the fully applied `v`, and perform beta reduction. E.g.
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -45,10 +49,10 @@ an under-approximation of how many arguments the term may need.
 e.g. consider the term @let id = \x -> x in id@: the variable @id@ has syntactic
 arity @[]@, but does in fact need an argument before it does any work.
 
-In the `body`, where `v` is *called*,
+In `body`, where `v` is called,
 if it was given the `n` term or type arguments in the correct order, then it is *fully applied*.
-If `VBody` is acceptable in size and cost, we inline the call site of the fully applied `v` in this
-case, i.e., we replace `v` in the `body` with `rhs`. E.g.
+If the changes in size and cost caused by the inlining is acceptable, and the inlining is
```
thealmarty, line 54:
```suggestion
If the changes in size and cost caused by the inlining is acceptable, and the inlining and the beta reduction is
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -60,120 +64,110 @@ becomes
...
-Because a function can be called in the `body` multiple times and may not be fully applied for all
-its calls, we cannot simply keep a substitution map like in `Inline`,
-which substitute *all* occurrences of a variable. Instead, we store all in scope variables in a
-map, `Utils.NonRecInScopeSet`, with all information needed for inlining saturated functions.
-
-To find out whether a function is fully applied, when we encounter a variable in the `body`,
-we find its arity from the `Utils.NonRecInScopeSet` map and compare it with the list
-of arguments it has been applied to at that site. If it is fully applied, we inline it there.
-
-Note that over-application of a function would also be inlined. E.g.:
```
thealmarty, line 84: You don't want to keep this part about over-application?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -60,120 +64,110 @@ becomes
...
+Note that because `f` occurs twice in the original term, and the RHS of `f` is not small enough,
+`f` is not unconditionally inlined. Instead, the second usage of `f` is inlined. The first
+usage is not inlined because it is not fully applied.
 
-(2) How do we decide whether cost and size are acceptable?
+(2) How do we decide whether a fully saturated application can be inlined?
 
-We currently reuse the heuristics 'Utils.sizeIsAcceptable' and 'Utils.costIsAcceptable'
-that are used in unconditional inlining. For
+For size, we compare the sizes (in terms of AST nodes before and after the inlining), and
```
thealmarty, line 75:
```suggestion
For size, we compare the sizes (in terms of AST nodes before and after the inlining and beta reduction), and
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -60,120 +64,110 @@ becomes
...
+inline only if it does not increase the size. In the above example, we count the number of AST
+nodes in `f a b` and in `a`. The latter is smaller, which means inlining reduces the size.
 
-let v = \x1.\x2...\xn.VBody in body
+For cost, we check whether the RHS (in this example, `\x. \y -> x`) has a small cost.
+If the RHS has a non-zero arity, then the cost is always small (since a lambda or a type
```
thealmarty, line 80: I think we can just mention [costIsAcceptable](https://github.com/input-output-hk/plutus/blob/2d0a953ffcdce0975595c3c9422ff1202d7ae63e/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs#L328) here. I don't see why we should discuss the case of lambda separately.

```suggestion
For cost, we check whether the RHS (in this example, `\x. \y -> x`) has an acceptable cost as specified in `Utils.costIsAcceptable`. 
```
```

File: plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs [hunk 8]
```diff
@@ -248,7 +248,7 @@ processSingleBinding body = \case
                 void $ modify' $
                     extendVarInfo
                         n
-                        (MkVarInfo s (Done (dupable processedRhs)) varArity bodyToCheck)
+                        (MkVarInfo s processedRhs arity bodyToCheck)
```
thealmarty, line 251: see below, should be an `InlineTerm`, not just `Term`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -60,120 +64,110 @@ becomes
...
+          args'
+      _ -> pure Nothing
+
+    -- Checks whether it is safe (i.e., preserves semantics) to perform beta reduction,
+    -- namely, turning `(\a -> body) arg` into `body [a := arg]`.
+    safeToSubst :: name -> Term tyname name uni fun ann -> InlineM tyname name uni fun ann Bool
+    safeToSubst n = effectSafe (varBody info) Strict n <=< checkPurity
+
+-- | Consider whether to inline an application.
+inlineSaturatedApp ::
```
thealmarty, line 143: I like the old name `considerInlineSat` better, it shows that it may or may not be inlined. Perhaps it should be changed to `considerInlineBetaReduceSatFn` or something like that.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -60,120 +64,110 @@ becomes
...
-                -- rename the term before substituting in
-                renamed <- renameTerm defAsInlineTerm
-                pure $ fillAppContext renamed ctx
-            else pure tm
-          -- The variable maybe a *recursive* let binding, in which case it won't be in the map,
-          -- and we don't process it. ATM recursive bindings aren't inlined.
-          Nothing -> pure tm
-      -- if the term being applied is not a `Var`, don't inline
-      _ -> pure tm
+            pure $ if sizeIsOk && costIsOk && defPure then fullyApplied else t
```
thealmarty, line 168: If we are worried about the cost of `fullyApplyAndBetaReduce`, we can do the `costIsOk` and `defPure` checks first, these checks don't need anything from `fullyApplyAndBetaReduce`. If either of them are false, we stop. Otherwise, do the case match on `fullyApplyAndBetaReduce`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -60,120 +64,110 @@ becomes
...
+      (TypeParam param : arity', TypeAppContext arg _ args') ->
+        go
+          (termSubstTyNames (\n -> if n == param then Just arg else Nothing) acc)
+          arity'
+          args'
+      _ -> pure Nothing
+
+    -- Checks whether it is safe (i.e., preserves semantics) to perform beta reduction,
+    -- namely, turning `(\a -> body) arg` into `body [a := arg]`.
+    safeToSubst :: name -> Term tyname name uni fun ann -> InlineM tyname name uni fun ann Bool
```
thealmarty, line 139: maybe `safeToBetaReduce` is better? We do do other substitutions.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -60,120 +64,110 @@ becomes
...
+  (InliningConstraints tyname name uni fun) =>
+  Term tyname name uni fun ann ->
+  InlineM tyname name uni fun ann (Term tyname name uni fun ann)
+inlineSaturatedApp t
+  | (Var _ann name, args) <- splitApplication t =
+      gets (lookupVarInfo name) >>= \case
+        Just varInfo -> fullyApplyAndBetaReduce varInfo args >>= \case
+          Just fullyApplied -> do
+            let def = varDef varInfo
+                -- Inline only if the size is no bigger than not inlining.
```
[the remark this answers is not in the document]
zliu41: Added the following:

```
The number of AST nodes is an approximate rather than a precise measure. It correlates,
but doesn't directly map to the size of the serialised script. Different kinds of AST nodes
may take different number of bits to encode; in particular, a `Constant` node always
counts as one AST node, but the constant in it can be of arbitrary size. Then, would it be
better to use the size of the serialised term, instead of the number of AST nodes? Not
necessarily, because other transformations, such as CSE, may change the size further;
specifically, if a large constant occurs multiple times in a program, it may get CSE'd.

In general there's no reliable way to precisely predict the size impact of an inlining
decision, and we believe the number of AST nodes is in fact a good approximation.
```

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
zliu41: That shouldn't be a problem. `fullyApplyAndBetaReduce` is linear.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: I remember writing something similar and both of you didn't think it was necessary! I guess I was right after all!

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: I don't really care that much, but this is in line with GHC's naming actually.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
thealmarty: OK looking at  [Dupable](https://github.com/input-output-hk/plutus/blob/2d0a953ffcdce0975595c3c9422ff1202d7ae63e/plutus-core/plutus-core/src/PlutusCore/Rename/Internal.hs#L46) again I've apparently used it wrong. And if we are using `Dupable` we don't need to rename, because that's the point. So it seems we should put it in as an `InlineTerm` and call `liftDupable` when we do the substitution. I think we may not need to rename in the unconditional inliner too in this case.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -60,120 +64,110 @@ becomes
...
                 -- The definition itself will be inlined, so we need to check that the cost
                 -- of that is acceptable. Note that we do _not_ check the cost of the _body_.
                 -- We would have paid that regardless.
-                -- Consider e.g. `let y = \x. f x`. We pay the cost of the `f x` at every call
-                -- site regardless. The work that is being duplicated is the work for the lambda.
-                defCostOk = costIsAcceptable def
+                -- Consider e.g. `let y = \x. f x`. We pay the cost of the `f x` at
+                -- every call site regardless. The work that is being duplicated is
+                -- the work for the lambda.
+                costIsOk = costIsAcceptable def
```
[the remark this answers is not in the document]
michaelpj: Hmmm, maybe we can actually glue together multiple pieces of argumentation. Suppose that `costIsAcceptable def` is true. Then we are going to:
- Inline the variable. This is a cost saving since `costIsAcceptable def` is true
- Do beta reduction. This is a cost saving because we are following our usual inlining heuristics, which should ensure that the transformation is a cost saving

So I think as long as we do match the existing inlining heuristics for step 2, the logic does actually check out? It's a bit subtle, though, worth writing down carefully.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
michaelpj: The point of `Dupable` is to signal that "this is a term that you need to rename when you use it". We indeed shouldn't be matching on the `Dupable` constructor, that does rather defeat the point! Sorry I missed that before!

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
michaelpj: Yes, this is suspicious.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -60,120 +66,130 @@ becomes
...
+      Arity tyname name ->
+      AppContext tyname name uni fun ann ->
+      InlineM tyname name uni fun ann (Maybe (Term tyname name uni fun ann))
+    go acc arity args = case (arity, args) of
+      -- success
+      ([], _) -> pure . Just $ fillAppContext acc args
+      (TermParam param : arity', TermAppContext arg _ args') ->
+        ifM
+          (safeToBetaReduce param arg)
+          ( do
```
michaelpj, line 138: https://github.com/input-output-hk/plutus/blob/master/STYLEGUIDE.adoc#avoid-multi-line-parenthesized-expressions

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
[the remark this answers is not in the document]
michaelpj: SGTM

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -60,120 +66,140 @@ becomes
...
+                  (\n -> if n == param then Just <$> PLC.rename arg else pure Nothing)
+                  acc
+              go acc' arity' args'
+            else pure Nothing
+        (TypeParam param : arity', TypeAppContext arg _ args') -> do
+          acc' <-
+            termSubstTyNamesM
+              (\n -> if n == param then Just <$> PLC.rename arg else pure Nothing)
+              acc
+          go
```
michaelpj, line 155: Is this some fourmolu thing? Seems weird to break this since it clearly fits on one line...

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -60,120 +66,140 @@ becomes
...
+            acc'
+            arity'
+            args'
+        _ -> pure Nothing
+
+      -- Is it safe to turning `(\a -> body) arg` into `body [a := arg]`?
```
michaelpj, line 161:
```suggestion
      -- Is it safe to turn `(\a -> body) arg` into `body [a := arg]`?
```

michaelpj (approved):

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
zliu41: Good catch; that was me.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
thealmarty: I don't think using `Dupable` on `varBody` makes sense. E.g., for `f = \ x y -> x + y`, renaming `varBody` means renaming `x + y`, which is not a great idea because now `x` and `y` are free. You need to do the renaming of `f` i.e., `varDef`.

thealmarty: I have since found that the renamer doesn't touch free variables. It's more of a problem that `varBody`  can't be renamed, because it looks like a bunch of free variables, and thus won't be touched. See [this comment](https://github.com/input-output-hk/plutus/pull/5395#discussion_r1238789249).

michaelpj: What's wrong with the way `varBody` is renamed? The renamer won't touch the free variables, which is correct, but it will rename any bound variables in there, which is what we want?

PR #5311, opened 2023-05-09
[PlutusTx] [Test] Add 'InterList'

effectfully:
This adds a test showing how compilation of a value of a non-regular data type to TPLC looks like.

### DIFF
--- a/plutus-tx-plugin/test/Lib.hs
+++ b/plutus-tx-plugin/test/Lib.hs
@@ -25,6 +25,7 @@ import PlutusPrelude
 import PlutusCore.Test
 
 import PlutusTx.Code
+import PlutusTx.Test ()
 
 import PlutusCore qualified as PLC
 import PlutusCore.Evaluation.Machine.ExBudgetingDefaults qualified as PLC
@@ -32,12 +33,6 @@ import PlutusCore.Pretty
 
 import UntypedPlutusCore qualified as UPLC
 import UntypedPlutusCore.Evaluation.Machine.Cek
-
-instance (PLC.Closed uni, uni `PLC.Everywhere` Flat, Flat fun) =>
-            ToUPlc (CompiledCodeIn uni fun a) uni fun where
-    toUPlc v = do
-        v' <- catchAll $ getPlcNoAnn v
-        toUPlc v'
 
 goldenPir
     :: (PLC.Closed uni, uni `PLC.Everywhere` PrettyConst, uni `PLC.Everywhere` Flat, PrettyParens (PLC.SomeTypeIn uni), Pretty fun, Flat fun)
--- a/plutus-tx-plugin/test/Plugin/Data/recursive/interListConstruct.tplc.golden
+++ b/plutus-tx-plugin/test/Plugin/Data/recursive/interListConstruct.tplc.golden
@@ -1,0 +1,749 @@
+(program
+  1.1.0
+  [
+    [
+      [
+        {
+          (abs
+            InterList_i0
+            (fun (type) (fun (type) (type)))
+            (lam
+              InterCons_i0
+              (all
+                a_i0
+                (type)
+                (all
+                  b_i0
+                  (type)
+                  (fun
+                    a_i2
+                    (fun
+                      b_i1
+                      (fun
+                        [ [ InterList_i4 b_i1 ] a_i2 ]
+                        [ [ InterList_i4 a_i2 ] b_i1 ]
+                      )
+                    )
+                  )
+                )
+              )
+              (lam
+                InterNil_i0
+                (all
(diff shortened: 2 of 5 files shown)


plutus-tx-plugin/test/Lib.hs
```diff
@@ -33,12 +34,6 @@ import PlutusCore.Pretty
 import UntypedPlutusCore qualified as UPLC
 import UntypedPlutusCore.Evaluation.Machine.Cek
 
-instance (PLC.Closed uni, uni `PLC.Everywhere` Flat, Flat fun) =>
-            ToUPlc (CompiledCodeIn uni fun a) uni fun where
-    toUPlc v = do
-        v' <- catchAll $ getPlcNoAnn v
-        toUPlc v'
```
effectfully, line 40: The instance was duplicated.

plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -198,6 +201,18 @@ instance (PLC.Closed uni, uni `PLC.Everywhere` Flat, Flat fun) =>
...
+         ( PLC.PrettyParens (PLC.SomeTypeIn uni)
+         , PLC.GEq uni, PLC.Typecheckable uni fun
+         , PLC.Closed uni, uni `PLC.Everywhere` PrettyConst, Pretty fun
+         , uni `PLC.Everywhere` Flat, Flat fun, Default (PLC.CostingPart uni fun)
+         ) => ToTPlc (CompiledCodeIn uni fun a) uni fun where
+    toTPlc v = do
+        mayV' <- catchAll $ getPir v
+        case mayV' of
+            Nothing -> fail "No PIR available"
+            Just v' -> toTPlc v'
```
effectfully, line 214: It took me a ridiculous amount of time to figure out where to put this instance and how to implement it. The testing machinery seems scattered across multiple files with a lot of ad hoc logic etc. Not sure if we can do much about it, though.

plutus-tx-plugin/test/Plugin/Data/recursive/interListConstruct.tplc.golden
```diff
@@ -0,0 +1,749 @@
...
+                          (fun (type) (fun (type) (type)))
+                          (sop [a_i6 b_i5 [ [ InterList_i1 b_i5 ] a_i6 ]] [])
+                        )
+                        (lam
+                          a_i0
+                          (type)
+                          (lam
+                            b_i0
+                            (type)
+                            (ifix
```
effectfully, line 374: @michaelpj this is what the line you pointed me to produces for the `InterList` example. It's indeed what one'd expect and it doesn't look problematic (i.e. I take back my suspicion that there can be something wrong with it). I do still find it unsettling. How come we have to inline the entire definition of the data type whenever we define a constructor of that data type? Maybe we're missing some interleaving opportunity or I don't know?

Anyways, I'm going to keep thinking about it, but probably after I find a good way of rendering this test using the readable syntax, as I'm not particularly keen on studying 749 lines of S-expression abomination.

plutus-tx-plugin/test/Plugin/Data/recursive/interListConstruct.tplc.golden
```diff
@@ -0,0 +1,749 @@
+(program
```
michaelpj, line 1: so much code :joy: :sob:

plutus-tx-plugin/test/Plugin/Data/recursive/interListConstruct.tplc.golden
michaelpj: I think I convinced myself that it essentially corresponds to the typing rule for `iwrap`, where we do exactly substitute the whole definition of the recursive type for itself.

plutus-tx/testlib/PlutusTx/Test.hs
michaelpj: It's terrible and I hate it. Also not many ideas on how to improve though.

plutus-tx-plugin/test/Plugin/Data/recursive/interListConstruct.tplc.golden
effectfully: I _think_, it indeed plays out like that, but via a different pathway, kinda. With `iwrap` you substitute both the data type _and_ the index -- that's the whole idea behind `ifix`. Here you only substitute the type in the original pattern functor body with `n` type variables, which is very much not was intended with the whole data type compilation machinery... But it works! And it does kinda make sense. My brain hurts.

I think I have an idea on how to reduce the duplication. I'll create a ticket and return to it once I'm done with BLS builtins.

effectfully:
Comments addressed. @michaelpj please merge.

PR #5317, opened 2023-05-11
Enable editorconfig checker and reduce indentation for a few modules

zliu41:
No code change, purely re-formatting.
Enabled editorconfig checker for some modules (which enforces line length <= 100), and reduced indentation to 2 spaces.


michaelpj:
> reduced indentation to 2 spaces

https://github.com/input-output-hk/plutus/blob/master/STYLEGUIDE.adoc#basic-formatting

If you want to use 2 spaces we should have the argument and change the style guide first :)

michaelpj:
Then, let's change the style guide too!

plutus-core/plutus-ir/src/PlutusIR/Compiler/Definitions.hs
```diff
@@ -11,25 +10,28 @@
...
-                                              , lookupTerm
-                                              , lookupOrDefineTerm
-                                              , lookupType
-                                              , lookupOrDefineType
-                                              , lookupConstructors
-                                              , lookupDestructor) where
+module PlutusIR.Compiler.Definitions (
+  DefT,
+  MonadDefs (..),
+  TermDefWithStrictness,
```
effectfully, line 18: So let's do that here...

plutus-core/plutus-ir/src/PlutusIR/Compiler/Definitions.hs
```diff
@@ -73,157 +75,238 @@ mapDefs :: (a -> b) -> DefMap key a -> DefMap key b
...
+  }
 makeLenses ''DefState
 
-newtype DefT key uni fun ann m a = DefT { unDefT :: StateT (DefState key uni fun ann) m a }
-    deriving newtype (Functor, Applicative, Monad, MonadTrans, MM.MFunctor, MonadError e, MonadReader r, MonadQuote, MonadWriter w)
+newtype DefT key uni fun ann m a = DefT {unDefT :: StateT (DefState key uni fun ann) m a}
+  deriving newtype
+    ( Functor
+    , Applicative
+    , Monad
```
effectfully, line 92: ... and that here.

plutus-core/plutus-ir/src/PlutusIR/Subst.hs
```diff
@@ -26,96 +26,111 @@ import Data.Set.Lens (setOf)
...
 funRes :: Lens' (Type tyname uni a) (Type tyname uni a)
 funRes f = \case
-    TyFun a dom cod -> TyFun a dom <$> funRes f cod
-    t               -> f t
+  TyFun a dom cod -> TyFun a dom <$> funRes f cod
+  t               -> f t
 
 -- TODO: these could be Traversals
+
 -- | Get all the term variables in a term.
```
effectfully, line 134: Another example.

effectfully (commented):
Nice try, but you won't be able to sneakily introduce `ormolu` like that :stuck_out_tongue:

zliu41:
> Nice try, but you won't be able to sneakily introduce ormolu like that stuck_out_tongue

It's fourmolu, much less opinionated than ormolu and very configurable, e.g., doesn't mess with import blocks :)

I don't really care about fourmolu and certainly won't mandate it, and am happy to address your comments. The things I definitely want are (1) two space indentation (2) normalizing at least the type signatures (in particular, `foo ::` rather than `foo\n  ::`, which makes it much easier to find the definition of `foo`) - there's no need to have so many variations of type signatures.

PR #5318, opened 2023-05-11
[PlutusIR] [Compile] Replace a lambda with 'unveilDatatype'

effectfully:
This should probably alleviate at least some of the confusion. @michaelpj what you were doing there with creating a lambda and immediately applying it is in fact a pretty common thing in the Plutus IR compiler. So common we've named a paper after it! It's just unveiling. You can even see how the same thing is done for Scott-encoding as well, albeit for a slightly different reason (with Scott-encoding we need unveiling for the types of arguments, with SOPs we need it for the type of the result).

Here's a comparison of compilation output for the `InterList` example:

```
                      SOP    Unveil SOP    Scott
lines of S-exps       749    764           817
occurrences of ifix   6      6             5

```

With unveiling we get slightly larger output, but we're also being consistent and making the dreaded line less dreaded.

Questions still remain though:

1. should the data type compilation machinery in Plutus Core adopt some of this logic?
2. should we try to reduce the amount of duplication? I'm rather skeptical that we're being optimal there
3. would compilation of mutually recursive affect things in a non-trivial way? I remember that incorporating mutually recursive data types wasn't an easy thing even without trying not to produce so many duplicated types
4. do we even care that much about type duplication, given that it all gets erased anyway? I'd argue that yes, because they make it harder to analyze compilation output, hiding optimization opportunities for us and making us reluctant to look into compilation output in the first place (just like the S-expression abomination)

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
@@ -431,7 +431,7 @@ SOPs:
 @
 -}
 mkConstructor :: MonadQuote m => DatatypeCompilationOpts -> PLCRecType uni fun a -> Datatype TyName Name uni (Provenance a) -> Word64 -> m (PIRTerm uni fun a)
-mkConstructor opts dty d@(Datatype ann tv tvs _ constrs) index = do
+mkConstructor opts dty d@(Datatype ann _ tvs _ constrs) index = do
     -- This is inelegant, but it should never fail
     let thisConstr = constrs !! fromIntegral index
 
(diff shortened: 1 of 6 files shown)


effectfully:
Wait, don't look here, I'm trying something out.

effectfully:
> Wait, don't look here, I'm trying something out.

Nah, that was stupid. Ready for review again.

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -452,9 +452,8 @@ mkConstructor opts dty d@(Datatype ann tv tvs _ constrs) index = do
 
             -- The pattern functor with a hole in it
             pf <- mkPatternFunctorBody opts ann d
-            -- ... and with the hole filled in with the datatype type, by making
-            -- a type lambda and immediately applying it
-            let unrolled = TyApp ann (PIR.mkIterTyLam [tv] pf) (getType dty)
+            -- ... and with the hole filled in with the datatype type
+            let unrolled = unveilDatatype (getType dty) d pf
```
effectfully, line 456: The only thing I'm still worried about here if the parameters in `pf` can somehow get out of sync with those of `d` (i.e. `fmap (PIR.mkTyVar ann) tvs`).

And yeah, it's still weird to reconstruct the pattern functor body.

PR #5320, opened 2023-05-11
Update style guide regarding indentation and type signature

zliu41:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/STYLEGUIDE.adoc
+++ b/STYLEGUIDE.adoc
@@ -42,7 +42,7 @@ Avoid lines over 100 characters. This is codified in `.editorconfig`.
 
 Format Haskell source with `stylish-haskell` (config in `.stylish-haskell.yaml`).
 This is enforced in the pre-commit hook and in CI.
-Formatting with `stylish-haskell` must be idempotent.
+Formatting with `stylish-haskell` must be a no-op.
 
 Type signatures should have the double-colon on the same line as the name, separated by one space.
 
--- a/STYLEGUIDE.adoc
+++ b/STYLEGUIDE.adoc
@@ -71,6 +71,19 @@ foo ::
   C
 ----
 
+Good:
+[source,Haskell]
+----
+foo ::
+  forall a b c.
+  ( C1 a
+  , C2 b
+  ) =>
+  A ->
+  B ->
+  C
+----
+
 ==== Reasoning
 
 The indentation is arbitrary. The line length is longer than the “traditional” 80 characters, but anecdotally Haskell code tends toward longer lines, and breaking them up doesn’t help readability much.
--- a/STYLEGUIDE.adoc
+++ b/STYLEGUIDE.adoc
@@ -1,7 +1,7 @@
 = Plutus Haskell style guide
 :toc:
 
-The purpose of this document is to codify various aspects of how we write Haskell code as we contribute to Plutus. 
+The purpose of this document is to codify various aspects of how we write Haskell code as we contribute to Plutus.
 
 The goal is twofold:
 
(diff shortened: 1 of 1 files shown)


effectfully (approved):
I'm not a fan, but I don't hate it either.

STYLEGUIDE.adoc
```diff
@@ -18,14 +18,65 @@ The contents of this document are “recommendations”, that is:
...
+f x = y
+  where
+    y = ...
+----
+
+Avoid lines over 100 characters. This is codified in `.editorconfig`.
 
 Format Haskell source with `stylish-haskell` (config in `.stylish-haskell.yaml`).
+This is enforced in the pre-commit hook and in CI.
+Formatting with `stylish-haskell` must be idempotent.
```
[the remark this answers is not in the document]
zliu41: Yes, no-op is the correct word.

STYLEGUIDE.adoc
```diff
@@ -18,14 +18,65 @@ The contents of this document are “recommendations”, that is:
...
+----
+foo :: A -> B -> C
+----
+
+Good:
+[source,Haskell]
+----
+foo ::
+  A ->
+  B -> C
```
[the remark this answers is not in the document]
zliu41: That's what I would do; I don't want to force it to be written that way but I agree the example should use the one that looks nicer.

STYLEGUIDE.adoc
```diff
@@ -18,14 +18,65 @@ The contents of this document are “recommendations”, that is:
...
+----
 
 ==== Reasoning
 
 The indentation is arbitrary. The line length is longer than the “traditional” 80 characters, but anecdotally Haskell code tends toward longer lines, and breaking them up doesn’t help readability much.
 
+Placing the double-colon on the same line makes it easier to find definitions.
+We don't necessarily have HLS running all the time, and it may not work on every module.
```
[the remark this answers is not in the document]
zliu41: At least for me that takes a few more seconds and key strokes though :)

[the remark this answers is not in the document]
effectfully: True. I personally virtually never do `git grep <definition>`, because I can just use jump-to-definition and I don't need HLS for that. Works even in a GHC Core file for me (it wasn't worth the pain of setting it up though).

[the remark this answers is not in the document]
zliu41: Yes, not to mention when reading PR in a browser etc.

STYLEGUIDE.adoc
```diff
@@ -34,6 +86,22 @@ In the past we tried `ormolu`, but its https://github.com/tweag/ormolu/issues/64
...
+foo :: Eq a => Ord a => a -> a -> a
+----
+
+Good:
+
+[source,Haskell]
+----
+foo :: (Eq a, Ord a) => a -> a -> a
+----
+
```
bezirg, line 104: I think this "curried" constraining has something to do with Haskell inferencing. I don't remember anymore about it, but there was a tweag's video about it sometime ago

bezirg: I mean the above (Bad) is not just the curried version of the below (Good). There is semantic difference in the type.

bezirg: Maybe it's this: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/quantified_constraints.html?highlight=constraints

STYLEGUIDE.adoc
```diff
@@ -18,14 +18,66 @@ The contents of this document are “recommendations”, that is:
...
+[source,Haskell]
+----
+foo :: A -> B -> C
+----
+
+Good:
+[source,Haskell]
+----
+foo ::
```
michaelpj, line 68: Example with constraints and explicit quantification, please

So the thing I think we actually currently do much more (and the one I prefer) is:
```
foo 
   :: forall a b c
   . (C1 a
   , C2 b)
   => A
   -> B
   -> C
```

STYLEGUIDE.adoc
zliu41: > There is semantic difference in the type.

I'm not sure how the semantics are different, but the reason I prefer uncurried constraints is because it's easier to tell, in a big type signature, which are constraints and where the term arguments start.

STYLEGUIDE.adoc
zliu41: Added. I still prefer to have `::` on the first line, so

```haskell
foo ::
  forall a b c.
  ( C1 a
  , C2 b
  ) =>
  A ->
  B ->
  C
```

This is also fine:

```haskell
foo ::
  forall a b c.
  (C1 a, C2 b) =>
  A -> B -> C
```

PR #5321, opened 2023-05-12
[Costing] Fix rendering in budgeting benchmarks

effectfully:
This should probably fix PLT-5935.


kwxm (approved):
Yes, this solves the problem, and the benchmark names look more sensible without "ExMemory " everywhere.  The R modelling code has to extract the numbers from the benchmark names, so I've updated it to deal with the new format and adjusted the benchmark results in `benching.csv` to match the new format.  I've checked that by running `generate-cost-model`and it succeeds, producing JSON output identical to that checked in in `builtinCostModel.json`.

effectfully:
Great, thanks, I'm merging the PR then.

PR #5323, opened 2023-05-15
Inline fully saturated applications in UPLC

zliu41:
This is quite effective, even given the callsite inliner for PIR, likely because more things become fully-applied functions in UPLC.

The implementation mirrors the callsite inliner for PIR.


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -67,11 +68,33 @@ newtype TermEnv name uni fun a =
...
 newtype Subst name uni fun a = Subst { _termEnv :: TermEnv name uni fun a }
     deriving stock (Generic)
-    deriving (Semigroup, Monoid) via (GenericSemigroupMonoid (Subst name uni fun a))
+    deriving newtype (Semigroup, Monoid)
 
 makeLenses ''TermEnv
 makeLenses ''Subst
 
+data VarInfo name uni fun ann = VarInfo
+  { _varBinders :: [name]
```
thealmarty, line 77: Haddock please

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -67,11 +68,33 @@ newtype TermEnv name uni fun a =
...
+  , _varRhs     :: Term name uni fun ann
+  , _varRhsBody :: InlineTerm name uni fun ann
```
thealmarty, line 79: maybe mention `InlineTerm` is needed instead of `Term` and point to the `Dupable` note.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -67,11 +68,33 @@ newtype TermEnv name uni fun a =
...
+  }
+
+makeLenses ''VarInfo
+
+-- | UPLC inliner state
+data S name uni fun a = S
```
thealmarty, line 85: Can we have a more informative name than `S`?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core.hs
```diff
@@ -1,7 +1,27 @@
...
+import Data.Bifunctor
+
+-- | Strips off lambda binders.
+splitLamAbs :: Term name uni fun a -> ([name], Term name uni fun a)
+splitLamAbs = \case
+  LamAbs _ n t -> first (n :) (splitLamAbs t)
+  t            -> ([], t)
+
+-- | Strip off arguments
+splitApplication :: Term name uni fun a -> (Term name uni fun a, [(a, Term name uni fun a)])
```
thealmarty, line 22: The PIR version of [splitApplication](https://github.com/input-output-hk/plutus/blob/5b7848040e932f3344ec31ff131f52371470da72/plutus-core/plutus-ir/src/PlutusIR/Contexts.hs#L27) is in a different module. In general can we be consistent with the (1) naming (2) module it belongs in for PIR and UPLC - we do want to be able to easily sync the two implementation if we need to fix a bug/add some features it would be nice to be able to easily find the counterparts.

thealmarty: Maybe at least add a comment about where to find the corresponding PIR part for each of the functions.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core.hs
```diff
@@ -1,7 +1,27 @@
...
+) where
 
 import UntypedPlutusCore.Core.Instance as Export
 import UntypedPlutusCore.Core.Plated as Export
 import UntypedPlutusCore.Core.Type as Export
+
+import Data.Bifunctor
+
+-- | Strips off lambda binders.
+splitLamAbs :: Term name uni fun a -> ([name], Term name uni fun a)
```
thealmarty, line 16: See below comment, this is similar to [computeArity](https://github.com/input-output-hk/plutus/blob/5b7848040e932f3344ec31ff131f52371470da72/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs#L109) maybe add a comment and rename one of them so they are more clearly doing something similar.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -369,6 +424,69 @@ isPure = go True
...
+      -- The criteria is the same as the criteria for unconditionally inlining `a`,
+      -- since inlining is the same as beta reduction.
+      safeToBetaReduce ::
+        name ->
+        Term name uni fun a ->
+        InlineM name uni fun a Bool
+      safeToBetaReduce a = shouldUnconditionallyInline a rhsBody
+  go rhsBody (info ^. varBinders) args0
+
+-- | This works in the same way as `PlutusIR.Transform.Inline.CallSiteInline.inlineSaturatedApp`.
```
thealmarty, line 465: :+1 yeah something like this but for more of the functions.

thealmarty (approved):
It's great to see the improvements!

PR #5337, opened 2023-05-19
Case-of-case for UPLC where the inner case is an `ifThenElse` application

zliu41:
This achieves the same effect as simplifying `Foo_match (ifThenElse x (Foo_Con_1...) (Foo_Con_2...)) b1 b2`. It is implemented in UPLC because it is easier - we simply need to do the following:

```
case ((force ifThenElse) b (constr t) (constr f)) alts
==>
(force ifThenElse) b (case (constr t) alts) (case (constr f) alts)
```

while doing it in PIR involves the fuss of identifying constructors and destructors, dealing with type instantiations etc.

I went and did this, without first implementing the general case-of-case optimisation and making it a special case. The reason is that this is trivial to implement, while the general case-of-case is potentially much more complex as it may increase the program size if not done carefully.


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Simplify.hs
```diff
@@ -62,5 +66,11 @@ simplifyTerm opts = simplifyNTimes (_soMaxSimplifierIterations opts)
...
         floatDelay
             >=> pure . forceDelayCancel
+            >=> pure . caseOfCase'
             >=> pure . caseReduce
             >=> inline (_soInlineHints opts)
+
+    caseOfCase' :: Term name uni fun a -> Term name uni fun a
+    caseOfCase' = case eqT @fun @DefaultFun of
+        Just Refl -> caseOfCase
+        Nothing   -> id
```
thealmarty, line 76: If you are checking that I think the specific builtin is in the default universe.
```suggestion
    caseOfCase' = case (eqT @fun @DefaultFun, eqT @uni @DefaultUni) of
        (Just Refl, Just Refl) ->caseOfCase
        Nothing   -> id
```

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,33 @@
...
+{- | A special case of case-of-case optimisation: transforms
+
+@
+    case ((force ifThenElse) b (constr t) (constr f)) alts
+@
+
+into
+
+@
+    force ifThenElse b (case (constr t) branches) (case (constr f) branches)
```
thealmarty, line 14: do we need `force` here?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Simplify.hs
zliu41: There's no need to check `eqT @uni @DefaultUni`. It needs to check `fun` because case-of-case only works on `DefaultFun`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
zliu41: I think so. `ifThenElse` is a polymorphic builtin needing a type argument.

thealmarty (approved):

plutus-core/plutus-core/src/PlutusCore/Compiler/Types.hs
```diff
@@ -0,0 +1,10 @@
+{-# LANGUAGE ConstraintKinds #-}
+
+module PlutusCore.Compiler.Types where
+
+import PlutusCore.Builtin
+import PlutusCore.Name
+import PlutusCore.Quote
+
+type Compiling m uni fun name =
+  (ToBuiltinMeaning uni fun, MonadQuote m, HasUnique name TermUnique, Ord name)
```
effectfully, line 10: `Ord name` is weird, do you know where it comes from?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core.hs
```diff
@@ -1,7 +1,27 @@
...
+
+type Arg name uni fun a = (Term name uni fun a, a)
+
+-- | Strip off arguments
+splitApplication :: Term name uni fun a -> (Term name uni fun a, [Arg name uni fun a])
+splitApplication = go []
+  where
+    go acc = \case
+      Apply ann fun arg -> go ((arg, ann) : acc) fun
+      t                 -> (t, acc)
```
effectfully, line 23: @michaelpj seems like you're right and we should resurrect the `View` module.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core.hs
```diff
@@ -1,7 +1,27 @@
...
+
+-- | Make an Apply node from the given function and arguments.
+mkApplication :: Term name uni fun a -> [Arg name uni fun a] -> Term name uni fun a
+mkApplication = foldl' $ \acc (arg, ann) -> Apply ann acc arg
```
effectfully, line 27: This is like

```haskell
mkIterApp
    :: TermLike term tyname name uni fun
    => ann
    -> term ann -- ^ @f@
    -> [term ann] -- ^@[ x0 ... xn ]@
    -> term ann -- ^ @[f x0 ... xn ]@
mkIterApp ann = foldl' (apply ann)
```

except doesn't share a single annotation for application nodes. Maybe we should keep this function in `MkPlc` too? And maybe we should generally do something about annotations being shared/not shared. Like provide some kind of convenient API and allow the user to choose or something idk.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core.hs
```diff
@@ -1,7 +1,27 @@
...
+  mkApplication,
+) where
 
 import UntypedPlutusCore.Core.Instance as Export
 import UntypedPlutusCore.Core.Plated as Export
 import UntypedPlutusCore.Core.Type as Export
+
+import Data.List (foldl')
+
+type Arg name uni fun a = (Term name uni fun a, a)
```
effectfully, line 15: I'd rather make it a `data` with two fields.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Simplify.hs
effectfully, line 76: Not sure how I feel about this. @michaelpj do you have an opinion?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE LambdaCase    #-}
+{-# LANGUAGE TypeFamilies  #-}
+{-# LANGUAGE TypeOperators #-}
+
+{- | A special case of case-of-case optimisation: transforms
+
+@
+    case ((force ifThenElse) b (constr t) (constr f)) alts
```
effectfully, line 8: Should `alts` be `branches`? What do we normally call those BTW? I'd call them `clauses` perhaps, but `branches` seems good too.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,34 @@
...
+import Control.Lens
+
+caseOfCase :: (fun ~ PLC.DefaultFun) => Term name uni fun a -> Term name uni fun a
+caseOfCase = transformOf termSubterms $ \case
+  Case ann scrut alts
+    | ( ite@(Force _ (Builtin _ PLC.IfThenElse))
+        , [cond, (true@Constr{}, trueAnn), (false@Constr{}, falseAnn)]
+        ) <-
+        splitApplication scrut ->
+        mkApplication ite [cond, (Case ann true alts, trueAnn), (Case ann false alts, falseAnn)]
```
effectfully, line 33: Such weird formatting. Is it fourmolu doing that?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,34 @@
...
+import PlutusCore qualified as PLC
+import UntypedPlutusCore.Core
+
+import Control.Lens
+
+caseOfCase :: (fun ~ PLC.DefaultFun) => Term name uni fun a -> Term name uni fun a
+caseOfCase = transformOf termSubterms $ \case
+  Case ann scrut alts
+    | ( ite@(Force _ (Builtin _ PLC.IfThenElse))
+        , [cond, (true@Constr{}, trueAnn), (false@Constr{}, falseAnn)]
```
effectfully, line 30: Wait, so it works for `ifThenElse cond nil (cons x xs)` too (with `nil` and `cons` being SOP-based)? I guess it's even better, but is worth explicit discussion and the naming probably shouldn't be limited to `true` and `false`, as well as the specification above.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core.hs
```diff
@@ -1,7 +1,27 @@
...
+
+-- | Strip off arguments
+splitApplication :: Term name uni fun a -> (Term name uni fun a, [Arg name uni fun a])
+splitApplication = go []
+  where
+    go acc = \case
+      Apply ann fun arg -> go ((arg, ann) : acc) fun
```
effectfully, line 22: I know it's horrible nitpicking, but it's counter-intuitive to me that in constructor `arg` comes after `ann`, but in the tuple it's the other way around.

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
```diff
@@ -31,6 +31,26 @@ interveningLambda = runQuote $ do
...
     pure $ Force () $ Apply () lam arg
 
+caseOfCase1 :: Term Name PLC.DefaultUni PLC.DefaultFun ()
+caseOfCase1 = runQuote $ do
+    b <- freshName "b"
+    let ite = Force () (Builtin () PLC.IfThenElse)
+        true = Constr () 0 []
+        false = Constr () 1 []
+        alts = [mkConstant @Integer () 1, mkConstant @Integer () 2]
+    pure $ Case () (mkApplication ite [(Var () b, ()), (true, ()), (false, ())]) alts
```
effectfully, line 41: And we should test `nil` & `cons` too, plus maybe some other stuff.

plutus-core/untyped-plutus-core/test/Transform/caseOfCase2.plc.golden
```diff
@@ -0,0 +1,5 @@
+(case
+  [ [ [ (force (builtin ifThenElse)) b_0 ] t_1 ] (constr 1) ]
+  (con integer 1)
+  (con integer 2)
+)
```
effectfully, line 5: Worth a comment on why we don't transform

```haskell
case (if b then t else False) of
     True  -> 1
     False -> 2
```

into

```haskell
if b
   then case t of
        True  -> 1
        False -> 2
   else 2
```

I assume this is because we don't want to duplicate branches? (and what we do right now does duplicate them, but since all constants are now, the duplication get removed by a subsequent pass of case-of-known-constructor or whatever).

effectfully (commented):
Looks good, but underspecified to me.

effectfully:
Now do the same for `Data` :sweat_smile:

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
zliu41: GHC uses "alt". @michaelpj do we have a standard name?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core.hs
zliu41: The PIR counterpart is `PlutusIR.Contexts.fillAppContext`. Perhaps they should all use the `TermLike` API in `MkPLC`?

I don't think sharing annotation makes sense. `splitApplication` and `mkApplication` should round-trip.

plutus-core/plutus-core/src/PlutusCore/Compiler/Types.hs
effectfully: Those are supposed to use `UniqueMap` and we could add `UniqueSet` too. Hard for me to imagine why we'd prefer `Map` over a properly `newtype`d `IntMap`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core.hs
effectfully: > I don't think sharing annotation makes sense.

I think it does sometimes, but I agree that not in this case.

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
effectfully: Sure, but it's better to test different cases than only 0-ary one (in general tests should assume at little as possible about the concrete thing that they're supposed to test).

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
effectfully: > The name `true/false` is because those are the true and false branches of `ifThenElse`.

I just realized that they are in fact not, in the general and perfectly realistic case. Think of how `not (if b then True else False) && c` or whatever would look like after `not` is optimized, but before `&&` is optimized (assuming both get inlined).

Also could you please add a test showing how `not (not (if b then True else False))` gets optimized (with `not` inlined)?

plutus-core/plutus-core/src/PlutusCore/Compiler/Types.hs
zliu41: Good point.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
zliu41: This PR is not about optimizing `not . not`. I think that requires the actual case-of-case optimization. This PR implements a special case of case-of-case, where the inner case is `(builtin ifThenElse)`. This special case occurs very frequently, namely every time one writes something similar to `if x < y then ... else ...`. `not . not` is not covered by this special case.

zliu41: Added a test case involving `not . not` (`compiledNotNot`), and it currently isn't fully optimized (but that's not the goal of this PR). See `not-not.uplc-readable.golden`.

effectfully: I think you only this optimization and case-of-known-constructor? Here's how interleaving those two would transform the `not-not` expression:

```
   not (not (if b then True else False))
~> not (if b then not True else not False)
~> not (if b then False else True)
~> if b then not False else not True
~> if b then True else False
```

effectfully: (that wasn't a response to your last message, which I hadn't seen up until after I wrote my comment)

effectfully: >  it currently isn't fully optimized (but that's not the goal of this PR). See `not-not.uplc-readable.golden`.

Look like we should propagate `force` the same way this PR propagates `case`?

plutus-core/untyped-plutus-core/test/Transform/caseOfCase2.plc.golden
zliu41: Added a comment:

```
-- Unless both branches are known constructors, the case-of-case transformation
-- may increase the program size.
```

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
zliu41: > Look like we should propagate force the same way this PR propagates case?

I think we can do that - as long as all the `delay` branches are pure. That is a separate optimization though.

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
zliu41: Added `caseOfCase3`.

michaelpj:
I'm sceptical of doing this in UPLC just because it's easier. We need COKC to make this work well... which is in PIR. And generally this is an optimization that is likely to unlock more optimization opportunities, so I think it's important to do it in PIR. So this seems like false economy to me - if we're just going to do it in PIR later, what's the point of this version?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -119,7 +119,7 @@ data DefaultFun
     | MkPairData
     | MkNilData
     | MkNilPairData
-    deriving stock (Show, Eq, Ord, Enum, Bounded, Generic, Ix)
+    deriving stock (Show, Eq, Ord, Enum, Bounded, Generic, Ix, Typeable)
```
bezirg, line 122: Is `deriving Typeable` redundant nowadays with newer GHC versions?

zliu41:
> I'm sceptical of doing this in UPLC just because it's easier. We need COKC to make this work well... which is in PIR.

`CaseReduce` is COKC in UPLC, no?

michaelpj:
> CaseReduce is COKC in UPLC, no?

It is if constructors are inlined, which they aren't currently.

zliu41:
> It is if constructors are inlined, which they aren't currently.

This is not done in PIR either, is it?

Case-of-case may also interact with CSE, which we also decided to do in UPLC. So I'm in favor of doing all these in UPLC.

michaelpj:
Constructors don't need to be inlined in PIR because we recognize them directly in COKC.

effectfully:
> So I'm in favor of doing all these in UPLC.

Me too, and we could add it to PIR as well if we wanted to. It's a simple optimization, so there isn't much cost (we'll probably end up discussing the design, corner cases etc way longer than actually implementing anything, and design applies to UPLC and PIR). More importantly, UPLC is for everybody, PIR is only for the users of Plutus Tx, and as such there's value in providing that optimization for UPLC, even if we do it for PIR too.

zliu41:
> Constructors don't need to be inlined in PIR because we recognize them directly in COKC.

I'm confused. Do you mean constructors or constructor applications? Constructors can be recognized in PIR, yes, but they can also be recognized in UPLC - they are just `Constr` nodes. Constructor applications like `Just 2` aren't inlined in either PIR or UPLC.

michaelpj:
> I'm confused. Do you mean constructors or constructor applications? Constructors can be recognized in PIR, yes, but they can also be recognized in UPLC - they are just Constr nodes. 

Even in UPLC there will be a _binding_ for the constructor, whose _body_ will contain `Constr` nodes. Unless you inline that binding, you won't see the `Constr`.

I haven't looked at your tests yet, but you should be able to see in some of the tests that come from PIR whether this is happening. Maybe we are sometimes inlining the constructor bindings... but if we don't, this optimization can't fire.

zliu41:
Going to merge this, since it is a strict improvement in terms of script sizes and costs. We can remove it once we implement a similar optimization in PIR.

PR #5340, opened 2023-05-19
Use readable uplc for budget.spec

zliu41:


effectfully (approved):
LGTM. I'll improve readable pretty-printing some time later, so that it doesn't generate that much whitespace.

PR #5344, opened 2023-05-23
PLT-6006: Purge the use of int-cast lib

bezirg:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-core/plutus-core/src/GHC/Natural/Extras.hs
+++ b/plutus-core/plutus-core/src/GHC/Natural/Extras.hs
@@ -3,12 +3,26 @@
 
 module GHC.Natural.Extras (naturalToWord64Maybe, Natural (..)) where
 
-import Data.IntCast (intCastEq)
 import Data.Word (Word64)
 import GHC.Natural
 
--- Note this will only work on 64bit platforms, but that's all we build on
--- so that's okay.
+{- | We only support 64-bit architectures, see Note [Index (Word64) (de)serialized through Natural].
+This implementation is safe w.r.t. cross-compilation, because WORD_SIZE_IN_BITS will point
+to the compiler's target word size when compiling this module.
+-}
 {-# INLINABLE naturalToWord64Maybe #-}
 naturalToWord64Maybe :: Natural -> Maybe Word64
-naturalToWord64Maybe n = intCastEq <$> naturalToWordMaybe n
+naturalToWord64Maybe n =
+#if WORD_SIZE_IN_BITS == 64
+    fromIntegral <$> naturalToWordMaybe n
+#else
+
+{- HLint does not know about WORD_SIZE_IN_BITS, so it calls cpphs to preprocess away these lines;
+cpphs fails on encountering the #error cpp directive (as it should).
+HLint treats this cpphs failure as an hlint failure. Skip HLint then.
+-}
+#ifndef __HLINT__
+#error unsupported WORD_SIZE_IN_BITS. We only support 64-bit architectures.
+#endif
+
+#endif
--- a/plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
@@ -1,4 +1,3 @@
--- editorconfig-checker-disable-file
 -- | The universe used by default and its instances.
 
 {-# OPTIONS -fno-warn-missing-pattern-synonym-signatures #-}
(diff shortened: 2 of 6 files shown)


michaelpj:
You have to pass a benchmark name.

bezirg:
@michaelpj you can take a look.

bezirg:
@michaelpj  The ghc core of naturalToWord64Maybe remains identical.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -323,15 +314,12 @@ internal one -- it's a normal evaluation failure, but unlifting errors have this
...
-instance KnownTypeAst DefaultUni Int64 where
-    toTypeAst _ = toTypeAst $ Proxy @Integer
-
 -- See Note [Integral types as Integer].
-instance HasConstantIn DefaultUni term => MakeKnownIn DefaultUni term Int64 where
+instance HasConstantIn DefaultUni term => MakeKnownIn DefaultUni term Int where
     makeKnown = makeKnown . toInteger
     {-# INLINE makeKnown #-}
 
-instance HasConstantIn DefaultUni term => ReadKnownIn DefaultUni term Int64 where
```
michaelpj, line 334: huh, did we not need the `Int64` thing? I'm confused why we could get rid of this

michaelpj (approved):
Looks fine if the CI is happy!

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
bezirg: We don't use it atm, all our builtins use Int. It was added during int-cast to go through Int64 and fail afterwards

plutus-core/plutus-core/src/GHC/Natural/Extras.hs
```diff
@@ -3,12 +3,22 @@
...
+naturalToWord64Maybe n =
+#if WORD_SIZE_IN_BITS == 64
+    fromIntegral <$> naturalToWordMaybe n
+#else
+
+-- hlint (or cpphs?) has problems parsing the #error cpp directive. Skip hlint then.
+#ifndef __HLINT__
+#error unsupported WORD_SIZE_IN_BITS. We only support 64-bit architectures.
+#endif
+
```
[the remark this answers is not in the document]
bezirg: Unfortunately it is not a parse error for hlint+non-64bit machine

bezirg:
don't merge yet, I am still fixing some stuff

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -340,25 +333,21 @@ instance HasConstantIn DefaultUni term => ReadKnownIn DefaultUni term Int64 wher
...
+#ifndef __HLINT__
+#error unsupported WORD_SIZE_IN_BITS. We only support 64-bit architectures.
+#endif
+#endif
     {-# INLINE readKnown #-}
 
 instance KnownTypeAst DefaultUni Int where
     toTypeAst _ = toTypeAst $ Proxy @Integer
 
--- See Note [Integral types as Integer].
```
michaelpj, line 351: I was suggesting to just keep these instances, but put the whole instance declaration in the CPP, i.e.
```
#if WORD_SIZE_IN_BITS == 64
instance HasConstantIn DefaultUni term => MakeKnownIn DefaultUni term Int where ...
#endif 
```
That way we can keep the `Int64` instance free of conditional logic. I think that's kind of nice because AFAICT the `Int64` instance is correct regardless of the word size, which is sort of nice. We also don't need the `#error`, since it will simply mean that any other uses of `Int` in unlifting will fail to compile because the instance won't exist.

I'm partly just trying to think what expresses the truth of the situation as clearly as possible. I _think_ the main thing we want to avoid is casting from `Int` to `Int64` or vice versa in builtins unless they're identical. And having a conditionally available instance for `Int` kind of expresses that.

Anyway, I think this is fine.

michaelpj (approved):

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -349,15 +347,19 @@ instance KnownTypeAst DefaultUni Int where
...
 
 -- See Note [Integral types as Integer].
+#if WORD_SIZE_IN_BITS == 64
 instance HasConstantIn DefaultUni term => MakeKnownIn DefaultUni term Int where
-    -- This could safely just be toInteger, but this way is more explicit and it'll
-    -- turn into the same thing anyway.
-    makeKnown = makeKnown . intCastEq @Int @Int64
+    -- Constraining this Int-to-Integer conversion instance to 64-bit arch is unnecessary,
+    -- because the conversion should work on any arch. However, we constrain it for simplicity.
+    makeKnown = makeKnown . toInteger
```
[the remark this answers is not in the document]
bezirg: I changed it to go through Int64, as it was before.

bezirg:
@michaelpj ghc core is identical to before.

PR #5348, opened 2023-05-23
Update `mkIterApp`, `mkIterInst` and `mkIterTyApp`

zliu41:
Changed `PlutusCore.MkPlc.mkIterApp`, `PlutusCore.MkPlc.mkIterInst` and `PlutusCore.MkPlc.mkIterTyApp`  to require an annotation to be provided for each argument. 

Also added `PlutusCore.MkPlc.mkIterAppNoAnn`, `PlutusCore.MkPlc.mkIterInstNoAnn` and `PlutusCore.MkPlc.mkIterTyAppNoAnn`.

### DIFF
--- a/plutus-core/plutus-core/examples/PlutusCore/Examples/Data/Vec.hs
+++ b/plutus-core/plutus-core/examples/PlutusCore/Examples/Data/Vec.hs
@@ -174,9 +174,9 @@ churchCons = runQuote $ do
         . TyAbs () r (KindArrow () natK $ Type ())
         . LamAbs () f stepFun
         . LamAbs () z (TyApp () (TyVar () r) zeroT)
-        $ mkIterApp () (TyInst () (Var () f) $ TyVar () n)
+        $ mkIterAppNoAnn (TyInst () (Var () f) $ TyVar () n)
             [ Var () x
-            , mkIterApp () (TyInst () (Var () xs) $ TyVar () r)
+            , mkIterAppNoAnn (TyInst () (Var () xs) $ TyVar () r)
                 [ Var () f
                 , Var () z
                 ]
@@ -214,9 +214,9 @@ churchConcat = runQuote $ do
         . TyAbs () r (KindArrow () natK $ Type ())
         . LamAbs () f stepFun
         . LamAbs () z (TyApp () (TyVar () r) zeroT)
-        $ mkIterApp () (TyInst () (Var () xs) . TyLam () p natK $ TyApp () (TyVar () r) plusTPM)
+        $ mkIterAppNoAnn (TyInst () (Var () xs) . TyLam () p natK $ TyApp () (TyVar () r) plusTPM)
             [ TyAbs () p natK $ TyInst () (Var () f) plusTPM
-            , mkIterApp () (TyInst () (Var () ys) $ TyVar () r)
+            , mkIterAppNoAnn (TyInst () (Var () ys) $ TyVar () r)
                 [ Var () f
                 , Var () z
                 ]
@@ -307,7 +307,7 @@ scottCons = runQuote $ do
         . TyAbs () r (KindArrow () natK $ Type ())
         . LamAbs () f stepFun
         . LamAbs () z (TyApp () (TyVar () r) zeroT)
-        $ mkIterApp () (TyInst () (Var () f) $ TyVar () n)
+        $ mkIterAppNoAnn (TyInst () (Var () f) $ TyVar () n)
             [ Var () x
             , Var () xs
             ]
@@ -334,7 +334,7 @@ scottHead = runQuote $ do
         . TyAbs () a (Type ())
         . TyAbs () n natK
         . LamAbs () xs (mkIterTyApp () scottVec [TyVar () a, TyApp () succT $ TyVar () n])
-        $ mkIterApp ()
+        $ mkIterAppNoAnn
             (   TyInst () (Unwrap () $ Var () xs)
               . TyLam () p natK
                 $ mkIterTyApp () (TyVar () p)
@@ -376,7 +376,7 @@ scottSumHeadsOr0 = runQuote $ do
         . TyAbs () n natK
         . LamAbs () xs (vecInteger $ TyVar () n)
         . LamAbs () ys (vecInteger $ TyVar () n)
-        $ mkIterApp ()
+        $ mkIterAppNoAnn
             (   TyInst () (Unwrap () $ Var () xs)
               . TyLam () p natK
               . TyFun () (TyFun () (vecInteger $ TyVar () n) $ vecInteger (TyVar () p))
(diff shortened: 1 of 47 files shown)


zliu41:
I think `fillAppContext` should also live in `MkPlc.hs` and be made polymorphic in `Term`, but I'm not doing that in this PR.

plutus-core/plutus-core/src/PlutusCore/MkPlc.hs
```diff
@@ -210,23 +213,39 @@ mkImmediateTyAbs
...
     :: TermLike term tyname name uni fun
-    => ann
-    -> term ann -- ^ @f@
-    -> [term ann] -- ^@[ x0 ... xn ]@
-    -> term ann -- ^ @[f x0 ... xn ]@
-mkIterApp ann = foldl' (apply ann)
+    => term ann
+    -> [(ann, term ann)]
+    -> term ann
+mkIterApp = foldl' (uncurry . flip apply)
```
effectfully, line 223: Nit: at that point I'd add some names, `uncurry`, `.` and `flip` together is a bit too much even for a point-free abuser like myself.

plutus-core/plutus-core/examples/PlutusCore/Examples/Data/Vec.hs
```diff
@@ -334,7 +334,7 @@ scottHead = runQuote $ do
         . TyAbs () a (Type ())
         . TyAbs () n natK
         . LamAbs () xs (mkIterTyApp () scottVec [TyVar () a, TyApp () succT $ TyVar () n])
-        $ mkIterApp ()
+        $ mkIterAppNoAnn
```
effectfully, line 337: `mkIterTyApp` shares the same annotation, `mkIterApp` doesn't and we have a separate `mkIterAppNoAnn`. Looks kinda messy. It also messes with one's reflexes: when creating an AST, one uses an explicit annotation for virtually every function/constructor except `mkIterApp` and `mkIterInst` all of a sudden.

effectfully (approved):
This does look like an improvement for me, but I'm feeling something a bit unsettling deep inside my soul about the new API.

plutus-core/plutus-core/src/PlutusCore/MkPlc.hs
zliu41: Understandable. I find it easier to read though :smile:

plutus-core/plutus-core/examples/PlutusCore/Examples/Data/Vec.hs
zliu41: Indeed, the same change should be made to `mkIterTyApp`, which I overlooked.

zliu41 (commented):
> This does look like an improvement for me, but I'm feeling something a bit unsettling deep inside my soul about the new API.

I think sharing a non-unit annotation when constructing nested Apply or TyInst nodes is, in most cases, wrong (e.g., what the parsers are doing currently).

plutus-core/plutus-core/src/PlutusCore/MkPlc.hs
michaelpj: > I find it easier to read though

I'm kind of amazed if this is true :D

plutus-core/plutus-core/src/PlutusCore/Parser.hs
```diff
@@ -46,7 +47,8 @@ lamTerm = withSpan $ \sp ->
 
 appTerm :: Parser PTerm
 appTerm = withSpan $ \sp ->
-    inBrackets $ mkIterApp sp <$> term <*> some term
+    -- TODO: should not use the same `sp` for all arguments.
```
michaelpj, line 50: Not clear to me what it _should_ be. Do you want something like this?
```
[ f a b c ]
--> 
[[ f a ](0:5) b](0:7) c ](0:11)
```

plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs
```diff
@@ -590,7 +591,10 @@ compileDatatype r body d = do
         vars = fmap PIR.defVar constrDefs ++ [ PIR.defVar destrDef ]
         vals = fmap PIR.defVal constrDefs ++ [ PIR.defVal destrDef ]
     -- See note [Abstract data types]
-    pure $ PIR.mkIterApp p (PIR.mkIterInst p (PIR.mkIterTyAbs tyVars (PIR.mkIterLamAbs vars body)) tys) vals
+    pure $
+      PIR.mkIterApp
+        (PIR.mkIterInst (PIR.mkIterTyAbs tyVars (PIR.mkIterLamAbs vars body)) ((p,) <$> tys))
```
michaelpj, line 596: I mean, this stuff reads much worse now but :shrug:

michaelpj (commented):
I'm not entirely convinced this is an improvement. We could have fixed the parsers without changing all the rest, and it seems to me that in all the other cases which were changed we are _not_ going to do anything different than using a single annotation for all of it. So we've made 90% of the usages worse for no benefit, when we could just have special-cased the ones that did benefit.

plutus-core/plutus-core/src/PlutusCore/MkPlc.hs
zliu41: I mean, `uncurry` and `flip` are just restructuring the arguments and can be ignored, so it's really just `fold' apply`.

plutus-core/plutus-core/src/PlutusCore/Parser.hs
zliu41: Yes, that's what I had in mind.

zliu41 (commented):
> We could have fixed the parsers without changing all the rest

I need the new `mkIterApp` in #5323. We can keep both versions, but I don't know how to name them. Am not a huge fan of either `mkIterApp'` or `mkIterAppSharedAnn`.

kwxm:
@zliu41 Should we be doing the same thing for the UPLC of `mkIterApp` [here](https://github.com/input-output-hk/plutus/blob/e2f664e52a5cfef99bf863de945e9f03d505b071/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/MkUPlc.hs#L16)?

zliu41:
> Should we be doing the same thing for the [UPLC version of mkIterApp](https://github.com/input-output-hk/plutus/blob/e2f664e52a5cfef99bf863de945e9f03d505b071/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/MkUPlc.hs#L16)?

Yes, it should be deleted. The other `mkIterApp` is polymorphic so we don't need the UPLC version.

zliu41:
It is in fact deleted in https://github.com/input-output-hk/plutus/pull/5323.

PR #5349, opened 2023-05-23
Add nofib and lists to the list of longitudinal benchmarks

zeme-wana:


michaelpj (approved):

michaelpj:
Hmm, this was maybe a mistake. Now the job takes >2 hrs rather than 1hr before, which I think may just make it totally clogged, and makes it harder for us to run manual benchmarks :thinking:

thealmarty:
I requested this. It's more helpful to have more longitudinal benchmarks. We can pick a time when it's more "quiet" for this?

michaelpj:
I think adding `nofib` is good, but `lists` is really long and not that interesting, so I removed it.

PR #5353, opened 2023-05-24
Remove zipped event dump right after unzipping in s3-sync-unzip.hs

zeme-wana:
In an attempt to fix an 'out of disk space' error in script-evaluation-test
See: https://github.com/input-output-hk/plutus/actions/runs/5067207532/jobs/9097996057


bezirg:
can we switch also to zstd perhaps? bzip2 is incredibly slow and outdated. Even lzma,xz are better than bzip2 i think

PR #5354, opened 2023-05-24
Avoid allocating VLamAbs and VDelay sometimes

michaelpj:
See note in the code.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


michaelpj:
I'm surprised this is so bad, but it's certainly not an improvement. Even just doing it for lambdas seems to have marginal benefit, whereas I'd expected it to be moderate, since it should often fire when doing case analysis.

The Core looks like what I expect: stuff is inlined where I want it and we do avoid the allocation.

michaelpj:
Might be interesting to @effectfully but I think we probably don't want to do this atm.

PR #5356, opened 2023-05-24
[Builtins] [Evaluation] Retain 'cost' in a different way

effectfully:
Our current approach to creating thunks for individual `BuiltinRuntime`s and sharing them during evaluation turned out to have a flaw: it randomly duplicates `BuiltinRuntime`s for a few builtins (while doing the right thing for the majority of them).

For example, in the current Core we have this for `AddInteger` :

```
    AddInteger -> lvl366_sbwr4;
```

as well as this:

```
    AddInteger ->
      let! { __DEFAULT ~ <- lvl366_sbwr4 } in jump go5_i86oX ys_i86p3;
```

and both point to the same thunk, `lvl366_sbwr4`.

But for `IndexByteString` the two thunks are different, it's 

```
    IndexByteString -> lvl370_sbwBp;
```

vs

```
    IndexByteString ->
      let! { __DEFAULT ~ <- lvl367_sbx56 } in jump go5_i86oX ys_i86p3;
```

i.e. `lvl370_sbwBp` vs `lvl367_sbx56`. Both bindings have to point to the same implementation, but duplicating the implementation still results in extra noise and confusion as to why one of them even exists, if it gets forced and dropped on the floor immediately afterwards.

This isn't a big deal, since we don't get any wrong behavior from `BuiltinRuntime` duplication or any noticeable slowdown, but it still makes sense to fix the problem if there's a way. This PR implements an alternative approach, which looks like it results in the same `BuiltinRuntime` sharing while being both simpler and much more reliable.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
@@ -382,21 +382,10 @@ toBuiltinsRuntime
     => BuiltinVersion fun -> cost -> BuiltinsRuntime fun val
 toBuiltinsRuntime ver cost =
     let runtime = BuiltinsRuntime $ toBuiltinRuntime cost . inline toBuiltinMeaning ver
-        {-# INLINE runtime #-}
-       -- Force each 'BuiltinRuntime' to WHNF. Inlining 'force' manually, since it doesn't have an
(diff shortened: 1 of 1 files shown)


effectfully:
^ noise, which is great. Now let's try to remove the pragma to make that the pragma does help.

michaelpj (approved):

effectfully:
^ lies.

Ready for review.

michaelpj (approved):

PR #5360, opened 2023-05-25
Add `commuteConst` to simplifier pass.

thealmarty:
Doing this in PIR in case we want to do CSE in PIR.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteConst.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteConst.hs
@@ -33,25 +33,23 @@ the one that will benefit the most. Plutonomy only commutes `EqualsInteger`.
 -}
 
 commuteConstDefault ::
-    forall m tyname name uni a.
-    ( PLC.MonadQuote m
-    ) => Term tyname name uni PLC.DefaultFun a ->
-    m (Term tyname name uni PLC.DefaultFun a)
+    forall tyname name uni a.
+    Term tyname name uni PLC.DefaultFun a ->
+    Term tyname name uni PLC.DefaultFun a
 commuteConstDefault (Apply ann (Builtin annB PLC.EqualsInteger) (Apply ann1 x y@(Constant{}))) =
-    pure $ Apply ann (Builtin annB PLC.EqualsInteger) (Apply ann1 y x)
+    Apply ann (Builtin annB PLC.EqualsInteger) (Apply ann1 y x)
 commuteConstDefault (Apply ann (Builtin annB PLC.EqualsByteString) (Apply ann1 x y@(Constant{}))) =
-    pure $ Apply ann (Builtin annB PLC.EqualsByteString) (Apply ann1 y x)
+    Apply ann (Builtin annB PLC.EqualsByteString) (Apply ann1 y x)
 commuteConstDefault (Apply ann (Builtin annB PLC.EqualsString) (Apply ann1 x y@(Constant{}))) =
-    pure $ Apply ann (Builtin annB PLC.EqualsString) (Apply ann1 y x)
+    Apply ann (Builtin annB PLC.EqualsString) (Apply ann1 y x)
 commuteConstDefault (Apply ann (Builtin annB PLC.AddInteger) (Apply ann1 x y@(Constant{}))) =
-    pure $ Apply ann (Builtin annB PLC.AddInteger) (Apply ann1 y x)
+    Apply ann (Builtin annB PLC.AddInteger) (Apply ann1 y x)
 commuteConstDefault (Apply ann (Builtin annB PLC.MultiplyInteger) (Apply ann1 x y@(Constant{}))) =
-    pure $ Apply ann (Builtin annB PLC.MultiplyInteger) (Apply ann1 y x)
-commuteConstDefault tm = pure tm
+    Apply ann (Builtin annB PLC.MultiplyInteger) (Apply ann1 y x)
+commuteConstDefault tm = tm
 
-commuteConst :: forall m tyname name uni fun a.
-    ( PLC.MonadQuote m, Typeable fun
-    ) => Term tyname name uni fun a -> m (Term tyname name uni fun a)
+commuteConst :: forall tyname name uni fun a. Typeable fun =>
+    Term tyname name uni fun a -> Term tyname name uni fun a
 commuteConst = case eqT @fun @PLC.DefaultFun of
     Just Refl -> commuteConstDefault
-    Nothing   -> \x -> pure x
+    Nothing   -> id
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteConst.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteConst.hs
@@ -1,0 +1,57 @@
+{-# LANGUAGE GADTs            #-}
+{-# LANGUAGE TypeApplications #-}
+
+module PlutusIR.Transform.CommuteConst (commuteConst) where
+
+import Data.Typeable
+import PlutusCore qualified as PLC
+import PlutusIR
+
+{- | Commute such that constants are the first arguments. Consider:
+
+(1)    equalsInteger 1 x
+
+(2)    equalsInteger x 1
+
+We have unary application, so these are two partial applications:
+
+(1)    (equalsInteger 1) x
+
+(2)    (equalsInteger x) 1
+
+With (1), we can share the `equalsInteger 1` node, and it will be the same across any place where
+we do this.
+
+With (2), both the nodes here include x, which is a variable that will likely be different in other
+invocations of `equalsInteger`. So the second one is harder to share, which is worse for CSE.
+
+So commuting `equalsInteger` so that it has the constant first both a) makes various occurrences of
+`equalsInteger` more likely to look similar, and b) gives us a maximally-shareable node for CSE.
+
+This applies to any commutative builtin function, although we might expect that `equalsInteger` is
+the one that will benefit the most. Plutonomy only commutes `EqualsInteger`.
+-}
+
+commuteConstDefault ::
+    forall m tyname name uni a.
+    ( PLC.MonadQuote m
+    ) => Term tyname name uni PLC.DefaultFun a ->
+    m (Term tyname name uni PLC.DefaultFun a)
+commuteConstDefault (Apply ann (Builtin annB PLC.EqualsInteger) (Apply ann1 x y@(Constant{}))) =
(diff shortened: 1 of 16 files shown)


plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteConst.hs
```diff
@@ -0,0 +1,57 @@
...
+    pure $ Apply ann (Builtin annB PLC.EqualsByteString) (Apply ann1 y x)
+commuteConstDefault (Apply ann (Builtin annB PLC.EqualsString) (Apply ann1 x y@(Constant{}))) =
+    pure $ Apply ann (Builtin annB PLC.EqualsString) (Apply ann1 y x)
+commuteConstDefault (Apply ann (Builtin annB PLC.AddInteger) (Apply ann1 x y@(Constant{}))) =
+    pure $ Apply ann (Builtin annB PLC.AddInteger) (Apply ann1 y x)
+commuteConstDefault (Apply ann (Builtin annB PLC.MultiplyInteger) (Apply ann1 x y@(Constant{}))) =
+    pure $ Apply ann (Builtin annB PLC.MultiplyInteger) (Apply ann1 y x)
+commuteConstDefault tm = pure tm
+
+commuteConst :: forall m tyname name uni fun a.
```
zliu41, line 52: `commuteConst` sounds weird to me - it's not the constants that commute, it's the functions that commute.

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteConst.hs
```diff
@@ -0,0 +1,57 @@
...
+This applies to any commutative builtin function, although we might expect that `equalsInteger` is
+the one that will benefit the most. Plutonomy only commutes `EqualsInteger`.
+-}
+
+commuteConstDefault ::
+    forall m tyname name uni a.
+    ( PLC.MonadQuote m
+    ) => Term tyname name uni PLC.DefaultFun a ->
+    m (Term tyname name uni PLC.DefaultFun a)
+commuteConstDefault (Apply ann (Builtin annB PLC.EqualsInteger) (Apply ann1 x y@(Constant{}))) =
```
zliu41, line 40: This is incorrect - you are pattern matching on `(==) (x y)`, but you should pattern match on `((==) x) y`. Please add some tests.

You can also simplify this by using a helper function `isCommutative :: DefaultFun -> Bool`.

I also suggest checking that `x` is not a constant. Otherwise you'd be switching `x` and `y` back and forth which doesn't make much sense.

zliu41 (changes requested):

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteConst.hs
thealmarty: Renamed to `commuteFunWithConst`. Feel free to suggest.

plutus-core/plutus-ir/test/recursion/factorial.golden
```diff
@@ -15,7 +15,7 @@
                   [
                     [
                       (force (builtin ifThenElse))
-                      [ [ (builtin equalsInteger) x_514 ] (con integer 0) ]
+                      [ [ (builtin equalsInteger) (con integer 0) ] x_514 ]
```
thealmarty, line 18: It's cool that it caught this but I'm not sure how. Looking at the code, I don't think that this test should have gone through the simplifier.

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteFnWithConst.hs
```diff
@@ -0,0 +1,130 @@
...
+
+commuteFnWithConst :: forall tyname name uni fun a. Typeable fun =>
+    Term tyname name uni fun a -> Term tyname name uni fun a
+commuteFnWithConst = case eqT @fun @DefaultFun of
+    Just Refl -> commuteDefaultFun
+    Nothing   -> id
+
+-- | Returns whether a `DefaultFun` is commutative with `Constant`'s as arguments. Not using
+-- catchall to make sure that this function catches newly added `DefaultFun`.
+isCommutativeWithConstant :: DefaultFun -> Bool
```
thealmarty, line 66: I could see an argument to just filter for commutative functions, and let the `isConstant` function filter out the non-constants. Lmk which you think is cleaner.

michaelpj: I don't understand what it's currently saying. What does it meant to be "commutative with constants as arguments"? Either the function is commutative or it isn't :thinking: So... yeah, I'd just have an `isCommutative` function.

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteFnWithConst.hs
```diff
@@ -0,0 +1,130 @@
...
+isCommutativeWithConstant = \case
+  AddInteger                      -> False
```
michaelpj, line 68:
```suggestion
  AddInteger                      -> True
```

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteFnWithConst.hs
```diff
@@ -0,0 +1,130 @@
...
+  MapData                         -> False
+  ListData                        -> False
+  IData                           -> False
+  BData                           -> False
+  UnConstrData                    -> False
+  UnMapData                       -> False
+  UnListData                      -> False
+  UnIData                         -> False
+  UnBData                         -> False
+  EqualsData                      -> False -- doesn't take constant
```
michaelpj, line 126: I don't understand what you're saying here. It does take constants. And it is commutative.

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteFnWithConst.hs
```diff
@@ -0,0 +1,130 @@
...
+  SerialiseData                   -> False
+  MkPairData                      -> False
+  MkNilData                       -> False
+  MkNilPairData                   -> False
```
michaelpj, line 130: I'm unsure, but I think given we're listing some exceptions, this could reasonably have a `_ -> False` case.

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteFnWithConst.hs
```diff
@@ -0,0 +1,130 @@
...
+  , commuteDefaultFun)
+  where
+
+import Control.Lens (over)
+import Data.Typeable (Typeable, eqT)
+import PlutusCore.Default
+import PlutusIR.Core.Plated (termSubterms)
+import PlutusIR.Core.Type (Term (Apply, Builtin, Constant))
+
+{- | Commute such that constants are the first arguments. Consider:
```
michaelpj, line 16: This doc comment attaches to `isConstant`, which I suspect isn't what you wanted. Do you want to put it as the module haddock?

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteFnWithConst.hs
```diff
@@ -0,0 +1,130 @@
...
+isConstant _          = False
+
+commuteDefaultFun ::
+    forall tyname name uni a.
+    Term tyname name uni DefaultFun a ->
+    Term tyname name uni DefaultFun a
+commuteDefaultFun = over termSubterms commuteDefaultFun . localCommute
+  where
+    localCommute tm@(Apply ann (Apply ann1 (Builtin annB fun) x) y) =
+      case (isCommutativeWithConstant fun, isConstant x, isConstant y) of
```
michaelpj, line 53: Use guards!

michaelpj: And for the constraint on `y` you can just inline the pattern into the one on the line above.

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteFnWithConst.hs
```diff
@@ -0,0 +1,130 @@
...
+        (True, False, True) -> Apply ann (Apply ann1 (Builtin annB fun) y) x
+        _                   -> tm
+    localCommute tm = tm
+
+commuteFnWithConst :: forall tyname name uni fun a. Typeable fun =>
+    Term tyname name uni fun a -> Term tyname name uni fun a
+commuteFnWithConst = case eqT @fun @DefaultFun of
```
michaelpj, line 60: I thought we decided we were going to take it as a parameter like `InlineHints`?

michaelpj: I don't love this at all: it just means it can't work for anything except `DefaultFun`.

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteFnWithConst.hs
zliu41: Yeah, this function is already `isCommutative`.

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteFnWithConst.hs
thealmarty: What I'm trying to do is to force us to add a case here whenever we add a new `DefaultFun`. Otherwise, any new builtins will just be false, because we're unlikely to remember this unless the compiler reminds us. Of course, not remembering is not the end of the world because we just won't commute, but I think it makes sense to add any upcoming commutative functions in here.

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteFnWithConst.hs
thealmarty: Yeah it's in this [PR](https://github.com/input-output-hk/plutus/pull/5337/files#r1199188364) too. Although I guess that's not merged. I'm fine with either way really. I have more preference for things to work similarly in the same code base. So it comes down to whether you want to change your #5337 Ziyang!

plutus-core/plutus-ir/src/PlutusIR/Transform/CommuteFnWithConst.hs
thealmarty: I guess I just thought in my head that `Data` isn't a "real" constant because anything can be in it.

zliu41 (approved):

PR #5361, opened 2023-05-25
Fix `firstEffectfulTerm`

zliu41:
In the `Apply` case (and similarly the `Constr` case), `firstEffectfulTerm` only looks at the function, not the argument, which is both inefficient and incorrect.

An example where it is inefficient:

```
let !a = error "a" in (\x -> x) a
```
The inliner is unable to inline `a`, because `firstEffectfulTerm ((\x -> x) a)` returns `Nothing`, but it should.

An example where it is incorrect:

```
let !a = error "a" in
let !b = (\x -> x) (error "b")
 in a b
```

The inliner incorrectly inlines `a`, changing the semantics of the program, because `firstEffectfulTerm (let !b = (\x -> x) (error "b") in a b)` returns `Just a`, which is, in turn, because `firstEffectfulTerm ((\x -> x) (error "b"))` returns `Nothing`.

This PR fixes it. There are some cost increases because previously it was inlining things that shouldn't have been inlined. That said, we should have a non-conservative mode that allows `a` to be inlined in the second example.


michaelpj:
> because firstEffectfulTerm (let !b = (\x -> x) (error "b") in a b) returns Just a, which is, in turn, because firstEffectfulTerm ((\x -> x) (error "b")) returns Nothing.

Bad past Michael! Conflating "we know there is no effectful term" with "we don't know whether there is an effectful term"!

michaelpj (approved):

PR #5364, opened 2023-05-29
[Triage] Help out with 'loadFromFile'

### LINKED ISSUE #5331: `PlutusTx.loadFromFile` can't load validator from file
### Summary

I have the following code:
```haskell
import Plutus.Script.Utils.V2.Typed.Scripts.Validators (UntypedValidator)
import PlutusTx qualified

compiledMyScript :: PlutusTx.CompiledCode (MyScriptParams -> UntypedValidator)
compiledMyScript = $$(PlutusTx.loadFromFile "data/script.txt")
```

And this is the contents of the `data/script.txt` file (ie. the validator's CBOR encoded in hex):
```
4e4d01000033222220051200120011
```

When I try and use a validator loaded from file it fails with the following type of error:
```
Exception: Failed to deserialise our own program! This is a bug, please report it. Caused by: TooMuchSpace (0x000000010124969a,S {currPtr = 0x00000001012419d7, usedBits = 6})
```

I've also tried placing PlutusCore to the file:
```
(program
  2.0.0
  (lam i_0 [ [ (builtin addInteger) (con integer 16) ] i_0 ])
)
```

But it also failed with the same error.

Both of the files above were textual but I've also tried binary representation of the PlutusCore posted above, the following are the bits in textual form, but I've provided the binary data in a binary file:
```
00000010 00000000 00000000 00100011
00110111 00000000 10010000 00010000
00000000 00001001
```

It failed with a different error:
```
Exception: Failed to deserialise our own program! This is a bug, please report it. Caused by: BadEncoding (0x000000010b5aa2fa,S {currPtr = 0x000000010b5aa2f3, usedBits = 7}) "usedBits /= 0"
```

Described above - trying to find the correct use for the function `PlutusTx.loadFromFile`. I haven't managed to find correct format that the function is expecting.

### Actual Result

Results with a runtime error.

### Expected Result

I'm expecting such compiled code would behave the same as I would load a classical Haskell written PlutusTx code.

I'm curious what format does it expect and in case it's something different how could I get that from the CBOR hex, or if there's other function that would create `CompiledCode` from CBOR hex.

_No response_

### System info

Plutus revision: f003b096dbc5abca9dbb74fbff847914dd39fa6b

effectfully:
This is to help resolve #5331.

@KristianBalaj please take a look.


effectfully:
It helped.

PR #5231, opened 2023-03-24
Kwxm/bls12-381/prototype (PLT-192, PLT-1557, PLT-1554, etc).

kwxm:
## Introduction

This PR adds new built-in types and functions to Plutus Core and PlutusTx for pairings over BLS12-381, as specified in [CIP-0381](https://cips.cardano.org/cips/cip0381/).  These depend on Haskell bindings to the [blst C library](https://github.com/supranational/blst) which are currently the subject of an [open PR](https://github.com/input-output-hk/cardano-base/pull/266) in `cardano-base`.  In order not to depend on a branch, the code for these has been coped into `plutus-core/cbits` and `plutus-core/src/plutus-core/Crypto/External`.  We should be able to merge this PR as-is, but ideally we'd wait until the `cardano-base` PR is merged and remove the copied files from `plutus-core`.

## Details
The PR introduces three new PLC types:
  * `bls12_381_G1_element`
  * `bls12_381_G2_element`
  * `bls12_381_MlResult`

and seventeen new built-in functions:
```
bls12_381_G1_add                : [ bls12_381_G1_element, bls12_381_G1_element ] -> bls12_381_G1_element
bls12_381_G1_neg                : [ bls12_381_G1_element ] -> bls12_381_G1_element
bls12_381_G1_scalarMul          : [ integer, bls12_381_G1_element ] -> bls12_381_G1_element
bls12_381_G1_equal              : [ bls12_381_G1_element, bls12_381_G1_element ] -> bool
bls12_381_G1_hashToGroup        : [ bytestring ] -> bls12_381_G1_element
bls12_381_G1_compress           : [ bls12_381_G1_element ] -> bytestring
bls12_381_G1_uncompress         : [ bytestring ] -> bls12_381_G1_element

bls12_381_G2_add                : [ bls12_381_G2_element, bls12_381_G2_element ] -> bls12_381_G2_element
bls12_381_G2_neg                : [ bls12_381_G2_element ] -> bls12_381_G2_element
bls12_381_G2_scalarMul          : [ integer, bls12_381_G2_element ] -> bls12_381_G2_element
bls12_381_G2_equal              : [ bls12_381_G2_element, bls12_381_G2_element ] -> bool
bls12_381_G2_hashToGroup        : [ bytestring ] -> bls12_381_G2_element
bls12_381_G2_compress           : [ bls12_381_G2_element ] -> bytestring
bls12_381_G2_uncompress         : [ bytestring ] -> bls12_381_G2_element

[...]

PR #5367, opened 2023-06-01
Kwxm/bls12 381/minor fixes

kwxm:
This is just responding to a couple of small things raised in the PR comments; the main branch got merged while I was fixing these.

### DIFF
--- a/plutus-benchmark/bls-benchmarks/src/PlutusBenchmark/BLS12_381/Common.hs
+++ b/plutus-benchmark/bls-benchmarks/src/PlutusBenchmark/BLS12_381/Common.hs
@@ -36,7 +36,7 @@ import Hedgehog.Internal.Gen qualified as G
 import Hedgehog.Internal.Range qualified as R
 import System.IO.Unsafe (unsafePerformIO)
 
-import Prelude (fromIntegral, ($!))
+import Prelude (fromIntegral)
 
 -------------------------------- PLC stuff--------------------------------
 
--- a/plutus-benchmark/bls-benchmarks/src/PlutusBenchmark/BLS12_381/Common.hs
+++ b/plutus-benchmark/bls-benchmarks/src/PlutusBenchmark/BLS12_381/Common.hs
@@ -1,4 +1,5 @@
 -- editorconfig-checker-disable-file
+{-# LANGUAGE BangPatterns      #-}
 {-# LANGUAGE DataKinds         #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE TemplateHaskell   #-}
@@ -35,7 +36,7 @@ import Hedgehog.Internal.Gen qualified as G
 import Hedgehog.Internal.Range qualified as R
 import System.IO.Unsafe (unsafePerformIO)
 
-import Prelude (fromIntegral)
+import Prelude (fromIntegral, ($!))
 
 -------------------------------- PLC stuff--------------------------------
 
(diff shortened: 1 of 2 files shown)


plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -258,14 +258,35 @@ instance ExMemoryUsage Data where
             I n        -> memoryUsage n
             B b        -> memoryUsage b
 
+
```
kwxm, line 261: This is in response to [this comment](https://github.com/input-output-hk/plutus/pull/5231#discussion_r1192712631) from @effectfully .

plutus-benchmark/bls-benchmarks/src/PlutusBenchmark/BLS12_381/Common.hs
```diff
@@ -274,7 +276,7 @@ groth16Verify
     :: BuiltinByteString  -- G1
     -> BuiltinByteString  -- G2
     -> BuiltinByteString  -- G2
-    -> BuiltinByteString  -- G1
+    -> BuiltinByteString  -- G2
```
effectfully, line 279: Seems like it might be worth having some `newtype`s here idk?

effectfully (approved):

plutus-benchmark/bls-benchmarks/src/PlutusBenchmark/BLS12_381/Common.hs
kwxm: Yeah, I was a bit confused when I looked at that today.  These types are wrapped in `newtypes` elsewhere in this file, but not here.  I think I did that because what we're really wanting to measure here is what happens on the chain, and having newtypes here would cost a little more.  I think we get `delay` or something in the uplc if we do that, which I was slightly surprised by when I found out.  I could be wrong about that though: I'll check.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -258,14 +258,35 @@ instance ExMemoryUsage Data where
...
+{- Note [Costing constant-size types]
+The memory usage of each of the BLS12-381 types is constant, so we may be able
+to optimise things a little by ensuring that we don't re-compute the size of
+(say) a G1 element every time one is used. GHC will probably do this anyway, but
+we make sure by defining a top level function for each of the size measures and
+getting the memoryUsage instances to call those.
+-}
+
```
kwxm, line 269: The stuff above and below is in response to https://github.com/input-output-hk/plutus/pull/5231#discussion_r1192712631 from @effectfully .

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -258,14 +258,35 @@ instance ExMemoryUsage Data where
...
+    memoryUsage _ = g2ElementCost
     -- Should be 24
 
+{-# NOINLINE mlResultElementCost #-}
+mlResultElementCost :: CostRose
+mlResultElementCost = singletonRose . unsafeToSatInt $ BLS12_381.Pairing.mlResultMemSizeBytes `div` 8
+
 instance ExMemoryUsage BLS12_381.Pairing.MlResult where
-    memoryUsage _ = singletonRose . unsafeToSatInt $ BLS12_381.Pairing.mlResultMemSizeBytes `div` 8
+    memoryUsage _ = mlResultElementCost
```
kwxm, line 291: The stuff above is in response to [this comment](https://github.com/input-output-hk/plutus/pull/5231#discussion_r1192712631) from @effectfully .

PR #5370, opened 2023-06-04
Kwxm/bls12 381/golden cost tests (PLT-5947)

kwxm:
`plutus-benchmarks` contains executables that run costing "benchmarks" which print out tables (like [this](https://github.com/input-output-hk/plutus/blob/6b1860ab978ca7746942300ac4ab660100bb4577/plutus-benchmark/bls12-381-costs/test/bls12-381-costs.golden)) of costs for various sequences of `ed25519` and `bls12-381` operations so that we can get an idea of what we can do in a single validation.  This (a) refactors the code to eliminate duplication between the two sets of benchmarks, and (b) adds golden tests which check the outputs against checked-in copies.  I've kept the executables as well because they're easier to use if you're experimenting with changes in the cryptographic builtins.


plutus-benchmark/bls12-381-benchmarks/test/bls12-381-costs.golden
kwxm, line 1: This file is deleted in my local branch  but `git push` is failing to delete it here.

plutus-benchmark/bls12-381-costs/test/Spec.hs
```diff
@@ -0,0 +1,13 @@
+module Main where
+
+import PlutusBenchmark.BLS12_381.RunTests (runTests)
+import PlutusBenchmark.Common (goldenVsTextualOutput)
+
+outputFile :: String
+outputFile = "bls12-381-costs.txt"
+
+goldenFile :: FilePath
+goldenFile = "bls12-381-costs/test/bls12-381-costs.golden"
```
kwxm, line 10: I don't like having this relative path here but I don't think I can get rid of it.  If we change the name of the directory and fail to change this then the test will still succeed but write a new golden file into the wrong directory (which will be created if it doesn't exist).  Something similar will happen if you type `cabal run bls12-381-costs-test` (__not__ `cabal test .,.,.`) anywhere but in the root of `plutus-core` (but then you shouldn't use `cabal run` to run tests).

plutus-benchmark/ed25519-costs/test/Spec.hs
```diff
@@ -0,0 +1,13 @@
+module Main where
+
+import PlutusBenchmark.Common (goldenVsTextualOutput)
+import PlutusBenchmark.Ed25519.Common (runTests)
+
+outputFile :: String
+outputFile = "ed25519-costs.txt"
+
+goldenFile :: FilePath
+goldenFile = "ed25519-costs/test/ed25519-costs.golden"
```
kwxm, line 10: See the remark about the path in `bls12-381-costs/test/Spec.hs`.

james-iohk:
minor typo in comment `btyestring` at https://github.com/input-output-hk/plutus/blob/kwxm/BLS12-381/golden-cost-tests-2/plutus-core/plutus-core/src/PlutusCore/Crypto/BLS12_381/G1.hs#L101

kwxm:
> minor typo in comment `btyestring` at https://github.com/input-output-hk/plutus/blob/kwxm/BLS12-381/golden-cost-tests-2/plutus-core/plutus-core/src/PlutusCore/Crypto/BLS12_381/G1.hs#L101

FIxed; thanks!

plutus-benchmark/bls12-381-costs/test/bls12-381-costs.golden
```diff
@@ -0,0 +1,102 @@
...
+   60     737   (4.5%)      4092143400  (40.9%)          221142   (1.6%) 
+   70     847   (5.2%)      4774230520  (47.7%)          257502   (1.8%) 
+   80     957   (5.8%)      5456317640  (54.6%)          293862   (2.1%) 
+   90    1067   (6.5%)      6138404760  (61.4%)          330222   (2.4%) 
+  100    1177   (7.2%)      6820491880  (68.2%)          366582   (2.6%) 
+  110    1287   (7.9%)      7502579000  (75.0%)          402942   (2.9%) 
+  120    1397   (8.5%)      8184666120  (81.8%)          439302   (3.1%) 
+  130    1507   (9.2%)      8866753240  (88.7%)          475662   (3.4%) 
+  140    1617   (9.9%)      9548840360  (95.5%)          512022   (3.7%) 
+  150    1727  (10.5%)     10230927480 (102.3%)          548382   (3.9%)
```
effectfully, line 20: 102.3% of what?

plutus-benchmark/common/PlutusBenchmark/Common.hs
```diff
@@ -80,14 +100,23 @@ benchTermCek :: Term -> Benchmarkable
...
 
-type Result = EvaluationResult Term
+-- | Evaluate a script and return the CPU and memory costs (according to the cost model)
+getCostsCek :: UPLC.Program UPLC.NamedDeBruijn DefaultUni DefaultFun () -> (Integer, Integer)
+getCostsCek (UPLC.Program _ _ prog) =
+    case Cek.runCekDeBruijn PLC.defaultCekParameters Cek.tallying Cek.noEmitter prog of
+      (_res, Cek.TallyingSt _ budget, _logs) ->
+          let ExCPU cpu = exBudgetCPU budget
+              ExMemory mem = exBudgetMemory budget
+          in (fromSatInt cpu, fromSatInt mem)
```
effectfully, line 119: Nit: I'd perhaps just match on `budget` with `ExBudget`.

effectfully (approved):
I don't understand much here, but it seems OK.

plutus-benchmark/bls12-381-costs/test/bls12-381-costs.golden
kwxm: 102.3% of the maximum CPU ExUnits, as specified in `PlutusBenchmark.ProtocolParameters`, which it appears that I forgot to add to git.  The question is why this PR is passing the CI checks given that that file is imported into `PlutusBenchmarks.Common` [here](https://github.com/input-output-hk/plutus/blob/a9797ed084df6578ffb2d3e61ca1f030ddc1beb7/plutus-benchmark/common/PlutusBenchmark/Common.hs#L24) and its contents are used [here](https://github.com/input-output-hk/plutus/blob/a9797ed084df6578ffb2d3e61ca1f030ddc1beb7/plutus-benchmark/common/PlutusBenchmark/Common.hs#L175).  I'll leave the file out of the repository until we work out what's going on, but it contains this:
```
{-# LANGUAGE NumericUnderscores #-}

-- | 
module PlutusBenchmark.ProtocolParameters (max_tx_size, max_tx_ex_steps, max_tx_ex_mem)
where
  
-- Protocol parameters (June 2023)

-- | This is the "maximum transaction size".  We're just comparing the size of
-- the script with this, so our results may be a little optimistic if the
-- transaction includes other stuff (I'm not sure exactly what "maximum
-- transaction size" means).
max_tx_size :: Integer
max_tx_size = 16384

max_tx_ex_steps :: Integer
max_tx_ex_steps = 10_000_000_000

max_tx_ex_mem :: Integer
max_tx_ex_mem = 14_000_000
```

plutus-benchmark/bls12-381-costs/test/Spec.hs
kwxm: > You could add a check that the file already exists before running the test.

Yes, that might be a sensible idea. The way that cabal and tasty-golden deal with these things is maybe a little annoying (tasty-golden does report when it creates a new golden file, but it's easy to miss because it's in a pale colour).  Once we've generated the golden file for the first time it should be safe to assume that it should exist for ever afterwards. Now I'm wondering what we do with all of the golden files that we already have.

plutus-benchmark/common/PlutusBenchmark/Common.hs
```diff
@@ -80,14 +100,23 @@ benchTermCek :: Term -> Benchmarkable
 benchTermCek term =
     nf (runTermCek) $! term -- Or whnf?
 
+type Result = EvaluationResult Term
+
 {- | Just run a term (used for tests etc.) -}
-runTermCek :: Term -> EvaluationResult Term
+runTermCek :: Term -> Result
```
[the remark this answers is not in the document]
kwxm: It seemed like a good idea at the time, but in fact there's only one place that it's used, so I'll take it out.  Maybe it was used more often in a previous version.

plutus-benchmark/bls12-381-costs/test/bls12-381-costs.golden
kwxm: If I check out this branch in a separate copy of the `plutus` repository and try to  build `plutus-benchmark` it does what I'd expect:
```
$ cabal build plutus-benchmark/
Build profile: -w ghc-9.2.7 -O1
In order, the following will be built (use -v for more details):
 - plutus-benchmark-0.1.0.0 (lib:plutus-benchmark-common) (first run)
 [...]
Preprocessing library 'plutus-benchmark-common' for plutus-benchmark-0.1.0.0..
Error: cabal: can't find source for PlutusBenchmark/ProtocolParameters in
common, [...]

Error: cabal: Failed to build lib:plutus-benchmark-common from
plutus-benchmark-0.1.0.0 (which is required by bench:validation-full from
plutus-benchmark-0.1.0.0, bench:validation-decode from
plutus-benchmark-0.1.0.0 and others).
```

effectfully: Weird. Ask Lorenzo?

kwxm:
Closing this in favour of #5374.

PR #5372, opened 2023-06-06
[Textual] [Readable] Improve indentation for a lot of nodes

effectfully:
This addresses most of the remaining issues with the readable pretty-printer. One thing that is still missing is handling of applications of lambdas, those are basically unreadable right now, even knowing what to expect it still takes me way too much time to comprehend code having those.


plutus-tx-plugin/test/Budget/elemCheap.uplc-readable.golden
```diff
@@ -1,28 +1,53 @@
...
+                                           [ 8
+                                           , (constr
+                                                1
+                                                [ 9
+                                                , (constr
+                                                     1
+                                                     [ 10
+                                                     , (constr
+                                                          0
+                                                          [  ]) ]) ]) ]) ]) ]) ]) ]) ]) ]) ]))
```
effectfully, line 53: I think this would be better if `]` was aligned with `[`.

kwxm (approved):
I don't understand the details of the prettyprinting, but I looked at the diffs for most of the golden files and it was pretty clear that the only differences were in indentation, so I'm willing to believe that it's OK.  The output's greatly improved: most of the examples fit into the width of a screen and many of them are only about half as long (in terms of numbers of lines) as they were before, so I think this is doing what it's supposed to.

I wonder if we could get away with even less indentation, maybe only one space?  More generally, is there any possibility of making the amount of indentation globally configurable? Sometimes it would be convenient to be able to turn it off altogether: when I'm working with large UPLC files (no in Readable format though) I get things that are dozens of gigabytes, and 99.98% of that is white space.  I usually pipe the output through `sed 's/^ *//'` to make everything flush with the left margin to make the files smaller and more amenable for use with other tools.  I suppose we could do that inside `plc/uplc` by modifying the output stream, but it would  be handy if we could just get flush output in the first place. That's not a big deal though.

effectfully:
> More generally, is there any possibility of making the amount of indentation globally configurable?

For configurable pretty-printing -- probably yes. For regular `Pretty` I'm not quite sure. Might be easier to modify the output stream in `plc/uplc` as you propose.

plutus-core/plutus-core/src/PlutusCore/Core/Instance/Pretty/Readable.hs
```diff
@@ -18,11 +16,24 @@ import PlutusCore.Core.Type
...
+    go kind                  = ([], kind)
+viewKindArrow _ = Nothing
+
+-- | Split an iterated 'TyFun' (if any) into the list of argument types and the resulting type.
+viewTyFun :: Type tyname uni ann -> Maybe ([Type tyname uni ann], Type tyname uni ann)
+viewTyFun ty0@TyFun{} = Just $ go ty0 where
+    go (TyFun _ dom cod) = first (dom :) $ go cod
+    go ty                = ([], ty)
+viewTyFun _ = Nothing
+
```
bezirg, line 36: Sorry for the late review, I just remembered about these:

https://github.com/input-output-hk/plutus/blob/a1b9bb57d5342d7e757b4d8de3c70f456b425a92/plutus-core/plutus-ir/src/PlutusIR/Compiler/Datatype.hs#L67-L85

Can we merge them together?

PR #5374, opened 2023-06-07
Kwxm/bls12 381/golden cost tests 3 (PLT-5947)

kwxm:
This is another attempt at #5370, since that has a mysterious CI problem.  I've deliberately failed to add `ProtocolParameters.hs` to see if the same thing still happens.


kwxm:
I've now added the missing file.  #5375 reproduces the error, so it's not my fault.

kwxm:
For Jira's benefit, this is for PLT-5947.

kwxm:
I think it's safe to merge this now.  It seems that the problem with CI failing to detect a missing file wasn't specific to this PR, and it's now  being looked into elsewhere.

PR #5375, opened 2023-06-07
Test

kwxm:
Something odd is happening in #5374, where there's a missing file but it still passes CI.  This is an attempt to make a small example with similar behaviour.


kwxm:
This still passes despite there being a missing file: see below (it goes in `plutus-benchmark/common/PlutusCore/NOSUCHFILE.hs`).

```
module PlutusBenchmark.NOSUCHFILE (nosuchconstant)
where

nosuchconstant :: Integer
nosuchconstant = 44
```

effectfully:
Well, for that matter, let's see what this does:

kwxm:
This is fixed (ie, it now fails, as expected) by #5382.

PR #5376, opened 2023-06-07
Another example with a missing file

kwxm:
Here's another example that uses a file that's not been committed to GitHub.  This time it's in the `plutus-core` library and it's used in the `uplc` executable.


kwxm:
Ah, this one fails, unlike #5375.

kwxm:
The problem is correctly detected here, so I'm closing this.

PR #5379, opened 2023-06-07
Introduce type error

kwxm:
This introduces a type error in a benchmark, to see if it still passes CI.


kwxm:
...  and it does!

kwxm:
The error is now detected successfully after #5382, so I'm going to close this.

PR #5382, opened 2023-06-12
Restore wanted hydraJobs (Work In Progress)

zeme-wana:


kwxm:
This will hopefully fix the issue discussed in #5375, #5378, and #5379, where errors such as missing files in benchmarking code were not being detected in CI.

thealmarty:
What's the status of this? I am noticing some weirdness regarding the CI and benchmarking job in #5385 too. So I'm hoping this will make it less weird.

kwxm:
> What's the status of this? I am noticing some weirdness regarding the CI and benchmarking job in #5385 too. So I'm hoping this will make it less weird.

I think it should be ready to go, but I'll check with @zeme-wana to make sure he doesn't want to do anything else.

kwxm:
I had to add `TypeOperators` in a couple of places because I was getting a warning that we soon won't be able to use `~` without it.  Also I had to make all the imports of `asum` be from `Data.Foldable` instead of `Control.Applicative` because some of the CI machines are using GHC 8.12,which doesn't support the latter(see [here](https://hackage.haskell.org/package/base-4.18.0.0/changelog) and [here](https://gitlab.haskell.org/ghc/ghc/-/issues/19575)).

kwxm:
@michaelpj FYI this reverts some things from #5351.

kwxm:
This is pretty small and will improve our testing, so I'm going to merge it. Reviews are still welcome!

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -16,7 +16,8 @@ module PlutusIR.Purity
 import PlutusCore.Builtin
 import PlutusIR
 
-import Control.Applicative
+import Control.Applicative ((<|>))
+import Data.Foldable (asum)
```
effectfully, line 20: Kinda nasty... Looks like we need to add this to the style guide or something as we'll inevitably break it again in the future.

effectfully (commented):
Sorry for the late and useless review.

I don't understand nix, so it all looks great to me!

kwxm:
I'm getting a failure in master because of a warning about needing `TypeOperators` if we want to use `~`.  I fixed a similar one in this PR but for some reason I didn't get a complaint about the other one. The new one should be fixed in #5386 .

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
kwxm: I think if we wait a while the issue will go away because 8.12 won't be involved in CI any more. We could change it back when that happens, but we'll probably have forgotten about it by then.

PR #5385, opened 2023-06-12
PLT-6152 Drop fully applied condition in callsite inliner.

thealmarty:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
@@ -156,12 +156,12 @@ applyAndBetaReduce info args0 = do
   go rhsBody (varArity info) args0
 
 -- | Consider whether to inline an application.
-inlineSaturatedApp ::
+inlineApp ::
   forall tyname name uni fun ann.
   (InliningConstraints tyname name uni fun) =>
   Term tyname name uni fun ann ->
   InlineM tyname name uni fun ann (Term tyname name uni fun ann)
-inlineSaturatedApp t
+inlineApp t
   | (Var _ann name, args) <- splitApplication t =
       gets (lookupVarInfo name) >>= \case
         Just varInfo ->
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
@@ -20,7 +20,7 @@ import PlutusIR.Transform.Substitute
 
 import Control.Monad.State
 
-{- Note [Inlining and beta reduction of fully applied functions]
+{- Note [Inlining and beta reduction of functions]
 
 We inline if its cost and size are acceptable.
 
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
@@ -68,17 +68,21 @@ splitParams = \case
   -- Whenever we encounter a body that is not a lambda or type abstraction, we are done counting
   tm -> ([], tm)
 
--- | Returns the function with the supplied term or type lambda abstractions.
-mkAbs ::
-  -- | The function body
-  Term tyname name uni fun ann ->
+-- | Returns a function with the supplied term or type lambda abstractions. The result is wrapped in
+-- `Dupable` for renaming.
+mkAbs ::forall tyname name uni fun ann.
+  (InliningConstraints tyname name uni fun) =>
+  -- | The function body wrapped in `Dupable`.
+  Dupable (Term tyname name uni fun ann) ->
   Arity tyname name uni ann -> -- The abstractions
-  Term tyname name uni fun ann
-mkAbs body (TermParam ann n ty : nextLam) =
-  mkAbs (LamAbs ann n ty body) nextLam
-mkAbs body  (TypeParam ann n kd : nextLam) =
-  mkAbs (TyAbs ann n kd body) nextLam
-mkAbs body _ = body
+  InlineM tyname name uni fun ann (Dupable (Term tyname name uni fun ann))
+mkAbs body (TermParam ann n ty : nextLam) = do
+  liftedBody <- liftDupable body
+  mkAbs (dupable (LamAbs ann n ty liftedBody)) nextLam
+mkAbs body  (TypeParam ann n kd : nextLam) = do
+  liftedBody <- liftDupable body
+  mkAbs (dupable (TyAbs ann n kd liftedBody)) nextLam
+mkAbs body _ = pure body
 
 
 {- | Apply the RHS of the given variable to the given arguments and beta-reduce
@@ -123,8 +127,11 @@ applyAndBetaReduce info args0 = do
         (TermParam{}:_, TypeAppContext{}) -> pure Nothing
         (TypeParam{}:_, TermAppContext{}) -> pure Nothing
         -- no more arguments to apply, just apply what we have
-        (remainingArity, AppContextEnd) -> pure Nothing
-            -- pure $ Just (mkAbs acc remainingArity)
+        (remainingArity, AppContextEnd) -> do
+          -- renaming is needed otherwise we get multiply defined errors.
+          bodyWithAbs <- mkAbs (dupable acc) remainingArity
+          renamed <- liftDupable bodyWithAbs
+          pure $ Just renamed
 
       -- Is it safe to turn `(\a -> body) arg` into `body [a := arg]`?
       -- The criteria is the same as the criteria for inlining `a` in
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
@@ -5,8 +5,7 @@
 {-# LANGUAGE TypeFamilies     #-}
 
 {- |
-Call site inlining machinery. For now there's only one part: inlining of fully applied functions.
-We inline fully applied functions if the cost and size are acceptable.
+Call site inlining machinery. We inline if the size of the inlined result is not larger.
 See note [Inlining of fully applied functions].
 -}
 module PlutusIR.Transform.Inline.CallSiteInline where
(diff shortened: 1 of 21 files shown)


plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -154,7 +105,11 @@ fullyApplyAndBetaReduce info args0 = do
               (\n -> if n == param then Just <$> PLC.rename arg else pure Nothing)
               acc
           go acc' arity' args'
-        _ -> pure Nothing
+        -- term/type argument mismatch, don't inline
+        (TermParam _:_, TypeAppContext{}) -> pure Nothing
+        (TypeParam _:_, TermAppContext{}) -> pure Nothing
```
thealmarty, line 110: I’m thinking when we have a mismatch of term/type arguments, we don’t inline. This shouldn’t happen and so we leave it alone. Do people agree? Like if we do inline in such case for the correctly applied earlier arguments, it may reduce size, but I feel there’s a risk of making things more wrong?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -154,7 +105,11 @@ fullyApplyAndBetaReduce info args0 = do
...
+        -- no more arguments to apply, just apply what we have
+        (_, AppContextEnd) -> pure . Just $ fillAppContext acc args
```
zliu41, line 112: This is wrong. `args` is `AppContextEnd` here, so `fillAppContext acc args` does nothing - it just returns `acc`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -154,7 +105,11 @@ fullyApplyAndBetaReduce info args0 = do
               (\n -> if n == param then Just <$> PLC.rename arg else pure Nothing)
               acc
           go acc' arity' args'
-        _ -> pure Nothing
+        -- term/type argument mismatch, don't inline
+        (TermParam _:_, TypeAppContext{}) -> pure Nothing
+        (TypeParam _:_, TermAppContext{}) -> pure Nothing
+        -- no more arguments to apply, just apply what we have
+        (_, AppContextEnd) -> pure $ Just acc
```
zliu41, line 112: You can't just return `acc` here - there are remaining lambdas.

Consider `f x y` where `f = \a b c d -> a + b + c + d`.
You need to return `\c d -> x + y + c + d`.
But `acc` is `x + y + c + d`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -470,7 +470,8 @@ fullyApplyAndBetaReduce info args0 = do
                   acc
               go acc' params args'
             else pure Nothing
-        _ -> pure Nothing
+        -- no more arguments to apply, just apply what we have
+        (_, []) -> pure $ Just acc
```
zliu41, line 474: This also needs to be changed

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -154,7 +119,29 @@ fullyApplyAndBetaReduce info args0 = do
...
               acc
           go acc' arity' args'
-        _ -> pure Nothing
+        -- term/type argument mismatch, don't inline
+        (TermParam _:_, TypeAppContext{}) -> pure Nothing
+        (TypeParam _:_, TermAppContext{}) -> pure Nothing
+        -- no more arguments to apply, just apply what we have
+        (remainingArity, AppContextEnd) ->
+          let numberOfUnappliedLam = length remainingArity
+              allLambdas = fst $ splitLambdas (varRhs info)
```
zliu41, line 128: This is not needed. The lambdas are already split - the binders are `varArity` and the body is `varBody`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -154,7 +119,29 @@ fullyApplyAndBetaReduce info args0 = do
...
+              -- | Returns the function with the unapplied lambda.
+              partiallyAppliedFun ::
+                -- | The function body
+                Term tyname name uni fun ann ->
+                -- | The number of arguments not applied
+                Int ->
+                [Lam tyname name uni ann] -> -- The whole list of lambdas of the function
```
zliu41, line 135: This is also not needed - why use the whole list of lambdas? You already have `remainingArity`.

zliu41 (commented):
`inlineSaturatedApp` should be renamed.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
thealmarty: Yes I just wanted to see if the PIR part was right first.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: The difference is that `Arity` doesn't have all the things needed to rebuild the term, e.g., for `Term`, it's missing `Type`, and for `Type`, it's missing `Kind`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: I don't really *need* the whole list, but it's what `splitLambdas` returns and it'll be work to filter out the needed ones.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
zliu41: Then we need to put additional things into `Arity`. We already have `splitParams`, so we don't need `splitLambdas`

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
zliu41: If you put additional things you need into `Arity`, then you can simply use `remainingArity`, without needing to `splitLambdas`

zliu41: As an example: `f x y` where `f = \a b c d -> a + b + c + d`.
Here we have `acc = x + y + c + d`, and `remainingArity = [c, d]`. And we need to make `\x y -> x + y + c + d`.
So we can simply add a lambda abstraction to `acc` for each element of `remainingArity`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: Yes I can alter `splitParams` instead.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: Yes, I understand, if `Arity` already has all the needed info, and I will make it so.

zliu41: There's `mkIterLamAbs` and `mkIterTyAbs` in `PlutusCore.MkPlc`. Since this function makes both LamAbs and TyAbs, we can perhaps call it `mkAbs`.

thealmarty:
This is still a draft, not ready for review yet. I still need to do a few things (add tests, remove the fully applied condition in some tests etc)

michaelpj:
Is there some explanation of why we're considering this?

thealmarty:
> Is there some explanation of why we're considering this?

You mean dropping the fully applied condition? See [this ticket](https://input-output.atlassian.net/browse/PLT-6152).

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -166,23 +129,23 @@ fullyApplyAndBetaReduce info args0 = do
         Term tyname name uni fun ann ->
         InlineM tyname name uni fun ann Bool
       safeToBetaReduce a = shouldUnconditionallyInline Strict a rhsBody
```
michaelpj, line 131: This is an example of it being wrong. It's looking at the fully stripped body of the function, which isn't rigth.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -111,27 +112,35 @@ an under-approximation of how many arguments the term may need.
...
     -- ^ its arity, storing to avoid repeated calculations.
     , varRhsBody    :: InlineTerm tyname name uni fun ann
     -- ^ the body of the function, for checking `acceptable` or not. Storing this to avoid repeated
     -- calculations.
     }
 -- | Is the next argument a term or a type?
-data Param tyname name =
-    TermParam name | TypeParam tyname
+data Param tyname name uni ann =
+    TermParam ann name (Type tyname uni ann)
```
michaelpj, line 130: why put the extra info in? It's not used, and it's confusing what the point of the arity is, I think?

thealmarty:
I'm sorry Michael, I should clarify this isn't done so don't look at it. I temporarily abandoned this when I saw a problem and worked on #5395 instead so this is a mess now.

michaelpj:
That's okay, I deliberately looked here since this is where you said you had the problem!

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: Do you think this https://github.com/input-output-hk/plutus/pull/5426/commits/ebb818ad067e712e59a0e5a174583f8ee9fae4a7 is more right? `acc` is the whole function that has been processed.

michaelpj: yes, that looks more like it

michaelpj: I continue to think that `varBody` should just be removed now.

thealmarty:
Superseded by #5426

PR #5392, opened 2023-06-16
Rename some misspelled conformance tests

kwxm:
This corrects a couple of misspelled filenames in the conformance tests.  The file names should match the name of the directory containing them but I mistyped `serialised` as `seriliased`.

This has revealed an odd feature of the testing machinery.  The [discoverTest](https://github.com/input-output-hk/plutus/blob/071ed813ef2e2ab101b1fdfe76083a7fc3a84bee/plutus-conformance/src/PlutusConformance/Common.hs#L57) function searches recursively through a directory tree.  When it finds a directory with no subdirectories it assumes that it contains a uplc file and a corresponding golden file. If we're in the directory `someTest` then it expects it to contain `someTest.uplc` and `someTest.uplc.expected`; however, the logic's a bit faulty.  On [line 75](https://github.com/input-output-hk/plutus/blob/071ed813ef2e2ab101b1fdfe76083a7fc3a84bee/plutus-conformance/src/PlutusConformance/Common.hs#L75) it constructs the name of the golden file based on the directory name, and then it calls [getTestedValue](https://github.com/input-output-hk/plutus/blob/071ed813ef2e2ab101b1fdfe76083a7fc3a84bee/plutus-conformance/src/PlutusConformance/Common.hs#L112) to find the `.uplc` file.  However, instead of looking for `someTest.uplc` (based on the directory name) it searches for `.uplc` files in the directory and if it finds precisely one then it uses that for a golden test.  If the base name of the file isn't the same as the directory _then it still runs the test but creates a new golden file_.

For example (and this is effectively what happened here) if we're in the`someTest` directory and it contains `smoeTest.uplc` and `smoeTest.uplc.expected` then it bases the golden file name that it uses on the directory name, so it'll use `someTest.uplc.expected` for the name of the golden file, despite the fact that the directory contains `smoeTest.uplc.expected`; it'll then run the test, discover that `someTest.uplc.expected` doesn't exist, and create it, so we end up with two golden files, with the new one probably not committed to the repository.

I won't try to fix this here, but we should fix it.  I guess that it should either base everything on the name of the directory or search for `.uplc` files and append `.expected` for the name of the golden file (ignoring the directory name).  I don't know which would be better.


effectfully (approved):
Great investigation, thank you.

@thealmarty I assume this is your territory? Could you please create a ticket for yourself for fixing the behavior?

effectfully:
> I won't try to fix this here, but we should fix it. I guess that it should either base everything on the name of the directory or search for .uplc files and append .expected for the name of the golden file (ignoring the directory name). I don't know which would be better.

I'd perhaps just remove all the directories. `x.uplc` and `x.uplc.expected` appear next to each other anyway, so in my view those directories are an unnecessary indirection, but maybe I misunderstand it.

PR #5397, opened 2023-06-22
PLT-473 TPLC with Open builtins

mjaskelioff:
This PR implements open builtins for TPLC.

* A new kind `#` is introduced. This required many changes across all the formalisation, and in the way Agda code interfaces with Haskell code.
* `list` and `pair` are just type constants with higher-order kinds.
* Builtin signatures are redesigned to distinguish between variables polymorphic over all types (that is, variables of kind `*`) and variables polymorphic over builtin-types (i.e. of kind `#`)
* *All* current builtins are implemented.
* Several postulated lemmas were proved (parts of their proofs were needed somewhere else anyway).


plutus-metatheory/src/Algorithmic/CEK.lagda.md
```diff
@@ -72,43 +68,47 @@ data Value : (A : ∅ ⊢Nf⋆ *) → Set where
...
+    → {A : ∅ ⊢Nf⋆ K}
     → (q : C ≡ B [ A ]Nf)
     → {σC : SigTy (bubble pt) pa C}
-    → (σq : σC ≡ convSigTy (sym q) (σB [ A ]SigTy))
     → BApp b C σC
+
+infixl 4 _$_
+infixl 4 _$$_
+pattern Λ̂ A = _$$_ base {A = A} refl 
+pattern _●_ bapp B = _$$_ bapp {A = B} refl
```
mjaskelioff, line 111: I redefined the syntax so that the BUILTIN function definition looks a bit less ugly (I had to modify the definition anyway, since I needed to eliminate the term constructors which were not need anyway).

plutus-metatheory/src/Algorithmic.lagda
```diff
@@ -89,13 +95,79 @@ data _∋_ : (Γ : Ctx Φ) → Φ ⊢Nf⋆ * → Set where
...
+ty≅sty₁ (ne (^ list · x))        = cong ne (cong (^ list ·_) (ty≅sty₁ x)) 
+ty≅sty₁ (ne (^ (atomic x)))      = refl
+
+ty≅sty₂ : ∀ (A : 0 ⊢♯) →  A ≡ ty2sty (sty2ty A)
+ty≅sty₂ (atomic x) = refl
+ty≅sty₂ (list A) = cong list (ty≅sty₂ A)
+ty≅sty₂ (pair A B) = cong₂ pair (ty≅sty₂ A) (ty≅sty₂ B)
+\end{code}
+
+The semantics of types is given by the following interpretation function
```
jmchapman, line 146: 'semantics of types' of kind # I guess

plutus-metatheory/src/Algorithmic/CEK.lagda.md
```diff
@@ -353,16 +355,16 @@ step (s ; ρ ▻ Λ L) = s ◅ V-Λ L ρ
 step (s ; ρ ▻ (L ·⋆ A / refl)) = (s , -·⋆ A) ; ρ ▻ L
 step (s ; ρ ▻ wrap A B L) = (s , wrap-) ; ρ ▻ L
 step (s ; ρ ▻ unwrap L refl) = (s , unwrap-) ; ρ ▻ L
-step (s ; ρ ▻ con c) = s ◅ V-con c
+step (s ; ρ ▻ con c refl) = s ◅ V-con c
```
jmchapman, line 358: could you use a pattern synonym here?

plutus-metatheory/src/Algorithmic/CK.lagda.md
```diff
@@ -70,33 +71,33 @@ discharge : ∀{A : ∅ ⊢Nf⋆ *}{t : ∅ ⊢ A} → Value t → ∅ ⊢ A
...
-step ((s , (V-ƛ t ·-)) ◅ V)       = s ▻ (t [ discharge V ])
-step ((s , (-·⋆ A)) ◅ V-Λ t)      = s ▻ (t [ A ]⋆)
-step ((s , wrap-) ◅ V)            = s ◅ (V-wrap V)
-step ((s , unwrap-) ◅ V-wrap V)   = s ▻ deval V
-step (s ▻ (builtin b / refl))     = s ◅ ival b
-step ((s , (V-I⇒ b {am = 0} bt ·-)) ◅ vu) = s ▻ BUILTIN' b (app bt vu)
+step (s ▻ ƛ L)                                = s ◅ V-ƛ L
+step (s ▻ (L · M))                            = (s , -· M) ▻ L
+step (s ▻ Λ L)                                = s ◅ V-Λ L
+step (s ▻ (L ·⋆ A / refl))                    = (s , -·⋆ A) ▻ L
```
jmchapman, line 77: pattern synonym?

plutus-metatheory/src/Algorithmic/CK.lagda.md
```diff
@@ -70,33 +71,33 @@ discharge : ∀{A : ∅ ⊢Nf⋆ *}{t : ∅ ⊢ A} → Value t → ∅ ⊢ A
...
+step (s ▻ wrap A B L)                         = (s , wrap-) ▻ L
+step (s ▻ unwrap L refl)                      = (s , unwrap-) ▻ L
```
jmchapman, line 79: pattern synonym?

plutus-metatheory/src/Algorithmic/CK.lagda.md
```diff
@@ -70,33 +71,33 @@ discharge : ∀{A : ∅ ⊢Nf⋆ *}{t : ∅ ⊢ A} → Value t → ∅ ⊢ A
...
+step (s ▻ con cn refl)                        = s ◅ V-con cn
+step (s ▻ error A)                            = ◆ A
+step (ε ◅ V)                                  = □ V
+step ((s , (-· M)) ◅ V)                       = ((s , V ·-) ▻ M)
+step ((s , (V-ƛ t ·-)) ◅ V)                   = s ▻ (t [ discharge V ])
+step ((s , (-·⋆ A)) ◅ V-Λ t)                  = s ▻ (t [ A ]⋆)
+step ((s , wrap-) ◅ V)                        = s ◅ (V-wrap V)
+step ((s , unwrap-) ◅ V-wrap V)               = s ▻ deval V
+step (s ▻ (builtin b / refl))                 = s ◅ ival b
```
jmchapman, line 88: pattern synonym?

plutus-metatheory/src/Algorithmic/Completeness.lagda
```diff
@@ -112,9 +107,83 @@ lemσ σ C _ refl q = trans
...
+                  (cong Π (trans (sig2type⇒-lem {zero} {suc n} {Norm.mkTy r} {Syn.mkTy r} (head ∷ tail) (mkTy-lem r)) 
+                                 (sym (reifyCR (idext exte-lem (Syn.sig2type⇒ tail (Syn.mkTy head ⇒ Syn.mkTy r)))))
+                   )) 
+sig2type-lem (sig (suc n) n♯ (head ∷ tail) r) = sig2typeΠ-lem {t = Π (Syn.sig2type⇒ tail (Syn.mkTy head ⇒ Syn.mkTy r))}
+                  {Π (Norm.sig2type⇒ tail (Norm.mkTy head ⇒ Norm.mkTy r))}
+                  (cong Π (trans (sig2type⇒-lem {suc n} {n♯} {Norm.mkTy r} {Syn.mkTy r} (head ∷ tail) (mkTy-lem r)) 
+                                 (sym (reifyCR (idext exte-lem (Syn.sig2type⇒ tail (Syn.mkTy head ⇒ Syn.mkTy r)))))
+                   )) 
+
+btype-lem : ∀ {Φ} b → Norm.btype {Φ} b ≡ nf (Syn.btype b)
```
jmchapman, line 172: postulate-- :)

plutus-metatheory/src/Algorithmic.lagda
mjaskelioff: Right

plutus-metatheory/src/Raw.hs
```diff
@@ -63,34 +66,41 @@ unIndex (Index n) = toInteger n
...
 convT :: Type NamedTyDeBruijn DefaultUni a -> RType
 convT (TyVar _ (NamedTyDeBruijn x)) = RTyVar (unIndex (ndbnIndex x))
 convT (TyFun _ _A _B)               = RTyFun (convT _A) (convT _B)
-convT (TyForall _ _ _K _A)          = RTyPi (() <$ _K) (convT _A)
-convT (TyLam _ _ _K _A)             = RTyLambda (() <$ _K) (convT _A)
+convT (TyForall _ _ _K _A)          = RTyPi (convK _K) (convT _A)
+convT (TyLam _ _ _K _A)             = RTyLambda (convK _K) (convT _A)
 convT (TyApp _ _A _B)               = RTyApp (convT _A) (convT _B)
-convT (TyBuiltin _ b)               = RTyCon (convTyCon b)
+convT (TyBuiltin _ b)               = convTyCon b
```
effectfully, line 79: This may fail, given `error "unsupported builtin type application"` below. You should handle the `DefaultUniApply` case (by elaborating it to `RTyApp`) and only call `convTyCon` if `b` is not a type application.

plutus-metatheory/src/Scoped.lagda
```diff
@@ -32,17 +32,13 @@ open import RawU using (TyTag;TmCon;tmCon;tagCon2TmCon;tmCon2TagCon)
...
+data ScopedTy (n : ℕ) : Set where
   `    : Fin n → ScopedTy n
   _⇒_  : ScopedTy n → ScopedTy n → ScopedTy n
   Π    : Kind → ScopedTy (suc n) → ScopedTy n
   ƛ    : Kind → ScopedTy (suc n) → ScopedTy n
   _·_  : ScopedTy n → ScopedTy n → ScopedTy n
-  con  : S.TyCon n → ScopedTy n
+  con  : ∀{K} → TyCon K → ScopedTy n
   μ    : ScopedTy n → ScopedTy n → ScopedTy n
   missing : ScopedTy n -- for when things compute to error
```
effectfully, line 43: Would be great to have an explanation why we need that `missing` thing.

plutus-metatheory/src/Utils.lagda
```diff
@@ -202,14 +202,20 @@ a kind called `#` which is used for sized integers and bytestrings.
...
+-- {-# FOREIGN GHC import PlutusCore                       #-}
+-- {-# FOREIGN GHC {-# LANGUAGE GADTs, PatternSynonyms #-} #-}
+
+-- {-# FOREIGN GHC type KIND = Kind ()                     #-}
+
+-- {-# FOREIGN GHC pattern Star    = Type ()               #-}
+
+-- {- TODO: Map Sharp to the correct kind once it is added to the Haskell implementation -}
+-- {-# FOREIGN GHC pattern Sharp   = Type ()               #-}
+-- {-# FOREIGN GHC pattern Arrow k j = KindArrow () k j    #-}
```
effectfully, line 217: Just to be clear, we are not even sure that adding `#` to the implementation is going to happen. Nobody is working on that and from what I know nobody plans to in any foreseeable future. There's plenty of things our users care about and having the builtin kind just isn't one of them.

effectfully (approved):
Looks great to me. I didn't review the proofs.

plutus-metatheory/src/Algorithmic/CEK.lagda.md
mjaskelioff: I could. The synonym cannot be named `con` (could be named something like `const`) even if we rename the real `con` constructor to `con_/_` because it clashes with the `con` constructor for types.

plutus-metatheory/src/Algorithmic.lagda
```diff
@@ -89,13 +95,79 @@ data _∋_ : (Γ : Ctx Φ) → Φ ⊢Nf⋆ * → Set where
...
           
-Let `x`, `y` range over variables.
+## Semantic of constant terms
+
+We define a predicate ♯Kinded for kinds that ultimately end in ♯.
+
+\begin{code}
+data ♯Kinded : Kind → Set where
+   ♯ : ♯Kinded ♯
+   K♯ : ∀{K J} → ♯Kinded J → ♯Kinded (K ⇒ J)
```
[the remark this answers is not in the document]
mjaskelioff: Yes, in principle. You wouldn't be able to construct one, though, given the current builtin type constants.

plutus-metatheory/src/Algorithmic/CK.lagda.md
mjaskelioff: I'm not convinced we should make every constructor with a `refl` a pattern synonym. Unlike constructors, pattern synonyms have poor overloading of names, so they are best used local to a module. We reuse a lot of syntax, and if we took the route of doing a pattern synonym to avoid every `refl`, we would need unique global names for every pattern synonym. This means introducing a lot of new names just for the sake of avoiding the ocasional `refl`.

plutus-metatheory/src/Algorithmic/CEK.lagda.md
```diff
@@ -72,43 +68,47 @@ data Value : (A : ∅ ⊢Nf⋆ *) → Set where
    → Value (nf (embNf A · ƛ (μ (embNf (weakenNf A)) (` Z)) · embNf B))
    → Value (μ A B)
 
-  V-con : {tcn : TyCon _}
-    → (cn : TermCon {∅} (con tcn))
-    → Value (con tcn)
+  V-con : ∀{A : ∅ ⊢Nf⋆ ♯}
+    → ⟦ A ⟧ 
+    → Value (con A)
```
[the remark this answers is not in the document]
mjaskelioff: Why is that? The `TermCon` that we would have here would we a dependent pair of a type and its meaning, so it would be equivalent but with one more level of indirection.

plutus-metatheory/src/Builtin.lagda.md
```diff
@@ -135,38 +138,46 @@ private module SugaredSignature where
...
+    -- names for type variables of kind ⋆
+    A :  ∀{n⋆ n♯} → suc n⋆ / n♯ ⊢⋆
+    A = ` Z
+
+    B :  ∀{n⋆ n♯} → suc (suc n⋆) / n♯ ⊢⋆
+    B = ` (S Z)
+
+    -- names for type variables of kind ♯
+    a : ∀{n♯} → suc n♯ ⊢♯
     a = ` Z
```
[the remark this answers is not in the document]
mjaskelioff: Do you think it's worth it adding one here? In any case we can discuss it on Slack. It would be a separate ticket anyway.

plutus-metatheory/src/Builtin/Constant/AtomicType.lagda.md
```diff
@@ -37,3 +44,18 @@ data AtomicTyCon : Set where
...
 unquoteDef decAtomicTyCon = defDec (quote AtomicTyCon) decAtomicTyCon
 ```
+
+## Meaning of Atomic Constants
+
+```
+⟦_⟧at : AtomicTyCon → Set
+⟦ aInteger ⟧at = ℤ
+⟦ aBytestring ⟧at = ByteString
+⟦ aString ⟧at = String
```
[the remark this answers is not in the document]
mjaskelioff: I'm not sure it would change anything. Is there any difference aside from performance? The strings get converted to Text to use Haskell functions such as pretty-printing.

plutus-metatheory/src/Algorithmic/CEK.lagda.md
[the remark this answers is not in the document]
effectfully: > The `TermCon` that we would have here would we a dependent pair of a type and its meaning

I want the implementation to have a `TermCon` that isn't just (morally) a dependent pair, but instead is a bunch of constructors with type tags in their names rather than in a separate field. That is faster than storing tags separately and isn't much more boilerplate (we'd still have to keep type tags around for fast unlifting of lists and pairs).

plutus-metatheory/src/Check.lagda.md
```diff
@@ -85,220 +81,303 @@ data TypeError : Set where
...
+len⋆ ∅        = 0
+len⋆ (Φ ,⋆ K) = suc (len⋆ Φ)
+
+inferTyVar : ∀ Φ (i : Fin (len⋆ Φ)) → Σ Kind (Φ ∋⋆_)
+inferTyVar (Φ ,⋆ K) zero    = K ,, Z
+inferTyVar (Φ ,⋆ K) (suc i) = let J ,, α = inferTyVar Φ i in  J ,, S α
+
+⊎bind : {A B C : Set} → A ⊎ C → (A → B ⊎ C) → B ⊎ C
+⊎bind (inj₁ a) f = f a
+⊎bind (inj₂ c) f = inj₂ c
```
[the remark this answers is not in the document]
mjaskelioff: I just left it the way it was before (the old lines 70-72).

plutus-metatheory/src/Builtin/Constant/AtomicType.lagda.md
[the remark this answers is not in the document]
effectfully: > Is there any difference aside from performance? 

Accuracy? It's already confusing on the implementation side that `String` maps to `Text` and here we add a bunch more confusion on top.

It doesn't really matter, I'm just nitpicking.

plutus-metatheory/src/Check.lagda.md
```diff
@@ -3,10 +3,17 @@ layout: page
 ---
 ## Type checker
 
+This module implements typechecking and inference for Scoped terms.
+
+Scoped terms `ScopedTm` have scoped types `ScopedTy` which don't have kinds, so
+kinds need to be inferred. Since we have two base kinds (`*` and `♯`) and the 
+latter embeds into the former, there is some subtleties discussed below.
```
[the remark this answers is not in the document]
effectfully: Right, silly me, we don't have the builtin kind in the syntax. Thanks!

plutus-metatheory/src/Scoped.lagda
mjaskelioff: I don't think it is needed anymore. I'll try and remove it.

plutus-metatheory/src/Utils.lagda
mjaskelioff: I understand. I can probably remove this comment since we now have code for mapping between Haskell and Agda kinds (`convK` and `unconvK`)

plutus-metatheory/src/Check.lagda.md
```diff
@@ -85,220 +81,303 @@ data TypeError : Set where
...
+isPat ((♯ ⇒ K₁) ,, A) = inj₁ (notPat * λ _ ())
+isPat (((K ⇒ K₂) ⇒ *) ,, A) = inj₁ (notPat * λ _ ())
+isPat (((K ⇒ K₂) ⇒ ♯) ,, A) = inj₁ (notPat * λ _ ())
+isPat (((K ⇒ *) ⇒ (K₁ ⇒ *)) ,, A) = do
+      refl ← withE (kindMismatch _ _) (dec2Either (decKind K K₁))
+      return (K ,, A) 
+isPat (((K ⇒ *) ⇒ (K₁ ⇒ ♯)) ,, A) = inj₁ (notPat * λ _ ())
+isPat (((K ⇒ *) ⇒ (K₁ ⇒ (K₃ ⇒ K₄))) ,, A) = inj₁ (notPat * λ _ ())
+isPat (((K ⇒ ♯) ⇒ (K₁ ⇒ K₃)) ,, A) = inj₁ (notPat * λ _ ())
+isPat (((K ⇒ (K₂ ⇒ K₄)) ⇒ (K₁ ⇒ K₃)) ,, A) = inj₁ (notPat * λ _ ())
```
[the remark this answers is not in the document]
mjaskelioff: As you mention below, the type errors are wrong. They are all different.

plutus-metatheory/src/Builtin/Constant/AtomicType.lagda.md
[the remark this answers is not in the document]
mjaskelioff: I'll make a ticket to fix this in the future.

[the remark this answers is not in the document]
mjaskelioff: Done. PLT-6378.

plutus-metatheory/src/Check.lagda.md
```diff
@@ -85,220 +81,303 @@ data TypeError : Set where
...
+  2.    con (ne (^ (atomic aInteger)))
+
+Whenever we have a constant of kind `♯` we embed it into `*` using `con`.
+This means that a composition of, for instance a list (kind `♯ ⇒ ♯`) applied 
+to a constant (which will be of kind `*`) doesn't match exactly. So we relax 
+this condition  when checking kinds and allow
+  1. Checking a `*`-kinded type against `♯`, whenever the type is of the form `con A` : 
+       we "downgrade" the type to A of kind `♯`. 
+  2. Checking a `♯`-kinded type against `*`: 
+       we "upgrade" the type to `*` using `con`.
```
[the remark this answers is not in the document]
mjaskelioff: One problem is that a standalone `con (atomic aInteger)` should be inferred as a kind star * (i.e. 
`con (ne (^ (atomic aInteger)))`. I guess one could try to rewrite the inferKind function as a two-step process, but I'm not sure it would work in general. I'll try.

plutus-metatheory/src/Algorithmic/CEK.lagda.md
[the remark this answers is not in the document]
mjaskelioff: OK. But you are talking about the untyped case. For the typed case, this version is potentially more efficient, because you only can get something of the correct type, so you wouldn't need to check for tags. Anyway, efficiency is not currently an issue for the formalisation, as long as its performance is not unreasonable and can be used for testing.

plutus-metatheory/src/Raw.hs
mjaskelioff: Aren't the good cases covered by lines 95-98?
```
convTyCon (SomeTypeIn (DefaultUniList a))             =
    RTyApp (RTyCon RTyConList) (convTyCon (SomeTypeIn a))
convTyCon (SomeTypeIn (DefaultUniPair a b))           =
    RTyApp (RTyApp (RTyCon RTyConPair) (convTyCon (SomeTypeIn a))) (convTyCon (SomeTypeIn b))
```

plutus-metatheory/src/Algorithmic/CEK.lagda.md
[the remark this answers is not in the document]
effectfully: > Anyway, efficiency is not currently an issue for the formalisation

Sure, I'm just letting you know what is planned for the implementation, you don't need to follow that in the formalization, but I thought you might want to know.

plutus-metatheory/src/Raw.hs
effectfully: :woman_facepalming: I have no idea how I missed that, sorry!

Anyways, those lines aren't quite correct. One can have a partially applied `DefaultUniProtoPair` for example and then neither of these cases would catch it. Just elaborate any `DefaultUniApply` to `RTyApp` and handle `DefaultUniProtoList` and `DefaultUniProtoPair` to handle all possible cases and you'll be able to avoid throwing that `error` there too.

Also I'd define `convTyCon` over `DefaultUni (Esc a)` rather than `SomeTypeIn DefaultUni`, so that you don't need to match on `SomeTypeIn` repeatedly.

plutus-metatheory/src/Check.lagda.md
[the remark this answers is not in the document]
mjaskelioff: As it happens it's not used at all. I'm going to erase the definition.

plutus-metatheory/src/Raw.hs
mjaskelioff: Thank you Roman.

> Also I'd define convTyCon over DefaultUni (Esc a) rather than SomeTypeIn DefaultUni, so that you don't need to match on SomeTypeIn repeatedly.

You mean define `convTyCon :: DefaultUni (Esc a) -> RType`?
Wouldn't this open the existential prematurely andfail to typecheck?

mjaskelioff: I mean fail to typecheck when calling it
```
convT (TyBuiltin _ (SomeTypeIn b))  = convTyCon b
```

plutus-metatheory/src/Check.lagda.md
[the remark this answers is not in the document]
mjaskelioff: Ok. Now I did implemented upgrading "as needed" and I remembered why we need to prematurely upgrade. You have functions like `isPat` that checks a type has kind `(K ⇒ *) ⇒ (K ⇒ *)` and returns a type `Φ ⊢Nf⋆ (K ⇒ *) ⇒ (K ⇒ *)`.  
If we have a kind, say `(K ⇒ ♯) ⇒ (K ⇒ ♯)` then we have the requirement to upgrade `♯`, but we can't get inside the type to insert the appropriate `con`s. (Well, we probably can make a function to do that, but that is even messier than the solution above.)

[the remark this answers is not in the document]
mjaskelioff: I'll add an example for this case.

plutus-metatheory/src/Utils.lagda
mjaskelioff: Comment Removed

PR #5407, opened 2023-06-28
PLT-6233: Fix Flat (Binder FakeNamedDeBruijn) instance. Add tests

effectfully:
An exact clone of #5390 created for benchmarking purposes.


effectfully:
That's what we wanted to see.

PR #5409, opened 2023-06-28
Benchmark the production code fully rather than some arbitrary thing

effectfully:
This is a continuation of #5200, which updated `BenchCek.hs` to use the correct evaluation context. This one updates `BenchFull.hs`.

### DIFF
--- a/plutus-benchmark/validation/BenchFull.hs
+++ b/plutus-benchmark/validation/BenchFull.hs
@@ -2,15 +2,14 @@
 module Main where
 
 import PlutusCore.Evaluation.Machine.ExBudget
-import PlutusLedgerApi.Test.EvaluationContext (evalCtxForTesting)
 import PlutusLedgerApi.V1
 import UntypedPlutusCore qualified as UPLC
 
 import Common
 import Control.DeepSeq (force)
+import Control.Exception
 import Criterion
 import Data.ByteString as BS
-import Data.Either
 
 {-|
 for each data/*.flat validation script, it benchmarks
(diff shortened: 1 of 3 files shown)


effectfully:
^ yay, noise.

plutus-benchmark/validation/BenchFull.hs
```diff
@@ -22,36 +21,29 @@ the whole time taken from script deserialization to script execution result.
...
+                -- We make some effort to mimic what happens on-chain, including the provision of
+                -- the script arguments. However, the inputs we have are *fully applied*. So we try
+                -- and reverse that by stripping off the arguments here. Conveniently, we know that
+                -- they will be Data constants. Annoyingly we can't just assume it's the first 3
+                -- arguments, since some of them are policy scripts with only 2.
+                (term, args) = peelDataArguments body
+                -- strictify and "short" the result cbor to create a real `SerialisedScript`
+                !benchScript = force . serialiseUPLC $ UPLC.Program () ver term
+                eval script = do
+                    either (error . show) (\_ -> ()) . snd $ evaluateScriptRestricting
```
bezirg, line 38: You can use i think `PlutusPrelude.unsafeFromRight`

plutus-benchmark/validation/BenchFull.hs
```diff
@@ -22,36 +21,29 @@ the whole time taken from script deserialization to script execution result.
...
+            let UPLC.Program () ver body = unsafeUnflat file bsFlat
+                -- We make some effort to mimic what happens on-chain, including the provision of
+                -- the script arguments. However, the inputs we have are *fully applied*. So we try
+                -- and reverse that by stripping off the arguments here. Conveniently, we know that
+                -- they will be Data constants. Annoyingly we can't just assume it's the first 3
+                -- arguments, since some of them are policy scripts with only 2.
+                (term, args) = peelDataArguments body
+                -- strictify and "short" the result cbor to create a real `SerialisedScript`
+                !benchScript = force . serialiseUPLC $ UPLC.Program () ver term
+                eval script = do
```
bezirg, line 37: `do` looks redundant here

plutus-benchmark/validation/BenchFull.hs
effectfully: I think that'd return `()` regardless of whether we get `Left` or `Right` and we do want to throw on `Left` instead.

bezirg: oups yeah

plutus-benchmark/validation/BenchFull.hs
effectfully: Yup, thanks, fixed.

kwxm:
Should we worry about this for the nofib  benchmarks too?  I'm getting confused because we have lots of different benchmarks that all seem to have their own ways of evaluating terms.

PR #4733, opened 2022-06-24
Bitwise operations

kozross:
This is a work-in-progress implementation of [this CIP](https://github.com/cardano-foundation/CIPs/pull/283), which ultimately aims to solve #4252. I have left the pre-submit checklist in place (below) to be filled in as this gets finished.

Edit: This is now complete. @michaelpj - requesting a review.

Below is the checklist of tasks that have been done, and those that still remain outstanding:

- `integerToByteString`
   - [X] Implementation
   - [X] Plutus Core builtin
   - [X] Tests for Plutus Core builtin
   - [X] `PlutusTx` primitive 
- `byteStringToInteger`
   - [X] Implementation
   - [X] Plutus Core builtin
   - [X] Tests for Plutus Core builtin
   - [X] `PlutusTx` primitive 
- `andByteString`
   - [X] Implementation
   - [X] Plutus Core builtin
   - [X] Tests for Plutus Core builtin
   - [X] `PlutusTx` primitive 
- `iorByteString`
   - [X] Implementation
   - [X] Plutus Core builtin
   - [X] Tests for Plutus Core builtin
   - [X] `PlutusTx` primitive 
- `xorByteString`
   - [X] Implementation
   - [X] Plutus Core builtin
   - [X] Tests for Plutus Core builtin
   - [X] `PlutusTx` primitive 
- `complementByteString`
   - [X] Implementation
   - [X] Plutus Core builtin
   - [X] Tests for Plutus Core builtin
   - [X] `PlutusTx` primitive 
- `shiftByteString`
   - [X] Implementation
   - [X] Plutus Core builtin
   - [X] Tests for Plutus Core builtin
   - [X] `PlutusTx` primitive 
- `rotateByteString`
   - [X] Implementation
   - [X] Plutus Core builtin
   - [X] Tests for Plutus Core builtin
   - [X] `PlutusTx` primitive 
- `popCountByteString`
   - [X] Implementation
   - [X] Plutus Core builtin
   - [X] Tests for Plutus Core builtin
   - [X] `PlutusTx` primitive 
- `testBitByteString`
   - [X] Implementation
   - [X] Plutus Core builtin
   - [X] Tests for Plutus Core builtin
   - [X] `PlutusTx` primitive 
- `writeBitByteString`
   - [X] Implementation
   - [X] Plutus Core builtin
   - [X] Tests for Plutus Core builtin
   - [X] `PlutusTx` primitive 
- `findFirstSetByteString`
   - [X] Implementation
[...]

PR #5411, opened 2023-07-02
Kwxm/costing/linear in x and y

kwxm:
We currently have two-variable costing functions like `(x,y) -> a + b*max(x,y)`, but not one that's a linear (or maybe more accurately, affine) combination of both variables `(x,y) -> a +bx +cy`.  This adds that because I think we'll need it for #4733.


plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
```diff
@@ -278,7 +278,7 @@ unsafeReadModelFromR formula rmodel = do
     Right x  -> pure x
 
 -- t = ax+by+c
-unsafeReadModelFromR2 :: MonadR m => String -> String -> (SomeSEXP (Region m)) -> m (CostingInteger, CostingInteger, CostingInteger)
+unsafeReadModelFromR2 :: MonadR m => String -> String -> (SomeSEXP (Region m)) -> m (Intercept, Slope, Slope)
```
kwxm, line 281: This was previously unused.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -253,6 +273,7 @@ data ModelTwoArguments =
     ModelTwoArgumentsConstantCost       CostingInteger
   | ModelTwoArgumentsLinearInX          ModelLinearSize
   | ModelTwoArgumentsLinearInY          ModelLinearSize
+  | ModelTwoArgumentsLinearInXAndY      ModelLinearSizeTwoVariables
```
kwxm, line 276: This has to be `XAndY` because `XandY` gives `linear_in_xand_y` in the JSON.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/BuiltinCostModel.hs
```diff
@@ -21,6 +21,7 @@ module PlutusCore.Evaluation.Machine.BuiltinCostModel
     , ModelConstantOrLinear(..)
     , ModelConstantOrTwoArguments(..)
     , ModelLinearSize(..)
+    , ModelLinearSizeTwoVariables(..)
```
kwxm, line 24: I wondered whether to call this `ModelLinearSizeTwoArguments` but then came down on the side of `Variables` instead because I thought that "arguments" referred to the arguments of the builtin and "variables` referred to the aruguments of the costing function.  I may change my mind about that.

kwxm: /benchmark validation

kwxm:
^ In case adding another constructor slows things down.

kwxm:
I don't believe these figures, but at least it doesn't seem to be getting worse.

kwxm:
Let's try that again.

/benchmark validation

kwxm:
/benchmark validation

Does it work if you say something afterwards?

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -179,6 +180,16 @@ scaleLinearly (Intercept intercept) (Slope slope) =
...
 {-# INLINE scaleLinearly #-}
 
+-- | Take an intercept, two slopes and two streams, and scale each element of
+-- the first stream by the first slope, each element of the second stream by the
+-- second slope, add the two scaled streams together, and cons the intercept to
+-- the stream afterwards.
+scaleLinearlyTwoVariables :: Intercept -> Slope -> CostStream -> Slope -> CostStream -> CostStream
+scaleLinearlyTwoVariables (Intercept intercept) (Slope slope1) stream1 (Slope slope2) stream2 =
+    addCostStream (CostLast intercept) $
+                  addCostStream (mapCostStream (slope1 *) stream1) (mapCostStream (slope2 *) stream2)
```
effectfully, line 190: Nit: weird indentation. Otherwise looks good to me.

effectfully (approved):
LGTM

I'm kinda tempted to say that `addCostStream` is `<>`, so that we can write `fold` to add up multiple streams.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -251,8 +275,9 @@ data ModelConstantOrTwoArguments = ModelConstantOrTwoArguments
 
 data ModelTwoArguments =
     ModelTwoArgumentsConstantCost       CostingInteger
-  | ModelTwoArgumentsLinearInX          ModelLinearSize
-  | ModelTwoArgumentsLinearInY          ModelLinearSize
+  | ModelTwoArgumentsLinearInX          OneVariableLinearFunction
```
kwxm, line 278: We can change the names of the things on the right without messing anything up, but changing the constructor names on the left changes the auto-derived JSON tags in the cost model file, which we probably don't want to do.

kwxm: It might be conceptually clearer if we renamed `ModelAddedSizes` and so on, but calling them all `Model...` helps to see that they're things of a similar type.  Another thing we could do would be to put these things in a separate file (`CostingFunction` or something) and use them in qualified form.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/BuiltinCostModel.hs
kwxm: I've now changed the names again and added a big comment [here](https://github.com/input-output-hk/plutus/blob/3ef1b15f7c65cb87ffc6e134ef627a6168ff10d3/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs#L195). As mentioned in some of the other comments that I made on the PR, some renaming/refactoring might make this stuff a bit clearer, but this isn't the place to do it.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
kwxm: Should be fixed now.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
effectfully: I'd prefer if we renamed `ModelAddedSizes` and co too, otherwise with `OneVariableLinearFunction` it's all wildly inconsistent.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -284,6 +309,20 @@ runCostingFunTwoArguments (CostingFun cpu mem) =
...
+-- second slope, add the two scaled streams together, and cons the intercept to
+-- the stream afterwards.
+scaleLinearlyTwoVariables :: Intercept -> Slope -> CostStream -> Slope -> CostStream -> CostStream
+scaleLinearlyTwoVariables (Intercept intercept) (Slope slope1) stream1 (Slope slope2) stream2 =
+    addCostStream
+    (CostLast intercept)
+    (addCostStream
+     (mapCostStream (slope1 *) stream1)
+     (mapCostStream (slope2 *) stream2)
+    )
```
effectfully, line 323: I liked the previous version more than this one with inconsistent indentation of arguments, one of which is 0 spaces and the other is 1 space, none of which we normally use.

Anyways, I'm nitpicking.

effectfully (approved):
Clearer now, but I wish we changed the legacy code as well to use similar names. Not necessarily in this PR.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
kwxm: That's just what emacs did for me!  I couldn't get it to do anything better with the previous version.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
kwxm: I'll think about that a bit and maybe do it in another PR.  I think I might put these things in a separate file so that it's clear that they're all the same sort of thing.  There are a lot of things starting with `Model` and I was getting confused trying to sort them out mentally.

kwxm:
This ls PLT-6540.

PR #5413, opened 2023-07-05
Add Typeable and IsData instances for BLS12-381 types (PLT-6512)

kwxm:
@james-iohk ran into problems when trying to use the BLS functions in some of his end-to-end tests (#5383) because I'd failed to add some instances in `plutus-tx`.  This should fix that.

### DIFF
--- a/plutus-tx/src/PlutusTx/IsData/Class.hs
+++ b/plutus-tx/src/PlutusTx/IsData/Class.hs
@@ -190,10 +190,10 @@ instance (TypeError ('Text "toBuiltinData is not supported for BuiltinBLS12_381_
     => ToData Builtins.BuiltinBLS12_381_MlResult where
     toBuiltinData = Haskell.error "unsupported"
 instance (TypeError ('Text "fromBuiltinData is not supported for BuiltinBLS12_381_MlResult"))
-    => FromData Haskell.Int where
+    => FromData Builtins.BuiltinBLS12_381_MlResult where
     fromBuiltinData = Haskell.error "unsupported"
 instance (TypeError ('Text "unsafeFromBuiltinData is not supported for BuiltinBLS12_381_MlResult"))
-    => UnsafeFromData Haskell.Int where
+    => UnsafeFromData Builtins.BuiltinBLS12_381_MlResult where
     unsafeFromBuiltinData = Haskell.error "unsupported"
 
 -- | Convert a value to 'PLC.Data'.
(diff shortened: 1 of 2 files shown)


plutus-tx/src/PlutusTx/IsData/Class.hs
```diff
@@ -139,6 +141,61 @@ instance UnsafeFromData Void where
...
+instance UnsafeFromData Builtins.BuiltinBLS12_381_G2_Element where
+    {-# INLINABLE unsafeFromBuiltinData #-}
+    unsafeFromBuiltinData = Builtins.bls12_381_G2_uncompress . unsafeFromBuiltinData
+
+{- | We do not provide instances of any of these classes for
+   BuiltinBLS12_381_MlResult since there is no serialisation format: we expect
+   that values of that type will only occur as the result of on-chain
+   computations.
+-}
+instance (TypeError ('Text "toBuiltinData is not supported for BuiltinBLS12_381_MlResult"))
```
kwxm, line 189: I just copied what it does for `Int` above here.  We could also just not provide any instances, but this may give more informative error messages.

kwxm:
I notice that we don't seem to have any tests for this stuff.  I'm not sure what form they would take though.

plutus-tx/src/PlutusTx/IsData/Class.hs
```diff
@@ -139,6 +141,61 @@ instance UnsafeFromData Void where
...
+    toBuiltinData = toBuiltinData . Builtins.bls12_381_G2_compress
+instance FromData Builtins.BuiltinBLS12_381_G2_Element where
+    {-# INLINABLE fromBuiltinData #-}
+    fromBuiltinData d =
+        case fromBuiltinData d of
+          Nothing -> Nothing
+          Just (BI.BuiltinByteString bs) ->
+              case BLS12_381.G2.uncompress bs of
+                Haskell.Left _  -> Nothing
+                Haskell.Right g -> Just $ toBuiltin g
```
effectfully, line 179: Can you write something like

```haskell
        BI.BuiltinByteString bs <- fromBuiltinData d
        reoption $ toBuiltin <$> BLS12_381.G2.uncompress bs
```

or is it annoying to make normal Haskell stuff work here properly?

effectfully (approved):
LGTM

I've created PLT-6538 for myself to write tests so that we don't forget to add these instance again. I'm going to it after PLT-1423 (which is in progress currently), because it'll make it easier, I think.

plutus-tx/src/PlutusTx/IsData/Class.hs
kwxm: I did spend a while looking for some standard function(s) that would help with that then gave up because the only things I could find were in slightly obscure libraries.  I had forgotten/didn't know about `reoption`, but maybe it's better to just be explicit, especially given that this will be used on the chain.

PR #5420, opened 2023-07-12
Fix safeToBetaReduce in UPLC inliner and update uplc test.

thealmarty:
Separating this out to see the changes in tests more easily.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-core/untyped-plutus-core/test/Transform/callsiteInline.uplc.golden
```diff
@@ -2,7 +2,7 @@
   (lam
     a_5
     [
-      [ f_1 [ [ a_5 (con integer 0) ] (con integer 1) ] ]
+      [ f_1 [ [ g_2 (con integer 1) ] (con integer 0) ] ]
```
michaelpj, line 5: okay, now _this_ looks busted. That's a free variable where we didn't have one before.

michaelpj: Oh wait, `g` is free in the input! and a is bound to g. So this is actually fine, we've just inlined a, which seems okay

thealmarty:
Unfortunately the script context sizes increased. :( Also, in `no-fib-test` and `script-contexts-test` some of the `SizeTest` failed. :(

michaelpj:
Well, it's going to change how much we inline, so I guess it's a toss up between getting better and worse!

plutus-core/untyped-plutus-core/test/Transform/callsiteInline.uplc.golden
thealmarty: I'm not sure I understand. These two outputs are semantically different, I think. Also I don't know why we say *inline* `a` here, I think this is beta reduction?

michaelpj: Sorry, a is not bound to g. a is bound to `\x y -> g y x`. This is UPLC, so it's an immediately applied lambda, see `a` at line 3 and the definition on line 9. We've indeed _inlined_ one occurrence of `a`, replacing it with `\x y -> g y x`, and then reducing that lambda. We haven't beta reduced the original application - it's still there!

thealmarty: I see that now. Thanks!

thealmarty: I guess we just didn't notice that `a` wasn't inlined before! Great news that this is more right now.

michaelpj (approved):

PR #5421, opened 2023-07-13
Test Marlowe benchmarks in CI

kwxm:
Just checking whether the new Marlowe benchmarks work in CI.


michaelpj:
I mean, those benchmark names are kind of terrible. Maybe we can do something about that @thealmarty ? Even just numbering them `role1` etc. would help

thealmarty:
> I mean, those benchmark names are kind of terrible. Maybe we can do something about that @thealmarty ? Even just numbering them `role1` etc. would help

It *should* have the transaction ids like the semantics ones. I'll try fix it.

PR #5422, opened 2023-07-14
Bump lower bound on cardano-crypto-class

michaelpj:
This is the version that actually added the BLST support.

### DIFF
--- a/plutus-core/plutus-core.cabal
+++ b/plutus-core/plutus-core.cabal
@@ -263,7 +263,7 @@ library
     , bimap
     , bytestring
     , cardano-crypto
-    , cardano-crypto-class        ^>=2.1.0.2
+    , cardano-crypto-class        ^>=2.1.2
     , cassava
     , cborg
     , composition-prelude         >=1.1.0.1


plutus-core/plutus-core.cabal
```diff
@@ -263,7 +263,7 @@ library
     , bimap
     , bytestring
     , cardano-crypto
-    , cardano-crypto-class        ^>=2.1.0.2
+    , cardano-crypto-class        ^>=2.1.2
```
kwxm, line 266: I was tempted to leave a massive multi-page review just to be annoying, but OK, I'm willing to believe this.

kwxm (approved):

plutus-core/plutus-core.cabal
michaelpj: My evidence is https://github.com/input-output-hk/cardano-haskell-packages/pull/291

kwxm: So why did it work?  It just happened to find the most recent version?

michaelpj: Yep! The issue was that if you don't have blst installed, cabal will just go looking for a version that doesn't require it, and this lets it pick an old version, which doesn't compile for other reasons. So this mostly just forces the solver error about missing blst.

PR #5423, opened 2023-07-14
Fix rolepayout benchmarks.

thealmarty:
This should fix the problem of the role payout benchmarks having bad names, as [promised](https://github.com/input-output-hk/plutus/pull/5421#issuecomment-1636000977).

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-benchmark/marlowe/bench/Bench.hs
+++ b/plutus-benchmark/marlowe/bench/Bench.hs
@@ -7,6 +7,7 @@ import Criterion.Main (Benchmark, Benchmarkable, bench, bgroup, defaultMainWith)
 
 import PlutusBenchmark.Common (benchTermCek, getConfig)
 import PlutusBenchmark.Marlowe.BenchUtil (rolePayoutBenchmarks, semanticsBenchmarks)
+import PlutusBenchmark.Marlowe.Scripts.RolePayout (rolePayoutValidator)
 import PlutusBenchmark.Marlowe.Scripts.Semantics (marloweValidator)
 import PlutusBenchmark.Marlowe.Types qualified as M
 import PlutusCore.DeBruijn.Internal (NamedDeBruijn)
@@ -66,7 +67,7 @@ main = do
       fmap (uncurriedBench . mkBenchmarkable marloweValidator) semanticsMBench
     rolePayoutBench :: [Benchmark] -- list of criterion role payout Benchmarks
     rolePayoutBench =
-      fmap (uncurriedBench . mkBenchmarkable marloweValidator) rolePayoutMBench
+      fmap (uncurriedBench . mkBenchmarkable rolePayoutValidator) rolePayoutMBench
 
   -- Run each benchmark for 5 secs by default. This benchmark runs on the longitudinal
   -- benchmarking flow so we don't want to set it higher by default. One can change this with -L or
--- a/plutus-benchmark/marlowe/bench/Bench.hs
+++ b/plutus-benchmark/marlowe/bench/Bench.hs
@@ -75,4 +75,4 @@ main = do
   config <- getConfig 5.0
   defaultMainWith config [
     bgroup "semantics" semanticsBench
-    , bgroup "role payout" rolePayoutBench]
+    , bgroup "role-payout" rolePayoutBench]
--- a/plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Core/V1/Semantics.hs
+++ b/plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Core/V1/Semantics.hs
@@ -32,24 +32,11 @@
 -----------------------------------------------------------------------------
 
 
-{-# LANGUAGE DataKinds                  #-}
-{-# LANGUAGE DefaultSignatures          #-}
-{-# LANGUAGE DeriveAnyClass             #-}
-{-# LANGUAGE DeriveDataTypeable         #-}
-{-# LANGUAGE DeriveGeneric              #-}
-{-# LANGUAGE DerivingVia                #-}
-{-# LANGUAGE FlexibleContexts           #-}
-{-# LANGUAGE FlexibleInstances          #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE ImportQualifiedPost        #-}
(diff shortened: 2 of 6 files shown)


kwxm (commented):
I think this still won't work.  The awk script that does the benchmark results comparison assumes that the benchmark names don't contain any whitespace.  It [gets the benchmark names](https://github.com/input-output-hk/plutus/blob/a25bbdbd51ccb18e9650fd67149b20d398134aee/plutus-benchmark/bench-compare-markdown#L39) from the lines where Cirterion prints out "benchmarking ..." by looking at the second field (`$2)`on the line, where it regards fields as being separated by whitespace.  When I run the benchmarks locally I get things like
```
benchmarking role payout/121a0a1b12030616111f02121a0e070716090a0e031c071419121f141409031d
```
so it will think that the benchmark name is "role".  To fix this we can either allow spaces in benchmark names and modify the awk script to take the whole of the rest of the line after `benchmarking " to be the name, or change the [name](https://github.com/input-output-hk/plutus/blob/8998a16a2ddcc08ae2edfe490df58f55d871aa44/plutus-benchmark/marlowe/bench/Bench.hs#L78) to be "role-payout/xxxx..." or something.

Note also that changes like this won't immediately show up in the output of `/benchmark` because the script just takes the names from the first run of the benchmarks and ignores the names in the second run (so the benchmarks in the two runs had better match up) so even if we change the names in this branch it'll still use the old ones from `master`.  

It might also make the output more readable if we just took the first few digits from the big hex string in the script name instead of the whole thing.  Incidentally, do you know what these numbers represent? I'd assumed that they were hashes but the first few [here](https://github.com/input-output-hk/plutus/tree/thealmarty/fix-marlowe-bench-bug/plutus-benchmark/marlowe/scripts/rolepayout) surely can't be hashes.

kwxm:
I notice that a number of the Haskell files contain [a lot of pragmas](https://github.com/input-output-hk/plutus/blob/8998a16a2ddcc08ae2edfe490df58f55d871aa44/plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Scripts/RolePayout.hs#L18) at the start and they're surely not all necessary.  Can we get rid of the extra ones, or is this some standard set of pragmas that we're inheriting from the Marlowe code?

thealmarty:
Thanks Kenneth!

> so it will think that the benchmark name is "role". To fix this we can either allow spaces in benchmark names and modify the awk script to take the whole of the rest of the line after `benchmarking " to be the name, or change the [name](https://github.com/input-output-hk/plutus/blob/8998a16a2ddcc08ae2edfe490df58f55d871aa44/plutus-benchmark/marlowe/bench/Bench.hs#L78) to be "role-payout/xxxx..." or something.

I used your last suggestion.
> 
> Note also that changes like this won't immediately show up in the output of `/benchmark` because the script just takes the names from the first run of the benchmarks and ignores the names in the second run (so the benchmarks in the two runs had better match up) so even if we change the names in this branch it'll still use the old ones from `master`.
> 

So how do we fix this? I don't think the script should do that right? It should always check the names because we may change the code.

> It might also make the output more readable if we just took the first few digits from the big hex string in the script name instead of the whole thing. Incidentally, do you know what these numbers represent? I'd assumed that they were hashes but it seems very unlikely that he first few [here](https://github.com/input-output-hk/plutus/tree/thealmarty/fix-marlowe-bench-bug/plutus-benchmark/marlowe/scripts/rolepayout) can be hashes.

I thought they are transaction IDs, I have no idea in what form. I don't think it'll be more readable if I just cut off the first few digits but I don't know how to make it more readable.

thealmarty:
hm...I guess I just need to merge this to see the updated names. I'll remove some more pragmas first.

kwxm:
> So how do we fix this? I don't think the script should do that right? It should always check the names because we may change the code.

I think that would be rather difficult.  If we rename benchmarks for example then there's no way to tell which old names correspond to which new ones, so the script would just have to fail.  I think the best that we could do would be to construct a mapping from benchmark names to execution times for both sets of benchmarks (one mapping for each set of benchmarks) then check that we have exactly the same set of names for both and match up the results and print out the statistics for the matching pairs.  If some name exists in one set but not in the other then it'd have to issue an error message (so in this case we'd get an error for every role-payout example). Even doing that would add quite significant complexity to the script, and there will probably still be situations we won't be able to deal with, like if we rename benchmark A to B and B to A (which would be a silly thing to do, but I suppose it could happen).  It is pretty obvious if something has gone wrong at the moment, like if you re-order the benchmarks then the numbers you get probably just won't make sense.

What do the longitudinal benchmarks do if the benchmarks are run in a different order or if you add some new ones?

thealmarty:
> > So how do we fix this? I don't think the script should do that right? It should always check the names because we may change the code.
> 
> I think that would be rather difficult. If we rename benchmarks for example then there's no way to tell which old names correspond to which new ones, so the script would just have to fail. I think the best that we could do would be to construct a mapping from benchmark names to execution times for both sets of benchmarks (one mapping for each set of benchmarks) then check that we have exactly the same set of names for both and match up the results and print out the statistics for the matching pairs. If some name exists in one set but not in the other then it'd have to issue an error message (so in this case we'd get an error for every role-payout example). Even doing that would add quite significant complexity to the script, and there will probably still be situations we won't be able to deal with, like if we rename benchmark A to B and B to A (which would be a silly thing to do, but I suppose it could happen). It is pretty obvious if something has gone wrong at the moment: if you re-order the benchmarks for example then the numbers you get probably just won't make sense.
> 

Good point.

> What do the longitudinal benchmarks do if the benchmarks are run in a different order or if you add some new ones?

Good question! No idea!

thealmarty:
> The reason I mentioned that is because on my screen the table is too narrow and I have to go to the bottom and scroll horizontally to see the numbers. I don't have any good suggestions for fixing it though!

The reason I'm a little hesitant to trim it down is because then it's not really anything anymore. Right now it's a transaction ID in some form.

thealmarty:
I've taken out a lot of the pragmas. @kwxm do you want to take another look? Thanks. After it's merged I'll try benchmarking again to make sure it's good.

kwxm (approved):
Yep, this all looks good.  Thanks!

PR #5425, opened 2023-07-17
Revert mistaken change

michaelpj:
This moved a check from phase 2 to phase 1, which we can't do.


effectfully (approved):
We need some principled and paranoid long-term, but short-term this will do.

PR #5426, opened 2023-07-17
Drop fully applied condition in PIR call site inliner

thealmarty:
This replaces part of #5385 - this PR contains the PIR part. #5385 will be closed and be replaced by this and a follow up PR for the UPLC inliner.  

- drop fully applied condition in the PIR callsite inliner. 
- we don't need `splitParam` altogether because we're just pattern matching the `LamAbs` and `TyAbs` as we go as suggested by @michaelpj 
- deleted all tests for `splitParam`
- remove the related fields too in `VarInfo`
- renaming the rhs now happens in `inlineApp`, before it gets passed into `applyAndBetaReduce`. This ensures we're renaming at each callsite, as desired.
- `safeToBetaReduce` is further fixed to check against the whole term instead of just the function body. 
- removed `letOverAppType2` test because it is a duplicate

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-core/plutus-ir/test/transform/inline/letTypeAppMultiNotSat.golden
```diff
@@ -3,5 +3,5 @@
   (termbind
     (strict) (vardecl idFun (all a (type) (fun a a))) (abs a (type) (lam x a x))
   )
-  { idFun (con integer) }
+  [ (lam x (fun (con integer) (con integer)) x) (lam x (con integer) x) ]
```
thealmarty, line 6: 🆗

plutus-core/plutus-ir/test/transform/inline/letTypeAppMulti.golden
```diff
@@ -3,5 +3,8 @@
   (termbind
     (strict) (vardecl idFun (all a (type) (fun a a))) (abs a (type) (lam x a x))
   )
-  [ [ (builtin addInteger) (con integer 3) ] (con integer 3) ]
+  [
+    [ (builtin addInteger) [ (lam x (con integer) x) (con integer 3) ] ]
+    [ (lam x (con integer) x) (con integer 3) ]
+  ]
```
thealmarty, line 9: 🆗

thealmarty: it seems somehow the applications get reduced further in most cases before the code change. I'm not sure why it is the case but the simplification pass should take care of that.

plutus-benchmark/bls12-381-costs/test/bls12-381-costs.golden
```diff
@@ -103,7 +103,7 @@ VRF example
 
     n     Script size             CPU usage               Memory usage
   ----------------------------------------------------------------------
-    -     783   (4.8%)      1547576705  (15.5%)          400191   (2.9%)
```
thealmarty, line 106: @kwxm any idea why dropping the fully applied condition for inlining functions in the PIR inliner could increase the script size? We do do checks that size shouldn't increase before inlining. So I wonder if we've missed something?

plutus-core/plutus-ir/test/transform/inline/letOverAppType2.golden
```diff
@@ -3,13 +3,11 @@
...
-    )
-    [ funApp (con integer 3) ]
-  )
+  [
+    [
+      (lam x (con integer) x)
+      [ (lam x (con integer) x) (lam x (con integer) x) ]
+    ]
+    (con integer 3)
+  ]
```
thealmarty, line 12: It seems `funApp` was not inlined before but now it is. As the original comment said, `funApp` should be inlined by unconditional inlining.

zliu41:
> However, I'm not sure why idFun in letFunInFunMulti isn't inlined?

I think it's because `\x -> x` is bigger than `idFun`.
Of course, if we inline both callsites of `idFun`, then we can remove the definition of `idFun`, which may ultimately reduce the size. But I don't think this is a factor the inliner currently considers.

thealmarty:
> > However, I'm not sure why idFun in letFunInFunMulti isn't inlined?
> 
> I think it's because `\x -> x` is bigger than `idFun`. Of course, if we inline both callsites of `idFun`, then we can remove the definition of `idFun`, which may ultimately reduce the size. But I don't think this is a factor the inliner currently considers.

OK thanks that was what I thought too but then I got confused because `idFun` when it's fully applied gets inlined. I guess it looks at the simplied term `n`, not `(lam x x) n`.

zliu41:
> I guess it looks at the simplied term n, not (lam x x) n.

Yes - this is why we are able to remove the fully-saturated condition. Before we made the inliner to look at the simplified term, the fully-saturated condition was necessary.

plutus-benchmark/bls12-381-costs/test/bls12-381-costs.golden
zliu41: Yeah, that sounds suspicious. Is there a golden test where the size increases?

thealmarty:
@input-output-hk/plutus-core folks, this still needs tests but I'd appreciate feedback already - the implementation is done I hope.

plutus-benchmark/bls12-381-costs/test/bls12-381-costs.golden
thealmarty: Not so far, or maybe I don't know how to find them.

thealmarty: OK we found out this was just from `master`.

michaelpj (commented):
I can't see the bug but I think there must be one :thinking:

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -103,58 +54,51 @@ not have side effects; (2) all arguments be pure, since otherwise it is unsafe t
...
-              acc' <-
-                termSubstNamesM
-                  (\n -> if n == param then Just <$> PLC.rename arg else pure Nothing)
-                  acc
-              go acc' arity' args'
+              acc' <- do
+                termSubstNamesM -- substitute the term param with the arg in the function body
+                  -- rename before substitution to ensure global uniqueness
+                  (\tmName -> if tmName == n then Just <$> PLC.rename arg else pure Nothing)
+                  (getFnBody acc) -- drop the beta reduced term lambda
```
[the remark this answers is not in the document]
thealmarty: Returning `tm` would be wrong I think, because then nothing will ever be substituted! We need the `acc` with one lambda dropped to get the processed part.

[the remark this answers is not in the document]
michaelpj: You're doing `case` precisely on `acc`, no?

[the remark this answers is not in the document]
thealmarty: Oh you're right this is the input actually.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -164,25 +108,28 @@ fullyApplyAndBetaReduce info args0 = do
...
   | (Var _ann name, args) <- splitApplication t =
       gets (lookupVarInfo name) >>= \case
-        Just varInfo ->
-          fullyApplyAndBetaReduce varInfo args >>= \case
-            Just fullyApplied -> do
-              let rhs = varRhs varInfo
-                  -- Inline only if the size is no bigger than not inlining.
-                  sizeIsOk = termSize fullyApplied <= termSize t
+        Just varInfo -> do
+          -- rename the rhs of the variable before any substitution
```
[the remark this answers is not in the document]
michaelpj: Why can't we do it when we decide "yes let's inline it"? And it's not a variable? It's the RHS of the binding which could be anything, right?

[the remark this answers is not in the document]
thealmarty: Sorry I didn't mean variable, I meant it's just the rhs of the binding, not the whole term. 

`rhs` (which is renamed) is passed on to `applyAndBetaReduce` to give us the processed application. Did you mean I should rename in `applyAndBetaReduce`? I don't think that'll save anything if I had to do it right away there anyway.

plutus-core/plutus-ir/test/transform/inline/letFunConstMulti.golden
```diff
@@ -5,5 +5,5 @@
     (vardecl constFun (fun (con integer) (fun (con integer) (con integer))))
     (lam x (con integer) (lam y (con integer) x))
   )
-  [ constFun (con integer 3) ]
+  [ constFun [ (lam y (con integer) (con integer 3)) (con integer 5) ] ]
```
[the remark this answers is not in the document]
thealmarty: right, looking into it.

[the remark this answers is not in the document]
zliu41: This is actually the right before. Terms are processed bottom-up, so `constFun 3` is processed first. It has a bigger size then `\y -> 3`, so `constFun` is inlined here, yielding `(\y -> 3) 5`. The `5` will be eliminated by dead binding elimination in the next round.

plutus-core/plutus-ir/test/transform/inline/letTypeAppMultiNotSat.golden
zliu41: It doesn't seem like this is bigger. The size of `{ idFun (fun (con integer) (con integer)) }` is 5, and so is the size of `(lam x (fun (con integer) (con integer)) x)`.

michaelpj: Ah right, I wasn't looking at the _original_ term.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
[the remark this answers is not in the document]
michaelpj: I'm not sure, I'd have to stare at it more. I just notice that we're not always going to inline the term, but we do always do the renaming work. So it seems to me that we might want to try and delay as much work as possible until we've decided that we _are_ going to inline it.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -103,58 +54,47 @@ not have side effects; (2) all arguments be pure, since otherwise it is unsafe t
...
-                termSubstNamesM
-                  (\n -> if n == param then Just <$> PLC.rename arg else pure Nothing)
-                  acc
-              go acc' arity' args'
-            else pure Nothing
-        (TypeParam param : arity', TypeAppContext arg _ args') -> do
+              acc' <- do
+                termSubstNamesM -- substitute the term param with the arg in the function body
+                  -- rename before substitution to ensure global uniqueness
+                  (\tmName -> if tmName == n then Just <$> PLC.rename arg else pure Nothing)
```
michaelpj, line 82: This is actually another example: do we need to rename here? If the input is globally unique, then I think this will just give us something with local uniqueness... which is fine so long as we then rename it before we substitute it into the original program.

OTOH this reasoning is a bit tricky and maybe it's better to just stick on the safe side and rename too much.

plutus-core/plutus-ir/test/transform/inline/letOverAppType3.golden
```diff
@@ -3,5 +3,11 @@
...
     (strict) (vardecl idFun (all a (type) (fun a a))) (abs a (type) (lam x a x))
   )
-  (con integer 3)
+  [
+    (lam x (con integer) x)
+    [
+      [ (lam x (fun (con integer) (con integer)) x) (lam x (con integer) x) ]
+      (con integer 3)
+    ]
+  ]
```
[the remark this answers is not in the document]
michaelpj: Actually, I think these examples are a much clearer example of something weird happening. I think what's happened is that we've decided to inline `const`... but as a 0-arity term, so we don't do any beta-reduction! And that's pretty weird! Surely it would be even smaller if we did the beta-reduction? So why are we not taking that path?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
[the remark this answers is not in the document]
thealmarty: In Ziyang's PR I did raise the point of checking `costIsOK` before `applyAndBetaReduce`. Actually Roman agrees with that too. I'll make that change here.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -164,25 +104,28 @@ fullyApplyAndBetaReduce info args0 = do
...
-                  -- Inline only if the size is no bigger than not inlining.
-                  sizeIsOk = termSize fullyApplied <= termSize t
+        Just varInfo -> do
+          -- rename the rhs of the variable before any substitution
+          rhs <- liftDupable (let Done rhs = varRhs varInfo in rhs)
+          applyAndBetaReduce rhs args >>= \case
+            Just applied -> do -- there is beta reduction of the application
+              let -- Inline only if the size is no bigger than not inlining.
+                  sizeIsOk = termSize applied <= termSize t
                   -- The definition itself will be inlined, so we need to check that the cost
```
michaelpj, line 129: This logic is no longer correct now that we do beta reduction also. It's not totally clear to me what the correct logic for the cost actually is now :thinking:

thealmarty:
I found that the inliner was not doing what we thought it was doing. [inlineApp](https://github.com/input-output-hk/plutus/blob/0c501c714bda73328f98c8a70b47be935474be1f/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs#L225) takes in a subterm - so e.g., in `letFunConstMulti`: `constFun (constFun 3 5)` has subterms `Apply constFun 3`, and `5`. So with `Apply constFun 3`, `splitApplication` gives a context of `3`. I.e., [appCtx](https://github.com/input-output-hk/plutus/blob/0c501c714bda73328f98c8a70b47be935474be1f/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs#L75) always has at most 1 `arg` in it - this shows in other tests as well. So [applyAndBetaReduce](https://github.com/input-output-hk/plutus/blob/0c501c714bda73328f98c8a70b47be935474be1f/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs#L68) doesn't get a chance to look at multiple `apply` like we thought, and so [sizeIsOk](https://github.com/input-output-hk/plutus/blob/0c501c714bda73328f98c8a70b47be935474be1f/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs#L128) will not be accurate, it will only look at the `termSize` after applying 1 argument. So I will alter the implementation such that `inlineApp` can process the whole chain of applications.

michaelpj:
> inlineApp takes in a subterm

Sorry, I don't understand this? The line you linked does two things:
- Recursively process the subterms
- Give the resulting term to `inlineApp`

So we work bottom up, but that doesn't mean that `inlineApp` will never see a multi-application. I believe you that something is going wrong, but I don't think this can be it!

zliu41:
The PR generally looks OK to merge, except that I still don't understand the cause of the free variable error and how it was fixed. From what I heard, this line was the problem:

```haskell
rhsBody <- liftDupable (let Done rhsBody = varRhsBody info in rhsBody)
```

But why? How can renaming create free variables? Is there a bug in the renamer?

zliu41:
Wait, what happened to `letFunConstMulti.golden`? It used to be changed by this PR, but it is no longer changed? How so?

plutus-core/plutus-ir/test/transform/inline/letFunConstMulti.golden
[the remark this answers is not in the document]
zliu41: > Okay, that does make sense, but I would also expect the other constFun to be inlined, no?

No, because of the types.
```
size (constFun 3) = 3
size (\(y :: Integer) -> 3)) = 3

size (constFun ((\(y :: Integer) -> 3) 5)) = 7
size (\(y' :: Integer -> Integer) -> ((\(y :: Integer) -> 3) 5)) = 9
```

The `Integer -> Integer` adds 3 to the size.

zliu41:
> [appCtx](https://github.com/input-output-hk/plutus/blob/0c501c714bda73328f98c8a70b47be935474be1f/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs#L75) always has at most 1 arg in it

That's not true. If we don't inline `constFun` when processing `constFun 3`, then it will subsequently process `constFun 3 5`, where `appCtx` will have 2 args.

effectfully:
> How can renaming create free variables? Is there a bug in the renamer?

Renaming cannot create and does not change free variables, unless you use some internal function (neither `rename` nor `liftDupable` is internal) or there's indeed some horrible bug in the renamer.

> But why? 

I thought I understood that, but now that I've spent another hour staring at the code, I'm back to being puzzled.

I do strongly believe that renaming the whole function is necessary simply because if you're going to inline a partial application of a function, you'd better ensure that the inlined uninstantiated lambdas are renamed, so that you don't break global uniqueness by inlining the same lambda twice. I'm quite puzzled as to why this causes free variable errors, though.

However I've seen cases when breaking global uniqueness for passes relying on it was causing free variable errors in the past. I've never bothered to investigate them properly, since fixing the uniqueness bugs was always resolving the errors.

effectfully:
I guess, the obvious question I should've asked long ago is "what errors are we talking about?".

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -212,17 +212,21 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
+            consider that in the future.
+
+            Ideally, we would process all subterms first, so that the rhs won't be processed more
+            than once. This is important because otherwise the number of times we process them can
+            grow exponentially in the case that it has nested `let`s.
+            However, if we do that then `CallSiteInline.inlineApp` would be given a subterm of each
+            application node, instead of a whole application chain, and it would not reduce
+            properly.-}
+            t' <- inlineApp t
+            forMOf termSubterms t' processTerm
```
michaelpj, line 229: Note for reviewers, this was the fix for the last set of problems. The issue is that the old behaviour would process the subterms first, so would always look at the _prefix_ of an application before the full application. This often led to it not beta-reducing enough. The problem didn't arise before because we only ever inlined fully-saturated applications, so it didn't matter that we looked at the prefix first.

michaelpj: Aha, I think I see a sensible way out:
- Split the term into an application context
- Process the sub-parts of the _application context_
- Now consider inlining the application context

plutus-core/plutus-ir/test/transform/inline/letOverAppType2
```diff
@@ -1,53 +0,0 @@
-{-
```
michaelpj, line 1: why is this gone?

plutus-core/plutus-ir/test/transform/inline/letOverAppType.golden
```diff
@@ -3,5 +3,8 @@
   (termbind
     (strict) (vardecl idFun (all a (type) (fun a a))) (abs a (type) (lam x a x))
   )
-  (con integer 3)
+  [
```
michaelpj, line 6: regression?

plutus-core/plutus-ir/test/transform/inline/letOverAppMulti.golden
```diff
@@ -17,6 +17,6 @@
       )
       (lam x (fun (con integer) (con integer)) idFun)
     )
-    (con integer 6)
+    [ [ k idFun ] (con integer 6) ]
```
michaelpj, line 20: regression?

michaelpj:
It still looks to me like there are some regressions.

michaelpj:
I am pretty sure that the free variable errors were just caused by substituting in the wrong thing. I agree with Roman that there is no way that our renamer can create new free variables.

plutus-core/plutus-ir/test/transform/inline/letOverAppType2
thealmarty: As I mentioned this is a duplicate to `letOverAppType`.

thealmarty:
@zliu41 regarding the free-variable error: as I explained, the renaming in the original code does nothing, because it's renaming the function body only, without any lambdas, so everything is free, and nothing gets changed. 

The free variable errors happened when I tried to drop the fully saturated condition in the original code. I truly cannot say what fixed it at this point, because I have fixed some bugs (including renaming more properly) and changed other things in between. I agree with Roman, I think to go back to investigate that is too high a cost and may not get us any answers.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
thealmarty: Trying. Although this probably wouldn't fix the regressions.

zliu41:
I really want to see an explanation of the free variable error (with an example) that makes sense. I can't figure out how it can possibly happen in my head (e.g., how is it possible that we substitute the wrong thing? Under which circumstances can this happen?), and none of what anyone said so far helped me understand it better. Next time please do a more detailed investigation when something like this happens.

plutus-core/plutus-ir/test/transform/inline/letOverAppMulti.golden
zliu41: This seems indeed a regression. I'm not sure why `k` is not inlined here. `k idFun` reduces to `idFun`, which is clearly a win.

zliu41 (commented):
> Wait, what happened to `letFunConstMulti.golden`? It used to be changed by this PR, but it is no longer changed? How so?

I still have this question - the file was changed earlier, but somehow a later commit restored it and it's unclear why. It seems to me that the earlier version (the changed one) was correct.

zliu41:
>  It seems to me that the earlier version (the changed one) was correct.

Did a later commit change either how many iterations the inliner runs, or the order in which the subterms are processed? Given

```
let constFun x y = x in constFun (constFun 3 5)
```

It should first process `constFun 3`, and decide to inline `constFun`, yielding

```
let constFun x y = x in constFun ((\y -> 3) 5)
```

and that should be what's in `letFunConstMulti.golden` (and indeed that was the case earlier).

thealmarty:
> I really want to see an explanation of the free variable error (with an example) that makes sense. I can't figure out how it can possibly happen in my head (e.g., how is it possible that we substitute the wrong thing? Under which circumstances can this happen?), and none of what anyone said so far helped me understand it better. Next time please do a more detailed investigation when something like this happens.

The logging I proposed would help. We could go back and investigate if you wish. The error occurred somewhere in this commit https://github.com/input-output-hk/plutus/pull/5385/commits/fd964bc7960bd038f8d986fc7570540ab3d69e94

thealmarty:
> > It seems to me that the earlier version (the changed one) was correct.
> 
> Did a later commit change either how many iterations the inliner runs, or the order in which the subterms are processed? Given
> 
Yes, this commit https://github.com/input-output-hk/plutus/pull/5426/commits/c67350621b8285285b98374b7960e36200ff9233 makes it so that the whole term is passed to `inlineApp` first, then the traversal happen. See my comments in the code too.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -212,17 +212,21 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
-            -- exponentially in the case that it has nested `let`s.
-            --
-            -- Then, consider call site inlining for each node that have gone through unconditional
-            -- inlining. Because `inlineSaturatedApp` traverses *all* application nodes for each
-            -- subterm, the runtime is quadratic for terms with a long chain of applications.
-            -- If we use the context-based approach like in GHC, this won't be a problem, so we may
-            -- consider that in the future.
-            inlineSaturatedApp =<< forMOf termSubterms t processTerm
+        t -> do
+            {- We consider call site inlining when we see an application node, then process the term
```
zliu41, line 218: This is not accurate - we don't do callsite inlining before unconditional inlining (and we shouldn't). Unconditional inlining is still being done first - if a binding should be unconditionally inlined, then it won't get to the callsite inliner.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -212,17 +212,21 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
+            for each subterm, the runtime is quadratic for terms with a long chain of applications.
+            If we use the context-based approach like in GHC, this won't be a problem, so we may
+            consider that in the future.
+
+            Ideally, we would process all subterms first, so that the rhs won't be processed more
+            than once. This is important because otherwise the number of times we process them can
+            grow exponentially in the case that it has nested `let`s.
+            However, if we do that then `CallSiteInline.inlineApp` would be given a subterm of each
+            application node, instead of a whole application chain, and it would not reduce
+            properly.-}
```
zliu41, line 229: Again, I don't know why this is a problem.

zliu41 (commented):
While it is probably fine to make the inliner work in top-down order, I don't see the necessity of doing so. I think it's more natural to process subterms first.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
michaelpj: Yeah, I think this should just stick to saying "process the subterms" and drop the bit saying it's unconditional inlining, which it isn't. It's running the whole combined inliner on the subterms.

michaelpj:
> I really want to see an explanation of the free variable error (with an example) that makes sense. I can't figure out how it can possibly happen in my head (e.g., how is it possible that we substitute the wrong thing?

My best guess (and it is a guess) is that:
- Before this PR we were substituting in `varDef`, which is the body without _all_ of the lambdas.
- In this PR we sometimes don't substitute in all the arguments
- At some point there was a state where we were substituting in `varDef` still but without all the arguments substituted, which led to free variable errors. I think getting rid of `varDef` probably led to doing the right thing.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
zliu41: But `(\y -> 1) 2` is not a problem, right? It will be reduced to 2 in the next iteration of the simplifier.
Yes, it requires one more iteration in this instance. But simplifying a term without simplifying its subterms can also lead to needing additional iterations.

michaelpj: I don't know if it's a problem, but it very much seems to me to be not what the code is _trying_ to do. It's pretty counter-intuitive and I had to think about it a lot. Why would we _not_ reduce the longest application that we can, given we have all the machinery for reducing multiple applications right there!

zliu41: Ok, fair enough.

thealmarty:
Thanks for the comments. I'll be working on the Marlowe PR and will come back to this.

thealmarty:
It seems `safeToBetaReduce` is still causing problems (the regressions).

plutus-core/plutus-ir/test/transform/inline/letOverAppType.golden
thealmarty: Nothing got through to `applyAndBetaReduce` in this case - that means that `k` and `funApp` got unconditionally inlined, and they were never added to the variable map, so `inlineApp` went to the `otherwise` branch. I have removed call site inlining in https://github.com/input-output-hk/plutus/commit/d4f68656860b834c813af5080f5d6f7a9d4ab45a and verified that `k` did indeed get inlined by unconditional inlining. I'll update the comment in the test.

zliu41: This test case is supposed to test over-application (as its name suggests). If everything is unconditionally inlined, then we aren't testing that, and so the test case should either be removed or updated. I wonder why it wasn't unconditionally inlined before, but it is now - there should be no change to the unconditionally inlining condition?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: @zliu41 any thought on this? I think maybe we don't need to check the cost at all? We already made sure it's safe to beta reduce - ie we made sure no effects etc.

zliu41: It is still needed, and I think it should still be correct.

> We already made sure it's safe to beta reduce 

That's different. Consider

```
let f = 1+2+3+...+1000
 in ...f...f...
```

Here `f` is not even a function, so `safeToBetaReduce` does not apply. But we can't inline `f`, because it will duplicate the costly computation (in this case, the size of `f` is also too big, but it is possible for size to be small but cost to be big).

thealmarty: I believe currently we don't inline non-functions in the call site inliner. See [here](https://github.com/input-output-hk/plutus/blob/99d4253dc071c68aa2dcaa3746d2feae2118a10f/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs#L98). So that is not a relevant case? Any function (with term arguments) would go through `safeToBetaReduce`.

thealmarty: right but there's no actual substitution in this case.

zliu41: The question here is whether checking the cost is needed. There is no substitution, but I think it is still needed, because of my example.

michaelpj:
How do you know it's that?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: Here we are talking about the check for the call site inliner. If you look at the line I [linked](https://github.com/input-output-hk/plutus/blob/99d4253dc071c68aa2dcaa3746d2feae2118a10f/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs#L98), no arguments are substituted, if the variable is a non-function to begin with. That was what I meant.

I can try with a test yes but I won't go up to 1000 lol.

thealmarty: But yeah I see it will still look at that because it's returned as `applied`.

thealmarty: but looking at that maybe wrong in the function case if it has been beta reduced, like Michael said. So maybe we should deal with non-function separately afterall?

thealmarty: of course, we could just keep the check as is and inline less.

thealmarty: maybe we can check `costIsAcceptable applied` instead?

thealmarty:
> How do you know it's that?



> How do you know it's that?

@michaelpj from tracing. It shows that it thinks it's not effect safe to inline `k` in some tests. I'm still working on tracing more and will give a more detailed report soon I hope.

plutus-core/plutus-ir/test/transform/inline/letOverAppMulti.golden
```diff
@@ -17,6 +17,6 @@
       )
       (lam x (fun (con integer) (con integer)) idFun)
     )
-    [ [ k idFun ] (con integer 6) ]
+    [ idFun (con integer 6) ]
```
thealmarty, line 20: regression not happening anymore, as desired.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: tested in https://github.com/input-output-hk/plutus/pull/5426/commits/127b9d382c3dd7d4ecfd8135c308f25bfcad48d3 not much difference.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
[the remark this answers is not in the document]
thealmarty: I've moved the renaming to as late as possible in https://github.com/input-output-hk/plutus/pull/5426/commits/4d9cf3004457ea167f44b52abb3a4763b96ad1ab lmk if you can think of further savings.

plutus-core/plutus-ir/test/transform/inline/letTypeApp2.golden
```diff
@@ -3,5 +3,8 @@
   (termbind
     (strict) (vardecl idFun (all a (type) (fun a a))) (abs a (type) (lam x a x))
   )
-  (con integer 3)
+  [
+    { idFun (con integer) }
+    [ { idFun (con integer) } [ (lam x (con integer) x) (con integer 3) ] ]
+  ]
```
thealmarty, line 9: This is fine I think. `k` and `funApp` are inlined in 1 pass.

plutus-benchmark/script-contexts/test/Spec.hs
```diff
@@ -41,7 +41,7 @@ testCheckSc1 = testGroup "checkScriptContext1"
         compiledCodeToTerm $ mkCheckScriptContext1Code (mkScriptContext 4)
     , testCase "fails on 5" . assertFailed $
         compiledCodeToTerm $ mkCheckScriptContext1Code (mkScriptContext 5)
-    , Tx.fitsInto "checkScriptContext1 (size)" (mkCheckScriptContext1Code (mkScriptContext 1)) 3721
+    , Tx.fitsInto "checkScriptContext1 (size)" (mkCheckScriptContext1Code (mkScriptContext 1)) 3851
```
thealmarty, line 44: some regression here.

plutus-ledger-api/test/Spec/Interval.hs
```diff
@@ -168,8 +190,8 @@ prop_intervalSetContains = property $ do
   let
     c1 = ivl1 `Interval.contains` ivl2
     c2 = s2 `Set.isSubsetOf` s1
-  Hedgehog.cover 30 "True" $ c1
-  Hedgehog.cover 30 "False" $ not c1
+  Hedgehog.cover 10 "True" $ c1
```
thealmarty, line 193: I had to lower this because CI was failing with one case only having 29% (I triggered it a couple times, same result). I think 10 is fine. I see the other ones only require 10 or less also.

plutus-core/plutus-ir/test/transform/inline/letTypeApp2
```diff
@@ -1,21 +1,16 @@
...
-let
+let 
     idFun :: forall a . a -> a
     idFun = \x -> x
     funApp :: Integer -> Integer
     funApp = idFun @Integer
     k :: forall a . a -> a
     k = idFun
-in k @Integer (k @(Integer -> Integer) funApp 3)
+in k @Integer (k @Integer (funApp 3))
```
zliu41, line 13: What's the reason for this change?

plutus-core/plutus-ir/test/transform/inline/letTypeApp2
```diff
@@ -1,21 +1,16 @@
 {- 
 funApp gets inlined because it only occurs once.
-k should not get inlined by the unconditional inline optimization because it occurs more
-than once.
-Both instances of k should get inlined by the optimization that inlines fully applied functions 
-because it is fully applied.
-The first instance of k being inlined verifies that an over-applied function still gets inline with
-type applications.
+Both instances of k get inlined by unconditional inliner.
```
zliu41, line 3: The original purpose of this test is to test callsite inlining. If everything gets inlined unconditionally then it defeats the purpose.

plutus-core/plutus-ir/test/transform/inline/letTypeApp2.golden
zliu41: But `k` and `funApp` are all unconditionally inlined? I thought we wanted to test that the latest implementation of the callsite inliner (which processes arguments first before the application) can inline more things in 1 pass. Which tests cover that?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -212,17 +214,115 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
+                processArgs AppContextEnd = pure AppContextEnd
+
+            case tm of
+                Var _ann name -> do
+                    gets (lookupVarInfo name) >>= \case
+                        Just varInfo -> do
+                            -- process the args if it's a variable that's in the map
+                            processedArgs <- processArgs args
+                            -- consider call site inlining since it's a variable
+                            t' <- callSiteInline t varInfo processedArgs
```
zliu41, line 239: `t` should not be passed to `callSiteInline`. `t` is the original term, but you'll want to pass the one whose arguments have been processed.

plutus-benchmark/script-contexts/test/Spec.hs
zliu41: Any idea why? I can't think of any reason why this PR makes anything bigger or more expensive.

zliu41 (changes requested):

plutus-core/plutus-ir/test/transform/inline/letTypeApp2
thealmarty: Yes the purpose has changed, but I left this here because it's another example of type applications.

plutus-core/plutus-ir/test/transform/inline/letTypeApp2
thealmarty: Frankly I can't remember, I think I wanted to make it type check, since `funApp 3` is an `Integer`.

plutus-core/plutus-ir/test/transform/inline/letTypeApp2.golden
thealmarty: After processing the arguments first, a few of the test cases did inline more things, including `letFunConstMulti`, `letOverAppMulti`, and `letTypeAppMultiNotSat`. (See https://github.com/input-output-hk/plutus/pull/5426/commits/70291fc96bae2f4400627ad78891ac3030430ccb)

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
thealmarty: Done (https://github.com/input-output-hk/plutus/pull/5426/commits/631ddbad8b6e90ba3b9cdcae86103e1939fb17d9), no change in test results.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -212,17 +214,117 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
+            case tm of
+                Var _ann name -> do
+                    gets (lookupVarInfo name) >>= \case
+                        Just varInfo -> do
+                            -- process the args if it's a variable that's in the map
+                            processedArgs <- processArgs args
+                            -- consider call site inlining since it's a variable
+                            t' <-
+                                callSiteInline
+                                    (fillAppContext tm processedArgs) varInfo processedArgs
```
zliu41, line 241: `processedArgs` is already passed in the last argument, so the first argument should not need to be `fillAppContext tm processedArgs`.

plutus-benchmark/script-contexts/test/Spec.hs
thealmarty: I suspect `costIsOk` isn't quite right. I can't think of anything else.

thealmarty: I will do some more experiments changing what `costIsOk` checks.

zliu41: You can also look at the readable PIR before and after (use `goldenPirReadable` to get them). The readable PIRs are fairly readable.

plutus-benchmark/nofib/test/formulaBudget.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 8791690908
-| mem: 36693948})
\ No newline at end of file
+({cpu: 9642046908
+| mem: 40391148})
```
thealmarty, line 2: I've doubled checked, this change is NOT caused by the latest change in `callSiteInline`, it was from the rebasing on master.

plutus-benchmark/script-contexts/test/Spec.hs
thealmarty: Maybe I'm doing something wrong but the golden file from the master branch is the same as the one from this branch.

zliu41: You may have forgot to `-fforce-recomp`.

thealmarty: I just remembered another possible reason for the regression: `safeToBetaReduce` is fixed in this PR also. The last PR that fixed it did also cause some regression.

thealmarty: I'll add some tests to `termSize`. It seems we don't have any, just a couple of doc tests for `kindSize`. These seems to all involve type applications too so I'll make sure to check that.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
thealmarty: actually there is regression when I apply the change https://github.com/input-output-hk/plutus/pull/5426/commits/11d27b41f6a05f3729019603e8023c11c1cc89a9 :( I reverted it for now.

plutus-benchmark/script-contexts/test/Spec.hs
zliu41: > It is inlining some things with bigger size

That "some things" is `t`, which should not be inlined because it increases the size. So it seems the inliner isn't checking the sizes properly.

michaelpj (approved):
LGTM apart from one remaining instance of repeat processing. Good solution to the traversal order problem!

thealmarty:
This commit https://github.com/input-output-hk/plutus/pull/5426/commits/0f3652f49564a1eee92920019ea96ebca31725c9 causes the following error:
```
GHC Core to PLC plugin: Error: Error from the PIR compiler:
                               Error from the PLC compiler:
                               Free type variable at  Ann {annInline = MayInline, annSrcSpans = { no-src-span }} :  a_10124
Context: Compiling expr at "plutus-benchmark-0.1.0.0-inplace-script-contexts-internal:PlutusBenchmark.ScriptContexts:(96,7)-(96,53)"
Error: cabal: Failed to build lib:script-contexts-internal from
plutus-benchmark-0.1.0.0 (which is required by
test:plutus-benchmark-script-contexts-tests from plutus-benchmark-0.1.0.0).
```

Looking into it. @zliu41 you may be interested it's a free type variable.

plutus-benchmark/script-contexts/test/Spec.hs
zliu41: While adding tests is good, I would try to figure out what exactly happened in this case, and then add some tests accordingly (if this is what you are doing, great), rather than just adding some tests and hoping some of them reveal the issue.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -212,17 +214,114 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
+                processArgs AppContextEnd = pure AppContextEnd
+
+            case hd of
+                Var _ann name -> do
+                    gets (lookupVarInfo name) >>= \case
+                        Just varInfo -> do
+                            -- process the args if it's a variable that's in the map
+                            processedArgs <- processArgs args
+                            -- consider call site inlining since it's a variable.
+                            callSiteInline hd varInfo processedArgs
```
zliu41, line 239: I don't think this is correct, and this may be the reason for the free variable error. In addition to processing each arg, you should also need to process `hd`. Otherwise `processTerm` may never be applied to `hd`.

plutus-benchmark/nofib/test/Spec.hs
```diff
@@ -46,7 +46,7 @@ testClausify = testGroup "clausify"
                , testCase "formula3" $ mkClausifyTest Clausify.F3
                , testCase "formula4" $ mkClausifyTest Clausify.F4
                , testCase "formula5" $ mkClausifyTest Clausify.F5
-               , Tx.fitsInto "formula1 (size)" (Clausify.mkClausifyCode Clausify.F1) 1703
+               , Tx.fitsInto "formula1 (size)" (Clausify.mkClausifyCode Clausify.F1) 3444
```
zliu41, line 49: I don't think this is correct; please double check.

zliu41 (changes requested):
Two things remaining:
- fix the logic in `processTerm`
- figure out the reason for the size increase

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
michaelpj: Yes, this is definitely right. In particular, if we have `f a b` and `f` is being unconditionally inlined, then we've already removed the binding for `f` so we might end up with a free variable if we don't inline it.

You should just process all the parts of the application first, before doing anything else.

thealmarty: I was thinking along that line as well about `hd`. Still working on it. Lots of weird things happening still.

thealmarty: Still tempted to do the logging ticket first to be honest!

michaelpj: Yeah, this is getting ridiculous. I really think we should do the logging thing. We are just constantly failing to work out what's going on and it's sucking up all the time.

plutus-benchmark/nofib/test/Spec.hs
thealmarty: Fixed.

thealmarty:
Fixed all errors. Still regression (but less) on the `fitsInto` script context tests. Everything else is improved. Looked at the PIR of the `fitsInto` tests, no obviously inlining of bigger terms like before.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -103,58 +52,47 @@ not have side effects; (2) all arguments be pure, since otherwise it is unsafe t
...
         AppContext tyname name uni fun ann ->
         InlineM tyname name uni fun ann (Maybe (Term tyname name uni fun ann))
-      go acc arity args = case (arity, args) of
-        -- success
-        ([], _) -> pure . Just $ fillAppContext acc args
-        (TermParam param : arity', TermAppContext arg _ args') -> do
-          safe <- safeToBetaReduce param arg
-          if safe
+      go acc args = case (acc, args) of
+        (LamAbs _ann n _ty tm, appCtx@(TermAppContext arg _ args')) -> do
```
zliu41, line 72: `appCtx` is redundant. It is the same as `args`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -103,58 +52,47 @@ not have side effects; (2) all arguments be pure, since otherwise it is unsafe t
...
-        _ -> pure Nothing
+            termSubstTyNamesM -- substitute the type param with the arg
+              (\tyName -> if tyName == n then Just <$> PLC.rename arg else pure Nothing)
+              tm -- drop the beta reduced type lambda
+          go acc' args'
+        -- term/type argument mismatch, don't inline
+        (LamAbs{}, TypeAppContext{}) -> pure Nothing
+        (TyAbs{}, TermAppContext{}) -> pure Nothing
+        -- no more lambda abstraction, just return the processed application
+        (_, appCtx) -> pure . Just $ fillAppContext acc appCtx
```
zliu41, line 95: ditto

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -164,40 +102,48 @@ fullyApplyAndBetaReduce info args0 = do
...
 
--- | Consider whether to inline an application.
-inlineSaturatedApp ::
+-- | Consider inlining a variable. For applications, consider whether to apply and beta reduce.
+callSiteInline ::
   forall tyname name uni fun ann.
   (InliningConstraints tyname name uni fun) =>
+  -- | The term.
+  Term tyname name uni fun ann ->
+  -- | The Rhs of the "head" of the term, already processed.
```
zliu41, line 117: You shouldn't need this. `VarInfo` already contains the RHS of the head.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -212,17 +214,132 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
+                    processedArg <- processTerm arg
+                    processedArgs <- processArgs ctx
+                    pure $ TermAppContext processedArg ann processedArgs
+                processArgs (TypeAppContext ty ann ctx) = do
+                    processedArgs <- processArgs ctx
+                    pure $ TypeAppContext ty ann processedArgs
+                processArgs AppContextEnd = pure AppContextEnd
+            case (hd, args) of
+                -- not an application, just process the subterms
+                (_, AppContextEnd) -> forMOf termSubterms t processTerm
```
zliu41, line 233: This case is redundant

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -212,17 +214,132 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
+                                            -- when we know there's substitution
+                                            rhs = inlineTermToTerm defAsInlineTerm
+                                        -- process the args
+                                        processedArgs <- processArgs args
+                                        maybeInlined <- callSiteInline t rhs varInfo processedArgs
+                                        case maybeInlined of
+                                            Just inlined -> pure inlined
+                                            -- we didn't inline at the call site, just process the
+                                            -- subterms
+                                            Nothing      -> forMOf termSubterms t processTerm
```
zliu41, line 255: Should just be `pure t` here. The subterms have already been processed

zliu41 (commented):
> Still regression (but less) on the fitsInto script context tests.

Dropping the fully applied condition should not cause any size regression at all, however small it is, so we should still understand why it happened.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
thealmarty: I don't think so? The processed args aren't returned when `callSiteInline` returns `Nothing`?

thealmarty: I tried `pure t` and one of the tests (`factorial`) got a free variable error.

michaelpj:
I thought you'd worked out that the regression was due to `safeToBetaReduce`?

> merging this PR enables loggings to be added and the regression can be examined more easily

I don't see why we can't do the logging beforehand if it helps us to debug?

thealmarty:
> I thought you'd worked out that the regression was due to `safeToBetaReduce`?
> 

Unfortunately I later found out that I was not getting correct test results locally. (even though the compiler told me it compiled the relevant files again)

> > merging this PR enables loggings to be added and the regression can be examined more easily
> 
> I don't see why we can't do the logging beforehand if it helps us to debug?

There's quite a bit of changes here that are done, so I'd like to do the logging on top of this, not `master`. I'll basically have to migrate the logging to here anyway if I did it on `master`.

thealmarty:
I have verified that the regressions are caused by the inliner's interaction with other passes. See the test results on [master](https://gist.githubusercontent.com/thealmarty/123f21d1ec32484fe6174b346081170d/raw/4cb5851be601e4d32a9962efdebe74096d449be3/masterSCtest.txt) and for [my branch](https://gist.githubusercontent.com/thealmarty/d8ec0b333e81fefd4d09b416b21d4587/raw/86c7780f85a40b0475a12b0385e5ebafea21dbee/branchSCtests.txt). You will see that there are indeed improvements rather than regressions when other passes are turned off.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -212,17 +213,123 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
+                (Var{}, _) -> do
+                    processedHd <- processTerm hd -- the variable may be unconditionally inlined.
+                    case processedHd of
+                        -- if it didn't get unconditionally inlined/got inlined to another variable
+                        (Var _ name) -> do
+                                gets (lookupVarInfo name) >>= \case
+                                    Just varInfo -> do
+                                        -- process the args
+                                        processedArgs <- processArgs args
+                                        maybeInlined <- callSiteInline t varInfo processedArgs
```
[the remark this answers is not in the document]
zliu41: In fact you don't need to pass a term. You can just pass the size, i.e., `termSize (fillAppContext processedHd processedArgs)`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
zliu41: Oh, not `pure t`, but `pure (fillAppContext processedHd processedArgs)`.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -212,17 +213,123 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
+                                            -- we didn't inline at the call site, just process the
+                                            -- subterms
+                                            Nothing      -> forMOf termSubterms t processTerm
+                                    -- The variable maybe a *recursive* let binding, in which case
+                                    -- it won't be in the map, and we don't process it.
+                                    -- ATM recursive bindings aren't inlined.
+                                    Nothing -> forMOf termSubterms t processTerm
+                        -- if the processed head is not a variable (because it's inlined), just
+                        -- process the subterms
+                        _ ->  forMOf termSubterms t processTerm
```
zliu41, line 252: Here the head of `t` is already processed (i.e., `processedHd`), and `forMof termSubterms ...` will process it again. May not be a problem but I'm curious what @michaelpj thinks.

zliu41 (changes requested):
The size comparison seems wrong.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -212,17 +213,123 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
+                processArgs (TermAppContext arg ann ctx) = do
+                    processedArg <- processTerm arg
+                    processedArgs <- processArgs ctx
+                    pure $ TermAppContext processedArg ann processedArgs
+                processArgs (TypeAppContext ty ann ctx) = do
+                    processedArgs <- processArgs ctx
+                    pure $ TypeAppContext ty ann processedArgs
+                processArgs AppContextEnd = pure AppContextEnd
+            case (hd, args) of
+                (Var{}, _) -> do
```
michaelpj, line 231: Why are you doing this? I don't think it adds anything? We can just always process the head and the args and then put them back together if the processed head isn't a variable.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
michaelpj: I'm indeed a bit confused. This looks rather different to what Marty and I came up with when we paired on it and I'm not sure why? I was expecting something like:
```
case args of
  -- not really an application, so hd is the term itself. Processing it will loop.
  -- I feel like we came up with a nicer way of handling this but I've forgotten
  AppContextEnd -> forMOf termSubterms t processTerm
  _ -> do 
    hd' <- processTerm head
    args' <- processArgs args
    let reconstructed = fillAppCtx hd' args'
    case hd' of
      (Var _name) -> do
           ...
           case maybeInlined oof
              Just inlined -> pure inlined
              Nothing -> pure reconstructed
      _ -> pure reconstructed
```
Here the logic is very clear: if we split it into an application, then we process the parts of the application and put it together again. We don't do the subterm traversal an additional time.

thealmarty: After I paired with you, Ziyang wanted me to only split the context when it's a variable. And then there are still the bugs. This result is after I modified that to get rid of the errors.

thealmarty:
> Oh, not pure t, but pure (fillAppContext processedHd processedArgs).

@zliu41 as I mentioned in another comment, when we don't inline we don't want to put back together the processed parts of the application context, otherwise we get errors.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
zliu41: > only split the context when it's a variable

Doing so minimizes the changes to the `processTerm` logic, but I think Michael's way is better - it's shorter and I can't think of any issue it may cause. Can you try it? I'd be surprised if it causes any errors or regression.

thealmarty: yes, Michael paired with me and we tried that, and got the errors. Yes I know we got surprised by the inliner over and over again.

zliu41: > yes, Michael paired with me and we tried that, and got the errors.

This is surprising since it seems correct to me. Can you push a commit that does what Michael suggested? I just want to see what happens.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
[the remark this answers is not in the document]
thealmarty: In https://github.com/input-output-hk/plutus/pull/5426/commits/e71d7dd4c0a00f75f45fc2ce5f6b6df5c182f7cd I check the term size of `forMOf termSubterms t processTerm` because as I explained when we're not inlining we're returning `forMOf termSubterms t processTerm`. This actually make it so that the script context tests are not impacted by this change. So it's very curious how when comparing the sizes the other way actually caused some improvements for the other script context tests - this actually seems a rather strong indicator that the current size metric is not very good. (The other tests that got improvements are measuring the budget, while the `fitsInto` tests were using the same size metric).

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
thealmarty: https://github.com/input-output-hk/plutus/pull/5426/commits/caf0d825d2f959c9351f55df7acee1171338a948

thealmarty: I have since fixed a few things since that commit though. But yeah you can see the idea.

zliu41: > I have since fixed a few things

That might have fixed the problem. Just put caf0d82 on top of your branch pls - I just want to see how it does.

thealmarty: There will be merge conflict, I can't "just" put that on top. In fact, I tried that and it's just too messy. I put the fixes I did on top of [this branch](https://github.com/input-output-hk/plutus/tree/thealmarty/michaels-approach-pir-inliner). When I ran it locally I got 
```
GHC Core to PLC plugin: Error: Error from the PIR compiler:
                               Error from the PLC compiler:
                               Free type variable at  Ann {annInline = MayInline, annSrcSpans = { no-src-span }} :  a_10327
Context: Compiling expr at "plutus-benchmark-0.1.0.0-inplace-script-contexts-internal:PlutusBenchmark.ScriptContexts:(96,7)-(96,53)"
Error: cabal: Failed to build lib:script-contexts-internal from
plutus-benchmark-0.1.0.0 (which is required by
test:plutus-benchmark-script-contexts-tests from plutus-benchmark-0.1.0.0).
```
Let me know if you get the same result.

(I don't want to spend a lot of time tidying this branch up so I only put the semantic relevant fixes in)

zliu41 (approved):
I pushed Michael's approach to your branch: https://github.com/input-output-hk/plutus/pull/5426/commits/c9151a541e92c0d87b3928c61b57c61f469f5a55
There you go. Fix the remaining errors and it will be ready to merge.

> There will be merge conflict, I can't "just" put that on top. In fact, I tried that and it's just too messy.

:question: All you needed to do is throwing out the part of `processTerm` you wrote, and replacing with [this](https://github.com/input-output-hk/plutus/pull/5426#discussion_r1315035929). Took me less than five minutes.

> When I ran it locally I got
> 
> ```
> GHC Core to PLC plugin: Error: Error from the PIR compiler:
>                                Error from the PLC compiler:
>                                Free type variable at  Ann {annInline = MayInline, annSrcSpans = { no-src-span }} :  a_10327
> Context: Compiling expr at "plutus-benchmark-0.1.0.0-inplace-script-contexts-internal:PlutusBenchmark.ScriptContexts:(96,7)-(96,53)"
> Error: cabal: Failed to build lib:script-contexts-internal from
> plutus-benchmark-0.1.0.0 (which is required by
> test:plutus-benchmark-script-contexts-tests from plutus-benchmark-0.1.0.0).
> ```
> 
> Let me know if you get the same result.

I did. It's caused by failing to substitute the type arguments. The free variable is a free _type_ variable, which is the clue.

Wrong:
```haskell
processArgs (TypeAppContext ty ann ctx) = do
    processedArgs <- processArgs ctx
    pure $ TypeAppContext ty ann processedArgs
```

Correct:
```haskell
processArgs (TypeAppContext ty ann ctx) = do
    processedArgs <- processArgs ctx
    ty' <- applyTypeSubstitution ty
    pure $ TypeAppContext ty' ann processedArgs
```

zliu41:
And that's why errors like the free variable error should be investigated, not just ignored because it somehow mysteriously went away.

PR #5432, opened 2023-07-20
1.5.0.1

zliu41:


michaelpj:
This is all wrong, apparently I did a 1.5.0.1 release ages ago off the 1.5.0.0 branch, so we need 1.5.0.2.

michaelpj:
Ugh ignore me, that's `plutus-core`. Nonetheless, the 1.5.0 branch is now missing the changes from 1.5.0.0, I think we should just use the other branch.

michaelpj:
Okay, I instead merged 1.5.0.0 into this branch.

michaelpj:
Yes, I think I thought it was fine because it was a patch release but I think that was in fact a mistake for the reason you say.

Issue #4781, opened 2022-07-21
Why DefaultUni is so complicated?


phadej:
Why cannot pair and list be represented as

```
DefaultUniList :: DefaultUni a -> DefaultUni [a]
DefaultUniPair :: DefaultUni a -> DefaultUni b -> DefaultUni (a, b)
```

Is `DefaultUni (Esc a)` used anywhere (where?) at any other kind than `Type`? In `ValueOf` it is at type `Type`, and above simplification would be massive.

EDIT: The problem is that I have no idea how to exhaustively match on `DefaultUni`. Indeed the `plutus-core` itself has
(somewhat ugly):

```haskell
    bring _ (f `DefaultUniApply` _ `DefaultUniApply` _ `DefaultUniApply` _) _ =
        noMoreTypeFunctions f
```

michaelpj:
cc @effectfully

effectfully:
> Why cannot pair and list be represented as

See [this](https://github.com/input-output-hk/plutus/blob/4b40b31893a8a31456946670af981b06d71be6d5/plutus-core/plutus-core/src/Universe/Core.hs#L95) Note for how we arrived at this representation.

> Is `DefaultUni (Esc a)` used anywhere (where?) at any other kind than `Type`? 

Yes, in the Plutus types. The Note has an abstract example.

> EDIT: The problem is that I have no idea how to exhaustively match on DefaultUni. Indeed the plutus-core itself has
(somewhat ugly):

Upvote [this](https://github.com/ghc-proposals/ghc-proposals/pull/400) proposal :)

phadej:
Ok.

I still think the design is too clever for too little (I don't really believe in any non `DefaultUni` instantiations), but this is not my call.

effectfully:
> I don't really believe in any non `DefaultUni` instantiations

We were told that Plutus was going to be used on other blockchains. But even if that never happens, we still need `DefaultUni` as a separate entity. Consider a function like `sum :: [Integer] -> Integer`: in order to apply it, you need a way of extracting a `[Integer]` from the AST. You could have

```haskell
data Constant
    = ConstantInteger Integer
    | ConstantBool Bool
    | ConstantList [Constant]
```

in the AST but then extracting a `[Integer]` from that is an O(n) operation in the length of the list, plus you practically allow heterogeneous lists, which is inappropriate for a reasonably typed language. You could do

```haskell
data Constant
    = ConstantInteger Integer
    | ConstantBool Bool
[...]

PR #5434, opened 2023-07-21
[Builtins] Make elaboration customizable and unhardcode 'mkTyBuiltin'

effectfully:
This separates the single `Includes` constraint into two constraints, `HasTypeLevel` and `HasTermLevel` (which together form `HasBothLevel`). This is a prerequisite for splitting the universe of built-in types into two universes, one for the type level and the other for the term level, which is going to bring plenty of benefits:

1. make the implementation complaint with the metatheory
2. allow for simpler and more efficient implementation of `DefaultUni` as per #4781
3. allow for supporting partially applied types as built-in types without `newtype`-wrapping them (which is what @kwxm had to do with `BlstBindings.Point1`, `BlstBindings.Point2` and `BlstBindings.PT`)
4. simplify turning `Some (ValueOf DefaultUni)` into a `data family`
5. allow for erasable built-in types (a.k.a built-in newtypes, we'll talk about those a bit later)
6. make it clear if a definition uses a built-in type or values of the built-in type or both, thus providing a bit more type-level insight

This PR doesn't change any behavior, only type-level stuff. Behavioral changes are for future PRs.

### DIFF
--- a/plutus-core/changelog.d/20230724_195443_effectfully_make_elaboration_of_builtins_customizable.md
+++ b/plutus-core/changelog.d/20230724_195443_effectfully_make_elaboration_of_builtins_customizable.md
@@ -1,3 +1,3 @@
 ### Changed
 
-- Separated the single `Includes` constraint into two constraints, `HasTypeLevel` and `HasTermLevel` (which together form `HasBothLevel`) in #5434.
+- Separated the single `Includes` constraint into two constraints, `HasTypeLevel` and `HasTermLevel` (which together form `HasTypeAndTermLevel`) in #5434.
--- a/plutus-core/changelog.d/20230724_195443_effectfully_make_elaboration_of_builtins_customizable.md
+++ b/plutus-core/changelog.d/20230724_195443_effectfully_make_elaboration_of_builtins_customizable.md
@@ -1,0 +1,3 @@
+### Changed
+
+- Separated the single `Includes` constraint into two constraints, `HasTypeLevel` and `HasTermLevel` (which together form `HasBothLevel`) in #5434.
--- a/plutus-core/cost-model/budgeting-bench/Benchmarks/Unit.hs
+++ b/plutus-core/cost-model/budgeting-bench/Benchmarks/Unit.hs
@@ -14,7 +14,7 @@ import System.Random (StdGen)
 
 
 createChooseUnitBench
-    :: (DefaultUni `Includes` a, ExMemoryUsage a, NFData a)
+    :: (DefaultUni `HasTermLevel` a, ExMemoryUsage a, NFData a)
     => Type TyName DefaultUni ()
     -> [a]
     -> Benchmark
--- a/plutus-core/cost-model/budgeting-bench/Common.hs
+++ b/plutus-core/cost-model/budgeting-bench/Common.hs
@@ -92,21 +92,23 @@ benchDefault = benchWith defaultCekParameters
 
 ---------------- Constructing Polymorphic PLC terms for benchmarking ----------------
 
-integer :: uni `Includes` Integer => Type tyname uni ()
+integer :: uni `HasTypeLevel` Integer => Type TyName uni ()
 integer = mkTyBuiltin @_ @Integer ()
 
-bytestring :: uni `Includes` BS.ByteString => Type tyname uni ()
+bytestring :: uni `HasTypeLevel` BS.ByteString => Type TyName uni ()
 bytestring = mkTyBuiltin @_ @BS.ByteString ()
 
 
 -- To make monomorphic terms, make tys equal to [] in the mkApp functions
 
 -- Just make the term (con unit ()), which is about the simplest possible.
-mkUnit :: uni `Includes` () => PlainTerm uni fun
+mkUnit :: uni `HasTermLevel` () => PlainTerm uni fun
 mkUnit = eraseTerm $  mkConstant () ()
 
 -- Create a term instantiating a builtin and applying it to one argument
-mkApp1 :: (uni `Includes` a, NFData a) => fun -> [Type tyname uni ()] -> a -> PlainTerm uni fun
+mkApp1
+    :: (uni `HasTermLevel` a, NFData a)
+    => fun -> [Type tyname uni ()] -> a -> PlainTerm uni fun
 mkApp1 !name !tys (force -> !x) =
     eraseTerm $ mkIterAppNoAnn instantiated [mkConstant () x]
     where instantiated = mkIterInstNoAnn (builtin () name) tys
@@ -114,8 +116,8 @@ mkApp1 !name !tys (force -> !x) =
 
 -- Create a term instantiating a builtin and applying it to two arguments
 mkApp2
-    :: (uni `Includes` a, uni `Includes` b, NFData a, NFData b)
-    =>  fun -> [Type tyname uni ()]-> a -> b -> PlainTerm uni fun
+    :: (uni `HasTermLevel` a, uni `HasTermLevel` b, NFData a, NFData b)
+    => fun -> [Type tyname uni ()]-> a -> b -> PlainTerm uni fun
 mkApp2 !name !tys (force -> !x) (force -> !y) =
     eraseTerm $ mkIterAppNoAnn instantiated [mkConstant () x,  mkConstant () y]
     where instantiated = mkIterInstNoAnn (builtin () name) tys
@@ -123,7 +125,9 @@ mkApp2 !name !tys (force -> !x) (force -> !y) =
 
 -- Create a term instantiating a builtin and applying it to three arguments
 mkApp3
-    :: (uni `Includes` a, uni `Includes` b, uni `Includes` c, NFData a, NFData b, NFData c)
+    :: ( uni `HasTermLevel` a, uni `HasTermLevel` b, uni `HasTermLevel` c
+       , NFData a, NFData b, NFData c
+       )
     => fun -> [Type tyname uni ()] -> a -> b -> c -> PlainTerm uni fun
 mkApp3 !name !tys (force -> !x) (force -> !y) (force -> !z) =
     eraseTerm $ mkIterAppNoAnn instantiated [mkConstant () x, mkConstant () y, mkConstant () z]
@@ -132,9 +136,10 @@ mkApp3 !name !tys (force -> !x) (force -> !y) (force -> !z) =
 
 -- Create a term instantiating a builtin and applying it to four arguments
 mkApp4
-    :: (uni `Includes` a, uni `Includes` b,
-        uni `Includes` c, uni `Includes` d,
-        NFData a, NFData b, NFData c, NFData d)
+    :: ( uni `HasTermLevel` a, uni `HasTermLevel` b
+       , uni `HasTermLevel` c, uni `HasTermLevel` d
+       , NFData a, NFData b, NFData c, NFData d
+       )
     => fun -> [Type tyname uni ()] -> a -> b -> c -> d -> PlainTerm uni fun
 mkApp4 !name !tys (force -> !x) (force -> !y) (force -> !z) (force -> !t) =
     eraseTerm $ mkIterAppNoAnn instantiated [ mkConstant () x, mkConstant () y
@@ -144,9 +149,10 @@ mkApp4 !name !tys (force -> !x) (force -> !y) (force -> !z) (force -> !t) =
 
 -- Create a term instantiating a builtin and applying it to five arguments
 mkApp5
-    :: (uni `Includes` a, uni `Includes` b, uni `Includes` c,
-        uni `Includes` d, uni `Includes` e,
-        NFData a, NFData b, NFData c, NFData d, NFData e)
+    :: ( uni `HasTermLevel` a, uni `HasTermLevel` b, uni `HasTermLevel` c
+       , uni `HasTermLevel` d, uni `HasTermLevel` e
+       , NFData a, NFData b, NFData c, NFData d, NFData e
+       )
     => fun -> [Type tyname uni ()] -> a -> b -> c -> d -> e -> PlainTerm uni fun
 mkApp5 !name !tys (force -> !x) (force -> !y) (force -> !z) (force -> !t) (force -> !u) =
     eraseTerm $ mkIterAppNoAnn instantiated [ mkConstant () x, mkConstant () y, mkConstant () z
@@ -156,9 +162,10 @@ mkApp5 !name !tys (force -> !x) (force -> !y) (force -> !z) (force -> !t) (force
 
 -- Create a term instantiating a builtin and applying it to six arguments
 mkApp6
-    :: (uni `Includes` a, uni `Includes` b, uni `Includes` c,
-        uni `Includes` d, uni `Includes` e, uni `Includes` f,
-        NFData a, NFData b, NFData c, NFData d, NFData e, NFData f)
+    :: ( uni `HasTermLevel` a, uni `HasTermLevel` b, uni `HasTermLevel` c
+       , uni `HasTermLevel` d, uni `HasTermLevel` e, uni `HasTermLevel` f
+       , NFData a, NFData b, NFData c, NFData d, NFData e, NFData f
+       )
     => fun -> [Type tyname uni ()] -> a -> b -> c -> d -> e -> f-> PlainTerm uni fun
 mkApp6 name tys (force -> !x) (force -> !y) (force -> !z) (force -> !t) (force -> !u) (force -> !v)=
     eraseTerm $ mkIterAppNoAnn instantiated [mkConstant () x, mkConstant () y, mkConstant () z,
@@ -179,7 +186,7 @@ mkApp6 name tys (force -> !x) (force -> !y) (force -> !z) (force -> !t) (force -
 {- | Given a builtin function f of type a -> _ together with a lists xs, create a
    collection of benchmarks which run f on all elements of xs. -}
 createOneTermBuiltinBench
-    :: (fun ~ DefaultFun, uni ~ DefaultUni, uni `Includes` a, ExMemoryUsage a, NFData a)
+    :: (fun ~ DefaultFun, uni ~ DefaultUni, uni `HasTermLevel` a, ExMemoryUsage a, NFData a)
     => fun
     -> [Type tyname uni ()]
     -> [a]
@@ -192,8 +199,11 @@ createOneTermBuiltinBench name tys xs =
    ys::[b], create a collection of benchmarks which run f on all pairs in
    {(x,y}: x in xs, y in ys}. -}
 createTwoTermBuiltinBench
-    :: (fun ~ DefaultFun, uni ~ DefaultUni, uni `Includes` a, DefaultUni `Includes` b,
-        ExMemoryUsage a, ExMemoryUsage b, NFData a, NFData b)
+    :: ( fun ~ DefaultFun, uni ~ DefaultUni
+       , uni `HasTermLevel` a, DefaultUni `HasTermLevel` b
+       , ExMemoryUsage a, ExMemoryUsage b
+       , NFData a, NFData b
+       )
     => fun
     -> [Type tyname uni ()]
     -> [a]
@@ -214,8 +224,11 @@ createTwoTermBuiltinBench name tys xs ys =
    builtin can spot that its arguments both point to the same heap object.
 -}
 createTwoTermBuiltinBenchElementwise
-    :: (fun ~ DefaultFun, uni ~ DefaultUni, uni `Includes` a, uni `Includes` b,
-            ExMemoryUsage a, ExMemoryUsage b, NFData a, NFData b)
+    :: ( fun ~ DefaultFun, uni ~ DefaultUni
+       , uni `HasTermLevel` a, uni `HasTermLevel` b
+       , ExMemoryUsage a, ExMemoryUsage b
+       , NFData a, NFData b
+       )
     => fun
     -> [Type tyname uni ()]
     -> [a]
@@ -232,8 +245,11 @@ createTwoTermBuiltinBenchElementwise name tys xs ys =
    inputs.
 -}
 createThreeTermBuiltinBenchElementwise
-    :: (fun ~ DefaultFun, uni ~ DefaultUni, uni `Includes` a, uni `Includes` b, uni `Includes` c,
-            ExMemoryUsage a, ExMemoryUsage b, ExMemoryUsage c, NFData a, NFData b, NFData c)
+    :: ( fun ~ DefaultFun, uni ~ DefaultUni
+       , uni `HasTermLevel` a, uni `HasTermLevel` b, uni `HasTermLevel` c
+       , ExMemoryUsage a, ExMemoryUsage b, ExMemoryUsage c
+       , NFData a, NFData b, NFData c
+       )
     => fun
     -> [Type tyname uni ()]
     -> [(a,b,c)]
--- a/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
+++ b/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
@@ -169,24 +169,25 @@ instance (Default (BuiltinVersion fun1), Default (BuiltinVersion fun2))
 newtype MetaForall name a = MetaForall a
 instance
         ( name ~ 'TyNameRep @kind text uniq, KnownSymbol text, KnownNat uniq
-        , KnownKind kind, KnownTypeAst uni a
-        ) => KnownTypeAst uni (MetaForall name a) where
-    type IsBuiltin (MetaForall name a) = 'False
-    type ToHoles (MetaForall name a) = '[TypeHole a]
-    type ToBinds acc (MetaForall name a) = ToBinds (Insert ('GADT.Some name) acc) a
+        , KnownKind kind, KnownTypeAst tyname uni a
+        ) => KnownTypeAst tyname uni (MetaForall name a) where
+    type IsBuiltin _ (MetaForall name a) = 'False
+    type ToHoles _ (MetaForall name a) = '[TypeHole a]
+    type ToBinds uni acc (MetaForall name a) = ToBinds uni (Insert ('GADT.Some name) acc) a
     toTypeAst _ = toTypeAst $ Proxy @a
 instance MakeKnownIn DefaultUni term a => MakeKnownIn DefaultUni term (MetaForall name a) where
     makeKnown (MetaForall x) = makeKnown x
 -- 'ReadKnownIn' doesn't make sense for 'MetaForall'.
 
 data PlcListRep (a :: GHC.Type)
-instance KnownTypeAst uni a => KnownTypeAst uni (PlcListRep a) where
-    type IsBuiltin (PlcListRep a) = 'False
-    type ToHoles (PlcListRep a) = '[RepHole a]
-    type ToBinds acc (PlcListRep a) = ToBinds acc a
+instance (tyname ~ TyName, KnownTypeAst tyname uni a) =>
+        KnownTypeAst tyname uni (PlcListRep a) where
+    type IsBuiltin _ (PlcListRep a) = 'False
+    type ToHoles _ (PlcListRep a) = '[RepHole a]
+    type ToBinds uni acc (PlcListRep a) = ToBinds uni acc a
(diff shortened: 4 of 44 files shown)


plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
```diff
@@ -143,77 +150,89 @@ data family RepHole x
...
+type HasBothLevel uni x = (uni `HasTypeLevel` x, uni `HasTermLevel` x)
 
-type KnownTypeAst :: forall a. (GHC.Type -> GHC.Type) -> a -> GHC.Constraint
-class KnownTypeAst uni x where
+-- TODO: make it @forall {a}@ once we have that.
+-- (see https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0099-explicit-specificity.rst)
+-- | Convert a Haskell representation of a possibly 0-ary application of a built-in type to
+-- arbitrary types implementing 'KnownTypeAst'.
+mkTyBuiltin :: forall a (x :: a) uni ann tyname. uni `HasTypeLevel` x => ann -> Type tyname uni ann
+mkTyBuiltin ann = ann <$ typeMapNames absurd (toTypeAst $ Proxy @(ElaborateBuiltin uni x))
```
effectfully, line 170: Note to self: explain the trick.

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
```diff
@@ -143,77 +150,89 @@ data family RepHole x
...
+-- | This class allows one to convert the type-level Haskell representation of a Plutus type into
+-- the corresponding Plutus type. Associated type families are needed to help elaboration.
+--
+-- Depending on the universe converting a Haskell type to a Plutus team can give different results
+-- (e.g. 'Int' can be a built-in type instead of being encoded via built-in 'Integer'), hence this
+-- class takes a @uni@ argument. Plus, elaboration is universe-specific too.
+--
+-- 'KnownTypeAst' class also takes a @tyname@ argument, so that we can define 'mkTyBuiltin' that is
+-- generic over the type of names (in addition to 'TyName' we also have 'TyDeBruijn', so genericity
+-- does help).
```
effectfully, line 186: This elaborates on pretty much all the changes in the PR. I generalized over `tyname` because of [this](https://github.com/input-output-hk/plutus/pull/5390#discussion_r1244582581) discussion -- wasn't hard in the end at all, just a bit of straightforward boilerplate.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -102,18 +103,18 @@ feature and have meta-constructors as built-in functions.
...
-    DefaultUniString               :: DefaultUni (Esc Text.Text)
-    DefaultUniUnit                 :: DefaultUni (Esc ())
-    DefaultUniBool                 :: DefaultUni (Esc Bool)
-    DefaultUniProtoList            :: DefaultUni (Esc [])
-    DefaultUniProtoPair            :: DefaultUni (Esc (,))
-    DefaultUniApply                :: !(DefaultUni (Esc f)) -> !(DefaultUni (Esc a)) -> DefaultUni (Esc (f a))
-    DefaultUniData                 :: DefaultUni (Esc Data)
+    DefaultUniInteger :: DefaultUni (Esc Integer)
+    DefaultUniByteString :: DefaultUni (Esc ByteString)
+    DefaultUniString :: DefaultUni (Esc Text)
```
effectfully, line 108: Got tired of maintaining alignment here.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -233,96 +234,108 @@ instance Pretty (DefaultUni a) where
...
-instance KnownBuiltinTypeAst DefaultUni ()                         => KnownTypeAst DefaultUni ()
-instance KnownBuiltinTypeAst DefaultUni Bool                       => KnownTypeAst DefaultUni Bool
-instance KnownBuiltinTypeAst DefaultUni [a]                        => KnownTypeAst DefaultUni [a]
-instance KnownBuiltinTypeAst DefaultUni (a, b)                     => KnownTypeAst DefaultUni (a, b)
-instance KnownBuiltinTypeAst DefaultUni Data                       => KnownTypeAst DefaultUni Data
-instance KnownBuiltinTypeAst DefaultUni BLS12_381.G1.Element       => KnownTypeAst DefaultUni BLS12_381.G1.Element
-instance KnownBuiltinTypeAst DefaultUni BLS12_381.G2.Element       => KnownTypeAst DefaultUni BLS12_381.G2.Element
-instance KnownBuiltinTypeAst DefaultUni BLS12_381.Pairing.MlResult => KnownTypeAst DefaultUni BLS12_381.Pairing.MlResult
-
-{- Note [Constraints of ReadKnownIn and MakeKnownIn instances]
```
effectfully, line 263: The Note is gone, because now the instances are all defined the same straightforward way regardless of whether the built-in type is monomorphic or polymorphic.

plutus-core/cost-model/budgeting-bench/Benchmarks/Unit.hs
```diff
@@ -14,7 +14,7 @@ import System.Random (StdGen)
 
 
 createChooseUnitBench
-    :: (DefaultUni `Includes` a, ExMemoryUsage a, NFData a)
+    :: (DefaultUni `HasTermLevel` a, ExMemoryUsage a, NFData a)
```
michaelpj, line 17: Bikeshedding: I think "universe 'has term' x" and "universe 'has type' x" are clear enough, not sure the "level" adds anything.

`HasBoth` is confusing, since "both" only makes sense in a context where you know the two things.

Suggestion:
- `HasTerms` -> ```uni `HasTerms` Data``` (`HasTermsOf`?)
- `HasType` -> ```uni `HasType` Data```
- `Includes` (not sure about this, but something that sounds like "fully has this type")

plutus-core/plutus-core/src/PlutusCore/Builtin/HasConstant.hs
```diff
@@ -38,7 +38,7 @@ fromValueOf uni = fromConstant . someValueOf uni
 {-# INLINE fromValueOf #-}
 
 -- | Wrap a Haskell value (provided its type is in the universe) as a @term@.
-fromValue :: (HasConstant term, UniOf term `Includes` a) => a -> term
+fromValue :: (HasConstant term, UniOf term `Contains` a) => a -> term
```
michaelpj, line 41: I already forgot we have both `Includes` and `Contains` :scream:

plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
```diff
@@ -209,7 +209,7 @@ class KnownMonotype val args res where
 
 -- | Once we've run out of term-level arguments, we return a
 -- 'TypeSchemeResult'/'RuntimeSchemeResult'.
-instance (Typeable res, KnownTypeAst (UniOf val) res, MakeKnown val res) =>
+instance (Typeable res, KnownTypeAst TyName (UniOf val) res, MakeKnown val res) =>
```
michaelpj, line 212: why do we need to pin it down here?

plutus-core/plutus-core/src/PlutusCore/Builtin/TypeScheme.hs
```diff
@@ -59,10 +59,10 @@ on readability of the Core output.
 -- E.g. @Text -> Bool -> Integer@ is encoded as @TypeScheme val [Text, Bool] Integer@.
 data TypeScheme val (args :: [GHC.Type]) res where
     TypeSchemeResult
-        :: (Typeable res, KnownTypeAst (UniOf val) res, MakeKnown val res)
+        :: (Typeable res, KnownTypeAst TyName (UniOf val) res, MakeKnown val res)
```
michaelpj, line 62: ditto?

plutus-core/testlib/PlutusCore/Generators/Hedgehog/Denotation.hs
```diff
@@ -28,7 +28,7 @@ import Data.Dependent.Map qualified as DMap
 import Data.Functor.Compose
 import Type.Reflection
 
-type KnownType val a = (KnownTypeAst (UniOf val) a, MakeKnown val a)
+type KnownType val a = (KnownTypeAst TyName (UniOf val) a, MakeKnown val a)
```
michaelpj, line 31: ditto

michaelpj: were we just implicitly pinned to `TyName` somehow before, and this just makes it explicit? or what?

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
```diff
@@ -143,77 +150,89 @@ data family RepHole x
...
--- 'ToBinds', so handling built-in types in a special way for each of those would be a hassle,
--- especially given the fact that type-level Haskell is not exactly good at computing things.
--- With the 'ElaborateBuiltin' approach we get 'KnownTypeAst', 'ToHoles' and 'ToBinds' for free.
-type ElaborateBuiltin :: forall k. k -> k
-type family ElaborateBuiltin a where
-    ElaborateBuiltin (f x) = ElaborateBuiltin f `TyAppRep` x
-    ElaborateBuiltin f     = BuiltinHead f
+-- | Specifies that the given type is a built-in one and can be embedded into a 'Type'.
+type HasTypeLevel :: forall a. (GHC.Type -> GHC.Type) -> a -> GHC.Constraint
+type HasTypeLevel uni x = KnownTypeAst Void uni (ElaborateBuiltin uni x)
```
michaelpj, line 155: I guess this could instead be quantified over the type name argument?

michaelpj (approved):
Not fully following, but seems okay. Main thoughts are
1. Can we name these more clearly. `Contains` and `Includes` are also confusing and I'm surprised that `Includes` is still around!
2. Some places we fix the new type parameter to `TyName` and I'm unsure if this is significant.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
kwxm: `M-x align-regexp` does the job, but I don't know if it's worth it.  Now that we have these very long type names you can end up with a lot of whitespace in the middle of a line if you align stuff.

plutus-core/testlib/PlutusCore/Generators/Hedgehog/Denotation.hs
effectfully: > were we just implicitly pinned to `TyName` somehow before, and this just makes it explicit?

Yes, exactly. It used to be

```
class KnownTypeAst uni x where
    <...>
    toTypeAst :: proxy x -> Type TyName uni ()
    <...>
```

Now we're generic over `tyname`, but only for the very specific purpose of allowing `mkTyBulitin` to construct a `Type` with any `tyname` and so I get out of that generality as quickly as possible in order not to proliferate `tyname` through the entire code base. Like, `HasTypeLevel` propagates up to `PlutusTx`, we don't want to mention implementation details such as `TyName` there.

plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
effectfully: We want it here, because we want it in `TypeScheme`.

We want it in `TypeScheme`, because we hardcode `TyName` in 

```haskell
typeSchemeToType :: TypeScheme val args res -> Type TyName (UniOf val) ()
```

anyway, so there isn't much point in abstracting something to only be able to instantiate it to a single thing in the end. 

And the reason why it has to be a single thing is that otherwise we'd need to generalize

```
typeSchemeToType (TypeSchemeAll proxy schK) = case proxy of
    (_ :: Proxy '(text, uniq, kind)) ->
        let text = Text.pack $ symbolVal @text Proxy
            uniq = fromIntegral $ natVal @uniq Proxy
            a    = TyName $ Name text $ Unique uniq
        in TyForall () a (demoteKind $ knownKind @kind) $ typeSchemeToType schK
```

to be able to construct other `tyname`s and that would pretty much require redoing all name handling in `TypeScheme.hs`/`Meaning.hs`/`Elaborate.hs`, because that `uniq` in there assumes that we're going to construct a `TyName` in the end. Which would be a ton of work for nothing in return, because we can always just convert names.

As per my other comment, we really only benefit from generality in a single place (to the best of my knowledge), so why bother.

plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
```diff
@@ -169,24 +169,25 @@ instance (Default (BuiltinVersion fun1), Default (BuiltinVersion fun2))
...
-    type ToBinds acc (PlcListRep a) = ToBinds acc a
+instance (tyname ~ TyName, KnownTypeAst tyname uni a) =>
+        KnownTypeAst tyname uni (PlcListRep a) where
+    type IsBuiltin _ (PlcListRep a) = 'False
+    type ToHoles _ (PlcListRep a) = '[RepHole a]
+    type ToBinds uni acc (PlcListRep a) = ToBinds uni acc a
     toTypeAst _ = TyApp () Plc.listTy . toTypeAst $ Proxy @a
 
-instance KnownTypeAst DefaultUni Void where
+instance tyname ~ TyName => KnownTypeAst tyname DefaultUni Void where
```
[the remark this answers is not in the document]
effectfully: Because it's determined by the other two, none of which can be determined from one another. I.e. in order to be able to implement a `KnownTypeAst` instance you need to pin down a universe and a type -- and only those two. The type of names follows the encoding then. In this case we encode `Void` as `all a. a`, so we need names (because of `all a`), so we use `TyName`.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: > `M-x align-regexp` does the job

Oh wow, thanks for telling me about that! I wish I knew about it earlier...

> Now that we have these very long type names you can end up with a lot of whitespace in the middle of a line if you align stuff.

Yeah, and `DefaultUniApply` no longer fits on a single line if we align stuff.

plutus-core/cost-model/budgeting-bench/Benchmarks/Unit.hs
effectfully: > `HasBoth` is confusing, since "both" only makes sense in a context where you know the two things.

Yes, but `HasBothLevel` isn't confusing, because we know what two levels we have.

> `HasTerms`

Meh, it's not terms, it's constants and we already use `HasConstant` for "has a `Constant`-like constructor, so adding a completely different `HasConstants` would be confusing. While ``uni `HasTermLevel` Integer`` does read (albeit a bit weirdly) like you can use an `Integer` at the term level when dealing with a term having access to the `uni` universe.

> `HasType`

This is way too abstract of a name to be stolen by the builtins machinery for very specific usage, I feel.

plutus-core/plutus-core/src/PlutusCore/Builtin/HasConstant.hs
effectfully: Yeah, we might want to discuss how to name those properly, I do agree they are confusing. But that shouldn't block this PR, `Contains` vs `Includes` is an old thing.

To add insult to injury `HasTermLevel` is defined as

```haskell
type HasTermLevel uni = Includes uni
```

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
effectfully: Yes, but this what I'm actively avoiding here, because see the [other](https://github.com/input-output-hk/plutus/pull/5434#discussion_r1272288449) comment.

kwxm: Is it true that ultimately the only reason that you ever need `HasTypeLevel` is to allow you to eventually call `mkTyBuiltin` and you only need `HasTermLevel` if you're eventually going to call `mkConstant`?  If that's the case then I hadn't grasped it before (maybe spittling the constraints has clarified things), so it might be worth mentioning it here.

plutus-core/plutus-core/src/PlutusCore/Builtin/Polymorphism.hs
```diff
@@ -173,6 +179,40 @@ data family TyAppRep (fun :: dom -> cod) (arg :: dom) :: cod
 -- | Representation of of an intrinsically-kinded universal quantifier: a bound name and a body.
 data family TyForallRep (name :: TyNameRep kind) (a :: GHC.Type) :: GHC.Type
 
+-- | For annotating an uninstantiated built-in type, so that it gets handled by the right instance
```
kwxm, line 182: Is this all here to deal with the thing that required the `newtype`, so it's kind of orthogonal to the rest of the stuff in this PR?  If I remember correctly we had a type that was actually some other type being instantiated at a phantom type and that was being processed by `DefaultUniApply`.  It might be worth mentioning that situation more explicitly here.

kwxm (approved):
As usual I'm not going to pretend that I understand everything that's going on here, but I think these changes have helped me to understand it a bit better than I did before.  Doing this as a separate step was very helpful: if it'd been mixed up with lots of other stuff I would have had no idea what was going on.

kwxm:
`grep` shows me several occurrences of `Contains` which I think can be replaced by `HasTermLevel`: [here](https://github.com/input-output-hk/plutus/blob/6134587a10cd6f2aed58a9dd6103db3b3c8e1ab7/plutus-core/cost-model/budgeting-bench/Benchmarks/Nops.hs#L315) (several times), [here](https://github.com/input-output-hk/plutus/blob/6134587a10cd6f2aed58a9dd6103db3b3c8e1ab7/plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs#L503) (and several other times in the same file), [here](https://github.com/input-output-hk/plutus/blob/6134587a10cd6f2aed58a9dd6103db3b3c8e1ab7/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs#L514), and maybe in this comment [here](https://github.com/input-output-hk/plutus/blob/6134587a10cd6f2aed58a9dd6103db3b3c8e1ab7/plutus-tx/src/PlutusTx/Show.hs#L111).  I think that if you change all of those the only place that the export of `Contains` from `PlutusCore` is used is [here](url).

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
```diff
@@ -83,7 +83,7 @@ test_Const =
         b <- forAll Gen.bool
         let tC = mkConstant () c
             tB = mkConstant () b
-            text = toTypeAst @_ @DefaultUni @Text Proxy
+            text = toTypeAst @_ @_ @DefaultUni @Text Proxy
```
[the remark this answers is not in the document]
effectfully: > Is there a specific reason why we have typed terms in tests in untyped-plutus-core?

I think in general it's largely accidental with the whole untyped-plutus-core depending on typed plutus-core situation, but I think in this specific module I just wanted to make sure that things evaluate correctly and wired in the type checker so that there are no surprises and to help with debugging.

plutus-core/changelog.d/20230724_195443_effectfully_make_elaboration_of_builtins_customizable.md
```diff
@@ -0,0 +1,3 @@
+### Changed
+
+- Separated the single `Includes` constraint into two constraints, `HasTypeLevel` and `HasTermLevel` (which together form `HasBothLevel`) in #5434.
```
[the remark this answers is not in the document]
effectfully: > Will we have something like `DefaultTypeUni` and `DefaultTermUni`

Yup. We might want to choose different names though. I'd really like to replace all `Default` business with `Cardano` for example.

> `(DefaultTypeUni HasType a, DefaultTermUni HasTerm a, ...)`

We still want to reference a single `uni` variable, because adding another type variable isn't going to appeal to anyone. We're probably going to associate a data family with that `uni` containing the type-level counterparts of types whose values are available at the term level (one of the worst sentences I've ever written, sorry about that). This way people who only care about the term level (anyone targeting UPLC) will not need to know anything about the type-level universe.

You can see a [prototype](https://github.com/input-output-hk/plutus/compare/effectfully/builtins/treating-type-heads?expand=1#diff-4c05db6daf04fdffe5653ce0426d2707197302397c7382e97f1b1c745973be2a) of the idea, but it's a bit old.

plutus-core/cost-model/budgeting-bench/Common.hs
```diff
@@ -92,38 +92,42 @@ benchDefault = benchWith defaultCekParameters
...
-bytestring :: uni `Includes` BS.ByteString => Type tyname uni ()
+bytestring :: uni `HasTypeLevel` BS.ByteString => Type TyName uni ()
 bytestring = mkTyBuiltin @_ @BS.ByteString ()
 
 
 -- To make monomorphic terms, make tys equal to [] in the mkApp functions
 
 -- Just make the term (con unit ()), which is about the simplest possible.
-mkUnit :: uni `Includes` () => PlainTerm uni fun
+mkUnit :: uni `HasTermLevel` () => PlainTerm uni fun
```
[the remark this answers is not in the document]
effectfully: > I think if you write `HasBothLevel` when you only really need `HasTermLevel` (like here, for example) then there won't be an error/warning, right? 

Unfortunately, right. No idea on how to improve on this.

> We should probably try to be as generic as we can when we're writing these constraints though since that might make things a bit more understandable.

I agree.

plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
```diff
@@ -169,24 +169,25 @@ instance (Default (BuiltinVersion fun1), Default (BuiltinVersion fun2))
...
-        , KnownKind kind, KnownTypeAst uni a
-        ) => KnownTypeAst uni (MetaForall name a) where
-    type IsBuiltin (MetaForall name a) = 'False
-    type ToHoles (MetaForall name a) = '[TypeHole a]
-    type ToBinds acc (MetaForall name a) = ToBinds (Insert ('GADT.Some name) acc) a
+        , KnownKind kind, KnownTypeAst tyname uni a
+        ) => KnownTypeAst tyname uni (MetaForall name a) where
+    type IsBuiltin _ (MetaForall name a) = 'False
+    type ToHoles _ (MetaForall name a) = '[TypeHole a]
+    type ToBinds uni acc (MetaForall name a) = ToBinds uni (Insert ('GADT.Some name) acc) a
```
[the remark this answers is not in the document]
effectfully: It's all elaboration business, so yeah, it's a bit too fancy.

plutus-core/plutus-core/src/Universe/Core.hs
```diff
@@ -432,7 +432,7 @@ someValueOf :: forall a uni. uni (Esc a) -> a -> Some (ValueOf uni)
 someValueOf uni = Some . ValueOf uni
 
 -- | Wrap a value into @Some (ValueOf uni)@, provided its type is in the universe.
-someValue :: forall a uni. uni `Includes` a => a -> Some (ValueOf uni)
+someValue :: forall a uni. uni `Contains` a => a -> Some (ValueOf uni)
```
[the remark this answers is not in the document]
effectfully: Because in this case `Includes` expands to `Contains` (this is always the case when the kind of its second argument is `*`), so I thought it was worth removing the indirection.

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
```diff
@@ -143,77 +150,101 @@ data family RepHole x
...
 
--- | A constraint for \"@a@ is a 'KnownTypeAst' by means of being included in @uni@\".
-type KnownBuiltinTypeAst uni a = KnownTypeAst uni (ElaborateBuiltin a)
+-- | Specifies that the given type is a built-in one and can be embedded into a 'Type'.
+type HasTypeLevel :: forall a. (GHC.Type -> GHC.Type) -> a -> GHC.Constraint
+type HasTypeLevel uni x = KnownTypeAst Void uni (ElaborateBuiltin uni x)
+
+-- | Specifies that the given type is a built-in one and its values can be embedded into a 'Term'.
+type HasTermLevel :: forall a. (GHC.Type -> GHC.Type) -> a -> GHC.Constraint
+type HasTermLevel uni = Includes uni
```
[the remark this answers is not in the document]
effectfully: > Could we put this in there instead and not export `Includes` at all? 

Well, `Universe.Core` doesn't know anything about the type-level vs term-level distinction, which is why I didn't want to introduce `HasTermLevel` there. 

> Also, if `Includes` is essentially confined to `Universe.Core` it might be possible to give it a longer but more illuminating name.

Agreed, I've created PLT-6851 for this.

Keep in mind that I'm going to change `Universe.Core` a lot too. The whole `Esc`, `Kinded` etc business is going to be removed simplifying the library. Unfortunately, `Includes` is there to stay, but maybe we'll decide to introduce the type-level vs term-level distinction right into the library (it's our internal library anyway) and replace `Includes` with `HasTermLevel`, although I'm not fond of this, because I think `Includes` is more general than that.

plutus-core/plutus-core/src/PlutusCore/Builtin/Polymorphism.hs
effectfully: > If I remember correctly we had a type that was actually some other type being instantiated at a phantom type and that was being processed by `DefaultUniApply`.

Yes, there's a [solution](https://github.com/input-output-hk/plutus/commit/8c41df16e75dfa6da5ab059244175ebfa28333b8) to that problem now with the changes this PR brings, but it's not going to be merged, because separating the universes will allow us to have a much simpler solution. And then we'll be able to remove all the `newtype` nonsense, provided it's safe to do so (no custom instances conflicting with the underlying types etc).

effectfully: > Is this all here to deal with the thing that required the `newtype`, so it's kind of orthogonal to the rest of the stuff in this PR? 

Yes, we could've preserved the hardcoded `ElaborateBuiltin` (given that the unhardcoded one does 100% the same thing for `DefaultUni` anyway), but it was an easy change, so I shoved it in too. It was nonsensical to hardcode a certain way of elaborating built-in types for all universes to begin with. But that's the general story of the builtins machinery: things are stupidly hardcoded until I get some spare time to fix that and make the machinery more flexible (and sometimes simpler at the same time).

I guess we could've just hardcoded `DefaultUni` everywhere without implementing any extensible built-in types. I do think having extensible built-in _functions_ pays off very well, but extensible built-in types seem useless until someone else decides to use Plutus on their chain or something (which is how extensible built-in types exist IIRC with the idea of running Plutus on Midnight).

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
effectfully: > Is it true that the only reason that you ever need `HasTypeLevel` is to allow you to eventually call `mkTyBuiltin` and you only need `HasTermLevel` if you're eventually going to call `mkConstant`?

I think this is correct, yes. Not necessarily `mkConstant`, can be `someValue` for example, but that's what `mkConstant` is defined in terms of.

>  If that's the case then I hadn't grasped it before (maybe spittling the constraints has clarified things), so it might be worth mentioning it here.

I'm not sure about that... What if somebody decides they want to call `toTypeAst` directly given the `HasTypeLevel` constraint? They might even decide to call `toTypeAst` over the argument of the built-in type in the constraint rather than the whole built-in type for example. 

Let's return to this question once we split the universe and everything regarding the design of builtins is finalized.

effectfully:
I'm going to claim that I've addressed the comments. Let's see if my changes broke anything.

> `grep` shows me several occurrences of `Contains` which I think can be replaced by `HasTermLevel`

Done.

Renamed `HasBothLevel` to `HasTypeAndTermLevel` too.

effectfully:
@michaelpj I think Darwin is stuck?

effectfully:
Nevermind, it worked.

PR #5437, opened 2023-07-26
PLT-6822 Add budget tests for Marlowe scripts

thealmarty:
Added golden budget tests for the Marlowe scripts.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


michaelpj:
I advised Marty to print the code since there's a difference between the CI build results and the local ones.

@thealmarty you almost certainly want to print the PIR, the UPLC will be useless.

plutus-core/testlib/PlutusCore/Test.hs
```diff
@@ -340,6 +342,39 @@ goldenUEvalProfile name values =
   nestedGoldenVsDocM name ".ueval-profile" $
     pretty . view _2 <$> (rethrow $ runUPlcProfile values)
 
+-- Budget testing
+
+goldenUplcBudget ::
```
zliu41, line 347: I don't see the need to add anything new in this module. We already have `goldenBudget` and we can just use that.

plutus-benchmark/marlowe/test/Spec.hs
```diff
@@ -0,0 +1,50 @@
...
+  semanticsMBench <- either error id <$> semanticsBenchmarks
+
+  -- Read the role payout benchmark files.
+  rolePayoutMBench <- either error id <$> rolePayoutBenchmarks
+
+  let allTests :: TestTree
+      allTests =
+        testGroup "plutus-benchmark Marlowe tests"
+            [
+            --     runTestNestedIn ["marlowe", "test"] $ testNested "semantics" $
```
zliu41, line 43: Remove?

plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/BenchUtil.hs
```diff
@@ -54,11 +59,31 @@ import PlutusPrelude ((.*))
...
+  :: CompiledCode a
+  -- ^ semantics or role payout validator.
+  -> M.Benchmark
+  -- ^ `PlutusBenchmark.Marlowe.Types.Benchmark`, benchmarking type used by the executable,
+  -- it includes benchmarking results along with script info.
+  -> UPLC.Term NamedDeBruijn PLC.DefaultUni PLC.DefaultFun ()
+  -- A named DeBruijn term, for turning to `Benchmarkable`.
+benchmarkToUPLC validator M.Benchmark{..} =
+  let
+    unsafeFromRight (Right x) = x
```
zliu41, line 76: There's already a `unsafeFromEither`.

thealmarty:
yeah @zliu41 the latest commit is just for testing only. I tried to change the min amount of code to nail down the diff. @michaelpj I'm getting the UPLC directly from applying. I think it's useful to see if the CI sees the same UPLC programs anyway.

thealmarty:
> > I advised Marty to print the code since there's a difference between the CI build results and the local ones.
> 
> But: +655,609 −46 I doubt anyone is ever going to read it!

@zliu41 we just need to see if the CI said failed on the test suite. If the results are different at all, then somehow the UPLC programs are different. I can't imagine how but this is to check where the discrepancy starts.

plutus-core/testlib/PlutusCore/Test.hs
thealmarty: yes this is similar to `goldenBudget` but it's for `UPLC` not `PlutusTx`. It's less awkward to use because I can just input the `UPLC` programs. We have many similar `golden_` functions I don't think this is inconsistent with the code base.

plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -158,8 +158,8 @@ measureBudget compiledCode =
    in case programE of
         Left _ -> Nothing
         Right program ->
-          let (_, UPLC.TallyingSt _ budget) =
-                UPLC.runCekNoEmit PLC.defaultCekParameters UPLC.tallying $
+          let (_, UPLC.CountingSt budget) =
```
zliu41, line 161: What's the reason for this change?

zliu41 (changes requested):

plutus-core/testlib/PlutusCore/Test.hs
thealmarty: The goal is *not* to minimize the number of lines here. Doing unneeded conversions just to save 9 lines is just not worth it. On top of that it makes the code harder to read. Best to keep the code most direct. The code will be less bug prone too. Converting the arguments into `CompiledCode` obviously doesn't help here as it doesn't save any conversion.

zliu41: Even if we want to keep `goldenUplcBudget`, isn't this much better?

```haskell
goldenUplcBudget name term = goldenBudget name (DeserializedCode term Nothing mempty)
```

How is this harder to read or more bug prone?

effectfully: Since you discuss two options and can't agree which one is better, I'm going to do god's work and dump a third one on you: instead of defining `goldenBudget` in terms of `goldenUplcBudget`, do the opposite, because converting `CompiledCode` to `UPLC.Program` is something that `goldenBudget` already does anyway via `getPlcNoAnn`.

I don't see a reason to duplicate 10 lines of code here, particularly given that we have already had at least one case with a budget calculations not being consistent across different measurement functions. Also, `goldenBudget` is still `TallyingSt` while `goldenUplcBudget` is `CountingSt`?

thealmarty: OK that direction of conversion does make more sense to me. I'll do that. 

I think I updated `goldenBudget` to be `CountingSt` already. Anyway it'll be combined anyway.

zliu41: > do the opposite

That seems sensible.

thealmarty: Big win. I overall decrease the number of lines by 3 when adding `goldenUplcBudget`. Also the error messages are better, plus, there's no extra conversions at all. :)

plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -180,8 +158,7 @@ goldenPirReadable ::
   TestNested
 goldenPirReadable name value =
   nestedGoldenVsDoc name ".pir-readable"
-    . fromMaybe "PIR not found in CompiledCode"
-    . fmap (pretty . AsReadable . view progTerm)
+    . maybe "PIR not found in CompiledCode" (pretty . AsReadable . view progTerm)
```
thealmarty, line 161: linter suggestion

plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -140,28 +139,7 @@ renderExcess tData mData diff =
...
-            traverseOf UPLC.progTerm UPLC.unDeBruijnTerm $
-              getPlcNoAnn compiledCode
-   in case programE of
-        Left _ -> Nothing
-        Right program ->
-          let (_, UPLC.TallyingSt _ budget) =
-                UPLC.runCekNoEmit PLC.defaultCekParameters UPLC.tallying $
-                  program ^. UPLC.progTerm
-           in Just budget
+goldenBudget name compiledCode = goldenUplcBudget name (getPlcNoAnn compiledCode)
```
zliu41, line 142: Does this compile? `getPlcNoAnn` returns a `Program` but `goldenUplcBudget` expects a `Term`.

zliu41 (approved):

plutus-tx/testlib/PlutusTx/Test.hs
thealmarty: it did locally...

thealmarty: oops sorry that's the `goldenBudget`, that's why. My HLS is not working well again today. Fixing.

PR #5438, opened 2023-07-27
Cleanup in api tests. Add some tests for keccak,bls

bezirg:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-core/testlib/UntypedPlutusCore/Test/DeBruijn/Bad.hs
```diff
@@ -83,8 +83,7 @@ deepOutMix1_0 n = timesA n lamAbs1 $
 -- Both branches are evaluated *before* the predicate,
 -- so it is clear that this should fail in every case.
 iteStrict0 :: Term DeBruijn DefaultUni DefaultFun ()
-iteStrict0 = mkIterAppNoAnn
-             (Force () (Builtin () IfThenElse))
+iteStrict0 = Force () (Builtin () IfThenElse) @@
```
michaelpj, line 86: :o when did we get this?

plutus-ledger-api/test/Spec/CostModelParams.hs
```diff
@@ -27,23 +26,24 @@ tests =
     "costModelParams"
     [ testCase "length" $ do
             166 @=? length (enumerate @V1.ParamName)
-            166 @=? length v1CostModelParamNames
             175 @=? length (enumerate @V2.ParamName)
-            175 @=? length v2CostModelParamNames
             223 @=? length (enumerate @V3.ParamName)
+            -- FIXME: the following three tests are misleading; better to directly test that
```
michaelpj, line 31: what's misleading about them? they seem okay? Is it because there could be duplicates or something?

michaelpj (approved):
looks fine!

plutus-ledger-api/test/Spec/CostModelParams.hs
bezirg: _misleading_ is not the right word indeed, I just couldn't come up with a better word.

the _purpose_ of these tests are not clear:

- Is their purpose a regression safety check for the number of expected arguments? Then they are redundant from the tests above.
- Is their purpose to check that `showParamName` is injective? Then it *indirectly* checks that through the length. The actual length is not important, what is important is they are injective and you don't see that directly.

Do I make sense? :D

plutus-core/testlib/UntypedPlutusCore/Test/DeBruijn/Bad.hs
bezirg: When u were on vacation I secretly introduced it :see_no_evil:

bezirg:
@michaelpj can u force this? I cannot comprehend the darwin error tbh

PR #5443, opened 2023-07-30
Add some strictness tests for the plugin in Strictness.Spec

zliu41:
Added some tests to verify that when the `Strict` extension is on, the compiler does the right things wrt strictness (it currently doesn't).

### DIFF
--- a/plutus-tx-plugin/test/Strictness/Spec.hs
+++ b/plutus-tx-plugin/test/Strictness/Spec.hs
@@ -1,0 +1,109 @@
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE Strict              #-}
+{-# LANGUAGE TemplateHaskell     #-}
+{-# OPTIONS_GHC -fplugin PlutusTx.Plugin #-}
+
+module Strictness.Spec where
+
+import Test.Tasty.Extras
+
+import PlutusTx qualified as Tx
+import PlutusTx.Code
+import PlutusTx.Prelude qualified as PlutusTx
+import PlutusTx.Test
+import PlutusTx.TH (compile)
+
+tests :: TestNested
+tests =
+  testNested
+    "Strictness"
+    [ goldenEvalCekCatch "lambda-default" $ [lambdaDefault `unsafeApplyCode` bot]
+    , goldenPirReadable "lambda-default" lambdaDefault
+    , goldenUPlcReadable "lambda-default" lambdaDefault
+
+    -- FIXME: This should not crash, but it currently does.
+    , goldenEvalCekCatch "lambda-nonstrict" $ [lambdaNonStrict `unsafeApplyCode` bot]
+    , goldenPirReadable "lambda-nonstrict" lambdaNonStrict
+    , goldenUPlcReadable "lambda-nonstrict" lambdaNonStrict
+
+    , goldenEvalCekCatch "lambda-strict" $ [lambdaStrict `unsafeApplyCode` bot]
+    , goldenPirReadable "lambda-strict" lambdaStrict
+    , goldenUPlcReadable "lambda-strict" lambdaStrict
+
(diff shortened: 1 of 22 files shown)


plutus-tx-plugin/test/Strictness/let-strict.pir-readable.golden
```diff
@@ -0,0 +1,17 @@
+let
+  data Unit | Unit_match where
+    Unit : Unit
+  data Bool | Bool_match where
+    True : Bool
+    False : Bool
+in
+\(m : integer) ->
+  let
+    !n : all t. t = /\t -> error {integer -> t} m
```
michaelpj, line 10: Weird! I have no idea why we're compiling the binding this way.

plutus-tx-plugin/test/Strictness/Spec.hs
```diff
@@ -0,0 +1,109 @@
...
+        [||
+        \n -> \m -> n PlutusTx.+ m
+        ||]
+    )
+
+lambdaNonStrict :: CompiledCode (Integer -> Integer -> Integer)
+lambdaNonStrict =
+  $$( compile
+        [||
+        \(~n) -> \m -> n PlutusTx.+ m
```
michaelpj, line 63: Interesting. I think we probably just can't support this. The non-strictness is gone by the time we get to Core, so then it's an issue of the fact that our function application is strict by default.

Generally, I suspect ~ patterns will work much worse than ! patterns. I think we might want to end up saying "use `Strict` and don't expect ~ patterns to work".

michaelpj (approved):

plutus-tx-plugin/test/Strictness/Spec.hs
zliu41: > The non-strictness is gone by the time we get to Core

With the `Strict` extension, GHC Core should distinguish `n` against `~n`. GHC does perform some non-semantic-preserving optimizations, but they should be rare and can be turned off by `-fpedantic-bottoms`, so I'm optimistic about this.

effectfully: > Generally, I suspect ~ patterns will work much worse than ! patterns.

I thought `~` would work with `Strict` more or less as well as `!` without `Strict`? No idea, though.

plutus-tx-plugin/test/Strictness/Spec.hs
```diff
@@ -0,0 +1,109 @@
...
+    , goldenPirReadable "lambda-nonstrict" lambdaNonStrict
+    , goldenUPlcReadable "lambda-nonstrict" lambdaNonStrict
+
+    , goldenEvalCekCatch "lambda-strict" $ [lambdaStrict `unsafeApplyCode` bot]
+    , goldenPirReadable "lambda-strict" lambdaStrict
+    , goldenUPlcReadable "lambda-strict" lambdaStrict
+
+    -- FIXME: This should crash (because the `Strict` extension is on), but it currently doesn't.
+    , goldenEvalCekCatch "let-default" $ [letDefault `unsafeApplyCode` one]
+    , goldenPirReadable "let-default" letDefault
```
effectfully, line 38: Why not pretty-print the same program that you're testing? Would be easier to review and generally make sense of.

plutus-tx-plugin/test/Strictness/let-strict.pir-readable.golden
effectfully: Weird indeed, I'm glad we've found it out, because that bug can really cripple up our users.

effectfully (commented):
Sorry for the late review. Looks great to me modulo the unexpected behavior.

plutus-tx-plugin/test/Strictness/Spec.hs
zliu41: I could do that but printing the argument doesn't really add anything. One should only ever need to check what the function is compiled into.

plutus-tx-plugin/test/Strictness/Spec.hs
zliu41: > I thought ~ would work with Strict more or less as well as ! without Strict?

Yes, that's what I thought.

PR #5444, opened 2023-07-30
Improve error messages in ScriptContext benchmarks

zliu41:
When a test fails, rather than `"evaluation failed"`, it now prints the actual error message and Cek logs.


effectfully (approved):
LGTM

Issue #5448, opened 2023-07-31
TxInfo for PlutusV3


lehins:
### Describe the feature you'd like

Here is the spec for all of the Conway related changes in the Ledger that affect TxInfo:

```haskell

newtype ColdCommitteeCredential = ColdCommitteeCredential Credential
  deriving (Show, Eq)

newtype HotCommitteeCredential = HotCommitteeCredential Credential
  deriving (Show, Eq)

newtype DRepCredential = DRepCredential Credential
  deriving (Show, Eq)

data DRep
  = DRepCredential DRepCredential
  | DRepAlwaysAbstain
  | DRepAlwaysNoConfidence
  deriving (Show, Eq)

data Delegatee
  = DelegStake PubKeyHash
  | DelegVote DRep
  | DelegStakeVote PubKeyHash DRep
  deriving (Show, Eq)

-- | There is probably no need for Plutus Scripts to even know about all this data in Anchors,
-- but it is here for completeness. So we might wanna omit them, especially since urls are a
-- bit tricky, since we can't guarantee their validity, not even that they are utf8-encoded.
data Anchor = Anchor
  { anchorUrl :: ByteString -- Arbitrary Url
  , anchorDataHash :: Hash -- Blake2b_256 of some off-chain data
  }
  deriving (Eq, Show)

-- | Note taht it no longer is caled DCert, since certificates are no longer all about
-- delegation, they are also about governance.
--
-- Notable changes:
--
-- * Pool certificates are unchanged
-- * Addition of deposits and refunds. Those values are validated according to respectful
--   PParams. Certificates that where present before get their deposit/refund as
--   optional. This is done for graceful tarnsition, since next era after Conway will require
--   deposit/refund amount
-- * New DRep entity
-- * New Constitutional Committee entity
data TxCert
  = -- | Register staking credential with an optional deposit amount
    TxCertRegStaking Credential (Maybe Value)
  | -- | Un-Register staking credential with an optional refund amount
    TxCertUnRegStaking Credential (Maybe Value)
  | -- | Delegate staking credential to a Delegatee
    TxCertDelegStaking Credential Delegatee
  | -- | Register and delegate staking credential to a Delegatee in one certificate. Noter that
    -- deposit is mandatory.
    TxCertRegDeleg Credential Delegatee Value
    -- | Register a DRep with mandatory deposit value and an optional Anchor
  | TxCertRegDRep DRepCredential Value (Maybe Anchor)
[...]

PR #5449, opened 2023-08-01
PlutusV3 ScriptContext

zliu41:
Based on https://github.com/input-output-hk/plutus/issues/5448.


plutus-tx/src/PlutusTx/IsData/Class.hs
```diff
@@ -92,6 +92,18 @@ instance UnsafeFromData Builtins.BuiltinByteString where
     {-# INLINABLE unsafeFromBuiltinData #-}
     unsafeFromBuiltinData = BI.unsafeDataAsB
 
+instance ToData Builtins.BuiltinString where
+    {-# INLINABLE toBuiltinData #-}
+    toBuiltinData = mkB . BI.encodeUtf8
```
zliu41, line 97: Mandating UTF-8 is unsatisfactory, but we need these instances since `BultinString` is used in `Anchor`.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+    a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
+  _ == _ = Haskell.False
+
+{- | There is probably no need for Plutus Scripts to even know about all this data in Anchors,
+but it is here for completeness. So we might wanna omit them, especially since urls are a
+bit tricky, since we can't guarantee their validity, not even that they are utf8-encoded.
+-}
+data Anchor = Anchor
+  { anchorUrl      :: PlutusTx.BuiltinString -- Arbitrary Url
+  , anchorDataHash :: PlutusTx.BuiltinByteString -- Blake2b_256 of some off-chain data
```
zliu41, line 90: Should we newtype these? Seems unnecessary.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+* Addition of deposits and refunds. Those values are validated according to respectful
+  PParams. Certificates that where present before get their deposit/refund as
+  optional. This is done for graceful tarnsition, since next era after Conway will require
+  deposit/refund amount
+* New DRep entity
+* New Constitutional Committee entity
+-}
+data TxCert
+  = -- | Register staking credential with an optional deposit amount
+    TxCertRegStaking V2.StakingCredential (Haskell.Maybe V2.Value)
```
zliu41, line 120: @lehins @WhatisRT There's no `StakingHash`; I'm assuming you meant `StakingCredential`.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+      ]
+
+instance PlutusTx.Eq Constitution where
+  {-# INLINEABLE (==) #-}
+  Constitution a b == Constitution a' b' =
+    a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
+
+data ProtocolVersion = ProtocolVersion
+  { pvMajor :: Haskell.Integer
+  , pvMinor :: Haskell.Integer
```
zliu41, line 270: We can perhaps reuse the existing `ProtocolVersion` type, but that one is faster due to `Int`. I don't know whether it matters or not.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+
+instance PlutusTx.Eq Delegatee where
+  {-# INLINEABLE (==) #-}
+  DelegStake a == DelegStake a' = a PlutusTx.== a'
+  DelegVote a == DelegVote a' = a PlutusTx.== a'
+  DelegStakeVote a b == DelegStakeVote a' b' =
+    a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
+  _ == _ = Haskell.False
+
+{- | There is probably no need for Plutus Scripts to even know about all this data in Anchors,
```
michaelpj, line 84: This shouldn't be a haddock comment!

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+but it is here for completeness. So we might wanna omit them, especially since urls are a
+bit tricky, since we can't guarantee their validity, not even that they are utf8-encoded.
+-}
+data Anchor = Anchor
+  { anchorUrl      :: PlutusTx.BuiltinString -- Arbitrary Url
+  , anchorDataHash :: PlutusTx.BuiltinByteString -- Blake2b_256 of some off-chain data
+  }
+  deriving stock (Generic, Haskell.Show, Haskell.Eq)
+
+instance Pretty Anchor where
```
michaelpj, line 94: why not PrettyShow here?

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+  _ == _             = Haskell.False
+
+-- | Similar to TxOutRef, but for GovActions
+data GovernanceActionId = GovernanceActionId
+  { gaidTxId        :: V2.TxId
+  , gaidGovActionIx :: Haskell.Integer
+  }
+  deriving stock (Generic, Haskell.Show, Haskell.Eq)
+
+instance Pretty GovernanceActionId where
```
michaelpj, line 200: ditto. generally unclear when you're picking one or the other. applies later too

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+      , "committeeQuorum:" <+> pretty committeeQuorum
+      ]
+
+instance PlutusTx.Eq Committee where
+  {-# INLINEABLE (==) #-}
+  Committee a b == Committee a' b' =
+    a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
+
+data Constitution = Constitution
+  { constitutionHash   :: PlutusTx.BuiltinByteString -- Balke2b_256 of some off-chain data
```
michaelpj, line 251: could newtype

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
michaelpj: Let's keep them separate for now, they have different purposes.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+instance PlutusTx.Eq ProtocolVersion where
+  {-# INLINEABLE (==) #-}
+  ProtocolVersion a b == ProtocolVersion a' b' =
+    a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
+
+data GovernanceAction
+  = -- | Proposed updates to PParams. This is a very tricky one, since PParamsUpdate is pretty
+    -- big and changes from one era to another. However, it could be very important for the
+    -- governance scripts to be able to see the values in the PParamsUpdate, but I am not
+    -- sure how to go forward about this, so for now I leave it as empty.
```
michaelpj, line 290: This seems like a massive TODO...

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+    ParameterChange -- Should contain: (PParamsUpdate era)
+    --
+  | -- | proposal to update protocol version
+    HardForkInitiation ProtocolVersion
+  | TreasuryWithdrawals !(Map V2.Credential V2.Value)
```
michaelpj, line 295: weird bang

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+  | NoConfidence
+  | NewCommittee
+      -- | Old committee
+      [ColdCommitteeHash]
```
michaelpj, line 299: kind of weird that this isn't a `Committee` but whatever

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+      PlutusTx.&& c PlutusTx.== c'
+      PlutusTx.&& d PlutusTx.== d'
+
+data ScriptPurpose
+  = Minting V2.CurrencySymbol
+  | Spending V2.TxOutRef
+  | Rewarding V2.StakingCredential
+  | Certifying TxCert
+  | Voting Voter GovernanceActionId
+  | Proposing -- For now this will only be used by Constitution, thus might turned out to be
```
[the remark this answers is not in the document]
michaelpj: Let's make sure we have a list of unresolved issues here and we mark it as blocking for the final release

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+    a PlutusTx.== a'
+  Voting a b == Voting a' b' =
+    a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
+  Proposing == Proposing = Haskell.True
+  _ == _ = Haskell.False
+
+-- | TxInfo for PlutusV3
+data TxInfo = TxInfo
+  { txInfoInputs                :: [V2.TxInInfo]
+  -- ^ Unchanged
```
michaelpj, line 373: haddock should be informative about the current state, not the diffs!

plutus-tx/src/PlutusTx/IsData/Class.hs
michaelpj: Hmmm. This is a pretty important constraint for the ledger. But they must have some similar constraint, since that string must be represented in the tx format as some bytestring. So we should be able to do the same thing... Let's ask what they do.

plutus-benchmark/script-contexts/test/checkScriptContexEqualityData-20.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 60538597
-| mem: 200402})
\ No newline at end of file
+({cpu: 65159541
```
michaelpj, line 1: oof

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
WhatisRT: Yeah. Same is true for the `DRepHash`, it's a credential as well.

lehins: `StakingCredential` is different from `Credential`. `StakingCredential` also contains pointers and and can be null:
https://github.com/input-output-hk/plutus/blob/a30229b3a07f47472460e5fe8c6135442d5c56a7/plutus-ledger-api/src/PlutusLedgerApi/V1/Credential.hs#L26-L38

Where Credential can be either KeyHash or ScriptHash. This one needs to be Credential
```suggestion
    TxCertRegStaking V2.Credential (Haskell.Maybe V2.Value)
```

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+  deriving stock (Generic, Haskell.Show, Haskell.Eq)
+  deriving (Pretty) via (PrettyShow HotCommitteeHash)
+  deriving newtype
+    ( PlutusTx.Eq
+    , PlutusTx.ToData
+    , PlutusTx.FromData
+    , PlutusTx.UnsafeFromData
+    )
+
+newtype DRepHash = DRepHash V2.Credential
```
lehins, line 45: My bad, I think I should have named it credential, rather than Hash. I have no idea how, but I mistakenly thought that's how in Plutus calls Credentials :facepalm: 
```suggestion
newtype DRepCredential = DRepCredential V2.Credential
```

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+  deriving stock (Generic, Haskell.Show, Haskell.Eq)
+  deriving (Pretty) via (PrettyShow ColdCommitteeHash)
+  deriving newtype
+    ( PlutusTx.Eq
+    , PlutusTx.ToData
+    , PlutusTx.FromData
+    , PlutusTx.UnsafeFromData
+    )
+
+newtype HotCommitteeHash = HotCommitteeHash V2.Credential
```
lehins, line 35: Same here:
```suggestion
newtype HotCommitteeCredential = HotCommitteeCredential V2.Credential
```

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+import Prettyprinter.Extras
+
+import PlutusLedgerApi.V2 qualified as V2
+import PlutusTx qualified
+import PlutusTx.AssocMap hiding (filter, mapMaybe)
+import PlutusTx.Prelude qualified as PlutusTx
+
+import Prelude qualified as Haskell
+
+newtype ColdCommitteeHash = ColdCommitteeHash V2.Credential
```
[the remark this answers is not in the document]
lehins: Yeah, it is good practice :slightly_smiling_face: 

Before there was only really only Payment and Staking credentials, now we have 3 more kinds. It'll be good to distinguish between them. For consistency sake it might be good to add `PaymentCredential` too. Just a suggestion.


Also, suggestion for my naming mishap: Credential
```suggestion
newtype ColdCommitteeCredential = ColdCommitteeCredential V2.Credential
```

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+  = -- | Register staking credential with an optional deposit amount
+    TxCertRegStaking V2.StakingCredential (Haskell.Maybe V2.Value)
+  | -- | Un-Register staking credential with an optional refund amount
+    TxCertUnRegStaking V2.StakingCredential (Haskell.Maybe V2.Value)
+  | -- | Delegate staking credential to a Delegatee
+    TxCertDelegStaking V2.StakingCredential Delegatee
+  | -- | Register and delegate staking credential to a Delegatee in one certificate. Noter that
+    -- deposit is mandatory.
+    TxCertRegDeleg V2.StakingCredential Delegatee V2.Value
+  | -- | Register a DRep with mandatory deposit value and an optional Anchor
```
lehins, line 128: Same reason, we can't register on unregister a Ptr or a NullAddr :slightly_smiling_face: 
```suggestion
    TxCertUnRegStaking V2.Credential (Haskell.Maybe V2.Value)
  | -- | Delegate staking credential to a Delegatee
    TxCertDelegStaking V2.Credential Delegatee
  | -- | Register and delegate staking credential to a Delegatee in one certificate. Noter that
    -- deposit is mandatory.
    TxCertRegDeleg V2.Credential Delegatee V2.Value
  | -- | Register a DRep with mandatory deposit value and an optional Anchor
```

I think confusion comes from the naming. We in Ledger call this `StakeCredential`, while this thing that is called `StakingCredential` in Plutus we actually call `StakeReference`:


```haskell
data StakeReference c
  = StakeRefBase !(StakeCredential c)
  | StakeRefPtr !Ptr
  | StakeRefNull
```

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+  DelegStakeVote a b == DelegStakeVote a' b' =
+    a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
+  _ == _ = Haskell.False
+
+{- | There is probably no need for Plutus Scripts to even know about all this data in Anchors,
+but it is here for completeness. So we might wanna omit them, especially since urls are a
+bit tricky, since we can't guarantee their validity, not even that they are utf8-encoded.
+-}
+data Anchor = Anchor
+  { anchorUrl      :: PlutusTx.BuiltinString -- Arbitrary Url
```
lehins, line 89: This is just a `ByteString` on the ledger side, so maybe using String here is not a good idea. I chose String in the ticket just to emphasize that it is textual data, but we do nothing to validate it.

plutus-tx/src/PlutusTx/IsData/Class.hs
lehins: We do not enforce utf8 encoding. ByteStrings are just ByteString for us.

So, the only time we should bother with this, if there will be desire to read this as text in Plutus. Then we can try figure out if we should do some sort of validation on it on the ledger side (which we could, but I know many people would be against it)

lehins (commented):
Here are some comments from me.

Looks good otherwise.

plutus-tx/src/PlutusTx/IsData/Class.hs
zliu41: Since we don't need to use String in Anchor, this is no longer a concern!

lehins (commented):
Couple more recommendations about StakingCredential

plutus-benchmark/script-contexts/test/checkScriptContexEqualityData-20.budget.golden
zliu41: This is V3 against V1, not against V2, so slightly less bad!

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
zliu41: Removed.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
zliu41: Elsewhere we usually define it manually for product types, and derive it for sum types.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+    vsep
+      [ "anchorUrl:" <+> pretty anchorUrl
+      , "anchorDataHash:" <+> pretty anchorDataHash
+      ]
+
+instance PlutusTx.Eq Anchor where
+  {-# INLINEABLE (==) #-}
+  Anchor a b == Anchor a' b' = a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
+
+{- | Note that it no longer is called DCert, since certificates are no longer all about
```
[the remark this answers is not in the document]
zliu41: This one seems pretty useful to me, as a user who used write V2 and want to migrate to V3?

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,577 @@
...
+  TxCertAuthHotCommittee a b == TxCertAuthHotCommittee a' b' =
+    a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
+  TxCertResignColdCommittee a == TxCertResignColdCommittee a' =
+    a PlutusTx.== a'
+  _ == _ = Haskell.False
+
+data Voter
+  = CommitteeVoter !HotCommitteeHash
+  | DRepVoter DRepHash
+  | StakePoolVoter V2.PubKeyHash
```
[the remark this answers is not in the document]
zliu41: We do use `PubKeyHash` directly in many places in the V1 and V2 types, so seems ok to me.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
zliu41: Yeah, we should just use `Strict` across the board. I'll remove the bang.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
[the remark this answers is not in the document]
zliu41: How about append `Experimental` to the module names? `PlutusLedgerApi.V3.Contexts.Experimental` and `PlutusLedgerApi.V3.Experimental`?

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
zliu41: I don't know. Like I said above, I think these are useful for someone migrating from V2 to V3, which is a lot of people.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
zliu41: Newtyped `anchorDataHash`.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
zliu41: Done.

zliu41 (commented):
Updated - please take another look!

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,625 @@
...
+    ( Haskell.Eq
+    , Haskell.Show
+    , PlutusTx.Eq
+    , PlutusTx.ToData
+    , PlutusTx.FromData
+    , PlutusTx.UnsafeFromData
+    )
+
+data DRep
+  = DRep DRepCredential
```
zliu41, line 64: Since there's already a `DRepCredential`, I renamed this to `DRep`. @lehins @WhatisRT Let me know if there's a better name!

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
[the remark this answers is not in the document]
michaelpj: I think the haddock for a type should say what it _is_. I think this belongs in the changelog!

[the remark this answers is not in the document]
michaelpj: Or it could be a section of the haddock saying "changes from V2" or something

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
michaelpj: Well, `StrictData` at least

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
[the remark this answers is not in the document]
michaelpj: I'm not sure how that helps?

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
zliu41: @lehins @WhatisRT Is it possible at all to make this contain only the delta, not the full params? Michael and I talked briefly about this and we suspect that including the full params here could be a serious problem due to the lack of efficient lookup operations in Plutus.

zliu41 (commented):
Removed some Haddock.

lehins:
> Is it possible at all to make this contain only the delta, not the full params? Michael and I talked briefly about this and we suspect that including the full params here could be a serious problem due to the lack of efficient lookup operations in Plutus.

We'll need to have a talk on how to best implement this. There are a few potential solutions, all with their own upsides and downsides, it will be a matter of deciding on what is the easiest for Plutus users to use and Plutus engineers to work with. I intentionally left it empty, because I can't make that decision myself, but I do have ideas on how it can be done.

`ParameterChange` will contain only the fields that are being updated, not the full set of PParams. That being said, nothing prevents a user of creating a proposal that changes all of the PParams, so it can very well be all of the PParams.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,581 @@
...
+  , txInfoFee                   :: V2.Value
+  , txInfoMint                  :: V2.Value
+  , txInfoTxCerts               :: [TxCert]
+  , txInfoWdrl                  :: Map V2.Credential Haskell.Integer
+  , txInfoValidRange            :: V2.POSIXTimeRange
+  , txInfoSignatories           :: [V2.PubKeyHash]
+  , txInfoRedeemers             :: Map ScriptPurpose V2.Redeemer
+  , txInfoData                  :: Map V2.DatumHash V2.Datum
+  , txInfoId                    :: V2.TxId
+  , txInfoVotingProcedures      :: Map Voter (Map GovernanceActionId VotingProcedure)
```
michaelpj, line 411: It's not totally obvious to me that it's best to represent this as a map rather than a flat list with the Voter and GovernanceID included in VotingProcedure?

michaelpj (approved):
I do think we should try to add Haddock, either based on the explanations in the spec or the CIP.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,581 @@
...
+    TxCertDelegStaking V2.Credential Delegatee
+  | -- | Register and delegate staking credential to a Delegatee in one certificate. Noter that
+    -- deposit is mandatory.
+    TxCertRegDeleg V2.Credential Delegatee V2.Value
+  | -- | Register a DRep with mandatory deposit value and an optional Anchor
+    TxCertRegDRep DRepCredential V2.Value (Haskell.Maybe Anchor)
+  | -- | Update DRep's optional Anchor
+    TxCertUpdateDRep DRepCredential (Haskell.Maybe Anchor)
+  | -- | UnRegister a DRep with mandatory refund value
+    TxCertUnRegDRep DRepCredential V2.Value
```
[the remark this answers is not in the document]
lehins: The answer to this is the as the other one. The CIP is wrong, because we unregister DRep immediately.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
lehins: Well on the ledger side it has to be a nested Map for correctness and efficiency.

* Order of votes does not matter, so we don't need it to be a list
* We ensure there are no duplicate votes in the same transaction, for that (Voter, GovernanceActionId) pair must be unique, hence the usage of Map.
* It is a nested Map, because it is a more efficient representation on the wire (eg. a single DRep making many votes in the same transaction, then we only encode Voter once, rather than as many times as there are votes). And it also makes the rules implementation easier as well.

In any case, those are Ledger related reasons for the structure. For Plutus, you'll know better how it should look like. We can just as easily convert it into a flat list of three tuples if you want:

```haskell
 , txInfoVotingProcedures      :: [Voter, GovernanceActionId, VotingProcedure]
```
or a flat Map:
```haskell
 , txInfoVotingProcedures      :: Map (Voter, GovernanceActionId) VotingProcedure
```

Doesn't matter to me.

michaelpj: I think it's fine this way, especially if that's how it is in the ledger. It just occured to me that we don't have a very clear rationale for why some of these fields are "indexed" by something and others aren't.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,581 @@
...
+      , "anchorDataHash:" <+> pretty anchorDataHash
+      ]
+
+instance PlutusTx.Eq Anchor where
+  {-# INLINEABLE (==) #-}
+  Anchor a b == Anchor a' b' = a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
+
+data TxCert
+  = -- | Register staking credential with an optional deposit amount
+    TxCertRegStaking V2.Credential (Haskell.Maybe V2.Value)
```
bezirg, line 132: What is the rationale with qualifying Maybe, Integer, Bool , etc with `Haskell`? These datatypes should be first and foremost targetting `PlutusTx`, right? 

The fact that these `Maybe,Integer,etc` are the same for PlutusTx and Haskell is orthogonal.

I am just asking why; not that there is sth wrong with them.

bezirg (approved):
Looks good. It would help to add some inline comments or a NOTE  in V3/Contexts.hs about what has changed since V2 (ADDED/CHANGED)

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
bezirg: I agree that this it is useful to have in haddock what has changed since previous version (v2 or v1)

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,581 @@
...
+    ( Haskell.Eq
+    , PlutusTx.Eq
+    , PlutusTx.Ord
+    , PlutusTx.Show
+    , PlutusTx.ToData
+    , PlutusTx.FromData
+    , PlutusTx.UnsafeFromData
+    )
+  deriving
+    ( -- | from hex encoding
```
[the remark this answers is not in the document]
bezirg: @michaelpj Buy @zliu41 derives `IsString` `via LedgerBytes` which seems to decode via hex:  https://github.com/input-output-hk/plutus/blob/6ad87e1a1c2bb85d6f1ab6adc6999f0f5045f5e5/plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs#L92-L93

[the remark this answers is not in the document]
michaelpj: Hmmmm. Does that work in plutus-tx? not sure

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
zliu41: I just want to make it abundantly clear which things are from Haskell, and which things are from the Plutus Tx library. Especially since we don't always use explicit or qualified imports.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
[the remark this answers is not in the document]
zliu41: I simply copied this part from `PubKeyHash` and several other ledger types, so yeah it should work for plutus tx.

zliu41:
> I do think we should try to add Haddock, either based on the explanations in the spec or the CIP.

> Looks good. It would help to add some inline comments or a NOTE in V3/Contexts.hs about what has changed since V2 (ADDED/CHANGED)

Agree with both. @lehins @WhatisRT Do you happen to have a description of the changes compared to V2 that we can put in the module as a Note, or link to?

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,581 @@
...
+  = -- TODO: this is currently empty.
+    ParameterChange
+  | -- | proposal to update protocol version
+    HardForkInitiation ProtocolVersion
+  | TreasuryWithdrawals (Map V2.Credential V2.Value)
+  | NoConfidence
+  | NewCommittee
+      [ColdCommitteeCredential] -- ^ Old committee
+      Committee -- ^ New Committee
+  | NewConstitution Constitution
```
lehins, line 329: Sorry for the last minute changes, but we realized today that one piece is missing from here, namely the previous governance action id:

```suggestion
    ParameterChange GovernanceActionId
  | -- | proposal to update protocol version
    HardForkInitiation GovernanceActionId ProtocolVersion
  | TreasuryWithdrawals (Map V2.Credential V2.Value)
  | NoConfidence GovernanceActionId
  | NewCommittee
      GovernanceActionId
      [ColdCommitteeCredential] -- ^ Old committee
      Committee -- ^ New Committee
  | NewConstitution GovernanceActionId Constitution
```

`TreasuryWithdrawal` and `InfoAction` do not require one.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
lehins: Yeah,I think `DRep` is a perfect name for this type. That's what we call it in ledger too.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,584 @@
...
+    , -- | using hex encoding
+      Haskell.Show
+    , -- | using hex encoding
+      Pretty
+    )
+    via V2.LedgerBytes
+
+data Anchor = Anchor
+  { anchorUrl      :: PlutusTx.BuiltinByteString -- Arbitrary Url
+  , anchorDataHash :: AnchorDataHash -- Blake2b_256 of some off-chain data
```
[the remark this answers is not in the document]
zliu41: I'm not familiar with the intricacies of CIP-1694 enough to tell whether or not people will want to access the Anchor URL in their scripts. I'll leave it here fr now but add a TODO to revisit it later.

lehins:
> Do you happen to have a description of the changes compared to V2 that we can put in the module as a Note, or link to?

@zliu41 No, we don't have anything like that yet.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -0,0 +1,586 @@
...
+      IsString
+    , -- | using hex encoding
+      Haskell.Show
+    , -- | using hex encoding
+      Pretty
+    )
+    via V2.LedgerBytes
+
+data Constitution = Constitution
+  { constitutionHash   :: ConstitutionHash -- Balke2b_256 of some off-chain data
```
lehins, line 286: One more thing. This will become an Anchor, since we need to record the url where the constitution lives. So, whatever you decide about the Anchor url, will also apply to the constitution url
```suggestion
  { constitutionAnchor :: Anchor
```

zliu41:
Going to merge once CI passes. Will add the Note summarizing the changes once we finalize the types.

PR #5461, opened 2023-08-08
PLT-894: Improve the error reporting of 'CBOR.DeserialiseFailure'

bezirg:
The ledger team has requested long time ago,  to have more descriptive errors when a script fails to deserialize because of some CBOR problem.

The `cborg` library returns plain messages (strings) as deserialization failure reasons, and this PR maps only those that can occur during a plutus script deserialisation to some datatypes.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-ledger-api/src/Codec/CBOR/Extras.hs
+++ b/plutus-ledger-api/src/Codec/CBOR/Extras.hs
@@ -1,11 +1,19 @@
-module Codec.CBOR.Extras where
+-- editorconfig-checker-disable-file
+{-# LANGUAGE LambdaCase #-}
+module Codec.CBOR.Extras
+    ( SerialiseViaFlat (..)
+    , decodeViaFlat
+    , DeserialiseFailureInfo (..)
+    , DeserialiseFailureReason (..)
+    , readDeserialiseFailureInfo
+    ) where
 
 import Codec.CBOR.Decoding as CBOR
+import Codec.CBOR.Read as CBOR
 import Codec.Serialise (Serialise, decode, encode)
 import Data.Either.Extras
 import Flat qualified
 import Flat.Decoder qualified as Flat
-
 
 -- | Newtype to provide 'Serialise' instances for types with a 'Flat' instance that
 -- just encodes the flat-serialized value as a CBOR bytestring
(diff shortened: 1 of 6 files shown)


plutus-ledger-api/src/Codec/CBOR/Extras.hs
```diff
@@ -20,3 +28,40 @@ decodeViaFlat decoder = do
...
+        -- <https://github.com/well-typed/cborg/blob/cborg-0.2.9.0/cborg/src/Codec/CBOR/Read.hs#L226>
+        -- <https://github.com/well-typed/cborg/blob/cborg-0.2.9.0/cborg/src/Codec/CBOR/Read.hs#L1424>
+        -- <https://github.com/well-typed/cborg/blob/cborg-0.2.9.0/cborg/src/Codec/CBOR/Read.hs#L1441>
+        "end of input"   -> EndOfInput
+        -- Relevant Sources:
+        -- <https://github.com/well-typed/cborg/blob/cborg-0.2.9.0/cborg/src/Codec/CBOR/Read.hs#L1051>
+        "expected bytes" -> ExpectedBytes
+        msg              -> OtherReason msg
+
+data DeserialiseFailureInterpreted
```
michaelpj, line 55: Haddock?

I think the "interpreted" here is not that useful for a consumer of this type. I think we could just say "DeserialiseFailureError" or something

michaelpj (approved):
Fine I guess. We could make an issue upstream, I think actually modifying it would probably be too much work.

plutus-ledger-api/src/Codec/CBOR/Extras.hs
bezirg: I don't like Interpreted, but I also don't like your suggestion DeserialiseFailureError  :D .

michaelpj: So the thing I don't like is that "interpreted" is about how it was produced, not what it is.

bezirg: I turned it into `DeserialiseFailureInfo` and `readDeserialiseFailureInfo`. What do u think?

michaelpj: Sure

PR #5462, opened 2023-08-08
One of the references in the PLC specification had the wrong title

kwxm:


kwxm:
This is a small correction to a reference in the PLC specification: I'll merge it without review.

PR #5463, opened 2023-08-08
PLT-5391: BLS12-381 conformance tests using independently-verified inputs.

kwxm:
See README.md.  This was extremely tedious.


michaelpj (approved):
I have not verified these, obviously, but I at least took a look that the test outputs seem to match success/failure with what is expected.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/bls12_381-cardano-crypto-tests/G1/arith/add/add.uplc.expected
kwxm, line 1: Looking at these again, maybe I should have put the expected value in the input file and checked for equality with the computed value there, so that all of the `.expected` files contain `(con bool True)` (or failure).  That would make the test file self-contained and make it more obvious that the stuff in the golden files is correct.

kwxm: >  I should have put the expected value in the input file and checked for equality with the computed value

I've` done that now

PR #5469, opened 2023-08-09
Add blake2b-224 and keccak-256 to the specification

kwxm:
This updates the specification to include the new hash functions.  Much of the work involved digging up definitive specifications of these (and the other hash functions) for the references.


kwxm:
This is PLT-6312.

doc/plutus-core-spec/plutus-core-specification.bib
```diff
@@ -704,3 +704,53 @@ @misc{BLS12-381
...
+  language = {en},
+}
+
+@misc{IETF-Blake2,
+    series =    {Request for Comments},
+    number =    7693,
+    howpublished =  {RFC 7693, \url{https://www.rfc-editor.org/info/rfc7693}},
+    publisher = {RFC Editor},
+    doi =       {10.17487/RFC7693},
+    url =       {https://www.rfc-editor.org/info/rfc7693},
```
iquerejeta, line 738: For IETF documents I usually cite `https://datatracker.ietf.org/`. Also seems that for other IETF documents we are citing datatracker, so maybe makes sense to be consistent here
https://datatracker.ietf.org/doc/html/rfc7693

iquerejeta (approved):
Modified text looks good. I saw that we are citing `IETF-pairing-friendly-curves`, which is already expired and archived, so I wonder if it's worth keeping that reference.

michaelpj (approved):

doc/plutus-core-spec/plutus-core-specification.bib
kwxm: > For IETF documents I usually cite `https://datatracker.ietf.org/`. Also seems that for other IETF documents we are citing datatracker, so maybe makes sense to be consistent here https://datatracker.ietf.org/doc/html/rfc7693

I got that entry by clicking on the `bibtex` entry on the relevant datatracker page!  Oddly, datatracker gives you a reference to rfc-editor.org.  There's also an RFC which proposes a [bibtex format](https://datatracker.ietf.org/doc/html/draft-carpenter-rfc-citation-recs-01#section-6.2) for RFCs, but the datatracker bibtex doesn't conform to that. :(

We do refer to quite a number of RFCs though, so I'll try to make the bibtex consistent for those.

kwxm:
> Modified text looks good. I saw that we are citing `IETF-pairing-friendly-curves`, which is already expired and archived, so I wonder if it's worth keeping that reference.

I wanted to have something definitive, and that seemed like the most reasonable candidate (and I think it's what `blst`implements).  Can you suggest anything better?

iquerejeta:
> wanted to have something definitive

Understood. I thought that the link of Sean Bowe was a bit more detailed, but just saw it is not the case. That's fine then 👍

PR #5471, opened 2023-08-10
[Test] [Names] Add more scoping tests

effectfully:
This utilizes scoping tests for all PIR transformations that we have (previously it was only renaming).

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Check/Scoping.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Check/Scoping.hs
@@ -1,20 +1,28 @@
 {-# LANGUAGE ConstraintKinds       #-}
 {-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE LambdaCase            #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE UndecidableInstances  #-}
+-- TODO: add @NoFieldSelectors@ once we are past antiquity, so that 'ScopeError' doesn't generate
+-- partial field selectors.
 
 module PlutusCore.Check.Scoping where
 
 import PlutusCore.Name
 import PlutusCore.Quote
 
-import Control.Monad (join, unless, (<=<))
+import Control.Monad (join, unless)
+import Data.Bifunctor
 import Data.Coerce
 import Data.List.NonEmpty (NonEmpty)
 import Data.List.NonEmpty qualified as NonEmpty
 import Data.Map.Strict as Map
 import Data.Maybe
 import Data.Set as Set
+import Text.Pretty
+import Text.PrettyBy
 
 {- Note [Example of a scoping check]
 Consider the following type:
@@ -99,6 +107,9 @@ data NameAnn
     | NotAName
     deriving stock (Show)
 
+instance Pretty NameAnn where
+    pretty = viaShow
+
 class ToScopedName name where
     toScopedName :: name -> ScopedName
 
@@ -161,9 +172,12 @@ data ScopeEntry
 -- | A 'ScopeInfo' is a set of 'ScopedName's for each of the 'ScopeEntry'.
 -- If a 'ScopeEntry' is not present in the map, the corresponding set of 'ScopeName's is considered
 -- to be empty.
-newtype ScopeInfo = ScopeInfo
-    { unScopeInfo :: Map ScopeEntry (Set ScopedName)
-    } deriving stock (Show)
+newtype ScopeInfo = ScopeInfo (Map ScopeEntry (Set ScopedName))
+    deriving stock (Show)
+
+-- Defining manually because of plan to move to @NoFieldSelectors@.
+unScopeInfo :: ScopeInfo -> Map ScopeEntry (Set ScopedName)
+unScopeInfo = coerce
 
 -- | Extract the set stored in the provided 'ScopeInfo' at the provided 'ScopeEntry'.
 to :: ScopeEntry -> ScopeInfo -> Set ScopedName
@@ -173,27 +187,40 @@ emptyScopeInfo :: ScopeInfo
 emptyScopeInfo = ScopeInfo Map.empty
 
 -- | Check if a set is empty and report an error with the set embedded in it otherwise.
-checkEmpty :: (Set ScopedName -> ScopeError) -> Set ScopedName -> Either ScopeError ()
-checkEmpty err s = unless (Set.null s) . Left $ err s
+checkEmptyOn
+    :: (Set ScopedName -> Set ScopedName -> Set ScopedName)
+    -> (Set ScopedName -> Set ScopedName -> ScopeError)
+    -> Set ScopedName
+    -> Set ScopedName
+    -> Either ScopeError ()
+checkEmptyOn f err s1 s2 = unless (Set.null $ f s1 s2) . Left $ err s1 s2
 
--- | Merge two 'ScopeInfo's checking that they do not intersect along the way.
+-- | Merge two 'ScopeInfo's checking that binders in them do not intersect along the way.
 mergeScopeInfo :: ScopeInfo -> ScopeInfo -> Either ScopeError ScopeInfo
 mergeScopeInfo si1 si2 = do
     let disappearedBindings1 = to DisappearedBindings si1
         disappearedBindings2 = to DisappearedBindings si2
         appearedBindings1    = to AppearedBindings    si1
         appearedBindings2    = to AppearedBindings    si2
-        duplicateDisappearedBindings = disappearedBindings1 `Set.intersection` disappearedBindings2
-        duplicateAppearedBindings    = appearedBindings1    `Set.intersection` appearedBindings2
-    checkEmpty DuplicateBindersInTheInput  duplicateDisappearedBindings
-    checkEmpty DuplicateBindersInTheOutput duplicateAppearedBindings
+    checkEmptyOn
+        Set.intersection
+        DuplicateBindersInTheInput
+        disappearedBindings1
+        disappearedBindings2
+    checkEmptyOn
+        Set.intersection
+        DuplicateBindersInTheOutput
+        appearedBindings1
+        appearedBindings2
     Right $ coerce (Map.unionWith Set.union) si1 si2
 
+-- We might want to use @Validation@ or something instead of 'Either'.
 -- @newtype@-ing it for the sake of providing very convenient 'Semigroup' and 'Monoid' instances.
-newtype ScopeErrorOrInfo = ScopeErrorOrInfo
-    -- We might want to use @Validation@ or something instead of 'Either'.
-    { unScopeErrorOrInfo :: Either ScopeError ScopeInfo
-    }
+newtype ScopeErrorOrInfo = ScopeErrorOrInfo (Either ScopeError ScopeInfo)
+
+-- Defining manually because of plan to move to @NoFieldSelectors@.
+unScopeErrorOrInfo :: ScopeErrorOrInfo -> Either ScopeError ScopeInfo
+unScopeErrorOrInfo = coerce
 
 instance Semigroup ScopeErrorOrInfo where
     ScopeErrorOrInfo errOrInfo1 <> ScopeErrorOrInfo errOrInfo2 =
@@ -201,6 +228,13 @@ instance Semigroup ScopeErrorOrInfo where
 
 instance Monoid ScopeErrorOrInfo where
     mempty = ScopeErrorOrInfo $ Right emptyScopeInfo
+
+-- | Whether it's OK if the pass removes bindings. A renamer isn't supposed to do that, but for
+-- example an inliner may do it, since it's basically the entire point of an inliner.
+data BindingRemoval
+    = BindingRemovalOk
+    | BindingRemovalNotOk
+    deriving stock (Show, Eq)
 
 -- ########################################################################
 -- ## Main class for collecting scope information and relevant functions ##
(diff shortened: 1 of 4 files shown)


plutus-core/plutus-ir/test/NamesSpec.hs
```diff
@@ -2,19 +2,69 @@ module NamesSpec
...
 
 import Test.Tasty
 
 names :: TestTree
-names =
-    testGroup "names"
-        [ test_scopingGood genProgram rename
-        , test_scopingBad genProgram markNonFreshProgram renameProgramM
-        ]
+names = testGroup "names"
```
effectfully, line 31: Note to self: this needs a Note.

plutus-core/plutus-ir/test/NamesSpec.hs
```diff
@@ -2,19 +2,69 @@ module NamesSpec
...
+    , -- We say that it's fine to remove bindings, because they never actually get removed,
+      -- because the scope checking machinery doesn't create unused bindings, every binding
+      -- gets referenced at some point at least once (usually very close to the binding site).
+      -- So this test doesn't really test much.
+      test_scopingGood "dead code elimination" genTerm BindingRemovalNotOk PrerenameNo $
+        removeDeadBindings def
+    , test_scopingGood "constant folding" genTerm BindingRemovalNotOk PrerenameYes $
+        pure . evaluateBuiltins False def defaultBuiltinCostModel
+    , test_scopingBad "inlining" genTerm BindingRemovalOk PrerenameYes $
+        Inline.inline mempty def
```
effectfully, line 47: It's `test_scopingBad`, because the inliner does not preserve global uniqueness as it promises due to the lack of marking (or initial renaming of the entire program, which would perform marking too). We discussed that in Slack, I'll create a Jira ticket for this issue.

plutus-core/plutus-ir/test/NamesSpec.hs
```diff
@@ -2,19 +2,69 @@ module NamesSpec
...
+    , test_scopingGood "match-against-known-constructor" genTerm BindingRemovalNotOk PrerenameNo $
+        knownCon
+    , test_scopingGood "floating bindings inwards" genTerm BindingRemovalNotOk PrerenameNo $
+        In.floatTerm def True
+    -- Can't test 'Out.floatTerm', because it requires the type of annotations to implement
+    -- 'Semigroup' and it's not clear what that means for 'NameAnn'.
+    , test_scopingGood "merging lets" genTerm BindingRemovalNotOk PrerenameYes $
+        pure . letMerge
+    , test_scopingBad "compilation of non-strict bindings" genTerm BindingRemovalOk PrerenameYes $
+        compileNonStrictBindings True
```
effectfully, line 57: Same thing as with the inliner, expect may not in fact be problematic (perhaps that transformation doesn't care about preserving global uniqueness).

Also, `test_scopingBad` isn't a very good name... All the names were based on the assumption that the machinery is only useful for testing renamers, I wish I realized it earlier that we could use the machinery for testing all the other passes too.

plutus-core/plutus-core/src/PlutusCore/Check/Scoping.hs
```diff
@@ -279,14 +301,14 @@ data ScopeError
...
-    | OldBindingsDiscordWithBoundVariables !(Set ScopedName)
-    | OldBindingsDiscordWithOutOfScopeVariables !(Set ScopedName)
-    | NewBindingsDiscordWithBoundVariables !(Set ScopedName)
-    | OldBindingsClashWithFreeVariables !(Set ScopedName)
-    | OldBindingsClashWithNewBindings !(Set ScopedName)
-    | NewBindingsClashWithFreeVariabes !(Set ScopedName)
+    | DuplicateBindersInTheInput !(Set ScopedName) !(Set ScopedName)
+    | DuplicateBindersInTheOutput !(Set ScopedName) !(Set ScopedName)
+    | DisappearedBindingsDiscordWithBoundVariables !(Set ScopedName) !(Set ScopedName)
+    | DisappearedBindingsDiscordWithOutOfScopeVariables !(Set ScopedName) !(Set ScopedName)
```
effectfully, line 307: Makes it much easier to debug stuff when you can see what exactly disappeared and what the out of scope variables are. Same for all the other errors.

plutus-core/plutus-ir/test/NamesSpec.hs
effectfully: > Note to self: this needs a Note.

Done.

plutus-core/testlib/PlutusCore/Test.hs
```diff
@@ -423,47 +425,145 @@ brokenRename ::
...
+- some passes such as inlining may remove bindings and there's special support implemented for
+  handling this: when invoking either 'test_scopingGood' or 'test_scopingBad' you need to provide a
+  'BindingRemoval' argument specifying whether binding removal is expected for the pass. Conversily,
+  make it 'BindingRemovalOk' whenever you use 'test_scopingBad' to emphasize that even with binding
+  removal allowed tests still fail
+- some passes do not perform renaming of the input themselves, in that case you need to provide
+  'PrerenameYes' for the 'Prerename' argument that both the test runners expect. It doesn't matter
+  whether the pass relies on global uniqueness itself, because the scoping tests rely on it anyway.
+  If the pass renames its input, and only in this case, provide 'PrerenameNo' for the 'Prerename'
+  argument, this will allow the scoping tests to ensure that the pass does indeed rename its input
```
michaelpj, line 455: Still think it would be easy enough to just test `rename . pass` in that case ...

plutus-core/testlib/PlutusCore/Test.hs
```diff
@@ -423,47 +425,145 @@ brokenRename ::
...
+
+There's also 'test_scopingSpoilRenamer', this one is used to test that the scoping tests do catch
+various kinds of bugs. Don't use it with any passes that aren't renamers.
+
+All in all, in order to use the scoping tests you have to understand how they work, which is an
+unfortunate but seemingly inevitable requirement. On the bright side, it's worth the effort, because
+the tests do catch bugs occasionally.
+
+Whenever you use 'test_scopingBad' make sure to explain why, so that it's clear whether there's
+something wrong with the pass or it's just a limitation of the scoping tests.
```
michaelpj, line 468: :+1:

plutus-core/testlib/PlutusCore/Test.hs
```diff
@@ -423,47 +425,145 @@ brokenRename ::
...
-  case catchEverything $ checkRespectsScoping (TPLC.runQuote . ren) prog of
-    Left exc         -> fail $ show exc
-    Right (Left err) -> fail $ show err
-    Right (Right ()) -> success
-
--- | Test that a good renamer does not destroy scoping.
+      prep = runPrerename preren
+  case catchEverything $ checkRespectsScoping bindRem prep (TPLC.runQuote . run) prog of
+    Left exc -> fail $ show exc
+    Right (before, after, Left err) -> fail $ concat
```
michaelpj, line 502: Could return a proper error with a pretty instance instead of a tuple

plutus-core/plutus-ir/test/NamesSpec.hs
effectfully: Damn it, draft PRs Roman, draft them! It's the only way to keep Michael away while you're working on the PR!

(just kidding, although I indeed had addressed your comment before seeing it)

michaelpj: I've got to procrastinate somehow!!

plutus-core/plutus-core/src/PlutusCore/Check/Scoping.hs
```diff
@@ -353,29 +375,61 @@ so we only need to consider three more relations:
...
-        disappearedBindings `symmetricDifference` disappearedVariables
-    checkEmpty NewBindingsDiscordWithBoundVariables $
-        appearedBindings `symmetricDifference` appearedVariables
-    checkEmpty OldBindingsClashWithFreeVariables $
-        disappearedBindings `Set.intersection` stayedFreeVariables
-    checkEmpty OldBindingsClashWithNewBindings $
-        appearedBindings  `Set.intersection` disappearedBindings
-    checkEmpty NewBindingsClashWithFreeVariabes $
-        appearedBindings `Set.intersection` stayedFreeVariables
+    unless (bindRem == BindingRemovalOk) $ do
```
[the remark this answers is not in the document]
effectfully: > Use a `Writer` of problems rather than `Either`. This makes sense since finding a problem doesn't stop you finding more problems!

Yeah, I put a similar comment into the file when I originally wrote it:

```
-- We might want to use @Validation@ or something instead of 'Either'.
```

I agree with you, just don't feel like refactoring the machinery...

plutus-core/plutus-core/src/PlutusCore/Check/Scoping.hs
```diff
@@ -353,29 +375,61 @@ so we only need to consider three more relations:
...
-checkRespectsScoping ren =
-    checkScopeInfo <=< unScopeErrorOrInfo . collectScopeInfo . ren . runQuote . establishScoping
+-- | Check if a pass respects scoping.
+--
+-- Returns the thing that the scoping tests run on, the result of the pass and the scope checking
+-- outcome, respectively.
+checkRespectsScoping
+    :: Scoping t
+    => BindingRemoval
+    -> (t NameAnn -> t NameAnn)  -- ^ For preparation before running the scoping tests for it.
```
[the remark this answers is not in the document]
effectfully: We can and it's how it used to be, but then we can't print the exact thing that `run` was run on, which, trust me, makes for a horrible debugging experience.

plutus-core/plutus-ir/test/NamesSpec.hs
```diff
@@ -2,19 +2,70 @@ module NamesSpec
...
+    , test_scopingGood "match-against-known-constructor" genTerm BindingRemovalNotOk PrerenameYes $
+        pure . recSplit
+    , test_scopingGood "renaming" genProgram BindingRemovalNotOk PrerenameNo $
+        rename
+    , test_scopingSpoilRenamer genProgram markNonFreshProgram
+        renameProgramM
+    -- Can't test substitution procedures at the moment, because that requires generating
```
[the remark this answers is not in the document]
effectfully: The `Transform` folder contains a file with substitution stuff and I'm saying here that can't test those. I just thought it would be a good idea to run through the whole folder and address every file one way or another.

plutus-core/testlib/PlutusCore/Test.hs
effectfully: Only until you get a failure and need to understand what's going on. This whole thing is written in blood and tears.

plutus-core/plutus-core/src/PlutusCore/Check/Scoping.hs
```diff
@@ -353,29 +375,61 @@ so we only need to consider three more relations:
...
-    checkEmpty OldBindingsClashWithFreeVariables $
-        disappearedBindings `Set.intersection` stayedFreeVariables
-    checkEmpty OldBindingsClashWithNewBindings $
-        appearedBindings  `Set.intersection` disappearedBindings
-    checkEmpty NewBindingsClashWithFreeVariabes $
-        appearedBindings `Set.intersection` stayedFreeVariables
+    unless (bindRem == BindingRemovalOk) $ do
+        checkEmptyOn
+            symmetricDifference
+            DisappearedBindingsDiscordWithOutOfScopeVariables
```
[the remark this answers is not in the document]
effectfully: > I'm actually slightly confused about why this is bad, anyway 🤔

This is due to the setup that the scoping tests use, quoting the docs:

```
    In addition to that every binder should be decorated with one out-of-scope variable (annotated
    with 'StaysOutOfScopeVariable') and one in-scope one (annotated with 'DisappearsVariable').
```

A renamer is supposed to change every single binding to a different one (no clashes with old bindings are allowed and can't happen if marking is used) and it's supposed not to change any free variable (of which the out-of-scope ones are a subset and yes, we distinguish between out-of-scope variables and free ones in the machinery), hence what bindings disappear must coincide perfectly with what out-of-scope variables stay. 

All this is elaborated on in the haddock of `checkScopeInfo`:

```
{-| Check that each kind of 'Set' from 'ScopeInfo' relates to all other ones in a certain way.
We start with these three relations that are based on the assumption that for each binder we add
at least one out-of-scope variable and at least one in-scope one:

1. disappeared bindings should be the same as stayed out of scope variables
     (ensures that old bindings disappear via renaming and not via removal)
2. disappeared bindings should be the same as disappeared variables
     (ensures that old names consistently disappear at the binding and use sites)
3. appeared bindings should be the same as appeared variables
     (ensures that new names consistently appear at the binding and use sites)

Once we've ensured all of that, we're left with only three sets and 3C2 equals 3,
so we only need to consider three more relations:

1. disappeared bindings should not intersect with free variables
     (an internal sanity check)
2. appeared bindings should not intersect with disappeared bindings
3. appeared bindings should not intersect with free variables

The last two ensure that no new name has an old name's unique.
-}
```

zliu41 (commented):
After reading `Note [Scoping tests API]` I still don't get how this works. Is it possible to give a more formal definition of the passing criteria than "free names to stay free, bound names to change together with their binders and global uniqueness to be preserved"?

A pass can produce an output that looks totally different than the input:
- Free variables can be moved around arbitrarily. Does it simply check that the set of unique free variables stay the same, regardless of the locations or number of occurrences?
- For bound variables it is even less clear what is being checked, since in general it is not possible to map the bound variables in the output to those in the input: (1) the pass may do a renaming at the beginning, changing all bound variable names; (2) it may move things around substantially, e.g., two bindings may be swapped; or a binding may be floated from the top level into a much deeper location.

A more formal description may help understand how exactly the check works.

effectfully:
> After reading `Note [Scoping tests API]` I still don't get how this works. 

You're not supposed to! That Note is specifically about the API. In order to understand the innards of the machinery you need to read `Note [Example of a scoping check]`, which will also point you to the haddocks of `checkScopeInfo` specifying what is being checked once `ScopeInfo` is collected. Although collecting the `ScopeInfo` can fail too and that isn't documented so well... but at least those cases are more straightforward: it's literally "a bound name is supposed to disappear, a free name is supposed to stay and no duplicate binders are allowed in the input or the output".

> Free variables can be moved around arbitrarily. Does it simply check that the set of unique free variables stay the same, regardless of the locations or number of occurrences?

Yes. The scoping tests are not designed to catch such errors, because any other test could, in particular we have the `rename x == x` tests (see e.g. [this](https://github.com/input-output-hk/plutus/blob/31c231faeb3d3829c83279ba5a7629457e62c10c/plutus-core/plutus-core/test/Names/Spec.hs#L55-L67) one). Scoping tests are designed to catch horrible alien undebuggable stuff such as name duplication and clash of bound variables with free ones due to the lack of marking (which is exactly what happened in this PR with the inliner). This is to prevent the situation where a test fails only after 8000+ iterations on a program whose size is 10+ screens of S-expressions and you have no way of reducing the size of that program, so the only option you have is to go and review every single line of relevant code hoping something will jump on you (this is not a hypothetical situation, it's exactly the experience I had with our type checker in 2018).

> in general it is not possible to map the bound variables in the output to those in the input

Yes, and we don't attempt to. We only check that what is supposed to disappear disappears, what is supposed to stay stays and that these things happen consistently. No usage counting, to attempt to guess which names became what etc. This is a pretty general machinery, which is how we're able to test so many passes using it.

zliu41:
Ok, sounds reasonable. One last question: what's the reason for the existence of `PrerenameYes` and `PrerenameNo`? I don't understand why a pass renaming the input (or not) would matter to scope checking.

effectfully:
> One last question: what's the reason for the existence of `PrerenameYes` and `PrerenameNo`? I don't understand why a pass renaming the input (or not) would matter to scope checking.

I thought I answered this in the docs?

```
- some passes do not perform renaming of the input themselves, in that case you need to provide
  'PrerenameYes' for the 'Prerename' argument that both the test runners expect. It doesn't matter
  whether the pass relies on global uniqueness itself, because the scoping tests rely on it anyway.
  If the pass renames its input, and only in this case, provide 'PrerenameNo' for the 'Prerename'
  argument, this will allow the scoping tests to ensure that the pass does indeed rename its input
```

Basically, `PrerenameNo` is to

>  allow the scoping tests to ensure that the pass does indeed rename its input

We could always rename the input before running the pass on it and feeding the result to the scoping tests (we need input to satisfy global uniqueness and the pass to preserve it for the scoping tests to do their magic -- if the pass doesn't preserve global uniqueness by design, then the scoping tests are only useful for confirming that the pass indeed doesn't preserve global uniqueness), but that would basically be testing less as then there would be no way to specify and test that some pass renames its input, which I believe is important information -- in particular, the bug in the inliner that we've been discussing wouldn't exist if the inliner simply renamed its input (without quitting `Quote`).

And if we're testing a renamer, it doesn't make any sense to prerename the input before testing the pass that is the very same renaming procedure, so at the very least we need `PrerenameYes` vs `PrerenameNo` for that reason.

zliu41 (approved):

plutus-core/testlib/PlutusCore/Test.hs
effectfully: > Could return a proper error with a pretty instance instead of a tuple

Done.

effectfully:
Gonna claim I've addressed the comments, merging.

effectfully:
CI doesn't like me :(

michaelpj:
I checked on hydra. Obviously you should be getting a notification, but it's failing on 8.10 because `NoFieldSelectors` doesn't exist there, sorry :(

PR #5472, opened 2023-08-10
Integrate IOGX

zeme-wana:
*This PR is a work in progress, please don't look yet*
-----

As part of the ongoing effort to build a Developer Platform for IOG, the smart-contracts-dev-empowerment team is releasing [IOGX - a Flake Template for Haskell Projects at IOG](https://github.com/input-output-hk/iogx)

The vision is to provide a JSON-like, declarative interface to Nix, so that developers unfamiliar with the language may independently maintain and add to the Nix sources with minimum effort.

The goal is to integrate with all Smart Contracts repositories, and maybe more.

There are currently several other repositories using this: marlowe-cardano, antaeus, quickcheck-dynamic, marconi, stablecoin-plutus, dapps-certification.

-----

You might find that the shell looks a little different, if you don't like it you can tell me.
Notably, the `fix-*formatter*` scripts have gone, because the same result can be obtained just by using `pre-commit`.
E.g.
```bash
fix-stylish-haskell
fix-shellcheck
# -> Now becomes
pre-commit run stylish-haskell
pre-commit run shellcheck

# To run all formatters
pre-commit run

# To run on all files (not just the committed ones)
pre-commit run --all-files
```

---- 

Various documentation-related files will also have to be updated.
I'm preparing a separate PR for that


nix/per-system-outputs.nix
```diff
@@ -0,0 +1,13 @@
...
+
+{ nix, projects, ... }:
+
+{
+  documents = nix.plutus.latex-docs;
+
+  packages.plutus-metatheory-site = nix.plutus.plutus-metatheory-site;
+
+  # Needed by .github/workflows/hlint.yml
+  hlint = projects.default.hsPkgs.hlint.components.exe.hlint;
```
zeme-wana, line 12: Although... `.github/.hlint.yml` is:
```
- ignore: { }
```

So it's no-op, always, right?

We could delete it, and the relative workflow file, and rely on `hlint.enable = true;` in `./nix/formatters.nix`.
With this, `hlint` will be run with each `git commit`, and also manually with `pre-commit run hlint` and also automatically run in CI along with all other formatters.

But because pre-commit only works on committed files, in order to lint existing .hs sources, you will have to run `pre-commit run hlint --all-files` and fix everything up (which you'd wanna do in a separate PR).
Thoughts?

nix/formatters.nix
```diff
@@ -0,0 +1,20 @@
...
+  png-optimization.enable = true;
+
+  # TODO
+  # open a new PR
+  # set this to true
+  # run 'pre-commit run hlint --all-files'
+  # fix everything up
+  # push & merge
+  # remove this comment
+  hlint.enable = false;
```
zeme-wana, line 19: I deleted `.github/workflows/hlint.yml`, and `.github/.hlint.yml`, which was just:
```
- ignore: { }
```
Which is a no-op, always, right?

By setting `hlint.enable = true;` in `./nix/formatters.nix` instead, `hlint` will be run (using the top-level `.hlint.yml` file) with each `git commit`, and also manually with `pre-commit run hlint` and also automatically run in CI along with all other formatters.

But because pre-commit only works on committed files, in order to lint existing .hs sources, you will have to run `pre-commit run hlint --all-files` and fix everything up (which you'd wanna do in a separate PR).
Thoughts?

nix/cabal-project.nix
```diff
@@ -0,0 +1,124 @@
...
+    # -Werror for CI
+    # Only enable on the newer compilers. We don't care about warnings on the old ones,
+    # and sometimes it's hard to be warning free on all compilers, e.g. the unused
+    # packages warning is bad in 8.10.7
+    # (https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6130)
+
+    # Werror everything.
+    # This is a pain, see https://github.com/input-output-hk/haskell.nix/issues/519
+
+    plutus-benchmark.ghcOptions = l.mkIf isNotGhc8107 [ "-Werror" ];
```
michaelpj, line 102: This seems worse than the previous structure that just had a single if for the whole block of things... Same with the cross-compiling block. I don't think that flattening all the conditionals like this is actually helpful?

nix/plutus/agda-project.nix
```diff
@@ -1,16 +1,15 @@
...
-in
-
-library.haskell-nix.hackage-project {
+pkgs.haskell-nix.hackage-project {
   name = "Agda";
 
   version = "2.6.2.2";
 
-  compiler-nix-name = library.ghc-compiler-nix-name;
+  compiler-nix-name = "ghc927";
```
michaelpj, line 8: Shouldn't this be the default compiler?

nix/plutus/latex-docs.nix
```diff
@@ -0,0 +1,146 @@
+{ nix, inputs, pkgs, l, ... }:
+
+let
+
+  latex-docs = l.flip l.mapAttrValues nix.plutus.build-latex-doc {
+
+    cost-model-notes = {
```
michaelpj, line 7: fine I guess. I kind of did like the std thing of one package per file, but whatever

scripts/prepare-release.sh
```diff
@@ -1,45 +0,0 @@
-#!/usr/bin/env bash
```
michaelpj, line 1: Um, no? we use these

michaelpj: You put them in IOGX? I think that's a bad idea. These are some things I hacked together for `plutus`: there is 0 guarantee they will work for anyone else, and I don't want to go through `iogx` if I want to hack on them further.

michaelpj (commented):
sorry, I looked

michaelpj:
Stuff in the shell today:
```
🔨 Welcome to plutus-shell                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                           
[docs]                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                           
  autobuild-docs          - Start the autobuild server on localhost:8000                                                                                                                                                                                                                                                   
  build-docs              - Build the docs locally in doc/read-the-docs-site/_build                                                                                                                                                                                                                                        
  jekyll                  - Static site builder, used for the metatheory documentation                                                                                                                                                                                                                                     
  serve-docs              - nix build and serve the doc site on localhost:8002                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                           
[general commands]                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                           
  R                       - Statistical modelling tool                                                                                                                                                                                                                                                                     
  ShellCheck              - Shell file checker                                                                                                                                                                                                                                                                             
  editorconfig-checker    - Checker for editorconfig conformance                                                                                                                                                                                                                                                           
  fix-png-optimization    - Fix all PNG files in-place                                                                                                                                                                                                                                                                     
  menu                    - prints this menu                                                                                                                                                                                                                                                                               
  scriv                   - Manage changelogs                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                           
[haskell]                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                           
  agda                    - Agda and its standard library                                                                                                                                                                                                                                                                  
  cabal                   - Haskell build tool                                                                                                                                                                                                                                                                             
  cabal-fmt               - Cabal file formatter                                                                                                                                                                                                                                                                           
  fix-cabal-fmt           - Format all cabal files in-place                                                                                                                                                                                                                                                                
  fix-stylish-haskell     - Run stylish-haskell on all haskell files in-place                                                                                                                                                                                                                                              
  haskell-language-server - Haskell Language Server                                                                                                                                                                                                                                                                        
  hlint                   - Haskell linting tool                                                                                                                                                                                                                                                                           
  stylish-haskell         - Haskell code formatter                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                           
[nix]                                                                                                                                                                                                                                                                                                                      

  check-the-flake         - For nix maintainers: build everything in the flake 
  nixpkgs-fmt             - Nix code formatter
```
Output of `info` on this branch:
```
λ changelog
— assemble-changelog ∷ Assembles the changelog for PACKAGE at VERSION
— prepare-release ∷ Prepares to release PACKAGEs at VERSION
— update-version ∷ Updates the version for PACKAGE to VERSION

λ general
— info ∷ Print this message
— list-flake-outputs ∷ List the flake outputs buildable by nix
```

flake.nix
```diff
@@ -1,140 +1,33 @@
...
+
   inputs = {
-    nixpkgs = {
-      follows = "haskell-nix/nixpkgs";
-    };
-    std = {
-      url = "github:divnix/std";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
-    haskell-nix = {
```
michaelpj, line 23: I suspect I will want to pin `haskell.nix` also, but I guess this is fine for now

nix/agda-project.nix
```diff
@@ -1,16 +1,15 @@
...
-in
-
-library.haskell-nix.hackage-project {
+pkgs.haskell-nix.hackage-project {
   name = "Agda";
 
   version = "2.6.2.2";
 
-  compiler-nix-name = library.ghc-compiler-nix-name;
+  compiler-nix-name = "ghc927";
```
michaelpj, line 8: I'm unsure how much it matters what compiler version we use here... probably not that much. But at least we can do `ghc92` to get the latest GHC 9.2

nix/cabal-project.nix
```diff
@@ -0,0 +1,159 @@
...
+    # We would expect R to be pulled in automatically as it's a dependency of
+    # plutus-core, but it appears it is not, so we need to be explicit about
+    # the dependency on R here. Adding it as a buildInput will ensure it's
+    # added to the pkg-config env var.
+    shell = {
+      withHoogle = false;
+      buildInputs = [ pkgs.R ];
+    };
+
+    flake.variants = {
```
michaelpj, line 23: Personally I don't like the flake variants. I don't understand why people want to do `f(args).adjustFToHaveDifferentArgs` rather than `f(differentArgs)`. It's really not hard, and it's super hard to work out what's going on with all the clever variants and overriding stuff. What's Wrong With Functions, Damnit. But I've made enough of a fuss, I won't fight it.

nix/cabal-project.nix
```diff
@@ -0,0 +1,159 @@
+# editorconfig-checker-disable-file 
+
+{ repoRoot, inputs, lib, ... }:
+
+let
+  cabalProjectArgs = { config, pkgs, ... }: {
+
+    name = "plutus";
+
+    compiler-nix-name = lib.mkDefault "ghc927";
```
michaelpj, line 10:
```suggestion
    compiler-nix-name = lib.mkDefault "ghc92";
```

nix/cabal-project.nix
```diff
@@ -0,0 +1,159 @@
...
+    # the dependency on R here. Adding it as a buildInput will ensure it's
+    # added to the pkg-config env var.
+    shell = {
+      withHoogle = false;
+      buildInputs = [ pkgs.R ];
+    };
+
+    flake.variants = {
+      ghc927 = { }; # Alias for the default project
+      ghc927-profiled.modules = [{ enableProfiling = true; }];
```
michaelpj, line 25: I just did some work to only write the major version everywhere, now that haskell.nix supports that as a way of writing `compiler-nix-name`, can we keep that please?

nix/cabal-project.nix
```diff
@@ -0,0 +1,159 @@
...
+      ghc8107.compiler-nix-name = "ghc8107";
+      ghc962.compiler-nix-name = "ghc962";
+    };
+
+    sha256map = {
```
michaelpj, line 30: This is dead anyway!

nix/cabal-project.nix
```diff
@@ -0,0 +1,159 @@
...
+          prettyprinter-configurable.ghcOptions = [ "-Werror" ];
+        };
+      })
+    ];
+  };
+
+
+  project = lib.iogx.mkHaskellProject {
+    inherit cabalProjectArgs;
+    readTheDocs.siteFolder = "doc/read-the-docs-site";
```
michaelpj, line 140: why _is_ this part of the configuration of my cabal project? from a external viewpoint that makes no sense

nix/latex-documents.nix
```diff
@@ -0,0 +1,146 @@
+{ repoRoot, inputs, pkgs, lib, ... }:
+
+let
+
+  latex-documents = lib.flip lib.mapAttrValues nix.plutus.build-latex-doc {
```
michaelpj, line 5: If this is supposed to be readable by non-nix experts, this sort of thing makes it harder. I'd just write `build-latex-doc` at each use site, it's not that bad

nix/latex-documents.nix
```diff
@@ -0,0 +1,146 @@
...
+    utxoma-paper = {
+      name = "utxoma-paper";
+      description = "utxoma";
+      src = inputs.self + /doc/papers/utxoma;
+      texFiles = [ "utxoma.tex" ];
+    };
+  };
+
+
+  unraveling-recursion-paper =
```
michaelpj, line 85: can we put this in its own file? I kind of liked the std thing of having one derivation per file...

nix/make-shell.nix
```diff
@@ -0,0 +1,45 @@
+{ repoRoot, pkgs, ... }:
+
+cabalProject:
+
+{
```
michaelpj, line 5: I find the naming confusing here. "make X" is usually a function that makes an X, which this is not. I'm _guessing_ that it's the _arguments_ to the `mkShell` function, so I'd call it `mkShellArgs` or something.

It's also not a module? Or is it? Why is it getting the cabal project as an argument instead of accessing module settings? I'm generally confused.

nix/cabal-project.nix
```diff
@@ -0,0 +1,159 @@
...
+    shellFor = repoRoot.nix.make-shell;
```
michaelpj, line 141: The control flow here is super confusing. I define the cabal project here, which gets passed into `mkHaskellProject`, which internally passes in the project to `make-shell`, and then uses the result to call some other function???

nix/make-shell.nix
```diff
@@ -0,0 +1,45 @@
...
+    cabal-fmt.enable = true;
+    shellcheck.enable = true;
+    editorconfig-checker.enable = true;
+    nixpkgs-fmt.enable = true;
+    optipng.enable = true;
+    hlint.enable = false;
+  };
+
+
+  tools.haskellCompiler = cabalProject.args.compiler-nix-name;
```
michaelpj, line 44: It seems strange to me that we need to do this. Shouldn't this be done generically?

nix/outputs.nix
```diff
@@ -0,0 +1,72 @@
...
+  }
+  {
+    latex-documents = repoRoot.nix.latex-documents;
+
+    packages.plutus-metatheory-site = repoRoot.nix.plutus-metatheory-site;
+    packages.pre-commit-check = ghc927.pre-commit-check;
+    packages.read-the-docs-site = ghc927.read-the-docs-site;
+  }
+  {
+    hydraJobs.required = lib.iogx.mkHydraRequiredJob;
```
michaelpj, line 32: I'm kind of confused how this works :thinking:

nix/outputs.nix
```diff
@@ -0,0 +1,72 @@
...
+  }
+  (lib.optionalAttrs (system == "x86_64-linux" || system == "x86_64-darwin")
+    {
+      hydraJobs.plutus-metatheory-site = repoRoot.nix.plutus-metatheory-site;
+      hydraJobs.pre-commit-check = ghc927.iogx.pre-commit-check;
+
+      hydraJobs.ghc8107 = ghc8107.iogx.hydraJobs;
+      hydraJobs.ghc927 = ghc927.iogx.hydraJobs;
+      hydraJobs.ghc962 = ghc962.iogx.hydraJobs;
```
michaelpj, line 41: what happened to automatic build matrix support?

michaelpj (commented):
Main things that seem odd to me:
- The input to `outputs` is weird
- The control flow between defining the shell and the "project" is very confusing

Overall, looking at what's here now... it looks a lot to me like what we had before `std`. The main difference is that it's using the new flake outputs for haskell.nix projects and the variants support to avoid having to have function arguments (which I have philosophical disagreements with, but whatever). `outputs.nix` is... pretty much just the flake outputs. 

Yes, there are some useful utility functions coming from `iogx`, but at this point it mostly looks like:
- A haddock builder
- A RTD builder
- Various convenience scripts
- Something for turning a haskell.nix shell into a shell with more stuff.

Perhaps this is the inevitable endpoint of what I've been asking for: just a directory of more-or-less vanilla Nix files containing derivations, with some utilities from elsewhere.

flake.nix
zeme-wana: Will do

nix/outputs.nix
```diff
@@ -0,0 +1,72 @@
...
+  ghc927-profiled = cabalProject.projectVariants.ghc927-profiled;
+  ghc962 = cabalProject.projectVariants.ghc962;
+
+in
+
+[
+  {
+    inherit cabalProject;
+  }
+  {
```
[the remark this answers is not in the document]
zeme-wana: These are not modules, we are not using modules anywhere.
We are just using options to validate the input arguments of `mkFlake`, `mkProject` and `mkShell`.

Yes `outputs` returning a list is a little weird, it should just return a single attrset i.e. the flake outputs for each system

nix/cabal-project.nix
```diff
@@ -0,0 +1,159 @@
...
+          # upstream fix, there's nothing we can do about it other than
+          # disabling -Werror on it.
+          prettyprinter-configurable.ghcOptions = [ "-Werror" ];
+        };
+      })
+    ];
+  };
+
+
+  project = lib.iogx.mkHaskellProject {
```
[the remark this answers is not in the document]
zeme-wana: What you get is the `iogx` overlay added to each `haskell-nix.cabalProject` variant.
In addition you get a `devShell` for each project variant, which includes the `pre-commit-check`.
Then you get the `combined-haddock` derivation.
plutus and plutus-apps use the combined-haddock to build the RTD site, so the RTD site needs to reference the haskell project somehow, since it also needs to add some scripts to the shell.

Separating the RTD into its own `lib.iogx.mkReadTheDocsSite` means more work for end user.

zeme-wana:
- Update `agda-project.nix` `compiler-nix-name` to `ghc92`
- Brought back the version scripts
- Removed` #!/usr/bin/env bash` from version scripts
- Renamed option `shellFor` into `shellArgsForProjectVariant`
- Pinned `haskell.nix`
- Added support for `ghcXX*` versioning 
- Removed `sha256map` from `cabal-project.nix`
- Use `build-latex-doc` at each use site
- Move `unraveling-recursion-paper` to its own file
- Simplified `nix/outputs.nix`

nix/cabal-project.nix
zeme-wana: Explained in a comment below

flake.nix
```diff
@@ -1,140 +1,40 @@
...
-      url = "github:input-output-hk/haskell.nix";
-      inputs = {
-        hackage.follows = "hackage-nix";
-      };
+    iogx = {
+      url = "github:input-output-hk/iogx?ref=v4";
+      inputs.hackage.follows = "hackage";
+      inputs.CHaP.follows = "CHaP";
+      inputs.haskell-nix.follows = "haskell-nix";
+      inputs.nixpkgs.follows = "haskell-nix/nixpkgs-2305";
```
michaelpj, line 11: this changes our nixpkgs pin, not sure if we want that. I guess it's fine

nix/latex-documents.nix
```diff
@@ -0,0 +1,86 @@
+{ repoRoot, inputs, pkgs, lib, ... }:
+
+let
+
+  latex-documents = {
+
+    cost-model-notes = repoRoot.nix.build-latex-doc {
```
michaelpj, line 7: okay, I mean you could do 
```
let 
   inherit (repoRoot.nix) build-latex-doc;
   latex-documents = {...
```

nix/project.nix
```diff
@@ -0,0 +1,160 @@
+# editorconfig-checker-disable-file 
+
+{ repoRoot, inputs, lib, ... }:
+
+let
+  cabalProjectArgs = { config, pkgs, ... }: {
+
+    name = "plutus";
+
+    compiler-nix-name = lib.mkDefault "ghc92";
```
michaelpj, line 10: If we're not merging the option sets, which I assume we aren't, then I guess this mkDefault does nothing?

michaelpj: Or does it mean we can override it later in the variants? worth a comment

nix/project.nix
```diff
@@ -0,0 +1,160 @@
...
+
+    src = ../.;
+
+    # We would expect R to be pulled in automatically as it's a dependency of
```
michaelpj, line 14: misplaced, should be over the buildInputs

nix/project.nix
```diff
@@ -0,0 +1,160 @@
...
+    # the dependency on R here. Adding it as a buildInput will ensure it's
+    # added to the pkg-config env var.
+    shell = {
+      withHoogle = false;
+      buildInputs = [ pkgs.R ];
+    };
+
+    flake.variants = {
+      ghc92 = { }; # Alias for the default project
+      ghc92-profiled.modules = [{ enableProfiling = true; }];
```
michaelpj, line 25: Is this enough? I have a vague feeling that we also need `enableLibraryProfiling`?

nix/project.nix
```diff
@@ -0,0 +1,160 @@
...
+        };
+      })
+    ];
+  };
+
+
+  project = lib.iogx.mkHaskellProject {
+    inherit cabalProjectArgs;
+    readTheDocs.siteFolder = "doc/read-the-docs-site";
+    shellArgsForProjectVariant = repoRoot.nix.shell;
```
michaelpj, line 142: Why "forProjectVariant"? That doesn't really mean anything to me here, isn't all the stuff above already per-variant?

michaelpj: It's kind of confusing that there are two sets of "shell args", one here and one above. I guess this is my fault for wanting the cabalProjectArgs to match the haskell.nix ones, though :)

nix/shell.nix
```diff
@@ -0,0 +1,63 @@
+{ repoRoot, pkgs, ... }:
+
+cabalProject:
+
+{
```
michaelpj, line 5: so... how do I change a tool version? Just override the derivation entirely?

nix/shell.nix
```diff
@@ -0,0 +1,63 @@
+{ repoRoot, pkgs, ... }:
+
+cabalProject:
```
michaelpj, line 3: dead? what is this for?

nix/project.nix
```diff
@@ -0,0 +1,160 @@
...
+          # disabling -Werror on it.
+          prettyprinter-configurable.ghcOptions = [ "-Werror" ];
+        };
+      })
+    ];
+  };
+
+
+  project = lib.iogx.mkHaskellProject {
+    inherit cabalProjectArgs;
```
michaelpj, line 140: So, I see that in iogx you literally just call `cabalProject'` on this. So I wonder whether it wouldn't be simpler to move that into the user side, and just have the user pass in the output of `cabalProject'`.

nix/cabal-project.nix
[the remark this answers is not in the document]
michaelpj: > plutus and plutus-apps use the combined-haddock to build the RTD site, so the RTD site needs to reference the haskell project somehow, since it also needs to add some scripts to the shell.
> Separating the RTD into its own lib.iogx.mkReadTheDocsSite means more work for end user.

IDK, I guess I just don't like having a big "do everything" function, but I guess that's kind of the point. How do I fix anything in the RTD site if it's not working properly?

nix/project.nix
```diff
@@ -0,0 +1,160 @@
...
+    readTheDocs.siteFolder = "doc/read-the-docs-site";
```
michaelpj, line 141: Shouldn't there be an "enable" option here also?

nix/cabal-project.nix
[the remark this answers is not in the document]
michaelpj: e.g. it looks like you've just taken a frozen snapshot of the python packages we happen to currently be using for Sphinx, but we might want to add more - there's 0 guarantee that that's the final set. So what would I do then?

nix/outputs.nix
michaelpj: Feels vaguely unsatisfying but I guess so

michaelpj (approved):
TBH I still prefer the more explicit approaches but I'm happy for you to go from here and iterate.

flake.nix
zeme-wana: These are parts of `iogx` inputs:

```nix
    haskell-nix = {
      url = "github:input-output-hk/haskell.nix/445086f35d450af485cd9200b86211355095b2a1";
      inputs.hackage.follows = "hackage";
    };

    nixpkgs.follows = "haskell-nix/nixpkgs-2305";

    hackage = {
      url = "github:input-output-hk/hackage.nix";
      flake = false;
    };
```

If one uses iogx one must use the `nixpkgs` provided by it, since it will have the necessary overlays, etc.. 

When pinning `haskell-nix` in your flake, nix complained and so I had to add the line `inputs.nixpkgs.follows = "haskell-nix/nixpkgs-2305";`

I guess if one must use another `nixpkgs` one can always add say `nixpkgs-2` to their inputs.

nix/project.nix
zeme-wana: haskell.nix complains if I don't do that. I'll add a comment

nix/project.nix
zeme-wana: Yes I agree, one might as well pass the project at this point.

nix/shell.nix
zeme-wana: The shell is called for each project variant.
Many repositories pull exes from `cabalProject` into shell.
`plutus` doesn't do it, so I guess we can rename `cabalProject` -> `_cabalProject` to make it more clear.

nix/shell.nix
zeme-wana: Yes, you'll have to provide your own derivation.

nix/project.nix
zeme-wana: > Why "forProjectVariant"?

Yeah it's a little too verbose. `shellArgs` is good enough.

zeme-wana:
> How do I fix anything in the RTD site if it's not working properly?

You have to rely on iogx for the RTD site derivation. If it breaks then it has to be patched in iogx. I'm not sure what an "escape hatch" for this would look like, short of providing the `read-the-docs-site` derivation yourself, which kind of defeats the purpose.

> it looks like you've just taken a frozen snapshot of the python packages we happen to currently be using for Sphinx, but we might want to add more - there's 0 guarantee that that's the final set. So what would I do then?

I can add a `shpinxToolchain` option to the `readTheDocs` config.
That's all that's needed to build the site. Although there is the issue with `plantum`:
```nix
  site = pkgs.stdenv.mkDerivation {

    name = "read-the-docs-site";

    buildInputs = [
      repoRoot.src.ext.sphinx-toolchain
      # We need this here in order to get the `plantuml` executable in PATH.
      # Unfortunately `python3.withPackages` (used by sphinx-toolchain above)
      # won't do it automatically.
      pkgs.python3Packages.sphinxcontrib_plantuml
    ];
  };
```

Maybe we can get `sphinxcontrib_plantuml` from `sphinx-toolchain` somehow. I'll look into that

nix/project.nix
zeme-wana: Good catch, thanks.

zeme-wana:
What do we do about hlint?

- With this PR, `hlint` would be run on each `git commit`, and also in CI as part of `pre-commit-check`, using the top-level `.hlint.yml` file, and using the "right" version of the `hlint` program. But this won't show the hints in the PR.
- The GH `hlint` workflow uses version 2.13.3. We would have to remember to keep that up-to-date. But again, it uses `./.github/.hlint.yaml`, which runs no hints.

We can either:
- Tell the GH `hlint` to use the top-level `.hlint.yml`
- Tell the GH `hlint` to also use *our* `hlint`. But this means that the workflow has to enter the nix shell, which means downloading 10-30 minute of stuff every time.
- Remove the GH `hlint` altogether and rely on `pre-commit-check`

flake.nix
michaelpj: > If one uses iogx one must use the nixpkgs provided by it, since it will have the necessary overlays, etc..

I don't see how this affects the choice of flake input?

My issue is just that we were previously using `haskell-nix/nixpkgs` and you changed it to `haskell-nix/nixpkgs-2305`.

nix/shell.nix
michaelpj: If I want to change the version of a formatter, does that mean I'll also need to provide a HLS derivation that uses that version correctly?

michaelpj: This is just making me want to immediately use our own derivations so this is easy to change later...

michaelpj: Also... doesn't this mean that my version of iogx determines the formatter version? so updating iogx can cause me to have haskell formatting changes, which as I said before seems undesirable. Maybe other people don't care :shrug:

michaelpj:
I'm not sure. I'm happy to not have the GH Hlint workflow, I don't find it that valuable. I think @zliu41 added it, what do you think?

nix/shell.nix
zeme-wana: iogx provides one toolchain per ghc, which includes hls, hlint, etc.
You can override any of those.
The version of iogx does not determine the formatter version, the GHC version does.

flake.nix
zeme-wana: Got it, will change to `haskell-nix/nixpkgs`

nix/shell.nix
michaelpj: >  The version of iogx does not determine the formatter version, the GHC version does.

So you promise to never ever change the version of a tool that's part of the toolchain for a given GHC? that seems unlikely - there will be bugs and you'll want to not recommend a broken version.

PR #5473, opened 2023-08-11
Kwxm/conformance/hash functions

kwxm:
PLT-6308 and 6340.  This adds conformance tests for the new hash functions.  I've taken the inputs either from appropriate standards or from reference implementations (documented in the tests).  While I was at it I took the opportunity to update the tests for the existing hash functions with data from similar sources.  There are now multiple unit tests for the same bultins in `untyped-plutus-core`, but I thought that more tests wouldn't do any harm.


kwxm:
@michaelpj I was going to ask someone else to review it, but I suppose you'll do it anyway!

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
```diff
@@ -638,6 +638,63 @@ test_Crypto = testCase "Crypto" $ do
     -- hex output: 47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad
     evals @ByteString "G\ETB2\133\168\215\&4\RS^\151/\198w(c\132\248\STX\248\239B\165\236_\ETX\187\250%L\176\US\173"
         Keccak_256 [cons @ByteString "hello world"]
+    -- Tests for blake2b_224: output obtained using the b2sum program from https://github.com/BLAKE2/BLAKE2
```
michaelpj, line 641: Is there much point having these if we have the conformance tests?

michaelpj (approved):

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
kwxm: > Is there much point having these if we have the conformance tests?

I did wonder about that, then thought that it wouldn't do any harm.  Maybe we should have a "tidy up the tests" ticket to sort that out more generally if we think it's worth doing.

PR #5476, opened 2023-08-11
Add strictifying bindings optimization

michaelpj:
This adds an optimization pass that tries to make non-strict bindings strict, if their RHSs are pure.

By itself, this is a fine but not very relevant optimization. But the reason it does not do much is because the Plutus Tx plugin is already doing similar logic! This is putting the responsibility in the wrong place: we can do this generically in PIR, which simplifies the plugin and would also benefit anyone else targeting PIR.

The test differences are just in non-simplified programs.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-core/plutus-ir/src/PlutusIR/Compiler.hs
```diff
@@ -116,6 +118,11 @@ availablePasses =
     , Pass "case reduce"          (onOption coDoSimplifierCaseReduce)         (pure . CaseReduce.caseReduce)
     , Pass "known constructor"    (onOption coDoSimplifierKnownCon)           KnownCon.knownCon
     , Pass "beta"                 (onOption coDoSimplifierBeta)               (pure . Beta.beta)
+    , Pass "strictify bindings"   (onOption coDoStrictifyBindings)            (\t ->
```
bezirg, line 121: Can you rename `coDoStrictifyBindings` to `coDoSimplifierStrictify` or sth? The rest seem to have uniformly the `coDoSimplifier...` prefix

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -343,11 +343,11 @@ In Haskell this is conditionally error, in PLC it is unconditionally error.
...
 These sorts of thing can be written by the user, or generated by GHC.
 
-We solve this by compiling such let-bindings *non-strictly*. That means we delay the body
-and force all uses of it. This is a bit overenthusiastic, as it will add overhead to
-effect-free let-bindings, but I don't know of an easy way to tell if an expression is pure.
+We solve this by compiling let-bindings *non-strictly*. That means we delay the body
+and force all uses of it.
 
 Conveniently, we can just use PIR's support for non-strict let bindings to implement this.
+The PIR optimizer will also strictify any such bindings that turn out to be pure.
```
bezirg, line 350: I don't know if we use the term "PIR optimizer" anywhere else, so the term seems a bit confusing. In any case, it would be good to say that:  "The PIR optimizer (which is enabled by default) ..."

bezirg (approved):

PR #5477, opened 2023-08-15
Kwxm/spec/cons byte string change (PLT-629)

kwxm:
This started out as an attempt to remedy the fact that we'd forgotten to specify a change in the behaviour of `consByteString` between V1 and V2 (see PLT-629 and issue #5475).  However I then had to talk about language versions and ledger language versions (see [CIP-35](https://cips.cardano.org/cips/cip35/) and [this document](https://github.com/input-output-hk/plutus/blob/master/doc/read-the-docs-site/explanations/language-versions.rst#id4)) and things got a bit out of hand, necessitating some reorganisation.  It's not easy to explain this version stuff. :(

Here's a rendered version: [plutus-core-specification.pdf](https://github.com/input-output-hk/plutus/files/12375932/plutus-core-specification.pdf).


doc/plutus-core-spec/flat-serialisation.tex
```diff
@@ -412,6 +412,10 @@ \subsubsection{Terms}
   \Dterm(\bits{1001} \cdot s)  &= (s', \Kase{u}{l}) &&\quad \text{if } \Dterm(s) = (s', u) &&\ \text{and}\ \Dlist_{\mathsf{term}}(s') = (s'', l)
 \end{alignat*}
 
+\paragraph{NOTE.} The decoder $\Dterm$ should fail if we are decoding a program
```
kwxm, line 415: Strictly I should have term decoders indexed by version numbers, but I was lazy and decided to just fudge it.

doc/plutus-core-spec/introduction.tex
```diff
@@ -18,3 +18,51 @@ \section{Introduction}
...
+  consists of three non-negative integers separated by decimal points, for
+  example \texttt{1.4.2}.  Language versions are ordered lexicographically.
+\item The built-in types and functions deployed on the chain along with a
+  particular language version can change with time, and we refer to Plutus Core
+  along with a particular set of types and functions as a \textit{Plutus Core
+    ledger language version}.  These are referred to using version names like
+  ``Plutus V1,'', ``Plutus V2'', and so on.
+  \end{itemize}
+ \noindent These two kinds of version are independent: all Plutus Core language
+ versions must support all ledger language versions.
```
kwxm, line 43: I think this isn't exactly true.  For example, the V1 cost parameters (as stored in the protocol parameters) don't contain a cost for `cekCaseCost`, so you can't use V1 with PLC 1.1.0.

kwxm: ^ I now think that the previous comment is wrong.  The `defaultCostModel` for 1.1.0 contains entries for `cekCaseCost` and `cekConstrCost`, and if we use [applyCostModelParams](https://github.com/input-output-hk/plutus/blob/b321575d9266b3358b9e728d064fc0bee4f355d7/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs#L273) to update that with an old set of V1 parameters `P` that don't contain those, there won't be an error: the entries for the original constructors like `cekApplyCost` will be overwritten with the corresponding entries from `P`, but the entries for `cekCaseCost` and `cekConstrCost` will be left untouched: see [this test](https://github.com/input-output-hk/plutus/blob/b321575d9266b3358b9e728d064fc0bee4f355d7/plutus-core/plutus-core/test/CostModelInterface/Spec.hs#L128) for example.

doc/plutus-core-spec/V2-builtins.tex
```diff
@@ -43,18 +42,31 @@ \subsection{Built-in functions}
...
+\hline 
 \end{longtable}
 
 
+\note{Changed semantics for \texttt{consByteString}.}
+\label{note:consbytestring-v2}
+In Plutus V1 the first argument of \texttt{consByteString} is an arbitrary
+integer which will be reduced modulo 256 before being prepended to the second
+argument.  In Plutus V2 we require that the first argument lies between 0 and 255
+(inclusive): an error will occur if this is not the case.
```
bezirg, line 68: Suggestion: in any other case the builtin returns the error value ×

doc/plutus-core-spec/flat-serialisation.tex
bezirg: Seems fine for now

doc/plutus-core-spec/introduction.tex
```diff
@@ -18,3 +18,51 @@ \section{Introduction}
...
+re-evaluated at any point in the future when the history of the chain is
+verified.  Plutus Core may evolve as time progresses, and because of this we
+have to be quite careful with versioning.  A detailed discussion of this issue
+can be found in~\cite{CIP-35}, but we reproduce two definitions here for
+convenience.
+\begin{itemize}
+\item The language itself may evolve: for example some fundamental new construct
+  may be added.  A \textit{Plutus Core language version} describes a version of
+  the basic language with a particular set of features. A language version
+  consists of three non-negative integers separated by decimal points, for
```
bezirg, line 34: Suggestion: "non-negative integers" => "natural numbers"

bezirg, line 34: Suggestion: "decimal points" => dots (.)

doc/plutus-core-spec/introduction.tex
bezirg: I am also not sure anymore. Is it going to be possible to run V1 together with plc-1.1.0? @michaelpj

doc/plutus-core-spec/introduction.tex
```diff
@@ -18,3 +18,51 @@ \section{Introduction}
...
+    ledger language version}.  These are referred to using version names like
+  ``Plutus V1,'', ``Plutus V2'', and so on.
+  \end{itemize}
+ \noindent These two kinds of version are independent: all Plutus Core language
+ versions must support all ledger language versions.
+
+\subsection{Cardano support for different Plutus Core language and ledger language versions}
+
+Cardano introduced support for Plutus Core in the Alonzo release, and
+Table~\ref{table:versions} shows later releases which have introduced new
```
bezirg, line 48: Suggestion: "shows later releases" => "shows all releases that have introduced new language versions and/or new ledger language versions"

doc/plutus-core-spec/untyped-grammar.tex
```diff
@@ -86,16 +86,29 @@ \subsection{Grammar}
...
+the Plutus Core language version (Section~\ref{sec:version-numbers}) used in the
+program.  The grammar above describes Plutus Core Version 1.1.0, which will be
+deployed in a forthcoming hard fork.  Version 1.0.0 (available on Cardano since
+the Alonzo hard fork) is identical, except that \texttt{constr} and
+\texttt{case} are not included.  Version 1.0.0 is fully forward-compatible with
+version 1.1.0, so any valid version 1.0.0 program is also a valid version 1.1.0
+program.  The semantics, evaluator and serialisation formats described later in
+this document all apply to both versions, except that it is an error to use
+\texttt{constr} or \texttt{case} in any program with a version prior to 1.1.0: a
+parser, deserialiser, or evaluator should fail immediately if \texttt{constr} or
```
bezirg, line 99: Suggestion ", or evaluator" => ", and evaluator"

bezirg (approved):

doc/plutus-core-spec/introduction.tex
kwxm: I'll revise that when I work out what the truth is.

effectfully (approved):

doc/plutus-core-spec/V1-builtins.tex
```diff
@@ -10,16 +10,16 @@
 \newcommand{\vk}{\textit{vk}}  %% Verification key (ie public key) for signature verification functions.
 \newpage
 
-\section{Built-in Types and Functions Supported in the Alonzo Release}
-\label{appendix:default-builtins-alonzo}
+\section{Plutus V1 Built-in Types and Functions}
+\label{appendix:default-builtins-V1}
```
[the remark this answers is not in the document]
kwxm: Yes!  I'll fix that.

doc/plutus-core-spec/introduction.tex
kwxm: > Suggestion: "non-negative integers" => "natural numbers"

Some people think that 0 isn't a natural number, although they're probably not computer scientists.  We do say that ℕ includes zero earlier on though, so maybe that's good enough.

bezirg:
Oh yes, the semantics of consByteString is going to take effect in PlutusV3 (which uses DefaultFunV2)

kwxm:
I've now added a discussion of PLC language versions (which we need to talk about the addition of `constr` and `case) and ledger language versions.  I've changed the appdendices that talk about builtins to use ledger language versions instead of Cardano releases because that seems more accurate.

Here's an updated version of the PDF.

[plutus-core-specification.pdf](https://github.com/input-output-hk/plutus/files/12402547/plutus-core-specification.pdf)

kwxm:
We still need to talk about de Bruijn indices(eg, for evaluation) (SCP-630) and concrete syntax for `data` and built-in polymorphic types (PLT-378).

doc/plutus-core-spec/versions.tex
```diff
@@ -0,0 +1,72 @@
...
+  example \texttt{1.4.2}.  Language versions are ordered lexicographically.  As
+  we have seen in Section~\ref{sec:grammar-notes}, the Plutus Core
+  language version used by a script is included in the \texttt{program} header.
+  
+\item The built-in types and functions deployed on the chain along with a
+  particular version of the Plutus Core language can change with time.  Plutus
+  Core scripts on the Cardano chain are equipped with a tag called a
+  \textit{ledger language version} which (among other things) determines which
+  set of types and functions are to be used during evaluation of the script.
+  Ledger language versions have names names like ``Plutus V1'', ``Plutus V2'',
```
michaelpj, line 24: Here and throughout: I strongly suggest using "PlutusV1" over "Plutus V1". The latter suggests that there are two components, and one of them is a "version". This is not true: "PlutusV1" is just an uninterpreted string, that we happen to have given a descriptive name. Or to put it another way, the ledger language _is_ the string "PlutusV1" not "Plutus V1".

doc/plutus-core-spec/versions.tex
```diff
@@ -0,0 +1,72 @@
...
+  and so on.  The semantics of a given built-in function may be different in
+  different ledger language versions.  Each Plutus Core ledger language version
+  is only valid with a specific Plutus Core language version, but a given
+  language version may support multiple ledger language versions.
```
michaelpj, line 28: I think this is a bit confusing. Perhaps just say "the ledger language version determines the semantics of the built-in functions and the allowable Plutus Core language versions"?

doc/plutus-core-spec/versions.tex
```diff
@@ -0,0 +1,72 @@
...
+Table~\ref{table:versions} shows all releases that have introduced new language
+versions and/or new ledger language versions since then.  The features provided
+by the different ledger langauge versions are described in subsequent appendices.
+\begin{table}[H]
+  \centering
+    \begin{tabular}{|l|l|l|l|}
+        \hline
+        Cardano release & Date & Language versions & Ledger language versions \\
+        \hline
+        Alonzo & September 2021 & 1.0.0 & V1 \\
```
michaelpj, line 43: There is no such thing as "V1". There is only "PlutusV1"

doc/plutus-core-spec/versions.tex
```diff
@@ -0,0 +1,72 @@
...
+
+Cardano introduced support for Plutus Core in the Alonzo release;
+Table~\ref{table:versions} shows all releases that have introduced new language
+versions and/or new ledger language versions since then.  The features provided
+by the different ledger langauge versions are described in subsequent appendices.
+\begin{table}[H]
+  \centering
+    \begin{tabular}{|l|l|l|l|}
+        \hline
+        Cardano release & Date & Language versions & Ledger language versions \\
```
michaelpj, line 41: Better to use protocol versions, I think. You can see which ones correspond to what in `plutus-ledger-api`

doc/plutus-core-spec/versions.tex
```diff
@@ -0,0 +1,72 @@
...
+attempting to evaluate it will cause an error.
+ 
+%% See PlutusLedgerApi.Common.Versions.plcVersionsIntroducedIn
+\begin{table}[H]
+  \centering
+    \begin{tabular}{|l|l|}
+        \hline
+        \thead{Plutus Core \\ language version} & \thead{Compatible ledger \\ language versions}\\
+        \hline
+        1.0.0 & V1, V2\\
```
michaelpj, line 65: I think this map should go the other way around: it's more "given the ledger langauge, what plutus core langauge versions are permissible"

michaelpj (approved):
I think this is fine. I think we could organize it more clearly in the longer term, maybe with a whole section for "Plutus Core on Cardano" that lays out all the Cardano-specific stuff (including the builtins!).

doc/plutus-core-spec/versions.tex
kwxm: > Better to use protocol versions, I think.

:( I was hoping to get away without mentioning protocol versions, but I'll have a go.

kwxm: > Better to use protocol versions, I think. You can see which ones correspond to what in plutus-ledger-api

Apparently not. There's no mention of protocol version 6.0 [here](https://github.com/input-output-hk/plutus/blob/master/plutus-ledger-api/src/PlutusLedgerApi/Common/ProtocolVersions.hs) for example.

doc/plutus-core-spec/versions.tex
michaelpj: Yep, you've got me there!

PR #5478, opened 2023-08-16
Reduce gh-pages size when uploading Haddock

zeme-wana:
The git history for the gh-pages branch makes the repo quite heavy.
This PR sets the `single-commit` flag of the github-pages-deploy action to true:
See here: https://github.com/JamesIves/github-pages-deploy-action#optional-choices

> This option can be toggled to true if you'd prefer to have a single commit on the deployment branch instead of maintaining the full history. Using this option will also cause any existing history to be wiped from the deployment branch.

### DIFF
--- a/.github/workflows/haddock.yml
+++ b/.github/workflows/haddock.yml
@@ -28,4 +28,6 @@ jobs:
         with:
           folder: dist
           target-folder: ${{ github.ref_name }}
-          # single-commit: true
+          # We publish our haddock, which is non-trivially big. 
+          # So keeping the whole history is expensive, and anyway we don't need it.
+          single-commit: true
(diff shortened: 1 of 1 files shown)


michaelpj:
How does this work? Comments?

.github/workflows/haddock.yml
```diff
@@ -28,3 +28,4 @@ jobs:
         with:
           folder: dist
           target-folder: ${{ github.ref_name }}
+          # single-commit: true
```
zeme-wana, line 31: Although workflow changes do not take effect until pushed to master, there are cases when it does.
If this PR is approved I'll uncomment this line

michaelpj:
```suggestion
          -- We publish our haddock, which is non-trivially big. So keeping the whole history
          -- is expensive, and anyway we don't need it.
          # single-commit: true
```

michaelpj (approved):

PR #5482, opened 2023-08-18
Dump full compilation trace for the plugin (PLT-269)

zliu41:
When the compilation fails it can dump a stack of things leading to the failure point. This is not sufficient for two reasons:

1. In most of my debugging sessions I find myself wanting to look at the traces of both the failed compilation and a succeeded compilation, and compare the two to figure out where it went wrong. Looking at the trace of the failed compilation alone is insufficient.
2. The stack can't be printed if the compiler loops infinitely.

This PR adds a compiler flag for dumping the full trace, that is, every step taken by `compileExpr`, `hoistExpr`, `compileType` and so on, as the compilation proceeds. To make it easier to figure out how the compilation got to a certain point, every step is annotated by an ID along with the ID of the parent step.

An example trace dump:

```
...
<Step 415, Parent Step: 414>
Compiling expr: PlutusTx.Foldable.sum
                  @[]
                  @GHC.Num.Integer.Integer
                  PlutusTx.Foldable.$fFoldable[]
                  PlutusTx.Numeric.$fAdditiveMonoidInteger
<Step 416, Parent Step: 415>
Compiling expr: PlutusTx.Foldable.sum
                  @[] @GHC.Num.Integer.Integer PlutusTx.Foldable.$fFoldable[]
<Step 417, Parent Step: 416>
Compiling expr: PlutusTx.Foldable.sum @[] @GHC.Num.Integer.Integer
<Step 418, Parent Step: 417>
Compiling expr: PlutusTx.Foldable.sum @[]
<Step 419, Parent Step: 418>
Compiling expr: PlutusTx.Foldable.sum
<Step 420, Parent Step: 419>
Compiling definition of: PlutusTx.Foldable.sum
<Step 421, Parent Step: 420>
Compiling type: forall (t :: * -> *) a.
                (PlutusTx.Foldable.Foldable t,
                 PlutusTx.Numeric.AdditiveMonoid a) =>
                t a -> a
<Step 422, Parent Step: 421>
Compiling kind: * -> *
<Step 423, Parent Step: 422>
Compiling kind: *
<Completed step 423, Returning to step 422>
<Step 424, Parent Step: 423>
Compiling kind: *
<Completed step 424, Returning to step 423>
<Completed step 422, Returning to step 421>
...
```

### DIFF
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Kind.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Kind.hs
@@ -17,7 +17,7 @@ import GHC.Plugins qualified as GHC
 import PlutusCore qualified as PLC
 
 compileKind :: Compiling uni fun m ann => GHC.Kind -> m (PLC.Kind ())
-compileKind k = withContextM 2 (sdToTxt msg) . traceCompilation msg $ case k of
+compileKind k = traceCompilation 2 ("Compiling kind:" GHC.<+> GHC.ppr k) $ case k of
     -- this is a bit weird because GHC uses 'Type' to represent kinds, so '* -> *' is a 'TyFun'
     (GHC.isLiftedTypeKind -> True)         -> pure $ PLC.Type ()
     (GHC.splitFunTy_maybe -> Just r) -> case r of
@@ -36,5 +36,3 @@ compileKind k = withContextM 2 (sdToTxt msg) . traceCompilation msg $ case k of
     (GHC.classifiesTypeWithValues -> True) -> pure $ PLC.Type ()
 #endif
     _                                      -> throwSd UnsupportedError $ "Kind:" GHC.<+> (GHC.ppr k)
-  where
-    msg = "Compiling kind:" GHC.<+> GHC.ppr k
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Kind.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Kind.hs
@@ -8,6 +8,7 @@
 module PlutusTx.Compiler.Kind (compileKind) where
 
 import PlutusTx.Compiler.Error
+import PlutusTx.Compiler.Trace
 import PlutusTx.Compiler.Types
 import PlutusTx.Compiler.Utils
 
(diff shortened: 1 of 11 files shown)


plutus-tx-plugin/src/PlutusTx/Compiler/Trace.hs
```diff
@@ -0,0 +1,41 @@
...
+  GHC.SDoc ->
+  -- | The compilation action
+  m a ->
+  m a
+traceCompilation sd compile = ifM (notM (asks ccDebugTraceOn)) compile $ do
+  CompileState nextStep prevSteps <- get
+  put $ CompileState (nextStep + 1) (nextStep : prevSteps)
+  let mbParentStep = listToMaybe prevSteps
+  s <- sdToStr sd
+  traceM $
```
michaelpj, line 28: This stuff starts to make me wonder if we should just be emitting logs in some more standard format that includes this kind of information, like https://hackage.haskell.org/package/opentelemetry-0.8.0 or something.

plutus-tx-plugin/src/PlutusTx/Compiler/Kind.hs
```diff
@@ -16,7 +17,7 @@ import GHC.Plugins qualified as GHC
 import PlutusCore qualified as PLC
 
 compileKind :: Compiling uni fun m ann => GHC.Kind -> m (PLC.Kind ())
-compileKind k = withContextM 2 (sdToTxt $ "Compiling kind:" GHC.<+> GHC.ppr k) $ case k of
+compileKind k = withContextM 2 (sdToTxt msg) . traceCompilation msg $ case k of
```
michaelpj, line 20: The tracking of the context here and the tracking of the parent steps seems somewhat redundant and could probably be merged?

michaelpj (approved):
Looks fine.

plutus-tx-plugin/src/PlutusTx/Compiler/Trace.hs
zliu41: Perhaps for this too, but I'd love to have some observability on the scripts being evaluated on mainnet.

PR #5484, opened 2023-08-21
Plutus Tx: avoid duplicating code for DEFAULT case bodies

michaelpj:
Turns out we were duplicating code for a slightly subtle reason. GHC can have just a single `DEFAULT` case that handles a ll the non-explicitly-handled cases. But we need to make all the cases explicit, which means that we were duplicating the `DEFAULT` case for all the non-handled cases. And of course, the moment we see "duplicating code" it's a short hop to "exponential code", and indeed that's what I was seeing for code that has many nested default cases (which it turns out you get from using pattern synonyms, apparently).

The fix is simple: just share the code.


zliu41 (commented):
There's a dramatic decrease in `formula1 (size)` but a 7.8% _increase_ in `formulaBudget`. Can you look into it?

michaelpj:
I took a look at the PIR diff. It's not at all obvious what's worse: the changes are exactly what we'd expect (there was a _lot_ of duplicated code from this, in fact). My guess is that
- The formula PIR has a lot of default cases
- The default cases are dynamically dead, i.e. we don't actually hit them
- Since there are many default cases, the inliner doesn't inline the default body binding
- So we now pay extra for the binding and don't benefit

The inlining decisions seem correct to me, since we would pay for it with a substantial increase in program size (as we can see). So I think we basically just get the opportunity to do a cost/size tradeoff differently, and we choose to not increase size, which seems fine.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -1047,3 +1058,52 @@ compileExprWithDefs e = do
...
+The solution is straightforward: share the default case. That means we produce a program more
+like:
+```
+f a = let defaultBody = 2 in case a of
+  B -> 1
+  C -> defaultBody
+  D -> defaultBody
+```
+Then the inliner can inline it as appropriate.
+-}
```
bezirg, line 1109: Great note

PR #5486, opened 2023-08-21
[Builtins] Add support for pattern matching builtins

effectfully:
The main change is replacing

```haskell
data BuiltinRuntime val
    = BuiltinCostedResult ExBudgetStream ~(BuiltinResult val)
    | <...>
```

with

```haskell
data BuiltinRuntime val
    = BuiltinCostedResult ExBudgetStream ~(BuiltinResult (HeadSpine val))
    | <...>
```

where `HeadSpine` is a fancy way of saying `NonEmpty`:

```haskell
-- | A non-empty spine. Isomorphic to 'NonEmpty', except is strict and is defined as a single
-- recursive data type.
data Spine a
    = SpineLast a
    | SpineCons a (Spine a)

-- | The head-spine form of an iterated application. Provides O(1) access to the head of the
-- application. Isomorphic to @NonEmpty@, except is strict and the no-spine case is made a separate
-- constructor for performance reasons (it only takes a single pattern match to access the head when
-- there's no spine this way, while otherwise we'd also need to match on the spine to ensure that
-- it's empty -- and the no-spine case is by far the most common one, hence we want to optimize it).
data HeadSpine a
    = HeadOnly a
    | HeadSpine a (Spine a)
```

(we define a separate type, because we want strictness, and you don't see any bangs, because it's in a module with `StrictData` enabled).

The idea is that a builtin application can return a function applied to a bunch of arguments, which is exactly what we need to be able to express `caseList`

```haskell
caseList xs0 f z = case xs0 of
   []   -> z
   x:xs -> f x xs
```

as a builtin:

```haskell
-- | Take a function and a list of arguments and apply the former to the latter.
headSpine :: Opaque val asToB -> [val] -> Opaque (HeadSpine val) b
headSpine (Opaque f) = Opaque . \case
    []      -> HeadOnly f
    x0 : xs ->
        -- It's critical to use 'foldr' here, so that deforestation kicks in.
        -- See Note [Definition of foldl'] in "GHC.List" and related Notes around for an explanation
        -- of the trick.
        HeadSpine f $ foldr (\x2 r x1 -> SpineCons x1 $ r x2) SpineLast xs x0

instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
    <...>
    toBuiltinMeaning _ver CaseList =
        let caseListDenotation
                :: Opaque val (LastArg a b)
                -> Opaque val (a -> [a] -> b)
                -> SomeConstant uni [a]
                -> BuiltinResult (Opaque (HeadSpine val) b)
            caseListDenotation z f (SomeConstant (Some (ValueOf uniListA xs0))) = do
                case uniListA of
                    DefaultUniList uniA -> pure $ case xs0 of
                        []     -> headSpine z []                                             -- [1]
                        x : xs -> headSpine f [fromValueOf uniA x, fromValueOf uniListA xs]  -- [2]
                    _ ->
                        -- See Note [Structural vs operational errors within builtins].
                        throwing _StructuralUnliftingError "Expected a list but got something else"
            {-# INLINE caseListDenotation #-}
        in makeBuiltinMeaning
            caseListDenotation
            (runCostingFunThreeArguments . unimplementedCostingFun)
```

Being able to express [1] (representing `z`) and [2] (representing `f x xs`) is precisely what this PR enables.

Adding support for the new functionality to the CEK machine is trivial. All we need is a way to push a `Spine` of arguments onto the context:

```haskell
    -- | Push arguments onto the stack. The first argument will be the most recent entry.
    pushArgs
        :: Spine (CekValue uni fun ann)
        -> Context uni fun ann
        -> Context uni fun ann
    pushArgs args ctx = foldr FrameAwaitFunValue ctx args
```

and a `HeadSpine` version of `returnCek`:

```haskell
    -- | Evaluate a 'HeadSpine' by pushing the arguments (if any) onto the stack and proceeding with
    -- the returning phase of the CEK machine.
    returnCekHeadSpine
        :: Context uni fun ann
        -> HeadSpine (CekValue uni fun ann)
        -> CekM uni fun s (Term NamedDeBruijn uni fun ())
    returnCekHeadSpine ctx (HeadOnly  x)    = returnCek ctx x
    returnCekHeadSpine ctx (HeadSpine f xs) = returnCek (pushArgs xs ctx) f
```

Then replacing

```haskell
                BuiltinSuccess x ->
                    returnCek ctx x
```

with

```haskell
                BuiltinSuccess fXs ->
                    returnCekHeadSpine ctx fXs
```

(and similarly for `BuiltinSuccessWithLogs`) will do the trick.

We used to define `caseList` in terms of `IfThenElse`, `NullList` and either `HeadList` or `TailList` depending on the result of `NullList`, i.e. three builtin calls in the worst and in the best case. Then we introduced `ChooseList`, which replaced both `IfThenElse` and `NullList` in `caseList` thus bringing total amount of builtin calls down to 2 in all cases. This turned out to have a [substantial](https://github.com/input-output-hk/plutus/pull/4119#pullrequestreview-781441048) impact on performance. This PR allows us to bring total number of builtin calls per `caseList` invokation down to 1 -- the `CaseList` builtin itself.


effectfully:
> I'm a bit confused about what this would entail for the PLC language.
> Would there be a new builtin-function? would case be a new language construct?

No new language constructs. New built-in functions (`caseList` in the description of the PR, plus `caseData` in the PR), plus a bit of changes to the builtins machinery and the CEK machine (see in the PR, it's a surprisingly small amount of changes).

mjaskelioff:
I saw the PR, but it's about code and I don't fully understand the consequences of the code for PLC. What would be the type of `caseList`?

plutus-core/plutus-core/test/Pretty/Golden/Readable/StdLib/Data/List/FoldrList.golden
```diff
@@ -25,15 +25,15 @@
...
                 \(z : r) (f : a -> list a -> r) ->
-                  chooseList
+                  caseList
                     {a}
                     {unit -> r}
-                    x
+                    xs
                     (\(u : unit) -> z)
-                    (\(u : unit) -> f (headList {a} x) (tailList {a} x))
+                    (\(x : a) (xs' : list a) (u : unit) -> f x xs')
```
mjaskelioff, line 36: If i understand correctly, we would need to allow higher-order builtins?

mjaskelioff: That is, is `caseList : all a b. list a -> (unit -> b) -> (a -> list a -> unit -> b)` ?
Why is the unit in the last argument needed?

plutus-core/plutus-core/test/Pretty/Golden/Readable/StdLib/Data/Data/caseData.golden
```diff
@@ -5,20 +5,12 @@
...
-             (unConstrData d))
-        (\(u : unit) -> fMap (unMapData d))
-        (\(u : unit) -> fList (unListData d))
-        (\(u : unit) -> fI (unIData d))
-        (\(u : unit) -> fB (unBData d))
+        (\(i : integer) (ds : list data) (u : unit) -> fConstr i ds)
+        (\(es : list (pair data data)) (u : unit) -> fMap es)
+        (\(ds : list data) (u : unit) -> fList ds)
+        (\(i : integer) (u : unit) -> fI i)
+        (\(b : bytestring) (u : unit) -> fB b)
```
mjaskelioff, line 15: I don't understand why the `unit`s are needed if computation is already under a function type.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1256,6 +1262,22 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
                 []    -> a
                 _ : _ -> b
           {-# INLINE choosePlc #-}
+    toBuiltinMeaning _ver CaseList =
+        makeBuiltinMeaning
+            caseListPlc
+            (\_ _ _ _ -> ExBudgetLast mempty)  -- TODO.
```
mjaskelioff, line 1268: I was wondering what happens to the cost...

effectfully: We always do a constant amount of work, so should be constant and not very hard to measure. One part that may be tricky is distinguishing (costing-wise) between returning a function application from a builtin and evaluating that application in the CEK machine. But we probably can adapt the current Kenneth's trick of using dedicated no-op builtins. Anyways, we should only discuss it if we think pattern matching builtins make sense, otherwise we'll just waste a bunch of time. I do believe costing shouldn't be an issue, but I might be wrong about that.

plutus-core/plutus-core/test/Pretty/Golden/Readable/StdLib/Data/Data/caseData.golden
effectfully: Right, I think I just mechanically updated the previous definition which was in terms of `chooseData` where `unit`s were needed without realizing that they're no longer needed. Thank you, will fix.

effectfully:
> What would be the type of `caseList`?

It's in the PR, the `plutus-core/plutus-core/test/TypeSynthesis/Golden/CaseList.plc.golden` file (there's a golden file with a type signature for every builtin, including the proposed ones):

```
(all
  a
  (type)
  (all
    b
    (type)
    (fun [ (con list) a ] (fun b (fun (fun a (fun [ (con list) a ] b)) b)))
  )
)
```

or if you're not a Cylon:

```
all a b. list a -> b -> (a -> list a -> b) -> b
```

This is the type of the built-in function, i.e. what this PR proposed to implement support for.

We also have a function that is confusingly named the same (the one in `stdlib`), but has `all b` and the subsequent `list a` swapped, because it's some extremely outdated legacy code that was introduced in order to make it easy to replace `unwrap`-based code with code that uses pattern matching. Basically, if you see `StdLib` in the name of a file in this PR, ignore it, it's not really useful for anything and it's not used other than for tests. I apologize for the confusion, I should've specified that `stdlib` stuff should be ignored, it was just for me to test that the proposed builtins type check and evaluate fine.

plutus-core/plutus-core/test/Pretty/Golden/Readable/StdLib/Data/List/FoldrList.golden
effectfully: > If i understand correctly, we would need to allow higher-order builtins?

Yes, but only in the metatheory, the implementation supports arbitrary Plutus types in type signatures of builtins (i.e. the metatheory does it better).

> is `caseList : all a b. list a -> (unit -> b) -> (a -> list a -> unit -> b)` ?

No, the type of `caseList` doesn't have any units, even though they're used inside. They probably shouldn't be used inside, like with the `caseData` definition discussion below. Note that this is the nonsensical stuff from the `stdlib`, so don't think too hard about it, the actual builtin (confusingly named the same) is defined elsewhere, here it's only tested.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
mjaskelioff: That's good to know. I assumed so, but when I asked why there is no higher-order builtin (especially for cases like the ones here where higher-order is the most natural formulation) I was told that it was because of the costs, so I was wondering if there is any problem here.

mjaskelioff (approved):
LGTM. The cases written like this is are the way they should be. 
We will need to decide what to do with the spec and metatheory regarding the types of builtins, but I think the benefits are worth it.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
effectfully: > That's good to know. I assumed so, but when I asked why there is no higher-order builtin (especially for cases like the ones here where higher-order is the most natural formulation) I was told that it was because of the costs

We _can_ define higher-order builtins already and we _can_ cost the definable ones properly. E.g. `\(f : integer -> integer) -> f` is a higher-order builtin, but it's definable and is clearly constant-cost, so there's nothing fancy about costing it.

I think the person giving you comments on higher-order builtins was likely thinking about applying functions and actually reducing them from within a builtin, which we do not support indeed, but

1. we did support that in the past
2. we'd never costed that functionality, but given that it was simply continuing evaluation before returning from the builtin using the same evaluator without any state reinitialization or discarding, I don't think there were any costing issues either, it just didn't matter if the CEK machine (or any other evaluator) execution was invoked directly or continued from within a builtin

> so I was wondering if there is any problem here.

Anyhow, take my words with a grain of salt, I might be failing to spot an issue with costing pattern matching builtins.

michaelpj:
> Which is why my original plan was to allow arbitrary trees of application (in the sense of [Data.Tree.Tree](https://hackage.haskell.org/package/containers-0.6.7/docs/Data-Tree.html#t:Tree)), not just a head and a spine-as-a-list.

Is the "real" answer here that we can return any term that introduces no new variable binders? e.g. introducing `force`/`delay`/`constr` all also seem fine, because it's adding new binders that really throws us off (because then we'd need to shift any binders in the input terms depending on where we put them, etc.

Of course, our arguments are values and not terms, so we can't actually construct a term, instead we have to construct an evaluation continuation or something, so that does make things more complicated. And maybe the most useful and straightforward evaluation continuation is just a spine of applications. Although even then... can I do this?

```
\h a -> h 1 a
```

or this?

```
\h a -> h (addInteger 1 2) a
```

plutus-core/plutus-core/test/Pretty/Golden/Readable/StdLib/Data/Data/caseData.golden
effectfully: In fact, we don't need anything more complicated than

```
-- | Pattern matching over 'Data' inside PLC.
--
-- > \(d : data) -> /\(r :: *) -> caseData {r} d
caseData :: TermLike term TyName Name DefaultUni DefaultFun => term ()
caseData = runQuote $ do
    r <- freshTyName "r"
    d <- freshName "d"
    return
        . lamAbs () d dataTy
        . tyAbs () r (Type ())
        . apply () (tyInst () (builtin () CaseData) $ TyVar () r)
        $ var () d
```

effectfully:
+3% on average on the builtins-heavy `nofib` benchmarks. Not too bad I guess and maybe I could squeeze some performance out of it.

effectfully:
@michaelpj seems like I forgot to hit "comment" last time I wrote a response to your latest comment. Here's another attempt.

> Is the "real" answer here that we can return any term that introduces no new variable binders? e.g. introducing `force`/`delay`/`constr` all also seem fine, because it's adding new binders that really throws us off (because then we'd need to shift any binders in the input terms depending on where we put them, etc.

Yeah, I guess, although we don't seem to need `force` or `delay`. Yet anyway.

Although we could allow constructing arbitrary terms as long as it's impossible to embed values supplied as arguments into them, so that there's no chance of mixing up original and created-by-the-builtin variables, but I'm not even sure what we could use that for, so probably not worth discussing it.

> Although even then... can I do this?
>
> ```
> \h a -> h 1 a
> ```

Yep, the machinery implemented in this PR is enough for that, just checked:

```haskell
this :: Opaque val (Integer -> a -> b) -> Opaque val a -> Opaque (HeadSpine val) b
this h (Opaque a) = headSpine h [fromValue (1 :: Integer), a]
```

> or this?
>
> ```
> \h a -> h (addInteger 1 2) a
> ```

Not with the machinery implemented in this PR (neither nested application nor builtin calls are supported within the denotation of a builtin), but we can add support for those later if we want to.

effectfully:
Now we're talking!

image

image



The data ones have smaller speedup compared to the list ones, because there's more stuff unaffected by the changes going on in them. Gonna experiment a bit to see if I can optimize those benchmarks only using the new builtins.

plutus-benchmark/lists/test/Sum/9.6/left-fold-built-in.eval.golden
effectfully, line 1: TODO: seems like some semantic git conflict or whatever resulted in creation of these spurious golden files. I'll remove them.

effectfully:
Classic.

image

image

plutus-tx-plugin/test/size/Golden/Rational/Serialization/fromBuiltinData.size.golden
```diff
@@ -1 +1 @@
-361
\ No newline at end of file
+275
```
effectfully, line 1: Not bad.

plutus-tx-plugin/test/size/fromBuiltinData.size.golden
effectfully, line 1: Seems accidental.

effectfully:
Closing in favor of #6530.

PR #5488, opened 2023-08-22
Add a nightly test suite

zliu41:
The nightly job runs via GH action on GH-hosted runners. It runs ~~1000x~~ 100x more test cases than CI does.

Test run: https://github.com/input-output-hk/plutus/actions/runs/5932925964/job/16087578544


plutus-core/testlib/PlutusCore/Test.hs
```diff
@@ -537,9 +538,9 @@ test_scopingGood ::
   -- | The runner of the pass.
   (t NameAnn -> TPLC.Quote (t NameAnn)) ->
   TestTree
-test_scopingGood pass gen bindRem preren run =
+test_scopingGood n pass gen bindRem preren run =
   testPropertyNamed (pass ++ " does not break scoping and global uniqueness") "test_scopingGood" $
-    prop_scopingFor gen bindRem preren run
+    withAtLeastTests n (prop_scopingFor gen bindRem preren run)
```
michaelpj, line 543: You can pass this via the CLI. I think you could just put this in your GHA and avoid the whole additional test suite, also it will run all the property tests in the test suite with 1000 tests.
```
cd plutus-core # it needs to run in the package directory
nix run .#hydraJobs.x86_64-linux.ghc92.components.tests.plutus-core.plutus-ir-tests -- --quickcheck-tests 1000
```
(I checked this, it works)

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,32 @@
...
+        uses: cachix/install-nix-action@v22
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+            accept-flake-config = true
+
+      - name: plutus-core-nightly
+        if: always()
+        run: |
+          nix run --no-warn-dirty --accept-flake-config \
```
michaelpj, line 25: You need to be in the package directory for this. You would probably want to build the derivation in `checks` if we were doing it this way.

plutus-core/testlib/PlutusCore/Test.hs
zliu41: When running the whole `plutus-core:plutus-core-test` with `--hedgehog-tests 100000`, the GH action got killed after about one hour ([attempt1](https://github.com/input-output-hk/plutus/actions/runs/5943963244/attempts/1), [attempt2](https://github.com/input-output-hk/plutus/actions/runs/5943963244/attempts/2)). Unclear whether they are OOM kills, but GH-hosted runners do have relatively tight resources. I lowered the number of test cases from 100k to 10k.

zliu41 (commented):
Updated.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,34 @@
...
+        if: always()
+        run: |
+          cd plutus-core
+          nix run --no-warn-dirty --accept-flake-config \
+            .#hydraJobs.x86_64-linux.ghc92.components.tests.plutus-core.plutus-core-test -- --hedgehog-tests 10000
+
+      - name: plutus-ir-nightly
+        if: always()
+        run: |
+          cd plutus-core
```
michaelpj, line 32: Aren't we already in `plutus-core`? I don't think the working directory resets between steps. You could put a `popd` at the end of each step for isolation, I guess.

michaelpj (approved):

.github/workflows/nightly.yml
zliu41: I think this is done automatically, but I added pushd/popd anyway.

PR #5492, opened 2023-08-23
Add one more impredicativity wrapper in PlutusTx.Lift.TH

zliu41:
This appeases core lint on `PlutusTx.Lift.TH`. Hopefully it will prevent some weird errors from happening.

The core lint error was caused by the fact that in `liftExprs' = $$(TH.liftSplice $ tyListE liftExprs)`, `tyListE liftExprs` has the wrong type (it should have an impredicative type, but GHC inferred a non-impredicative type where `fun` is not in scope).

### DIFF
--- a/plutus-tx/src/PlutusTx/Lift/TH.hs
+++ b/plutus-tx/src/PlutusTx/Lift/TH.hs
@@ -15,6 +15,9 @@
 {-# LANGUAGE TemplateHaskell       #-}
 {-# LANGUAGE TypeApplications      #-}
 {-# LANGUAGE ViewPatterns          #-}
+-- It is complex to mix TH with polymorphisms. Core lint can sometimes catch problems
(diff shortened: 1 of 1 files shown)


plutus-tx/src/PlutusTx/Lift/TH.hs
```diff
@@ -15,6 +15,9 @@
 {-# LANGUAGE TemplateHaskell       #-}
 {-# LANGUAGE TypeApplications      #-}
 {-# LANGUAGE ViewPatterns          #-}
+-- It is complex to mix TH with polymorphisms. Core lint can sometimes catch problems
+-- caused by using polymorphisms the wrong way, e.g., accidentally using impredicative types.
+{-# OPTIONS_GHC -dcore-lint #-}
```
michaelpj, line 20: good idea!

michaelpj (approved):
Great catch!

PR #5494, opened 2023-08-25
Handle integerNegate in the plugin

zliu41:
GHC 9.6 unboxes Integer literals, and `-123` becomes `integerNegate (IS 123)` in Core. We can't compile `integerNegate`, so it must be handled specially.


plutus-tx-plugin/testlib/IntegerLiteral/NoStrict.hs
```diff
@@ -0,0 +1,44 @@
...
+import PlutusTx.Code
+import PlutusTx.Prelude qualified as PlutusTx
+import PlutusTx.TH (compile)
+
+integerLiterals :: CompiledCode (Integer -> Integer)
+integerLiterals =
+  $$( compile
+        [||
+        \x ->
+          let !smallStrict = 123
```
michaelpj, line 23: I remember how I've handled this in the past: `LANGUAGE NegativeLiterals`. That avoids the `integerNegate` calls. Nonetheless, I think handling `integerNegate` is reasonable.

plutus-tx-plugin/plutus-tx-plugin.cabal
```diff
@@ -115,6 +115,17 @@ executable gen-plugin-opts-doc
 
   default-language: Haskell2010
 
+library testlib-internal
```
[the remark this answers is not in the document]
zliu41: I can't put it in `plutus-tx-tests`, because we don't build the latter with GHC 9.6.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -903,6 +905,22 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
     GHC.Coercion _ -> throwPlain $ UnsupportedError "Coercions as expressions"
 
+compileIntegerNegate ::
+  (CompilingDefault uni fun m ann) =>
+  GHC.CoreExpr ->
+  m (PIRTerm uni fun)
+compileIntegerNegate e = case e of
+  GHC.Var n `GHC.App` arg
+    | GHC.getName n == GHC.integerNegateName ->
+        -- `integerNegate (integerNegate arg)` is equivalent to `arg`
```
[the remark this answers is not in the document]
zliu41: > What happens if I do a triple negate?

It cancels the first two. `compileExpr (negate (negate (negate arg))) ==> compileExpr (negate arg)`.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -903,6 +905,22 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
+        compileExpr arg
+  GHC.Var (GHC.idDetails -> GHC.DataConWorkId dc) `GHC.App` GHC.Lit (GHC.LitNumber _ i)
```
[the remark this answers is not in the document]
zliu41: You mean turn `integerNegate arg` into `0 - arg`? That sounds reasonable; I haven't thought of that approach.

plutus-tx-plugin/testlib/IntegerLiteral/NoStrict.hs
```diff
@@ -0,0 +1,44 @@
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell     #-}
+
+{-# LANGUAGE NoStrict            #-}
+
+-- | We use this module to ensure that integer literals are handled correctly by all
+-- supported GHC versions, when @Strict@ is off (we get different Core depending on
```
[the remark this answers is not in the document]
zliu41: The Strict version of this doesn't compile - we need to be able to handle `ByteArray#` first.

plutus-tx-plugin/plutus-tx-plugin.cabal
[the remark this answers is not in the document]
michaelpj: ... but the _test_ is in there. So what does this achieve? I guess it shows that the plugin doesn't blow up, but that's not that much. 

If we want to start testing the behaviour on 9.6 I think we could do that but I'd prefer to do it systematically, e.g. use CPP to put the GHC version into the golden file and check in both versions.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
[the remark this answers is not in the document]
michaelpj: Yes exactly.

plutus-tx-plugin/plutus-tx-plugin.cabal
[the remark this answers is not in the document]
michaelpj: Regardless, it's not at all obvious what is going on here so we need to write it down!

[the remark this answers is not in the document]
zliu41: Ensuring that plugin does not blow up in any GHC version is the main purpose. Yes we also want to check the golden files, but we can't do that for both 9.2 and 9.6 until, like you said, we do it in a systematic way.

plutus-tx-plugin/testlib/IntegerLiteral/NoStrict.hs
zliu41: `NegativeLiterals` work for most cases (it turns `integerNegate (IP ...)` into `IN ...`), but for double negation, we still get `integerNegate`.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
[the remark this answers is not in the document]
zliu41: Oh, I guess the callsite inliner should take care of that.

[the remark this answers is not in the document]
michaelpj: That's the optimizer's problem :)

plutus-tx-plugin/plutus-tx-plugin.cabal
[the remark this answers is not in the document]
michaelpj: Okay, but let's be systematic. That suggests moving _all_ of the test definitions into this library.

plutus-tx-plugin/testlib/IntegerLiteral/NoStrict.hs
michaelpj: You just [told me](https://github.com/input-output-hk/plutus/pull/5494#discussion_r1305671039) the compiler cancels double negation!

Anyway, if we just turn `integerNegate` into a function, it doesn't matter any more.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -765,11 +764,17 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
       case maybeDef of
         Just term -> pure term
         Nothing ->
-          throwSd FreeVariableError $
-            "Variable"
-              GHC.<+> GHC.ppr n
-              GHC.$+$ (GHC.ppr $ GHC.idDetails n)
-              GHC.$+$ (GHC.ppr $ GHC.realIdUnfolding n)
+          -- No other cases apply; compile the unfolding of the var
```
michaelpj, line 767: I would mildly prefer to keep this as a separate case and put it later. That way the fallthrough case is more obviously part of the series of `Var`-related matches. We can put the necessary checks into the pattern here:
```
GHC.Var n | let lexName = LexName $ GHC.getName n, Just term <- PIR.lookupTerm annMayInline lexName = ...
```

plutus-tx-plugin/testlib/IntegerLiteral/NoStrict.hs
```diff
@@ -0,0 +1,44 @@
+{-# LANGUAGE BangPatterns        #-}
```
michaelpj, line 1: I'm confused, why do we have this file and also `IntegerLIterals/NoStrict`?

plutus-tx-plugin/test/IntegerLiterals/NoStrict/9.6/integerLiterals-NoStrict.pir.golden
```diff
@@ -0,0 +1,76 @@
...
+    )
+    (lam
+      x
+      (con integer)
+      (let
+        (nonrec)
+        (termbind
+          (strict)
+          (vardecl y (con integer))
+          [ integerNegate (con integer 24680135792468013579) ]
```
michaelpj, line 18: This is pretty surprising! So in 9.2 GHC turns a double negation into nothing, but in 9.6 it turns it into two calls to `integerNegate`, even with `NegativeLiterals` turned on??? Is `NegativeLiterals` broken?

plutus-tx-plugin/test/IntegerLiterals/NoStrict/9.6/integerLiterals-NoStrict.pir.golden
```diff
@@ -0,0 +1,76 @@
+(program
+  1.1.0
+  (let
+    (nonrec)
+    (termbind
+      (strict)
+      (vardecl integerNegate (fun (con integer) (con integer)))
```
michaelpj, line 7: I'd like to see some examples in tests where we have some optimization. I hope we inline this: it's a guarantee win if the argument is a constant since we'll then reduce it.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -1050,13 +1055,34 @@ coverageCompile originalExpr exprType src compiledTerm covT =
     findHeadSymbol (GHC.Cast t _) = findHeadSymbol t
     findHeadSymbol _              = Nothing
 
+-- | We cannot compile the unfolding of `GHC.Num.Integer.integerNegate`, so we directly
```
michaelpj, line 1058: and why does that matter?
```suggestion
-- | We cannot compile the unfolding of `GHC.Num.Integer.integerNegate`, which is important because GHC inserts calls to it when it sees negations, even negations of literals (unless NegativeLiterals is on, which it usually isn't). So we directly
```

plutus-tx-plugin/test/IntegerLiterals/NoStrict/9.6/integerLiterals-NoStrict.pir.golden
zliu41: No, GHC doesn't do anything. It is my first implementation of handling `integerNegate` that cancels double negation!

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
zliu41: `lookupTerm` is a monadic action and can't be put in the guards!

plutus-tx-plugin/test/IntegerLiterals/NoStrict/9.6/integerLiterals-NoStrict.pir.golden
zliu41: It is now inlined, thanks to `annAlwaysInline`!

plutus-tx-plugin/test/IntegerLiterals/NoStrict/NoNegativeLiterals/Spec.hs
```diff
@@ -0,0 +1,58 @@
...
+import PlutusTx.Code
+import PlutusTx.Prelude qualified as PlutusTx
+import PlutusTx.Test
+import PlutusTx.TH (compile)
+
+import Test.Tasty.Extras
+
+tests :: TestNested
+tests = testNestedGhc "IntegerLiterals"
+  [ goldenPir "integerLiterals-NoStrict-NoNegativeLiterals" integerLiterals
```
michaelpj, line 29: why not just include `NoStrict` as part of the test nesting?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
michaelpj: bah

plutus-tx-plugin/test/IntegerLiterals/NoStrict/9.6/integerLiterals-NoStrict.pir.golden
michaelpj: I'm not sure if that's the right call either. It _might_ be bad to inline if the argument isn't a constant. Probably we should just try and make sure the inliner is smart enough to see that. Not sure.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -1050,13 +1062,43 @@ coverageCompile originalExpr exprType src compiledTerm covT =
     findHeadSymbol (GHC.Cast t _) = findHeadSymbol t
     findHeadSymbol _              = Nothing
 
+-- | We cannot compile the unfolding of `GHC.Num.Integer.integerNegate`, which is important because GHC inserts calls to it when it sees negations, even negations of literals (unless NegativeLiterals is on, which it usually isn't). So we directly
+-- define a PIR term for it: @integerNegate = \x -> 0 - x@.
+defineIntegerNegate :: (CompilingDefault PLC.DefaultUni fun m ann) => m ()
+defineIntegerNegate = do
+  ghcId <- GHC.tyThingId <$> getThing 'GHC.Num.Integer.integerNegate
+  -- Always inline `integerNegate`.
```
michaelpj, line 1070: why?

plutus-tx-plugin/test/IntegerLiterals/NoStrict/NoNegativeLiterals/Spec.hs
zliu41: This way all golden files are in the same folder, which is a bit more convenient imo

plutus-tx-plugin/test/IntegerLiterals/NoStrict/NoNegativeLiterals/Spec.hs
```diff
@@ -0,0 +1,58 @@
...
+  ]
+
+integerLiterals :: CompiledCode (Integer -> Integer)
```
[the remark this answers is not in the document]
zliu41: Probably not

plutus-tx-plugin/test/IntegerLiterals/NoStrict/9.6/integerLiterals-NoStrict.pir.golden
zliu41: There's no other good way to make it inline, because `let !x = negateInteger 100` is not considered safe to inline, unless we somehow special-case it in the inliner. And the argument is almost always constant.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
zliu41: Otherwise we get a lot of `integerNegate`s in _optimized_ PIR!

michaelpj: Right, but is that bad?

plutus-tx-plugin/test/IntegerLiterals/NoStrict/9.6/integerLiterals-NoStrict.pir.golden
michaelpj: > There's no other good way to make it inline, because let !x = negateInteger 100 is not considered safe to inline

That's different. I'm talking about inlining the `negateInteger` call, not `x`. And `negateInteger` is certainly safe to inline, it just won't tend to be smaller, which is why we won't inline it.

>  And the argument is almost always constant.

... in your test cases. Now the plugin supports it, people can just write `-(some expression)` throughout their code. Who knows what's in there. It's probably not a big deal, but worth discussing.

plutus-tx-plugin/test/IntegerLiterals/NoStrict/NegativeLiterals/Spec.hs
```diff
@@ -0,0 +1,58 @@
...
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE TypeApplications      #-}
+
+{-# LANGUAGE NegativeLiterals      #-}
+{-# LANGUAGE NoStrict              #-}
+
+{-# OPTIONS_GHC -fplugin PlutusTx.Plugin #-}
```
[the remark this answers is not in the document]
michaelpj: worth a comment, since basically all the other tests run with simplification off

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
michaelpj: > And without the annMayInline annotation, the inliner cannot make this transformation.

not today! Maybe we should fix that! 

I just want you to write down the reasoning, it's not obvious.

zliu41: Added:
> -- Always inline `integerNegate`.
  -- `let integerNegate = \x -> 0 - x in integerNegate 1 + integerNegate 2`
  -- is much more expensive than `(-1) + (-2)`. The inliner cannot currently
  -- make this transformation without `annAlwaysInline`, because it is not aware
  -- of constant folding.

plutus-tx-plugin/test/IntegerLiterals/NoStrict/NegativeLiterals/Spec.hs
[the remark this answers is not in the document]
zliu41: Added:
> -- This module runs the PIR and UPLC simplifiers, because (1) we want to verify that
-- `integerNegate` is compiled away; (2) it is easier to tell from the optimized PIR
-- whether or not the signs of the numbers are correct, which is ultimately what we
-- care about.

zliu41 (commented):
Added some comments.

michaelpj (approved):

PR #5499, opened 2023-08-29
Handle ByteArray# in the plugin

zliu41:
This enables the plugin to compile `let !x = 12345678901234567890 in PlutusTx.equalsInteger x y` with GHC 9.6.


plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -793,6 +799,31 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
     GHC.Lam b@(GHC.isTyVar -> True) body -> mkTyAbsScoped b $ compileExpr body
     -- otherwise it's a normal lambda
     GHC.Lam b body -> mkLamAbsScoped b $ compileExpr body
+    -- Handle the following case:
```
michaelpj, line 802: I think we could just roll this into the case below? We would just always use `withVarTyScoped`, and have something like
```
let isWeirdBytestringIntegerLiteral = GHC.eqType (GHC.varType b) GHC.byteArrayPrimTy && isNumericLiteral rhs
ty <- if isWeirdBytestringIntegerLiteral then pure integerTy else compileType ghcType
withVarTyScoped b ty $ ...
```

michaelpj (approved):
Nice. I hadn't realised that the underlying Core expression is actually still a numeric literal! That makes this kind of okay.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
zliu41: Yeah, good point.

PR #5502, opened 2023-08-30
 Rename "builtin versions" to "builtin semantics variants" to reduce risk of confusion

kwxm:
[PLT-7306]

The `ToBuiltinMeaning` class in PlutusCore.Builtin.Meaning has an associated class called `BuiltinVersion` and in PlutusCore.Default.Builtins this is instantiated as `data BuiltinVersion DefaultFun = DefaultFunV1 | DefaultFunV2`
This risks confusion with PlutusV1, PlutusV2 etc. (as seen in [this issue](https://github.com/input-output-hk/plutus/issues/5475)).

After some Slack discussion we’ve decided to rename `BuiltinVersion` to `BuiltinSemanticsVariant` and `DefaultFunV1` to `DefaultFunSemanticsVariant1` , etc. I've also tried to remove all mentions of "builtin versions" in comments.

[PLT-7306]: https://input-output.atlassian.net/browse/PLT-7306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
@@ -103,7 +103,8 @@ typeOfBuiltinFunction semvar fun =
     case toBuiltinMeaning @_ @_ @(Term TyName Name uni fun ()) semvar fun of
         BuiltinMeaning sch _ _ -> typeSchemeToType sch
 
-{- Note [Builtin semantics variants] The purpose of the "builtin semantics
+{- Note [Builtin semantics variants] 
+The purpose of the "builtin semantics
 variant" feature is to provide multiple, different denotations (implementations)
 for the same builtin(s).  An example use of this feature is for "fixing" the
 behaviour of `ConsByteString` builtin to throw an error instead of overflowing
--- a/plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
@@ -70,57 +70,76 @@ data BuiltinMeaning val cost =
 type HasMeaningIn uni val = (Typeable val, ExMemoryUsage val, HasConstantIn uni val)
 
 -- | A type class for \"each function from a set of built-in functions has a 'BuiltinMeaning'\".
-class (Typeable uni, Typeable fun, Bounded fun, Enum fun, Ix fun, Default (BuiltinVersion fun)) =>
-            ToBuiltinMeaning uni fun where
+class ( Typeable uni
+      , Typeable fun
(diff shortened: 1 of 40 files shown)


plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
```diff
@@ -119,17 +119,21 @@ newtype EvaluationContext = EvaluationContext
     deriving stock Generic
     deriving anyclass (NFData, NoThunks)
 
-{-|  Create an 'EvaluationContext' for a given plutus-builtins' version.
+{-|  Create an 'EvaluationContext' for a given builtin semantics variant.
```
kwxm, line 123: Is that right?  I presume it's not referring to the ledger language version, although this is in fact what gets called in the V1/V2/V2 `mkEvaluationContext` functions.

plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
```diff
@@ -70,57 +70,76 @@ data BuiltinMeaning val cost =
...
+typeOfBuiltinFunction semvar fun =
+    case toBuiltinMeaning @_ @_ @(Term TyName Name uni fun ()) semvar fun of
         BuiltinMeaning sch _ _ -> typeSchemeToType sch
 
-{- Note [Versioned builtins]
-The purpose of the "versioned builtins" feature is to provide multiple, different denotations
-(implementations) for the same builtin(s).
-An example use of this feature is for "fixing" the behaviour of `ConsByteString` builtin to throw an
-error instead of overflowing its first argument.
+{- Note [Builtin semantics variants] The purpose of the "builtin semantics
```
michaelpj, line 106:
```suggestion
{- Note [Builtin semantics variants] 
The purpose of the "builtin semantics
```

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
michaelpj: Yes, I think it is right. It's getting called with the corresponding `BuiltinSemanticsVariant`. Possibly we should just use this generically, and instead have for each language a clearly associated `BuiltinSemanticsVariant` or something.

michaelpj (approved):
Tedious, but good!

plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
kwxm: Oh, that seems to have broken something.  I'll try to fix it manually.

PR #5503, opened 2023-08-30
Fix `testOrd`

zliu41:
It fails if ratio denominator is 0: https://github.com/input-output-hk/plutus/runs/16359392302

### DIFF
--- a/plutus-tx-plugin/test/StdLib/Spec.hs
+++ b/plutus-tx-plugin/test/StdLib/Spec.hs
@@ -116,9 +116,7 @@ testOrd :: Property
 testOrd = Hedgehog.property $ do
     let gen = Gen.integral (Range.linear (-10000) 100000)
         -- Ratio must have non-zero denominator or else an ArithException will be thrown.
-        gen' =
-          Gen.choice
-            [Gen.integral (Range.linear (-10000) (-1)), Gen.integral (Range.linear 1 100000)]
+        gen' = Gen.filter (/= 0) gen
     n1 <- Hedgehog.forAll $ (%) <$> gen <*> gen'
     n2 <- Hedgehog.forAll $ (%) <$> gen <*> gen'
     ghcResult <- tryHard $ n1 <= n2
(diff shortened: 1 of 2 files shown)


plutus-tx-plugin/test/StdLib/Spec.hs
```diff
@@ -115,8 +115,12 @@ testReduce = Hedgehog.property $ do
 testOrd :: Property
 testOrd = Hedgehog.property $ do
     let gen = Gen.integral (Range.linear (-10000) 100000)
-    n1 <- Hedgehog.forAll $ (%) <$> gen <*> gen
-    n2 <- Hedgehog.forAll $ (%) <$> gen <*> gen
+        -- Ratio must have non-zero denominator or else an ArithException will be thrown.
+        gen' =
+          Gen.choice
+            [Gen.integral (Range.linear (-10000) (-1)), Gen.integral (Range.linear 1 100000)]
```
effectfully, line 121: [`filter`](https://hackage.haskell.org/package/hedgehog-1.4/docs/Hedgehog-Gen.html#v:filter) would perhaps be a bit more readable, but this is fine too.

effectfully (approved):

plutus-tx-plugin/test/StdLib/Spec.hs
zliu41: Thanks, `filter` is much better.

plutus-tx/src/PlutusTx/IsData/Instances.hs
```diff
@@ -4,6 +4,7 @@
 {-# LANGUAGE TemplateHaskell   #-}
 {-# OPTIONS_GHC -fno-specialise #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
+{-# OPTIONS_GHC -ddump-splices #-}
```
michaelpj, line 7: this shouldn't have been committed

PR #5506, opened 2023-08-31
Kwxm/cost model representations

kwxm:
We had an apparent cost regression a while ago (see PLT-5956), but it turned out to be due to the cost model parameters being supplied in the wrong order.  There are multiple representations of the cost model and this is an attempt to explain what they're all for.  Contributions and corrections are most welcome.

### DIFF
--- a/doc/notes/cost-model-representations/cost-model-representations.md
+++ b/doc/notes/cost-model-representations/cost-model-representations.md
@@ -158,7 +158,7 @@ various versions used in Cardano.
 The ledger interacts with the Plutus Core evaluator using functions defined in
 [plutus-ledger-api](https://github.com/input-output-hk/plutus/tree/master/plutus-ledger-api).
 The provides functions
-[evaluateScriptCounting](https://github.com/input-output-hk/cardano-ledger/blob/330b42db03fec425ad72c98cb6931f979e59941b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs#L330)
+[evaluateScriptCounting](https://github.com/input-output-hk/plutus/blob/e773e58ea0e4a8088fed0ea5f934a7c413caa5b3/plutus-ledger-api/src/PlutusLedgerApi/V1.hs#L158)
 and
 [evaluateScriptRestricting](https://github.com/input-output-hk/plutus/blob/e773e58ea0e4a8088fed0ea5f934a7c413caa5b3/plutus-ledger-api/src/PlutusLedgerApi/V1.hs#L173)
 (there are PlutusV2 and PlutusV3 versions of these too) to run scripts on the
--- a/doc/notes/cost-model-representations/cost-model-representations.md
+++ b/doc/notes/cost-model-representations/cost-model-representations.md
@@ -1,0 +1,216 @@
+# Representations of the Plutus Core cost model
+
+The `plutus` codebase uses several different representations of the Plutus Core
+cost model.  It can be somewhat difficult to keep track of what all of these are
+used for; this document attempts to give an overview to clarify matters.  An
+essential point is that if you're using `mkEvaluationContext` then cost model
+parameters much be provided in the correct order, otherwise script costs
+will be inaccurate: see [`EvaluationContext`](#evaluationcontext) below.
+
+## Introduction
+
+The Plutus Core language incorporates a _cost model_ which assigns costs for CPU
+usage and memory usage to the evaluation of an Untyped Plutus Core program on
+the standard Plutus evaluator (the _CEK machine_).  CPU costs are measured in
+abstract units called `ExUnit`s (_execution units_), with one `ExUnit`
+corresponding to one picosecond on a specific dedicated benchmarking machine.
+We expect that that the actual time required to execute a program on some other
+machine will be (approximately) proportional to the total number of `ExUnit`s.
+
+Memory usage, or "size", is measured in units of 64-bit words (although with
+hindsight the number of bytes required to store an object might have been a
+better measure).
+
+The cost model consists of two components:
+
+  1. A CPU and memory cost for every basic step of the evaluator. 
+  2. A pair of _costing functions_ for each built-in function which give an
+     approximation to the CPU usage and memory usage of the function.  The cost
+     generally depends on the size of the input(s) to the builtin, which is why
+     we need a function rather than just a constant.  The CPU costing functions are
+     determined by running benchmarks and fitting regression models to the
+     results.  The memory costs of a built-in function (which measure only the
+     memory required to store the output of the function, and not any transient
+     memory allocated during execution) are determined by understanding what the
+     builtin does, and perhaps by examining the implementation.
+
+### Serialised representation
+
+Concrete values for the coefficients of the costing functions for the builtins
+in the current version of Plutus Core are stored in the file
+[`builtinCostModel.json`](https://github.com/input-output-hk/plutus/blob/master/plutus-core/cost-model/data/builtinCostModel.json)
+and values for the machine steps are stored in
+[`cekMachineCosts.json`](https://github.com/input-output-hk/plutus/blob/master/plutus-core/cost-model/data/cekMachineCosts.json
+) (both in `plutus-core/cost-model/data`).
+
+For example, the costing functions for `lessThanEqualsByteString` are
+represented by the following entry, which says that the CPU cost of comparing
+two bytestrings `r` and `s` is `156 + 197145 * min {size(r), size(s)}`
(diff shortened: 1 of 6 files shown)


bezirg:
I agree with what the doc says.

doc/notes/cost-model-representations/cost-model-representations.md
```diff
@@ -0,0 +1,216 @@
+# Representations of the Plutus Core cost model
```
michaelpj, line 1: Can we move any of this to Haddock on the appropriate bits of the codebase?

doc/notes/cost-model-representations/cost-model-representations.md
```diff
@@ -0,0 +1,216 @@
...
+
+The `plutus` codebase uses several different representations of the Plutus Core
+cost model.  It can be somewhat difficult to keep track of what all of these are
+used for; this document attempts to give an overview to clarify matters.  An
+essential point is that if you're using `mkEvaluationContext` then cost model
+parameters much be provided in the correct order, otherwise script costs
+will be inaccurate: see [`EvaluationContext`](#evaluationcontext) below.
```
michaelpj, line 8: e.g. this information is crucial for users of `mkEvaluationContext`, so it should probably be discoverable there.

doc/notes/cost-model-representations/cost-model-representations.md
```diff
@@ -0,0 +1,216 @@
...
+     we need a function rather than just a constant.  The CPU costing functions are
+     determined by running benchmarks and fitting regression models to the
+     results.  The memory costs of a built-in function (which measure only the
+     memory required to store the output of the function, and not any transient
+     memory allocated during execution) are determined by understanding what the
+     builtin does, and perhaps by examining the implementation.
+
+### Serialised representation
+
+Concrete values for the coefficients of the costing functions for the builtins
```
michaelpj, line 39: Worth being clear that these are not the values on mainnet, since people get confused about that. Also this could go in the Haddock for the Haskell values that are the read-in versions of these files.

doc/notes/cost-model-representations/cost-model-representations.md
```diff
@@ -0,0 +1,216 @@
...
+The costs for the different types of CEK steps (all of which currently have the
+same cost) are defined in
+[cekMachineCosts.json](https://github.com/input-output-hk/plutus/blob/master/plutus-core/cost-model/data/cekMachineCosts.json)
+which is compiled into a CekMachineCosts object called [defaultCekMachineCosts](https://github.com/input-output-hk/plutus/blob/b321575d9266b3358b9e728d064fc0bee4f355d7/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetingDefaults.hs#L65),
+and this is bundled together with the builtin costs in [defaultCekCostModel](https://github.com/input-output-hk/plutus/blob/b321575d9266b3358b9e728d064fc0bee4f355d7/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetingDefaults.hs#L69) (of
+type [CostModel CekMachineCosts BuiltinCostModel](https://github.com/input-output-hk/plutus/blob/e773e58ea0e4a8088fed0ea5f934a7c413caa5b3/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/MachineParameters.hs#L28)).
+
+#### Flattened cost model entries
+The cost model is sometimes flattened into a list of pairs of (key, value)
```
michaelpj, line 96: Maybe something like: "While the nested structure is convenient for representing the structure of the parameters, for interoperating with other sources of parameter values (such as the ledger) it is convenient to have a flat representation of the parameters."?

michaelpj: This should all be in the haddock for `CostModelParams` or `CostModelInterface`

doc/notes/cost-model-representations/cost-model-representations.md
```diff
@@ -0,0 +1,216 @@
...
+See also the Notes at the top of
+[CostModelInterface.hs](https://github.com/input-output-hk/plutus/blob/e773e58ea0e4a8088fed0ea5f934a7c413caa5b3/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs).
+
+
+#### MachineParameters
+
+The [MachineParameters](https://github.com/input-output-hk/plutus/blob/e773e58ea0e4a8088fed0ea5f934a7c413caa5b3/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/MachineParameters.hs#L39) type bundles together a cost model and the denotations of
+the builtins, and this is used by the machinery which evaluates builtins to both
+evaluate a call to a builtin and cost it.  It's generic over a type `fun` which
+describes the built-in functions.  Usually the `MachineParameters` object that we
```
michaelpj, line 125:
```suggestion
describes the built-in functions.  In testing, the `MachineParameters` object that we
```

doc/notes/cost-model-representations/cost-model-representations.md
```diff
@@ -0,0 +1,216 @@
...
+use is `defaultCekParameters`, where `fun` is
```
michaelpj, line 126: ... which is built from the default cost models above?

michaelpj (approved):
This is all correct, I just think we should try and put it in places where it's going to be easier to find and keep up to date as the codebase changes.

doc/notes/cost-model-representations/cost-model-representations.md
kwxm: Can we put it there, but keep this too?  I wrote this down because I was finding it hard to see how all the pieces fit together.

michaelpj: Sure, I just think if we're going to the trouble of writing this we should put as much of it into the hopefully more discoverable haddocks.

doc/notes/cost-model-representations/cost-model-representations.md
kwxm: It does already [say this](https://github.com/input-output-hk/plutus/blob/e098652b6985c6ae52477e2f0931fe2364732aaa/plutus-ledger-api/src/PlutusLedgerApi/V1/EvaluationContext.hs#L22):
> The input is a list of integer values passed from the ledger and
are expected to appear in correct order.

I'll expand on this and some of the other comments, but I think that a lot of the content of this document is already in the Haddock, but spread around in such a way that it's difficult to see the big picture.

kwxm:
I've got an annoying circularity problem here.  The document contains permalinks to definitions of various objects, but I'm just about to update some of the comments for those to include content from this file, and the links will point to previous versions with the old comments.

doc/notes/cost-model-representations/cost-model-representations.md
kwxm: > Can we move any of this to Haddock on the appropriate bits of the codebase?

I've done that for some of the stuff here, but I gave up after some time because this document is mostly a summary of stuff that's spread out over numerous Haddock comments in different parts of the codebase and I was just finding myself repeating stuff that's already there.  I think the Haddock's a bit problematic because it's hard to know where to start: the point of this document is to give you an overview of the whole thing, and splitting it up and putting it back in the Haddock kind of defeats the purpose.  Maybe we need a big giant Haddock comment in a single place, with links to relevant bits of code. I'm not sure where would be a sensible place to put that though.

doc/notes/cost-model-representations/cost-model-representations.md
kwxm: > > have a flat representation of the parameters."?
> 
> I don't like "flat representation" because it reminds me flat, the serialization format

I see what you mean.  Would "flattened" be OK?  I can't think of a better way of putting it.

bezirg: Yes, flattened better

bezirg:
> Comparing benchmark results of ' plutus-benchmark:validation' on '4849f777d' (base) and '75ebb3eb8' (PR)
> 
> Results table

So does this mean that the machine is inconsistent again?

kwxm:
> > Comparing benchmark results of ' plutus-benchmark:validation' on '4849f777d' (base) and '75ebb3eb8' (PR)
> > Results table
> 
> So does this mean that the machine is inconsistent again?

I think the problem may be that it's comparing against master and other changes have been committed since the start of this branch (that doesn't mean that the machine isn't inconsistent though). There's a lengthy Slack discussion about this.  

[I'm quite proud of the triple negative in the parenthetical comment there.]

kwxm:
I think I'll merge this later today if no-one objects: I'm going to be off for a couple of weeks and it would be good to get it out of the way.

PR #5507, opened 2023-09-01
Update the download link and the date on the specification

kwxm:
The README file in the directory containing the PLC specification had a broken link to the PDF (we don't have a checked-in PDF any more) and this is supposed to fix it.  I also changed the date on the PDF because it's not 2024 yet.


michaelpj:
Could also just direct people to the README, but sure. Not sure what I was thinking with the 2024 date :joy:

michaelpj (approved):

PR #5508, opened 2023-09-01
Add --builtin-semantics-version option to plc and uplc commands

kwxm:
This adds an option to the `plc` and `uplc` commands so that we can say which variant of the builtin semantics we want to use.
This is a step towards PLT-372, but we may want to make it take the ledger language version and/or the PLC language version configurable too.  That would probably require us to move the executables elsewhere though, maybe into a package of their own.

```
$ cat cons.plc
(program 1.0.0
[(builtin consByteString) (con integer 400) (con bytestring #213123213421)]
)

$ cabal run uplc evaluate -- -i cons.uplc -B1
(con bytestring #90213123213421)

$ cabal run uplc evaluate -- -i cons.uplc -B2
An error has occurred:  User error:
Caused by: (consByteString 400 #213123213421)

### DIFF
--- a/plutus-core/executables/plc/Main.hs
+++ b/plutus-core/executables/plc/Main.hs
@@ -7,6 +7,7 @@ module Main (main) where
 import PlutusCore qualified as PLC
 import PlutusCore.Compiler.Erase qualified as PLC (eraseProgram)
 import PlutusCore.Data
+import PlutusCore.Default (BuiltinSemanticsVariant (..))
 import PlutusCore.Evaluation.Machine.Ck qualified as Ck
 import PlutusCore.Evaluation.Machine.ExBudgetingDefaults qualified as PLC
 import PlutusCore.Executable.Common
@@ -29,7 +30,13 @@ plcInfoCommand :: ParserInfo Command
 plcInfoCommand = plutus plcHelpText
 
 data TypecheckOptions = TypecheckOptions Input Format
-data EvalOptions      = EvalOptions Input Format PrintMode TimingMode
+data EvalOptions =
+    EvalOptions
+      Input
+      Format
+      PrintMode
+      TimingMode
+      (BuiltinSemanticsVariant PLC.DefaultFun)
 data EraseOptions     = EraseOptions Input Format Output Format PrintMode
 
 
@@ -56,7 +63,7 @@ eraseOpts = EraseOptions <$> input <*> inputformat <*> output <*> outputformat <
 
 evalOpts :: Parser EvalOptions
 evalOpts =
-  EvalOptions <$> input <*> inputformat <*> printmode <*> timingmode
+  EvalOptions <$> input <*> inputformat <*> printmode <*> timingmode <*> builtinSemanticsVariant
 
 plutus ::
   -- | The @helpText@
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetingDefaults.hs [hunk 1]
```diff
@@ -90,7 +91,16 @@ unitCekParameters =
     noinline mkMachineParameters def $
         CostModel unitCekMachineCosts unitCostBuiltinCostModel
 
-defaultBuiltinsRuntime :: HasMeaningIn DefaultUni term => BuiltinsRuntime DefaultFun term
+defaultBuiltinsRuntimeForSemanticsVariant
```
kwxm, line 94: @effectfully Do we maybe want to make this the thing that's actually exported and get the user to say which variant to use?  `BuiltinSemanticsVariant2` is what gets used by default at the moment.

plutus-core/executables/plc/Main.hs
```diff
@@ -179,9 +186,9 @@ runOptimisations (OptimiseOptions inp ifmt outp ofmt mode) = do
 ---------------- Evaluation ----------------
 
 runEval :: EvalOptions -> IO ()
-runEval (EvalOptions inp ifmt printMode timingMode) = do
+runEval (EvalOptions inp ifmt printMode timingMode semvar) = do
   prog <- readProgram ifmt inp
-  let evaluate = Ck.evaluateCkNoEmit PLC.defaultBuiltinsRuntime
+  let evaluate = Ck.evaluateCkNoEmit (PLC.defaultBuiltinsRuntimeForSemanticsVariant semvar)
```
kwxm, line 191: We should perhaps get the typechecker to take account of the variant too, but currently the types are the same for all variants.  Would it even be feasible (or desirable) to have different types for different variants?

michaelpj: Yeah, currently we're assuming it just affects the dynamic semantics and not the static semantics... which seems reasonable for now.

michaelpj: Yeah, currently we're assuming it just affects the dynamic semantics and not the static semantics... which seems reasonable for now.

michaelpj (approved):
Seems fine

plutus-core/executables/src/PlutusCore/Executable/Parsers.hs
```diff
@@ -159,6 +160,32 @@ exampleSingle = ExampleSingle <$> exampleName
...
 
+builtinSemanticsVariantReader :: String -> Maybe (BuiltinSemanticsVariant DefaultFun)
+builtinSemanticsVariantReader =
+    \case
+     "1" -> Just DefaultFunSemanticsVariant1
+     "2" -> Just DefaultFunSemanticsVariant2
+     _   -> Nothing
+
+-- This is used to make the help message show you what you actually need to type.
+showBuiltinSemanticsVariant :: BuiltinSemanticsVariant DefaultFun -> String
```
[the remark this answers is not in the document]
kwxm: [There is one](https://github.com/input-output-hk/plutus/blob/3ff3901e7766647a135fa8442aaeec6d98a8b4c5/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs#L1059), but it's the standard one, and if you use that it says that the default option is `DefaultFunSemanticsVariant2`.  I just want to say `1` or `2` instead of having to type all of that in, so I had to add a special function to get a sensible help message.

PR #5509, opened 2023-09-01
Marlowe validator using Data

michaelpj:
This is not ready to be merged, but it does now work. Includes:
- Some fixes to the plugin for pattern synonyms, which I will pull out separately.
- Some TH for generating "as data" pattern synonyms for a datatype declaration, which I'll probably also pull out and contribut separately since it does seem to be useful.
- A POC of using this on the Marlowe validator.

There is some definite clumsiness, although most of it is due to the pattern synonyms and having to define standalone deriving instances. But the rest is not that bad, and the performance impact is big.


plutus-benchmark/marlowe/test/semantics/f339f59bdf92495ed2b14e2e4d3705972b4dda59aa929cffe0f1ff5355db8d79.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 81685303867
-| mem: 231960658})
\ No newline at end of file
+({cpu: 100225050
+| mem: 327890})
```
effectfully, line 2: That's a 815x reduction in cpu and a 707x reduction in mem. Not %, _times_.

plutus-benchmark/marlowe/test/semantics/2797d7ac77c1b6aff8e42cf9a47fa86b1e60f22719a996871ad412cbe4de78b5.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 39583373357
-| mem: 111889350})
\ No newline at end of file
+({cpu: 106197130
+| mem: 348246})
```
effectfully, line 2: 373x -- cpu, 321x -- mem.

plutus-benchmark/marlowe/test/semantics/202d273721330b31193405101e0637202e2a0f1140211c3e3f171e26312b0220.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 194862795670
-| mem: 553697090})
\ No newline at end of file
+({cpu: 106197130
+| mem: 348246})
```
effectfully, line 2: 1835x and 1590x again.

plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Core/V1/Semantics/Types.hs
```diff
@@ -26,49 +33,18 @@
...
-  , ValueId(..)
-    -- * Error Types
-  , IntervalError(..)
-    -- * Utility Functions
-  , emptyState
-  , getAction
-  , getInputContent
-  , inBounds
-  ) where
+{-# OPTIONS_GHC -ddump-simpl -ddump-to-file -dsuppress-uniques -dsuppress-coercions -dsuppress-type-applications -dsuppress-unfoldings -dsuppress-idinfo -dppr-cols=200 -dumpdir /tmp/dumps #-}
```
zliu41, line 36: You can do `-dsuppress-all -dno-suppress-the-thing-i-want`

plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Core/V1/Semantics/Types.hs
```diff
@@ -97,12 +67,20 @@ deriving stock instance Data Ledger.ScriptHash
 deriving stock instance Data Ledger.StakingCredential
 
 
--- | A Party to a contract.
-data Party =
-    Address Network Ledger.Address  -- ^ Party identified by a network address.
-  | Role TokenName                  -- ^ Party identified by a role token name.
-  deriving stock (Generic,Haskell.Eq,Haskell.Ord,Haskell.Show,Data)
+$(asData
```
zliu41, line 70: Hmm this is not super nice syntax, and may not work very well with HLS. How about the following?

```haskell
-- | A Party to a contract.
data _Party =
    _Address Network Ledger.Address  -- ^ Party identified by a network address.
  | _Role TokenName                  -- ^ Party identified by a role token name.

-- Generates newtype Party and pattern synonyms
makeData ''_Party
```

zliu41 (commented):
I remember you mentioned that we need to implement some optimizations in order to ensure `toData` and `unsafeFromData` are properly cancelled out, but nothing is in this PR, and it already works??

plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Core/V1/Semantics/Types.hs
michaelpj: HLS won't have any more trouble with the `$()` TH quotes than the top level ones - the only difference with the latter is that the parser inserts the implicit `$()` around `makeData ''_Party`. 

And it's not like having the separate data declaration helps you otherwise: it just adds extra names that you don't want.

michaelpj: Actually, this is also a top-level splice, so you don't need the `$( )` either. It's still not lovely, but that makes it a bit better. I'm not going to go through this and change it though, you can see it in https://github.com/input-output-hk/plutus/pull/5512

michaelpj:
> I remember you mentioned that we need to implement some optimizations in order to ensure toData and unsafeFromData are properly cancelled out, but nothing is in this PR, and it already works??

The AsData PR just generates the "more optimal" code already. It's still not _very_ optimal, but not as bad. GHC generates matching functions for each pattern synonym, and then calls them in sequence. At the moment each of those calls `unsafeFromConstr` in order to get the tag and check it, which is wasteful. The argument extraction work at least isn't repeated because it's guarded by the constructor check.

zliu41:
> The AsData PR just generates the "more optimal" code already. It's still not very optimal, but not as bad.

But the pattern synonyms use `toData` and `unsafeFromData`. Are they present in the final UPLC, or are they cancelled out?

If they are cancelled out, how is it done? I thought it requires case-of-case, case-of-known-constructor etc., which we don't have.

If they are present in the final UPLC, I'd be surprised that the improvements are so big.

michaelpj:
> But the pattern synonyms use toData and unsafeFromData. Are they present in the final UPLC, or are they cancelled out?

Ah, that's different. That's for the _arguments_. The bit I was worried about was that my old approach was using those functions to convert the the _scrutinee_ back and forth to a helper type. That's all gone.

The conversion of the arguments is still there, but that's inherent to the approach that I'm taking!

bwbush:
@michaelpj, When I run the original benchmarking code, all of the Marlowe semantics validator test cases fail with `error`:

```console
$ cd plutus-benchmark

$ cabal run exe:marlowe-validators
Semantics:
  Validator hash: ca0dc81eec347965a8a7f3cc92951acda18fcad9b687ff4b27f48038
  Validator file: /extra/iohk/plutus.md/plutus-benchmark/./marlowe/exe/marlowe-semantics.plutus
  Measurements file: /extra/iohk/plutus.md/plutus-benchmark/./marlowe/exe/marlowe-semantics.tsv

Role payout:
  Validator hash: a97312a82f41ca31242004e7ed36ce40e1d49dce459cb42f25f38e7d
  Validator file: /extra/iohk/plutus.md/plutus-benchmark/./marlowe/exe/marlowe-rolepayout.plutus
  Measurements file: /extra/iohk/plutus.md/plutus-benchmark/./marlowe/exe/marlowe-rolepayout.tsv

$ head marlowe/exe/marlowe-semantics.tsv 
Validator       Script  TxId    Measured CPU    Measured Memory Reference CPU   Reference Memory        Message
Semantics       ca0dc81eec347965a8a7f3cc92951acda18fcad9b687ff4b27f48038        0000020002010200020101020201000100010001020101020201010000020102        NA      NA      974886831       3267600 ([],CekError An error has occurred:  User error:
Caused by: (unBData
              (Constr 0
                 [B #1b9af43b0eaafc42dfaefbbf4e71437af45454c7292a6b6606363741])))
Semantics       ca0dc81eec347965a8a7f3cc92951acda18fcad9b687ff4b27f48038        0001000101000000010101000001000001010101010100000001000001010000        NA      NA      1703126152      5803500 ([],CekError An error has occurred:  User error:
Caused by: (unBData
              (Constr 0
```

I did some bisecting, and observe the following:
* `mpj/marlowe-data`: 100/100 failures
* `1.12.0.0`: 46/100 failures
* `1.11.0.0`: 0/100 failures

So, any ideas what might have broken this regression test?

michaelpj:
Nope! I didn't realise that the existing tests were not checking that evaluation succeeds, that seems like a bug. We should have caught this at the time.

michaelpj:
It looks like maybe there is an issue with the script not finding the transaction output that it is looking for, perhaps something to do with hashes. I note that the setup for these tests does some cleverness with substituting hashes that I don't understand, perhaps that has gone wrong somehow?

bwbush:
I'll begin investigating this as PLT-7536, which starts today, and let you know what I discover.

michaelpj:
I did some more investigation. I had to do some hackery to deal with traces, but I've confirmed that the failing check is `checkOwnOutputConstraint`.

michaelpj:
Notes:
- Get rid of the fake implementations of `trace` and `traceError` so tracing actually works
- Set `conservative-optimisation` so we don't optimise traces away
- Set up some evaluation tests
- Fix the evaluation test machinery to actually show logs if it fails
- Modify the trace in `checkState` to not be "o" since that's duplicated
- Observe that "o" is the last trace

That let me narrow it down to `checkOwnOutputConstraint`

michaelpj:
Ah, also just to note that this investigation was done on master, not this branch, so as to minimize the amount of difference we're looking at!

bwbush:
The source of the validation failures is a change in the strictness of `getContinuingOutputs :: TxOut` (which is in a `where` clause):
1. Up to `plutus-1.11.0.0`, it was not unconditionally evaluated.
2. Starting with `plutus-1.12.0.0`, it is unconditionally evaluated, resulting in the Marlowe semantics validator failing because `getContinuingOutputs` is evaluated even though `getContinuingOutputs` is not used on the execution path where it would call `traceError`. 

Specifically, `getContinuingOutputs` (called by `checkOwnOutputConstraint`) is never called on the `Close` branch of the following pattern match, but the UPLC evaluates it nonetheless:
https://github.com/input-output-hk/plutus/blob/7a918a315a6c3f08f6a85d607b9e5cd2643f4bf4/plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Scripts/Semantics.hs#L197-L211

The following fix (cf. https://github.com/input-output-hk/plutus/compare/mpj/fix-marlowe...bwbush:plutus:fix-marlowe) provisionally solves this problem.
```diff
     checkOwnOutputConstraint ocDatum ocValue =
         let hsh = findDatumHash' ocDatum
         in traceIfFalse "d" -- "Output constraint"
-        $ checkScriptOutput (==) ownAddress hsh ocValue getContinuingOutput
+        $ checkScriptOutput (==) ownAddress hsh ocValue (getContinuingOutput ())
 
-    getContinuingOutput :: TxOut
-    getContinuingOutput = case filter (\TxOut{txOutAddress} -> ownAddress == txOutAddress) allOutputs of
+    getContinuingOutput :: () -> TxOut
+    getContinuingOutput _ = case filter (\TxOut{txOutAddress} -> ownAddress == txOutAddress) allOutputs of
         [out] -> out
         _     -> traceError "o" -- No continuation or multiple Marlowe contract outputs is forbidden.
```

This failure does not have anything to do with the `AsData` that is the subject of this PR.

So, the compilation result of `getContinuingOutput` has changed between `1.11.0.0` and `1.12.0.0`. This change could affect other validators.

bwbush:
Here are histograms showing memory and cpu reduction for the 100 Marlowe benchmark cases.

marlowe-semantics-cpu-histogram
marlowe-semantics-memory-histogram

michaelpj:
Alright, good stuff. So after having rebased on master, it seems that this PR _is_ making things fail, and so we can't trust the numbers yet. Time to find the bug!

michaelpj:
Okay, with https://github.com/input-output-hk/plutus/pull/5542 and some tweaks we're back in action. I'm going to try and get this over the line now.

plutus-benchmark/marlowe/test/semantics/f339f59bdf92495ed2b14e2e4d3705972b4dda59aa929cffe0f1ff5355db8d79.budget.golden
michaelpj: That's what happens when you fail immediately :joy:

michaelpj:
This is much more mixed now, with some actual regressions. I'm going to investigate some more.

michaelpj:
Okay, I think this is good enough for now.

PR #5510, opened 2023-09-01
Support pattern synonyms

michaelpj:
This requires very little from us (thanks GHC!). GHC generates matching/building functions for the pattern synonyms, we can just compile those as they are.

The only wrinkle is that the functions it generates are runtime-rep polymorphic in a way we haven't dealt with before, so we need to touch up our support for that a little.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
@@ -379,49 +379,27 @@ the plugin compilation mode, so we have a special function that's not a builtin 
 just get turned into a function in PLC).
 -}
 
-{- Note [Unboxed tuples]
-This note describes the support of unboxed tuples which are different from boxed tuples.
-The difference between boxed and unboxed types is available in GHC manual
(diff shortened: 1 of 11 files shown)


plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -379,49 +379,27 @@ the plugin compilation mode, so we have a special function that's not a builtin
...
-- using 'dropRuntimeRepArgs' in 'compileType' and 'getMatchInstantiated'
-to handle the initial runtime rep arguments in a 'TyCon' application  of `(#,#)';
-
-- using 'dropRuntimeRepVars' in 'compileTyCon' to ignore 'RuntimeRep' type variables
-and to compile the kind of '(#,#)' properly.
-
-3. 'compileExpr' uses 'isRuntimeRepKindedTy' to match on type application and to ignore
-'RuntimeRep' type arguments.
-
+{- Note [Runtime reps]
```
zliu41, line 382: Much more readable than the original note!

zliu41 (approved):

bezirg (approved):

PR #5512, opened 2023-09-01
AsData: TH for backing datatypes by Data

michaelpj:
The doc comment explains what it does. 

Depends on https://github.com/input-output-hk/plutus/pull/5510


plutus-tx-plugin/test/IsData/9.6/matchAsData.pir-readable.golden
```diff
@@ -0,0 +1,209 @@
+let
+  data (Maybe :: * -> *) a | Maybe_match where
```
zliu41, line 2: I guess this (and other datatypes in PIR) will go away once we apply the same approach to `PlutusTx.Maybe`? Is that the plan?

plutus-tx/src/PlutusTx/AsData.hs
```diff
@@ -0,0 +1,148 @@
...
+
+  -- Other stuff is data families and so on
+  unless (dVariant == TH.Datatype) $ fail $ "asData: can't handle datatype variant " ++ show dVariant
+  -- a fresh name for the new datatype, but same lexically as the old one
+  cname <- TH.newName (show name)
+  -- The newtype declaration
+  let ntD =
+        let con = TH.NormalC cname [(TH.Bang TH.NoSourceUnpackedness TH.NoSourceStrictness, TH.ConT ''BuiltinData)]
+            ntDeriv ctx = TH.DerivClause (Just TH.NewtypeStrategy) ctx
+            deriv = ntDeriv [TH.ConT ''ToData, TH.ConT ''FromData, TH.ConT ''UnsafeFromData]
```
zliu41, line 90: Why not derive other instances, particularly `Eq`, automatically?

plutus-tx-plugin/test/IsData/Spec.hs
```diff
@@ -67,6 +70,23 @@ isDataRoundtrip a =
         unsafeRoundtrip = IsData.unsafeFromBuiltinData d P.== a
     in safeRoundtrip && unsafeRoundtrip
 
+AsData.asData [d|
+  data SecretlyData = FirstC () | SecondC Integer
+  |]
+
+AsData.asData [d|
+  data RecordConstructor a = RecordConstructor { x :: a, y :: Integer }
+  |]
```
zliu41, line 79: I'd be tempted to add a test case converting between these two types, plus equality comparison, say

```haskell
dataToData :: CompiledCode (RecordConstructor Integer -> SecretlyData)
dataToData = plc (Proxy @"dataToData")
  (\(x@(RecordConstructor a b)) ->
    if x == RecordConstructor 3 4
      then SecondC (Builtins.addInteger a b)
      else FirstC ())
```

plutus-tx-plugin/test/IsData/Spec.hs
```diff
@@ -67,6 +70,23 @@ isDataRoundtrip a =
         unsafeRoundtrip = IsData.unsafeFromBuiltinData d P.== a
     in safeRoundtrip && unsafeRoundtrip
 
+AsData.asData [d|
+  data SecretlyData = FirstC () | SecondC Integer
+  |]
+
+AsData.asData [d|
+  data RecordConstructor a = RecordConstructor { x :: a, y :: Integer }
```
zliu41, line 78: There should be a test case testing the field accessors, `x` and `y`.

plutus-tx/src/PlutusTx/AsData.hs
```diff
@@ -0,0 +1,148 @@
...
+    fieldNames <- case cVariant of
+      TH.RecordConstructor names -> pure names
+      -- otherwise whatever
+      _                          -> ifor fields $ \fieldIx _ -> TH.newName $ "arg" ++ show fieldIx
+    let extractFieldNames = fieldNames
+    createFieldNames <- for fieldNames (TH.newName . show)
+    patSynArgs <- case cVariant of
+      TH.NormalConstructor   -> pure $ TH.PrefixPatSyn extractFieldNames
+      TH.RecordConstructor _ -> pure $ TH.RecordPatSyn extractFieldNames
+      TH.InfixConstructor    -> fail "asData: Can't handle infix data constructors"
```
zliu41, line 106: TH.InfixPatSyn?

plutus-tx-plugin/test/IsData/Spec.hs
```diff
@@ -89,4 +109,7 @@ tests = testNestedGhc "IsData" [
     , goldenUEval "bytestring" [plc (Proxy @"bytestring") (isDataRoundtrip (WrappedBS Builtins.emptyByteString))]
     , goldenPir "deconstructData" deconstructData
     , goldenPir "unsafeDeconstructData" unsafeDeconstructData
+    , goldenPirReadable "matchAsData" matchAsData
```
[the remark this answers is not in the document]
michaelpj: I am, though. If we care about what codegen is doing (which I do), then we need to see what codgen produces!

Anyway, I can add versions of these to the `Optimization` tests which do test that.

plutus-tx-plugin/test/IsData/9.6/matchAsData.pir-readable.golden
michaelpj: I'd be very wary of applying this approach to Maybe, that would make it no longer a generically usable datatype. We could define a separate type for the Data version, I guess. Then we'd really need an entirely different library with copies of almost everything...

Also, since normal datatypes and AsData datatypes can interoperate, it's not necessarily a problem. If you have a field of type `Maybe X` in an AsData type where X is also an AsData type, you just add one level of unwrapping and wrapping when dealing with it, which isn't too bad.

plutus-tx/src/PlutusTx/AsData.hs
michaelpj: Just not being opinionated. For example, sometimes people don't _want_ an `Eq` instance. The job of this function is to define the datatype, users should still have full control over what instances they want or don't want.

I was going to say that the To/FromData instances are necessary, but that's not true. They're only necessary if you use the type in another AsData type, which you might not do. So we shouldn't provide those either.

plutus-tx-plugin/test/IsData/Spec.hs
michaelpj: Yes, good idea.

plutus-tx/src/PlutusTx/AsData.hs
michaelpj: To some degree I was just being lazy: I can't remember the last time I saw an infix data constructor, and it's not quite as simple (since it always has exactly two binders).

plutus-tx-plugin/test/IsData/Spec.hs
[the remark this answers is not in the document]
michaelpj: I added one. Note that there _are_ no conversions between Data and SOP in the way that there were in my first version.

plutus-tx/src/PlutusTx/AsData.hs
michaelpj: I added it anyway.

plutus-tx-plugin/test/IsData/Spec.hs
michaelpj: added

PR #5515, opened 2023-09-01
1.12.0.0

zliu41:


bezirg:
The PR looks fine, but can we stall it until 1.11.0.0 is integrated to  `cardano-ledger` ?

michaelpj:
I think it's fine - we can either finish the 1.11 PR and then the 1.12 one will be easy, or just switch the PR to integrate 1.12 instead.

bezirg:
I need a patch for 1.11  to export some extra stuff from plutus-ledger-api (for cardano-ledger integration). Shall we stick this patch instead here?

michaelpj:
Yes, let's just merge it to master and include it here. Or: we can work on what we need using a s-r-p of plutus, and then when we have it all working we just include the changes in the next release and then finally do the upgrade?

PR #5518, opened 2023-09-04
PLT-7110: Add costModelParamsForTesting for all versions

bezirg:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/CekMachineCosts.hs
```diff
@@ -27,32 +34,48 @@ cekMachineCostsPrefix = "cek"
...
-    deriving (FromJSON, ToJSON) via CustomJSON '[FieldLabelModifier LowerIntialCharacter] CekMachineCosts
+    deriving stock (Generic)
+    deriving anyclass (FunctorB, TraversableB, ConstraintsB)
+
+deriving via CustomJSON '[FieldLabelModifier LowerIntialCharacter]
+             (CekMachineCostsBase Identity) instance ToJSON (CekMachineCostsBase Identity)
+deriving via CustomJSON '[FieldLabelModifier LowerIntialCharacter]
+             (CekMachineCostsBase Identity) instance FromJSON (CekMachineCostsBase Identity)
+
+-- This instance will emit the generation of JSON for Nothing fields,
```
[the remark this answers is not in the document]
bezirg: thanks @kwxm, fixed

bezirg:
I think I should actually add some tests and some comments, especially about this reordering thing

bezirg:
I added tests and I think it is ready for review

kwxm:
[I'm just checking the consistency of the benchmarks].

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
```diff
@@ -1,29 +1,119 @@
...
+import PlutusLedgerApi.V3 qualified as V3
+
+import Barbies
+import Data.Aeson
+import Data.Functor.Identity
+import Data.Map qualified as Map
 import Data.Maybe
 
-{-| only for testing purposes: The default, raw cost model params.
+-- | Example values of costs for plutus-v3 , in expected ledger order, i.e. can be used in testing
```
michaelpj, line 28:
```suggestion
-- | Example values of costs for @PlutusV3@ , in expected ledger order. Suitable to be used in testing.
```
etc

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
```diff
@@ -1,29 +1,119 @@
...
+v1_costModelParamsForTesting = Map.toList $ fromJust $ extractCostModelParamsLedgerOrder v1_liftedCostModel
+
+v3_liftedCostModel, v2_liftedCostModel, v1_liftedCostModel
+    :: CostModel (CekMachineCostsBase Maybe)  (BuiltinCostModelBase MCostingFun)
+v3_liftedCostModel = liftToJust defaultCekCostModel
+v2_liftedCostModel = v3_liftedCostModel
+                    & machineCostModel
+                    %~ clearV3machineCostModel
+                    & builtinCostModel
+                    %~ clearV3builtinCostModel
```
michaelpj, line 50: nice

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
```diff
@@ -1,29 +1,119 @@
...
+v1_liftedCostModel = v2_liftedCostModel
+                    & builtinCostModel
+                    %~ clearV2builtinCostModel
+
+-- Assign to `mempty` those builtins that the specific plutus version X introduces, and thus
+-- are missing from version X-1.
+-- Calling this function on a vX_BuiltinCostModel will turn it into a vX-1_BuiltinCostModel,
+-- by omitting the generation of costs concerning the missing builtins.
```
michaelpj, line 58: This is kind of ugly. We already list the param names that we know about! Can we not do something with them?

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
```diff
@@ -1,29 +1,119 @@
...
-extracted default cost model parameters (resulting in no change to params).
-An alternative implementation would be to pass an empty map of params.
+-- A variant of `extractCostModelParams` that returns the `CostModelParams` not in alphabetical order,
+-- but in the `ParamName` order, i.e. the order expected by the ledger
+extractCostModelParamsLedgerOrder :: (Common.IsParamName p, Ord p, ToJSON machinecosts, ToJSON builtincosts)
+                     => CostModel machinecosts builtincosts
+                     -> Maybe (Map.Map p Integer)
+extractCostModelParamsLedgerOrder cm =
+    let mapInAlphaOrder = extractCostModelParams cm
+        mapInLedgerOrder = Map.mapKeys (fromJust . readParamName) <$> mapInAlphaOrder
```
michaelpj, line 107: the fromJust seems bad, we're already returning a maybe, can't we fold that in?

plutus-ledger-api/src/PlutusLedgerApi/V1/ParamName.hs
```diff
@@ -181,6 +181,7 @@ data ParamName =
   | VerifyEd25519Signature'cpu'arguments'intercept
   | VerifyEd25519Signature'cpu'arguments'slope
   | VerifyEd25519Signature'memory'arguments
-    deriving stock (Eq, Ord, Enum, Ix, Bounded, Generic)
+    -- NB: Show.show != IsParamName.showParamName
```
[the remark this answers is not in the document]
bezirg: I originally didn't have the Show instance because of this potential confusion, but then I added hunit tests which want to show values when comparing (@=?) . I will fix it with using assertEqual instead.

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
bezirg: ok good idea

bezirg: I hope I can find a way to fold this, because `mapKeys` interface seems to be pure

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
bezirg: > This is kind of ugly. 
Yes it is certainly duplication.

>We already list the param names that we know about! Can we not do something with them?

You mean take advantage of this?

https://github.com/input-output-hk/plutus/blob/master/plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs#L83-L114

michaelpj: No, I mean this! https://github.com/input-output-hk/plutus/blob/master/plutus-ledger-api/src/PlutusLedgerApi/V1/ParamName.hs

bezirg: But then I have to write all the intercept and number of ModelThreeArguments, etc

bezirg: Hmmm, ok let me think about it a bit more

michaelpj: Oh I see, those are for the flat version and this is the structured version... tricky :thinking:

plutus-ledger-api/test/Spec.hs
kwxm, line 1: Could we maybe have a comment in here about what it's actually testing?  I don't look at this file very often, but when I do I have to work out what's going on.  For example my first thought here is to wonder why we've only got `v1_evalCtxForTesting` and not V2 and V3 versions too.

plutus-ledger-api/test/Spec/CostModelParams.hs
```diff
@@ -25,39 +26,43 @@ tests =
...
+    , testCase "default values costmodelparamsfortesting" $ do
+            defaultCostModelParams @=? Just (toCostModelParams v3_costModelParamsForTesting)
     , testCase "context length" $ do
-            let costValuesForTesting = Map.elems costModelParamsForTesting
+            let costValuesForTesting = fmap snd v3_costModelParamsForTesting
             -- the `costModelParamsForTesting` reflects only the latest version V3, so this should succeed because the lengths match
             assertBool "wrong number of arguments in V3.mkEvaluationContext" $ isRight $ runExcept $ runWriterT $ V3.mkEvaluationContext costValuesForTesting
             -- this one should succeed because we pass more params
-            assertBool "larger number of params did not warn" $ hasWarnMoreParams (length v3CostModelParamNames) (length v3CostModelParamNames+1) $
+            assertBool "larger number of params did not warn" $ hasWarnMoreParams (length v3_ParamNames) (1 + length v3_ParamNames) $
```
kwxm, line 49: I've forgotten why we allow this.  Is it written down somewhere and if so could we have a reference here?

plutus-ledger-api/test/Spec/CostModelParams.hs
```diff
@@ -25,39 +26,43 @@ tests =
...
                 runExcept $ runWriterT $ V3.mkEvaluationContext $ costValuesForTesting ++ [1] -- dummy param value appended
     , testCase "cost model parameters" $ do
          -- v1 is missing some cost model parameters because new builtins are added in v2
-         assertBool "v1 params is proper subset of v2 params" $ Set.fromList v1CostModelParamNames `Set.isProperSubsetOf` Set.fromList v2CostModelParamNames
+         assertBool "v1 params is not proper subset of v2 params" $ v1_ParamNames `paramProperSubset` v2_ParamNames
```
kwxm, line 53:
```suggestion
         assertBool "v1 params is not a proper subset of v2 params" $ v1_ParamNames `paramProperSubset` v2_ParamNames
```

plutus-ledger-api/test/Spec/CostModelParams.hs
```diff
@@ -25,39 +26,43 @@ tests =
...
+            let costValuesForTesting = fmap snd v3_costModelParamsForTesting
             -- the `costModelParamsForTesting` reflects only the latest version V3, so this should succeed because the lengths match
             assertBool "wrong number of arguments in V3.mkEvaluationContext" $ isRight $ runExcept $ runWriterT $ V3.mkEvaluationContext costValuesForTesting
             -- this one should succeed because we pass more params
-            assertBool "larger number of params did not warn" $ hasWarnMoreParams (length v3CostModelParamNames) (length v3CostModelParamNames+1) $
+            assertBool "larger number of params did not warn" $ hasWarnMoreParams (length v3_ParamNames) (1 + length v3_ParamNames) $
                 runExcept $ runWriterT $ V3.mkEvaluationContext $ costValuesForTesting ++ [1] -- dummy param value appended
     , testCase "cost model parameters" $ do
          -- v1 is missing some cost model parameters because new builtins are added in v2
-         assertBool "v1 params is proper subset of v2 params" $ Set.fromList v1CostModelParamNames `Set.isProperSubsetOf` Set.fromList v2CostModelParamNames
```
kwxm, line 45: This catches me out too.

plutus-ledger-api/test/Spec/CostModelParams.hs
kwxm: > Is it written down somewhere

Ah, I found it in `Note [Cost model parameters from the ledger's point of view]`.

plutus-ledger-api/test/Spec/Eval.hs
```diff
@@ -9,13 +9,14 @@ import PlutusCore.MkPlc
 import PlutusCore.StdLib.Data.Unit
 import PlutusCore.Version as PLC
 import PlutusLedgerApi.Common.Versions
-import PlutusLedgerApi.Test.EvaluationContext (evalCtxForTesting)
+import PlutusLedgerApi.Test.EvaluationContext
```
kwxm, line 12: You could still say what you're importing here (I think just `v1_costModelParamsForTesting`). I find that quite helpful (I won't demand that you do that though).

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
```diff
@@ -1,29 +1,119 @@
...
 
-{-| only for testing purposes: The default, raw cost model params.
+-- | Example values of costs for plutus-v3 , in expected ledger order, i.e. can be used in testing
+-- `V3.mkEvaluationContext . fmap snd`
+v3_costModelParamsForTesting :: [(V3.ParamName, Integer)]
+v3_costModelParamsForTesting = Map.toList $ fromJust $ extractCostModelParamsLedgerOrder v3_liftedCostModel
+
+-- | Example values of costs for plutus-v2, in expected ledger order, i.e. can be used in testing
+-- `V2.mkEvaluationContext . fmap snd`
+v2_costModelParamsForTesting :: [(V2.ParamName,Integer)]
```
kwxm, line 35:
```suggestion
v2_costModelParamsForTesting :: [(V2.ParamName, Integer)]
```

Inconsistent spacing.

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
```diff
@@ -1,29 +1,119 @@
...
+v2_costModelParamsForTesting = Map.toList $ fromJust $ extractCostModelParamsLedgerOrder v2_liftedCostModel
+
+-- | Example values of costs for plutus-v1, *in expected ledger order*, i.e. can be used in testing
+-- `V1.mkEvaluationContext . fmap snd`
+v1_costModelParamsForTesting :: [(V1.ParamName,Integer)]
```
kwxm, line 40:
```suggestion
v1_costModelParamsForTesting :: [(V1.ParamName, Integer)]
```

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
kwxm: Would they be unsuitable if they were in the wrong order?

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
```diff
@@ -1,29 +1,119 @@
...
+v1_costModelParamsForTesting = Map.toList $ fromJust $ extractCostModelParamsLedgerOrder v1_liftedCostModel
+
+v3_liftedCostModel, v2_liftedCostModel, v1_liftedCostModel
```
kwxm, line 43: !!!  I had no idea you were allowed to do this.

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
```diff
@@ -1,29 +1,119 @@
...
+               , paramKeccak_256 = mempty
+               , paramBlake2b_224 = mempty
+               }
+clearV2builtinCostModel r = r
+               { paramSerialiseData = mempty
+               , paramVerifyEcdsaSecp256k1Signature = mempty
+               , paramVerifySchnorrSecp256k1Signature = mempty
+               }
+
+-- Assign to `mempty` those CEK construct that the specific plutus version X introduces, and thus
```
kwxm, line 88: This is the ledger language `PlutusV<X>`, right?  `Constr` and `Case` appear in Plutus Core 1.0.1.  I guess that in fact the machine parameters depend on the Plutus Core language version and the buitlin parameters depend on the ledger language version, which is a pain (just like everything to do with these versions).

kwxm (approved):
This looks fine to me.  Most of my comments were just minor things.

plutus-ledger-api/changelog.d/20230905_103632_bezirg_cmparams_testing.md
```diff
@@ -0,0 +1,13 @@
+### Added
+
+- costModelParamsForTesting for all plutus versions (v1,v2,v3)
```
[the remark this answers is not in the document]
bezirg: Indeed there is no V1,V2,V3 only PlutusV* , but our module scheme seems to suggest that. I will perhaps split this module into 3 parts to reflect that V* module scheme

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
bezirg: You are semantically correct, but I see it more like "plutus version X introduces ledger language version Y which introduces those CEK machine constructs"

plutus-ledger-api/testlib/PlutusLedgerApi/Test/EvaluationContext.hs
kwxm: >  It is not super bad, but it is not "representable" of the costs.

Right, so they _could_ just be random numbers if you only want to check that this code does what you expect, but using the correct numbers doesn't do any harm.  I just wondered if it was doing something that I hadn't spotted that needed the actual values.

michaelpj:
Error: cabal: Unrecognised target syntax for ' plutus-benchmark:validation'.

Somehow it's including the space before the benchmark name??? Or something? I don't really know.

michaelpj:
Okay, I see the problem. @zeme-iohk your PR added some quoting of shell variables. Previously the use of `$BENCHMARK` wasn't quoted so the shell would have split on the whitespace, but now we're passing the value explicitly with the leading space. Quoting variable usage is good, so I think the fix is to extract the benchmark name without the leading space.

bezirg:
> Okay, I see the problem. @zeme-iohk your PR added some quoting of shell variables. Previously the use of `$BENCHMARK` wasn't quoted so the shell would have split on the whitespace, but now we're passing the value explicitly with the leading space. Quoting variable usage is good, so I think the fix is to extract the benchmark name without the leading space.

Thanks Michael

PR #5523, opened 2023-09-05
Add Documentation on GHC Extensions, Flags and Pragmas

zliu41:
The `Strict` extension section can potentially be extended and probably deserves its own page. That is future work.

### DIFF
--- a/doc/read-the-docs-site/extensions-flags-pragmas.rst
+++ b/doc/read-the-docs-site/extensions-flags-pragmas.rst
@@ -75,7 +75,7 @@ The most reliable solution, however, is to simply turn these optimizations off.
 Another option is to bump ``-funfolding-creation-threshold`` to make it more likely for GHC to retain unfoldings for functions without the ``INLINEABLE`` pragma.
 ``-fexpose-all-unfoldings`` should be used as a last resort.
 
-There's also the ``INLINE`` pragma, which does more than the ``INLINEABLE`` in that the ``INLINE`` pragma forces the Plutus IR optimizer to inline the function.
+There's also the ``INLINE`` pragma, which does more than ``INLINEABLE`` in that the ``INLINE`` pragma forces the Plutus IR optimizer to inline the function.
 This is not currently useful, and should be avoided.
 In the future, we plan to make the ``INLINE`` pragma part of the mechanism for making certain functions non-strict in their arguments.
 This is not yet implemented.
--- a/doc/read-the-docs-site/extensions-flags-pragmas.rst
+++ b/doc/read-the-docs-site/extensions-flags-pragmas.rst
@@ -40,12 +40,11 @@ GHC has a variety of optimization flags, many of which are on by default.
 Although Plutus Tx is, syntactically, a subset of Haskell, it has different semantics and a different evaluation strategy (Haskell: non-strict semantics, call by need; Plutus Tx: strict semantics, call by value).
 As a result, some GHC optimizations are not helpful for Plutus Tx programs, and can even be harmful, in the sense that it can make Plutus Tx programs less efficient, or fail to be compiled.
 An example is the full laziness optimization, controlled by GHC flag ``-ffull-laziness``, which floats let bindings out of lambdas whenever possible.
-Since Untyped Plutus Core does not employ lazy evaluation, the full laziness optimization is never beneficial, and can sometimes make a Plutus Tx program more expensive.
+Since Untyped Plutus Core does not employ lazy evaluation, the full laziness optimization is usually not beneficial, and can sometimes make a Plutus Tx program more expensive.
 Conversely, some GHC features must be turned on in order to ensure Plutus Tx programs are compiled successfully.
 
 All Plutus Tx modules should use the following GHC flags: ::
 
-  -fplugin PlutusTx.Plugin
   -fno-ignore-interface-pragmas
   -fno-omit-interface-pragmas
   -fno-full-laziness
@@ -55,9 +54,17 @@ All Plutus Tx modules should use the following GHC flags: ::
   -fno-unbox-strict-fields
   -fno-unbox-small-strict-fields
 
-``-fplugin PlutusTx.Plugin`` enables the Plutus Tx compiler.
 ``-fno-ignore-interface-pragmas`` and ``-fno-omit-interface-pragmas`` ensure unfoldings of Plutus Tx functions are available.
-The rest are GHC optimizations that are bad for Plutus Tx, and should thus be turned off.
+The rest are GHC optimizations that are generally bad for Plutus Tx, and should thus be turned off.
+
+These flags can be specified either in a Haskell module, e.g.: ::
+
+  {-# OPTIONS_GHC -fno-ignore-interface-pragmas #-}
+
+or in a build file. E.g., if your project is built using Cabal, you can add the flags to the ``.cabal`` files, like so:
+
+  ghc-options:
+    -fno-ignore-interface-pragmas
 
 Note that this section only covers GHC flags, not Plutus Tx compiler flags.
 Information about the latter can be found in :ref:`plutus_tx_options`.
@@ -76,8 +83,3 @@ Therefore, ``INLINEABLE`` is preferred over ``-fexpose-all-unfoldings`` even tho
 The most reliable solution, however, is to simply turn these optimizations off.
 Another option is to bump ``-funfolding-creation-threshold`` to make it more likely for GHC to retain unfoldings for functions without the ``INLINEABLE`` pragma.
 ``-fexpose-all-unfoldings`` should be used as a last resort.
-
-There's also the ``INLINE`` pragma, which does more than ``INLINEABLE`` in that the ``INLINE`` pragma forces the Plutus IR optimizer to inline the function.
-This is not currently useful, and should be avoided.
-In the future, we plan to make the ``INLINE`` pragma part of the mechanism for making certain functions non-strict in their arguments.
-This is not yet implemented.
--- a/doc/read-the-docs-site/extensions-flags-pragmas.rst
+++ b/doc/read-the-docs-site/extensions-flags-pragmas.rst
@@ -1,3 +1,5 @@
+.. _extensions_flags_pragmas:
+
 GHC Extensions, Flags and Pragmas
 =================================
 
--- a/doc/read-the-docs-site/extensions-flags-pragmas.rst
+++ b/doc/read-the-docs-site/extensions-flags-pragmas.rst
@@ -17,7 +17,7 @@ Plutus Tx modules should use the ``Strict`` extension: ::
 
 Unlike in Haskell, function applications in Plutus Tx are strict.
 In other words, when evaluating ``(\x -> 42) (3 + 4)`` the expression ``3 + 4`` is evaluated first, before evaluating the function body (``42``), even though ``x`` is not used in the function body.
-The ``Strict`` extension ensures that let bindings are also (by default) strict, for instance, evaluating
+The ``Strict`` extension ensures that let bindings and patterns are also (by default) strict, for instance, evaluating
 ``let x = 3 + 4 in 42`` evaluates ``3 + 4`` first, even though ``x`` is not used.
 
 Bang patterns and lazy patterns can be used to explicitly specify whether a let binding is strict or non-strict, as in ``let !x = 3 + 4 in 42`` (strict) and ``let ~x = 3 + 4 in 42`` (non-strict).
--- a/doc/read-the-docs-site/extensions-flags-pragmas.rst
+++ b/doc/read-the-docs-site/extensions-flags-pragmas.rst
@@ -1,0 +1,81 @@
+GHC Extensions, Flags and Pragmas
+=================================
+
+Plutus Tx is a subset of Haskell and is compiled to Untyped Plutus Core by the Plutus Tx compiler, a GHC (Glasgow Haskell Compiler) plugin.
+
+In order to ensure the success and correct compilation of Plutus Tx programs, all Plutus Tx modules (that is, Haskell modules that contain code to be compiled by the Plutus Tx compiler) should use the following GHC extensions, flags and pragmas.
+
+
+Extensions
+-------------------------------------------------
+
+Plutus Tx modules should use the ``Strict`` extension: ::
+
+  {-# LANGUAGE Strict #-}
+
+Unlike in Haskell, function applications in Plutus Tx are strict.
+In other words, when evaluating ``(\x -> 42) (3 + 4)`` the expression ``3 + 4`` is evaluated first, before evaluating the function body (``42``), even though ``x`` is not used in the function body.
+The ``Strict`` extension ensures that let bindings are also (by default) strict, for instance, evaluating
+``let x = 3 + 4 in 42`` evaluates ``3 + 4`` first, even though ``x`` is not used.
+
+Bang patterns and lazy patterns can be used to explicitly specify whether a let binding is strict or non-strict, as in ``let !x = 3 + 4 in 42`` (strict) and ``let ~x = 3 + 4 in 42`` (non-strict).
+At this time, it is not possible to make function applications non-strict: ``(\(~x) -> 42) (3 + 4)`` still evaluates ``3 + 4`` strictly.
+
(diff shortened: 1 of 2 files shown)


doc/read-the-docs-site/extensions-flags-pragmas.rst
```diff
@@ -0,0 +1,83 @@
...
+  On the other hand, if ``x`` is strict, it is always evaluated once, which is at most one more time than what is necessary.
+
+Flags
+-------------------------------------------------
+
+GHC has a variety of optimization flags, many of which are on by default.
+Although Plutus Tx is, syntactically, a subset of Haskell, it has different semantics and a different evaluation strategy (Haskell: non-strict semantics, call by need; Plutus Tx: strict semantics, call by value).
+As a result, some GHC optimizations are not helpful for Plutus Tx programs, and can even be harmful, in the sense that it can make Plutus Tx programs less efficient, or fail to be compiled.
+An example is the full laziness optimization, controlled by GHC flag ``-ffull-laziness``, which floats let bindings out of lambdas whenever possible.
+Since Untyped Plutus Core does not employ lazy evaluation, the full laziness optimization is never beneficial, and can sometimes make a Plutus Tx program more expensive.
```
effectfully, line 43: > the full laziness optimization is never beneficial

How come? What about transforming, say,

```haskell
map (\x -> let y = <expensive> in x + y) xs
```

into

```haskell
let y = <expensive> in map (\x -> x + y) xs
```

assuming the `let`  is non-strict and  `xs` is non-empty?

doc/read-the-docs-site/extensions-flags-pragmas.rst
```diff
@@ -0,0 +1,83 @@
...
+  -fno-strictness
+  -fno-unbox-strict-fields
+  -fno-unbox-small-strict-fields
+
+``-fplugin PlutusTx.Plugin`` enables the Plutus Tx compiler.
+``-fno-ignore-interface-pragmas`` and ``-fno-omit-interface-pragmas`` ensure unfoldings of Plutus Tx functions are available.
+The rest are GHC optimizations that are bad for Plutus Tx, and should thus be turned off.
+
+Note that this section only covers GHC flags, not Plutus Tx compiler flags.
+Information about the latter can be found in :ref:`plutus_tx_options`.
```
effectfully, line 63: Clicking on the link from the rendered file doesn't lead me anywhere.

doc/read-the-docs-site/extensions-flags-pragmas.rst
```diff
@@ -0,0 +1,83 @@
...
+
+The ``-fexpose-all-unfoldings`` flag also makes GHC expose all unfoldings, but unfoldings exposed this way can be more optimized than unfoldings exposed via ``INLINEABLE``.
+In general we do not want GHC to perform optimizations, since GHC optimizes a program based on the assumption that it has non-strict semantics and is evaluated lazily (call by need), which is not true for Plutus Tx programs.
+Therefore, ``INLINEABLE`` is preferred over ``-fexpose-all-unfoldings`` even though the latter is simpler.
+
+``-fexpose-all-unfoldings`` can be useful for functions that are generated by GHC and do not have the ``INLINEABLE`` pragma.
+``-fspecialise`` and ``-fspec-constr`` are two examples of optimizations that can generate such functions.
+The most reliable solution, however, is to simply turn these optimizations off.
+Another option is to bump ``-funfolding-creation-threshold`` to make it more likely for GHC to retain unfoldings for functions without the ``INLINEABLE`` pragma.
+``-fexpose-all-unfoldings`` should be used as a last resort.
```
effectfully, line 78: I'm not sure if I'm doing something wrong, but the rendered file does not render those as a list, all the bullet points are lumped into a single paragraph.

doc/read-the-docs-site/extensions-flags-pragmas.rst
```diff
@@ -0,0 +1,83 @@
...
+
+There's also the ``INLINE`` pragma, which does more than ``INLINEABLE`` in that the ``INLINE`` pragma forces the Plutus IR optimizer to inline the function.
```
effectfully, line 80: Does it? The Plutus IR one? Even on the Haskell side `INLINE` may not cause the definition to get inlined without an additional call to `inline` or at all (sometimes for a good reason like the definition being recursive).

doc/read-the-docs-site/extensions-flags-pragmas.rst
```diff
@@ -0,0 +1,83 @@
...
+This is not currently useful, and should be avoided.
+In the future, we plan to make the ``INLINE`` pragma part of the mechanism for making certain functions non-strict in their arguments.
```
effectfully, line 82: Sounds rather weird and unreliable.

doc/read-the-docs-site/extensions-flags-pragmas.rst
```diff
@@ -0,0 +1,83 @@
...
+
+  -fplugin PlutusTx.Plugin
+  -fno-ignore-interface-pragmas
+  -fno-omit-interface-pragmas
+  -fno-full-laziness
+  -fno-spec-constr
+  -fno-specialise
+  -fno-strictness
+  -fno-unbox-strict-fields
+  -fno-unbox-small-strict-fields
```
effectfully, line 56: You may want to specify how to add those (`.cabal` file vs `OPTIONS_GHC`). I know it's not the best place to do it, but it could be helpful. Or at least provide a reference.

effectfully (approved):
Great stuff, very helpful.

doc/read-the-docs-site/extensions-flags-pragmas.rst
zliu41: This is a single paragraph :joy:

doc/read-the-docs-site/extensions-flags-pragmas.rst
zliu41: This does not render properly on GitHub (although it's not too bad). To see the properly rendered text you'd need to check out the branch, run `autobuild-docs` and look at the HTML.

doc/read-the-docs-site/extensions-flags-pragmas.rst
zliu41: Unless I'm missing something, `<expensive>` is evaluated the same number of times in both cases. Moving a non-strict binding never matters.

doc/read-the-docs-site/extensions-flags-pragmas.rst
zliu41: Yes, it indeed forces the Plutus IR inliner to inline the function: https://github.com/input-output-hk/plutus/pull/5183

doc/read-the-docs-site/extensions-flags-pragmas.rst
zliu41: Sure. This will just be used for a few selected functions, like `(&&)` and `(||)`. As it stands, they are sometimes non-strict in the second argument, sometimes strict, which is very bad.

doc/read-the-docs-site/extensions-flags-pragmas.rst
effectfully: Sorry, I said the wrong thing. I mean, assuming `let` is strict, of course.

zliu41: I doubt `-ffull-laziness` moves strict bindings; it'd need to prove `xs` is non-empty. And it's called full _laziness_ after all. I guess it might be possible. But turning it off makes the test cases we have strictly better.

michaelpj: I think there is a misunderstanding here. Floating out bindings can be helpful - indeed, we have our own FloatOut pass. The problem with `full-laziness` is that specifically in GHC it does the reverse binder-swap transformation, which can lead to us re-evaluating things. _That_ is the thing that is bad for us. If it didn't do that, it would probably be fine and maybe even beneficial.

doc/read-the-docs-site/extensions-flags-pragmas.rst
```diff
@@ -0,0 +1,83 @@
...
+Extensions
+-------------------------------------------------
+
+Plutus Tx modules should use the ``Strict`` extension: ::
+
+  {-# LANGUAGE Strict #-}
+
+Unlike in Haskell, function applications in Plutus Tx are strict.
+In other words, when evaluating ``(\x -> 42) (3 + 4)`` the expression ``3 + 4`` is evaluated first, before evaluating the function body (``42``), even though ``x`` is not used in the function body.
+The ``Strict`` extension ensures that let bindings are also (by default) strict, for instance, evaluating
```
michaelpj, line 20:
```suggestion
The ``Strict`` extension ensures that let bindings and patterns are also (by default) strict, for instance, evaluating
```

doc/read-the-docs-site/extensions-flags-pragmas.rst
```diff
@@ -0,0 +1,83 @@
...
+GHC has a variety of optimization flags, many of which are on by default.
+Although Plutus Tx is, syntactically, a subset of Haskell, it has different semantics and a different evaluation strategy (Haskell: non-strict semantics, call by need; Plutus Tx: strict semantics, call by value).
+As a result, some GHC optimizations are not helpful for Plutus Tx programs, and can even be harmful, in the sense that it can make Plutus Tx programs less efficient, or fail to be compiled.
+An example is the full laziness optimization, controlled by GHC flag ``-ffull-laziness``, which floats let bindings out of lambdas whenever possible.
+Since Untyped Plutus Core does not employ lazy evaluation, the full laziness optimization is never beneficial, and can sometimes make a Plutus Tx program more expensive.
+Conversely, some GHC features must be turned on in order to ensure Plutus Tx programs are compiled successfully.
+
+All Plutus Tx modules should use the following GHC flags: ::
+
+  -fplugin PlutusTx.Plugin
```
michaelpj, line 48: No, you do not need this. `compile` adds it for you.

doc/read-the-docs-site/extensions-flags-pragmas.rst
```diff
@@ -0,0 +1,83 @@
...
+  -fno-ignore-interface-pragmas
+  -fno-omit-interface-pragmas
+  -fno-full-laziness
+  -fno-spec-constr
+  -fno-specialise
+  -fno-strictness
+  -fno-unbox-strict-fields
+  -fno-unbox-small-strict-fields
+
+``-fplugin PlutusTx.Plugin`` enables the Plutus Tx compiler.
```
michaelpj, line 58: delete

doc/read-the-docs-site/extensions-flags-pragmas.rst
```diff
@@ -0,0 +1,83 @@
...
+-------------------------------------------------
+
+GHC has a variety of optimization flags, many of which are on by default.
+Although Plutus Tx is, syntactically, a subset of Haskell, it has different semantics and a different evaluation strategy (Haskell: non-strict semantics, call by need; Plutus Tx: strict semantics, call by value).
+As a result, some GHC optimizations are not helpful for Plutus Tx programs, and can even be harmful, in the sense that it can make Plutus Tx programs less efficient, or fail to be compiled.
+An example is the full laziness optimization, controlled by GHC flag ``-ffull-laziness``, which floats let bindings out of lambdas whenever possible.
+Since Untyped Plutus Core does not employ lazy evaluation, the full laziness optimization is never beneficial, and can sometimes make a Plutus Tx program more expensive.
+Conversely, some GHC features must be turned on in order to ensure Plutus Tx programs are compiled successfully.
+
+All Plutus Tx modules should use the following GHC flags: ::
```
michaelpj, line 46: "Plutus Tx module" is not well-defined at the moment. Is this "all modules that are used transitively in Plutus Tx programs" or "all modules where you use the plugin"? I think you mean the former.

doc/read-the-docs-site/extensions-flags-pragmas.rst
michaelpj: And indeed you do _not_ want this in modules that are _used_ in Plutus Tx programs but do not themselves contain `compile` calls.

doc/read-the-docs-site/extensions-flags-pragmas.rst
```diff
@@ -0,0 +1,83 @@
...
+Therefore, ``INLINEABLE`` is preferred over ``-fexpose-all-unfoldings`` even though the latter is simpler.
+
+``-fexpose-all-unfoldings`` can be useful for functions that are generated by GHC and do not have the ``INLINEABLE`` pragma.
+``-fspecialise`` and ``-fspec-constr`` are two examples of optimizations that can generate such functions.
+The most reliable solution, however, is to simply turn these optimizations off.
+Another option is to bump ``-funfolding-creation-threshold`` to make it more likely for GHC to retain unfoldings for functions without the ``INLINEABLE`` pragma.
+``-fexpose-all-unfoldings`` should be used as a last resort.
+
+There's also the ``INLINE`` pragma, which does more than ``INLINEABLE`` in that the ``INLINE`` pragma forces the Plutus IR optimizer to inline the function.
+This is not currently useful, and should be avoided.
```
michaelpj, line 81: Why not? It can absolutely be useful if the user knows something should be inlined - it seems pretty much just as useful as it is in Haskell.

doc/read-the-docs-site/extensions-flags-pragmas.rst
michaelpj: I would just not include this - it's a weird special case that we may use but users have no reason to know about, it's just confusing.

michaelpj (approved):
I think this should obsolete the information in `plutus-tx/README.md`. It mostly does, there is just one bit about GHCi that we might want to copy in. Then I think we should just delete `plutus-tx/README.md`: nobody knows it's there or looks at it.

doc/read-the-docs-site/extensions-flags-pragmas.rst
zliu41: > The problem with full-laziness is that specifically in GHC it does the reverse binder-swap transformation, which can lead to us re-evaluating things.

Understood, but I don't think we need to explain this detail here. Should suffice to say that `full-laziness` can be bad. I'll change "never beneficial" to "rarely beneficial".

doc/read-the-docs-site/extensions-flags-pragmas.rst
zliu41: Good point!

doc/read-the-docs-site/extensions-flags-pragmas.rst
zliu41: The INLINE pragma causes inlining only, not beta reduction, so I have a hard time thinking of a situation where it is useful. If you override the inliner's size/cost measures, you are going to get bigger/more expensive programs. If you override `effectSafe`, that's even worse since the dead binding won't be removed. So it's fair to say that it's rarely useful.

Anyway, I'll remove this paragraph.

michaelpj: You have a lot of faith in the ability of our inliner to get things right :) In practice, it is very conservative and there may well be times when the programmer wants to override it. GHC's inliner is much smarter, and there are still many times when you want manual INLINE pragmas

doc/read-the-docs-site/extensions-flags-pragmas.rst
effectfully: Meh, do folks actually do that? I read all the docs on GitHub.

doc/read-the-docs-site/extensions-flags-pragmas.rst
effectfully: Impressive! Thanks for explaining.

doc/read-the-docs-site/extensions-flags-pragmas.rst
zliu41: Most things do render nicely, except a few things like links.

doc/read-the-docs-site/extensions-flags-pragmas.rst
zliu41: It is defined at the top:

> all Plutus Tx modules (that is, Haskell modules that contain code to be compiled by the Plutus Tx compiler)

And yes that means the former.

PR #5526, opened 2023-09-06
Add workflow to run the cost model benchmark.

shlevy:


shlevy:
Draft pending https://github.com/input-output-hk/plutus/actions/runs/6098181769/job/16547457361

shlevy:
Looks like @kwxm cancelled the run to free up the bench machine, so let's just assume the process will work.

michaelpj:
Do we not need to test that this won't exceed the size limit for uploaded artifacts? I thought that was the worry?

kwxm:
@shlevy I'll run this over the weekend to check that it works (plus the numbers would be useful to us).

PR #5527, opened 2023-09-07
PLT-7064: Redo PIR purity analysis

michaelpj:
This PR went down a bit of a rabbit-hole. I was looking at `firstEffectfulTerm`, and it is tricky to get right. In particular, it reasons about the _evaluation order_ of the program. So I thought: the best way to find the first thing satisfying a predicate in a list is to just produce the whole list and then look for the thing. (It's also easier to test, since we can produce the whole evaluation order for a temrm and check to see if it makes sense.)

That led to `termEvaluationOrder`. Then I realised that you can use `termEvaluationOrder` to implement `isPure` (and otherwise they share a lot of the same logic!).

Finally, the use of `firstEffectfulTerm` itself was slightly weird: the thing we are looking for is in fact _not_ an effectful term, since it's an occurrence of a strict variable. So I wrote something more accurate in the module where it's used.

So this amounts to a bit of a rewrite of the purity analysis, but I think the result is more general and easier to understand and test.

We also get some modest PIR and budget improvements. Most of the PIR changes seem to be the `firstEffectfulTerm` check just working properly.

Since we have this logic in UPLC also, I'll update UPLC to use the same strategy if we decide we like this.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-core/plutus-ir/test/transform/inline/firstEffectfulTerm2
```diff
@@ -1,17 +1,19 @@
 -- `a` should not be inlined since it changes the semantics of the program.
+-- However, `b` _can_ be inlined: `a` is already evaluated by then, and after
+-- that `b` is the first variable to be evaluated.
```
michaelpj, line 3: I was initially surprised by this one but it's actually kind of great.

plutus-core/plutus-ir/test/transform/inline/firstEffectfulTerm2
```diff
@@ -1,17 +1,19 @@
...
 (let
   (nonrec)
   (termbind
     (strict)
     (vardecl a (fun (con integer) (con integer)))
-    { error (fun (con integer) (con integer)) }
```
michaelpj, line 7: FYI this is not how you write error terms. This is a type instantiation of a free variable called "error". I suspect a bunch of our tests don't typecheck because of this. It _probably_ isn't throwing off the purity analysis because it will assume that the unknown variable is impure, but still.

plutus-core/plutus-ir/test/transform/inline/effectfulBuiltinArg
```diff
@@ -0,0 +1,12 @@
+-- x should not be inlined, because the `trace "msg1"` call comes before
```
zliu41, line 1: I still think we should allow `x` to be inlined if `-conservative-optimisation` is off. In practice it is almost always acceptable to inline in cases like this.

Doing that separately is totally fine, although I'd imagine it is easier to implement that, and shouldn't need the `termEvaluationOrder` machinery.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -1,80 +1,213 @@
...
-asBuiltinApp :: Term tyname name uni fun a -> Maybe (BuiltinApp tyname name uni fun a)
-asBuiltinApp = go []
-    where
-        go argsSoFar = \case
-            Apply _ t arg  -> go (TermArg arg:argsSoFar) t
-            TyInst _ t arg -> go (TypeArg arg:argsSoFar) t
-            Builtin _ b    -> Just $ BuiltinApp b argsSoFar
-            _              -> Nothing
+-- | Is this pure? Either yes, or maybe not.
+data Purity = MaybeImpure | Pure
```
zliu41, line 60: If you split `Pure` into `PureAndWorkFree` and `PureAndNotWorkFree` then `termEvaluationOrder` can also be reused by `isEssentiallyWorkFree`.

zliu41 (commented):
I'm slightly surprised that all test cases affected are improvements. There are cases where this change causes more inlining, and there are also cases where it causes less inlining. The former must be much more common than the latter!

> So I thought: the best way to find the first thing satisfying a predicate in a list is to just produce the whole list and then look for the thing.

That sounds counter intuitive. If you can build the whole list surely it's not harder to find the first effectful term? You are using `Seq` for the list which is strict and can lead to building a large list of which you only need the first few elements. 

You can do two things to improve the efficiency: (1) no need to put anything `Pure` in it unless it is a variable; (2) stop building the list once you hit the first `Unknown`. But if you do these then it basically becomes the same thing as `firstEffectfulTerm`, so I don't understand why fixing `firstEffectfulTerm` is trickier than this.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
michaelpj: I wondered about that, and I think you probably could do, but I'm mostly just unsure what our precise definition of "work free" is, though.

plutus-core/plutus-ir/test/transform/inline/effectfulBuiltinArg
michaelpj: Yes, if we aren't being conservative then I think we can change the check from "is this the _first_ evaluated term" to "is this _an_ evaluated term".

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -1,80 +1,213 @@
...
+          <> evalThis (EvalTerm Pure t) -- then the whole term
+        t@(Case _ _ scrut _)    ->
+          goTerm scrut -- first the scrutinee
+          <> evalThis (EvalTerm Pure t) -- then the whole term
+          <> evalThis Unknown -- then we go to an unknown scrutinee
+
+        -- Leaf terms
+        t@(Var _ name)      ->
+          -- See Note [Purity, strictness, and variables]
+          let purity = case varStrictness name of { Strict -> Pure; NonStrict -> MaybeImpure}
```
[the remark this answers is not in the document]
michaelpj: Quite possibly. More generally, merging the underlying logic of `isPure` and `firstEffectfulTerm` means that we only have to do this sort of thing once in order to get consistent improvements.

plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
```diff
@@ -14,6 +14,14 @@ data AppContext tyname name uni fun ann =
   | TypeAppContext (Type tyname uni ann) ann (AppContext tyname name uni fun ann)
   | AppContextEnd
 
+splitApplicationNE :: Term tyname name uni fun ann
```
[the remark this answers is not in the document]
michaelpj: Ah yes, this was leftover.

michaelpj:
> That sounds counter intuitive. If you can build the whole list surely it's not harder to find the first effectful term? You are using Seq for the list which is strict and can lead to building a large list of which you only need the first few elements.

I meant it to be lazy! But you're right, Seq is strict in the structure, but not the elements (oops). I'll write some tests with some `undefined` terms to make sure I get it lazy enough.

I do think that phrasing a search as a generate+filter is generally nice, especially since you can use the generator for other things. For example, I think to do your suggestion [here](https://github.com/input-output-hk/plutus/pull/5527#discussion_r1319613838) we would need to look _past_ the first effectful term, so then we'd be rewriting anything.

> stop building the list once you hit the first Unknown

The semigroup instance should do this if it's lazy enough. But I need to test that.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -1,80 +1,213 @@
...
+
+        -- If we can view as a builtin application, then handle that specially
+        (splitApplication -> (Builtin a fun, args)) -> goBuiltinApp a fun args
+        -- We could handle functions and type abstractions with *known* bodies
+        -- here. But there's not much point: beta reduction will immediately
+        -- turn those into let-bindings, which we do see through already.
+        t@(Apply _ fun arg)    ->
+          goTerm fun -- first the function
+          <> goTerm arg -- then the arg
+          <> evalThis (EvalTerm Pure t) -- then the whole term
```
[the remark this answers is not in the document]
michaelpj: So, this would be relevant if e.g. you wanted to do the work-free check. You'd say something like "it's okay if the evaluation order includes work-free terms like `LamAbs`, but not if it includes e.g. actual applications". So you need to see the application step in the order. Plus it corresponds to what actually happens in the machine and the reduction rules: we evaluate the function, then the argument, then we do something for the application.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
michaelpj: I gave it a shot and I think it's actually pretty nice. I still think it's unclear what "work-free" means, but I think this way of looking at it makes that more obvious: we can look and see "what steps do we want to say do work?" more directly.

michaelpj:
> There are cases where this change causes more inlining, and there are also cases where it causes less inlining.

Where are the cases with less inlining? Just the ones where it was buggy before? I think those were (hopefully) rare. I think the reason this is a general win is that it's enabling more inlining of single-occurrence bindings, which is basically always a win so long as we let ourselves do it!

michaelpj:
I made it properly lazy, which turned out to be easy with just a DList. Test shows that we can analyze a term with `undefined` in it just fine.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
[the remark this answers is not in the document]
michaelpj: Also, there are a few places where we don't pass the variable strictness map to `isPure`, which means it will work less well.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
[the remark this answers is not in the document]
michaelpj: This is in the code now - we need those nodes otherwise we can't say why e.g. `[(lam x x) (con integer 1)]` is not work-free.

plutus-core/plutus-ir/test/transform/inline/effectfulBuiltinArg
michaelpj: I added a note about this.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -1,80 +1,244 @@
...
+
+instance Pretty WorkFreedom where
+  pretty MaybeWork = "maybe work?"
+  pretty WorkFree  = "work-free"
+
+-- | Either the "next" term to be evaluated, along with its 'Purity' and 'WorkFreedom',
+-- or we don't know what comes next.
+data EvalTerm tyname name uni fun a =
+  Unknown
+  | EvalTerm Purity WorkFreedom (Term tyname name uni fun a)
```
zliu41, line 77: I'd prefer to rule out the combination of `MaybeImpre` and `WorkFree` in the types, since an impure term cannot be work free.

zliu41 (approved):
> I do think that phrasing a search as a generate+filter is generally nice

Fair enough. I recall that several algorithms in the functional pearls book are done this way. As long as we use lazy structures, some algorithms can be expressed very elegantly in this manner.

> Where are the cases with less inlining? Just the ones where it was buggy before? I think those were (hopefully) rare.

Yeah, I thought this would lead to a number of performance degradations, but the opposite happened, which is a pleasant surprise!

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
michaelpj: I'm actually not sure I agree with that! For example, I think it kind of makes sense to consider `error` to be impure but work-free.

PR #5528, opened 2023-09-07
PLT-6722 Convert `fitsInto` tests to golden tests

thealmarty:
This will make it more convenient when the sizes change because we will be able to run `--test-options=--accept` to update the tests instead of doing it manually.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


michaelpj (commented):
Looks fine, would just be nice to tweak the test names to avoid weird filenames.

plutus-benchmark/nofib/test/Spec.hs
```diff
@@ -47,7 +47,7 @@ testClausify = testGroup "clausify"
                , testCase "formula4" $ mkClausifyTest Clausify.F4
                , testCase "formula5" $ mkClausifyTest Clausify.F5
                , runTestNested $
-                  Tx.goldenSize "formula1 (size)" (Clausify.mkClausifyCode Clausify.F1)
+                  Tx.goldenSize "formula1-size" (Clausify.mkClausifyCode Clausify.F1)
```
thealmarty, line 50: the golden files have the suffix `uplc-size`, so I can take out the "size" here if you like. Actually yeah I'll do that.

plutus-benchmark/script-contexts/test/checkScriptContext1.uplc-size.golden
```diff
@@ -0,0 +1 @@
+3432
```
thealmarty, line 1: I have verified that `master` has improved the sizes.

plutus-benchmark/nofib/test/Spec.hs
michaelpj: Yeah, I was going to suggest that but then the test names are the same as other tests, which does show up in the test output from `tasty`. IDK if it makes anything not work, but it seems maybe bad.

michaelpj: like, if you wanted to rerun just that one test it might be a problem, but that's not such a big deal

thealmarty: hmmm....yeah. I can add the `-size` back in then? or maybe add a `-golden` in?

michaelpj (approved):
I think the current names are fine but feel free to tweak a bit more if you want.

PR #5529, opened 2023-09-11
Change commit used when running benchmarks

zeme-wana:
The `benchmark.yml` workflow is used to compare benchmark results between two commits.

The two commits user are:
- The latest commit in the PR that triggers the workflow
- The current `master`.

Instead of using `master` we now want to use the commit before the first commit in the PR.


kwxm:
> /benchmark plutus-benchmark:nofib

Won't this still be running the old version and comparing with master?

michaelpj:
Ah right, this is a repository workflow, so it's working off master and won't get the new definition until this is merged. I guess let's just try it.

PR #5530, opened 2023-09-11
[PlutusTx] Add tail-recursion cost tests and optimize the stdlib

effectfully:
This adds some budget tests for tail-recursion vs no tail-recursion and then optimized the stdlib of Plutus Tx a bit, since I was there anyway.


plutus-tx-plugin/test/Budget/9.2/constAccR.budget.golden
```diff
@@ -0,0 +1,2 @@
+({cpu: 1624333576
+| mem: 4909700})
```
effectfully, line 2: Compared to `constAccL`

```
({cpu: 1693471576
| mem: 5210300})
```

this is better.

I.e. right-folding wins over left-folding.

plutus-tx-plugin/test/Budget/9.2/constElR.budget.golden
```diff
@@ -0,0 +1,2 @@
+({cpu: 1693333576
+| mem: 5209700})
```
effectfully, line 2: A teeny tiny win over left-folding:

```
({cpu: 1693471576
| mem: 5210300})
```

plutus-tx-plugin/test/Budget/9.2/show.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 2926804219
-| mem: 8741204})
\ No newline at end of file
+({cpu: 2695424219
+| mem: 7735204})
```
effectfully, line 2: I didn't optimize any `show` logic specifically, this is just from the general optimizations of the stdlib code.

plutus-tx-plugin/test/Budget/9.2/sumR.budget.golden
```diff
@@ -0,0 +1,2 @@
+({cpu: 2057711053
+| mem: 5912702})
```
effectfully, line 2: Another small win over left-folding:

```
({cpu: 2126918053
| mem: 6213602})
```

effectfully (commented):
Need one more self-review.

plutus-benchmark/script-contexts/test/checkScriptContextEqualityTerm-20.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 614150251
-| mem: 2512246})
\ No newline at end of file
+({cpu: 557731251
+| mem: 2266946})
```
effectfully, line 2: Not bad given that I didn't work on script context equality at all.

plutus-tx/src/PlutusTx/Foldable.hs
```diff
@@ -95,7 +95,7 @@ toList t = build (\ c n -> foldr c n t)
 -- | Plutus Tx version of 'Data.Foldable.length'.
 {-# INLINABLE length #-}
 length :: Foldable t => t a -> Integer
-length = foldl (\c _ -> c + 1) 0
+length = foldr (\_ acc -> acc + 1) 0
```
effectfully, line 98: Since `foldr` appears to be a tiny bit more efficient, why not utilize that.

plutus-tx-plugin/test/Budget/9.2/gte0.budget.golden
```diff
@@ -0,0 +1,2 @@
+({cpu: 2210128053
+| mem: 6212402})
```
effectfully, line 2: As expected, defining this in terms of the new inlinable `all` makes this noticeably faster. Here are the stats with non-worker-wrapper-transformed `recursiveAll`:

```
({cpu: 2509519053
| mem: 7514102})
```

plutus-tx-plugin/test/Budget/9.2/lte0.budget.golden
```diff
@@ -0,0 +1,2 @@
+({cpu: 2206118053
+| mem: 6212402})
```
effectfully, line 2: As expected `<= 0` is a teeny tiny bit faster than `>= 0` (because there's a builtin for the former, but not the latter, which is defined via the former + `not`). The stats for `gte0` are

```
({cpu: 2210128053
| mem: 6212402})
```

I'm not sure I understand how `mem` is exactly the same though.

plutus-benchmark/script-contexts/test/checkScriptContextEqualityTerm-20.budget.golden
effectfully: Not bad given that I didn't work on script context equality at all.

plutus-tx/src/PlutusTx/Numeric.hs
```diff
@@ -55,15 +56,15 @@ newtype Additive a = Additive a
 
 instance Semigroup a => AdditiveSemigroup (Additive a) where
     {-# INLINABLE (+) #-}
-    Additive x + Additive y = Additive (x <> y)
+    (+) = coerce ((<>) :: a -> a -> a)
```
zliu41, line 59: Is it more efficient this way?

plutus-tx-plugin/test/Budget/9.2/lte0.budget.golden
zliu41: I'm actually surprised that the CPU is different, since `x<=0` is `\x -> ifThenElse (x<=0) True False` and `x>=0` is `\x -> ifThenElse (x<0) False True` which should be the same.

plutus-tx-plugin/test/Budget/9.2/gte0.budget.golden
zliu41: Yeah, this makes a lot of difference and we should probably point it out in the doc.

zliu41 (approved):

plutus-tx-plugin/test/Budget/9.2/lte0.budget.golden
zliu41: Oh, `<=` and `<` have different CPU cost parameters, that's why!

plutus-tx/src/PlutusTx/Numeric.hs
michaelpj: It's an example of eta-reducing stuff. I'm also unsure why the eta-reduction helps except insofar as it shrinks the size of terms, which maybe makes the inliner more likely to inline things. For cases like this it means we'll end up with a binding which is just a variable reference, which we'll almost certainly inline, I hope.

effectfully:
> Okay, so this seems to confirm the hypothesis that tail-recursion is slightly faster?

It's the opposite from what I can tell. `foldl` is tail-recursive and it's slightly slower than the non-tail-recursive `foldr`, most likely because of the additional argument that we need to toss around in the worker of `foldl`.

Hence why I change `length` from being `foldl`-based to being `foldr`-based.

plutus-tx/src/PlutusTx/Numeric.hs
effectfully: > Is it more efficient this way?

Yes, fewer lambdas and fewer applications.

It even helps in Haskell, which is why we do it all over the place on the evaluation path. Quoting a Note from `base`:

```
{-
Note [Function coercion]
~~~~~~~~~~~~~~~~~~~~~~~

Several functions here use (#.) instead of (.) to avoid potential efficiency
problems relating to #7542. The problem, in a nutshell:

If N is a newtype constructor, then N x will always have the same
representation as x (something similar applies for a newtype deconstructor).
However, if f is a function,

N . f = \x -> N (f x)

This looks almost the same as f, but the eta expansion lifts it--the lhs could
be _|_, but the rhs never is. This can lead to very inefficient code.  Thus we
steal a technique from Shachaf and Edward Kmett and adapt it to the current
(rather clean) setting. Instead of using  N . f,  we use  N #. f, which is
just

coerce f `asTypeOf` (N . f)

That is, we just *pretend* that f has the right type, and thanks to the safety
of coerce, the type checker guarantees that nothing really goes wrong. We still
have to be a bit careful, though: remember that #. completely ignores the
*value* of its left operand.
-}
```

Also what Michael said.

plutus-tx-plugin/test/Budget/9.2/gte0.budget.golden
effectfully: It's kinda "common sense". Pull arguments out of recursion if they never change during evaluation. Not only will you pass fewer arguments around, you'll also provide more opportunities for optimizing usages of those that don't change. Works in Plutus the same way it works in Haskell (except in Haskell you have to deal with rewrite rules, deforestation etc, which complicate the picture).

effectfully:
> Lots of eta-reduction here. I'm not sure why that helps tbh. I guess it makes the definitions smaller.

I don't feel like it helps a lot. I think what really helps is worker-wrapper-transforming functions like `all`, `foldr`, `map` etc.

michaelpj:
> It's the opposite from what I can tell. foldl is tail-recursive and it's slightly slower than the non-tail-recursive foldr, most likely because of the additional argument that we need to toss around in the worker of foldl.

Okay, so the lesson is perhaps that it's so minor that it's swamped by even relatively minor other changes.

plutus-tx-plugin/test/Budget/9.2/gte0.budget.golden
michaelpj: We also have a ticket for doing this automatically: this is the "static argument transformation".

PR #5532, opened 2023-09-12
AsData: transfer deriving clauses from the quoted declaration

michaelpj:
This is much more ergonomic, and not much more nonsensical.


michaelpj:
I'm just going to merge this cos it's not that interesting and people agreed it was an improvement.

PR #5533, opened 2023-09-12
Produce more sensible matcher names

zliu41:
Nil_match becomes List_match


plutus-tx-plugin/plutus-tx-plugin.cabal
```diff
@@ -121,10 +121,6 @@ test-suite plutus-tx-tests
   if flag(use-ghc-stub)
     buildable: False
 
-  -- test output changes after 9.2, bad for golden tests
-  if impl(ghc >=9.3)
-    buildable: False
```
effectfully, line 126: Is this a related change?

effectfully (approved):
Much better, thanks.

plutus-tx-plugin/plutus-tx-plugin.cabal
zliu41: No but this was committed by accident in https://github.com/input-output-hk/plutus/pull/5528 and I need to remove it to run the tests for 9.6.

michaelpj (commented):
Good catch!

PR #5537, opened 2023-09-13
[Test] Add forgotten budget golden files

effectfully:
Guess who forgot a `git add .`.


effectfully:
Not waiting for a review or even requesting one, since this just adds missing golden files.

thealmarty:
The CI didn't complain?

effectfully:
I think it doesn't complain about missing golden files and just creates them silently instead.

thealmarty:
Right! That's not ideal. Not sure how to improve this though.

PR #5538, opened 2023-09-13
Add ScriptForEvaluation to the ledger API (PLT-7327)

zliu41:
- Added `ScriptForEvaluation`, containing both a serialised script and the corresponding deserialised script.
- Renamed `ScriptForExecution` to `ScriptNamedDeBruijn`.
- Renamed `fromSerialisedScript` to `deserialiseScript`, which produces a `ScriptForEvaluation`. This is the only way to obtain `ScriptForEvaluation`.
- Removed `assertScriptWellFormed`.

By the way, three kinds of `ScriptDecodeError` (CBORDeserialiseError, RemainderError, LedgerLanguageNotAvailableError) are checked in phase 1 (in `deserialiseScript`), while the fourth kind (PlutusCoreLanguageNotAvailableError) is checked in phase 2 (in `mkTermToEvaluate`). This can be very bewildering, and begs documentation and testing (PLT-6622).


plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
```diff
@@ -133,48 +146,33 @@ scriptCBORDecoder lv pv =
...
     toScripDecodeError :: CBOR.DeserialiseFailure -> ScriptDecodeError
     toScripDecodeError = CBORDeserialiseError . CBOR.Extras.readDeserialiseFailureInfo
-
-
-{-| Check if a 'Script' is "valid" according to a protocol version. At the moment this means "deserialises correctly", which in particular
-implies that it is (almost certainly) an encoded script and the script does not mention any builtins unavailable in the given protocol version.
-
-Note: Parameterized over the ledger-plutus-version since the builtins allowed (during decoding) differs.
--}
-assertScriptWellFormed :: MonadError ScriptDecodeError m
```
michaelpj, line 172: We agreed that we were going to leave the old interface in place, so that we can update without the ledger team having to update immediately.

So either we should leave this function in, or we should be prepared to re-create it on the ledger side when we update them. Not sure which is better.

plutus-benchmark/validation/BenchFull.hs
```diff
@@ -25,7 +25,8 @@ main = do
     evalCtx <- evaluate $ force mkEvalCtx
     let mkFullBM :: FilePath -> BS.ByteString -> Benchmarkable
         mkFullBM file bsFlat =
-            let UPLC.Program () ver body = unsafeUnflat file bsFlat
+            let pv = ProtocolVersion 6 0
```
michaelpj, line 28: `futurePV`?

plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/BenchUtil.hs
```diff
@@ -287,9 +287,14 @@ executeBenchmark serialisedValidator Benchmark{..} =
...
-      $ evaluateScriptCounting (ProtocolVersion 8 0) Verbose ec serialisedValidator
-        [bDatum, bRedeemer, toData bScriptContext]
+    case deserialiseScript pv serialisedValidator of
+      Left err -> Left (show err)
+      Right validator ->
+        Right
+          $ evaluateScriptCounting pv Verbose ec validator
+              [bDatum, bRedeemer, toData bScriptContext]
+  where
+    pv = ProtocolVersion 8 0
```
michaelpj, line 297: ditto

michaelpj (approved):
Looks basically fine to me so long as we sort the backwards compatibility thing.

michaelpj:
> By the way, three kinds of ScriptDecodeError (CBORDeserialiseError, RemainderError, LedgerLanguageNotAvailableError) are checked in phase 1 (in deserialiseScript), while the fourth kind (PlutusCoreLanguageNotAvailableError) is checked in phase 2 (in mkTermToEvaluate)

Yep, this is exactly the thing that tripped me before: I moved it to be with the others, which is not safe :(

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
```diff
@@ -113,14 +115,25 @@ uncheckedDeserialiseUPLC = unSerialiseViaFlat . deserialise . BSL.fromStrict . f
   where
     unSerialiseViaFlat (SerialiseViaFlat (UPLC.UnrestrictedProgram a)) = a
 
--- | A variant of `Script` with a specialized decoder.
-newtype ScriptForExecution = ScriptForExecution (UPLC.Program UPLC.NamedDeBruijn DefaultUni DefaultFun ())
+-- | A script with named de-bruijn indices.
+newtype ScriptNamedDeBruijn = ScriptNamedDeBruijn (UPLC.Program UPLC.NamedDeBruijn DefaultUni DefaultFun ())
+
+-- | A Plutus script ready to be evaluated on-chain, via @evaluateScriptRestricting@.
+data ScriptForEvaluation = UnsafeScriptForEvaluation SerialisedScript ScriptNamedDeBruijn
```
lehins, line 122: We need a few instances:

`NFData` instance is required for benchmarking.

Also `NoThunks` and `Eq` will be needed.

`Show` would be nice, but not an absolute requirement.

lehins (commented):
I think this will work. Thank you for doing this!

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
```diff
@@ -133,48 +146,33 @@ scriptCBORDecoder lv pv =
...
 
--- | The deserialization from a serialised script to a Script (for execution).
--- Called inside phase-1 validation (assertScriptWellFormed) and inside phase-2's `mkTermToEvaluate`
-fromSerialisedScript :: forall e m. (AsScriptDecodeError e, MonadError e m)
+-- | The deserialization from a serialised script into a `ScriptForEvaluation`,
+-- ready to be evaluated on-chain.
+-- Called inside phase-1 validation (i.e., deserialisation error is a phase-1 error).
+deserialiseScript :: forall m. (MonadError ScriptDecodeError m)
                      => PlutusLedgerLanguage -- ^ the Plutus ledger language of the script.
                      -> ProtocolVersion -- ^ which protocol version the script was submitted in.
```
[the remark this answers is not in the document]
michaelpj: I think we could just change our `ProtocolVersion` type to be `MajorProtocolVersion`. It sounds like the basic situation is that `plutus` should never care about the minor protocol version, so we can represent that in the types.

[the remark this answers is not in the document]
zliu41: > I think we could just change our ProtocolVersion type to be MajorProtocolVersion.

I'll open a separate PR for this.

> should not change for intra-era hardforks

Do you mean "should not be more restrictive"? It can become more permissive as new builtins or new Plutus Core versions are added in an intra-era HF.

> since ledger will not be able to supply to this function a major protocol version that was bumped during an intra-era hardfork.

Can you elaborate? Are you saying we'll still get the old major PV after an intra-era HF? Why is that?

zliu41:
Added the instances. The NFData instance for `ScriptForEvaluation` is derived via `OnlyCheckWhnf`, which should be sufficient since `deserialiseScript` is the only way to obtain a `ScriptForEvaluation`.

michaelpj:
> The NFData instance for ScriptForEvaluation is derived via OnlyCheckWhnf, which should be sufficient since deserialiseScript is the only way to obtain a ScriptForEvaluation.

Should be in a comment on the instance.

lehins:
> The NFData instance for ScriptForEvaluation is derived via OnlyCheckWhnf,

I think you meant to say NoThunks, not NFData :slightly_smiling_face:

zliu41:
> I think you meant to say NoThunks, not NFData 🙂

lol

PR #5541, opened 2023-09-14
Redo UPLC purity analysis to match PIR

michaelpj:
This is the UPLC counterpart of the PIR work. Broadly small improvements.

The new logic _almost_ entirely subsumes the old `delayAndVarIsPure` logic. There is one exception: we would effectively propagate the "forcing" context through several steps of evaluation, which is quite tricky, and the current code does not try to do. Now we only go into the body of a delay if it is _immediately_ enclosed by a force. I don't think this is a huge loss, though.

Also this finally fixes the broken testcase (it's still weird, but now it's just GHC's fault).

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


michaelpj:
(I think this is covered by the previous changelog entry)

zliu41 (commented):
I don't think the `<>` short circuits once it hits the first `Unknown` does it? It could build a much longer DList than needed, which then needs to be converted to a regular list in linear time, before it can be inspected.

michaelpj:
There's a test in the PIR version that the whole thing is lazy enough: it successfully processes a term with an `undefined` subterm, which it couldn't do if it was too strict. Note that it also can't be creating the "spine" of the DList, since knowing what the spine looks like requires analyzing later terms, which we know it doesn't.

Which reminds me, I should port the tests over too.

zliu41 (approved):
> Note that it also can't be creating the "spine" of the DList, since knowing what the spine looks like requires analyzing later terms, which we know it doesn't.

Yeah, you are right.

PR #5542, opened 2023-09-15
Fix strictness in generated IsData instances

michaelpj:
This is a bit nasty. We don't want to evaluate `indexMatchCase` and `fallthrough` unconditionally. We know that - we even have a `~` pattern to make sure of it. But then we use `const` on them! And function application is strict!

Why did this not go horribly wrong immediately? Because _GHC_ inlines `const` very aggressively. AFAICT, it only doesn't inline `const` with `-O0`... which is what I was using in my tests on Marlowe. So then I started seeing weird stuff. Regardless, using `const` in this way is clearly wrong according to the strict semantics we're using.

I audited the other uses of `const` in `plutus-tx` and they all look fine.

(I split this into two commits so you can see a readable diff by just looking at the second commit)

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-tx-plugin/test/IsData/9.2/unsafeDeconstructData.pir-readable.golden
```diff
@@ -1,30 +1,29 @@
 let
   data Unit | Unit_match where
     Unit : Unit
+  !error : all a. unit -> a = /\a -> \(thunk : unit) -> error {a}
+  ~reconstructCaseError : string = "PT1"
+  !trace : all a. string -> a -> a = trace
+  !unitval : unit = ()
+  ~fallthrough : all a. a
```
michaelpj, line 8: This is probably also a reason we got saved sometimes: if `fallthrough` is polymorphic then the type abstraction saves us.

michaelpj:
I think this is fairly safe and I need it for the Marlowe PR, but do take a look.

PR #5543, opened 2023-09-15
Testlibs: various changes

michaelpj:
- Avoids the duplcation in the many `run*` functions
- Ensures that when we have a failure we always have the final budget and logs available and printed.
- Remove the `X` vs `XCatch` distinction. In practice, producing a different output for a golden file is going to fail the test, so the non-catch variants provided only a small safeguard against not noticing that a golden file has become a failure, which didn't seem worth the duplication.
- Simplify our golden file extensions. There were a lot of things in there that were not really that useful. Now it's either: source (with language extension), output of evaluation, or size.

Going to have miserable conflicts, but I'll deal with them until we can decide if  we prefer this.


plutus-core/testlib/PlutusCore/Test.hs
```diff
@@ -302,69 +320,57 @@ goldenUPlc ::
...
-goldenUPlcBudget ::
+goldenUEvalBudget ::
   (ToUPlc a TPLC.DefaultUni TPLC.DefaultFun)
   => String
   -> [a]
   -> TestNested
-goldenUPlcBudget name values =
-  nestedGoldenVsDocM name ".budget" $ ppCatch $ view _2 <$> runUPlcBudget values
+goldenUEvalBudget name values =
+  nestedGoldenVsDocM name ".eval" $ ppCatch $ runUPlcBudget values
```
zliu41, line 373: Any particular reason to use `.eval` for all these cases rather than differentiating them?

zliu41 (approved):

plutus-core/testlib/PlutusCore/Test.hs
michaelpj: Is it useful? It's unclear that the suffix helps you at all here! I think it's somewhat useful to distinguish the files containing programs, since you could imagine syntax highlighting or tools that process them. TBH... that makes me wonder if we should just not use an extra suffix for these ones at all :thinking: 

The other advantage of using the same suffix is that if you change the type of output you don't have to remove and recreate the file. e.g. if you want to change a test to also print the logs you can just change the test function.

plutus-core/testlib/PlutusIR/Test.hs
```diff
@@ -120,6 +120,10 @@ withGoldenFileM name op = do
   where
     currentDir = joinPath <$> ask
 
+-- TODO: deduplicate with the PlutusuCore one
```
effectfully, line 123: A TODO.

effectfully (commented):
I like most of it, but I don't like that

1. the result of evaluation
2. the budget
3. the logs

now all have the same extension, despite all being rather different things.

Except... while we're here, should we just give up on printing budgets separately? This makes it super easy to miss a change in the result of evaluation (like it's happened recently with the Marlowe benchmarks), so should we just always print the result of evaluation together with budget?

Though, even then I still find it useful to reflect whether there's a budget inside of a file in the extension of the file. But I don't feel strongly about it (maybe yet).

plutus-core/testlib/PlutusCore/Test.hs
zliu41: Just a bit more convenient when you want to run different tests (e.g. godlenUEvalBudget and goldenUEvalLogs) on the same code, since you can use the same name.

michaelpj: Turns out we had precisely one case of name collision in the repository, which was some typechecking tests, which makes me think it doesn't matter. It's also not hard to change the test name.

plutus-core/testlib/PlutusIR/Test.hs
michaelpj: I'm going to do this later

michaelpj:
I'm going to merge this, it's progress.

PR #5548, opened 2023-09-19
Order V3.ParamName alphabetically

zliu41:
Since V3 is new, there's no reason not to order it alphabetically. Doing so makes it easier to generate the genesis file for Plutus V3.

Only when we add new builtins in the future, after V3 is already in production, do we need to append the new builtins at the end.

### DIFF
--- a/plutus-ledger-api/src/PlutusLedgerApi/V3/ParamName.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V3/ParamName.hs
@@ -29,15 +29,60 @@ data ParamName =
   | AppendString'memory'arguments'slope
   | BData'cpu'arguments
   | BData'memory'arguments
+  | Blake2b_224'cpu'arguments'intercept
+  | Blake2b_224'cpu'arguments'slope
+  | Blake2b_224'memory'arguments
   | Blake2b_256'cpu'arguments'intercept
   | Blake2b_256'cpu'arguments'slope
   | Blake2b_256'memory'arguments
+  | Bls12_381_G1_add'cpu'arguments
+  | Bls12_381_G1_add'memory'arguments
+  | Bls12_381_G1_compress'cpu'arguments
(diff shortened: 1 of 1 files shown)


michaelpj:
Hmmm. I guess. It was sort of nice to have each one be an extension of the previous set. It meant that e.g. they should all have an identical prefix if we update them in sync. But I guess it doesn't actually matter :thinking:

zliu41:
> It was sort of nice to have each one be an extension of the previous set.

But V2.ParamName is not an extension of V1.ParamName. It is alphabetically ordered, with the new builtins inserted in the middle.

Perhaps we can add a prefix, denoting when a builtin is added, so that they will stay alphabetically ordered even if we add new builtins in the future.

michaelpj:
> Perhaps we can add a prefix, denoting when a builtin is added, so that they will stay alphabetically ordered even if we add new builtins in the future.

:see_no_evil: 

I guess reordering them is fine. But perhaps we can decide on a policy and write it down in the haddock so it's clear.

james-iohk:
If we're doing this now it would be good to apply this at each new language version. The following could be reworded to "will be re-ordered alphabetically"
https://github.com/input-output-hk/plutus/blob/71459bffdd0b0f12e15be17d07157f47c811e921/doc/notes/cost-model-representations/cost-model-representations.md?plain=1#L199-L202

zliu41:
> apply this at each new language version

The orders for the language versions in production (V1, V2) cannot be changed. We are free to use whatever order, only when we are introducing a new language version.

zliu41:
On a second thought I don't think ordering them alphabetically is the right thing to do. Doing so is only convenient because of

```haskell
type CostModelParams = Map.Map Text.Text Integer
```

i.e., in `CostModelParams`, the keys are ordered alphabetically, because they are of type `Text`. As long as we can make the order of `ParamName` constructors consistent with the order of keys in `CostModelParams`, we don't have a problem.

PR #5549, opened 2023-09-19
Add golden tests to `size` tests

thealmarty:
This will make it easier to investigate the regressions in the UPLC/PIR optimization passes.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-tx/testlib/PlutusTx/Test.hs
+++ b/plutus-tx/testlib/PlutusTx/Test.hs
@@ -26,8 +26,9 @@ module PlutusTx.Test (
   goldenEvalCekLog,
 
   -- * Budget testing
-  goldenBudget
+  goldenBudget,
 
+  goldenTests
 ) where
 
 import Prelude
@@ -227,3 +228,12 @@ runPlcCekTrace values = do
         UPLC.runCek PLC.defaultCekParameters UPLC.tallying UPLC.logEmitter (p ^. UPLC.progTerm)
   res <- fromRightM (throwError . SomeException) result
   pure (logOut, tally, res)
+
+-- | A group of golden tests including UPLC size and UPLC and PIR readable.
+goldenTests :: String -> CompiledCode a -> [TestTree]
+goldenTests testName code =
+  [
+  runTestNested $ goldenSize testName code,
+  runTestNested $ goldenUPlcReadable testName code,
+  runTestNested $ goldenPirReadable testName code
+  ]
--- a/plutus-tx-plugin/abs-specialized.pir-readable.golden
+++ b/plutus-tx-plugin/abs-specialized.pir-readable.golden
@@ -1,0 +1,24 @@
+let
+  !`$fAdditiveMonoidInteger_$czero` : integer = 0
+  data Bool | Bool_match where
+    True : Bool
+    False : Bool
+  data Rational | Rational_match where
+    Rational : integer -> integer -> Rational
+in
+\(rat : Rational) ->
+  Rational_match
+    rat
+    {Rational}
+    (\(n : integer) (d : integer) ->
+       Bool_match
+         (ifThenElse
+            {Bool}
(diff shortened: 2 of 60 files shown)


plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -227,3 +228,12 @@ runPlcCekTrace values = do
         UPLC.runCek PLC.defaultCekParameters UPLC.tallying UPLC.logEmitter (p ^. UPLC.progTerm)
   res <- fromRightM (throwError . SomeException) result
   pure (logOut, tally, res)
+
+-- | A group of golden tests including UPLC size and UPLC and PIR readable.
+goldenTests :: String -> CompiledCode a -> [TestTree]
```
zliu41, line 233: Consider renaming to "goldenOpt". This is the group of test you'd want when you turn on the simplifier and want to verify that the program is optimized correctly. It makes little sense to use it on tests where the simplifier is disabled.

plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -227,3 +228,12 @@ runPlcCekTrace values = do
...
+goldenTests testName code =
+  [
+  runTestNested $ goldenSize testName code,
```
michaelpj, line 236: You should actually run them in a directory, look where the golden files have been generated. You want something like `runTestNestedIn ["test", "size"]`

michaelpj: Ah, or rather you want this to return `TestNested` so the size tests can discharge that correctly

plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -227,3 +228,12 @@ runPlcCekTrace values = do
...
+  runTestNested $ goldenUPlcReadable testName code,
```
michaelpj, line 237: personally I don't think that including the UPLC output is usually helpful. The final PIR is generally good enough

plutus-tx/testlib/PlutusTx/Test.hs
michaelpj: I think it's also maybe just not that generic? Perhaps it should just be a local helper function in the size tests

plutus-tx/testlib/PlutusTx/Test.hs
thealmarty: The golden UPLC files change with the new UPLC inliner code for the other `plutus-tx-tests` so I definitely want the UPLC golden files.

plutus-tx/testlib/PlutusTx/Test.hs
thealmarty: I originally wrote it in the size test module, but then I thought it may be useful elsewhere. I'm fine either way though.

plutus-tx/testlib/PlutusTx/Test.hs
michaelpj: Why these tests though? In the plutus tx tests we mostly look at the PIR, we have a few exceptions but they should be justified, though. It seems unlikely to me that we're regularly going to want to look at the UPLC here?

thealmarty: There are regressions in some of these tests. I didn't think it's costly to add more golden files, so I did it to all of these tests. Next time it may be other tests that will have regression so I was hoping to save us time. Also, all in all it'll be good to know that the UPLC terms haven't changed anyway.

michaelpj: It's a bit costly, in that it's more stuff to look through (and usually ignore). If there's an interesting example I'd be inclined to just add that one to the optimization tests.

thealmarty: Ok sure will do.

PR #5551, opened 2023-09-20
Optimization refactoring: VarsInfo and BuiltinsInfo, pure conapps

michaelpj:
This is some preparatory refactoring for doing case-of-case:
- Add a generic VarsInfo pass that collects information about PIR variables, such as strictness, arity, and whether they are constructors/matchers.
    - This sets us up to use this information to do case-of-case generically.
- Adds a BuiltinsInfo type that (for now) just contains the semantics variant, to hold user-provided information about builtins.
    - We'll want to add more information about builtins in order to know that e.g. `ifThenElse` is a matcher-like function that should participate in case-of-case.
- Use these throughout
    - This includes using them in some places that previously weren't handling variable strictness.
- Demonstrate the utility by using the new info to consider constructor applications to be pure

The actual optimization changes come from making more passes use variable info (which they should!) and making conapps pure (which they are!). It's a bit of a wash: some things are a little better, some things are a little worse, but the PIR looks better to me, and I think it's better to have accurate information.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-benchmark/nofib/test/formulaBudget.pir-readable.golden
```diff
@@ -39,7 +39,7 @@
     !equalsInteger : integer -> integer -> Bool
       = \(x : integer) (y : integer) ->
           ifThenElse {Bool} (equalsInteger x y) True False
-    ~`$fOrdInteger` : Ord integer
+    !`$fOrdInteger` : Ord integer
```
michaelpj, line 42: Now we know that the dictionary constructor is pure, so it's safe to strictify this!

plutus-benchmark/nofib/test/formulaBudget.pir-readable.golden
```diff
@@ -312,62 +312,63 @@
   in
   let
     ~`$dOrd` : Ord (List integer)
-      = let
-        !v : Ord integer = `$fOrdInteger`
```
michaelpj, line 316: this got inlined thanks to the above strictification

plutus-benchmark/nofib/test/formulaBudget.pir-readable.golden
```diff
@@ -905,205 +867,39 @@
...
-                                  Eqv (Sym 1) dt)
-                                  dt
-                            in
                             Eqv
-                              (let
-                                !dt : Formula = Sym 1
-                              in
-                              Eqv (Sym 1) dt)
-                              dt)
+                              (Eqv (Sym 1) (Sym 1))
```
michaelpj, line 871: All these conapps were pure and singly used, so we inline lots here

plutus-benchmark/script-contexts/test/checkScriptContext1.pir-readable.golden
```diff
@@ -96,11 +96,12 @@
                    {unit -> Extended a}
                    (equalsInteger 1 index)
                    (\(ds : unit) ->
-                      let
-                        !arg : data
-                          = headList {data} (sndPair {integer} {list data} tup)
-                      in
-                      Finite {a} (`$dUnsafeFromData` arg))
```
michaelpj, line 103: Now we know that `Finite` is strict, so `arg` is the first possibly-effectful thing evaluated here, so we can inline it!

plutus-tx-plugin/test/Budget/9.2/allEmptyList.pir-readable.golden
```diff
@@ -26,7 +26,4 @@ letrec
                  {all dead. dead})
           {all dead. dead}
 in
-let
-  !eta : List integer = Nil {integer}
-in
-go eta
```
michaelpj, line 32: first-effectful-term firing again

plutus-tx-plugin/test/Budget/9.2/applicative.pir-readable.golden
```diff
@@ -17,20 +15,12 @@ let
...
 in
 Maybe_match
   {integer -> integer}
   ds
   {all dead. Maybe integer}
-  (\(ipv : integer -> integer) ->
-     /\dead ->
-       Maybe_match
-         {integer}
-         y
```
michaelpj, line 30: y is singly used and now known to be pure, so we inline and COKC fires!

plutus-core/plutus-core/src/PlutusCore/Core/Type.hs
```diff
@@ -105,6 +109,26 @@ data Type tyname uni ann
     deriving stock (Show, Functor, Generic)
     deriving anyclass (NFData)
 
+-- | Get recursively all the domains and codomains of a type.
+-- @funTySections (A->B->C) = [A, B, C]@
+-- @funTySections (X) = [X]@
+funTySections :: Type tyname uni a -> NE.NonEmpty (Type tyname uni a)
```
zliu41, line 115: Since we have `splitApplication`, it's better to name it `splitFunTys`, and put them both in similar modules (if not the same module) so that they are easier to find. As it stands, if I want to do the same thing this function does, it's difficult to know that it exists.

The function in the GHC API that does this is also called `splitFunTys`, which makes it even more discoverable. If I need to split a function type, this name would be the first thing I look for.

plutus-core/plutus-ir/src/PlutusIR/Analysis/VarInfo.hs
```diff
@@ -0,0 +1,138 @@
...
+  -> VarsInfo tyname name
+  -> Maybe (TyVarInfo tyname name)
+lookupTyVarInfo name (VarsInfo _ vim) = PLC.lookupName name vim
+
+data TyVarInfo tyname name =
+  NormalTyVar {}
+  | DatatypeTyVar { dtNumTyVars :: Int, dtConstructors :: [name] }
+
+data VarInfo tyname name =
+  NormalVar { varStrictness :: Strictness, varArity :: Maybe Arity }
```
zliu41, line 42: Can we do without these partial field accessors?

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -142,6 +147,20 @@ termEvaluationOrder semvar varStrictness = goTerm
...
+        -- If we can view as a constructor application, then handle that specially.
+        -- Constructor applications are always pure: if under-applied they don't
+        -- reduce; if fully-applied they are pure; if over-applied it's going to be
+        -- a type error since they never return a function. So we can ignore the arity
+        -- in this case!
+        (splitApplication -> (h@(Var _ n), args))
+          | Just (DatatypeConstructor{}) <- lookupVarInfo n vinfo ->
+            evalThis (EvalTerm Pure MaybeWork h)
+            <>
+            goAppCtx args
```
zliu41, line 159: Aren't the args evaluated before `this`?

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -142,6 +147,20 @@ termEvaluationOrder semvar varStrictness = goTerm
...
         -- If we can view as a builtin application, then handle that specially
         (splitApplication -> (Builtin a fun, args)) -> goBuiltinApp a fun args
+        -- If we can view as a constructor application, then handle that specially.
+        -- Constructor applications are always pure: if under-applied they don't
+        -- reduce; if fully-applied they are pure; if over-applied it's going to be
+        -- a type error since they never return a function. So we can ignore the arity
+        -- in this case!
+        (splitApplication -> (h@(Var _ n), args))
+          | Just (DatatypeConstructor{}) <- lookupVarInfo n vinfo ->
+            evalThis (EvalTerm Pure MaybeWork h)
```
zliu41, line 157: Why `MaybeWork`? Does a constructor application do any non-trivial work?

plutus-core/plutus-ir/src/PlutusIR/Analysis/VarInfo.hs
```diff
@@ -0,0 +1,138 @@
...
+  | DatatypeConstructor { dcArity :: Arity , dcParentTyname :: tyname }
```
zliu41, line 43: Need some Haddock explaining what the "parent" of a constructor or a matcher means.

plutus-core/plutus-core/src/PlutusCore/Core/Type.hs
michaelpj: Yep, good call. I don't think we have an obvious home for it in `PlutusCore`, but we can certainly name it better.

plutus-core/plutus-ir/src/PlutusIR/Analysis/VarInfo.hs
michaelpj: The pattern matching is much nicer with them... Wish we could drop 8.10 so I can use `NoFieldSelectors`... but sure

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
michaelpj: `h` is the head of the application, applications are still evaluated left-to-right, so the head is always first.

plutus-core/plutus-ir/src/PlutusIR/Analysis/VarInfo.hs
michaelpj: If I get rid of the field names it will be even more obscure :(

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
michaelpj: Unclear - I'm generally unclear on what "work free" should actually mean. At the moment the only things that are "work free" are immediately returning a value, which a constructor application certainly does not do.

plutus-core/plutus-ir/src/PlutusIR/Analysis/VarInfo.hs
michaelpj: Added haddock anyway

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
zliu41: Oh, I thought `h` is the whole constructor application. If it is the head - the constructor itself, then shouldn't it always be WorkFree?

Also, shouldn't there be an `evalThis (EvalTerm Pure MaybeWork t)` after `goAppCtx args`? Because `t` is the constructor application, and `t` may not be workfree like you said.

Btw, `Unknown` seems redundant - can we use `EvalTerm Impure MaybeWork` in place of `Unknown`?

zliu41: > Also, shouldn't there be an evalThis (EvalTerm Pure MaybeWork t) after goAppCtx args?

Nvm about this, I see you just added it!

michaelpj: > If it is the head - the constructor itself, then shouldn't it always be WorkFree?

Unsure: this is the same as we do for variables. Is accessing a variable work-free? You have to look it up... maybe that counts as work? IDK, I just stuck to what was there.

> Btw, Unknown seems redundant - can we use EvalTerm Impure MaybeWork in place of Unknown?

Yes, that would be a slightly different way of looking at it. Unknown has two purposes: it tells us that the evaluation goes somewhere unknown, which lets us distinguish terms for which we can see the whole evaluation order with ones where we actually just don't know what happens. If we only care about purity/work, then we can approximate that with "something that is maybe impure and does work", but I think it's kind of nice to say that it's an unknown jump. Also, it's what we use to terminate the traversal: we don't do any more stuff after we hit an Unknown, which is what stops us traversing the whole term in that case.

zliu41 (approved):
LGTM.

PR #5554, opened 2023-09-21
PLT-506: Case-of-case

michaelpj:
Based on https://github.com/input-output-hk/plutus/pull/5551, look at the last commit if you're interested.

This is a bit tricky. The key is to extract out enough of the code for picking apart datatype match terms that the pass itself is comprehensible. We also need this because then we can abstract over builtin "matchers" also, which lets us handle e.g. the combination of `Bool_match` and `ifThenElse1`, which is very common.

There's a question of whether to only do it when we know it's a win. I've linked that to the conservative optimisation flag, which feels right. To my surprise, it seems to be generally better to just let it run regardless!

Improvements are, again, modest.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
```diff
@@ -0,0 +1,124 @@
...
+      -- 4. Then branches
+      , (TypeAppContext resTy resTyAnn (TermAppContext scrut scrutAnn branches)) <- args
+      =
+        let
+          sm = SplitMatchContext mempty (scrut, scrutAnn) (resTy, resTyAnn) branches
+          -- both branches have no args
+          branchArities = [[], []]
+          reconstruct (SplitMatchContext _ (scrut, scrutAnn) (resTy, resTyAnn) branches) =
+            TypeAppContext resTy resTyAnn (TermAppContext scrut scrutAnn branches)
+        in Just (sm, branchArities, reconstruct)
```
effectfully, line 104: Lots of name shadowing.

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,148 @@
...
+@
+    case (case s of { C1 a -> x; C2 b -> y; }) of
+      D1 -> w
+      D2 -> z
+
+    -->
+
+    case s of
+      C1 a -> case x of { D1 -> w; D2 -> z; }
+      C2 b -> case y of { D1 -> w; D2 -> z; }
```
effectfully, line 33: Any considerations regarding the fact that this duplicates branches and potentially makes the output exponentially big?

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,148 @@
...
+---> case-of-case
+ifThenElse {integer} (lessThanInteger 1 2) (Bool_match {integer} True 3) (Bool_match {integer} False 4)
+---> case-of-known-constructor
+ifThenElse {integer} (lessThanInteger 1 2) 3 4
+
+That is, this guarantees that case-of-known constructor will fire and get rid of one of the
+matches entirely, which is great.
+
+If we *don't* have this property, then case-of-case can duplicate code, making the program bigger.
+So the conservative option is to only do case-of-case when this is true.
```
effectfully, line 147: Aha, here they are. Is case-of-case interleaved with case-of-known-constructor? If not, can there still be an exponential blowup in code size before case-of-known-constructor cleans everything up? Worth discussing in this Note I think.

effectfully (commented):
Quite a diff! I've only had a very cursory look so far, I'd need a ticket and some allocated time to review this one properly.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
```diff
@@ -0,0 +1,124 @@
...
+    => BuiltinsInfo uni fun
+    -> fun
+    -> Arity
+builtinArity binfo fun =
+  let semvar = binfo ^. biSemanticsVariant
+  in case toBuiltinMeaning @uni @fun @(Term TyName Name uni fun ()) semvar fun of
+        BuiltinMeaning sch _ _ -> typeSchemeArity sch
+
+-- | Get the arity of a builtin function.
+builtinArity'
```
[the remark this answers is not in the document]
michaelpj: One just takes the `BuiltinSemanticsVariant`, one takes a `BuiltinsInfo`. I meant to define one in terms of the other... I wanted `builtinArity'` because it's more convenient to use in `defaultUniMatcherLike` below :thinking:

plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
```diff
@@ -0,0 +1,124 @@
...
+            SplitMatchContext tyname name uni fun a -> AppContext tyname name uni fun a)
+splitBuiltinDatatypeMatch binfo f args
+  | Just (BuiltinMatcherLike split) <- Map.lookup f (binfo ^. biMatcherLike)
+  = split args
+  | otherwise = Nothing
+
+defaultUniMatcherLike :: Map.Map DefaultFun (BuiltinMatcherLike DefaultUni DefaultFun)
+defaultUniMatcherLike = Map.fromList
+  [ (IfThenElse, BuiltinMatcherLike splitIfThenElse)
+  , (ChooseUnit, BuiltinMatcherLike splitChooseUnit)
```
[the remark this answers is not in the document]
michaelpj: Lists we can do, for pairs we don't actually have a matcher function!

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseOfCase.hs
michaelpj: Also note that I'm currently turning the non-conservative version on by default since it seems to help! I feel pretty nervous about that...

Good point about the interleaving: indeed, they run one after the other and only once in each simplifier iteration, so we should be safe, but worth pointing out.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
```diff
@@ -0,0 +1,124 @@
...
+import PlutusIR.Contexts
+import PlutusIR.Core.Type hiding (Type)
+import PlutusPrelude (Default (..))
+
+-- | The information we need to work with a builtin that is like a datatype matcher.
+data BuiltinMatcherLike uni fun = BuiltinMatcherLike
+  { _bmlSplitMatchContext :: forall tyname name a .
+      (AppContext tyname name uni fun a ->
+       Maybe (SplitMatchContext tyname name uni fun a, [Arity],
+              SplitMatchContext tyname name uni fun a -> AppContext tyname name uni fun a))
```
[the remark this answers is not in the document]
michaelpj: Yeah, I'm not super happy with this. It's very nearly a prism, except for this annoying arity business.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
michaelpj: Yes, I should tidy this up.

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseOfCase.hs
michaelpj: Good point, that seems possible. Let me try and come up with a case where that could happen.

michaelpj: Yep, I can make it happen. I think if we're in conservative mode then COKC is still guaranteed to clear it up, so the worst that can happen is a big intermediate tree.

michaelpj:
So I realised that
1. Roman is right, and case-of-case can create exponentially large programs in a single pass
2. Even if we only trigger it when the branch bodies are conapps, we can still get duplication (and hence exponential programs).

That means I don't think we can have a strictly conservative case-of-case at all (or we'd have to assert that the branch bodies were all _distinct_ conapps, which seems annoying). I'm not totally sure what to do then: turn it off entirely when in conservative mode, and turn it on universally otherwise (just forgetting the conapps-only mode?).

plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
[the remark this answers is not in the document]
michaelpj: Yeah. And it's slightly weird to treat `chooseList` as a "matcher": I'd be treating it as a matcher with 0 arity branches, which is in fact fine, but is a little strange...

michaelpj:
> processTerm is a single step, right? And transformOf in caseOfCase performs it to the fixpoint. 

`transformOf` does not run it to fixpoint, but it does run it recursively bottom up, which is enough. That means you can generate duplicate code in the branches before then duplicating the branches, exponential goes brrr.

> Can't you just interleave the step function of caseOfCase with the step function of COKC and run that to the fixpoint? Wasn't it what GHC was doing?

Hmmm maybe we could fuse the passes. I think the issue will be that case-of-case creates opportunities for COKC in a _subterm_ of the term currently being processed, so naive fusion won't work.

We could directly call the COKC step from the case-of-case step if we think we've made an opportunity for it. That would avoid the problem partly, although if you look in the comment I added you'll see there's still a case where we can get duplication :(

plutus-core/plutus-ir/src/PlutusIR/Compiler/Types.hs
```diff
@@ -133,7 +137,7 @@ data CompilationCtx uni fun a = CompilationCtx {
 makeLenses ''CompilationCtx
 
 toDefaultCompilationCtx
-    :: (Default (PLC.BuiltinSemanticsVariant fun), Default (PLC.CostingPart uni fun))
+    :: (Ord fun, Default (PLC.BuiltinSemanticsVariant fun), Default (PLC.CostingPart uni fun))
```
bezirg, line 140: This `Default CostingPart` always makes me uneasy. It sucks that we have to pass some fake (zero) consting just so we can have a runtime available for the EvaluateBuiltins pass (and perhaps other passes as well)

effectfully: Is there another way to do it? Costing _is_ a part of the evaluation semantics of a builtin.

michaelpj: I guess in principle we could make it optional? But then we'd need to pass `Maybe cost` around in various places which would have overhead in the machine, so this seems better.

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,198 @@
...
+processTerm ::
+    forall tyname name uni fun a .
+    (Ord fun, PLC.HasUnique name PLC.TermUnique
+    , PLC.HasUnique tyname PLC.TypeUnique) =>
+    BuiltinsInfo uni fun ->
+    VarsInfo tyname name ->
+    Bool ->
+    Term tyname name uni fun a ->
+    Term tyname name uni fun a
+processTerm binfo vinfo conappsOnly t
```
bezirg, line 63: very long/dense hard-to-read code

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,198 @@
...
+        mkNewOuterMatch newScrut =
+          reconstructOuter $ SplitMatchContext outerVars (newScrut, outerScrutAnn) (outerResTy, outerResTyAnn) outerBranches
+
+    -- See Note [Case-of-case and conapps]
+    , Just newInnerBranches <- mapBranches (\b -> if not conappsOnly || isConApp b then Just (mkNewOuterMatch b) else Nothing) innerBranches innerBranchArities
+    , let
+        newInnerMatch =
+          reconstructInner $ SplitMatchContext innerVars (innerScrut, innerScrutAnn) (outerResTy, outerResTyAnn) newInnerBranches
+
+      = newInnerMatch
```
bezirg, line 80: TBH, I don't like  all these `let`s inside the guard. It makes it hard to separate what is the "matching/condition part" and what is the "value returning/reconstruction part"

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,198 @@
...
+    | otherwise = t
+    where
+      -- We know this value gets cased on, so if it's a constructor application it _must_ be saturated:
+      -- if it was undersaturated the result would be a function (which can't be cased on), and if it was
+      -- oversaturated then it wouldn't typecheck since you can't apply a datatype value
+      isConApp (splitApplication -> (Var _ n, _)) | Just (DatatypeConstructor{}) <- lookupVarInfo n vinfo = True
+      isConApp _ = False
```
bezirg, line 87: oh no,  lets + a where!

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,198 @@
...
+
+    -- The scrutinee is itself an application
+    -- Note: we don't use the inner result type, we're going to replace it
+    , Just (SplitMatchContext innerVars (innerScrut, innerScrutAnn) _ innerBranches, innerBranchArities, reconstructInner) <- splitMatch binfo vinfo outerScrut
+    , let
+        mkNewOuterMatch newScrut =
+          reconstructOuter $ SplitMatchContext outerVars (newScrut, outerScrutAnn) (outerResTy, outerResTyAnn) outerBranches
+
+    -- See Note [Case-of-case and conapps]
+    , Just newInnerBranches <- mapBranches (\b -> if not conappsOnly || isConApp b then Just (mkNewOuterMatch b) else Nothing) innerBranches innerBranchArities
```
bezirg, line 75: Bind to some var this argument function

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,198 @@
...
+  -> Maybe (AppContext tyname name uni fun a)
+mapBranches f = go
+    where
+    go :: AppContext tyname name uni fun a -> [Arity] -> Maybe (AppContext tyname name uni fun a)
+    go AppContextEnd [] = Just AppContextEnd
+    go (TermAppContext branch ann ctx) (arity:arities) =
+      TermAppContext <$> underBindersMaybe arity f branch <*> pure ann <*> go ctx arities
+    go _ _ = Nothing
+
+    -- I tried to come up with a way to do this in terms of the underBinders traversal
```
bezirg, line 105: I don't get it. underbinders should just work since its a traversal with Maybe as the Applicative?

```
    go (TermAppContext branch ann ctx) (arity:arities) =
      TermAppContext <$> underBinders arity f branch <*> pure ann <*> go ctx arities
```

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,198 @@
...
+  C3 c -> w
+
+i.e. we still duplicate v!
+
+It's not all that uncommon to end up with multiple case branches that have the same
+result, e.g. if you write a wildcard fallthrough case in Haskell.
+-}
+
+{- Note [Case-of-case and conapps]
+Case-of-case will be much better if the bodies of the branches are all constructor applications.
```
bezirg, line 187: It's not jus t"will be much better", but to me it seems caseofcase is only applied if all the branches are conapps, correct?

plutus-tx-plugin/test/size/max.uplc-size.golden
```diff
@@ -1 +1 @@
-36
\ No newline at end of file
+33
```
bezirg, line 1: That's nice, I wish i could see the simplified uplc here to enjoy the benefit of caseofcase+caseknown+inline!

plutus-core/plutus-ir/src/PlutusIR/Compiler/Types.hs
effectfully: > I guess in principle we could make it optional? 

image

effectfully: @bezirg 

> But the compiler does not make use of the costing part at all. It's just for compile-time evaluation, disregarding the costs (more precisely treating all as zero-cost)

Good point, though. I'll think about it, thank you.

effectfully: If you only need compile-time evaluation, you can use the first two fields of `BuiltinMeaning` like we do in the well-typed Hedgehog program [generators](https://github.com/input-output-hk/plutus/blob/de48aa5ddce4dc3c571e84fb763e9e88cb7239c2/plutus-core/testlib/PlutusCore/Generators/Hedgehog/Denotation.hs#L39-L42) for example. Although it's probably not a good idea, since you probably don't want to roll out your own lifting and unlifting logic etc. Maybe we should indeed have some middle-ground. 

Although I still feel like insisting everywhere that costing is a part of operational semantics of builtins will save us so much trouble in the long run that it may not be optimal to break this rule even when it makes sense. I may be wrong about that in this particular case, though.

effectfully:
@michaelpj 

> > Can't you just interleave the step function of caseOfCase with the step function of COKC and run that to the fixpoint? Wasn't it what GHC was doing?
>
> Hmmm maybe we could fuse the passes. I think the issue will be that case-of-case creates opportunities for COKC in a subterm of the term currently being processed, so naive fusion won't work.
>
> We could directly call the COKC step from the case-of-case step if we think we've made an opportunity for it. That would avoid the problem partly, although if you look in the comment I added you'll see there's still a case where we can get duplication :(

Thanks for the explanation. Can you then just create a `let` binding for each of the branches that get duplicated and let the inliner deal with it down the line? And maybe inlining + COKC is easier to arrange than case-of-case + COKC? "This is safe to inline, because it immediately reduces by COKC to an acceptable expression".

michaelpj:
> Can you then just create a let binding for each of the branches that get duplicated and let the inliner deal with it down the line? 

Yeah, I was also wondering about this. This is essentially what GHC does with join points. In our case it's not "free" like join points are, but it might still be worth it. And it does avoid the exponential code problem. So maybe it is indeed the right thing to do. In fact, I already have most of the code there to do this...

> And maybe inlining + COKC is easier to arrange than case-of-case + COKC? "This is safe to inline, because it immediately reduces by COKC to an acceptable expression".

Right, and that's a generally useful inlining heuristic that will help us anyway. For most cases we don't even need a new heuristic, though: in the non-duplicative cases the binding will be single-occurrence, so the inliner will already inline it.

This seems promising, I think I should try this.

michaelpj:
> For most cases we don't even need a new heuristic, though: in the non-duplicative cases the binding will be single-occurrence, so the inliner will already inline it.

No, this is totally wrong: we get _one_ binding which is used in every case branch, so it will almost never be single-use. So we are unlikely to inline it currently, which is a bit of a problem. Boo.

michaelpj:
I guess we could make a binding in conservative mode? That avoids the program growth problem, and _currently_ won't optimize well, but maybe that's okay.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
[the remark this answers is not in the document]
bezirg: > I don't get what is the actual need of having this datatype as _existential_. Isn't TypeScheme (that Arity originates from) hardcoded to work on TyName/Name? That only leaves us a , which could just be left as a type param?

@michaelpj I got confused. This is not an existential type :man_facepalming:

plutus-core/plutus-ir/src/PlutusIR/Analysis/VarInfo.hs
```diff
@@ -144,8 +123,38 @@ bindingVarInfo = \case
         in VarsInfo (PLC.insertByName matcher info mempty) mempty
       constrArity constrTy =
         -- One parameter for all function type arguments
-        fmap (const TypeParam) (funTyArgs constrTy)
+        fmap (const TermParam) (funTyArgs constrTy)
```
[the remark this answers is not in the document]
michaelpj: It was wrong before! now it's still slightly wrong actually, but I'll fix it

plutus-core/plutus-ir/src/PlutusIR/Arity.hs
```diff
@@ -0,0 +1,38 @@
...
+abstractions, this is not a simple argument count, but rather a list of values
+indicating whether the next argument should be a term or a type.
+
+Note that this is the syntactic arity, i.e. it just corresponds to the number of
+syntactic lambda and type abstractions on the outside of the term. It is thus
+an under-approximation of how many arguments the term may need.
+e.g. consider the term @let id = \x -> x in id@: the variable @id@ has syntactic
+arity @[]@, but does in fact need an argument before it does any work.
+-}
+type Arity = [Param]
```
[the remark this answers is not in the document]
michaelpj: I guess I think about it differently. For me arity is a "summary of how the function is going to be used in the program". As such it makes sense to me to include information about the parameters, because that's what you need!

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -244,18 +245,19 @@ termEvaluationOrder binfo vinfo = goTerm
...
         let
-          saturated = isSaturated binfo hd args
+          saturated = saturates args (builtinArity binfo hd)
           reconstructed = fillAppContext (Builtin a hd) args
           evalEffect = case saturated of
-            -- If it's saturated, we might have an effect here
-            Just True  -> evalThis (EvalTerm MaybeImpure MaybeWork reconstructed)
+            -- If it's saturated or oversaturated, we might have an effect here
+            Just Saturated      -> evalThis (EvalTerm MaybeImpure MaybeWork reconstructed)
+            Just Oversaturated  -> evalThis (EvalTerm MaybeImpure MaybeWork reconstructed)
```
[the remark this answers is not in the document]
michaelpj: Yes

plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
```diff
@@ -1,19 +1,125 @@
...
   { _biSemanticsVariant :: PLC.BuiltinSemanticsVariant fun
+  , _biMatcherLike      :: Map.Map fun (BuiltinMatcherLike uni fun)
   }
 
 makeLenses ''BuiltinsInfo
 
-instance (Default (BuiltinSemanticsVariant fun)) => Default (BuiltinsInfo uni fun) where
-  def = BuiltinsInfo def
+instance (Ord fun, Default (BuiltinSemanticsVariant fun)) => Default (BuiltinsInfo uni fun) where
+  def = BuiltinsInfo def mempty
```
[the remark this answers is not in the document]
michaelpj: You can't make instances for instantiated versions of datatypes like that. You'd have to have a newtype.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
```diff
@@ -1,19 +1,125 @@
...
+    => Proxy uni
+    -> PLC.BuiltinSemanticsVariant fun
+    -> fun
+    -> Arity
+builtinArity' _ semvar fun =
+  case toBuiltinMeaning @uni @fun @(Term TyName Name uni fun ()) semvar fun of
+        BuiltinMeaning sch _ _ -> typeSchemeArity sch
+
+-- | Split a builtin 'match'.
+splitBuiltinDatatypeMatch ::
```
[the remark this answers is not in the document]
michaelpj: `splitBultinDatatypeMatch` is how you consume `BuiltinMatcherLike`,really

plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
```diff
@@ -58,6 +60,72 @@ lengthContext = go 0
...
+instance Semigroup (AppContext tyname name uni fun a) where
+  (<>) = appendAppContext
+
+instance Monoid (AppContext tyname name uni fun a) where
+  mempty = AppContextEnd
+
+data Saturation = Oversaturated | Undersaturated | Saturated
+
+-- | Do the given arguments saturate the given arity?
+saturates :: AppContext tyname name uni fun a -> Arity -> Maybe Saturation
```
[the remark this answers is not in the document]
michaelpj: Not sure? `semialign` constructs a new list or whatever, this computes a summary...

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseOfCase.hs
```diff
@@ -0,0 +1,198 @@
...
+    ) =>
+    BuiltinsInfo uni fun ->
+    Bool ->
+    Term tyname name uni fun a ->
+    m (Term tyname name uni fun a)
+-- See Note [Case-of-case and conapps]
+caseOfCase binfo conappsOnly t0 = do
+    -- We are going to record information about variables in a global map, so we
+    -- need global uniqueness
+    t <- PLC.rename t0
```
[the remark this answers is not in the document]
michaelpj: I copied this from KnownCon. Really I just want all the passes to properly declare if they need renaming or not. The current situation is a mess.

plutus-core/plutus-ir/src/PlutusIR/Compiler.hs
```diff
@@ -116,10 +118,13 @@ availablePasses :: [Pass uni fun]
 availablePasses =
     [ Pass "unwrap cancel"        (onOption coDoSimplifierUnwrapCancel)       (pure . Unwrap.unwrapCancel)
     , Pass "case reduce"          (onOption coDoSimplifierCaseReduce)         (pure . CaseReduce.caseReduce)
+    , Pass "case of case"         (onOption coDoSimplifierCaseOfCase)         (\t -> do
```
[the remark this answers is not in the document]
michaelpj: Now there is one to turn it on and one to be conservative

michaelpj:
This is good to go now, I think.
- We always do CoC if all the branches are _distinct_ conapps, this is guaranteed to be good
- If we don't know this, then we do it unless we are being conservative, in which case we bind it to a variable
    - The conservative case is not a noticeable regression: it looked like it was but it was a weird interaction with the UPLC case-of-case. I checked without that and it's fine.
- Refactored the code a bit following suggestions.

PR #5560, opened 2023-09-25
PLT-871: ThunkRecursions conservative mode

bezirg:
The current thunk recursions transformation has 2 problems:

1) does not preserve order of effects, so it may rearrange an error or a trace
2) does not respect the original strictness of some pure-strict bindings, turning them into lazy (not extremely bad)

This PR fixes (1) and (2) 

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-core/plutus-ir/src/PlutusIR/Compiler.hs
```diff
@@ -241,7 +249,7 @@ compileReadableToPlc (Program a v t) =
         >=> NonStrict.compileNonStrictBindings False
         >=> through check
         >=> (<$ logVerbose "  !!! thunkRecursions")
-        >=> (withBuiltinsInfo . fmap pure . flip ThunkRec.thunkRecursions)
+        >=> (\t' -> withBuiltinsInfo $ \semvar -> withCcOpts $ \opts -> pure $ ThunkRec.thunkRecursions semvar opts t')
```
michaelpj, line 252: at this point it's easier to just write an inline `\t -> do ...`, I think?

plutus-core/plutus-ir/src/PlutusIR/Transform/ThunkRecursions.hs
```diff
@@ -93,51 +94,74 @@ becomes
...
+    isStrictEffectful :: Binding tyname name uni fun a -> Bool
+    isStrictEffectful = \case
+        TermBind _ Strict _ rhs -> not $ isPure binfo vinfo rhs
+        _                       -> False
+
+    needsNonStrictify :: Binding tyname name uni fun a -> Bool
+    needsNonStrictify b =
+        if opts^.coPreserveEffectOrder
+        then isProblematic b || isStrictEffectful b
+        else isProblematic b
```
michaelpj, line 149: I'd like to see some discussion about why it's worth having two modes! Why not just do the conservative thing always? It doesn't look like a lot of savings: the advantage of non-conservative mode is that we keep the recursive binding strict, and so we avoid some overhead from that. But it's not clear to me how much difference that makes. And it only applies in a niche case where we have something that _is_ of function type (so doesn't get strictified anyway), but is impure.

So I'd be pretty inclined to just make the conservative mode the only mode.

michaelpj (commented):
Change definitely makes sense. Various thoughts...

plutus-core/plutus-ir/src/PlutusIR/Transform/ThunkRecursions.hs
```diff
@@ -93,51 +94,74 @@ becomes
...
     -- See Note [Thunking recursions]
-    let (toThunk, noThunk) = partition strictNonFunctionBinding bs
-        newBindings = fmap nonStrictify toThunk ++ noThunk
-        strictifiers = mapMaybe (mkStrictifyingBinding binfo vinfo) toThunk
-    in mkLet a Rec newBindings $ mkLet a NonRec strictifiers t
-thunkRecursionsStep _ _ t = t
+    let (toNonStrictify, rest) = NE.partition needsNonStrictify bs
+        -- MAYBE: use some prism/traversal to keep the original arrangement of the let group
+        -- this is not a semantic problem, but just a stylistic/debugging issue where the original
+        -- let-group will have reordered the (lazified) bindings
```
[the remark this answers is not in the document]
bezirg: I am also not sure if i want to rely on a later pass

plutus-core/plutus-ir/test/TransformSpec.hs
```diff
@@ -65,13 +64,28 @@ transform =
...
 thunkRecursions :: TestNested
 thunkRecursions =
-    testNested "thunkRecursions" $
-        map
-            (goldenPir (ThunkRec.thunkRecursions def) pTerm)
-            [ "listFold"
-            , "monoMap"
-            , "errorBinding"
-            ]
+    testNested "thunkRecursions"
```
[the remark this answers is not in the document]
bezirg: I also thought of adding that, but I was hit by an unrelated trace missing bug

plutus-core/plutus-ir/src/PlutusIR/Compiler.hs
bezirg: this whole thing sucks a bit. I pushed the monadreader compilationctx down to each pass and then I was super excited how the code looked great. But then half of the test cases broke, so i reverted it

plutus-core/plutus-ir/src/PlutusIR/Transform/ThunkRecursions.hs
```diff
@@ -93,51 +93,69 @@ becomes
...
+        editedLet = mkLet a Rec $ fmap nonStrictifyB toNonStrictify ++ rest
+        -- we now have to consider for which of those non-strictified we are going to introduce strictifiers
+        -- to preserve the original strictness annotations, we have to introduce strictifiers *for all* those we previously non-strictified
```
michaelpj, line 140: The real reason is to preserve any potential effects, right?

michaelpj (approved):
lgtm

plutus-core/plutus-ir/src/PlutusIR/Transform/ThunkRecursions.hs
bezirg: Indeed, this comment came from somewhere else. The comment made sense previously but not now. Fixing

PR #5561, opened 2023-09-27
Bump dependencies

erikd:
* Update both index-states.
* Remove un-needed constraints.


erikd:
Where do I put "No Changelog Required" to disable that check?

erikd:
The Nix failures are all in upstream packages.

michaelpj:
I'm assuming it built for you locally. I think the issue is that you didn't bump hackage.nix. I did that, and also bumped haskell.nix because then we hit another bug that I know got fixed.

The turtle issue I managed to fix upstream by getting inline-c updated, so I would indeed hope that we could get rid of those constraints, but let's see.

PR #5565, opened 2023-09-28
PLT-1398 Set up typechecking property tests for all PIR compiler passes

thealmarty:
Add typechecking property tests to each simplifier pass and the overall simplifier pass.

Some of the code is a bit repetitive but I think it's ok. This way it's tidy and easy to follow which tests are which.

The `inline` and overall simplifier passes are failing atm so they're marked as expected to fail. Debugging to follow in another PR. See [ticket](https://input-output.atlassian.net/browse/PLT-7746).

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


thealmarty:
I'll be adding other PIR compiler passes. Feel free to look at the tests for the simplifier passes.

plutus-core/plutus-ir/test/TypecheckSpec.hs
```diff
@@ -0,0 +1,141 @@
...
+
+-- Convert Either Error () to Either String () to match with the Testable (Either String ())
+-- instance.
+convertToEitherString :: Either (Error DefaultUni DefaultFun ()) ()
+  -> Either String ()
+convertToEitherString = \case
+  Left err -> Left $ show err
+  Right () -> Right ()
+
+-- | Typechecking a term after one of the pure passes.
```
michaelpj, line 75: This is actually doing more than that! This is not just checking that the term typechecks afterwards, it is checking that it has the same type as it had beforehand. Worth mentioning.

plutus-core/plutus-ir/test/TypecheckSpec.hs
```diff
@@ -0,0 +1,141 @@
...
+prop_typecheck_pure ::
+  (Term TyName Name DefaultUni DefaultFun ()
+  -> Term TyName Name DefaultUni DefaultFun ())
+  -> Property
+prop_typecheck_pure pass =
+  -- generate type and term in debug mode for easier debugging
+  forAllDoc "ty,tm" genTypeAndTermDebug_ (const []) $ \ (_ty, tm) -> do
+  convertToEitherString $ tyCheckPure pass tm
+
+-- | Typechecking a term after a `PlutusIR.Compiler.simplifyTerm` pass.
```
michaelpj, line 95: ditto

plutus-core/plutus-ir/test/TypecheckSpec.hs
```diff
@@ -0,0 +1,141 @@
...
+prop_typecheck :: Property
```
michaelpj, line 96:
```suggestion
prop_typecheck_simplify :: Property
```

plutus-core/plutus-ir/test/TypecheckSpec.hs
```diff
@@ -0,0 +1,141 @@
...
+prop_typecheck =
+  -- generate type and term in debug mode for easier debugging
+  forAllDoc "ty,tm" genTypeAndTermDebug_ (const []) $ \ (_ty, tm) ->
+    convertToEitherString $ do
+    config <- getDefTypeCheckConfig ()
+    plcConfig <- PLC.getDefTypeCheckConfig ()
+    inferredType <- runQuoteT $ inferType config tm
+    case runReaderT (runQuoteT (simplifyTerm (original tm))) (toDefaultCompilationCtx plcConfig) of
```
michaelpj, line 104: I'd have thought we'd want to start with the individual passes, rather than running the whole simplify loop?

plutus-core/plutus-ir/test/TypecheckSpec.hs
```diff
@@ -0,0 +1,141 @@
...
+  forAllDoc "ty,tm" genTypeAndTermDebug_ (const []) $ \ (_ty, tm) -> do
+  convertToEitherString $ tyCheckPure pass tm
+
+-- | Typechecking a term after a `PlutusIR.Compiler.simplifyTerm` pass.
+prop_typecheck :: Property
+prop_typecheck =
+  -- generate type and term in debug mode for easier debugging
+  forAllDoc "ty,tm" genTypeAndTermDebug_ (const []) $ \ (_ty, tm) ->
+    convertToEitherString $ do
+    config <- getDefTypeCheckConfig ()
```
michaelpj, line 101: weird indentation

plutus-core/plutus-ir/test/TypecheckSpec.hs
```diff
@@ -0,0 +1,141 @@
...
+    case runReaderT (runQuoteT (simplifyTerm (original tm))) (toDefaultCompilationCtx plcConfig) of
+      Left err -> Left $ void err
+      Right simplified ->
+        runQuoteT $ checkType config () (void simplified) inferredType
+
+-- | Typechecking a term after a `PlutusIR.Transform.StrictifyBindings` pass.
+prop_typecheck_strictifyBindings :: BuiltinSemanticsVariant DefaultFun -> Property
+prop_typecheck_strictifyBindings biVariant =
+  -- generate type and term in debug mode for easier debugging
+  forAllDoc "ty,tm" genTypeAndTermDebug_ (const []) $ \ (_ty, tm) ->
```
michaelpj, line 113: It seems to me that these all fit into `prop_typecheck_pure`, you just have to capture the other stuff in the pass function, e.g.
```
prop_typecheck_strictifyBindings biVariant = prop_typecheck_pure (strictifyBindings (BuiltinsInfo biVariant)
```

plutus-core/plutus-ir/test/TypecheckSpec.hs
```diff
@@ -0,0 +1,141 @@
...
+
+-- | Typechecking property tests for all PIR compiler passes.
+-- The argument allows the caller to scale the number of tests.
+-- The default for the argument is @1@.
+typecheck_test ::
+  Int -> TestNested
+typecheck_test factor = return $ testGroup "typechecking"
+  [
+  -- pure passes
+    testProperty "unwrap pass" $
```
michaelpj, line 39: IMO think these tests should live with the tests for the pass in question. I think it's more salient that we're checking property X _of the inliner_ than that we're checking _property X_ of the inliner.

plutus-core/plutus-ir/test/TypecheckSpec.hs
```diff
@@ -0,0 +1,141 @@
...
+    config <- getDefTypeCheckConfig ()
+    inferredType <- runQuoteT $ inferType config tm
+    let evaluated = evaluateBuiltins True (BuiltinsInfo biVariant) def (original tm)
+    runQuoteT $ checkType config () (void evaluated) inferredType
+
+-- | Typechecking a term after a `PlutusIR.Transform.Inline` pass.
+prop_typecheck_inline :: BuiltinSemanticsVariant DefaultFun -> Property
+prop_typecheck_inline biVariant =
+  -- generate type and term in debug mode for easier debugging
+  forAllDoc "ty,tm" genTypeAndTermDebug_ (const []) $ \ (_ty, tm) ->
```
michaelpj, line 135: Does the generated term have globally unique names? If not, we need to rename before we start otherwise the inliner will not work. (Yes, we're totally inconsistent in whether or not we rename inside the pass or not).

plutus-core/plutus-ir/test/TypecheckSpec.hs
```diff
@@ -0,0 +1,141 @@
...
+import PlutusIR.TypeCheck
+import PlutusPrelude (def)
+import Test.QuickCheck (Property, withMaxSuccess)
+import Test.Tasty (testGroup)
+import Test.Tasty.ExpectedFailure
+import Test.Tasty.Extras (TestNested)
+import Test.Tasty.QuickCheck (testProperty)
+
+-- | Typechecking property tests for all PIR compiler passes.
+-- The argument allows the caller to scale the number of tests.
```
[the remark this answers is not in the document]
thealmarty: I was following the convention of some of the other property tests. But yeah I was also wondering if it's needed. I'll remove it.

plutus-core/plutus-ir/test/TypecheckSpec.hs
thealmarty: hmmm...I was thinking along the line of the ticket: these are all typechecking tests for PIR passes. But yeah I could imagine organizing it the other way. I guess the question is what's more likely when we want to examine these tests, in terms of each pass, or in terms of typechecking? I'm not sure.

plutus-core/plutus-ir/test/TypecheckSpec.hs
thealmarty: Sorry I'm not sure what you mean. Did you want me to remove the test for the whole loop? Why?

plutus-core/plutus-ir/test/TypecheckSpec.hs
thealmarty: Oops yeah only `inline` is not pure actually.

plutus-core/plutus-ir/test/TypecheckSpec.hs
thealmarty: Frankly I think I'll just be annoyed by the inconsistency of the layout. Currently the layout is vaguely: `XSpec` modules which have the `X` tests in them. If I disperse the tests into the various passes and leave the code for them in this module it breaks the consistency.

plutus-core/plutus-ir/test/TypecheckSpec.hs
thealmarty: Good point. I'm just gonna `rename` to be sure. Doesn't seem to make the tests pass though.

plutus-core/plutus-ir/test/TypecheckSpec.hs
michaelpj: Well, I'm thinking of these as the first set of pre- and post-condition tests. I think it's a lot more natural to have "property tests showing that pass P obeys its pre- and post-conditions" in the tests for P, rather than spread out amongst N different modules for each property.

plutus-core/plutus-ir/test/TypecheckSpec.hs
michaelpj: We don't test the "simplifier loop as a whole" anywhere else. Also it seems pretty clear to me that if all the individual passes in the simplifier preserve typechecking, then so must the simplifier. I guess it's not bad per se, but if I was going to do an "integration test" for the compiler like this, I'd probably run the _whole_ compiler with pedantic mode on (which typechecks at every step).

plutus-core/plutus-ir/test/TypecheckSpec.hs
thealmarty: OK here are two options:
1. Add a `PrePostPropertyTestSpec.hs` and put all the related tests in there. The hierarchy would be "Tests that a pass meet the pre and post conditions"-> "typechecking".
2. Put all these code in `TransformSpec` and all the upcoming pre and post condition tests as well.

I prefer option 1 because `TransformSpec` is already over 400 lines. It's getting hard to find things in there.

plutus-core/plutus-ir/test/TypecheckSpec.hs
thealmarty: I don't see a strong reason to remove it. I think it makes sense to test more increments when possible. Although I don't mind removing it if it annoys you.

plutus-core/plutus-ir/test/TypecheckSpec.hs
michaelpj: I'm about to split up `TransformSpec` :)

michaelpj: But we can leave it for now, and I can rearrange it later I guess.

thealmarty: Is #5570 merging soon? I can wait. And I can add the harness for the pre- and post-conditions?

thealmarty:
As discussed I've changed the tests to testing that the terms typecheck after a pass. Fewer tests are failing but still there are some. They are marked as expected failures and will be investigated in another PR (see [ticket](https://input-output.atlassian.net/browse/PLT-7746)).

plutus-core/plutus-ir/test/PlutusIR/Transform/CaseReduce/Test.hs
```diff
@@ -0,0 +1,12 @@
+module PlutusIR.Transform.CaseReduce.Test where
```
michaelpj, line 1:
```suggestion
module PlutusIR.Transform.CaseReduce.Tests where
```

plutus-core/plutus-ir/test/PlutusIR/Transform/DeadCode/Tests.hs
```diff
@@ -31,3 +35,14 @@ test_deadCode = runTestNestedIn ["plutus-ir/test/PlutusIR/Transform"] $
             , "recBindingComplex"
             , "pruneDatatype"
             ]
+
+test_typecheck :: TestTree
+test_typecheck = testGroup "typechecking"
+  [ testProperty "Builtin Variant1" $
+  withMaxSuccess 3000
```
michaelpj, line 42: why are we setting this everywhere? is the default setting too low? too high?

plutus-core/plutus-ir/test/PlutusIR/Transform/DeadCode/Tests.hs
```diff
@@ -31,3 +35,14 @@ test_deadCode = runTestNestedIn ["plutus-ir/test/PlutusIR/Transform"] $
...
+  (non_pure_typecheck_prop (removeDeadBindings (BuiltinsInfo DefaultFunSemanticsVariant1)))
+
+  , testProperty "Builtin Variant2" $
```
michaelpj, line 45: I don't think this is really worth it. Which one we use is pretty irrelevant to the dead code pass

plutus-core/plutus-ir/test/PlutusIR/Transform/EvaluateBuiltins/Tests.hs
```diff
@@ -21,3 +26,17 @@ test_evaluateBuiltins = runTestNestedIn ["plutus-ir/test/PlutusIR/Transform"] $
...
+-- pass.
+typecheck_evaluateBuiltins_prop :: BuiltinSemanticsVariant DefaultFun -> Property
+typecheck_evaluateBuiltins_prop biVariant =
+  pure_typecheck_prop (evaluateBuiltins True (BuiltinsInfo biVariant) def)
+
+test_typecheck :: TestTree
+test_typecheck = testGroup "typechecking"
+  [ testProperty "Builtin Variant 1" $
+      withMaxSuccess 3000 (typecheck_evaluateBuiltins_prop DefaultFunSemanticsVariant1)
+  , testProperty "Builtin Variant 2" $
```
michaelpj, line 40: same throughout

plutus-core/plutus-ir/test/PlutusIR/Transform/LetFloatIn/Tests.hs
```diff
@@ -61,3 +65,13 @@ test_letFloatInRelaxed = runTestNestedIn ["plutus-ir/test/PlutusIR/Transform/Let
         _ <- runQuoteT . flip inferType (() <$ pirFloated) =<< TC.getDefTypeCheckConfig ()
         -- letmerge is not necessary for floating, but is a nice visual transformation
         pure $ LetMerge.letMerge pirFloated
+
+-- | Check that a term typechecks after a
+-- `PlutusIR.Transform.LetFloatIn.floatTerm` pass.
+typecheck_floatTerm_prop :: Property
+typecheck_floatTerm_prop =
+  non_pure_typecheck_prop $ LetFloatIn.floatTerm def True
```
michaelpj, line 73: I think we maybe _do_ want to test both ways with the conservative flags for various passes, though! In particular, we can easily generate a random bool and use that as the argument.

plutus-core/plutus-ir/test/PlutusIR/Transform/DeadCode/Tests.hs
michaelpj: Although I guess if you want you could just generate a random one!

plutus-core/plutus-ir/test/PlutusIR/Transform/LetFloatOut/Tests.hs
```diff
@@ -64,3 +68,13 @@ test_letFloatOut = runTestNestedIn ["plutus-ir/test/PlutusIR/Transform"] $
         _ <- runQuoteT . flip inferType (() <$ pirFloated) =<< TC.getDefTypeCheckConfig ()
         -- letmerge is not necessary for floating, but is a nice visual transformation
         pure $ LetMerge.letMerge pirFloated
+
+-- | Check that a term typechecks after a
+-- `PlutusIR.Transform.LetFloatOut.floatTerm` pass.
+typecheck_floatTerm_prop :: Property
+typecheck_floatTerm_prop =
+  pure_typecheck_prop $ LetFloatOut.floatTerm def
```
michaelpj, line 76: here you're not trying both builtin semantics variants! let's be consistent

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
```diff
@@ -0,0 +1,100 @@
...
+import PlutusIR.Compiler
+import PlutusIR.Compiler.Provenance (original)
+import PlutusIR.Core
+import PlutusIR.Generators.QuickCheck.GenerateTerms
+import PlutusIR.TypeCheck
+import PlutusPrelude (($>))
+import Test.QuickCheck (Property, elements)
+import Test.QuickCheck.Arbitrary
+
+instance Arbitrary (BuiltinSemanticsVariant DefaultFun) where
```
michaelpj, line 31: you even wrote this but I don't think you're using it!

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
```diff
@@ -0,0 +1,100 @@
...
+  Right () -> Right ()
+
+-- | Check that a term typechecks after one of the pure passes.
+pure_typecheck_prop ::
+  -- | The pure simplification pass.
+  (Term TyName Name DefaultUni DefaultFun ()
+  -> Term TyName Name DefaultUni DefaultFun ())
+  -> Property
+pure_typecheck_prop pass =
+  -- generate type and term in debug mode for easier debugging
```
michaelpj, line 49: I don't think we're using this? might as well not do it

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
```diff
@@ -0,0 +1,100 @@
...
+  forAllDoc "ty,tm" genTypeAndTermDebug_ (const []) $ \ (_ty, tm) ->
+    convertToEitherString $ do
+      config <- getDefTypeCheckConfig ()
+      -- the generated term may not have globally unique names
+      renamed <- runQuoteT $ rename tm
+      _ <- runQuoteT $ inferType config (pass renamed)
+      pure ()
```
michaelpj, line 56:
```suggestion
     runQuoteT $ inferType config (pass renamed)
```

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
```diff
@@ -0,0 +1,100 @@
...
+  -> Term TyName Name DefaultUni DefaultFun ())
+  -> Property
+pure_typecheck_prop pass =
+  -- generate type and term in debug mode for easier debugging
+  forAllDoc "ty,tm" genTypeAndTermDebug_ (const []) $ \ (_ty, tm) ->
+    convertToEitherString $ do
+      config <- getDefTypeCheckConfig ()
+      -- the generated term may not have globally unique names
+      renamed <- runQuoteT $ rename tm
+      _ <- runQuoteT $ inferType config (pass renamed)
```
michaelpj, line 55: You should be running all the `QuoteT` stuff together, otherwise you might generate duplicate names. I think that's safe here, but might not be in general

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
```diff
@@ -0,0 +1,100 @@
...
+-- Convert Either Error () to Either String () to match with the Testable (Either String ())
+-- instance.
+convertToEitherString :: Either (Error DefaultUni DefaultFun ()) ()
+  -> Either String ()
+convertToEitherString = \case
+  Left err -> Left $ show err
+  Right () -> Right ()
+
+-- | Check that a term typechecks after one of the pure passes.
+pure_typecheck_prop ::
```
michaelpj, line 43: we don't generally use snake-case?

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
```diff
@@ -0,0 +1,100 @@
...
+      (Term TyName Name DefaultUni DefaultFun a))
+  -> Property
+non_pure_typecheck_prop pass =
+  -- generate type and term in debug mode for easier debugging
+  forAllDoc "ty,tm" genTypeAndTermDebug_ (const []) $ \ (_ty, tm) ->
+    convertToEitherString $ do
+      config <- getDefTypeCheckConfig ()
+      -- the generated term may not have globally unique names
+      renamed <- runQuoteT $ rename tm
+      processed <- runQuoteT $ pass (original renamed)
```
michaelpj, line 71: okay, this one seems potentially dangerous

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
```diff
@@ -0,0 +1,100 @@
...
+      (Term TyName Name DefaultUni DefaultFun a))
+  -> Property
+extra_constraint_typecheck_prop pass =
+  -- generate type and term in debug mode for easier debugging
+  forAllDoc "ty,tm" genTypeAndTermDebug_ (const []) $ \ (_ty, tm) ->
+    convertToEitherString $ do
+      config <- getDefTypeCheckConfig ()
+      plcConfig <- PLC.getDefTypeCheckConfig ()
+      renamed <- runQuoteT $ rename tm
+      let processed =
```
michaelpj, line 91: ditto

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
```diff
@@ -0,0 +1,100 @@
...
+      processed <- runQuoteT $ pass (original renamed)
+      _ <- runQuoteT $ inferType config (processed $> ())
+      pure ()
+
+-- | Check that a term typechecks after a non-pure pass that requires extra constraints.
+extra_constraint_typecheck_prop ::
+  (Term TyName Name DefaultUni DefaultFun (Provenance ())
+  -> QuoteT
+      (ReaderT
+         (CompilationCtx DefaultUni DefaultFun c)
```
michaelpj, line 80: This is perhaps too coarse, but fine for now.

plutus-core/plutus-ir/test/PlutusIR/Transform/DeadCode/Tests.hs
thealmarty: Funny enough the CI failed the second variant only (it passed locally). The only reason I'm not generating a random one is that I can't mark a specific variant as expected failure.

plutus-core/plutus-ir/test/PlutusIR/Transform/DeadCode/Tests.hs
thealmarty: I was using it to change the max indiviually when sometimes some of them pass locally. With the weaker typechecking it's not needed maybe.

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
thealmarty: `inferType` returns a `(Normalized (Type TyName uni ()))`, not `()`.

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
thealmarty: I was under the impression that we do in tests - maybe not.

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
thealmarty: It won't compile without it. And I'm not using it because it makes it hard to mark expect failures. Although I can use it for the ones without failures atm. But then I'm worried it's inconsistent/messy. What do you think?

plutus-core/plutus-ir/test/PlutusIR/Transform/LetFloatIn/Tests.hs
thealmarty: OK I tried just letting `tasty-discover` detect this prop with `Bool` as argument. The default number of tests is only 100, which is way too small. I just wrote it out again - that way we can see which mode fails which is useful.

plutus-core/plutus-ir/test/PlutusIR/Transform/LetFloatOut/Tests.hs
thealmarty: Since we have some evidence of different results with different variants, I'm gonna try both for all tests.

plutus-core/plutus-ir/test/PlutusIR/Transform/DeadCode/Tests.hs
michaelpj: I'm happy to just mark the whole thing as expected to fail in that case! A property test fails if you find a counterexample with some combination of random inputs. I don't see why the semantics variant should be different.

plutus-core/plutus-ir/test/PlutusIR/Transform/LetFloatIn/Tests.hs
michaelpj: Then I do think this is a case to run with more tests. We should just need twice as many tests!

plutus-core/plutus-ir/test/PlutusIR/Transform/LetFloatOut/Tests.hs
michaelpj: Maybe you're catching my drift by now, which is: just generate all the arguments, configuration included!

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
michaelpj: > It won't compile without it

That doesn't tell me anything - why doesn't it compile? I think you mean "we're using it somewhere", but where?

Not that it matters much, since I'm suggesting you _do_ use it. I think it's fine as an instance, I just think we should be exactly consistent and generate it everywhere.

plutus-core/testlib/PlutusIR/Properties/Typecheck.hs
michaelpj: Take a look! We use it just for the prefix for tasty-discover, which I think is using snake-case for that part precisely because it's unusual.

thealmarty: I don't think it's just `tasty-discover`, if you look at the `QuickCheck` documentation it uses snake-case too. I'm happy to stick to only using snake-case for tests/properties for `tasty-discover` though.

plutus-core/plutus-ir/test/PlutusIR/Transform/DeadCode/Tests.hs
thealmarty: I don't see why either. But remember we should throw away the assumption that we know what's happening! That's the whole reason we have tests! Yes it does make the code bigger this way but I think it makes sense to get as much fine grain detail as possible so we can debug more easily. With the random arguments all in one property tests, if it encounters one counter example it stops. Maybe I can use some of the functions in [this section](https://hackage.haskell.org/package/QuickCheck-2.14.3/docs/Test-QuickCheck.html#g:21)? It does seem wrong and silly to not use the random input feature but I feel I'm losing info/control if I do. @james-iohk any advice?

michaelpj: I'm confused: what specifically do you think is going to happen here? I think: quickcheck will find the failing example and fail, thus proving a counterexample to the more-general property that this should work for all semantics variants. We can then investigate that failure.

Are you worried that the builtin semantics variant generator won't explore both options? Then you could use `label` or `cover` to check, but the generator is so simple I don't see the worry.  

> But remember we should throw away the assumption that we know what's happening! 

No, I'm saying that I know what is happening with _quickcheck_. I'm not saying I know what is happening with our code.

The whole point of quickcheck is to generate random arguments to functions and then run them. It's good at that, let's let it do its job!

> With the random arguments all in one property tests, if it encounters one counter example it stops.

So? We only need one counterexample at a time.

thealmarty: >The whole point of quickcheck is to generate random arguments to functions and then run them. It's good at that, let's let it do its job!

I agree, hence my comment "It does seem wrong and silly to not use the random input feature"

Let's say we got a counter example, how do we start debugging? Do we know it fails because it's from a specific argument? If we test every argument separately, then we have some idea of that, because we'll know if e.g., the other variant or the other mode also fails or not. I was hoping I wouldn't lose this.

Perhaps I'm too worried about this. I'll just do the regular thing. People can do whatever they want to debug the failed tests.

michaelpj: The counterexample will show the arguments, so you can just run it with those arguments to get the failure. And quickcheck will try and shrink them also. We might want to think about how we want the configuration arguments to shrink, but we can worry about that later.

plutus-core/plutus-ir/test/PlutusIR/Compiler/Let/Test.hs
```diff
@@ -2,11 +2,44 @@ module PlutusIR.Compiler.Let.Test where
...
 test_lets :: TestTree
 test_lets = runTestNestedIn ["plutus-ir/test/PlutusIR/Compiler"] $ testNested "Let"
     [ goldenPlcFromPir pTermAsProg "letInLet"
     , goldenPlcFromPir pTermAsProg "letDep"
     ]
+
+-- FIXME
+-- | Check that a term typechecks after a
+-- `PlutusIR.Compiler.Let.compileLets` (recursive terms) pass.
+prop_TypecheckCompileLetsRecTerms :: Property
```
thealmarty, line 22: I could write an arbitrary instance for [LetKind](https://github.com/input-output-hk/plutus/blob/f36d6fc5521f8740f71de109466bade38c9d10b9/plutus-core/plutus-ir/src/PlutusIR/Compiler/Let.hs#L39) too but I feel in this case it's more obviously useful to separate them. Let me know if you think otherwise.

michaelpj: Yeah, in this case it actually means "do something completely different", so I think it's good to separate them.

plutus-core/plutus-ir/test/PlutusIR/Compiler/Let/Test.hs
```diff
@@ -2,11 +2,44 @@ module PlutusIR.Compiler.Let.Test where
...
+prop_TypecheckCompileLetsTypes :: Property
+prop_TypecheckCompileLetsTypes =
+  withMaxSuccess 10000 $ extraConstraintTypecheckProp (compileLets Types)
+
+-- FIXME this test sometimes fails so ignoring it to make CI pass.
+-- | Check that a term typechecks after a
+-- `PlutusIR.Compiler.Let.compileLets` (data types) pass.
+test_TypecheckCompileLetsDataTypes :: TestTree
+test_TypecheckCompileLetsDataTypes =
+  ignoreTest $ testProperty "typechecking" $
```
michaelpj, line 43: So the reason we can't say it's expected to fail is because it sometimes passes?

michaelpj (approved):

michaelpj:
Looks like the letrec test also fails sometimes.

plutus-core/plutus-ir/test/PlutusIR/Compiler/Let/Test.hs
thealmarty: Yeah.

PR #5566, opened 2023-09-28
[PlutusTx] [Test] 'Eq' for 'Value'

effectfully:
This adds some budget tests for `(==) @Value`.

Needs docs, don't look here yet.


effectfully (commented):
After redoing it several times the current state of the PR is what I settled on. 

No property tests, because so far I'm only concerned with making sure that whatever I implement as a better version of `(==) @Value` is indeed faster. Plus I need to benchmark my version against not only the existing one, but also the winner from #5135 (`valueEqualsValue4`). I hope mine's gonna be faster, but you never know.

effectfully:
This is ready for review, but

1. CI throws random nix errors that I don't understand
2. there are some true errors related to name duplication etc, but it's hard to dig them out because of 1
3. I need to create golden files per GHC version, which I just realized

I'll handle 2 and 3 and they shouldn't really affect reviews, so please review. I'll need help with 1, I've no idea what's going on there.

plutus-tx/src/PlutusTx/AssocMap.hs
```diff
@@ -149,6 +152,10 @@ instance (Pretty k, Pretty v) => Pretty (Map k v) where
 fromList :: [(k, v)] -> Map k v
 fromList = Map
 
+{-# INLINEABLE fromListSafe #-}
+fromListSafe :: Eq k => [(k, v)] -> Map k v
```
michaelpj, line 156: unused?

michaelpj:
It's a bit hidden, but you can eventually find that it's a failure to solve for the dependencies on 8.10: https://ci.iog.io/build/1252601/nixlog/1

This is because the plugin is not buildable on < 9.2, so we now can't build the `plutus-ledger-api` tests on 8.10, which we were previously able to do. I'm unsure about that - thus far we've kept `plutus-ledger-api` entirely free of the plugin. But annoyingly `Value` is defined there, so not sure what else to do. We could either just accept that the tests don't work on 8.10, or make another test component that has stuff that depends on the plugin and doesn't build on 8.10.

plutus-ledger-api/test/Spec/Value.hs
```diff
@@ -0,0 +1,231 @@
...
+haskellEqValue value1 value2 = toMap value1 Haskell.== toMap value2 where
+    toMap
+        = Map.filter (Haskell.not . Map.null)
+        . Haskell.fmap (Map.filter (Haskell./= 0))
+        . Map.fromListWith (Map.unionWith (Haskell.+))
+        . Haskell.map (Haskell.fmap $ Map.fromListWith (Haskell.+))
+        . valueToLists
+
+-- | Check whether all currencies and tokens within each of the currencies occur uniquely.
+allDistinct :: Value -> Bool
```
michaelpj, line 182: Can't you use your `safeFromList` to avoid needing this?

michaelpj (approved):

michaelpj:
To be clear, to get stuff to be ignored for 8.10 you just have to do what we do elsewhere and add an GHC version conditional guarding `buildable: False`.

plutus-ledger-api/test/Spec/Value.hs
```diff
@@ -0,0 +1,231 @@
...
+    (errOrRes, cost)
+        = PLC.runCekNoEmit PLC.defaultCekParameters PLC.counting
+        . PLC.runQuote
+        . PLC.unDeBruijnTermWith (Haskell.error "Free variable")
+        . PLC._progTerm
+        $ getPlc prog
+    res = either Haskell.throw id $ errOrRes >>= PLC.readKnownSelf
+
+-- | Check equality of two compiled 'Value's directly in Haskell.
+haskellEqValue :: Value -> Value -> Bool
```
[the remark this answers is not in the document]
effectfully: > Or is the whole point to have another implementation?

Exactly, I'm testing the actual implementation against a much simpler one, not just computing the budgets.

plutus-ledger-api/test/Spec/Value.hs
```diff
@@ -0,0 +1,231 @@
...
+        Haskell.and . Map.fromListWith (\_ _ -> False) . Haskell.map (Haskell.fmap $ \_ -> True))
+    . valueToLists
+
+-- | Return all the pairs of elements of the given list.
+--
+-- > (x, y) `elem` pairs xs ==> fromJust (x `elemIndex` xs) <= fromJust (y `elemIndex` xs)
+--
+-- >>> pairs "abc"
+-- [('a','a'),('a','b'),('b','b'),('b','c'),('c','c')]
+pairs :: [a] -> [(a, a)]
```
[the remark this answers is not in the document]
effectfully: Nope, that would produce

```haskell
[('a','a'),('a','b'),('a','c'),('b','a'),('b','b'),('b','c'),('c','a'),('c','b'),('c','c')]
```

not

```haskell
[('a','a'),('a','b'),('b','b'),('b','c'),('c','c')]
```

I'll rename it to `pairsUnord` or something.

plutus-ledger-api/test/Spec/Value.hs
```diff
@@ -0,0 +1,231 @@
...
+                    ]
+                else Pretty.group $ Pretty.pretty budget
+
+test_EqValue :: TestTree
+test_EqValue =
+    runTestNestedIn ["test", "Spec"] $
+        testNestedGhc "golden"
+            [ test_EqCurrencyList "Short" currencyListOptions
+            , test_EqCurrencyList "Long" currencyLongListOptions
+            ]
```
effectfully, line 231: I don't like these tests... They cover some interesting corner cases, but not nearly enough. I'll make a ticket for improving them.

plutus-tx/src/PlutusTx/AssocMap.hs
effectfully: Yep, but I believe it's worth having it if just to emphasize that the default version isn't safe.

plutus-ledger-api/test/Spec/Value.hs
effectfully: Maybe, but the point is that it's for tests and I want an alternative implementation for testing purposes that does not rely on anything from `PlutusTx.AssocMap` and is self-contained etc.

effectfully:
@michaelpj I've separated the new tests into a separate stanza, but CI still seems to complain. Any suggestions what I should do next?

michaelpj:
I pushed a couple of commits:
1. Annoying Nix stuff since we need to turn off anything using the plugin when cross-compiling.
2. Some unnecessary dependencies that were causing a `-Werror` failure

PR #5568, opened 2023-09-29
Add empty string check for git command in ci-plutus-benchmark.sh

zeme-wana:
It's possible that one has to `git pull --rebase origin master` first in order for the `ci-plutus-benchmark.sh` to work in one's PR.

This PR adds a check that `git merge-base HEAD origin/master` does not return the empty string at the top of the `ci-plutus-benchmark.sh` file.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -20,6 +20,8 @@ jobs:
         with:
           # We need at least one commit before master to compare against
           fetch-depth: 5
+          # And we want to prevent `git reset --hard HEAD` before fetching
+          clean: false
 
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark.yml
```diff
@@ -20,6 +20,8 @@ jobs:
         with:
           # We need at least one commit before master to compare against
           fetch-depth: 5
+          # And we want to prevent `git reset --hard HEAD` before fetching
```
michaelpj, line 23: why?

michaelpj: nag

PR #5577, opened 2023-10-06
[PlutusTx] Optimize the 'Eq' instance of 'Value'

effectfully:
Do not look here, currently this PR contains a lot of irrelevant noise, which will be removed once #5566 is in.


plutus-tx/src/PlutusTx/SortedMap.hs
```diff
@@ -0,0 +1,78 @@
...
+-- is, the function throws).
+unsafeInsertOneUnique :: forall k v. Ord k => k -> v -> SortedMap k v -> SortedMap k v
+unsafeInsertOneUnique ~k0 ~v0 = coerce go where
+    go :: [(k, v)] -> [(k, v)]
+    go []                  = [(k0, v0)]
+    go kvs@((k, v) : kvs') =
+        case k0 `compare` k of
+            LT -> (k0, v0) : kvs
+            -- TODO: make this @traceError duplicateElements@.
+            EQ -> (k, v0) : kvs'
```
effectfully, line 49: I'm not making the change in this PR, so that the budgets comparison is fair. I'll make the change and I'll update the golden tests in general, since the current ones were written with the assumption that duplicate currencies/tokens are allowed.

effectfully:
This fixes the performance part of #5135, but we need tests for this stuff too, it's planned for a future PR.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -368,6 +364,118 @@ split (Value mp) = (negate (Value neg), Value pos) where
...
     (l, r) = Map.mapThese (\i -> if i <= 0 then This i else That i) mp'
 
+-- | The type of values that 'matchKVs' returns. See its Haddock for details.
+data MatchResult k v
+    = MatchSuccess
+    | MatchPartial [(v, v)] [(k, v)] [(k, v)]
+    | MatchFailure
+
+{-# INLINE matchKVs #-}
+{- | Take a function checking whether a value is zero\/empty, a function checking /structural/
```
bezirg, line 374:
```suggestion
{- | Takes
```

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -368,6 +364,118 @@ split (Value mp) = (negate (Value neg), Value pos) where
...
+equality of two values and two key-value lists and return the result of matching the lists
```
bezirg, line 375:
```suggestion
returns...
```

plutus-tx/src/PlutusTx/SortedMap.hs
```diff
@@ -0,0 +1,79 @@
...
+    { unSortedMap :: [(k, v)]
+    } deriving stock (Haskell.Show)
+
+{-# INLINABLE empty #-}
+empty :: SortedMap k v
+empty = UnsafeSortedMap []
+
+{-# INLINE singleton #-}
+singleton :: k -> v -> SortedMap k [v]
+singleton ~k ~v = UnsafeSortedMap [(k, [v])]
```
bezirg, line 36: Why the arguments to this function are lazy? Also aren't we supposed to use Strict nowadays to the whole of plutustx prelude?

plutus-tx/src/PlutusTx/SortedMap.hs
```diff
@@ -0,0 +1,79 @@
...
+    go [] kvs2 = all (is0 . snd) kvs2  -- If one of the lists is empty then all elements in the
+    go kvs1 [] = all (is0 . snd) kvs1  -- other list need to be zero for lists to be equal.
+    go ((k1, v1) : kvs1) ((k2, v2) : kvs2) =
+        -- As with 'matchKVs' we check equality of all the keys first and only then check equality
+        -- of values.
+        if k1 == k2
+            then if go kvs1 kvs2
+                then eqV v1 v2
+                else False
+            else False
```
bezirg, line 79:
```suggestion
        if k1 == k2 && go kvs1 kvs2
        then eqV v1 v2
        else False
```

plutus-tx/src/PlutusTx/SortedMap.hs
```diff
@@ -0,0 +1,79 @@
...
+eqWith
+    :: forall k v. Eq k
+    => (v -> Bool) -> (v -> v -> Bool) -> SortedMap k v -> SortedMap k v -> Bool
+eqWith ~is0 ~eqV = coerce go where
+    go :: [(k, v)] -> [(k, v)] -> Bool
+    go [] []   = True
+    go [] kvs2 = all (is0 . snd) kvs2  -- If one of the lists is empty then all elements in the
+    go kvs1 [] = all (is0 . snd) kvs1  -- other list need to be zero for lists to be equal.
+    go ((k1, v1) : kvs1) ((k2, v2) : kvs2) =
+        -- As with 'matchKVs' we check equality of all the keys first and only then check equality
```
bezirg, line 73: This looks like it shares some logic with matchKVs , but I don't have any suggestion on how to share it. Actually maybe it is better to keep them separate

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -368,6 +364,118 @@ split (Value mp) = (negate (Value neg), Value pos) where
...
+        MatchPartial valPairs currs1' currs2' ->
+            if SortedMap.eqWith
+                    (Map.all (== 0))
+                    (eqMapVia SortedMap.unsafeFromListUnique)
+                    (SortedMap.unsafeFromListUnique currs1')
+                    (SortedMap.unsafeFromListUnique currs2')
+                then
+                    -- Check equality of values that come from the common key prefix of the original
+                    -- lists.
+                    all (uncurry (==)) valPairs
```
bezirg, line 466: Here valpairs are maps themselves. Don't you have to do proper sorted equality check (i.e. eqMapVia)?

plutus-tx/src/PlutusTx/SortedMap.hs
```diff
@@ -0,0 +1,79 @@
...
+unsafeFromListUnique :: forall k v. Ord k => [(k, v)] -> SortedMap k v
+unsafeFromListUnique = go where
+    go :: [(k, v)] -> SortedMap k v
+    go []             = UnsafeSortedMap []
+    go ((k, v) : kvs) = unsafeInsertOneUnique k v $ go kvs
+
+-- The pragma trades a bit size for potentially plenty of budget.
+{-# INLINE eqWith #-}
+-- | Check equality of 'SortedMap's by matching the underlying lists pointwise.
+eqWith
```
michaelpj, line 64: This is a bit `Value`-specific: the `is0` function only really makes sense if you are working with a map that has the structure that `Value` does, as opposed to a generic sorted map. That's not necessarily a problem but we should say so.

plutus-tx/src/PlutusTx/SortedMap.hs
michaelpj: This will perform worse, remember that `&&` is not lazy

michaelpj (approved):
Given how many intermediate structures this must create, I'm amazed it performs okay, but I guess that just means the old one was truly appalling :grimacing:

plutus-tx/src/PlutusTx/SortedMap.hs
bezirg: Omg i forgot that (&&) in plutustx is not short-circuitting

michaelpj: application is still strict. This is a weird corner and I think Ziyang wants to make it somehow work, but at the moment `f x y` evaluates x and y always.

plutus-tx/src/PlutusTx/SortedMap.hs
```diff
@@ -0,0 +1,79 @@
...
+{-# INLINE unsafeFromListUnique #-}
+-- | Turn a list into a 'SortedMap' assuming all of its keys are unique (if they are not, the
+-- function throws).
+unsafeFromListUnique :: forall k v. Ord k => [(k, v)] -> SortedMap k v
+unsafeFromListUnique = go where
+    go :: [(k, v)] -> SortedMap k v
+    go []             = UnsafeSortedMap []
+    go ((k, v) : kvs) = unsafeInsertOneUnique k v $ go kvs
+
+-- The pragma trades a bit size for potentially plenty of budget.
```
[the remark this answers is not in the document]
effectfully: Screenshot from 2023-10-10 18-07-07

> Maybe GHC is doing something?

Quite likely.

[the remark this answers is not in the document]
effectfully: ... and

Screenshot from 2023-10-10 18-14-09

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
effectfully: I like the first sentence to have imperative mood. I don't mind changing it, but I do that for virtually every Haddock that I wrote, so if I do it only here, it would only make things inconsistent.

I don't mind changing this habit, if it's something that we want. @zliu41 @michaelpj do you have an opinion?

plutus-tx/src/PlutusTx/SortedMap.hs
```diff
@@ -0,0 +1,79 @@
...
+-- | Check equality of 'SortedMap's by matching the underlying lists pointwise.
+eqWith
+    :: forall k v. Eq k
+    => (v -> Bool) -> (v -> v -> Bool) -> SortedMap k v -> SortedMap k v -> Bool
+eqWith ~is0 ~eqV = coerce go where
+    go :: [(k, v)] -> [(k, v)] -> Bool
+    go [] []   = True
+    go [] kvs2 = all (is0 . snd) kvs2  -- If one of the lists is empty then all elements in the
+    go kvs1 [] = all (is0 . snd) kvs1  -- other list need to be zero for lists to be equal.
+    go ((k1, v1) : kvs1) ((k2, v2) : kvs2) =
```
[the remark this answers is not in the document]
kwxm: Later: I think this is actually a bug.  See the example in Value.hs.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -368,6 +364,118 @@ split (Value mp) = (negate (Value neg), Value pos) where
...
+                MatchFailure -> MatchFailure
+        -- If the keys aren't equal, then maybe the first value is empty, in which case we throw it
+        -- out and proceed.
+        | is0 v1 = go kvs1' kvs2
+        -- Or if the second one is empty, then we throw that one and proceed.
+        | is0 v2 = go kvs1 kvs2'
+        -- Otherwise the keys in the lists have diverged and we return the remaining parts.
+        | otherwise = MatchPartial [] kvs1 kvs2
+
+{-# INLINEABLE eq #-}
```
kwxm, line 444: I think this doesn't actually work: there's a problem with non-trailing entries with zero values.  In the code below, `v1 == v2` is `False` but `v2 == v3` is `True`.

```

import PlutusLedgerApi.V1.Value
import PlutusTx.AssocMap as Map
    
ada :: CurrencySymbol
ada = ""

a :: TokenName
a = "aa"

b :: TokenName
b = "bb"

c :: TokenName
c = "cc"
    
v1 :: Value
v1 = Value $ Map.fromList [(ada, Map.fromList [(a,0),(b,100)])]

v2 :: Value
v2 = Value $ Map.fromList [(ada, Map.fromList [(b,100)])]

v3 :: Value
v3 = Value $ Map.fromList [(ada, Map.fromList [(b,100),(c,0)])]
```

kwxm (commented):
I think there's still a bug in this.  I found it quite hard to understand what the code was supposed to be doing and I wonder if we could try to convert it to Agda to get some more assurance of correctness.  This would be a good example to try because it's pretty much standalone code where we don't have to rely on lots of external stuff.  @mjaskelioff do you have an opinion on this?

plutus-tx/src/PlutusTx/SortedMap.hs
effectfully: > Oh wait, it is short-circuitting , I don't know then what you mean

Only if GHC inlines it, which we probably can't 100% rely upon.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
effectfully: Thank you for spotting the bug, it's fixed now.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -368,6 +364,118 @@ split (Value mp) = (negate (Value neg), Value pos) where
...
+    -- non-empty lists only consists of empty values.
+    go [] kvs2 = if all (is0 . snd) kvs2 then MatchSuccess else MatchFailure
+    -- Symmetric to the previous case.
+    go kvs1 [] = if all (is0 . snd) kvs1 then MatchSuccess else MatchFailure
+    -- Both spines are non-empty.
+    go kvs1@((k1, v1) : kvs1') kvs2@((k2, v2) : kvs2')
+        -- If keys are equal
+        | k1 == k2 =
+            -- then continue checking equality of spines.
+            case go kvs1' kvs2' of
```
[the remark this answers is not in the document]
effectfully: It does a bit, as per #5530, not having tail recursion is better than having an extra argument, for example. So even if it costs anything, it's pretty negligible still.

plutus-tx/src/PlutusTx/SortedMap.hs
effectfully: Moved it to `Value.hs` and got rid of `SortedMap` entirely.

plutus-tx/src/PlutusTx/SortedMap.hs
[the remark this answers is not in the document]
effectfully: Now in the same file.

plutus-tx/src/PlutusTx/SortedMap.hs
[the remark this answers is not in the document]
effectfully: *a bit of size. Fixed.

plutus-tx/src/PlutusTx/SortedMap.hs
```diff
@@ -0,0 +1,79 @@
...
+        case k0 `compare` k of
+            LT -> (k0, v0) : kvs
+            -- TODO: make this @traceError duplicateElements@.
+            EQ -> (k, v0) : kvs'
+            GT -> (k, v) : go kvs'
+
+{-# INLINE unsafeFromListUnique #-}
+-- | Turn a list into a 'SortedMap' assuming all of its keys are unique (if they are not, the
+-- function throws).
+unsafeFromListUnique :: forall k v. Ord k => [(k, v)] -> SortedMap k v
```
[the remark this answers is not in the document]
effectfully: > Did you try any other sorting method here?

Yes, I've tried the one we have in `PlutusTx.List` and it basically doesn't finish. It's order of magnitudes worse, I've tested it on much smaller inputs and it still was much worse. I've no idea why, I investigated it shortly, but couldn't quickly find the culprit, so I moved on. I'll make a ticket.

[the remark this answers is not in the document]
effectfully: > Also, would it be worth trying to be tail recursive here and/or in the insertion function?

According to #5530 probably not.

plutus-ledger-api/test-plugin/Spec/golden/9.2/Long.stat.golden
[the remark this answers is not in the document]
effectfully: > Do you know how big the compiled function is, eg as a fraction of the maximum transaction size (16384 bytes)?

I don't know, but here's the diff of [plutus-benchmark/marlowe/test/semantics/semantics.size.golden](https://github.com/input-output-hk/plutus/pull/5577/files#diff-8d29f73c346b97d05a7c8a7940d9cbdb90a59b2e52763eec7dfdcc82edb6fadd) in this PR:

```diff
@@ -1 +1 @@
-13603
+14911
```

Does not look very good... Perhaps I should optimize for size more than for CPU and memory. I'll look into it in a future PR.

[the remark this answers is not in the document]
effectfully: I've managed to make it

```diff
-14911
+14617
```

It's what I get if I remove all `INLINE` pragmas. So I need to either define `eqKVs` (used to be `eqWith`) in terms of `matchKVs` to share more code or change the algorithm, if I want to bring that number further down. I have an idea on how to improve the algorithm generally, so I'll do that in a subsequent PR (after adding tests) instead of dumbing the current algorithm down in this PR.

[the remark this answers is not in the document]
effectfully: Ah, the cost for that size reduction is

```diff
index f5a2f50f0..93fa88acb 100644
--- a/plutus-benchmark/script-contexts/test/checkScriptContextEqualityTerm-20.eval.golden
+++ b/plutus-benchmark/script-contexts/test/checkScriptContextEqualityTerm-20.eval.golden
@@ -1,2 +1,2 @@
-({cpu: 243201611
-| mem: 951066})
\ No newline at end of file
+({cpu: 268570611
+| mem: 1061366})
```

So not very good, but not too bad either. Size is probably more important.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -368,6 +364,118 @@ split (Value mp) = (negate (Value neg), Value pos) where
...
+  where
+    -- Check equality of two @Map@s given a function converting a list to a 'SortedMap' (two options
+    -- are embedding the list directly to check structural equality or sorting the list first).
+    eqMapVia
+        :: ([(TokenName, Integer)] -> SortedMap.SortedMap TokenName Integer)
+        -> Map.Map TokenName Integer
+        -> Map.Map TokenName Integer
+        -> Bool
+    eqMapVia asSortedMap (Map.toList -> tokens1) (Map.toList -> tokens2) =
+        SortedMap.eqWith (== 0) (==) (asSortedMap tokens1) (asSortedMap tokens2)
```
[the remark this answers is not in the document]
effectfully: I've removed the `INLINE` pragma from that function, it did help a bit, see the thread above. Thanks for bringing it up!

plutus-tx/src/PlutusTx/SortedMap.hs
```diff
@@ -0,0 +1,79 @@
...
+unsafeFromListUnique = go where
```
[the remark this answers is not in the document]
effectfully: Probably not. We still have to keep this logic for `matchKVs` and replicating it for `eqKVs` is straightforward, while making a perfectly generic insertion sort aware of zeroes is kinda nasty.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
effectfully: Yup, thanks a lot for debugging the issue and providing the example. It's a bug indeed: I forgot to replicate the `is0` logic of `matchKVs` in `eqKVs`. My imaginary tests would catch the bug, but I'm rushing (for PI planning reasons) and imaginary tests don't get run in the real world, unfortunately. I've fixed the bug now and I'll add proper tests in a subsequent PR.

effectfully:
I addressed most of the comments. I'm going to hit "merge", so that we can claim this is done and I'll start working on proper property tests for it, then I'll revise the unit tests that we have and that were written with a different set of constraints in mind, then I'll try tweaking the main algorithm to make it smaller and possibly more efficient.

michaelpj:
I think maybe we shouldn't have merged this. It's tricky code and we found some bugs, so we probably want to make sure we get the tests in. But we're doing releases every sprint now! So it's quite possible that this could get into a release in a buggy state. In fact, I think @kwxm is planning to do a release today? So I think we should make sure this isn't in the release, or delay the release until the tests land.

kwxm:
> I think @kwxm is planning to do a release today?

Probably tomorrow (Thursday).

effectfully:
I've reverted the merge. @kwxm feel free to create a new release, everything should be fine. I'll reopen the PR or open another one once I have tests.

PR #5579, opened 2023-10-09
Just call adjustModel once

kwxm:
This involves a small refactoring of the R cost modelling code, prompted by PLT-881.  It's possible that we generate a modelling function that has negative slope (eg, if the time is more or less constant) or negative intercept (eg, if the time is linear in x but the regression line happens to cross the t-axis below the origin).  There's a function called `adjustModel` that checks the coefficients of the generated models and replaces any non-positive values with a small positive value in order to avoid the possibility of negative costs.  That was previously called at the end of the functions that generate the individual models, but in this PR I've changed the code so that it's mapped over the full set of models before they're returned to the outside world (which is a bit tricky to do neatly because of R's weird `*apply`functions:  maybe I just haven't found the best way to do it). That should make it less easy to forget the check.  I've also taken to opportunity to be more explicit about what's returned by some functions.  

I've tested this by running `generate-cost-model` and comparing the output with `builtinCostModel.json` (there are small differences due to re-generated costs for the BLS12-381 builtins),  and also by running `cost-model-test`.

### DIFF
--- a/plutus-core/cost-model/data/models.R
+++ b/plutus-core/cost-model/data/models.R
@@ -206,6 +206,20 @@ adjustModel <- function (m, fname) {
     return (m)
 }
 
+adjustModels <- function(models) {
+    ## Given a list of named models, apply `adjustModel` to each entry to
+    ## ensure that all of the model coefficients are positive.  This is applied
+    ## to the entire list of models before returning them to the outside world.
+    adjustEntry <- function(name) {
+        adjustModel(models[[name]], sub("Model", "", name))
+    }
+    models2 <- lapply(names(models), adjustEntry)
+    names(models2) <- names(models)
+    # Can we just mutate the entries of `models` in place, leaving the names untouched?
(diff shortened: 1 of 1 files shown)


plutus-core/cost-model/data/models.R
```diff
@@ -374,8 +390,7 @@ modelFun <- function(path) {
         filtered <- data %>%
             filter.and.check.nonempty (fname)  %>%
             discard.overhead ()
-        m <- lm(t ~ pmax(x_mem, y_mem), filtered)
-        adjustModel(m, fname)
+        lm(t ~ pmax(x_mem, y_mem), filtered)
     }
```
kwxm, line 394: We *don't* want `return` here.  This is like a nested let-expression whose value is the final line.  If you put `return` here then you return from the enclosing function, which definitely isn't what we want (as I discovered by accident...).

michaelpj (approved):

PR #5584, opened 2023-10-12
PLT-7745: Basic rewrite rules for builtins

bezirg:
This is something like GHC's RULES, but for PIR. The complication is that the rewrite pass is supposed to be parameterized over the universe, which requires some boilerplate similar to the BuiltinsInfo boilerplate.

More rewrite rules to come. I just want to get this into master so I can further add more rules.
Also turned the previous CommuteFnWithConst transformation to a RewriteRule

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


bezirg:
Can you take a first look? I am going to refactor it a bit and add comments.

Also I am thinking to incorporate commuteFnWithConst  pass as a rewrite rule,
and add more rules.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
```diff
@@ -1,19 +1,29 @@
...
 
+import Data.Monoid
 import Control.Lens.TH
 import Data.Kind
 import PlutusCore.Builtin
 import PlutusCore.Builtin qualified as PLC
 import PlutusPrelude (Default (..))
+import PlutusIR qualified as PIR
+
+newtype RewriteRules uni fun = RewriteRules {
```
bezirg, line 14: i am not fond of this newtype, but I had to do it to make makeLenses happy.

Also i could make derive-newtype semigroup,monoid but then I had to write custom instances.

plutus-core/plutus-ir/src/PlutusIR/Transform/Rewrite.hs
```diff
@@ -0,0 +1,36 @@
...
+    let RewriteRules f = bi^.rewriteRules
+    in transformOf termSubterms (appEndo $ getDual f) t
+
+defaultUniRewriteRules :: RewriteRules DefaultUni DefaultFun
+defaultUniRewriteRules = RewriteRules $ combineRules
+    -- rules are applied from left to right because of Dual
+    [ decodeEncode
+    ]
+  where
+    combineRules = foldMap (Dual . Endo)
```
michaelpj, line 27: If it were me, I'd probably make `RewriteRules` instead hold a `[Term -> Term]`, and then apply them here. It doesn't matter too much though, I guess.

plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
michaelpj: I'm confused by the makeLenses problem?

michaelpj: I would have thought you could just derive semigroup and monoid via `Dual Endo` whatever if that's what you need.

plutus-core/plutus-ir/src/PlutusIR/Transform/Rewrite.hs
```diff
@@ -0,0 +1,36 @@
...
+            -> t
+            -> t
+userRewrite bi t =
+    let RewriteRules f = bi^.rewriteRules
+    in transformOf termSubterms (appEndo $ getDual f) t
+
+defaultUniRewriteRules :: RewriteRules DefaultUni DefaultFun
+defaultUniRewriteRules = RewriteRules $ combineRules
+    -- rules are applied from left to right because of Dual
+    [ decodeEncode
```
michaelpj, line 24: As discussed, I'm not 100% sure this is safe. Maybe we should do `head (mkCons x xs) = let !_ = xs in x` or something?

michaelpj (commented):
General approach looks fine

plutus-core/plutus-ir/src/PlutusIR/Transform/Rewrite.hs
```diff
@@ -0,0 +1,36 @@
...
+    ]
+  where
+    combineRules = foldMap (Dual . Endo)
+
+decodeEncode :: Semigroup a => Term tyname name uni DefaultFun a -> Term tyname name uni DefaultFun a
+decodeEncode = \case
+    BA DecodeUtf8 a1 a2 (BA EncodeUtf8 a3 a4 t) ->
+        -- place the missed annotations inside the rewritten term
+        (<> a1 <> a2 <> a3 <> a4) <$> t
```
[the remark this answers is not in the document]
bezirg: Sounds complicated :(

plutus-core/plutus-ir/src/PlutusIR/Transform/Rewrite.hs
bezirg: Y I am also not certain.

plutus-core/plutus-ir/src/PlutusIR/Transform/Rewrite.hs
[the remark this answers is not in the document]
michaelpj: Just a function argument?

plutus-core/plutus-ir/src/PlutusIR/Analysis/Builtins.hs
bezirg: Can you try it out? RankNTypes get in the way for me both for makeLenses and for derive-newtype Semigroup,Monoid

michaelpj: The lenses seem fine, deriving doesn't work because of the `forall`s. I would either just write the instances by hand or have a list in there and evaluate it where you use it.

plutus-core/plutus-ir/src/PlutusIR/Transform/Rewrite.hs
bezirg: Turns out that this rewrite is safe in Haskell because  Text.fromString will handle gracefully non-unicode literal characters, see: https://hackage.haskell.org/package/text-2.1/docs/Data-Text.html#g:2

It is not safe currently in our case because:
1) unicode text is broken on plutus-tx
2) we don't do this graceful transformation in ghc-core that Haskell normally does. We have to replicate this behaviour in our plugin: https://hackage.haskell.org/package/text-2.1/docs/src/Data.Text.Internal.html#pack

plutus-core/plutus-ir/src/PlutusIR/Transform/RewriteRules.hs
```diff
@@ -0,0 +1,58 @@
...
+    let vinfo = termVarInfo t
+    in runQuote $ transformMOf termSubterms (rules vinfo) t
+
+-- | A bundle of composed `RewriteRules`, to be passed at entrypoint of the compiler.
+newtype RewriteRules uni fun = RewriteRules {
+    unRewriteRules :: forall tyname name m a
+                   . (MonadQuote m, Semigroup a)
+                   => VarsInfo tyname name uni a
+                   -> PIR.Term tyname name uni fun a
+                   -> m (PIR.Term tyname name uni fun a)
```
bezirg, line 45: @zliu41 @effectfully You probably wonder why MonadQuote m and VarsInfo are here. It is needed by all further rewrite rules (which are not yet in this PR).

plutus-core/plutus-ir/src/PlutusIR/Transform/RewriteRules/DecodeEncodeUtf8.hs
```diff
@@ -0,0 +1,32 @@
...
+  ) where
+
+
+import PlutusCore.Default
+import PlutusIR
+
+{- | This rewrites: `(decodeUtf8 (encodeUtf8 x)) => x`
+
+This rewrite is safe because x is either valid by construction *In Haskell, but not in GHC-CORE*
+(see <https://hackage.haskell.org/package/text-2.1/docs/Data-Text.html#g:2>)
```
effectfully, line 17: I'm not sure I understand the relevance of the link.

plutus-core/plutus-ir/src/PlutusIR/Transform/RewriteRules/DecodeEncodeUtf8.hs
```diff
@@ -0,0 +1,32 @@
...
+
+The opposite rewrite `(encodeUtf8 (decodeUtf8 x)) => x` is definitely not safe.
+-}
+-- FIXME: Unfortunately, unicode text is currently broken (at least on plutus-tx level), so
+-- we disable this rewrite until fix is in and further tested. See PLT-8314
+decodeEncodeUtf8 :: (Semigroup a, t ~ Term tyname name uni DefaultFun a) => t -> t
+decodeEncodeUtf8 = \case
+    BA DecodeUtf8 a1 a2 (BA EncodeUtf8 a3 a4 t) ->
+        -- place the missed annotations inside the rewritten term
+        (<> a1 <> a2 <> a3 <> a4) <$> t
```
effectfully, line 28: Just nitpicking, but I'd use `sconcat :: NonEmpty a -> a` instead, because we have

```
data Provenance a = 
                  <...>
                  | MultipleSources (S.Set (Provenance a))
```

and that would be a natural target for `sconcat`.

... except no implementation for `sconcat` is provided in

```haskell
instance Ord a => Semigroup (Provenance a) where
    x <> y = MultipleSources (toSet x `S.union` toSet y)
      where
        toSet = \case
            MultipleSources ps -> ps
            other              -> S.singleton other
```

but I think it's just an oversight and so I'd provide an implementation for `sconcat` and define `<>` in terms of that instead.

Do feel free to ignore this comment.

plutus-core/plutus-ir/src/PlutusIR/Transform/RewriteRules/DecodeEncodeUtf8.hs
```diff
@@ -0,0 +1,32 @@
...
+    t -> t
+
+pattern BA :: fun -> a -> a -> Term tyname name uni fun a -> Term tyname name uni fun a
+pattern BA b a1 a2 t <- Apply a1 (Builtin a2 b) t
```
effectfully, line 32: Too short of a global name IMO.

I'd just write `Apply a1 (Builtin a2 b) t` directly, it's not that longer and it's much more readable, but do feel free to ignore that.

plutus-core/plutus-ir/test/PlutusIR/Transform/RewriteRules/decodeEncodeUtf8.golden
effectfully, line 1: The test is disabled but the golden file exists?

plutus-core/plutus-ir/src/PlutusIR/Transform/RewriteRules/DecodeEncodeUtf8.hs
```diff
@@ -0,0 +1,32 @@
...
+
+module PlutusIR.Transform.RewriteRules.DecodeEncodeUtf8
+  ( decodeEncodeUtf8
+  ) where
+
+
+import PlutusCore.Default
+import PlutusIR
+
+{- | This rewrites: `(decodeUtf8 (encodeUtf8 x)) => x`
```
effectfully, line 14: I'd add a property test that `decodeUtf8 . encodeUtf8` is `id`. Who knows what kind of horrible infernal UTF8 weirdness might be going on within those functions.

plutus-core/plutus-ir/src/PlutusIR/Transform/RewriteRules.hs
```diff
@@ -0,0 +1,59 @@
...
+              , MonadQuote m
+              )
+            => RewriteRules uni fun
+            -> t
+            -> m t
+userRewrite (RewriteRules rules) t =
+    -- We collect `VarsInfo` on the whole program term and pass it on as arg to each RewriteRule.
+    -- This has the limitation that any variables newly-introduced by the rules would
+    -- not be accounted in `VarsInfo`. This is currently fine, because we only rely on VarsInfo
+    -- for isPure; isPure is safe w.r.t "open" terms.
```
effectfully, line 36: Kinda nasty tbh, but I'm not sure what to do about it.

plutus-core/plutus-ir/test/PlutusIR/Transform/RewriteRules/Tests.hs
```diff
@@ -0,0 +1,23 @@
...
+import Test.Tasty.Extras
+
+test_RewriteRules :: TestTree
+test_RewriteRules = runTestNestedIn ["plutus-ir/test/PlutusIR/Transform"] $
+    testNested "RewriteRules" $
+    fmap
+        (goldenPir (runQuote . RewriteRules.userRewrite def)  pTerm)
+        [ "equalsInt" -- this tests that the function works on equalInteger
+        , "divideInt" -- this tests that the function excludes not commutative functions
+        , "multiplyInt" -- this tests that the function works on multiplyInteger
```
effectfully, line 19: I'd consider adding a property test that tests whether optimizing a term permutes all bulitin applications that are supposed to be permuted and leaves others intact, but this is quite a bit more work, so having several golden tests is probably enough indeed.

plutus-tx/src/PlutusTx/Lift.hs
```diff
@@ -63,6 +65,8 @@ safeLift
        , PLC.Typecheckable uni fun
        , PrettyUni uni, Pretty fun
        , Default (PLC.CostingPart uni fun)
+       , Default (PIR.BuiltinsInfo uni fun)
+       , Default (PIR.RewriteRules uni fun)
```
effectfully, line 69: Constraints in Plutus Tx are completely out of order.

effectfully (approved):
Looks good to me, comments are basically just nitpicking.

plutus-core/plutus-ir/src/PlutusIR/Transform/RewriteRules.hs
```diff
@@ -0,0 +1,59 @@
...
+import PlutusPrelude
+
+import Control.Lens
+
+
+-- | Rewrite a `Term` using the given `RewriteRules` (similar to functions of Term -> Term)
+-- Normally the rewrite rules are configured at entrypoint time of the compiler.
+userRewrite :: ( Semigroup a, t ~ Term tyname name uni fun a
+              , HasUnique name TermUnique
+              , HasUnique tyname TypeUnique
```
effectfully, line 26: You can say `HasUniques t` instead of `HasUnique name TermUnique, HasUnique tyname TypeUnique`.

plutus-tx-plugin/src/PlutusTx/Options.hs
```diff
@@ -223,6 +224,9 @@ pluginOptions =
         , let k = "simplifier-remove-dead-bindings"
               desc = "Run a simplification pass that removes dead bindings"
            in (k, PluginOption typeRep (setTrue k) posDoSimplifierRemoveDeadBindings desc [])
+        , let k = "simplifier-rewrite"
```
zliu41, line 227: Is there any reason to want to turn this flag off? I can't think of any.

plutus-core/plutus-ir/src/PlutusIR/Transform/RewriteRules.hs
```diff
@@ -0,0 +1,59 @@
...
+import PlutusIR.Analysis.VarInfo
+import PlutusIR.Transform.RewriteRules.CommuteFnWithConst
+import PlutusPrelude
+
+import Control.Lens
+
+
+-- | Rewrite a `Term` using the given `RewriteRules` (similar to functions of Term -> Term)
+-- Normally the rewrite rules are configured at entrypoint time of the compiler.
+userRewrite :: ( Semigroup a, t ~ Term tyname name uni fun a
```
zliu41, line 24: How does a user specify a rewrite rule? It doesn't seem possible.

zliu41 (commented):
I don't see the rule for `constrData` and `unConstrData`; are you going to add it separately?

plutus-tx/src/PlutusTx/Lift.hs
bezirg: By order you mean, i should stick to a certain e.g. alphabetical order accoss the file?

plutus-core/plutus-ir/test/PlutusIR/Transform/RewriteRules/decodeEncodeUtf8.golden
bezirg: I will get rid of this rewrite rule completely for now

plutus-core/plutus-ir/src/PlutusIR/Subst.hs
```diff
@@ -129,7 +129,6 @@ funRes f = \case
   TyFun a dom cod -> TyFun a dom <$> funRes f cod
   t               -> f t
 
--- TODO: these could be Traversals
 -- | Get all the term variables in a term.
 vTerm :: Fold (Term tyname name uni fun ann) name
 vTerm = termSubtermsDeep . termVars
```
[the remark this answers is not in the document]
bezirg: I think I tried it, but I was hit but `cosmosOf`.  CosmosOf requires Contravariant f, which Traversals cannot provide

plutus-tx-plugin/src/PlutusTx/Options.hs
bezirg: Maybe u r right. But isn't this also the case for some other options as well? 

I am ok to remove it as long as there is at least 1 option to turn the simplifier off. And i think there is {simplifier-max-iterations=0} for that purpose

plutus-core/plutus-ir/src/PlutusIR/Transform/RewriteRules.hs
bezirg: userRewrite is a misnomer. What I wanted to express is that this `userRewrite` function is a general transformation that accepts "configurable" rewrite rules, based on a specific fun/universe. 

Currently our tx->pir->uplc pipeline fixes fun/uni to be `DefaultFun`, `DefaultUni`. The "user" does not have the ability to specify custom rules (like RULES of GHC), but he/she could statically configure another list of rewrite rules.

Anyway a misnomer, I will change it to sth like `applyRewriteRules`

bezirg:
> I don't see the rule for `constrData` and `unConstrData`; are you going to add it separately?

Yes

plutus-tx/src/PlutusTx/Lift.hs
effectfully: Sorry, I mean in the sense of "If something someone says or does is out of order, it is unpleasant or not suitable and it is likely to upset or offend people". They're way too bloated, but I don't know if we can have sensible type synonyms there.

plutus-core/plutus-ir/src/PlutusIR/Subst.hs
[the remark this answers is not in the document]
effectfully: I mean, is it in principle possible to have a `Traversal` focused on all the variables of a term? I think it is. If `cosmosOf` is getting in the way, then it just means `cosmosOf` isn't suitable for the job and we should use something else. Maybe we need to define this one in terms of `deepOf` or something, I'm not intimately familiar with `lens`.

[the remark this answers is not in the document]
bezirg: I do think it is possible, it's just we don't crucially need atm. Also it can be misused and lead to more variable capturing problems? Anyway, I will bring back the note in another PR but with a MAYBE tag.

plutus-core/plutus-ir/src/PlutusIR/Transform/RewriteRules/DecodeEncodeUtf8.hs
bezirg: I will address it in a new PR solely for this decode/encode rewriterule

PR #5547, opened 2023-09-18
[Test] [Builtins] Ensure 'Typeable', 'Lift' etc instances are present

effectfully:
This adds static tests ensuring the existence of `Lift` and `Typeable` instances for built-in types.

Not ready for review.

PR #5585, opened 2023-10-12
Add some comments on FromBuiltin/ToBuiltin/Lift

michaelpj:
Hopefully this helps a bit.

### DIFF
--- a/plutus-tx/src/PlutusTx/Builtins/Class.hs
+++ b/plutus-tx/src/PlutusTx/Builtins/Class.hs
@@ -27,15 +27,6 @@ import PlutusTx.Bool (Bool (..))
 import PlutusTx.Integer (Integer)
 import Prelude qualified as Haskell (String)
 
-{- Note [Fundeps versus type families in To/FromBuiltin]
-We could use a type family here to get the builtin representation of a type. After all, it's
-entirely determined by the Haskell type.
-
-However, this is harder for the plugin to deal with. It's okay to have a type variable
-for the representation type that needs to be instantiated later, but it's *not* okay to
-have an irreducible type application on a type variable. So fundeps are much nicer here.
--}
-
 {-|
 A class witnessing the ability to convert from the builtin representation to the Haskell representation.
 -}
@@ -62,17 +53,6 @@ instance ToBuiltin Bool BuiltinBool where
     {-# INLINABLE toBuiltin #-}
     toBuiltin b = if b then true else false
 
-{- Note [Strict conversions to/from unit]
-Converting to/from unit *should* be straightforward: just ``const ()`.`
-*But* GHC is very good at optimizing this, and we sometimes use unit
-where side effects matter, e.g. as the result of `trace`. So GHC will
-tend to turn `fromBuiltin (trace s)` into `()`, which is wrong.
-
-So we want our conversions to/from unit to be strict in Haskell. This
-means we need to case pointlessly on the argument, which means we need
-case on unit (`chooseUnit`) as a builtin. But then it all works okay.
--}
-
 instance FromBuiltin BuiltinUnit () where
     -- See Note [Strict conversions to/from unit]
     {-# INLINABLE fromBuiltin #-}
@@ -88,27 +68,6 @@ instance FromBuiltin BuiltinByteString ByteString where
 instance ToBuiltin ByteString BuiltinByteString where
     {-# INLINABLE toBuiltin #-}
     toBuiltin = BuiltinByteString
-
-{- Note [noinline hack]
-For some functions we have two conflicting desires:
-- We want to have the unfolding available for the plugin.
-- We don't want the function to *actually* get inlined before the plugin runs, since we rely
-on being able to see the original function for some reason.
-
-'INLINABLE' achieves the first, but may cause the function to be inlined too soon.
-
-We can solve this at specific call sites by using the 'noinline' magic function from
-GHC. This stops GHC from inlining it. As a bonus, it also won't be inlined if
-that function is compiled later into the body of another function.
-
-We do therefore need to handle 'noinline' in the plugin, as it itself does not have
-an unfolding.
-
-Another annoying quirk: even if you have 'noinline'd a function call, if the body is
-a single variable, it will still inline! This is the case for the obvious definition
-of 'stringToBuiltinString' (since the newtype constructor vanishes), so we have to add
-some obfuscation to the body to prevent it inlining.
--}
 
 -- We can't put this in `Builtins.hs`, since that force `O0` deliberately, which prevents
 -- the unfoldings from going in. So we just stick it here. Fiddly.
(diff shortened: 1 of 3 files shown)


plutus-tx/src/PlutusTx/Builtins/Class.hs
```diff
@@ -220,4 +165,80 @@ instance ToBuiltin BLS12_381.Pairing.MlResult BuiltinBLS12_381_MlResult where
...
 
+This is what the ToBuiltin and FromBuiltin classes do. They let us write wrappers
+for builtins relatively consistently by just calling toBuiltin on their arguments
+and fromBuiltin on the result. They shouldn't really be used otherwise.
+
+To keep the consistency we define dummy instances that are just the identity function
+for types that don't have a separate "normall Haskell" version. For example:
+integer. Integer in Plutus Tx user code _is_ the opaque builtin type, we don't
+expose a different one. Essentially: if it's a datatype then there's a substantive
+conversion.
```
effectfully, line 188: This sounds like we have two cases:

1. `Bool` vs `BuiltinBool` -- Plutus Tx type _vs_ built-in type, both appear in the `ToBuiltin`/`FromBuiltin` instance
2. `Integer` vs `BuiltinInteger` -- Plutus Tx type _and_ built-in type, both appear in the `ToBuiltin`/`FromBuiltin` instance

but we actually have a third one:

3. `Data` vs `BuiltinData` -- Plutus Tx type _vs_ built-in type, only the latter appears in the `ToBuiltin`/`FromBuiltin` instance

So this Note doesn't clarify this particular quirk for me.

@zliu41 did provide some explanations in [slack](https://input-output-rnd.slack.com/archives/G01EE4NQ9U0/p1695125397674719?thread_ts=1695059755.064939&cid=G01EE4NQ9U0), so could you please incorporate those if you agree with them?

plutus-tx/src/PlutusTx/Builtins/Class.hs
```diff
@@ -220,4 +165,80 @@ instance ToBuiltin BLS12_381.Pairing.MlResult BuiltinBLS12_381_MlResult where
...
+We could use a type family here to get the builtin representation of a type. After all, it's
+entirely determined by the Haskell type.
+
+However, this is harder for the plugin to deal with. It's okay to have a type variable
+for the representation type that needs to be instantiated later, but it's *not* okay to
+have an irreducible type application on a type variable. So fundeps are much nicer here.
+-}
+
+{- Note [Strict conversions to/from unit]
+Converting to/from unit *should* be straightforward: just ``const ()`.`
```
effectfully, line 201:
```suggestion
Converting to/from unit *should* be straightforward: just `const ()`.
```

plutus-tx/src/PlutusTx/Lift/Class.hs
```diff
@@ -135,56 +135,88 @@ instance (TypeError ('Text "Int is not supported, use Integer instead"))
...
+to put inside the constant.
+
+Fortunately, we have To/FromBuiltin, which happen to do what we want.
+See Note [Builtin types and their Haskell versions].
+This is arguably slightly an abuse: the versions of the types that we want in
+Plutus Tx source code and the versions that we use as the implementations of
+the builtin types in the universe could be different. But in practice they
+aren't. So we can write fairly straightforward instances for most types.
+
+Similarly, for Typeable we may have to use a different type from the opaque one.
```
effectfully, line 221: Thank you, that makes sense to me. But it still doesn't address the inconsistencies, I described those in the Jira [ticket](https://input-output.atlassian.net/browse/PLT-6538?focusedCommentId=213564), but I'll duplicate them here as well.

#5547 adds a bunch of instances and I wonder if that's what we want. In particular:

```diff
+ instance uni `PLC.HasTypeLevel` () => Typeable uni BuiltinUnit where
+     typeRep _proxyUnit = typeRepBuiltin (Proxy @())
+
+ instance uni `PLC.HasTermLevel` () => Lift uni BuiltinUnit where
+     lift = liftBuiltin . fromBuiltin
+
+ instance uni `PLC.HasTypeLevel` Bool => Typeable uni BuiltinBool where
+     typeRep _proxyBool = typeRepBuiltin (Proxy @Bool)
+
+ instance uni `PLC.HasTermLevel` Bool => Lift uni BuiltinBool where
+     lift = liftBuiltin . fromBuiltin
+
+ instance uni `PLC.HasTypeLevel` [] => Typeable uni BuiltinList where
+     typeRep _proxyBuiltinList = typeRepBuiltin (Proxy @[])
+
+ instance (FromBuiltin arep a, uni `PLC.HasTermLevel` [a]) => Lift uni (BuiltinList arep) where
+     lift = liftBuiltin . fromBuiltin
+
+ instance uni `PLC.HasTypeLevel` (,) => Typeable uni BuiltinPair where
+     typeRep _proxyBuiltinPair = typeRepBuiltin (Proxy @(,))
+
+ instance (FromBuiltin arep a, FromBuiltin brep b, uni `PLC.HasTermLevel` (a, b)) =>
+         Lift uni (BuiltinPair arep brep) where
+     lift = liftBuiltin . fromBuiltin
```

Also, why do we have

```haskell
instance uni `PLC.HasTermLevel` BS.ByteString => Lift uni BS.ByteString where
    lift = liftBuiltin
```

? It appears to be the only instance for a non-`Builtin*` type that isn't equal to its built-in version (i.e. `BuiltinInteger`/`Integer`).

effectfully (commented):
This does help me understand the setup better, but it doesn't answer my questions regarding various inconsistencies, I replicated those in the review comments.

plutus-tx/src/PlutusTx/Lift/Class.hs
michaelpj: I think those instances are indeed correct. And the one for ByteString looks wrong :scream:

plutus-tx/src/PlutusTx/Builtins/Class.hs
michaelpj: Hmm, I agree that the `Data` case is odd. I am actually not sure why it's like that. I'm going to try changing it and see what breaks.

plutus-tx/src/PlutusTx/Builtins/Class.hs
```diff
@@ -220,4 +165,80 @@ instance ToBuiltin BLS12_381.Pairing.MlResult BuiltinBLS12_381_MlResult where
...
+case on unit (`chooseUnit`) as a builtin. But then it all works okay.
+-}
+
+{- Note [noinline hack]
+For some functions we have two conflicting desires:
+- We want to have the unfolding available for the plugin.
+- We don't want the function to *actually* get inlined before the plugin runs, since we rely
+on being able to see the original function for some reason.
+
+'INLINABLE' achieves the first, but may cause the function to be inlined too soon.
```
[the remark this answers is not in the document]
michaelpj: Yeah, although the current hack has worked okay for some time.

plutus-tx/src/PlutusTx/Lift/Class.hs
michaelpj: We needed this for a derived lift instance for Data, which we also shouldn't have. I removed both and it's fine.

plutus-tx/src/PlutusTx/Builtins/Class.hs
michaelpj: It's actually quite odd, I'm trying to pin down how it should work.

PR #5587, opened 2023-10-12
Version-sensitive plutus-benchmark tests (PLT 7544)

kwxm:
This is PLT-7544: "Run plutus-benchmark tests on both 9.2 and 9.6".  The golden tests in `plutus-benchmark` now all have two lots of results, one for GHC 9.2 and one for GHC 9.6.


plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -12,6 +12,10 @@
 {-# OPTIONS_GHC -fno-omit-interface-pragmas #-}
 {-# OPTIONS_GHC -fno-spec-constr #-}
 {-# OPTIONS_GHC -fno-specialise #-}
+{-# OPTIONS_GHC -fexpose-all-unfoldings #-}
```
kwxm, line 15: We should remember to get someone to look at this,

plutus-core/testlib/Test/Tasty/Extras.hs
```diff
@@ -21,32 +24,47 @@ import Data.ByteString.Lazy qualified as BSL
...
+-- members of the second list.
+makeVersionedFilePath :: [FilePath] -> [FilePath] -> FilePath
+makeVersionedFilePath left right = joinPath left </> ghcVersion </> joinPath right
+
 -- | A 'TestTree' of tests under some name prefix.
-type TestNested = Reader [String] TestTree
+type TestNested = Reader [FilePath] TestTree
 
 -- | Run a 'TestTree' of tests with a given name prefix.
-runTestNestedIn :: [String] -> TestNested -> TestTree
```
kwxm, line 33: I''m not sure if I like this name.  It sounds like it runs a test, but really it's running a computation in the Reader monad.

plutus-core/testlib/Test/Tasty/Extras.hs
```diff
@@ -21,32 +24,47 @@ import Data.ByteString.Lazy qualified as BSL
...
 
 -- | Like `testNested` but adds a subdirectory corresponding to the GHC version being used.
-testNestedGhc :: String -> [TestNested] -> TestNested
-testNestedGhc folderName = testNested (folderName </> ghcVer)
-  where
-    ghcVer = showVersion compilerVersion
+testNestedGhc :: FilePath -> [TestNested] -> TestNested
+testNestedGhc folderName = testNested (folderName </> ghcVersion)
+
+-- Create a TestTree which runs in the directory 'path/<ghc-version>/dir'
```
kwxm, line 61: I don't like this name very much, but I couldn't think of anything better.

plutus-core/testlib/Test/Tasty/Extras.hs
```diff
@@ -21,32 +24,47 @@ import Data.ByteString.Lazy qualified as BSL
...
+runTestGroupNestedGhcIn :: [FilePath] -> FilePath -> [TestNested] -> TestTree
```
kwxm, line 62: I'm conflicted about whether to represent paths like `["dir1", "dir2", "dir3"]` or `"dir1/dir2/dir3".  We use both of these in the codebase, but in theory we might be in trouble on a system that doesn't use "/" as the path separator.  That's presumably very unlikely though.

kwxm, line 62: I don't like this name very much, but I couldn't think of anything better.

plutus-benchmark/nofib/test/Spec.hs
```diff
@@ -24,9 +24,10 @@ import PlutusCore.Default
 import PlutusTx qualified as Tx
 import PlutusTx.Test qualified as Tx
 
-runTestNested :: TestNested -> TestTree
-runTestNested = runTestNestedIn ["nofib", "test"]
-
+-- Make a set of golden tests with results stored in subdirectories determined
+-- by the GHC version.
+testGroupGhc :: [TestNested] -> TestTree
```
kwxm, line 29: ... or maybe `versionedTestGroup` or something.

plutus-core/testlib/Test/Tasty/Extras.hs
zliu41: The first options seems strictly better. I'm not sure whether or not the second option works on Windows.

zliu41: I'm not sure why this is needed. There's doesn't seem to be a use case where `dir` isn't `[]`?

zliu41 (approved):
Looks good. When merging, please use a better commit message than "Kwxm/test/plt 7544".

plutus-core/testlib/Test/Tasty/Extras.hs
kwxm: Oh, that's wrong.  It says it puts it in `path/version/dir`, but in fact it puts it in `path/dir/version`. 

This gets used in `marlowe/test` and `lists/test`.  See [here](https://github.com/input-output-hk/plutus/blob/74860632463e08989a5504e75477163e6b07125c/plutus-benchmark/marlowe/test/Spec.hs#L48) for example, where the directory is `semantics`.  You end up with golden files in `plutus-benchmark/marlowe/test` in the following directories:

* role-payout/9.2
* role-payout/9.6
* semantics/9.2
* semantics/9.6

What I meant to do in that function was to move the version directories further up, so you'd have

* 9.2/role-payout
* 9.2/semantics
* 9.6/role-payout
* 9.6/semantics

I don't know if that's any better.  Maybe we should just leave the directory layout as it is and get rid of the final `dir` argument in that function.

kwxm: > The first options seems strictly better. 

Maybe we can do that in a later PR.  You can work around it though, like [this](https://github.com/input-output-hk/plutus/blob/4886d40c75be5ef39f8f91daaed61a9b7df7c52c/plutus-core/plutus-ir/test/PlutusIR/Core/Tests.hs#L44).

kwxm: I found some tests in `plutus-ir` that had paths like `["a/b/c"]` so I changed them to `["a", "b", "c"]` while I was at it.

kwxm: I've fixed this now: all of the `9.2`/`9.6`/etc directories should be at the leaves of the directory hierarchy now.

PR #5593, opened 2023-10-17
[PlutusTx] Optimize the 'Eq' instance of 'Value'

effectfully:
Attempt number 3. Now with proper tests finally. I checked that the tests find all bugs from the previous PR (and one more). I absolutely didn't overfit them for that purpose, I'd written something most general first and only then checked. I entertained the idea of manually sorting out all possible corner cases, then building tests based on that, but the most straightforward QuickCheck thing really does seem to have a more or less equal power than whatever I would've wasted at least a week on, so I decided not to bother.

The new algorithm for checking equality of `Value`s is both faster and simpler. It's practically a reimplementation of `valueEqualsValue4` from #5135 (it may or may not differ, I didn't attempt to formally compare the two), so kudos to @manupadillaph.


plutus-benchmark/script-contexts/test/checkScriptContextEqualityTerm-20.eval.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 530200251
-| mem: 2147246})
\ No newline at end of file
+({cpu: 310936611
+| mem: 1245566})
```
effectfully, line 2: A bit worse than last time (it was `243063611`, but with a buggy version, so who knows if it even was correct), but still good.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Utils.hs
effectfully, line 1: Relocated some definitions from a test stanza, feel free to ignore this file, everything new here has already been reviewed.

plutus-ledger-api/test-plugin/Spec/golden/9.2/Long.stat.golden
```diff
@@ -1,15 +1,15 @@
...
-({cpu: 6808761202 | mem: 24376412})
-({cpu: 7477360785 | mem: 27205524})
-({cpu: 3869670543 | mem: 14302732})
-({cpu: 2884402957 | mem: 10629710})
-({cpu: 3952840806 | mem: 14128932})
-({cpu: 3754703216 | mem: 13744010})
-({cpu: 927859238 | mem: 3396606})
\ No newline at end of file
+({cpu: 3012083773 | mem: 11175942})
+({cpu: 713195980 | mem: 3098404})
```
effectfully, line 2: Note that those tests were written with the assumption that duplicated `CurrencySymbol`s/`TokenName`s were allowed, hence they're pretty useless for an algo that quits early in certain cases when we have duplicated names. Fixing that is a future work unrelated to this PR.

plutus-ledger-api/test-plugin/Spec/golden/9.2/Short.stat.golden
```diff
@@ -1,21 +1,21 @@
...
-({cpu: 198957295 | mem: 766946})
-({cpu: 204775120 | mem: 790948})
-({cpu: 236346454 | mem: 914568})
-({cpu: 390817706 | mem: 1513172})
-({cpu: 927859238 | mem: 3396606})
\ No newline at end of file
+({cpu: 4439100 | mem: 19400})
+({cpu: 6164100 | mem: 26900})
+({cpu: 8715491 | mem: 36702})
+({cpu: 10430980 | mem: 42904})
```
effectfully, line 4: These tests are a bit better, but still suffer from the same problem. But even then, at least some of the numbers we see here _are_ representative of the actual cost changes.

plutus-tx/src/PlutusTx/List.hs
```diff
@@ -244,14 +246,24 @@ xs0 !! n0 = go n0 xs0
...
+-- | Cons each element of the first list to the second one in reverse order (i.e. the last element
+-- of the first list is the head of the result).
+--
+-- > appendR xs ys === reverse xs ++ ys
+--
+-- >>> appendR "abc" "de"
+-- "cbade"
+appendR :: forall a. [a] -> [a] -> [a]
+appendR = rev where
+    rev :: [a] -> [a] -> [a]
```
effectfully, line 259: Just replicating the old logic to keep the diff minimal. We could've recursed in `appendR` directly, but that would be more costly due to that type-level `a` argument becoming a bunch of `delay` and `force` on the UPLC side (which is really annoying).

plutus-benchmark/marlowe/test/semantics/e34b48f80d49360e88c612f4016f7d68cb5678dd8cd5ddb981375a028b3a40a5.eval.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 605107070
-| mem: 2197734})
\ No newline at end of file
+({cpu: 552055724
+| mem: 1974822})
```
effectfully, line 2: Some of these are pretty decent.

plutus-benchmark/marlowe/test/semantics/7cbc5644b745f4ea635aca42cce5e4a4b9d2e61afdb3ac18128e1688c07071ba.eval.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 972095206
-| mem: 3457744})
\ No newline at end of file
+({cpu: 859760620
+| mem: 2993308})
```
effectfully, line 2: Another example.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -368,6 +372,70 @@ split (Value mp) = (negate (Value neg), Value pos) where
...
+This function recurses on both the lists in parallel and checks whether the key-value pairs are
+equal pointwise. If there is a mismatch, then it tries to find the left key-value pair in the right
+list. If that succeeds then the pair is removed from both the lists and recursion proceeds pointwise
+as before until there's another mismatch. If at some point a key-value pair from the left list is
+not found in the right one, then the function returns 'False'. If the left list is exhausted, but
+the right one still has some non-zero elements, the function returns 'False' as well.
+
+We check equality of values of two key-value pairs right after ensuring that the keys match. This is
+disadvantageous if the values are big and there's a key that is present in one of the lists but not
+in the other, since in that case computing equality of values was expensive and pointless. However
```
effectfully, line 388: Different to the previous PR, 'cause I changed my mind.

effectfully (commented):
Modest improvements across the board for the Marlowe benchmarks.

kwxm:
You might have some difficulty merging this now: see #5587.

effectfully:
> You might have some difficulty merging this now: see https://github.com/input-output-hk/plutus/pull/5587.

Somewhat surprisingly, it wasn't troubling at all. `git` is pretty smart.

plutus-benchmark/marlowe/test/semantics/9.2/0000020002010200020101020201000100010001020101020201010000020102.eval.golden
kwxm, line 1: This would be a lot easier to review if we had all of the test outputs in one file ...

kwxm: Also, what  happened to `role-payout`?  No change?

plutus-ledger-api/testlib/PlutusLedgerApi/Test/V1/Value.hs
```diff
@@ -0,0 +1,95 @@
...
+-- | 'arbitrary' throws, 'shrink' neither throws nor shrinks.
+instance Arbitrary (NoArbitrary a) where
+    arbitrary = error "No such 'Arbitrary' instance"
+    shrink _ = []
+
+instance Arbitrary Value where
+    arbitrary = do
+        -- Generate values for all of the 'TokenName's in the final 'Value' and split them into a
+        -- list of lists.
+        faceValues <- fmap (map getNonEmpty) . multiSplit . map unFaceValue =<< arbitrary
```
kwxm, line 87: This is generating Values where no CurrencySymbol has an empty list of tokens, is that correct?  Do we actually want to exclude such things?  I can create them manually and they seem to behave as you'd expect (for example, applying `split` to such a thing gives me back two Values with a currency symbol with an empty list of tokens).  

I was wondering if this situation was prohibited by the ledger, but presumably it isn't since enforcing it would require quite a lot of checking.  Now I'm worrying about whether it effects equality.  Do we consider a value with a currency symbol which has no tokens to be equal to the same thing with that currency symbol missing?  The version of equality implemented in this PR doesn't consider them to be equal.  We should check whether equality of values is specified in the ledger specification or somewhere.

kwxm (approved):
This looks good.  It's a lot easier to understand than the previous version.  I'd like to resolve the question of currency symbols with no tokens though.

kwxm:
Having looked at the tests again, it might be worth testing that `==` really is an equivalence relation (reflexive, symmetric, and  transitive) and that `geq` is a partial order (not strictly in the scope of this PR, but it would be straightforward now that there's a good generator for Values).

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -165,6 +166,13 @@ newtype AssetClass = AssetClass { unAssetClass :: (CurrencySymbol, TokenName) }
 assetClass :: CurrencySymbol -> TokenName -> AssetClass
 assetClass s t = AssetClass (s, t)
 
+{- Note [Value vs value]
+We call two completely different things "values": the 'Value' type and a value within a key-value
+pair. To distinguish between the two we write the former with a capital "V" and enclosed in single
+quotes and we write the latter with a lower case "v" and without the quotes, i.e. 'Value' vs value.
```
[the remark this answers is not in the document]
effectfully: > When rendered ass Haddock, the single quotes will disappear and turn into a link, right? I don't think that's likely to lead to any confusion though.

Yeah, but the link is properly highlighted, so the difference is clear in that case too.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -372,6 +376,82 @@ split (Value mp) = (negate (Value neg), Value pos) where
...
+    goBoth ((kL, vL) : kvsL') kvsR0@(kvR0@(kR0, vR0) : kvsR0')
+        -- We could've avoided having this clause if we always searched for the right key-value pair
+        -- using @goRight@, however the sheer act of invoking that function, passing an empty list
+        -- to it as an accumulator and calling 'appendR' afterwards affects performance quite a bit,
+        -- considering that all of that happens for every single element of the left list. Hence we
+        -- handle the special case of lists being equal pointwise (or at least their prefixes being
+        -- equal pointwise) with a bit of additional logic to get some easy performance gains.
+        | kL == kR0  = if vL `eqV` vR0 then goBoth kvsL' kvsR0' else False
+        | is0 vL     = goBoth kvsL' kvsR0
+        | otherwise  = goRight [kvR0 | not $ is0 vR0] kvsR0'
```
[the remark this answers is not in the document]
effectfully: I don't think it's any different, I'm just very used to writing zero-or-one-elements-depending-on-a-condition lists like that. I can change it if you find it confusing, although there are many places where I use that idiom. Do feel free to request not to use the idiom.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -372,6 +376,82 @@ split (Value mp) = (negate (Value neg), Value pos) where
...
+            -- @kL@, i.e. the first key in the remaining part of the left list. The accumulator
+            -- contains (in reverse order) all elements of the right list processed so far whose
+            -- keys are not equal to @kL@ and values are non-zero.
+            goRight :: [(k, v)] -> [(k, v)] -> Bool
+            goRight _   []                     = False
+            goRight acc (kvR@(kR, vR) : kvsR')
+                | is0 vR    = goRight acc kvsR'
+                -- @appendR@ recreates @kvsR0'@ with @(kR, vR)@ removed, since that pair
+                -- equals @(kL, vL)@ from the left list, hence we throw both of them away.
+                | kL == kR  = if vL `eqV` vR then goBoth kvsL' (appendR acc kvsR') else False
```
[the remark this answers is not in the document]
effectfully: > that's unlikely to make much difference since we'll only reach this code if the two lists already fail to be in the same order. 

Yeah, but what if the difference is only in two elements being swapped or something. I think it'd be too surprising to reverse the order of elements all the time for anybody who tried to understand the performance of `(==)`, so I opted for the predictable version that is likely more efficient as well.

> (Incidentally, I think `appendR` is often called `revAppend` in the strict world).

I'll rename, thanks.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -372,6 +376,82 @@ split (Value mp) = (negate (Value neg), Value pos) where
...
+    goBoth kvsL               []                               = all (is0 . snd) kvsL
+    -- Both spines are non-empty.
+    goBoth ((kL, vL) : kvsL') kvsR0@(kvR0@(kR0, vR0) : kvsR0')
+        -- We could've avoided having this clause if we always searched for the right key-value pair
+        -- using @goRight@, however the sheer act of invoking that function, passing an empty list
+        -- to it as an accumulator and calling 'appendR' afterwards affects performance quite a bit,
+        -- considering that all of that happens for every single element of the left list. Hence we
+        -- handle the special case of lists being equal pointwise (or at least their prefixes being
+        -- equal pointwise) with a bit of additional logic to get some easy performance gains.
+        | kL == kR0  = if vL `eqV` vR0 then goBoth kvsL' kvsR0' else False
```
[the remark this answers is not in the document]
effectfully: Yeah, it is.

plutus-ledger-api/testlib/PlutusLedgerApi/Test/V1/Value.hs
effectfully: > This is generating Values where no CurrencySymbol has an empty list of tokens, is that correct? 

Yes, and thank you for spotting that, it's really a generation bug. I've now fixed it, please review the [commit](https://github.com/input-output-hk/plutus/pull/5593/commits/4e89171e5176d5fcf49f80d8cd94992f5c5d58ee). I'll post some example distributions tomorrow or so.

> I originally thought that the version of == implemented [here](https://github.com/input-output-hk/plutus/blob/a0c01fbf08a441a7c16291c3ed69b550251eb40b/plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs#L449) didn't consider currency symbols with no tokens to be the same as absent currency symbols, but it does (and it says so here, which is a relief).

Yep, I didn't need to change anything in the implementation once I added generation of empty lists, tests kept being green (hope CI doesn't disprove that).

plutus-ledger-api/test/Spec/V1/Value.hs
```diff
@@ -0,0 +1,109 @@
...
+                        [ value1 <> value2 </> value1
+                        , value1 <> value2 </> value2
+                        , value1 <> value2 <=> value2 <> value1
+                        , forAll arbitrary $ \value3 ->
+                            not (isZero value3) ==>
+                                (value1 <> value2) <> value3 <=> value1 <> (value2 <> value3)
+                        ]
+            ]
+
+-- | Test that changing the values of some of the 'TokenName's creates a different 'Value'.
```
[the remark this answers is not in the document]
effectfully: That was a bit more annoying, but it's done now, thank you for the suggestion.

plutus-benchmark/marlowe/test/semantics/9.2/0000020002010200020101020201000100010001020101020201010000020102.eval.golden
effectfully: > No change?

I suppose. I have no idea really what to expect w.r.t. changes to Marlowe benchmarks.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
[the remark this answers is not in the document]
kwxm: No that's fine, just leave it.  I'd never seen this before, but the problem that it's solving is something that I've come across many times.

plutus-benchmark/marlowe/test/semantics/9.2/0000020002010200020101020201000100010001020101020201010000020102.eval.golden
kwxm: I checked the code and the [`role-payout` script](https://github.com/input-output-hk/plutus/blob/master/plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Scripts/RolePayout.hs) (it's quite simple) doesn't use `==` on Values.   The `semantics` one does though ([here](https://github.com/input-output-hk/plutus/blob/c688495eb77d9dc1e68abde2cb46540da11e3e9d/plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Scripts/Semantics.hs#L317)).

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Utils.hs
```diff
@@ -218,3 +220,90 @@ multiSplit xs = do
...
+
+-- | Split a list into chunks at random. Concatenating the resulting lists gives back the original
+-- one. Generates empty chunks. The first argument is the probability of generating at least one
+-- empty chunk as the first element of the resulting list. It is also the probability of generating
+-- an empty chunk as the last element of the resulting list. The probability of generating empty
+-- chunks decreases as we go from either of the ends of the resulting list (this is so that we are
+-- more likely to hit a corner case related to handling elements at the beginning or the end of a
+-- list).
+multiSplit :: Double -> [a] -> Gen [[a]]
+multiSplit endProb = multiSplit1 >=> insertManyPreferEnds endProb [] . coerce
```
kwxm, line 309: The tests are more complicated than the code they're testing now!  I'm reasonably convinced (mostly by looking at the output of the generators) that they're doing the right things though.

plutus-ledger-api/testlib/PlutusLedgerApi/Test/V1/Value.hs
kwxm: Looks good.  Thanks!

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Utils.hs
effectfully: Oh, I forgot to report here. I did look at the output of the generators closely for much larger probability of zeroes and more test runs and it was quite great (newlines added for readability):

```
          +++ OK, passed 100000 tests:
          12.967% [1]

           9.641% [0,1]
           9.614% [1,0]

           6.346% [1,0,0]
           6.341% [0,0,1]
           6.305% [0,1,0]

           4.087% [0,1,0,0]
           3.898% [1,0,0,0]
           3.875% [0,0,1,0]
           3.850% [0,0,0,1]

           2.367% [0,1,0,0,0]
           2.344% [1,0,0,0,0]
           2.336% [0,0,0,0,1]
           2.308% [0,0,0,1,0]
           2.293% [0,0,1,0,0]

           1.384% [0,0,0,1,0,0]
           1.377% [0,1,0,0,0,0]
           1.360% [0,0,1,0,0,0]
           1.314% [0,0,0,0,1,0]
           1.295% [1,0,0,0,0,0]
           1.271% [0,0,0,0,0,1]

           0.809% [0,0,0,1,0,0,0]
           0.759% [0,0,1,0,0,0,0]
           0.758% [0,0,0,0,1,0,0]
           0.753% [0,1,0,0,0,0,0]
           0.745% [0,0,0,0,0,0,1]
           0.720% [1,0,0,0,0,0,0]
           0.718% [0,0,0,0,0,1,0]
           <...>
```

and

```
           6.538% [1,1]
           6.390% [2]

           4.830% [0,2]
           4.828% [2,0]

           3.326% [0,1,1]
           3.269% [0,2,0]
           3.221% [1,1,0]
           3.146% [1,0,1]
           3.140% [0,0,2]
           3.133% [2,0,0]

           1.969% [2,0,0,0]
           1.949% [0,2,0,0]
           1.918% [0,0,2,0]
           1.901% [0,0,0,2]

           1.658% [1,1,0,0]
           1.634% [0,1,1,0]
           1.582% [0,0,1,1]
           1.559% [0,1,0,1]
           1.508% [1,0,1,0]
           1.475% [1,0,0,1]

           1.188% [0,0,2,0,0]
           1.158% [2,0,0,0,0]
           1.147% [0,0,0,0,2]
           1.138% [0,2,0,0,0]
           1.119% [0,0,0,2,0]

           0.874% [0,0,0,1,1]
           0.859% [1,1,0,0,0]
           0.823% [0,0,1,0,1]
           0.817% [0,0,1,1,0]
           0.807% [0,1,1,0,0]
           0.771% [1,0,1,0,0]
           0.741% [0,1,0,0,1]
           0.733% [0,1,0,1,0]
           0.732% [1,0,0,1,0]
```

You can see that there's some skewness in for example

```
           1.658% [1,1,0,0]
           1.634% [0,1,1,0]
           1.582% [0,0,1,1]
           1.559% [0,1,0,1]
           1.508% [1,0,1,0]
           1.475% [1,0,0,1]
```

but that's by design, since I made it so that inserting a zero (or an empty list etc) in the middle of a list happens a bit less often than doing it at one of the ends, which is precisely what we're observing.

PR #5603, opened 2023-10-31
Remarks about replacement of Unicode surrogates

kwxm:
The PLC parser uses `Data.Text.pack` when parsing Unicode strings, which replaces surrogate character codes with U+FFFE.  This adds a comment and a remark in the specification to explain that.


bezirg (approved):

kwxm:
This has been stuck in  a queue for about a week.  It's just modifiying some documents, so I'll merge it without waiting for CI.

Issue #4178, opened 2021-10-31
GCOmpare instances for DefaultUni and ValueOf


phadej:
## Area

[] *Plutus Foundation* Related to the GHC plugin, Haskell-to-Plutus compiler, on-chain code
[x] *Plutus Application Framework* Related to the Plutus application backend (PAB), emulator, Plutus libraries
[] *Marlowe* Related to Marlowe
[] *Other* Any other topic (Playgrounds, etc.)

## Describe the feature you'd like

GCompare, and Ord instances for DefaultUni and ValueOf

## Describe alternatives you've considered

Orphans:

```haskell
deriveGCompare ''DefaultUni

instance (GCompare uni, Closed uni, uni `Everywhere` Ord, uni `Everywhere` Eq) => GCompare (ValueOf uni) where
    ValueOf uni1 x1 `gcompare` ValueOf uni2 x2 =
        case uni1 `gcompare` uni2 of
            GLT -> GLT
            GGT -> GGT
            GEQ -> do
                bring (Proxy @Ord) uni1 $ case compare x1 x2 of
                    EQ -> GEQ
                    LT -> GLT
                    GT -> GGT
```

michaelpj:
Thoughts @effectfully

effectfully:
:+1: 

The instances make perfect sense, we just didn't need them.

@phadej I'm curious: what are doing that requires them? Using `Term`s as keys in a `Map`?

phadej:
> Using Terms as keys in a Map?

Something like that yes.

phadej:
Thanks!

PR #5609, opened 2023-11-02
[Build] Remove 'dependent-sum-template'

effectfully:
This removes the `dependent-sum-template` package from dependencies, so that it doesn't complicate supporting GHC 9.8. We don't need the package for anything, however it was requested that we add a `GCompare` instance for `DefaultUni` in #4178. We could implement the instance manually, but given that we don't need it my inclination is to say that if somebody needs the instance, they can either implement it themselves or indeed just depend on `dependent-sum-template` as we used to (unless they need it for GHC 9.8 and support for that version is never added to the package).


zliu41 (approved):
:+1:

PR #5611, opened 2023-11-03
[Builtins] Annotate each builtin with its type signature

effectfully:
This adds a type signature to the denotation of each builtin. See the added docs for why it's important.

I've also made the code more consistent.

Please review this one very carefully. Updating 70+ definitions and then a bunch of docs is an error-prone process even when emacs does half the job for you. It's not urgent, so please do take the time to review the changes.


michaelpj (approved):
I checked each one, they all look right to me.

effectfully:
Something weird is happening with this PR... I don't get any notifications and my review requests apparently were not sent either. Anyways, asked for @bezirg's review just to be sure I didn't screw anything up.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1072,51 +1083,96 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
+            {-# INLINE multiplyIntegerDenotation #-}
+        in makeBuiltinMeaning
+            multiplyIntegerDenotation
             (runCostingFunTwoArguments . paramMultiplyInteger)
+
     toBuiltinMeaning _semvar DivideInteger =
-        makeBuiltinMeaning
-            (nonZeroArg div)
+        let divideIntegerDenotation :: Integer -> Integer -> EvaluationResult Integer
+            divideIntegerDenotation = nonZeroArg div
```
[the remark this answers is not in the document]
effectfully: Good point, thanks. Fixed.

effectfully:
^ -2% on average. I think it's a lie, gonna rerun, there shouldn't be any performance differences.

effectfully:
^ -1.5%, let's try again.

effectfully:
-1.4%. Dunno, I doubt this shallow refactoring can make things faster.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1207,301 +1313,495 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
-          tailPlc (SomeConstant (Some (ValueOf uniListA xs))) = do
-              DefaultUniList _ <- pure uniListA
-              _ : xs' <- pure xs
-              pure $ fromValueOf uniListA xs'
-          {-# INLINE tailPlc #-}
+
     toBuiltinMeaning _semvar NullList =
-        makeBuiltinMeaning
-            nullPlc
+        let nullListDenotation :: SomeConstant uni [a] -> EvaluationResult Bool
```
kwxm, line 1478: Maybe we should have called this `IsNullList`.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1207,301 +1313,495 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
+            {-# INLINE traceDenotation #-}
+        in makeBuiltinMeaning
+            traceDenotation
             (runCostingFunTwoArguments . paramTrace)
+
     -- Pairs
     toBuiltinMeaning _semvar FstPair =
-        makeBuiltinMeaning
-            fstPlc
+        let fstPairDenotation :: SomeConstant uni (a, b) -> EvaluationResult (Opaque val a)
```
kwxm, line 1403: I should read the documentation again, but why does this need `EvaluationResult`?  I initially thought it was because it'll fail if you give it a non-pair, but anything can fail if you give it something of the wrong type.  Does this have to do with #-polymorphism, like it could in theory get a pair with an entry that isn't a built-in constant?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1207,301 +1313,495 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
+        in makeBuiltinMeaning
+            bls12_381_G1_compressDenotation
             (runCostingFunOneArgument . paramBls12_381_G1_compress)
+
     toBuiltinMeaning _semvar Bls12_381_G1_uncompress =
-        makeBuiltinMeaning
-            (eitherToEmitter . BLS12_381.G1.uncompress)
+        let bls12_381_G1_uncompressDenotation
+                :: BS.ByteString -> Emitter (EvaluationResult BLS12_381.G1.Element)
+            bls12_381_G1_uncompressDenotation = (eitherToEmitter . BLS12_381.G1.uncompress)
```
kwxm, line 1677: I think we can do away with these parentheses now.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1207,301 +1313,495 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
+        in makeBuiltinMeaning
+            bls12_381_G1_uncompressDenotation
             (runCostingFunOneArgument . paramBls12_381_G1_uncompress)
+
     toBuiltinMeaning _semvar Bls12_381_G1_hashToGroup =
-        makeBuiltinMeaning
-            (eitherToEmitter .* BLS12_381.G1.hashToGroup)
+        let bls12_381_G1_hashToGroupDenotation
+                :: BS.ByteString -> BS.ByteString -> Emitter (EvaluationResult BLS12_381.G1.Element)
+            bls12_381_G1_hashToGroupDenotation = (eitherToEmitter .* BLS12_381.G1.hashToGroup)
```
kwxm, line 1686: Parentheses again.

kwxm (approved):
This looks fine, and I think it's good to be explicit about the types.  Is there actually any scope to write a bad type and have it accepted, for example by missing out `EvaluationResult` or something?  I still haven't internalised all of the details of this  stuff.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -330,14 +328,17 @@ of them that are important to know to be able to use 'makeBuiltinMeaning' in cas
 complicated than a simple monomorphic or polymorphic function. But for now let's talk about a few
 more simple cases.
 
-4. Certain types are not built-in, but can be represented via built-in ones. For example, we don't
-have 'Int' as a built-in, but we have 'Integer' and we can represent the former in terms of the
+3. Certain types are not built-in, but can be represented via built-in ones. For example, we don't
+have 'Int' built-in, but we have 'Integer' and we can represent the former in terms of the
```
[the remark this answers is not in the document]
effectfully: > conceivable that we might add something like `Word8` or `Word64`

What for? We can already express those via `Integer` (and we already do for `Word8`) and it's unlikely that adding a new built-in type would help performance, since the cost of bounds checking that we currently do has to be virtually negligible compared to the cost of computing a builtin application. And a new built-in type is more code to write and maintain, plus potentially reduced performance due to GHC compiling big sum types less efficiently compared to small ones. Or do you have a use case in mind?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1128,69 +1185,118 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
             -- For builtin semantics variants other (i.e. larger) than
             -- DefaultFunSemanticsVariant1, the first input must be in range
             -- [0..255].  See Note [How to add a built-in function: simple
             -- cases]
-            DefaultFunSemanticsVariant2 -> makeBuiltinMeaning
-              @(Word8 -> BS.ByteString -> BS.ByteString)
-              BS.cons
-              costingFun
+            DefaultFunSemanticsVariant2 ->
+                let consByteStringDenotation :: Word8 -> BS.ByteString -> BS.ByteString
```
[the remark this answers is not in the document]
effectfully: > At PLC level this one takes an `integer` as the first argument and fails if it's not in the range 0..255 (the other variant reduces it modulo 256 and never fails). 

Correct.

> Is this the right type for that? We don't need `EvaluationResult` here because the failure occurs during the conversion from `Integer` to `Word8`

Yes, exactly.

> before we even get to the denotation, is that right?

Yes, before we get to the denotation (i.e. `consByteStringDenotation`), but within the runtime denotation (i.e. what `consByteStringDenotation` becomes after `makeBuiltinMeaning` adds unlifting/lifting and costing to it).

Note that even if there was no `Word8`, there would still be possibility of evaluation failure, since the user may supply a `Bool` or even a lambda for the `ByteString` argument.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
effectfully: >  I initially thought it was because it'll fail if you give it a non-pair, but anything can fail if you give it something of the wrong type. Does this have to do with #-polymorphism, like it could in theory get a pair with an entry that isn't a built-in constant?

No, a pair always has constants inside at runtime, there's no way one could sneak anything else in there, since it's unrepresentable in the AST.

This is described in the docs in this file but tl;dr is that we cannot have 

```haskell
fstPairDenotation :: (a, b) -> Opaque val a
```

due to some technical and annoying reasons, i.e. we cannot automatically unlift values of unmonomorphized polymorphic built-in types in the builtins machinery, unlike values of monomorphic types (e.g. `Integer`) or values of monomorphized polymorphic built-in types (e.g. `[(Bool, ByteString)]`). Therefore we have to unlift a pair manually within the denotation of the builtin and that operation can fail, hence `EvaluationResult`.

_In theory_ we could make automatic unlifting work for `fstPair`. But that is non-trivial and making automatic unlifting work for something like `mkCons` sounds like a complete nightmare, since we need to check equality of type tags extracted from the two arguments there and doing that automatically is just way too complicated and I'm sure nobody would appreciate the complexity of such a machinery, hence I didn't attempt to implement that. Present-day situation is "if the builtin doesn't have unmonomorphized polymorphic built-in types, the machinery does everything for you, but otherwise you have to understand how to do things manually" and I think it's a fine balance between complexity of the machinery and it's user-friendliness.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
effectfully: > Maybe we should have called this `IsNullList`.

Too late for that, I guess.

effectfully: > Too late for that, I guess.

(duh, that's what "should have" means Roman, learn some English already)

effectfully:
Comments addressed, I'm enabling auto-merge.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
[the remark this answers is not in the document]
bezirg: We could use fromInteger instead of fromIntegral in the semvar==1 case, but I guess ghc inlines fromIntegral , so it should be the same in the ghc core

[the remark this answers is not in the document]
effectfully: Just checked out of curiosity, it inlines to `integerToWord#` and later `wordToWord8#` is called over the result, so we probably can't do better than that.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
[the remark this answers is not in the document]
kwxm: > What for?

Who knows?  We have had requests for a full range of fixed-width types and operations in the past (although hopefully the bitwise thing will address that), so it's not out of the question that someone might request it in the future.  I was mostly just curious about what would actually happen though.

PR #5616, opened 2023-11-08
Update to new nothunks

michaelpj:
`ouroboros-network-framework` now has a lower bound on this, so we need to be compatible with it. It adds a conflicting instance, so we need to remove that and also update our lower bound too.


kwxm (approved):
This looks OK, as far as I can tell.  If it passes CI it should be good to merge.

Issue #5589, opened 2023-10-14
Compilation errors report do not point to the location 


abailly-iohk:
### Summary

When one gets a PlutusTx compilation error like 

```
 GHC Core to PLC plugin: Error: Reference to a name which is not a local, a builtin, or an external INLINABLE function: Variable GHC.Num.Integer.integerGt#
```
there is no indication of the origin of the error in the original source code, and one gets a 3000+ lines long error report which is pretty much useless.

### Steps to reproduce the behavior

An example contract that's failing to build can be found [here](https://github.com/abailly/black-jack/blob/3386f8ecd5d4cbae1e107b41053f24b76effc4f1/black-jack-core/src/Chess/Game.hs) but it's easy to reproduce with simpler code.

### Actual Result

An example output from Plutus compilation.

[out.md](https://github.com/input-output-hk/plutus/files/12903432/out.md)

### Expected Result

As a PlutusTx developer, I would like the compiler to tell me where the problem is in my source code.
In this particular case, I expect something like:

```
Plutus compiler error in line 167, column 22:  GHC.Num.Integer.integerGt# is not a local, a builtin, or an external INLINABLE function
Possible fix: You are trying to use a literal number inside a multiway-if expression which is not supported by the Plutus compiler, please check https://plutus.readthedocs.io/en/latest/troubleshooting.html#non-inlinable-functions
```

### Describe the approach you would take to fix this

I am not a compiler engineer and I am not aware of the details of the internals of the PlutusTx plugin, so it's very hard for me suggest an approach.

### System info

OS: Mac OS X Ventura 13.6 arm64
GHC: 9.2.8
Plutus: 1.7.0.0

effectfully:
I've tried adding `GHC.ppr (GHC.getSrcSpan n)` to the error message, but all it gave me was `<no location info>`, since if GHC attempted to print the location of the binding site of a variable that it doesn't recognize. I've tried to persuade GHC to do _something_ with the `Tickish` thing, but to no avail.

I guess we'll have to wait for @michaelpj to come from his vacation for this one, since he's probably the only person who's intimately familiar with GHC internals in our team. @zliu41 or do you happen to know how to reliably get source locations out of GHC?

michaelpj:
We should print the src spans in the context wherever we have them. But we often don't have them. Adding `-g` can help (it seems to add more src spans), but in general I think we can't go from an arbitrary GHC Core expression to its src span, which sucks :(

effectfully:
[...]

Issue #3395, opened 2021-06-17
Use of ranges in Plutus


SaIG:
## Area

[X] *Plutus Foundation* Related to the GHC plugin, Haskell-to-Plutus compiler, on-chain code
[] *Plutus Application Framework* Related to the Plutus application backend (PAB), emulator, Plutus libraries
[] *Marlowe* Related to Marlowe
[] *Other* Any other topic (Playgrounds, etc.)

## Describe the feature you'd like

I'd like be able to use Ranges in lists within Plutus.
Like [0..10] instead i have to write [0,1,2,3,4,5,6,7,8,9,10] otherwise it wont compile

VladimirLogachev:
It is possible to use `enumFromTo`.

effectfully:
Indeed you can use `enumFromTo`. Unfortunately GHC elaborates ranges to `enumFromTo` from `base` and we use a different one, so we'd either have to insist on everybody enabling `RebindableSyntax` by default, which is very pervasive, or to replace `enumFromTo` from `base` with our own behind the user's back and hardcode that into the compiler -- and neither of those is a good option. So for now, the only workaround is to use `enumFromTo` manually, but we'll keep this issue in mind to hopefully address it somehow in future.

effectfully:
I'm going to claim that we've spent quite a bit of time on this already (see the linked PRs above) and are unlikely to spend even more, so I'm closing the issue.

Feel free to reopen if you believe it should stay open.

PR #5619, opened 2023-11-09
Better errors for literal ranges in PlutusTx (PLT-8174)

kwxm:
[See also issues #5589 and #3395].

Suppose you have this code in a Haskell program:

```
   let !_ = $$(compile [|| [1..99] :: [Integer] ||])
```
The plugin currently gives you a 186-line error message like this:
```
GHC Core to PLC plugin: Error: Reference to a name which is not a local, a builtin, or an external INLINABLE function: Variable GHC.Num.Integer.integerGt#
                               No unfolding
Context: Compiling expr: GHC.Num.Integer.integerGt#
Context: Compiling expr: GHC.Num.Integer.integerGt# x1
Context: Compiling expr: GHC.Num.Integer.integerGt# x1 y1
Context: Compiling expr: case GHC.Num.Integer.integerGt# x1 y1 of {
                           __DEFAULT -> c x1 (go3 (GHC.Num.Integer.integerAdd x1 1));
                           1# -> n
...
<lots of nested context information>
...
Context: Compiling definition of: GHC.Enum.$fEnumInteger_$cenumFromTo
Context: Compiling expr: GHC.Enum.$fEnumInteger_$cenumFromTo
Context: Compiling expr: GHC.Enum.$fEnumInteger_$cenumFromTo 1
Context: Compiling expr: GHC.Enum.$fEnumInteger_$cenumFromTo 1 99
Context: Compiling expr at "main:Main:(26,15)-(26,52)"
```
This happens because the range is desugared to a call to [enumFromTo](https://hackage.haskell.org/package/base-4.19.0.0/docs/Prelude.html#v:enumFromTo) in the `Enum` class, but things are complicated by the fact that in the case of `Integer` there is extensive [inlining and rewriting](https://hackage.haskell.org/package/base-4.19.0.0/docs/src/GHC.Enum.html#line-902).

This PR improves things slightly by spotting uses of `enumFromTo` and related functions and suggesting that the user should use `PlutusTx.enumFromTo` instead.  I've also added `PlutusTx.enumFromThenTo` for ranges like `[1, 5..101]` with non-unit steps.

With this change the above error message is replaced by 
```
GHC Core to PLC plugin: Error: Unsupported feature: Literal ranges are not supported: please use PlutusTx.Enum.enumFromTo or PlutusTx.Enum.enumFromThenTo
Context: Compiling expr: GHC.Enum.$fEnumInteger_$cenumFromTo
Context: Compiling expr: GHC.Enum.$fEnumInteger_$cenumFromTo 1
Context: Compiling expr: GHC.Enum.$fEnumInteger_$cenumFromTo 1 99
Context: Compiling expr at "main:Main:(26,15)-(26,52)"
```

This is an improvement, but is still not ideal:  if the range is nested inside multiple functions then you'll still get  lots of nested contexts with the final line (which is the only one containing a source location) pointing to the splice, which may be a long way from where the range occurs.  I think we should merge this for now and think about ways to improve the situation.

### Possible improvements
I thought about getting the plugin to replace occurrences of `enumFromTo` with `PlutusTx.enumFromTo`.  This is tricky for a number of reasons:
* Because of the inlining/rewriting for `Integer` some other function may be called and it's difficult to work out what's going on.  
* We'd need a GHC [Name](https://hackage.haskell.org/package/ghc-9.8.1/docs/GHC.html#t:Name) for `PlutusTx.enumFromTo`: those are quite complicated (they include unique IDs, for example) and I'm not  sure how to get hold of one.
* Maybe it's just not a good idea to silently replace a standard Haskell function with our own version.

I looked at trying to improve the context information as well, but it looked as if that'd take quite a lot of work.
* The plugin provides a `PlutusTx.Plugin:context-level` option which looks as if it should let you change the amount of context information, but I couldn't get it to work.  I think it may be getting overridden by calls to `traceCompilation` like [this](url).
* Is there any way to get accurate location information for stuff that's reachable transitively from the splice? You can kind of see what might be going on, but there can be huge amounts of GHC Core to work through.  See issues #5589 and #3395 for complaints about this.


plutus-tx/src/PlutusTx/Enum.hs
```diff
@@ -43,9 +47,24 @@ instance Enum Integer where
   fromEnum x = x
 
   {-# INLINABLE enumFromTo #-}
-  enumFromTo x y
-    | x > y = []
-    | otherwise = x : enumFromTo (succ x) y
+  enumFromTo x lim
```
kwxm, line 50: I changed `y` to `lim` so that it looks more like what's in `GHC.Enum`.  That changed a lot of golden files too.

plutus-tx-plugin/test/Budget/9.2/show.pir.golden
```diff
@@ -200,12 +200,12 @@ let
 in
 letrec
   !`$fEnumBool_$cenumFromTo` : integer -> integer -> List integer
-    = \(x : integer) (y : integer) ->
+    = \(x : integer) (lim : integer) ->
```
kwxm, line 203: See later for an explanation of why all of these have changed.

plutus-tx-plugin/plutus-tx-plugin.cabal
```diff
@@ -115,7 +115,7 @@ executable gen-plugin-opts-doc
 
   default-language: Haskell2010
 
-test-suite plutus-tx-tests
+test-suite plutus-tx-plugin-tests
```
kwxm, line 118: I changed this because `plutus-tx` contains `plutus-tx-test` and I was getting very confused.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -228,6 +228,40 @@ isProbablyIntegerEq (GHC.getName -> n)
...
 isProbablyIntegerEq _ = False
 
+-- | Check for literal ranges like [1..9] and [1, 5..101].  This will also
+-- return `True` if there's an explicit use of `enumFromTo` or similar.
+isProbablyBoundedRange :: GHC.Id -> Bool
+isProbablyBoundedRange (GHC.getName -> n)
+    | Just m <- GHC.nameModule_maybe n
+    , GHC.moduleNameString (GHC.moduleName m) == "GHC.Enum" =
+        ("$fEnum" `isPrefixOf` methodName &&
+         (  "_$cenumFromTo"     `isSuffixOf` methodName  -- [1..100]
```
kwxm, line 238: Is there any better way of getting names of particular instances of methods like `enumFromTo`?  This was the best I could do and I'm not sure how robust it will be.  I suppose that if the names that GHC produces change then the tests will start to fail.

kwxm:
The `checkScriptContextEquality` test in `plutus-benchmark` is failing because its CPU budget has increased slightly.  I'm finding it difficult to imagine why this might be happening.

plutus-ledger-api/test-plugin/Spec/Value.hs
```diff
@@ -2,7 +2,6 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TemplateHaskell   #-}
 
-{-# OPTIONS_GHC -fplugin PlutusTx.Plugin #-}
```
kwxm, line 5: Nothing to do with this PR: I'm just taking the opportunity to remove it while I'm here.

kwxm: Nope, I'm going to revert that.  It may be causing problems with at least one GHC version.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
zliu41: Oh, this is too bad. I hadn't realized that GHC might specialize `enumFromTo`.

The answer to your question is, unfortunately, no. There's no way to prevent GHC from specializing a method. It is only possible to disable user written rules, not builtin rules.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -730,6 +764,16 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
     GHC.Var n
       | isProbablyBytestringEq n ->
           throwPlain $ UnsupportedError "Use of Haskell ByteString equality, possibly via the Haskell Eq typeclass"
+    GHC.Var n
+      -- Try to produce a sensible error message if a range like [1..9] is encountered.  This works
+      -- by looking for occurrences of GHC.Enum.enumFromTo and similar functions; the same error
+      -- occurs if these functions are used explicitly.
+      | isProbablyBoundedRange n ->
+          throwPlain $ UnsupportedError "Literal ranges are not supported: please use PlutusTx.Enum.enumFromTo or PlutusTx.Enum.enumFromThenTo"
```
zliu41, line 772: We don't know for sure that the user used list range syntax. You can instead say something like: "Unsupported feature: GHC.Enum.Enum. You probably used list range syntax, which is not currently supported. You can use PlutusTx.Enum.Enum instead".

plutus-tx-plugin/test/Plugin/Errors/Spec.hs
```diff
@@ -95,3 +99,17 @@ instance Eq AType where
 
 literalCaseOther :: CompiledCode (AType -> AType)
 literalCaseOther = plc (Proxy @"literalCaseOther") (\x -> case x of { "abc" -> ""; x -> x})
+
+-- Tests for literal ranges (and the corresponding methods in GHC.Enum). These
+-- should all fail with informative error messages.
+rangeEnumFromTo :: CompiledCode [Integer]
```
zliu41, line 105: Nice tests.

zliu41 (approved):
Looks good.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
kwxm: > We don't know for sure that the user used list range syntax.

Yes, I did worry about that but decided to go for the shorter version to avoid a giant string in the source code. I'll change it.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
michaelpj: I'm not sure if we should match on the specialized name. I don't know if we can guarantee that that isn't an arbitrarily messed up version :thinking:

michaelpj: We pass some information about GHC names through to the compiler in `NameInfo` in the `CompileContext`. Then you can use `getThing` to get the `TypedThing` and `getName` on that to get the `Name`. See how the stuff in `Builitns` works.

plutus-tx/src/PlutusTx/Enum.hs
```diff
@@ -26,8 +26,12 @@ class Enum a where
   toEnum :: Integer -> a
   -- | Convert to an 'Integer'.
   fromEnum :: a -> Integer
-  -- | Construct a list from the given range.
+  -- | Construct a list from the given range (corresponds to [a..b]).
   enumFromTo :: a -> a -> [a]
+  -- | Construct a list from the given range (corresponds to [a,b..c]).  This
+  -- has the same semantics as the Haskell version,so if a==b and c>=b then you
+  -- get an infinite list, which you probably don't want in Plutus Core.
+  enumFromThenTo :: a -> a -> a -> [a]
```
michaelpj, line 34: I'm not sure if we should add this. It's not that useful and it makes the dictionaries bigger.

kwxm: > I'm not sure if we should add this. It's not that useful and it makes the dictionaries bigger.

Yeah, I wondered about that.  I thought I might as well add it while I was looking at this, but it didn't initially occur to me that it would make the dictionary bigger.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
kwxm: > We pass some information about GHC names through to the compiler in `NameInfo` in the `CompileContext`. Then you can use `getThing` to get the `TypedThing` and `getName` on that to get the `Name`. See how the stuff in `Builitns` works.

Maybe we can come back to this in a later PR. I've run out  of  JIRA points on this.

kwxm: > I'm not sure if we should match on the specialized name. I don't know if we can guarantee that that isn't an arbitrarily messed up version 🤔

That's what I was asking about in my original comment.  Can we actaully get at the original name, especially after the inlining/rewriting?

michaelpj: No, `NameInfo` is just for getting an actual GHC `Name` instead of string matching. I'm not sure it will match up if it has in fact been specialized. I'm not sure what we do in other cases, perhaps we just fail to spot stuff :thinking:

michaelpj: I guess it's not too bad since we're failing instead of silently compiling differently.

michaelpj: I guess since this is inherently heuristic rather than matching a precise name there's no point doing the `NameInfo` stuff and this is fine.

kwxm:
I think I'd like to merge this to get it out of the way.  There are a number of things I'd like to revisit at a later date though: see the original PR comments.

bezirg:
@kwxm @michaelpj  Have you looked into using RebindableSyntax instead?

From https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/rebindable_syntax.html :+1: 

```
So the [RebindableSyntax](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/rebindable_syntax.html#extension-RebindableSyntax) extension causes the following pieces of built-in syntax to refer to whatever is in scope, not the Prelude versions:

List notation, such as [x,y] or [m..n] can also be treated via rebindable syntax if you use -XOverloadedLists; see [Overloaded lists](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/overloaded_lists.html#overloaded-lists).
```

PR #5618, opened 2023-11-09
Expose 'unSatInt' and 'fromSatInt' and add instances for 'CostModelApplyError'

effectfully:
As requested in #5613.

PR #5621, opened 2023-11-09
Expose 'unSatInt' and 'fromSatInt' from V2 and V3 as well

effectfully:
#5618 wasn't ready for merging, this one does for V2 and V3 the same as the previous one did for V1 and adds a changelog entry.


effectfully:
I think this one is entirely uncontroversial, so I'll merge it just to close the issue without waiting for a review, but please still review it just in case.

PR #5625, opened 2023-11-11
Remove some redundant pragmas

kwxm:
A number of files in `plutus-tx-plugin/test` contain
```
{-# OPTIONS_GHC -fplugin PlutusTx.Plugin #-}
```
where it isn't needed.  This confused me a bit, so I've removed them.

I found some other unnecessary pragmas in these files, so I removed them too.  There are probably similar things in other tests in this package, but I didn't go through them all.

### DIFF
--- a/plutus-tx-plugin/test/IntegerLiterals/NoStrict/NoNegativeLiterals/Spec.hs
+++ b/plutus-tx-plugin/test/IntegerLiterals/NoStrict/NoNegativeLiterals/Spec.hs
@@ -11,8 +11,6 @@
 {-# LANGUAGE NoNegativeLiterals    #-}
 {-# LANGUAGE NoStrict              #-}
 
-{-# OPTIONS_GHC -fplugin PlutusTx.Plugin #-}
-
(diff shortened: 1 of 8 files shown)


plutus-tx-plugin/test/IntegerLiterals/NoStrict/NoNegativeLiterals/Spec.hs
```diff
@@ -1,17 +1,9 @@
-{-# LANGUAGE BangPatterns          #-}
-{-# LANGUAGE DataKinds             #-}
-{-# LANGUAGE FlexibleContexts      #-}
-{-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE NegativeLiterals      #-}
```
kwxm, line 5: This file had  both `NegativeLiterals` and `NoNegativeLiterals`, in that order.  Presumably the latter (which is what's wanted here) took precedence.

zliu41 (approved):

PR #5626, opened 2023-11-11
Use Readable UPLC in the golden files of simplifier tests

zliu41:
This makes the golden files more readable; the uniques can be important for these tests.


plutus-core/plutus-core/src/PlutusCore/Pretty/Readable.hs
```diff
@@ -100,6 +100,28 @@ deriving via
   instance
     (PrettyDefaultBy config (AsReadable a)) => PrettyBy config (AsReadable a)
 
+newtype AsReadableWithUniques a = AsReadableWithUniques
+  { unAsReadableWithUniques :: a
+  }
```
effectfully, line 105: `AsDefaultWithUniques` would be useful too... I wonder if we could keep `AsReadable` and `AsWithUniques` separate and define `AsDefaultWithUniques` in terms of that. Or have `As '[Readable, WithUniques]` or something.

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
```diff
@@ -227,11 +227,16 @@ multiApp = runQuote $ do
...
+  (PrettyBy (PrettyConfigReadable PrettyConfigName) a) =>
+  String ->
+  String ->
+  a ->
+  TestTree
 goldenVsPretty extn name value =
   goldenVsString name ("untyped-plutus-core/test/Transform/" ++ name ++ extn) $
-    pure . BSL.fromStrict . encodeUtf8 . render $
-      prettyPlcClassicDebug value
+    pure . BSL.fromStrict . encodeUtf8 . render . pretty $
```
effectfully, line 238: `render . pretty` is `display` IIRC.

plutus-core/untyped-plutus-core/test/Transform/inlinePure4.uplc.golden
```diff
@@ -1,9 +1 @@
...
-  (lam a_4 (lam b_5 a_4))
-  (force
-    [
-      (lam x_6 (delay (lam y_7 [ x_6 x_6 ])))
-      (delay [ (error) (con integer 1) ])
-    ]
-  )
-]
\ No newline at end of file
+(\a_4 b_5 -> a_4) (force ((\x_6 -> delay (\y_7 -> x_6 x_6)) (delay (error 1))))
```
effectfully, line 1: Hm, why is it a single line? I thought it would render as two. I'll look into the UPLC pretty-printer.

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
```diff
@@ -227,11 +227,16 @@ multiApp = runQuote $ do
...
+goldenVsPretty ::
+  (PrettyBy (PrettyConfigReadable PrettyConfigName) a) =>
+  String ->
+  String ->
+  a ->
+  TestTree
 goldenVsPretty extn name value =
   goldenVsString name ("untyped-plutus-core/test/Transform/" ++ name ++ extn) $
-    pure . BSL.fromStrict . encodeUtf8 . render $
-      prettyPlcClassicDebug value
```
effectfully, line 234: You could just use `prettyPlcReadableDebug` here without defining a new data type. The `AsReadable` thing is useful for satisfying a `Show`, `Pretty`, `PrettyBy` etc constraint in a way that renders the thing readably regardless of the constraint.

effectfully (approved):
LGTM

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
zliu41: Oh that's convenient.. I didn't know that. I'll remove `AsReadableWithUniques` then.

effectfully: > I'll remove `AsReadableWithUniques` then.

Nah, keep it. It's definitely useful and I can see myself replacing some of `AsReadable` usages with `AsReadableWithUniques`.

michaelpj:
I'm not sure I agree with this. We explicitly moved away from printing uniques because it leads to _huge_ amounts of churn in golden files. Any peturbation to the compiler pipeline likely means that _every_ unique changes in _every_ golden file, obscuring any real change. It was pretty bad.

zliu41:
The uniques were already there, it's just readable vs. unreadable.

michaelpj:
:facepalm: sorry, I somehow got the completely wrong idea about this

PR #5628, opened 2023-11-12
Add evaluation results for some budget tests

zliu41:
... to make sure optimizations we do in the future do not accidentally change the results.


plutus-tx-plugin/test/Budget/Spec.hs
```diff
@@ -39,165 +40,212 @@ tests = testNestedGhc "Budget" [
...
 
   , goldenBudget "findIndexEmptyList" compiledFindIndexEmptyList
   , goldenUPlcReadable "findIndexEmptyList" compiledFindIndexEmptyList
   , goldenPirReadable "findIndexEmptyList" compiledFindIndexEmptyList
+  , goldenEvalCekCatch "findIndexEmptyList-result" [compiledFindIndexEmptyList]
 
   , goldenBudget "filter" compiledFilter
   , goldenUPlcReadable "filter" compiledFilter
   , goldenPirReadable "filter" compiledFilter
+  , goldenEvalCekCatch "filter-result" [compiledFilter]
```
effectfully, line 108: That is getting repetitive, how about defining a function encompassing the four types of tests? We'll probably need it in other places too.

effectfully (approved):
Fantastic, we should've done that long ago, thank you for doing it.

I didn't check whether the golden files are as expected, tell me if I should.

plutus-tx-plugin/test/Budget/Spec.hs
michaelpj: Yeah, I have also wondered about e.g. just combining the result value and the budget into a single file...

michaelpj (approved):

PR #5629, opened 2023-11-12
Add `Hashable` instances for UPLC `Term` and related types

zliu41:
This will be needed by CSE.

@kwxm let me know what you think of the instances for the BLS types.


plutus-core/plutus-core/src/PlutusCore/Crypto/BLS12_381/Pairing.hs
```diff
@@ -49,6 +50,9 @@ instance Flat MlResult where
 instance NFData MlResult where
     rnf _ = ()
 
+instance Hashable MlResult where
+    hashWithSalt salt = const salt
```
zliu41, line 54: This is needed to make `Term` hashable, and there doesn't seem to be anything else we can do. @kwxm

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Eq.hs
```diff
@@ -33,14 +44,44 @@ deriving stock instance
...
    Eq (Term DeBruijn uni fun ann)
 
+instance
+  ( GEq uni
+  , Closed uni
+  , uni `Everywhere` Eq
+  , uni `Everywhere` Hashable
+  , Hashable ann
+  , Hashable fun
+  ) =>
```
effectfully, line 82: I'd define a type synonym.

effectfully (approved):
LGTM

zliu41:
Merging since I want to open the next PR. If there are any comments I'll address them later.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Eq.hs
```diff
@@ -19,11 +20,23 @@ import PlutusCore.Rename.Monad
...
+type HashableTermConstraints uni fun ann =
+  ( GEq uni
+  , Closed uni
+  , uni `Everywhere` Eq
+  , uni `Everywhere` Hashable
+  , Hashable ann
+  , Hashable fun
+  )
+
+instance HashableTermConstraints uni fun ann => Hashable (Term Name uni fun ann)
```
michaelpj, line 38: Why don't we just have a `Hashable name` constraint and have it work for any `name`?

plutus-core/plutus-core/src/PlutusCore/Crypto/BLS12_381/Pairing.hs
kwxm: This is over a year and a half late, but I don't think this matters.  Values of this type can only be created during program execution: you should never be able to have a constant at compile time, unless maybe GHC or some other compiler has been overly helpful and done some compile-time computation for us before we get the AST.  I'm not sure if that can happen though; probably not since there's no way to serialise these things, so if we got such an AST we probably wouldn't be able to do anything with it.

PR #5631, opened 2023-11-12
Turn on PIR and UPLC optimizers when lifting

zliu41:
I cannot think of any reason not to.


plutus-tx/src/PlutusTx/Lift.hs
```diff
@@ -72,16 +72,13 @@ safeLift
 safeLift v x = do
     pir <- liftQuote $ runDefT () $ Lift.lift x
     tcConfig <- PLC.getDefTypeCheckConfig $ Original ()
-    -- NOTE:  Disabling simplifier, as it takes a lot of time during runtime
```
michaelpj, line 75: This is the reason! At the moment, we run the optimizer at user-program-compile time, i.e. as part of the plugin which runs during GHC compiling a module. In contrast, lifting happens at _user-program-runtime_, i.e. when they're busy submitting transactions or whatever. That's quite a different proposition! That means that if we, say, make the compiler accidentally quadratic, that could make people's actually deployed systems work badly, rather than "merely" giving them a slow build.

Maybe it's fine, but I do think it's a bit risky. OTOH it does seem to make some stuff better so :shrug:

zliu41: Our optimizers are quite fast, and the difference is too big not to do it. User-program-runtime is still off-chain time which is a lot less valuable than on-chain time.

michaelpj: > Our optimizers are quite fast

Sure, but now we _have_ to keep them fast.

zliu41:
```
- ({cpu: 49082100
- | mem: 213500})
+ ({cpu: 8602100
+ | mem: 37500})
```

Given numbers like these, even if it's not that fast, I think it's still worth it.
What's the big deal if your off-chain code takes a few more seconds to run?

michaelpj: Once we accidentally started inlining all types, making the compiler at least quadratic. It was quite noticeable, modules took tens of seconds to compile. It's very easy for us to introduce bad asymptotics in our compiler, and with bad asymptotics you can get noticeable-to-humans slowdowns quite easily. 

>  What's the big deal if your off-chain code takes a few more seconds to run?

Seriously? People might have systems that submit many transactions, or where latency is important. They might get quite annoyed if we add multiple seconds of delay into these processes.

Obviously, we should just not make the compiler slow. I just think it makes compiler performance _much more critical_ than it is otherwise. I'm not saying we shouldn't do it, just that it's risky.

zliu41: Even if someone needs to submit many transactions in a short period of time, I don't imagine them needing to lift something over and over again. If they do need that, then they may also need to compile things over and over again, and the compiler performance would affect them anyway.

How about providing both optimized version and unoptimized version of lift?

michaelpj: The point of lifting things is usually to provide transaction-specific parameters to scripts, so I think you might well call it many times.

Maybe we should let people pass the compilation options to `lift` or something, IDK.

PR #5634, opened 2023-11-14
CSE for UPLC

zliu41:
See Note [CSE] in `Cse.hs` for how it works.

This currently doesn't help `AsData`'s field accessors much, since those TH-generated field accessors are not CSE-friendly. I added a test case with manually written field accessors (See `recordFieldsManual` in `AsData.Budget.Spec`), and its performance is now comparable with `patternMatching`. We can easily generate those more efficient field accessors instead.

Besides `recordFieldsManual`, two unit tests are added in plutus-core/untyped-plutus-core/test/Transform/Simplify.hs.


plutus-core/plutus-ir/test/PlutusIR/Compiler/Recursion/factorial.golden
```diff
@@ -2,38 +2,38 @@
   1.1.0
   [
     [
-      (lam s_800 [ s_800 s_800 ])
+      (lam s_836 [ s_836 s_836 ])
```
effectfully, line 5: Kinda weird that only uniques change in this file and nothing else. CSE performs an additional renaming?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Simplify.hs
```diff
@@ -74,3 +89,10 @@ simplifyTerm opts = simplifyNTimes (_soMaxSimplifierIterations opts)
     caseOfCase' = case eqT @fun @DefaultFun of
         Just Refl -> caseOfCase
         Nothing   -> id
+
+    cseStep :: Int -> Term name uni fun a -> m (Term name uni fun a)
+    cseStep _ = case (eqT @name @Name, eqT @uni @PLC.DefaultUni) of
```
effectfully, line 94: Hardcoding `Name` here is kinda weird. Any name should work as long as it has a `TermUnique`.

effectfully (commented):
I've reviewed everything apart from the actual algorithm. I'll create a ticket for myself to review it, since it's definitely going to take me a while. Great docs by the way.

effectfully:
UPD: no, sorry, not everything, I didn't review all the golden files as it's infeasible.

plutus-core/plutus-core/src/PlutusCore/Compiler/Types.hs
```diff
@@ -6,5 +6,14 @@ import PlutusCore.Builtin
...
-type Compiling m uni fun name =
-  (ToBuiltinMeaning uni fun, MonadQuote m, HasUnique name TermUnique, Ord name)
+import Data.Hashable
+
+type Compiling m uni fun name a =
+  ( ToBuiltinMeaning uni fun
+  , MonadQuote m
+  , HasUnique name TermUnique
+  , Ord name
+  , Typeable name
```
michaelpj, line 16: As I commented on your other PR, I don't see why we can't just have a `Hashable name` constraint?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+
+mkCseTerm ::
+  forall uni fun ann m.
+  (MonadQuote m, Eq (Term Name uni fun ())) =>
+  [Term Name uni fun (Path, ann)] ->
+  -- | The original annotated term
+  Term Name uni fun (Path, ann) ->
+  m (Term Name uni fun ann)
+mkCseTerm ts t = do
+  cs <- sortOn (negate . termSize . ccTerm) <$> traverse mkCseCandidate ts
```
bezirg, line 348: Perhaps have a `cseStrategy = sortOn (negate . termSize)` because I think I saw this sortOn also somewhere before

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+import Data.List.Extra
+import Data.Traversable
+
+{- Note [CSE]
+
+-------------------------------------------------------------------------------
+1. Simplifications
+-------------------------------------------------------------------------------
+
+This is a simplified (i.e., not fully optimal) implementation of CSE. The two simplifications
```
michaelpj, line 33: Did you use any references for the implementation? Is this a standard algorithm? What variants do we know of? What does GHC do?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+        (case y [ (1+(2+x)) + (3+x)
+                , (2+x) + (3+x)
+                , 4+x
+                ]
+        )
+
+The implementation makes several passes on the given term. The first pass collects builtin
+arity information as described above.
+
+In the second pass, we assign a unique ID to each `LamAbs`, `Delay`, and each `Case` branch.
```
michaelpj, line 59: Why do we need one for a case branch? A case branch can be an arbitrary term, and if it's a function then there will be a lambda inside it, which you will give an id to?

michaelpj: I guess I'm missing the "abstract picture". What are we trying to do here? Identify unique scopes? or what?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+  Term name uni fun ann ->
+  HashMap fun Int
+calcBuiltinArity = foldrOf termSubtermsDeep go Map.empty
+  where
+    go :: Term name uni fun ann -> HashMap fun Int -> HashMap fun Int
+    go = \case
+      t@Apply{}
+        | (Builtin _ fun, args) <- splitApplication t ->
+            Map.insertWith max fun (length args)
+      _ -> id
```
bezirg, line 232: Again, I would rather have a hardcoded DefaultFun builtinarity :P

plutus-tx-plugin/plutus-tx-plugin.cabal
```diff
@@ -40,6 +40,7 @@ common lang
     -Wincomplete-record-updates -Wredundant-constraints -Widentities
     -Wunused-packages -Wmissing-deriving-strategies -fobject-code
     -fno-ignore-interface-pragmas -fno-omit-interface-pragmas
+    -fforce-recomp
```
bezirg, line 43: Why?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+`(4+x, "0.1.4")`.
+
+Now, each `(term, path)` pair whose count is greater than 1 is a CSE candidate.
+In the above example, the CSE candidates are `(2+x, "0.1")` and `(1+(2+x), "0.1")`.
+Note that `3+x` is not a CSE candidate, because it has two paths, and neither has a count
+greater than 1. `2+` is also not a CSE candidate, because it is workfree.
+
+The CSE candidates are then processed in descending order of their `termSize`s. For each CSE
+candidate, we generate a fresh variable, create a LamAbs for it under its path, and substitute
+it for all occurrences in the original term whose paths are descendents (or self) of
```
michaelpj, line 86: So we do a substitution pass for every CSE candidate? Sounds quadratic.

plutus-benchmark/plutus-benchmark.cabal
```diff
@@ -47,7 +47,7 @@ common lang
     -fno-specialise -fno-spec-constr -fno-ignore-interface-pragmas
     -fno-omit-interface-pragmas -fno-strictness
     -fno-unbox-strict-fields -fno-unbox-small-strict-fields
-    -fno-full-laziness
+    -fno-full-laziness -fforce-recomp
```
bezirg, line 50: Why is this needed?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+Here's another example:
+
+force (force ifThenElse
+         (lessThanEqualsInteger 0 0)
+         (delay ((1+2) + (1+2)))
+         (delay (1+2))
+      )
+
+In this case, the first two occurrences of `1+2` can be CSE'd, but the third occurrence
+can not. This is ensured by checking the path when substituting `cse1` for `1+2`. The result is
```
michaelpj, line 121: Why not? I think the answer is:
1. There is no scope enclosing both of them? If so, this is probably a bug, there should be a scope for "the top level" for sure.
2. They have differing paths, corresponding to the two different delays. In this instance, since the scopes come from `delay`s, you _could_ share the value across both (since there isn't a newly introduced variable that they might depend on). However, that might lead to the value being computed sooner than it would otherwise have been, which seems bad.

So, not at all obvious. And I don't know how to reason about it because you haven't told me what your algorithm is trying to do!

bezirg (approved):
This looks like a clear win in terms of execution (cpu&mem), and has mostly a positive effect on the script size as well. Nice work!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+may cause many optimization opportunities to be permanently lost.
+
+Therefore, this is what we do: first run the simplifier iterations. Then, run the CSE iterations,
+interleaving with the simplifier. For example, suppose max-simplifier-iterations-uplc=12, and
+max-cse-iterations=4. We first run 12 iterations of the simplifier, then run 4 iterations
+of CSE, with a simplifier pass after each iteration of CSE (i.e., the simplifier is run for a
+total of 16 times).
+
+Finally, since CSE can change the order or the number of occurrences of effects, it is only run
+when conservative optimization is off.
```
michaelpj, line 185: Can't we make it conservative by only CSE-ing pure expressions?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+run it along with the rest of the simplifier. That is, however, a bad idea. CSE does the reverse
+of inlining; inlining tends to expose more optimization opportunities, and conversely, CSE
+tends to destroy optimization opportunities. Running CSE on a not-fully-optimized program
+may cause many optimization opportunities to be permanently lost.
+
+Therefore, this is what we do: first run the simplifier iterations. Then, run the CSE iterations,
+interleaving with the simplifier. For example, suppose max-simplifier-iterations-uplc=12, and
+max-cse-iterations=4. We first run 12 iterations of the simplifier, then run 4 iterations
+of CSE, with a simplifier pass after each iteration of CSE (i.e., the simplifier is run for a
+total of 16 times).
```
michaelpj, line 182: Have you observed that this makes a difference? i.e. that we actually get new inlining opportunities that we want to exploit? I have a feeling that GHC for example just runs CSE once at the end and then doesn't run the inliner again. Not sure.

michaelpj (commented):
Just reviewed the algorithm description (thanks for that!) first, since that may well spur a bunch of discussion.

> since those TH-generated field accessors are not CSE-friendly

Since I guess you've looked at it, can you tell us why not? Note that the accessors are _not_ generated by TH. The TH generates pattern synonyms, and then _GHC_ generates the accessors. So I don't know how easily we can fix them.

michaelpj:
https://kar.kent.ac.uk/21455/1/Common_Subexpression_Elimination_in_a_Lazy.pdf seems relevant?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+{- Note [CSE]
+
+-------------------------------------------------------------------------------
+1. Simplifications
+-------------------------------------------------------------------------------
+
+This is a simplified (i.e., not fully optimal) implementation of CSE. The two simplifications
+we made are:
+
+- No alpha equivalence check, i.e., `\x -> x` and `\y -> y` are considered different expressions.
```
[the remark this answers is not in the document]
effectfully: Can we simply compute the hash after renaming the expression? Haven't reviewed the algorithm description yet.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: > However, that might lead to the value being computed sooner than it would otherwise have been, which seems bad.

Ok, this is not a great example because (1+2) appears in both branches, so it will definitely be computed. A better example would be an expression that appears in two of the three case branches, in which case we definitely can't CSE it.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: Yep, I tried making CSE part of the simplifier and it's much worse that way.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: Yeah we can do that later as an improvement.

zliu41 (commented):
Responded to some of the comments. Will address the other comments later.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
bezirg: Alpha-renaming may perhaps bring a huge benefit. But personally, I am ok to bring it later on a separate PR

plutus-benchmark/plutus-benchmark.cabal
zliu41: Oops, I always forget to remove this!

plutus-core/plutus-core/src/PlutusCore/Compiler/Types.hs
```diff
@@ -6,5 +6,14 @@ import PlutusCore.Builtin
...
 
-type Compiling m uni fun name =
-  (ToBuiltinMeaning uni fun, MonadQuote m, HasUnique name TermUnique, Ord name)
+import Data.Hashable
+
+type Compiling m uni fun name a =
+  ( ToBuiltinMeaning uni fun
+  , MonadQuote m
+  , HasUnique name TermUnique
+  , Ord name
```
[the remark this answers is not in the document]
zliu41: Yes this has been discussed before but hasn't been fixed.

plutus-core/plutus-core/src/PlutusCore/Compiler/Types.hs
zliu41: Probably you are right; let me try and see if it works.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: And it is difficult to do for several reasons: (1) what about non-closed terms; (2) substitution is also harder

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+- The builtin function arity information is approximate: rather than using the accurate arities,
```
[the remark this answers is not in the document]
zliu41: See https://github.com/input-output-hk/plutus/pull/5634#discussion_r1392716595

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: Yeah, I don't know how to describe this precisely, but the purpose is to avoid evaluating something that otherwise may not be evaluated. E.g.,

- If there's a common subexpression in a parent scope and a child scope (e.g., "0.1", and "0.1.2"), then it is safe to perform CSE
- On the other hand, if it is in two scopes that don't have parent-child relationship (e.g., "0.1.2" and "0.1.3"), then it is not safe to perform CSE. Because if you do perform CSE, the new "binding" has to be in scope "0.1", but it is possible that "0.1" is evaluated without evaluating either "0.1.2" or "0.1.3".

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+
+Now, each `(term, path)` pair whose count is greater than 1 is a CSE candidate.
+In the above example, the CSE candidates are `(2+x, "0.1")` and `(1+(2+x), "0.1")`.
+Note that `3+x` is not a CSE candidate, because it has two paths, and neither has a count
+greater than 1. `2+` is also not a CSE candidate, because it is workfree.
+
+The CSE candidates are then processed in descending order of their `termSize`s. For each CSE
+candidate, we generate a fresh variable, create a LamAbs for it under its path, and substitute
+it for all occurrences in the original term whose paths are descendents (or self) of
+the candidate's path. The order is because a bigger expression may contain a small subexpression.
```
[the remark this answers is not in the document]
zliu41: No because the bigger CS contains the smaller CS.
You can do it the other way around, but then you not only need to perform substitution in the original term, but also in the CS themselves

zliu41 (commented):
> Since I guess you've looked at it, can you tell us why not? Note that the accessors are not generated by TH. The TH generates pattern synonyms, and then GHC generates the accessors. So I don't know how easily we can fix them.

The field accessors from pattern synonyms are less inlined, and thus less CSE-friendly.

```
-- Manual
(\d -> (\tup -> ...something... (force (force fstPair) tup))
       (unConstrData d)
       +
       (\tup -> ...something_else... (force (force fstPair) tup))
       (unConstrData d)
)

-- From pattern synonyms

(\d -> (\f -> f something + f something_else)
       (\tup x -> ...x... (force (force fstPair) tup) (unConstrData d))
)
```

In the latter, `f` is not inlined, and as a result, `unConstrData d` is not a common subexpression, and has to be evaluated multiple times.

GHC doesn't have this problem since it can perform full-laziness, but ours is not a lazy language and we can't do that (unless in some very clever way).

Also I verified that the improvements are due to CSE, not the additional simplifier passes. Increasing the number of simplifier passes from 12 to 16 alone has no effect whatsoever.

zliu41:
> In the latter, f is not inlined, and as a result, unConstrData d is not a common subexpression, and has to be evaluated multiple times.

And this makes me think that we should probably make the inliner "dumber" (that is, inline more things, to expose opportunities for CSE).

plutus-benchmark/marlowe/test/role-payout/9.2/fc8c5f45ffcdb024c21e0f34b22c23de8045a94d5e1a5bda1555c45ddb059f82.eval.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 257402171
-| mem: 905340})
\ No newline at end of file
+({cpu: 252572171
+| mem: 884340})
```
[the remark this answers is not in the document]
kwxm: If the filenames said `budget` somewhere we could probably write a shell script that you could locally pipe the git diff though to get some kind of summary.

kwxm:
There are a lot of comments on this already so maybe I'll wait unt

kwxm:
il

michaelpj:
> The field accessors from pattern synonyms are less inlined, and thus less CSE-friendly.

I also checked and I think it's not quite what you wrote. With the pattern synonym version, all the field accessors _use_ the "matching function" (`mInts` in this case). The field accessors themselves are inlined, but the matching function is not, because it is used several times, and is quite large. I suspect that's the same situation as we would see with your manually written accessors if they were used multiple times in the program.

zliu41:
> I suspect that's the same situation as we would see with your manually written accessors if they were used multiple times in the program.

See https://github.com/input-output-hk/plutus/pull/5636#issuecomment-1813400215
Looks like CSE is more effective than I thought!

zliu41:
By the way, I retract what I said about "CSE making the compilation of marlowe validator much slower". The marlowe validator already takes about 5s to compile on master on my laptop, and it takes a similar amount of time with CSE (I didn't measure precisely, though)

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
effectfully: > Yep, I tried making CSE part of the simplifier and it's much worse that way.

It's worth mentioning that in the docs.

effectfully (commented):
Only reviewed the description so far. _Amazing_ docs, thank you. I need to meditate on it for a bit.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
michaelpj: > I don't understand why though, I find our type checking strategy convoluted for no good reason, but perhaps I'm missing something

Didn't you write the typechecker? :p

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
michaelpj: I think there are maybe three answers?
1. They represent places where computation stops, i.e. subexpressions are not immediately evaluated, and as such they shouldn't be hoisted above that point lest they be evaluated too soon.
2. They represent places where evaluation may happen multiple times with a different environment (lambdas) or not (delays) i.e. subexpressions may not always be evaluated to the same thing so shouldn't be shared between multiple evaluations.
3. They may bind new variables, in which case it's actually incorrect to hoist a subexpression out of one.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
michaelpj: Or you go over the whole program building up a single subtitution, but applying the substitution to new substitution-rhss that you insert. Maybe that's also quadratic, though! It seems less quadratic...

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
michaelpj: I wasn't assuming we knew it would definitely be computed, just that we could lift it anyway (and then get computation too soon).

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+In the above example, there are three occurrences of `2+x`, whose paths are "0.1", "0.1.2"
+and "0.1.3", respectively. The first path is an ancestor of the latter two. Therefore,
+the count for `(2+x, "0.1")` is 3, while the count for `(2+x, "0.1.2")` and `(2+x, "0.1.3")`
+is 0. The following all have a count of 1: `(3+x, "0.1.2")`, `(3+x, "0.1.3")` and
+`(4+x, "0.1.4")`.
+
+Now, each `(term, path)` pair whose count is greater than 1 is a CSE candidate.
+In the above example, the CSE candidates are `(2+x, "0.1")` and `(1+(2+x), "0.1")`.
+Note that `3+x` is not a CSE candidate, because it has two paths, and neither has a count
+greater than 1. `2+` is also not a CSE candidate, because it is workfree.
```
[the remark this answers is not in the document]
michaelpj: Arguably if you do that you might want to lift out of lambdas too by making the subexpression into a lambda that takes the free variables as arguments.

Also join points have to be tail-called and expressions we CSE need not be.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
michaelpj: It's not that complicated IMO. It is a little fiddly because we need to pass `BuiltinsInfo` around, but that's very mechanical.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
effectfully: I basically rewrote it, not just wrote it, and I was opposed to using global uniqueness there from before the beginning:

Screenshot from 2023-11-16 11-45-15

It took me a ridiculous amount of time to weed all the bugs out of the original implementation and to this day I still don't understand why we need global uniqueness there instead of having type values at the type level and reducing them using an abstract machine or whatever like people do.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+variable `cse1` for it, perform substitution, and create a `LamAbs` under path "0.1" (i.e., around
+the body of `y`). After processing this CSE candidate, the original term becomes
+
+\x y -> (\cse1 -> cse1
+                  +
+                  (case y [ cse1 + (3+x)
+                          , (2+x) + (3+x)
+                          , 4+x
+                          ]
+        ) (1+(2+x))
```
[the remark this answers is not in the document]
michaelpj: Yeah, this seems a little tricky. What if the new CSE term now includes `cse1`? The term below hoists `cse2` outside the scope of `cse1`! At least it should probably be inside. I notice the the examples are inconsistent: below there is an example where we do put `cse2` inside, which seems more correct.

I guess that if we do that it's safe, but the argument is non-trivial:
- Suppose we have two CSE candidates `cse1` and `cse2`, where `cse1` occurs within `cse2`.
- Then if we decide to hoist `cse1`, we must do so to the level indicated by the path to `cse2`, since `cse1` is a subexpression of `cse2`
- We thus in theory need to extend the path for `cse2` by one step for the new binding, but since it occurs right under the previous last step, we can just use the old path so long as we put the binding for `cse2` under the other CSE bindings.

I'm not even sure if that is correct. I hope we have some tests for this and we need a careful argument!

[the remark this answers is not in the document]
effectfully: > The term below hoists cse2 outside the scope of cse1! At least it should probably be inside. 

Seems correct to me? Not sure which term you're referring too. I've been ignoring indices in the names, only looking at semantics.

mjaskelioff (approved):
Looks good to me.  I appreciate the long explanation at the beginning. 
I would definitely use the real builtin arity. Calculating it in this way seems messy and adds unnecessary complexity.
The path looks like a good abstraction for the boundaries where CSE may happen. I wonder if having an explicit transformation into a tree of top level, LamAbs, Delay, and Case nodes would make the correctness easier to see, but I guess that reasoning about prefixes is easy enough.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: https://input-output.atlassian.net/browse/PLT-8659

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+-------------------------------------------------------------------------------
+
+This is a simplified (i.e., not fully optimal) implementation of CSE. The two simplifications
+we made are:
+
+- No alpha equivalence check, i.e., `\x -> x` and `\y -> y` are considered different expressions.
+- The builtin function arity information is approximate: rather than using the accurate arities,
+  we simply use the maximum number of arguments applied to a builtin function in the program
+  as the builtin function's arity. The arity information is used to determine whether a builtin
+  application is possibly saturated.
```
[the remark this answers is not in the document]
zliu41: Good point. Will address this later: https://input-output.atlassian.net/browse/PLT-8659

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+
+In the above example, the ID of `\x` is 0, the ID of `\y` is 1, and the IDs of the
+three case branches are 2, 3, 4 (the actual numbers don't matter, as long as they are unique).
+The path for the first `1+(2+x)` and the first `2+x` is "0.1"; the path for the second
+`1+(2+x)` and the second `2+x` is "0.1.2"; the path for `4+x` is "0.1.4".
+
+In the third pass, we calculate a count for each `(term, path)` pair, where `term` is a
+non-workfree term, and `path` is its path. If the same term has two paths, and one is an
+ancestor (i.e., prefix) of the other, we increment the count for the ancestor path in both
+instances.
```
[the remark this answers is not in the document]
zliu41: Because we need to aggregate the counts on each path via `Map.fromListWith`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+
+In the above example, there are three occurrences of `2+x`, whose paths are "0.1", "0.1.2"
+and "0.1.3", respectively. The first path is an ancestor of the latter two. Therefore,
+the count for `(2+x, "0.1")` is 3, while the count for `(2+x, "0.1.2")` and `(2+x, "0.1.3")`
+is 0. The following all have a count of 1: `(3+x, "0.1.2")`, `(3+x, "0.1.3")` and
+`(4+x, "0.1.4")`.
```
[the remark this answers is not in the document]
zliu41: Because an expression may have multiple occurrences in non-0.1 pairs, e.g., one in 0.1.3 and one in 0.1.3.2.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: We could implement a version of CSE that optimizes for size in the future. Some users to value size more than cost.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: I think it is quadratic only in very pathological cases; and as such I agree we can leave it as it is in the initial version.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: > But you do perform substitution in CSs?

No, what I meant to say: if we process `1+(2+x)` first, then we only need to perform substitution in the original term itself. On the other hand, if we process `2+x` first, we must also perform substitution in the CSs, i.e., turning `1+(2+x)` into `1+cse2`. Otherwise, when we process `1+(2+x)` later, we will not find any `2+x` in the original term.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: > But then shouldn't we handle such lambdas the same way regardless of whether they were originally in the program or were added by CSE?

Good point. What I'm not sure is: whether CSE should handle this, or should we implement another optimization, that floats terms inside an immediately-applied lambda that don't depend on the lambda-bound variable, outside of the lambda (i.e., similar to full-laziness).

> Suppose we have two CSE candidates cse1 and cse2, where cse1 occurs within cse2.

That's why the CSE candidates are processed in descending order of size. If we process cse2 first, then there's no problem.

plutus-core/plutus-core/src/PlutusCore/Compiler/Types.hs
zliu41: Oh, I remember why I didn't do that: the current implementation of CSE only works correctly for `Name`. We can't just do the substitutions that we do for `DeBruijn` terms.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Simplify.hs
zliu41: I think the current implementation of CSE only works correctly for `Name`. We can't just do the substitutions that we do for `DeBruijn` terms.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: I think point 1 is correct. I'll add it to the doc. I don't agree with 2 or 3: if it is incorrect to hoist something (because, e.g., it depends on the lambda-bound variable), then it can't be a common subexpression anyway.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: > It's worth mentioning that in the docs.

Added "Give it a try if you want to see how bad it is!"

zliu41 (commented):
Addressed/responded to some comments.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
effectfully: Sorry, still don't get it. Can you have a map from hashes of terms to lists of paths, without even any counts, and then post-process each list of paths? If you have `2+x :-> ["0.1", "0.1.2", "0.1.3"]`, then during post-processing you figure out that `"0.1"` is the root and insert a binding at that point, ignoring all the path extensions. Technically, it's not even a list of paths as a value, it's a trie of paths and "post-processing" means simply taking the root of each forest at the beginning of the trie (unless the forest consists only of the root, i.e. the subexpression only occurs once).

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
effectfully: > What I'm not sure is: whether CSE should handle this, or should we implement another optimization, that floats terms inside an immediately-applied lambda that don't depend on the lambda-bound variable, outside of the lambda (i.e., similar to full-laziness).

I don't believe it's going to fly. Plutus Core programs tend to have a long prefix of immediately applied lambdas, are you going to extract half a program outside those by picking every single subexpression that doesn't happen to reference some of the variables? Plus it'll create bindings for things that we don't really want to create bindings for, like what's the point of turning `(\x -> 1 + 2 + x) e` into `(\y -> (\x -> y + x) e) (1 + 2)` -- no work is saved and some evaluation cycles are wasted on creating a variable. And each such transformation only creates more immediately applied lambdas.

I _suspect_ that full-laziness-like transformations only help when you lift a subexpression past a non-applied lambda. CSE is different in that regard: it helps whenever a subexpression occurs more than one time, regardless of whether you're inside an applied lambda or not.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
effectfully: > Because an expression may have multiple occurrences in non-0.1 pairs, e.g., one in 0.1.3 and one in 0.1.3.2.

Sure, but once you have `0.1`, what's even the point of storing other ones? Just increment the counter of `0.1`. Regardless of whether you encounter `0.1.3.2` or `0.1.3.6.3.2.1.6.7.5.8.4` later, it'll still become a variable bound at `0.1`. Is that what you mean by their count be 0? Still haven't reviewed the code.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
michaelpj: > if it is incorrect to hoist something (because, e.g., it depends on the lambda-bound variable), then it can't be a common subexpression anyway.

I guess the problem is the related notions of "common sub-expression" and "structurally identical term". In the presence of non-globally-unique names you might have structurally identical terms that are not in fact identical, which might require this kind of care. Are we assuming global uniqueness?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
michaelpj: I think this comment makes more sense in the context of your other idea about indexing via path.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
michaelpj: > Good point, but mixing lambda-lifting and CSE into a single pass sounds really complicated.

If we think of delay as like `\() -> ...` then lifting out of a delay and delaying the value is pretty similar to lambda-lifting, so I would argue that your proposal is already mixing things.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
michaelpj: Can we make a pathological test case to see?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
michaelpj: > That's why the CSE candidates are processed in descending order of size. If we process cse2 first, then there's no problem.

That's a bit subtle and needs to be written down. I had to think about it for a while. Here's my argument:

By processing CSE candidates in size order, we also implicitly process them in dependency order: if we have two CSE candidates `cse1` and `cse2`, then we cannot end up in a situation where `cse2` would appear in `cse1`, because then the original `cse1` would have had to have been bigger than `cse2`.

I'm actually not sure this is a sound argument, because doing some rounds of CSE can change the size ordering. Can we end up with a situation like this?
1. `cse1` < `cse2` < `cse3`
2. Pull out `cse1`
3. `cse3` < `cse2`, and `cse3` appears a subexpression of `cse2`
4. Pull out `cse2`
5. Pull out `cse3`, break the RHS of `cse2`

Looking at that, in fact I think what we need to be safe is just:
1. Bindings for subsequent CSE candidates go inside the previous ones
2. We only substitute for a CSE candidate inside the scope of the binding
Which is much simpler and probably holds.

That still implies that this example is wrong and `cse2` should be bound inside the scope of `cse1`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,398 @@
...
+tends to destroy optimization opportunities. Running CSE on a not-fully-optimized program
+may cause many optimization opportunities to be permanently lost.
+
+Therefore, this is what we do: first run the simplifier iterations. Then, run the CSE iterations,
+interleaving with the simplifier. For example, suppose max-simplifier-iterations-uplc=12, and
+max-cse-iterations=4. We first run 12 iterations of the simplifier, then run 4 iterations
+of CSE, with a simplifier pass after each iteration of CSE (i.e., the simplifier is run for a
+total of 16 times).
+
+Finally, since CSE can change the order or the number of occurrences of effects, it is only run
```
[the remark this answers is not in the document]
michaelpj: Yeah, we do worry about log order.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
effectfully: I mean, lambda-lifting and getting rid of free variables. Mine doesn't attempt to get rid of free variables. Not sure if I understood you correctly.

effectfully:
I've spent some time pondering the general algorithm and I think what this PR does is possible to do better performance-wise. I don't kwow it we care about that though, maybe for a typical Plutus program the current algorithm is just fine, so feel free to ignore.

You check whether one path is an extension of another and do different things depending on the answer, but I think we can avoid this logic entirely. We're going to construct a DAG where each node is a `HashMap` from subexpressions to their counts and each edge is an ID (i.e. a step in a path). If we take your example

```haskell
\x y -> (1+(2+x))
        +
        (case y [ (1+(2+x)) + (3+x)
                , (2+x) + (3+x)
                , 4+x
                ]
        )
```

then `2+x`, `1+(2+x)`, `case y <...>` and `1+(2+x) + case y <...>` are all going to be keys in a `HashMap` stored in a node, with values all being `1`.

Similarly, `3+x` is going to be stored together with `2+x`, `1+(2+x)` and `(1+(2+x)) + (3+x)` in another node, with the ID of the first branch of the case expression being an edge between the previosly described node and this one.

Each time you enter a new scope, you start from an empty `HashMap` and populate it with all the subexpressions appearing at the current path. Once this pass is completed, you have a DAG where all common subexpressions appearing at the same path are accounted for by their count being greater than 1 in the nodes of the graph. It only remains to process common subexpressions that occur at different paths, but since you have a DAG that guarantees that travelling by its edges always gives you an extension of the current path you can do exactly that: travel by the edges, merge the accumulator hashmap with the hashmap of each of the nodes and when you reach a leaf of the AST compute which subexpressions occurring at any point in the current path to pull out and where to put them.

The actual implementation may avoid materializing DAGs and just add some in-place logic doing the same.

Fundamentally, the idea of your algorithm is reader-based, but you handle it with a state-based implementation and then you need to check path prefixes to account for the extra power of the implementation that you don't use.

But as per other comments, maybe that's fine, maybe we want a more general algorithm, so that in future we can optimize not only for evaluation costs, but also for the size of the script, which requires pulling common subexpressions occurring at different paths.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: @michaelpj I don't get your point at all. What do you think is the correct output for this example?

`cse2` is smaller than `cse1`. Shouldn't the smaller one be in the outer scope? I.e., you'd want

```
let cse2 = 2 + x in
let cse1 = 1 + cse2 in body
```

If you flip it you'd get

```
let cse1 = 1 + (2 + x) in
let cse2 = 2 + x in body
```

> Can we end up with a situation like this?

In your example you are processing CSE candidates in asencding order of size, while I'm proposing descending order.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: > Can we make a pathological test case to see?

Something like

```
((1+2) + (3+4) + (5+6) + ... + (999999 + 1000000))
+
((1+2) + (3+4) + (5+6) + ... + (999999 + 1000000))
```
? Do you have a worse pathological case in mind?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: > but once you have 0.1, what's even the point of storing other ones?

I guess I don't understand your point - but this is exactly what I'm doing, no? If the same expression has "0.1", "0.1.3" and "0.1.3.2", then the count for "0.1" will be incremented in all three cases, and in the end, the count of "0.1" is 3, and the count of the other two paths is 0.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: Here's an example. Suppose for `2+x`, we currently have the following counts: 
```
"0.1.2" -> 10, "0.1.3" -> 20, "0.2.2" -> 30
```

Now we process path "0.1". After processing it, the counts will become: 
```
"0.1" -> 10, "0.1" -> 20, "0.2.2" -> 30`, "0.1" -> 1
```

Finally we use `Map.fromListWith` to aggregate them, and get

```
"0.1" -> 31, "0.2.2" -> 30
```

Does that make sense?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: > you might have structurally identical terms that are not in fact identical, which might require this kind of care. Are we assuming global uniqueness?

Hmm, I guess that is indeed possible. I'll make it do renaming before CSE.

zliu41 (commented):
Responded to more comments.

zliu41:
> travel by the edges, merge the accumulator hashmap with the hashmap of each of the nodes and when you reach a leaf of the AST compute which subexpressions occurring at any point in the current path to pull out and where to put them.

How exactly does this work? Say a common subexpression `e` occurs at 0.1, 0.1.2 and 0.1.3. When you travel from 0.1 to 0.1.2, you decide to pull out `e`. That's fine. What happens if you then travel from 0.1 to 0.1.3, and also want to pull out `e`?

> I think what this PR does is possible to do better performance-wise.

My implementation can be slightly improved asymptotically: currently for each expression, it is `O(p^2)` due to `combinePaths`, where `p` is the number of paths this expression occurs. But it can be changed to: for each path, simply check all its ancestors, in which case it becomes `O(p * depth)`. If I understand what you propose correctly, it is not asymptotically better than that.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
michaelpj: And write something!!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
michaelpj: Sure, I think Roman's point is just that there's no point even _storing_ the counts for subsumed paths.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
michaelpj: Yes, that looks like a good example!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
michaelpj: Okay, I just had it backwards. You're doing them in _descending_ order of size :facepalm: 

Then yes, you need to put each subsequent one outside the other ones. And putting the new ones outside should always be safe.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
michaelpj: To be clear, when you say "perform substitution in the original term itself", you mean "the partially processed term with some candidates hoisted out"? Because you do want to hoist out the instance of `(2+x)` in the RHS of the newly created binding for the first CSE candidate, right?

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
```diff
@@ -226,6 +226,35 @@ multiApp = runQuote $ do
       app = mkIterAppNoAnn lam [mkConstant @Integer () 1, mkConstant @Integer () 2, mkConstant @Integer () 3]
   pure app
 
+-- | This is the first example in Note [CSE].
+cse1 :: Term Name PLC.DefaultUni PLC.DefaultFun ()
```
michaelpj, line 230: I feel like we've discussed quite a number of examples so far - this should have more than two tests!

effectfully:
> How exactly does this work? Say a common subexpression `e` occurs at 0.1, 0.1.2 and 0.1.3. When you travel from 0.1 to 0.1.2, you decide to pull out `e`. That's fine. What happens if you then travel from 0.1 to 0.1.3, and also want to pull out `e`?

When you reach a leaf you record that `e` is chosen to be pulled out and the same happens in the second case, then you merge all the decisions into a single map (or whatever) from subexpressions to where to insert new bindings for them and then in one fell swoop you insert the bindings and replace all subexpressions with their respective bindings in the original term. Maybe it's more passes than it could be, but I'd rather have more passes for testability and debugging purposes than lump all the logic into as few passes as possible.

> But it can be changed to: for each path, simply check all its ancestors, in which case it becomes `O(p * depth)`. If I understand what you propose correctly, it is not asymptotically better than that.

Yes, what I propose is one way to check only ancestors without iterating through all possible paths. I used a DAG for the model to be "tangible" (easier to think about, easier to explain), but if you do it directly, then that'll be great too.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,402 @@
...
+  forall uni fun ann.
+  (Eq (Term Name uni fun ())) =>
+  CseCandidate uni fun ann ->
+  Term Name uni fun (Path, ann) ->
+  Term Name uni fun (Path, ann)
+applyCse c = mkLamApp . transformOf termSubterms subst
+  where
+    candidatePath = fst (termAnn (ccAnnotatedTerm c))
+
+    subst :: Term Name uni fun (Path, ann) -> Term Name uni fun (Path, ann)
```
effectfully, line 365: Nit: I'd perhaps call it `cosubst` or something, to emphasize that it replaces a term by a variable, i.e. does the opposite of what "substitution" usually means.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,402 @@
...
+mkCseTerm ::
+  forall uni fun ann m.
+  (MonadQuote m, Eq (Term Name uni fun ())) =>
+  [Term Name uni fun (Path, ann)] ->
+  -- | The original annotated term
+  Term Name uni fun (Path, ann) ->
+  m (Term Name uni fun ann)
+mkCseTerm ts t = do
+  cs <- traverse mkCseCandidate ts
+  pure . fmap snd $ foldl' (flip applyCse) t cs
```
effectfully, line 353: Why not handle the whole list of candidates in `applyCse` instead of doing it one-by-one? Doesn't seem to be much more complicated and will save you some runtime. Maybe it doesn't matter.

effectfully (approved):
The code looks somewhat intimidating, but it's actually surprisingly straightforward. I've reviewed everything and it looks good to me, even though it's not optimal as we've discussed a lot here, but maybe it's efficient enough and we don't care about optimizing the whole thing.

I may have another run to reconcile the description of the algorithm with the implementation, but that shouldn't block the PR.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,402 @@
...
+      HashMap (Term name uni fun ()) (HashMap Path (Term name uni fun (Path, ann), Int))
+    addToMap t0
+      -- We don't consider work-free terms for CSE, because doing so may or may not
+      -- have a size benefit, but certainly doesn't have any cost benefit (the cost
+      -- will in fact be slightly higher due to the additional application).
+      --
+      -- `isWorkFree` currently doesn't check whether a builtin application is saturated,
+      -- or whether an term is the (possibly repeated) forcing of a builtin (which should
+      -- be workfree), so we check it separately.
+      | isWorkFree t0 || not (isBuiltinSaturated t0) || isForcingBuiltin t0 = id
```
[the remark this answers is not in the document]
effectfully: ^ this comment is an important one. Perhaps you can even remove the entire first pass?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: Ok, I'll expand the notes a bit

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: Actually paths indeed don't need to be map keys... I'll update the code

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: > you mean "the partially processed term with some candidates hoisted out"?

Yes.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: I initially did this and it made things worse, I think because the on the one hand, we avoid evaluating `force someBuiltin` multiple times, but on the other hand, we evaluate `LamAbs` and `Apply` more times, which is more expensive!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: > Why not handle the whole list of candidates in applyCse instead of doing it one-by-one?

It can be bit tricky to implement (if not a lot trickier). Different CSE candidates need to be inserted into different places in the original term. And different CSE candidates may contain each other. It is not immediately obvious to me what an elegant way of doing it might look like.

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
zliu41: I'll add the third example in the Note, as well as the pathological example.

zliu41 (commented):
Replied to a few new review comments.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: The inner HashMap (with Path as key) is replaced with a list in b32cd3cc99f9fd31b1cc0e52b838d050cedc4850, because in my example above, we can easily go from step 1 directly to step 3, bypassing step 2.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: I added a much smaller test case in 8147cbbfa8434c5cbd78aa1b7a237fe66a1b5e16.

It's not about CSE, but our test infrastructure simply can't handle anything close to the above term.
Even for a much smaller one, e.g., `(1+2) + (3+4) + ... + (99999 + 100000)` (i.e., more than 20 times smaller than the above term), the test times out _without doing CSE or anything else_!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -0,0 +1,402 @@
...
+      | otherwise =
+          Map.alter
```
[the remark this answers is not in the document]
zliu41: It makes `combinePaths` more awkward, since it would have to combine two lists of paths, while currently, it only needs to combine a single path with a list of paths.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: I renamed it to `substCseVarForTerm`. `cosubst` is a bit of a weird term.

zliu41:
> but if you do it directly, then that'll be great too.

I didn't end up doing it, because (1) it isn't enough to check for ancestors, but we also need to check for descendants; (2) in some cases, such as the pathological example I added, the number of paths per expression (two) is much smaller than the depth of the tree, so `path * depth` is worse than `path * path`. And in reality, deep ASTs may not be uncommon.

> then you merge all the decisions into a single map

I see. Time complexity wise, for each subexpression `e` that has `p` paths, in the worst case each path is in a different leaf node, so this is `O(p * depth)`. Like I said above, this is not necessarily better than `O(p * p)`.

plutus-tx-plugin/test/AsData/Budget/9.2/patternMatching-budget.eval.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 55946147
-| mem: 167776})
\ No newline at end of file
+({cpu: 9927795
```
zliu41, line 1: If only we could do something like this to the Marlowe validators!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
effectfully: > the test times out without doing CSE or anything else!

Could you please create a ticket for that just so we have it recorded?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
effectfully: > we evaluate `LamAbs` and `Apply` more times, which is more expensive!

How come? Were you eta-expanding builtin applications? Or is it just the cost of creating a binding? Sounds strange to me.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
effectfully: Right, thank you.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
effectfully: Fair enough.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Simplify.hs
michaelpj: What about requiring `HasUnique`? That's usually what we do in some situations, and it's properly polymorphic instead of this abomination.

plutus-core/plutus-core/src/PlutusCore/Compiler/Types.hs
michaelpj: HasUnique?

plutus-tx-plugin/test/AsData/Budget/9.6/recordFields.uplc.golden
```diff
@@ -1,60 +1,45 @@
...
+              (\cse ->
+                 (\cse ->
+                    (\lessThanInteger ->
+                       (\addInteger ->
+                          (\cse ->
+                             (\cse ->
+                                addInteger
+                                  (addInteger
+                                     (addInteger
+                                        (addInteger (addInteger cse cse) cse)
```
michaelpj, line 16: What is going on here? We're not CSE-ing single variables are we? That would definitely not be a win...

plutus-tx-plugin/test/AsData/Budget/9.6/recordFields.uplc.golden
```diff
@@ -1,60 +1,45 @@
...
-                                 (\ds ds ds ds -> ds)
-                                 (\void -> error)))
-                         (int d))
-                      (\ds ->
-                         `$mInts` ds (\ds ds ds ds -> ds) (\void -> error)))
-                   (int d))
-                (\ds -> `$mInts` ds (\ds ds ds ds -> ds) (\void -> error)))
-             (int d))
-          (\ds -> `$mInts` ds (\ds ds ds ds -> ds) (\void -> error)))
-       (\scrut cont fail ->
```
michaelpj, line 57: This is actually very nice: we CSE `$mInts ds`, i.e. the partial application.

plutus-tx-plugin/test/AsData/Budget/9.6/recordFields.uplc.golden
effectfully: > We're not CSE-ing single variables are we? 

I had it written in my notes to check specifically for that and then failed to actually do that, it's been a while since I felt like a complete moron so much...

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
effectfully: Another thing that I forgot to comment on:

> At least in this implementation, it doesn't CSE work-free expressions anyway.

Such an expression isn't necessarily work-free! You may have `f (\x -> x)` for example.

effectfully:
Oh yeah, my approval only holds as long as `cse` is changed to perform renaming before processing the term as per Michael's comment.

plutus-tx-plugin/test/AsData/Budget/9.6/recordFields.uplc.golden
zliu41: Hmm this seems like a bug, since it applies CSE on partially applied `addInteger`, which it isn't supposed to do.

zliu41: Actually not a bug - the partially applied `addInteger` is in fact not a builtin..

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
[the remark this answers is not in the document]
zliu41: > How come? Were you eta-expanding builtin applications? Or is it just the cost of creating a binding? Sounds strange to me.

No, it's because when we CSE something, it creates an additional LamAbs and Apply.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Simplify.hs
zliu41: > What about requiring HasUnique?

I also need `freshName`...

michaelpj: Okay, then just monomorphise to `Name` and `TyName`. The PIR compiler in fact runs on those types, and if that's the truth you might as well say it.

bezirg: Bummer, and I was looking forward to generalize PIR to work with any name/tyname. But of course this i s noot the only place where we do fix to Name/TyName. Other places that I can think of is the PIR typechecker and the rewrite rules

PR #5635, opened 2023-11-14
Kwxm/count mainnet builtin usage

kwxm:
[PLT_8615] This adds a mainnet script dump analysis to count the number of times each builtin appears, together with the results for the entire dump up to mid-October 2023.


zliu41 (approved):
:+1:

PR #5636, opened 2023-11-14
Add recordFieldsManual test

zliu41:
This is extracted from the CSE PR, so that we can see how much improvement CSE does to this test case.

```
-- Without CSE:
({cpu: 7629255
| mem: 20346})

-- With CSE:
({cpu: 5804520
| mem: 15162})
```

This is the test case that CSE has the biggest impact on, because the manually written field accessors are very CSE-friendly. See [this comment](https://github.com/input-output-hk/plutus/pull/5634#pullrequestreview-1730823270) for more details. This makes me think that we should probably make the inliner "dumber" (that is, inline more things), to expose more opportunities for CSE.


plutus-tx-plugin/test/AsData/Budget/Spec.hs
```diff
@@ -70,5 +77,19 @@ recordFields =
...
+recordFieldsManual :: CompiledCode (PlutusTx.BuiltinData -> Integer)
+recordFieldsManual =
+  $$( compile
+        [||
+        \d ->
+          let ints = PlutusTx.unsafeFromBuiltinData d
+              x = int1Manual ints
+              y = int2Manual ints
+              z = int3Manual ints
+              w = int4Manual ints
```
michaelpj, line 89: At the moment the accessor functions are all used exactly once, which means that we're pretty much guaranteed that the inliner will inline them. And that's very important, since in order for CSE to help it has to see the inlined versions. But in reality field accessors may well be used multiple times, which might inhibit inlining (and hence CSE). So I'd like to have at least one of them used repeatedly.

zliu41: > since in order for CSE to help it has to see the inlined versions

I don't think that's true. In this particular case, I think, e.g., `unConstrData d` will be processed regardless of inlining. But I can modify the test as you suggested.

zliu41:
@michaelpj updated the test; now not only is each the field accessors (`ints1Manual` etc.) used three times, but each of `x`, `y`, `z`, `w` is also used three times.

Without CSE:
```
-- pattern matching
({cpu: 55946147
| mem: 167776})

-- field accessor (manual)
({cpu: 23015765
| mem: 62058})

-- field accessor (TH)
({cpu: 36195493
| mem: 107926})
```

With CSE:

```
-- pattern matching
({cpu: 9927795
| mem: 27176})

-- field accessor (manual)
({cpu: 11292601
| mem: 27874})

-- field accessor (TH)
({cpu: 17541361
| mem: 49126})
```

Two interesting observations:

1. Contrary to what you said, now CSE actually has a much bigger impact on all three cases.
2. Without CSE, pattern matching is now the most expensive. With CSE, it becomes the cheapest. Field accessor (manual) is not far behind, while field accessor (TH) is still much worse.

plutus-tx-plugin/test/AsData/Budget/Spec.hs
michaelpj: That seems wrong. If we have two record accessors which both have `\d -> ... unConstrData d ...` then the two lambdas (correctly) create two different scopes (after all, there might be two different datas there!). The only way we can get sharing is if we inline and reduce them so that it's the _same_ d in both cases.

michaelpj:
Interesting! I'll be interested to see how that works out, I can't see it in my head!

PR #5643, opened 2023-11-18
Generate more efficient field accessors for asData

zliu41:
More work than using `RecordPatSyn`, but more efficient.


michaelpj:
One downside of this is that record pattern synonyms don't just generate accessors, they let you use the field name in all the ways you can use a record field name: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/pattern_synonyms.html#record-pattern-synonyms

Specifically, we lose the ability to do:
- `x = Foo { y = 100 }`
- `Foo { y = example }` (pattern)
- `Foo {y}` (pattern with puns)
- `f { y = 200 }`

I wonder if it would be possible to get the effect that you want by just making the matching function always-inline?

plutus-tx/src/PlutusTx/AsData.hs
```diff
@@ -87,25 +90,69 @@ asDataFor dec = do
         let con = TH.NormalC cname [(TH.Bang TH.NoSourceUnpackedness TH.NoSourceStrictness, TH.ConT ''BuiltinData)]
         in TH.NewtypeD [] name tTypeVars Nothing con derivs
 
+  -- Generate field accessor functions, if applicable. This is more efficient than the
+  -- field accessors obtained from record pattern synonyms (`RecordPatSyn`).
```
michaelpj, line 94: Why? I had to work this out myself, please write it down.

michaelpj:
Inlining the matcher functions didn't help much, since the main thing here is that you're only accessing the field that is asked for.

It's annoying since the only thing preventing the existing code from doing that is that all the bindings are strict so we can't eliminate them (which is correct from GHC's point of view).

michaelpj:
I have been attempting to try out making some of the pattern matching in `AsData` lazier, which I think is correct, but I'm hitting the stupid reboxing-of-builtin wrappers issue :(

michaelpj:
> That is far less of a problem. The criticisms of Plutus Tx are (as far as I know) script costs, script sizes, and difficulty to set up and get going, but not that it doesn't support a big enough subset of Haskell.

Well, it creates a particular problem with `AsData`. At the moment, it's a drop-in replacement for a datatype declaration. This change would make that not true. In particular the Marlowe validator using `AsData` will not compile with these changes since it uses a bunch of the record syntax features that work with normal record declarations and pattern synonyms, but won't work with this proposal.

I do think that this means it would be significantly better if we could make the existing stuff work better.

But that's going to be difficult. The thing I was trying won't work, for the reason we already discussed: pattern synonyms use a matching function to implement everything, which takes a continuation that gets all the fields. Thus the function application alone will force all the fields (and our inliner can't help since it preserves the semantics). So it's a bit like `&&`, where we would want to inline it in the GHC inliner or something.

michaelpj:
To elaborate, here's a typical workflow with using `AsData`:
1. Convert a datatype to `AsData`
2. Re-run the tests, see if things got better or not

With this change I would have to have another step in between which was "rewrite any code that uses record syntax", which would be sad.

zliu41:
@michaelpj To me the key observation is that record patterns like `Foo { y = example }` is (after CSE) quite efficient - more efficient than the improved field accessors.

Somehow such record patterns never crossed my mind, because I never use them in my code.

Given that, I think we can keep the inefficient field accessors as they are, and simply tell people to always use record patterns instead. How does that sound?

michaelpj:
Yes, if you use record patterns (or named field puns) instead of accessor functions then it should indeed be more efficient.

PR #5653, opened 2023-11-24
PLT-1567 Extend agda executable to handle costing reporting features

mjaskelioff:
Adds options `--counting` and `--tallying` for Untyped PLutus Core (--mode U).

### DIFF
--- a/plutus-metatheory/src/Cost.lagda.md
+++ b/plutus-metatheory/src/Cost.lagda.md
@@ -258,7 +258,7 @@ tallyingReport (mp , budget) =
     ++ printBuiltinReport mp 
     ++ "\n" 
     ++ "Total builtin costs:   " ++ budgetToString totalBuiltinCosts ++ "\n"
-     -- We would like to be ble to print the following  number as "%4.2f" 
+     -- We would like to be able to print the following  number as "%4.2f" 
      -- but Agda's printf currently doesn't support it.
     ++ printf "Time spent executing builtins:  %f%%\n" (fromℕ 100 * (getCPU totalBuiltinCosts) ÷ (getCPU budget)) ++ "\n"
     ++ "\n"
--- a/plutus-metatheory/src/Cost.lagda.md
+++ b/plutus-metatheory/src/Cost.lagda.md
@@ -18,7 +18,7 @@ open import Data.Bool
 open import Data.List using (List;foldr)
 open import Data.Nat using (ℕ;_+_)
 open import Data.Integer using (ℤ)
-open import Data.Float using (Float;fromℕ;_÷_;_*_;_≤ᵇ_)
+open import Data.Float using (Float;fromℕ;_÷_;_*_;_≤ᵇ_) renaming (show to show𝔽)
 open import Data.Nat.Properties using (+-assoc;+-identityʳ)
 open import Data.Nat.Show using () renaming (show to showℕ)
 open import Data.Maybe using (Maybe;just;nothing;fromMaybe;maybe)
@@ -26,6 +26,7 @@ open import Data.Product using (_×_;_,_)
 open import Data.String using (String;_++_;tail;padLeft;padRight)
 open import Algebra using (IsMonoid)
 open import Relation.Binary.PropositionalEquality using (_≡_;refl;trans;isEquivalence;cong₂)
+open import Text.Printf using (printf)
 
 open import Utils.Reflection using (defDec;defShow;defListConstructors)
 
@@ -110,6 +111,10 @@ stepKindList : List StepKind
 unquoteDef stepKindList = defListConstructors (quote StepKind) stepKindList 
 ``` 
 
+## Recording expenditure
+
+The following data structure is use to defines the categories for which we
+record expenditure.
 
 ```
 data ExBudgetCategory : Set where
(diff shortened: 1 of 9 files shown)


plutus-metatheory/src/Main.lagda
```diff
@@ -410,16 +432,16 @@ postulate execP : IO Command
...
+evalInput m _ fmt inp  = fmap (evalProgramN m) (parse fmt inp)
 
 tcInput : Format → Input → IO (Either ERROR String)
 tcInput fmt inp = fmap typeCheckProgramN (parse fmt inp)
 
 main' : Command → IO ⊤
-main' (Eval (EvalOpts inp fmt m)) = do
-  inj₂ s ← evalInput m fmt inp
+main' (Eval (EvalOpts inp fmt m bm)) = do
+  inj₂ s ← evalInput m bm fmt inp
```
kwxm, line 444: I think that a couple of lines below here it won't report the tally/costs in the case of failure.  That's just been changed for `uplc` in #5652.  It's not a big deal though.

plutus-metatheory/src/Main.lagda
```diff
@@ -280,34 +288,48 @@ just t' ← withE runtimeError $ U.stepper maxsteps (ε ; [] ▻ erase t)
...
 
-evalProgramNU : ProgramNU → Either ERROR String
-evalProgramNU namedprog = do
+showUPLCResult : U.State → Either ERROR String
+showUPLCResult (□ V) = return $ prettyPrintUTm (U.extricateU0 (U.discharge V))
+showUPLCResult ◆     = inj₁ (runtimeError userError)
+showUPLCResult _     = inj₁ (runtimeError gasError)
+
+executeUPLC : BudgetMode → ⊥ U.⊢ → Either ERROR String
+executeUPLC Silent t = (withE runtimeError $ U.stepper maxsteps (ε ; [] ▻ t)) >>= showUPLCResult
```
kwxm, line 315: You can't actually turn this on at the moment, right?  I think it's maybe doing the opposite to what `uplc` does: there it shows you the budget/tally without printing out the evaluation result.  I added that because sometimes I just wanted to see the cost or builtin usage of a program but it was also printing some massive term

plutus-metatheory/src/Cost.lagda.md
```diff
@@ -104,13 +163,146 @@ defaultCekMachineCost s = mkExBudget 23000 100
...
+    ++ printStepReport mp
+    ++ "\n"
+    ++ "startup    " ++ budgetToString (lookup mp BStartup) ++ "\n"
+    ++ "compute    " ++ budgetToString totalComputeCost ++ "\n"
+    -- ++ "AST nodes  " ++ ++ "\n"
+    ++ "\n\n"
+    ++ printBuiltinReport mp 
+    ++ "\n" 
+    ++ "Total builtin costs:   " ++ budgetToString totalBuiltinCosts ++ "\n"
+     -- We would like to be ble to print the following  number as "%4.2f"
```
kwxm, line 261:
```suggestion
     -- We would like to be able to print the following  number as "%4.2f" 
```

plutus-metatheory/src/Cost.lagda.md
```diff
@@ -104,13 +163,146 @@ defaultCekMachineCost s = mkExBudget 23000 100
...
+countingReport (mkExBudget cpu mem) = 
+      "\nCPU budget:    " ++ showℕ cpu 
+   ++ "\nMemory budget: " ++ showℕ mem
 ```
+ 
+ ## Tallying budget 
+
+We need a map from `ExBudgetCategory` into `ExBudget`. 
+It's not the most efficient, but the simplest thing to do is to 
+transform `ExBudgetCategory` into a string, ans use that as keys.
```
kwxm, line 190:
```suggestion
transform `ExBudgetCategory` into a string, and use that as keys.
```

plutus-metatheory/src/Untyped/CEKWithCost.lagda.md
```diff
@@ -36,7 +42,15 @@ CekM : Set → Set
 CekM = Writer Budget
 
 spend : StepKind → CekM ⊤
-spend st = tell (cekMachineCost st)
+spend st = tell (cekMachineCost (BStep st))
```
kwxm, line 45: This isn't attempting to deal with slippage, but that's OK because the only case where it'll make a difference is when you fail due to exceeding your budget, is that correct?  It might be worth adding a comment about that.

kwxm (approved):
This all seems good, as far as I can tell. The Agda's not too complicated so it's more understandable than usual in fact the CEK implementation's more understandable than the Haskell version because it doesn't have all of the obscure optimisations!  I also tried this on some large and complicated examples and got the same results as `uplc` (modulo the stuff that's not yet implemented), so I think it's good to go.

plutus-metatheory/src/Main.lagda
mjaskelioff: Actually, `Silent` here is meant to be the default option (not printing any budget, just the result). I should change the name to something else. I was not planning to implement `Silent` as in `uplc`, but maybe I should.

plutus-metatheory/src/Untyped/CEKWithCost.lagda.md
mjaskelioff: `plc-agda` does not implement a `restricting` mode. So there is no need to implement slippage.

plutus-metatheory/src/Main.lagda
mjaskelioff: Now I see where the confusion came from. In the implementation the constructor for "no cost printing" is called `Silent`, as I've done here. The option `--silent` is an *output* option (with constructor `NoOutput`). So I will leave this as it is so that constructors are similar to the implementation.

plutus-metatheory/src/Untyped/CEKWithCost.lagda.md
mjaskelioff: Done

PR #5657, opened 2023-11-29
less strict chooseList and chooseData

mjaskelioff:
Teamwork at it best. 
chooseList and chooseData were too strict in their typechecking.


bezirg (approved):

kwxm (approved):
Yes!

PR #5658, opened 2023-11-29
Extra conformance tests for chooseList

kwxm:
`chooseList` is *-polymorphic in its second and third arguments and we weren't checking this.  We should do this for all *-polymorphic arguments and check that builtins with #-polymorphic arguments fail when given non-constant terms (but later).

The present PR adds tests for the problem that got fixed in #5657.


bezirg (approved):

mjaskelioff (approved):
Thank you Kenneth!

Issue #4751, opened 2022-07-04
Syntax of builtins


effectfully:
### Describe the feature you'd like

Once #4738 gets merged, the classic syntax of builtins will look as follows:

1. built-in `Integer`:

```
(con integer)
```

2. built-in `[Integer]`:

```
(con (list integer))
```

3. `5` of built-in `Integer`:

```
(con integer 5)
```

4. `[1, 2, 3]` of built-in `[Integer]`:

```
(con (list integer) [1, 2, 3])
```

### Describe alternatives you've considered

There have been two proposals to improve the syntax:

1. render `Integer` as `(con (integer))` instead of `(con integer)` and render `5` as `(con (integer) 5)` instead of `(con integer 5)`. This is for consistency with rendering of constants of polymorphic built-in types
2. make a more radical change and render `[Integer]` as `(con list(integer))` and `(Integer, Bool)` as `(con pair(integer,bool))`

Kenneth explains:

> What we're looking for is a language for writing down built-in types.  We want to write down `(con <type> <value>)` and we want `<type>` to be unambiguous.  To keep the parsing simple I think

> (a) we should have one and only one way of writing down each type
> (b) white space should not be part of the syntax.

> I don't like `list integer` because then we need to write `list (list integer)` and then maybe we want to allow `list (integer)` as well and we have two ways of writing down the same type. This seems like trying to re-use Haskell syntax, but that's making things a bit too complicated.

> My proposal (and coincidentally what I use in the specification) is that a type is either an atomic type, which is just a string like `integer` or `string`, or a type application, `op(t₁, ..., tₙ)`, where the `tᵢ` are types. This is completely unambiguous and very easy to parse.

> We could also allow curried type applications like `pair integer string`, but I object to that on the grounds that then you once again need to introduce parentheses if the arguments are complex: `pair (list integer) string`.

> Another disadvantage of this is that it's more difficult to tell where the type ends. You have to work your way along parsing types separated by white space and possibly also involving parentheses, and when you see `t a b c` you have to parse `t` and then notice that there's something else after it, and probably end up with a nested application like `(((t a) b) c)` which you have to disentangle later, checking that `t` is a type operator and that you have the right number of nested applications.

> If everything's fully parenthesised then this doesn't arise: you see `t` , then you see `(` and you know that you've got a list of types coming up, so you parse those and stop when you see the corresponding `)`. You've got the operator and a list of argument types, and it's easy to check that you've got the right number of arguments.

My comments:

[...]

PR #5662, opened 2023-12-01
Kwxm/conformance/more polymorphism tests

kwxm:
This is a follow-up to #5658 , making some conformance test inputs a little more polymorphic.  I also noticed that the inputs for the various `chooseData*` tests used the wrong concrete syntax, so I fixed that (we should remember to fix that in the specification too: it was quite difficult to work out what the syntax is).

To test that some builtin arguments are #-polymorphic but not *-polymorphic, we'd need an input like `(con list ??? [(lam x x)])`, where `???` is some type.  However, all you can put in there are the names of builtin types, so we can't actually write a built-in list that contains a term:  such a thing is in fact rejected by the _parser_ because it knows what type of constant it's expecting inside the list and uses the type-specific parser for that.  If we really want such a test we'd have to try constructing the AST directly, but I don't think the builtin machinery will even let you do that.  So I don't think we can write failing tests of the sort we thought we should have.


mjaskelioff (approved):
LGTM. Thank you for doing these.

mjaskelioff:
BTW, the concrete syntax of builtins was specified (I remember I had to decide on that). 
There is a message on the Plutus mailing list and a comment on [a relevant PR](https://github.com/input-output-hk/plutus/issues/4751#issuecomment-1538377273).

PR #5663, opened 2023-12-03
Kwxm/bls flat uncompression

kwxm:
Currently this is just an experiment and I want so see if it gets through CI.


kwxm:
I think that this should be ready to go once #5664 is merged.  We should rebase this branch on that and then run 
```
cabal test bls12-381-costs-test --test-option --accept
```
to update the costs of the BLS12-381 test cases in `plutus-benchmark`.

kwxm:
We should probably also add a test to make sure that serialising script containing a G1 or G2 constant fails. I'm nor sure how do that right now though.

PR #5673, opened 2023-12-09
Improve beta reduction in callsite inliner for PIR

zliu41:
The `applyAndBetaReduce` function uses `safeToBetaReduce` to determine whether arguments can be beta-reduced. It is an oversimplified implementation and does not detect beta reductions that take multiple steps to perform.

E.g., given `(\x y -> addInteger x y) arg_x arg_y`, where `arg_x` and `arg_y` are effectful terms, `safeToBetaReduce` thinks it is unsafe. This is not true: it can be reduced to `addInteger arg_x arg_y`, but it takes two steps: first reducing `arg_y`, then `arg_x`.

This PR removes `safeToBetaReduce`, and simply makes `applyAndBetaReduce` run the beta and inlining passes for PIR that we already have. To avoid exponential worst case behavior, the inlining passes invoked by `applyAndBetaReduce` only run unconditional inlining. The `callSiteInline` function itself is unchanged.


plutus-tx-plugin/test/Budget/9.6/monadicDo.pir.golden
```diff
@@ -2,24 +2,5 @@ let
...
-  {integer}
-  (Just {integer} 1)
-  (\(x' : integer) ->
-     `$fMonadMaybe_$c>>=`
-       {integer}
-       {integer}
-       (Just {integer} 2)
-       (\(y' : integer) -> Just {integer} (addInteger x' y')))
\ No newline at end of file
+Just {integer} 3
```
michaelpj, line 6: :muscle:

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -407,3 +432,124 @@ maybeAddTySubst tn rhs = do
...
+  -- | The arguments, already processed
+  AppContext tyname name uni fun ann ->
+  InlineM tyname name uni fun ann (Term tyname name uni fun ann)
+applyAndBetaReduce rhs args = do
+  info <- ask
+  -- We run the beta and inlining passes for `len args` times, since sometimes the args
+  -- can only be beta-reduced one by one. An example is
+  -- `(\x y -> f x y) arg_x arg_y`, where `arg_x` and `arg_y` are effectful terms.
+  -- It needs two beta+inlining passes to reduce to `f arg_x arg_y`: the first pass
+  -- inlines `y` and the second pass `x`.
```
michaelpj, line 450: I had to think about this for a bit but yeah, this should work. Can we get a variant of https://github.com/input-output-hk/plutus/pull/5672 that tests this case. Obviously we have a bug at the moment, but in future hopefully not.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -407,3 +432,124 @@ maybeAddTySubst tn rhs = do
...
+  (InliningConstraints tyname name uni fun) =>
+  -- | The rhs of the variable, should have been renamed already
+  Term tyname name uni fun ann ->
+  -- | The arguments, already processed
+  AppContext tyname name uni fun ann ->
+  InlineM tyname name uni fun ann (Term tyname name uni fun ann)
+applyAndBetaReduce rhs args = do
+  info <- ask
+  -- We run the beta and inlining passes for `len args` times, since sometimes the args
+  -- can only be beta-reduced one by one. An example is
```
michaelpj, line 447: I'm somewhat tempted to say that that's a limitation of the inliner that we should try to fix. I'm not sure how big a deal it is if this doesn't work as effectively for effectful args, but only running the inliner once seems attractive.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -407,3 +432,124 @@ maybeAddTySubst tn rhs = do
...
+  -- `(\x y -> f x y) arg_x arg_y`, where `arg_x` and `arg_y` are effectful terms.
+  -- It needs two beta+inlining passes to reduce to `f arg_x arg_y`: the first pass
+  -- inlines `y` and the second pass `x`.
+  --
+  -- Note that the inlining passes we run here only run unconditional inlining.
+  -- Technically it can be helpful to also run callsite inlining, but it rarely
+  -- makes a difference and leads to exponential worst-case behavior.
```
michaelpj, line 454: I'm not convinced that this isn't also exponential... why do we think that?

I notice that what we actually want to do is even simpler than running unconditional inlining: we want to do unconditional inlining, _only considering bindings from the current term_, and then just apply the substitution if we have one. I'm not sure if it's easy to express that, but it might make this less expensive.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -407,3 +432,124 @@ maybeAddTySubst tn rhs = do
...
+            >=> inline'
+              UnconditionalOnly
+              (info ^. iiHints)
+              (info ^. iiBuiltinsInfo)
+              (info ^. iiVarInfo)
+        )
+    )
+    (fillAppContext rhs args)
+  where
+    len :: AppContext tyname name uni fun ann -> Int
```
michaelpj, line 470: `lengthContext` is already defined

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -407,3 +432,124 @@ maybeAddTySubst tn rhs = do
...
+  -- Note that the inlining passes we run here only run unconditional inlining.
+  -- Technically it can be helpful to also run callsite inlining, but it rarely
+  -- makes a difference and leads to exponential worst-case behavior.
+  foldl'
+    (>=>)
+    pure
+    ( replicate
+        (len args)
+        ( pure . Beta.beta
+            >=> inline'
```
michaelpj, line 461: `inline'` does still redundantly compute the term usages. 

I'm not sure why we can't just call `processTerm` here? We're *in* `InlineM`, so we can just do that directly and not define `inline'` at all.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
michaelpj: I'd like to see how much a difference it makes if we just do it once.

michaelpj: Okay, I tried it and things are much worse. However, I think this is not so much to do with effectful terms. It's to do with type arguments. We have long had the problem that you can't beta-reduce all the arguments at once for a term with type and term arguments. You end up with something like this:
```
map = (/\ a b . \(f :: a -> b) (l :: List a) . ...) {integer} {integer} incInteger ls
--->
(let
  a = integer
  b = integer
in \(f :: a -> b) (l :: List a) . ...)) incInteger ls
```
Whereas if we produced this:
```
let
  a = integer
  b = integer
  f :: a -> b = incInteger
  l :: List a = ls
in ...
```
the bindings for `f` and `l` don't typecheck, because their signatures reference the type variables, but their RHSs need the type variables to have been substituted in.

So we end up awkwardly having to do one beta, inline the type variable (which we don't always do!) and then do the rest of the arguments. This plausibly blocks a bunch of optimzation, with this being one example.

It would be good to improve this. It's not as simple as just always inlining type variables, because this is exactly the situation that we have when we compile down datatypes, and we really _don't_ want to inline all of those otherwise we get truly enormous programs and the typechecker grinds to a halt.

However, I think it _would_ be safe to always inline type variables just when doing call-site inlining, because we have the size check, so that won't let us actually grow the program!

michaelpj (commented):
I'm pretty suspicious about the loop of beta reduction and inlining. I think this is probably ultimately the right thing to do rather than relying on our current awkward tricks, though.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
zliu41: In fact, because the RHS is renamed, we'll also need to re-calculate the `VarsInfo` (which I'm not doing currently), and merge it with the original `VarsInfo`. I'll update it.

michaelpj: Ouch. If we're going to do that, you could compute it locally, merge it with the ones in the context using `local`, and then just call `processTerm`. I think that would be neater IMO.

michaelpj: > A further observation

Grr, this is not true because of the inliner state, which is kept in `StateT`, so if we ran `processTerm`, we'd add mappings for stuff we inline to that, and they would _persist_ and potentially be used later where they shouldn't be. So we would need to either rename or only locally modify the state.

michaelpj:
What I am concluding is that this is very subtle :sweat_smile: I wonder if our call-site inlining heuristic is the right one. But I don't know what else to do :thinking:

effectfully:
@zliu41 do we still want to do this? It's been a year.

zliu41:
Yes I do still want this, and I don't remember why I stopped working on it. Thanks for digging it up!

PR #5674, opened 2023-12-10
More efficient `Map.insert` and `Map.unionWith`

zliu41:


plutus-tx/src/PlutusTx/AssocMap.hs
```diff
@@ -222,7 +222,20 @@ union (Map ls) (Map rs) =
...
-unionWith merge ls rs = these id id merge <$> union ls rs
+unionWith merge (Map ls) (Map rs) =
+  let
+    f :: a -> Maybe a -> a
+    f a b' = case b' of
+      Nothing -> a
+      Just b  -> merge a b
+
+    ls' :: [(k, a)]
+    ls' = P.fmap (\(c, i) -> (c, f i (lookup c (Map rs)))) ls
```
michaelpj, line 233: I was going to say, surely this is absolutely atrocious and then I looked up and saw it's the code for `union` :see_no_evil: what a disaster

michaelpj (approved):

PR #5675, opened 2023-12-10
Trial, no "automatic" plugin loading

angerman:
Just don't force load it; this is a nightmare when trying to make cross work.


effectfully:
@zliu41 could you please take a look at this PR? It's been a year.

zliu41:
@angerman could you elaborate what problem you are trying to solve? The only other way to load a plugin is via the `-fPlugin` flag, is that more desirable?

PR #5676, opened 2023-12-11
Split multi-lets in the PIR inliner

zliu41:
This fixes the problem shown in https://github.com/input-output-hk/plutus/pull/5672.


plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
```diff
@@ -196,18 +195,23 @@ processTerm = handleTerm <=< traverseOf termSubtypes applyTypeSubstitution where
...
+                -- carry over into "sibling" terms. This is fine because we have global uniqueness
+                -- (see Note [Inlining and global uniqueness]), if somewhat wasteful.
+                b' <- processSingleBinding t b
+                t' <- processTerm t
+                -- Use 'mkLet': which takes a possibly empty list of bindings (rather than
+                -- a non-empty list)
+                pure $ mkLet ann NonRec (maybeToList b') t'
+            -- If it is a multi-let, split it and process the bindings one by one, because
+            -- the bindings after `b` need to be considered to determine whether `b` can
+            -- be unconditionally inlined.
```
michaelpj, line 213:
```suggestion
            -- If it is a multi-let, split it and process the bindings one by one, because
            -- the bindings after `b` need to be considered to determine whether `b` can
            -- be unconditionally inlined. Consider e.g. 
            -- `let !x = <effectful>; !y = <effectful> in x`, we need to look at the binding for 
            -- `y` to know that inlining `x` would change the order of effects.
            -- It's awkward to do this when they are part of the same term, (we would be 
            -- looking at "the let term but only considering all the bindings after x") and 
            -- much easier when they are just separate terms.
```

plutus-core/plutus-ir/test/PlutusIR/Transform/Inline/nameCapture.golden
```diff
@@ -1,5 +1,7 @@
 let
   !a_4 : integer = 10
+in
+let
```
michaelpj, line 4: We could consider running `LetMerge` after the inliner now, since it splits up all non-recursive lets and doesn't put them back together again.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Inline.hs
michaelpj: Could be a note

plutus-benchmark/plutus-benchmark.cabal
```diff
@@ -47,7 +47,7 @@ common lang
     -fno-specialise -fno-spec-constr -fno-strictness
     -fno-ignore-interface-pragmas -fno-omit-interface-pragmas
     -fno-unbox-strict-fields -fno-unbox-small-strict-fields
-    -fno-full-laziness
+    -fno-full-laziness -fforce-recomp
```
michaelpj, line 50:
```suggestion
    -fno-full-laziness
```

plutus-core/plutus-ir/test/PlutusIR/Transform/Inline/nameCapture.golden
michaelpj: Not a big deal, since we run `LetMerge` before the readable step, which is where it matters most, since everything else should handle split-up lets just the same. Might mean we're less likely to spot bugs relating to non-split lets :joy: 

I can do this nicely in the pass framework.

michaelpj (changes requested):
I agree this is the best thing to do, the alternative is too awkward. Few changes though.

plutus-core/plutus-ir/test/PlutusIR/Transform/Inline/nameCapture.golden
zliu41: Exactly, the single-lets will be merged later.

michaelpj:
Still conflicts.

PR #5677, opened 2023-12-12
More efficient Value.geq and leq

zliu41:


plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -316,40 +316,22 @@ flattenValue v = goOuter [] (Map.toList $ getValue v)
...
-    in checkPred unThese l r
-
 {-# INLINABLE geq #-}
 -- | Check whether one 'Value' is greater than or equal to another. See 'Value' for an explanation
 -- of how operations on 'Value's work.
 geq :: Value -> Value -> Bool
--- If both are zero then checkBinRel will be vacuously true, but this is fine.
-geq = checkBinRel (>=)
+geq l (Value r) =
+  all
```
michaelpj, line 324: Maybe define `valueEntries :: Value -> [(CurrencyId, TokenId, Integer)]` function? Useful here and probably elsewhere.

zliu41: It already exists: `flattenValue`, but flattening both values before checking `geq` is more expensive than this.

michaelpj: Aren't you pretty much exactly flattening the second value? I guess you don't construct an intermediary list this way. Perhaps instead we want `allValueEntries :: Value -> (CurrencyId, TokenId, Integer) -> Bool`. Or I guess the direct way is fine too.

michaelpj: At least worth a comment explaining why you didn't use the existing function so I don't just refactor it back again later ;)

zliu41: > Aren't you pretty much exactly flattening the second value?

There's no need to flatten the _entire_ second value. We can short circuit.

zliu41: I guess this is the limitation with strict languages - it's much harder to fuse transformations without materializing intermediate results.

michaelpj: > There's no need to flatten the entire second value. We can short circuit.

Good point. Comment!

PR #5678, opened 2023-12-12
Use Lovelace instead of Value in some TxInfo fields

zliu41:
This should simplify scripts using these fields, and reduce script costs and sizes even if a script doesn't use them.


plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -337,24 +337,44 @@ instance PlutusTx.Eq ScriptPurpose where
     a PlutusTx.== a'
   _ == _ = Haskell.False
 
+newtype Lovelace = Lovelace Haskell.Integer
```
michaelpj, line 340: IMO we should put this in `Value.hs` and use it where we use `Integer` there. There's no representation change since it's a newtype.

michaelpj (commented):
I have mixed feelings about this as discussed on Slack. I still somewhat feel like we should use `Value` everywhere that there's value moving around, but this is also reasonable.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
lehins: Placing `Lovelace` next to `Value` type certainly makes sense. However, I don't think there is currently a single function in the `Value.hs` module that could use `Lovelace` as a type. It definitely cannot be used instead of `Integer` in the `Value` type itself, since "Lovelace" is a name for 1/10^6 of ADA, not any other token.

That being said a few helper functions could be useful. Something like:

```haskell
promoteLovelace :: Lovelace -> Value

extractLovelace :: Value -> Maybe Lovelace
```
There is some mention of `'Ledger.Ada.adaValueOf'` in the haddock, but I could not find what exactly it refers to.

lehins (approved):
Changes to TxInfo look awesome!!!

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
michaelpj: Yes, I'm being silly, you can't use it in many places in `Value` after all. I did think we had some such helper functions, but perhaps we now don't :thinking:

zliu41: > That being said a few helper functions could be useful.

Does `valueOf v adaSymbol adaToken` return Ada or Lovelace? I assume it's Lovelace but can either of you confirm?

michaelpj: Yes, it's lovelace. It always returns the underlying quantity used, which is lovelace. Ada as a unit is "virtual", the quantity tracked in the ledger is always lovelace.

zliu41: Will do separately

PR #5682, opened 2023-12-13
Add functions for converting between Lovelace and Value

zliu41:


michaelpj (approved):

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -262,6 +261,15 @@ symbols (Value mp) = Map.keys mp
 singleton :: CurrencySymbol -> TokenName -> Integer -> Value
 singleton c tn i = Value (Map.singleton c (Map.singleton tn i))
 
+lovelaceValue :: Lovelace -> Value
```
james-iohk, line 264: why does `lovelaceValue` not need to be inlinable?

zliu41: Good catch. I overlooked that.

PR #5683, opened 2023-12-14
Add some benchmarks for a list matching task

michaelpj:
This mirrors what some DEXs do, and is very expensive since you do quadratic work. It would be much faster with an array primitive.

I'm doing both kinds of list just for completeness. Compiled lists are multiple times faster, but here the asyptotics dominate.


effectfully:
> I'm doing both kinds of list just for completeness.

Thanks a lot for that, I think it can help my pattern matching builtins endeavor.

PR #5691, opened 2023-12-20
Add constitution script hash to `ParameterChange` and `TreasuryWithdrawals`

zliu41:


WhatisRT (approved):
LGTM

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -261,10 +261,15 @@ newtype ChangedParameters = ChangedParameters {getChangedParameters :: PlutusTx.
     )
 
 data GovernanceAction
-  = ParameterChange (Haskell.Maybe GovernanceActionId) ChangedParameters
+  = ParameterChange
+      (Haskell.Maybe GovernanceActionId)
+      ChangedParameters
+      V2.ScriptHash -- ^ Hash of the constitution script
```
lehins, line 267: Governance policy is optional, in a sense it might not exist. In which case this field will be Nothing
```suggestion
      (Maybe V2.ScriptHash) -- ^ Hash of the constitution script
```

lehins (commented):
Script hash should be optional, other than that it looks good

PR #5692, opened 2023-12-21
Specify concrete syntax for built-in lists, pairs, and data (PLT-378)

kwxm:
This updates the specification to include a description of the syntax for the built-in `data` type and built-in list and pairs (see [this comment](https://github.com/IntersectMBO/plutus/issues/4751#issuecomment-1538377273)).  The syntax accepted by the `uplc` parser (and others) differs slightly and we'll need to fix that: see PLT-9074. 

~The page numbers in the index of notation at the end are all off by one, I think because the title page and the abstract page don't have page numbers. I'm trying to work out how to fix this.~

Rendered PDF: [plutus-core-specification.pdf](https://github.com/IntersectMBO/plutus/files/13743487/plutus-core-specification.pdf).


doc/plutus-core-spec/cardano/cardano.tex
```diff
@@ -31,7 +36,48 @@ \section{Built-in types and functions}
...
+                & & \texttt{(}\op \ \bitn_1 \ldots \bitn_{\valency{\op}}) & \textrm{Type application}\\
+    \end{array}\]
+    \label{fig:built-in-type-concrete-syntax}
+\end{minipage}
+\noindent Note that we again require that all type operators are fully applied.
+We refer to the syntactic objects $\bitn$ above as \textit{concrete built-in
+ types}. There is an obvious bijection between these and the abstract built-in
+types used elsewhere in this document, and given an abstract built-in type $T$
+we will denote the corresponding concrete built-in type by $\bar{T}$.
+
```
kwxm, line 65: ^ The fact that the concrete syntax for polymorphic types differs from the notation that we use everywhere else in the specification is a bit annoying.  Maybe we should use the curried notation everywhere (although that suggests that partial application might be possible, which I sincerely hope we'll never allow).  Changing that would have required a lot of rewriting and would have been easy to get wrong, so for the time being I've just added a fudge here.

mjaskelioff: I think this is totally fine. One thing is the concrete syntax, and another the abstract one. We don't need to conflate them.

doc/plutus-core-spec/cardano/builtins1.tex
```diff
@@ -144,6 +138,37 @@ \subsubsection{Built-in types and type operators}
...
+    \end{array}\]
+    \label{fig:data-concrete-syntax}
+\end{minipage}
+
+\noindent We interpret these syntactic constants as elements of $\denote{\ty{data}}$ using
+the various $\inj$ functions defined earlier.  Some valid \texttt{data}
+constants are
+
+\begin{verbatim}
+   (con data (Constr 1 [I 2, B #, Map []])
```
mjaskelioff, line 163: We need to fix the parser, but in the specification I'd put the one that we are goin to choose. That is:
`(con data (Constr 1 [(I 2),(B #), (Map [])])`.

doc/plutus-core-spec/cardano/builtins1.tex
```diff
@@ -144,6 +138,37 @@ \subsubsection{Built-in types and type operators}
...
+    \label{fig:data-concrete-syntax}
+\end{minipage}
+
+\noindent We interpret these syntactic constants as elements of $\denote{\ty{data}}$ using
+the various $\inj$ functions defined earlier.  Some valid \texttt{data}
+constants are
+
+\begin{verbatim}
+   (con data (Constr 1 [I 2, B #, Map []]) 
+   (con data (Map [(I 0, B #00), (I 1, B #0F)]))
```
mjaskelioff, line 164: Same here.

doc/plutus-core-spec/cardano/builtins1.tex
```diff
@@ -144,6 +138,37 @@ \subsubsection{Built-in types and type operators}
...
+\end{minipage}
+
+\noindent We interpret these syntactic constants as elements of $\denote{\ty{data}}$ using
+the various $\inj$ functions defined earlier.  Some valid \texttt{data}
+constants are
+
+\begin{verbatim}
+   (con data (Constr 1 [I 2, B #, Map []]) 
+   (con data (Map [(I 0, B #00), (I 1, B #0F)])) 
+   (con data (List [I 0, I 1, B #7FFF, List []]))
```
mjaskelioff, line 165: And here

doc/plutus-core-spec/cardano/cardano.tex
```diff
@@ -31,7 +36,48 @@ \section{Built-in types and functions}
...
+\paragraph{Concrete syntax for built-in constants.}
+We provide concrete syntax for constants of most (but not all) built-in types.
+For a built-in type $T$ which has a concrete syntax we specify a set $\bitc_T$
+of strings (using either regular expressions or a BNF-style grammar), and a
+constant of type $T$ is then represented in the concrete syntax by an expression
+of the form \texttt{(con $\bar{T}$ $c_T$)} with $c_T \in \bitc_T$.  Each string
+$c_T$ will have an interpretation as a value of type $T$ (ie, an element of
+$\denote{T}$) and since this will generally be the obvious interpreration we
```
mjaskelioff, line 73:
```suggestion
$\denote{T}$) and since this will generally be the obvious interpretation we
```

mjaskelioff (approved):
LGTM. Thank you for doing this!

doc/plutus-core-spec/cardano/builtins1.tex
kwxm: Oh thanks:  I forgot about that.

doc/plutus-core-spec/cardano/cardano.tex
zliu41: Why does the abstract syntax need to use the uncurried form?

zliu41 (approved):
Looks quite reasonable, and much better than the status quo.

doc/plutus-core-spec/cardano/builtins1.tex
```diff
@@ -57,31 +57,25 @@ \subsubsection{Built-in types and type operators}
...
+are thus
+
 \begin{verbatim}
-   (con [(con pair) (con bool) (con string)] 
-        ((con bool True), (con string "Plutus"))
-   ).
+   (con (list integer) [11, 22, 33])
+   (con (pair bool string) (True, "Plutus")).
+   (con (list (pair bool (list bytestring)))
+      [(True, []), (False, [#,#1F]), (True, [#123456, #AB, #ef2804])])
```
zliu41, line 71: Can a bytestring be `#`? I thought so, but according to Section 4.3.1.1, the bytestring syntax is `#([0-9A-Fa-f][0-9A-Fa-f])*`, which must be non-empty.

kwxm: `*` means zero or more repetitions, so that should be OK.

zliu41: OK, I misread it.

kwxm:
This finally passes CI (it only modifies some LaTeX, so I don't know what the problem was), so I'll merge it.

PR #5697, opened 2024-01-03
plutus-tx: add red-black tree implementation for Maps, again

michaelpj:
It still sucks. It manages to sometimes be better, which is cool, but overall it's quite a lot worse and it uses a lot of space. So still not worth it unless we somehow figure out a way to make it much faster.

Amusingly, with this version the dumb equality function is almost as good as @effectfully 's clever one. Also `toList` produces an ordered list, so maybe we could optimize equality further, IDK.

I spent a bit of time on it to get it this fast, mostly fusing things together so we don't do multiple passes. `biFoldr` is the workhorse.

Some things I have not tried:
- Pass around `k -> k -> Ordering` instead of `Ord k`, since apparently our `Ord` class has a ton of members
- Manually specialise to `ByteString` keys
- Come up with a fast way to iterate through the entries of two maps in order without either materializing them into lists or doing lots of lookups (which is what the current `biFoldr` does effectively by calling `split`)
- Actually profiling to see what's happening instead of just looking at it.


effectfully:
Let's close it. If we ever want to revisit it, reopening is always an option.

PR #5700, opened 2024-01-03
PLT-1568 Cost for all builtins

mjaskelioff:
Add support for costing all builtins, both in `counting` and `tallying` mode.

When run it will get the builtin costing parameters from the `builtinCostModel.json` file.  Since the file has to be under the `plutus-metatheory` folder for cabal to find it, the file is currently copied from the one in `plutus-core/cost-model/data/builtinCostModel.json`.

It has only been tested manually that cost reports by `plc-agda` coincide with the ones reported by `uplc`. There is another [ticket](https://input-output.atlassian.net/browse/PLT-1569) for adding automated tests.


plutus-metatheory/src/Cost/BuiltinModelAssignment.agda
```diff
@@ -0,0 +1,230 @@
...
+
+module Cost.BuiltinModelAssignment where
+
+open import Data.Fin using (suc;zero)
+open import Builtin using (Builtin;arity)
+open Builtin.Builtin
+open import Cost.Model using (BuiltinModel;CostingModel;costingCPU;costingMem)
+open CostingModel
+
+assignModel : (b : Builtin) → BuiltinModel (arity b)
```
michaelpj, line 11: do we need to know the definition of this at compile time? could we just take a function of this type as a parameter, and then make it at runtime by reading the file?

mjaskelioff: I don't think there is a way to dynamically load code at runtime. You would need to read a file with a representation of the data, parse it into a data structure, and then use that data structure to define the function. Which is what I'm trying to avoid because not only it would be a very long data structure, but also it would be another place to touch when you add a new builtin function.

michaelpj: Right, I'm suggesting exactly that. I'm not sure why that would be difficult? Wont the data structure be just a dependent map? Which isn't that hard to make? Or you can just return a dependent function? Fundamentally I don't quite see the problem, we just have to write a `Map BuiltinName CostModelDefinition -> Maybe ((b : Builtin) -> BuiltinModel (arity b))` function, which seems doable?

mjaskelioff: I wanted to do it statically, so that if there is any problem with the json file you find about it at compile time.
Maybe not that important, but it's also more efficient.

mjaskelioff: The file is now loaded dynamically at runtime.

kwxm:
@mjaskelioff I've made a ticket at PLT-9238 to see if we can add a function to the `plutus-core` library to expose the cost model to other packages.

kwxm (approved):
This all looks good.  I ran it on some examples and got the same results as `uplc`, so I'm pretty convinced that it must be doing the right thing.  In tallying mode, both `plc-agda` and `uplc` print out the costs for the builtins in apparently random order, but _different_ random orders.  Maybe we should make them both do it in alphabetical order.

plutus-metatheory/src/Cost.lagda.md
```diff
@@ -126,27 +127,37 @@ For each constant we return the corresponding size.
 byteStringSize : Utils.ByteString → CostingNat
```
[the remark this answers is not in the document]
mjaskelioff: We could try. But this is such a critical part of the costing that I think it is a good idea to keep them separate, as a safeguard. Any accidental change to one should be caught by the other in testing.
Also, I think the formalization is more readable and can serve as specification (we still don't have a complete story for sizes in the pdf specification).

mjaskelioff:
> In tallying mode, both `plc-agda` and `uplc` print out the costs for the builtins in apparently random order, but _different_ random orders. Maybe we should make them both do it in alphabetical order.

I think for `uplc` the order is the order they appear in the program. For `plc-agda` is the order of the constructors in the `Builtin` datatype.

plutus-metatheory/src/Cost/Model.lagda.md
```diff
@@ -0,0 +1,194 @@
...
+... | true = map (b ,_) (convertCpuAndMemoryModel rm)
+``` 
+
+Once we have a list of all the builtins and their corresponding model, 
+we need to turn this into a function.  
+
+However Agda doesn't know 
+that we are providing a model for *every* constructor, so we 
+provide a `dummyModel` to answer in the `[]` case.
+
```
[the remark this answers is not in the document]
mjaskelioff: Fixed

plutus-metatheory/src/Cost/Model.lagda.md
```diff
@@ -0,0 +1,194 @@
...
+## Creation of mapping function 
+
+Creates a function mapping builtins to their corresponding costing models, 
+starting from a `BuiltinCostMap`.
+
+We need to construct a `BuiltinModel (arity b)` for each `b`, and this may fail if
+the model int the map doesn't correspond to the arity. 
+
+```
+getModel : Builtin → BuiltinCostMap → Maybe (Σ Builtin (λ b → (BuiltinModel (arity b))))
```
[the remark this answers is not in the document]
mjaskelioff: It works by trying to obtain a model for every `Builtin` in the meta-theory. So, if in the JSON file there is a cost model for a new function that has not been added to the metatheory it will simply be ignored. As you say, I think this is good, as usually new functions are added to the production implementation first. 

(BTW, at first I've got confused by your comment because the function returns a `Maybe` with a `nothing` constructor)

PR #5701, opened 2024-01-04
Data-backed map for Plutus Tx

zliu41:
To avoid very large PRs I'd like to merge this part first.

This PR contains:
- A Data-backed `Map` type in `PlutusTx.DataMap`, and some tests.
- A new `Value` type backed by the new `Map` type in `PlutusLedgerApi.Data.V1.Value`.
- Updated the Marlowe validator to use the new `Map` type in `State`.

I need to write more tests, but otherwise this is ready for review.


plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Core/V1/Semantics.hs
```diff
@@ -322,19 +322,20 @@ evalObservation env state obs = let
...
+    if Map.null accounts
+        then Nothing
+        else
+            -- SCP-5126: The return value of this function differs from
+            -- Isabelle semantics in that it returns the least-recently
+            -- added account-token combination rather than the first
+            -- lexicographically ordered one. Also, the sequence
+            -- `Map.fromList . tail . Map.toList` preserves the
+            -- invariants of order and non-duplication.
+            let (((accId, token), balance), rest) = Map.unsafeUncons accounts
```
effectfully, line 335: Is the comment still true? At the very least the `Map.fromList . tail . Map.toList point seems obsolete. Someone from the Marlowe team should look at this, I think.

plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Scripts/Semantics.hs
```diff
@@ -265,29 +265,16 @@ mkMarloweValidator
...
            -- [Marlowe-Cardano Specification: "Constraint 18. Final balance."]
            traceIfFalse ("v"  <> tag) (totalBalance accounts == expected)
            -- [Marlowe-Cardano Specification: "Constraint 13. Positive balances".]
-        && traceIfFalse ("b"  <> tag) (all positiveBalance $ AssocMap.toList accounts)
+        && traceIfFalse ("b"  <> tag) (Map.all (> 0) accounts)
            -- [Marlowe-Cardano Specification: "Constraint 19. No duplicates".]
-        && traceIfFalse ("ea" <> tag) (noDuplicates accounts)
-        && traceIfFalse ("ec" <> tag) (noDuplicates choices)
-        && traceIfFalse ("eb" <> tag) (noDuplicates boundValues)
+        && traceIfFalse ("ea" <> tag) (Map.noDuplicateKeys accounts)
```
effectfully, line 275: I'm surprised the `Map` module provides such a function, but OK.

plutus-benchmark/marlowe/test/role-payout/9.6/057ebc80922f16a5f4bf13e985bf586b8cff37a2f6fe0f3ce842178c16981027.eval.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 194025187
-| mem: 663336})
\ No newline at end of file
+({cpu: 195267187
```
effectfully, line 1: Here we go one bazillion golden files. This would be so much easier to review if the diff was accompanied by a graph clearly showing the difference for all the data points at once.

plutus-ledger-api/src/PlutusLedgerApi/Data/V1/Value.hs
```diff
@@ -0,0 +1,134 @@
...
+  Map k a ->
+  Map k a ->
+  Bool
+unordEqWith is0 eqV = goBoth
+  where
+    -- Recurse on the spines of both the lists simultaneously.
+    goBoth kvsL kvsR
+      -- One spine is longer than the other one, but this still can result in a
+      -- succeeding equality.
+      -- check if the non-empty list only contains zero values.
```
effectfully, line 77: That was probably my typo, but please remove that dot and the newline after "equality", it's a single sentence.

plutus-tx-plugin/test/Map/Spec.hs
```diff
@@ -0,0 +1,94 @@
...
+  CompiledCode
+    ( Integer ->
+      ( Maybe PlutusTx.BuiltinByteString
+      , Maybe PlutusTx.BuiltinByteString
+      , Maybe PlutusTx.BuiltinByteString
+      , Maybe PlutusTx.BuiltinByteString
+      , Maybe PlutusTx.BuiltinByteString
+      )
+    )
+map1 =
```
effectfully, line 51: A short description of what we're testing here?

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+  * Other operations are slower than @PlutusTx.AssocMap.Map@, although equality
+    checks on keys can be faster due to `P.equalsData`.
+  * Many operations involve converting the keys and/or values to/from `P.BuiltinData`.
+
+Therefore this map implementation is likely a better choice than @PlutusTx.AssocMap.Map@
+if it is part of a data type defined using @asData@, and the key and value types
+have efficient `P.toBuiltinData` and `P.unsafeFromBuiltinData` operations (e.g., they
+are primitive types or types defined using @asData@).
+-}
+newtype Map k a = Map P.BuiltinData
```
effectfully, line 33: A comment on whether there are any invariants such as orderness? (Stipulating the lack of orderness would also be helpful)

effectfully (commented):
Ran out of time, will continue reviewing tomorrow.

plutus-benchmark/marlowe/test/role-payout/9.6/057ebc80922f16a5f4bf13e985bf586b8cff37a2f6fe0f3ce842178c16981027.eval.golden
kwxm: > Here we go one bazillion golden files. This would be so much easier to review if the diff was accompanied by a graph clearly showing the difference for all the data points at once.

It would help a lot if we even had all the results printed in a table in a single file.  That should be doable for the marlowe files, although I'm not sure if/how you'd be able to do it more generally.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts   #-}
+{-# LANGUAGE LambdaCase         #-}
+{-# LANGUAGE NoImplicitPrelude  #-}
+{-# LANGUAGE TupleSections      #-}
+{-# LANGUAGE ViewPatterns       #-}
+
+module PlutusTx.DataMap where
```
kwxm, line 9: Do you want to say what you're exporting here?

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+{- | An associative map implementation backed by `P.BuiltinData`.
+
+This map implementation has the following characteristics:
+
+  * The `P.toBuiltinData` and `P.unsafeFromBuiltinData` operations are no-op.
+  * Other operations are slower than @PlutusTx.AssocMap.Map@, although equality
+    checks on keys can be faster due to `P.equalsData`.
+  * Many operations involve converting the keys and/or values to/from `P.BuiltinData`.
+
+Therefore this map implementation is likely a better choice than @PlutusTx.AssocMap.Map@
```
kwxm, line 28: Oh, are you going to keep both of them and let the user choose?  If you're going to do that you could maybe have property tests to check that both versions do the same thing.  Also in that case it would be worth mentioning it in `PlutusTx.AssocMap.Map`.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+  BI.BuiltinList (BI.BuiltinPair BuiltinData BuiltinData) ->
+  Maybe BuiltinData
+lookup' k = go
+  where
+    go xs =
+      P.matchList
+        xs
+        Nothing
+        ( \hd tl ->
+            let k' = BI.fst hd
```
kwxm, line 65: The alignment seems a bit odd here (and elsewhere in this file).

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+        False
+        ( \hd tl ->
+            let k' = BI.fst hd
+             in if P.equalsData k k'
+                  then True
+                  else go tl
+        )
+
+{-# INLINEABLE insert #-}
+insert :: forall k a. (P.ToData k, P.ToData a) => k -> a -> Map k a -> Map k a
```
kwxm, line 92: Maybe have some Haddock saying how this behaves if the key already exists in the map.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+insert (P.toBuiltinData -> k) (P.toBuiltinData -> a) m =
+  unsafeFromBuiltinList (go (toBuiltinList m))
+  where
+    go ::
+      BI.BuiltinList (BI.BuiltinPair BuiltinData BuiltinData) ->
+      BI.BuiltinList (BI.BuiltinPair BuiltinData BuiltinData)
+    go xs =
+      P.matchList
+        xs
+        (BI.mkCons (BI.mkPairData k a) nil)
```
kwxm, line 102: I think that if the key's not already in the map then this always inserts the new entry at the end and has to duplicate all of the original list when it does `BI.mkCons hd (go tl)` a few lines below here.  Could we somehow just cons the new entry onto the front in that case? That would presumably be more memory-efficient.

Later: I thought you could avoid duplicating the original input by binding it to a variable at the start and consing onto that if you arrive at the end without finding the key, but you _do_ have to copy the front of the list if the key actually is present and you can't tell that it's safe not to  do that until you get to the end.  Hmmm.

Later still: I suppose you could scan the whole list first to see if your key's already there and do the cheap cons if it isn't, but then you're paying for a possible memory saving with the CPU cost of an extra traversal.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+      BI.BuiltinList (BI.BuiltinPair BuiltinData BuiltinData)
+    go xs =
+      P.matchList
+        xs
+        nil
+        ( \hd tl ->
+            let k' = BI.fst hd
+             in if P.equalsData k k'
+                  then tl
+                  else BI.mkCons hd (go tl)
```
kwxm, line 125: Again, I think this duplicates the entire list if you delete a non-existent entry.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+null = P.null . toBuiltinList
+
+{-# INLINEABLE unsafeFromList #-}
+unsafeFromList :: (P.ToData k, P.ToData a) => [(k, a)] -> Map k a
+unsafeFromList =
+  unsafeFromBuiltinList
+    . toBuiltin
+    . PlutusTx.Prelude.map (\(k, a) -> (P.toBuiltinData k, P.toBuiltinData a))
+
+{-# INLINEABLE uncons #-}
```
kwxm, line 149: Would you ever want to do this?

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+    res :: BI.BuiltinList (BI.BuiltinPair BuiltinData BuiltinData)
+    res = go rs' ls'
+      where
+        go acc xs =
+          P.matchList
+            xs
+            acc
+            (\hd -> go (BI.mkCons hd acc))
+
+-- | Combine two 'Map's with the given combination function.
```
kwxm, line 272: It would be useful to have a comment saying what the semantics are if you've got an overlap.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+        xs
+        False
+        ( \hd tl ->
+            let a = P.unsafeFromBuiltinData (BI.snd hd)
+             in if p a then True else go tl
+        )
+
+{-# INLINEABLE union #-}
+
+-- | Combine two 'Map's.
```
kwxm, line 217: I'm not sure what this is doing, but it's bedtime so I'll submit what I've said so far and come back for another look next week.

kwxm (commented):
I've not looked at the whole thing: I'll come back and finish next week.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+{-# INLINEABLE empty #-}
+empty :: forall k a. Map k a
+empty = unsafeFromBuiltinList nil
+
+{-# INLINEABLE null #-}
+null :: forall k a. Map k a -> Bool
+null = P.null . toBuiltinList
+
+{-# INLINEABLE unsafeFromList #-}
+unsafeFromList :: (P.ToData k, P.ToData a) => [(k, a)] -> Map k a
```
michaelpj, line 143: lots of this stuff needs haddock, explaining what things do and in this case why it's unsafe

plutus-tx/src/PlutusTx/DataMap.hs
michaelpj: Yeah, this is a bit weird, since it depends on the ordering

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+  Map k a ->
+  ((k, a), Map k a)
+unsafeUncons m =
+  ((P.unsafeFromBuiltinData k, P.unsafeFromBuiltinData a), unsafeFromBuiltinList tl)
+  where
+    (hd, tl) = P.unsafeUncons (toBuiltinList m)
+    (k, a) = P.pairToPair hd
+
+{-# INLINEABLE noDuplicateKeys #-}
+noDuplicateKeys :: forall k a. Map k a -> Bool
```
michaelpj, line 174: this is quadratic, worth saying

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+      P.matchList
+        xs
+        True
+        ( \hd tl ->
+            let k = BI.fst hd
+             in if member' k tl then False else go tl
+        )
+
+{-# INLINEABLE all #-}
+all :: forall k a. (P.UnsafeFromData a) => (a -> Bool) -> Map k a -> Bool
```
michaelpj, line 188: why not define a fold and use that for a bunch of these?

michaelpj: generally it seems like you've avoided defining any higher-order functions and using them, which seems like overkill to me, and makes this much harder to review...

plutus-tx/src/PlutusTx/DataMap.hs
michaelpj: I think it's an inlined and de-abstracted version of the one in AssocMap. I'm not sure why it's not a more direct port.

plutus-ledger-api/src/PlutusLedgerApi/Data/V1/Value.hs
```diff
@@ -0,0 +1,134 @@
+{-# LANGUAGE MultiWayIf        #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE ViewPatterns      #-}
+
+module PlutusLedgerApi.Data.V1.Value where
```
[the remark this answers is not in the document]
michaelpj: Yeah, I feel like maybe we should talk about how we're going to do the API. Duplicating it all is just going to suck, but I don't know what alternative we have if the new version isn't a strict improvement on the old one (which it doesn't seem like it is...).

plutus-tx-plugin/test/Map/Spec.hs
michaelpj: Also I think it would be appropriate to do some "model" property-based tests where we test that this map behaves like a Haskell map and/or AssocMap under various series of operations.

plutus-tx/src/PlutusTx/DataMap.hs
michaelpj: I think the invariant is just no-duplicates, and not orderedness. I wonder if having it be ordered would help... then it wouldn't be a drop-in replacement for `AssocMap` since you would need `Ord`, I guess.

michaelpj (commented):
I would like to know why it's not always faster!

plutus-tx/src/PlutusTx/DataMap.hs
kwxm: Isn't the notion of invariant a bit tenuous here?  It would be easy to construct a data object that looks like a map but doesn't satisfy the invariants and stick it in a datum or something. You might even do that by accident if you're using some external tool to produce data.

effectfully: > Isn't the notion of invariant a bit tenuous here?

Either way, I just wanna know if there are any or not.

>  It would be easy to construct a data object that looks like a map but doesn't satisfy the invariants and stick it in a datum or something. You might even do that by accident if you're using some external tool to produce data.

A `Map` isn't necessarily taken verbatim from a datum. It may be produced by the validator itself or come from a datum but be properly sanitized, so I don't think having invariants is out of question. But maybe I misunderstand the intended use cases.

plutus-tx/src/PlutusTx/DataMap.hs
effectfully: > why not define a fold and use that for a bunch of these?

That would make it trickier to make sure that things short-circuit properly, although inlining probably is more or less reliable in that case, although relying on inlining is always dangerous.

effectfully (commented):
As others said, tests and docs are missing, otherwise seems good, although I do share Kenneth's performance concerns.

plutus-tx/src/PlutusTx/DataMap.hs
zliu41: Yes, will add explicit exports.

plutus-tx/src/PlutusTx/DataMap.hs
zliu41: > I think the invariant is just no-duplicates

It doesn't guarantee no-duplicates. This is really an association list, rather than a map, and should probably be named so.

plutus-tx/src/PlutusTx/DataMap.hs
zliu41: It's fourmolu - it would take me forever to hand-format this module!

plutus-tx/src/PlutusTx/DataMap.hs
zliu41: > I suppose you could scan the whole list first to see if your key's already there and do the cheap cons if it isn't, but then you're paying for a possible memory saving with the CPU cost of an extra traversal.

I doubt doing two traversals for insertion is the more efficient implementation in any programming language (do people do that in any language?), and thus isn't the right thing to do in PLC either.

plutus-tx/src/PlutusTx/DataMap.hs
zliu41: The Marlowe validator needs `unsafeUncons`.

This data structure is really an association list rather than a map, and should probably be named so. For an association list, it makes a lot more sense to have an `uncons` operation.

plutus-tx/src/PlutusTx/DataMap.hs
zliu41: Again, if we call this an association list, then this would be a lot more obvious.

plutus-tx/src/PlutusTx/DataMap.hs
zliu41: > why not define a fold

Then `fold` will have to be inlined and beta-reduced, but there's no guarantee, even if it has the INLINE pragma. Recall that the INLINE pragma only makes it inline, not beta-reduce.

The code duplication shouldn't be a big deal due to CSE.

zliu41 (commented):
Responded to some comments.

plutus-tx/src/PlutusTx/DataMap.hs
kwxm: > It's fourmolu - it would take me forever to hand-format this module!

Ahhhh.  I've never tried using fourmolu.  I just let stylish-haskell do its thing, and it does something totally different!

plutus-tx/src/PlutusTx/DataMap.hs
kwxm: I suppose in general you'd use some structure that lets you do efficient insertion, but trying to do something like that in PLC might have too much overhead. The trouble is that we're trying to maintain an invariant of every key only occurring once, and that's maybe a bad fit with lists.  I suppose we could do some experiments to see what the space/time tradeoff is like if we do two traversals. Maybe the space usage isn't too bad though: you do have to copy all of the nodes if the thing you're inserting isn't in the map, but it's a shallow copy where you just need to allocate new cons-cells that contain pointers to the contents without copying the contents themselves.

kwxm: > Maybe the space usage isn't too bad 

... although if you create a map with n entries by calling `insert` n times you'll allocate n(n+1)/2 cons cells in total.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+        False
+        ( \hd tl ->
+            let a = P.unsafeFromBuiltinData (BI.snd hd)
+             in if p a then True else go tl
+        )
+
+{-# INLINEABLE union #-}
+
+-- | Combine two 'Map's.
+union ::
```
kwxm, line 218: I'm still failing to gasp what this is doing.  If I have
```
m1 = Map Map [(I 22,I 222),(I 11,I 111)]
m2 = Map Map [(I 22,I 999),(I 33,I 333)]
```
then `union m1 m2` is
```
Map Map [(I 11,Constr 1 [I 111]),(I 22,Constr 3 [I 222,I 999]),(I 33,I 333)]
```
The keys that are in m1 have their values tagged with `This`, the key's that are in both have both values tagged with `These`, but the stuff that's in m2 only is untagged.  Is there a `That` missing is the definition of rs' or somewhere?

kwxm: I was trying to think what would happen if either or both of the inputs contained duplicate keys, but my mind couldn't handle it.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+        go xs =
+          P.matchList
+            xs
+            nil
+            ( \hd tl ->
+                let k' = BI.fst hd
+                    tl' = go tl
+                 in if member' k' ls
+                      then tl'
+                      else BI.mkCons hd tl'
```
kwxm, line 260: Further to my earlier comment, think this should be something like
```
   else BI.mkCons (BI.mkPairData k' (P.toBuiltinData (That (P.unsafeFromBuiltinData (BI.snd hd)) :: These a b))) tl'
```
which is not exactly what you'd call elegant.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+
+-- | Combine two 'Map's with the given combination function.
+unionWith ::
+  forall k a.
+  (P.UnsafeFromData a, P.ToData a) =>
+  (a -> a -> a) ->
+  Map k a ->
+  Map k a ->
+  Map k a
+unionWith f (toBuiltinList -> ls) (toBuiltinList -> rs) =
```
kwxm, line 280: Ah, this is the thing that lets users resolve clashes themselves, but it's less general than `union` because the values in both maps have to be of the same type.  I suppose one could generalise this by making it something like 
```
  forall k a.
  <suitable constraints> =>
  (a -> c) ->
  (b -> c) ->
  (a -> b -> c) ->
  Map k a ->
  Map k b ->
  Map k c
  ```
and then it would subsume `union`.

File: plutus-tx/src/PlutusTx/DataMap.hs [hunk 23]
```diff
@@ -0,0 +1,355 @@
...
+            ( \hd tl ->
+                let k' = BI.fst hd
+                    tl' = go tl
+                 in if member' k' ls
+                      then tl'
+                      else BI.mkCons hd tl'
+            )
+
+    res :: BI.BuiltinList (BI.BuiltinPair BuiltinData BuiltinData)
+    res = go rs' ls'
```
kwxm, line 264: Can you make this act on `rs` and cons the missing stuff (ie the stuff not in the overlap) directly onto `ls'`? I think that at the moment it's using heap memory to construct `rs'` as an intermediate list which is then appended to `ls'` and discarded.  If you can avoid constructing `rs'` then it would be more efficient in terms of both CPU and memory.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,355 @@
...
+                    v'' = case lookup' k' rs of
+                      Just r ->
+                        P.toBuiltinData
+                          (f (P.unsafeFromBuiltinData v') (P.unsafeFromBuiltinData r))
+                      Nothing -> v'
+                 in BI.mkCons (BI.mkPairData k' v'') (go tl)
+            )
+
+    rs' :: BI.BuiltinList (BI.BuiltinPair BuiltinData BuiltinData)
+    rs' = go rs
```
kwxm, line 302: Again it might be possible to avoid constructing `rs'`.

kwxm (commented):
OK, I've finished now.  I think some of it could be made more efficient, but the problem is that that would make it more complicated and less comprehensible, and hence more error-prone.

ana-pantilie:
Closing as I am splitting this up into multiple PRs.

PR #5704, opened 2024-01-05
[Builtins] Add 'ListToConstr' and 'DataToConstr'

effectfully:
This is a competitor of #5486, let's see how it performs.

This adds the `listToConstr` builtin whose Plutus type is

```
all a. list a -> sop [] [a, (list a)]
```

The idea is that we can use this builtin to implement faster pattern matching over built-in lists by piggy-backing on `case` that we use for pattern matching over sums-of-products:

```
matchList : all a r. list a -> r -> (a -> list a -> r) -> r
  = /\a r ->
      \(xs : list a) (nilCase : r) (consCase : a -> list a -> r) ->
        case r (listToConstr {a} xs) [nilCase, consCase]
```

Implementation-wise this requires us to add the following class:

```haskell
class ToConstr term where
    toConstr
        :: forall rep. KnownTypeAst TyName (UniOf term) rep
        => Word64 -> [term] -> Opaque term rep
```

and add it to `HasMeaningIn uni val`.

The meaning of the `listToConstr` builtin is implemented as follows:

```haskell
    toBuiltinMeaning _semvar ListToConstr =
        let listToConstrDenotation
                :: KnownTypeAst TyName DefaultUni a
                => SomeConstant uni [a]
                -> EvaluationResult (Opaque val (TySopRep '[ '[], '[a, [a]] ]))
            listToConstrDenotation (SomeConstant (Some (ValueOf uniListA xs0))) = do
                DefaultUniList uniA <- pure uniListA
                pure $ case xs0 of
                    []     -> toConstr 0 []
                    x : xs -> toConstr 1 [fromValueOf uniA x, fromValueOf uniListA xs]
            {-# INLINE listToConstrDenotation #-}
        in makeBuiltinMeaning
            listToConstrDenotation
            (\_ _ -> ExBudgetLast mempty)
```

i.e. it simply turns the outer constructor of a list into an SOP term leaving all other constructors of the list intact.

The only interesting remaining part is that we make `matchList` an `Internal` builtin (i.e. a `NOINLINE` one) and tell the plugin to replace that builtin with the definition of `matchList` given above.


effectfully:
We are not doing this as per #5711, hence closing.

PR #5706, opened 2024-01-08
Remove hex-text workaround

michaelpj:


michaelpj:
Ah, it hasn't quite hit hackage.nix yet?

PR #5708, opened 2024-01-09
Remove `GovernanceActionId` from the `Voting` script purpose

zliu41:
It is not needed because the script for a given voter will be run only once for all votes.


WhatisRT (approved):

lehins (approved):
:+1:

PR #5712, opened 2024-01-11
Update the `Certifying` and `Proposing` script purposes

### LINKED ISSUE #5709: One more change is needed for ScriptPurpose

@zliu41 Sorry for not catching this earlier. Instead of an `Integer` here we need to specify `ProposalProcedure`
https://github.com/IntersectMBO/plutus/blob/5e95bc2db868e7592ae39c0ea95b9cc1b0f9337c/plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs#L328

_No response_

zliu41:
Their arguments now consist of both an integer index and the actual argument (`TxCert` and `ProposalProcedure`). Closes #5709


lehins (approved):
Perfect

michaelpj:
Given the long discussion on the other thread, I feel like we should have a Note here about the decision.

WhatisRT (approved):

zliu41:
> I feel like we should have a Note here about the decision.

Good point. Will add the note in another PR.

PR #5715, opened 2024-01-12
Add benchmarks for Agda CEK machine

kwxm:
Don't look here yet.  I'm just checking that I haven't messed anything up ...

PR #5725, opened 2024-01-17
Benchmarks for the Agda CEK machine (PLT-9187)

kwxm:
This adds benchmarks `validation-agda-cek`, `marlowe-agda-cek`, and `nofib-agda-cek` which benchmark the performance of the Agda CEK machine with our standard benchmarks.   Mostly it's just a bit of refactoring to let the basic benchmark runners take an evaluator as a parameter.  It might also be worth adding benchmarks for evaluation using the Agda version of the contextual semantics, but that might require a bit of work in the Agda code, so let's look at that in another ticket.

There was a slight danger that the refactoring might have affected the time required to run the existing benchmarks. I checked this in #5715 and that didn't seem to be the case, although the benchmarking machine appears to be a bit inconsistent.


kwxm:
Here are the results of the benchmarks.
[validation.txt](https://github.com/IntersectMBO/plutus/files/13960074/validation.txt)
[marlowe.txt](https://github.com/IntersectMBO/plutus/files/13960075/marlowe.txt)
[nofib.txt](https://github.com/IntersectMBO/plutus/files/13960076/nofib.txt)

plutus-benchmark/nofib/bench/BenchPlc.hs
kwxm, line 1: Renamed to BenchCek.hs for consistency with other benchmarks.

michaelpj (approved):

PR #5728, opened 2024-01-17
[Builtins] Expose 'BuiltinResult'

effectfully:
This renames `MakeKnownM` to `BuiltinResult`, exposes it to the user and updates a bunch of crypto primitives to use it instead of `Emitter (EvaluationResult Smth)`. The latter was very error-prone: an `EvaluationFailure` is supposed to terminate evaluation, but not when it's wrapper in `Emitter`, meaning code like

```haskell
do
    guard condition
    rest
```

wouldn't terminate on `condition` being `False`, which could easily lead to a security issue if we weren't extra careful with the crypto builtins. This PR fixes the problem.

Plus `BuiltinResult` is designed do be fast, unlike `Emitter (EvaluationResult Smth)`, although the difference for our specific use cases is likely negligible.

I also had to move a bunch of stuff around, 'cause I was getting pretty annoying import cycle errors.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
@@ -16,15 +16,15 @@
 {-# LANGUAGE StrictData             #-}
 
 module PlutusCore.Builtin.KnownType
-    ( KnownTypeError
-    , throwKnownTypeErrorWithCause
+    ( BuiltinError
+    , throwBuiltinErrorWithCause
     , KnownBuiltinTypeIn
     , KnownBuiltinType
-    , MakeKnownM (..)
+    , BuiltinResult (..)
     , ReadKnownM
+    , MakeKnownIn (..)
     , liftReadKnownM
     , readKnownConstant
-    , MakeKnownIn (..)
     , MakeKnown
     , ReadKnownIn (..)
     , ReadKnown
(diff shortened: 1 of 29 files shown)


plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
```diff
@@ -266,89 +264,19 @@ typeMismatchError uniExp uniAct = fromString $ concat
...
--- Moving from @ExceptT KnownTypeError Emitter@ to this data type gave us a speedup of 8% of total
--- evaluation time.
---
--- Logs are represented as a 'DList', because we don't particularly care about the efficiency of
--- logging, since there's no logging on the chain and builtins don't emit much anyway. Otherwise
--- we'd have to use @text-builder@ or @text-builder-linear@ or something of this sort.
-data MakeKnownM a
-    = MakeKnownFailure (DList Text) KnownTypeError
-    | MakeKnownSuccess a
-    | MakeKnownSuccessWithLogs (DList Text) a
```
effectfully, line 286: Moved that stuff to its own file.

plutus-core/plutus-core/src/PlutusCore/Evaluation/ErrorWithCause.hs
effectfully, line 1: Had to introduce this file to break up an import cycle. That stuff used to live in `Machine.Exception`, but it's not in any way specific to abstract machines. so it now has a better home here. I guess technically it's not specific to evaluation either, but we only use for that (unlike with machines, 'cause we also use this stuff for evaluation of builtins).

michaelpj (approved):
looks fine

plutus-tx/src/PlutusTx/Builtins/Internal.hs
```diff
@@ -258,10 +256,10 @@ keccak_256 (BuiltinByteString b) = BuiltinByteString $ Hash.keccak_256 b
...
 verifyEd25519Signature (BuiltinByteString vk) (BuiltinByteString msg) (BuiltinByteString sig) =
   case PlutusCore.Crypto.Ed25519.verifyEd25519Signature_V1 vk msg sig of
-    Emitter f -> case runWriter f of
-      (res, logs) -> traceAll logs $ case res of
-        EvaluationFailure   -> mustBeReplaced "Ed25519 signature verification errored."
-        EvaluationSuccess b -> BuiltinBool b
+    BuiltinFailure logs err       -> traceAll (logs <> pure (display err)) $
+        mustBeReplaced "Ed25519 signature verification errored."
+    BuiltinSuccess b              -> BuiltinBool b
+    BuiltinSuccessWithLogs logs b -> traceAll logs $ BuiltinBool b
```
effectfully, line 262: @michaelpj should I abstract this? It gets repeated a lot. 

(I failed to open a _draft_ PR again, this wasn't fully ready for merging, but it's fine that you merged it)

michaelpj: could do, not sure it's a big deal either way :shrug:

PR #5729, opened 2024-01-18
Add documentation for AsData

michaelpj:
This adds a basic howto guide showing how to use it and discussing what it does.

### DIFF
--- a/doc/read-the-docs-site/howtos/asdata.rst
+++ b/doc/read-the-docs-site/howtos/asdata.rst
@@ -4,65 +4,50 @@
 How to use ``AsData`` to optimize scripts
 =========================================
 
-The latest Plutus libraries contain a new ``PlutusTx.asData`` module that contains Template Haskell (TH) code for encoding algebraic data types (ADTs) as ``Data`` objects in Plutus Core, as opposed to sums-of-products terms.
+The Plutus libraries contain a ``PlutusTx.asData`` module that contains Template Haskell (TH) code for encoding algebraic data types (ADTs) as ``Data`` objects in Plutus Core, as opposed to sums-of-products terms.
 In general, ``asData`` pushes the burden of a computation nearer to where a value is used, in a crude sense making the evaluation less strict and more lazy.
 This is intended for expert Plutus developers.
 
 Purpose
 -------
 
-When writing and optimizing a Plutus script, one of the challenges is finding the right balance for your specific use case between which method to use for handling ``Data`` objects and how expensive that method will be.
+Values stored in datums or redeemers need to be encoded into ``Data`` objects.
+When writing and optimizing a Plutus script, one of the challenges is finding the right approach to handling ``Data`` objects and how expensive that method will be.
 To make an informed decision, you may need to benchmark and profile your smart contract code to measure its actual resource consumption.
 The primary purpose of ``asData`` is to give you more options for how you want to handle ``Data``.
 
-Choice of two pathways
-----------------------
+Choice of two approoaches
+-------------------------
 
 When handling ``Data`` objects, you have a choice of two pathways.
 It is up to you to determine which pathway to use depending on your particular use case.
 There are trade offs in performance and where errors occur.
 
-Method one: proactively do all of the parsing
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Approach one: proactively do all of the parsing
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The first approach is to parse the object immediately (using ``fromBuiltinData``) into a native Plutus Core datatype, which will also identify any problems with the structuring of the object.
 However, this performs all the work up front.
 
-Method two: only do the parsing if and when necessary
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+This is the normal style that has been promoted in the past.
 
-In the second approach, the validator doesn't necessarily do anything. It only does the work when it needs to. In order to work with the data objects, the validator has to call the builtin functions, which is a little bit more expensive. It might be that this saves you a lot of work, because you may never need to parse the entire object. Instead, the validator will just carry the item around as if it were a data object. Some time later, when the validator needs to do an operation with this integer, then the validator will parse it. If it determines that it is not an integer, there will be errors.
-Using this method, every time the validator parses it, it is going to look at the data object to find out if it is an integer. If it is an integer, it will get the integer out and do its processing. The analysis work may be repeated depending on how your script is written. In some cases, you might do less work, in some cases you might do more work, depending on your specific use case.
+Approach two: only do the parsing if and when necessary
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the second approach, the script doesn't do any parsing work immediately, and instead does it later, when it needs to.
+It might be that this saves you a lot of work, because you may never need to parse the entire object.
+Instead, the script will just carry the item around as a ``Data`` object.
+
+Using this method, every time the script uses the object, it will look at it to find out if it has the right shape.
+If it does have the right shape, it will deconstruct the ``Data`` object and do its processing; if not, it will throw an error.
+This work may be repeated depending on how your script is written.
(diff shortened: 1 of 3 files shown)


joseph-fajen:
@michaelpj I'm working on a suggestion or two about how this is organized but I haven't yet worked out the specifics. More soon.

doc/read-the-docs-site/howtos/asdata.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Caveats
+-------
+
+The most important caveat to using ``asData`` is that ``Data`` objects encoding datatypes must also encode the *fields* ``Data``.
+However, when using the pattern synonyms they try to give you exactly what you asked for, which might mean having the fields *not* encoded as ``Data``.
+
+For example, in the ``Bid`` case above the ``Lovelace`` field is represented as a Plutus Core builtin integer.
+However, in order for it to be encoded into the ``Bid`` structure, we need to encode it into ``Data``.
+That means that when you construct a ``Bid`` object you do an ``Integer`` to ``Data`` conversion, and when you pattern match on a ``Bid`` object you do a ``Data`` to ``Integer`` conversion.
```
zliu41, line 78: I don't think `Integer` is a very good example, since there's almost certainly no need to bother with a Data-encoded Integer type, vs. just the plain Integer.

doc/read-the-docs-site/howtos/asdata.rst
```diff
@@ -0,0 +1,89 @@
...
+That is:
+
+- It creates a newtype wrapper around ``BuiltinData``
+- It creates pattern synonyms corresponding to each of the constructors you wrote
+
+This lets you write code "as if" you were using the original declaration that you wrote, while in fact the pattern synonyms are handling conversion to/from ``Data`` for you.
+But any values of this type actually are represented with ``Data``.
+That means that when we newtype-derive the instances for converting to and from ``Data`` we get the instances for ``BuiltinData`` - which are free!
+
+Caveats
```
zliu41, line 70: There's another important caveat: it is advisable to access fields with pattern matching, since it is probably faster than field accessors. Record patterns can be used if there are too many fields.

doc/read-the-docs-site/howtos/asdata.rst
```diff
@@ -0,0 +1,89 @@
...
+-------
+
+The most important caveat to using ``asData`` is that ``Data`` objects encoding datatypes must also encode the *fields* ``Data``.
```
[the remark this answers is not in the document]
michaelpj: I don't know what you mean. It absolutely is the case that the fields of data objects must be data. `Constr` takes a list of data objects as fields, there is no way to put anything else in there.

doc/read-the-docs-site/howtos/asdata.rst
michaelpj: I'm confused. The point is that you're data-encoding `Bid`, and that _forces_ you to data-encode its `Integer` fields (which is cheap, but not free)

doc/read-the-docs-site/howtos/asdata.rst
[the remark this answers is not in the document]
zliu41: I took this sentence to mean: if you use `AsData` on `data Foo = Foo { bar :: Bar, ... }`, then you must also use `AsData` on `Bar`. But you don't have to: `Bar` can just be a regular Haskell datatype.

But now I don't think this is what you mean.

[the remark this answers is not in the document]
michaelpj: I said that the _data object_ must encode its fields as Data. So if you use `AsData` on Foo, then `bar` must be encoded using Data. It's just that that's going to be _expensive_ if `Bar` is a normal datatype and cheap if it's using `AsData`.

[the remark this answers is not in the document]
michaelpj: Anyway, I clearly need to make this clearer!

doc/read-the-docs-site/howtos/asdata.rst
```diff
@@ -0,0 +1,89 @@
+.. highlight:: haskell
+.. _asdata:
+
+How to use ``AsData`` to optimize scripts
+=========================================
+
```
joseph-fajen, line 6:
```suggestion
The latest Plutus libraries contain a new ``PlutusTx.asData`` module that contains Template Haskell (TH) code for creating algebraic data types (ADTs) as Data objects in Plutus Core, as opposed to sums-of-products terms. In general, ``asData`` pushes the burden of a computation nearer to where a value is used, in a crude sense making the evaluation less strict and more lazy. This is intended for expert Plutus developers. 

Purpose
-------

When writing and optimizing a Plutus script, one of the challenges is finding the right balance for your specific use case between which method to use for the data handling and how expensive that method will be. To make an informed decision, you may need to benchmark and profile your smart contract code to measure its actual resource consumption. The primary purpose of ``asData`` is to give you two options for how you want to handle your data. 

Choice of two pathways
----------------------

Through ``PlutusTx.asData``, you have a choice of two pathways. It is up to you to determine which pathway to use depending on your particular use case. There are trade offs. The pathway you choose changes where any errors might happen. 

Method one: proactively do all of the parsing
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

The first approach is to parse the object ahead of time to determine whether or not it is an integer. If it is not an integer, there are issues to resolve. If it is an integer, it will be packaged up. 

Method two: only do the parsing if and when necessary 
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

In the second approach, the validator doesn't necessarily do anything. It only does the work when it needs to. In order to work with the data objects, the validator has to call the builtin functions, which is a little bit more expensive. It might be that this saves you a lot of work, because you may never need to parse the entire object. Instead, the validator will just carry the item around as if it were a data object. Some time later, when the validator needs to do an operation with this integer, then the validator will parse it. If it determines that it is not an integer, there will be errors. 

Using this method, every time the validator parses it, it is going to look at the data object to find out if it is an integer. If it is an integer, it will get the integer out and do its processing. The analysis work may be repeated depending on how your script is written. In some cases, you might do less work, in some cases you might do more work, depending on your specific use case. 

Using ``asData``
------------------

``asData`` works best when you use it for a type and all the types that go into it. The ``asData`` function takes the definition of a data type and replaces it with an equivalent definition whose representation uses ``Data`` directly.

``Data`` objects versus Plutus Tx's datatypes
---------------------------------------------

There are tradeoffs relating to the slower processing speed of ``Data`` objects versus Plutus Tx's datatypes.

Values stored in datums or redeemers
------------------------------------

```

doc/read-the-docs-site/howtos/asdata.rst
```diff
@@ -0,0 +1,89 @@
...
+Values stored in datums or redeemers need to be encoded into ``Data`` objects.
+This means that when a script starts to work with its datum or redeemer, it often wants to parse it into a more structured format using Plutus Tx's support for datatypes.
+Usually this is done with the ``fromBuiltinData`` function.
+
+The problem is that ``fromBuiltinData`` has to traverse the entire ``Data`` object in order to perform the transformation, which is costly.
```
joseph-fajen, line 11:
```suggestion
However, ``fromBuiltinData`` has to traverse the entire ``Data`` object in order to perform the transformation, which is costly.
```

doc/read-the-docs-site/howtos/asdata.rst
```diff
@@ -0,0 +1,89 @@
...
+The alternative is to work directly with the raw ``Data`` object.
+
+These two approaches have tradeoffs:
+
+1. If the resulting object is going to be processed in its entirety, or have parts of it repeatedly processed, then it can be better to convert it, since Plutus Tx's datatypes are faster to work with than ``Data``.
+2. If it is important to check that the entire structure is well-formed, then it is better to convert it, since the conversion will check for well-formedness.
+3. If correctness is important, then it can be better to convert it, since the compiler can help you more if you are using Plutus Tx's datatypes.
+
```
joseph-fajen, line 19:
```suggestion
When conversion to Plutus Tx's datatypes is appropriate
-------------------------------------------------------

It is appropriate to convert an object to Plutus Tx's datatypes under the following circumstances: 

1. When the resulting object is going to be processed in its entirety, or have parts of it repeatedly processed, since Plutus Tx's datatypes are faster to work with than ``Data``.
2. If it is important to check that the entire structure is well-formed, since the conversion will check for well-formedness.
3. If correctness is important, since the compiler can help you more if you are using Plutus Tx's datatypes.

```

PR #5732, opened 2024-01-19
cost-model-bench: Allow passing extra arguments

shlevy:


kwxm (approved):
This looks as if it should do what's required.

PR #5733, opened 2024-01-19
Bitwise costing (PLT-8790)

kwxm:
This extends the costing infrastructure to handle the `ByteStringToInteger` and `IntegerToByteString` functions added in #5654.  The CPU usage of these is quadratic at the moment, and that required quite a few additions.  Once we no longer he tavo support GHC 8.10 it's hoped that we can have a linear-time implementation instead:  this won't require any further changes to the costing infrastructure since a linear function can be represented by a quadratic function with a zero coefficient (or more likely a very small coefficient) in degree 2.   Also, `IntegerToByteString` will currently fail if the length of the output bytestring would be greater than 8000 (accurate costing becomes difficult beyond this point).  We may remove that limit when the implementation is updated, and if we do that will require a new semantic variant.  

`IntegerToByteString` takes a size argument that says how long the output should be, and that required another small extension so that the value of that argument is treated as a literal size for memory costing purposes (ie we don't take the memory usage to be the size of that argument, but the argument itself).

I've added costing benchmarks and run them, and added the resulting data to `builtinCostModel.json`.  The benchmarks may still need a little tuning, but that can be done later.

I also fixed a few small things that I ran into related to plugin integration.


plutus-core/plutus-core/src/PlutusCore/Bitwise/Convert.hs
kwxm, line 1: I moved this into this directory because that's where similar code for crypto buitlins is.  Maybe stuff like this should have its own directory.

plutus-core/plutus-core/src/PlutusCore/Builtin/Runtime.hs
kwxm, line 1: Drive-by typo elimination.

plutus-core/plutus-core/src/PlutusCore/Builtin/Runtime.hs
```diff
@@ -21,7 +21,7 @@ import NoThunks.Class
 -- 'BuiltinCostedResult' contains the cost (an 'ExBudget') and the result (a @BuiltinResult val@) of
 -- the builtin application. The cost is stored strictly, since the evaluator is going to look at it
 -- and the result is stored lazily, since it's not supposed to be forced before accounting for the
--- cost of the application. If the cost exceeds the available budget, the evaluator discards the the
+-- cost of the application. If the cost exceeds the available budget, the evaluator discards the
 -- result of the builtin application without ever forcing it and terminates with evaluation
```
kwxm, line 25: Drive-by typo elimination.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1801,19 +1803,28 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
     toBuiltinMeaning _semvar IntegerToByteString =
-      let integerToByteStringDenotation :: Bool -> Integer -> Integer -> BuiltinResult BS.ByteString
-          integerToByteStringDenotation = integerToByteStringWrapper
+      let integerToByteStringDenotation :: Bool -> LiteralByteSize -> Integer -> BuiltinResult BS.ByteString
+          {- The second argument is wrapped in a LiteralByteSize to allow us to interpret it as a size during
+             costing.  It appears as an integer in UPLC: see Note [Integral types as Integer]. -}
+          {- We disallow inputs of more than 8000 bytes because the execution time becomes difficult to predict
+             accurately beyond this point.  This restriction will be removed in a later variant. -}
+          integerToByteStringDenotation b (LiteralByteSize w) n =
+              if w > 8000 || (w == 0 && (sumCostStream . flattenCostRose . memoryUsage $ n) > 1000)
```
kwxm, line 1812: This is a bit annoying, but it has to go here because a later implementation may be able to dispense with this check, and then we'll need two variants and two denotations.

kwxm: The `sumCostStream . flattenCostRose . memoryUsage` pattern appears in several places in the codebase ...

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -195,6 +207,17 @@ instance ExMemoryUsage BS.ByteString where
...
+   should use `div`, but the second one says to use `quot` instead (and that is
+   what is used).  The only difference here is for negative numbers: (-1) `div`
+   8 == -1 and (-1) `quot` 8 == 0.  Thus the current memoryUsage for the empty
+   bytestring is 1: replacing `quot` with `div` would change the memoryUsage of
+   the empty bytestring from 1 to 0 and leave the memoryUsages of all other
+   bytestrings unchanged.  We can probably change this safely since all that it
+   would mean is that programs which use the empty bytestring as a builtin
+   argument would become a little cheaper.  The difference would be very small
+   though, so it's maybe not worth fixing. -}
+
```
kwxm, line 220: I just noticed this in passing: it's irrelevant to this PR.

plutus-benchmark/bls12-381-costs/test/9.6/bls12-381-costs.golden
```diff
@@ -3,143 +3,143 @@ Hash n bytestrings onto G1 and add points
...
   ----------------------------------------------------------------------
-    -     418   (2.6%)       361167811   (3.6%)          264488   (1.9%) 
+    -     372   (2.3%)       248182411   (2.5%)           13996   (0.1%) 
 
 Schnorr Signature G2
 
     n     Script size             CPU usage               Memory usage
   ----------------------------------------------------------------------
-    -     562   (3.4%)       606243489   (6.1%)          264656   (1.9%) 
+    -     516   (3.1%)       493258089   (4.9%)           14164   (0.1%)
```
effectfully, line 142: 1.9% to 0.1%? Does something get optimized away? Or does it now fail? Or why such a drastic difference?

plutus-core/cost-model/data/builtinCostModel.json
```diff
@@ -204,7 +204,7 @@
     },
     "bls12_381_G2_neg": {
         "cpu": {
-            "arguments": 307813,
+            "arguments": 129165,
```
effectfully, line 207: That's quite a reduction.

plutus-core/cost-model/data/builtinCostModel.json
```diff
@@ -237,34 +237,51 @@
...
         "memory": {
             "arguments": 1,
             "type": "constant_cost"
         }
     },
-    "bls12_381_mulMlResult": {
+    "bls12_381_millerLoop": {
         "cpu": {
-            "arguments": 2544991,
+            "arguments": 401885761,
```
effectfully, line 250: In case somebody is confused like I was, this is `bls12_381_mulMlResult` being replaced with `bls12_381_millerLoop`, not a 160x growth in cost.

plutus-core/cost-model/data/builtinCostModel.json
```diff
@@ -237,34 +237,51 @@
...
         }
     },
+    "byteStringToInteger": {
+        "cpu": {
+            "arguments": {
+                "c0": 936157,
+                "c1": 49601,
+                "c2": 237
+            },
+            "type": "quadratic_in_y"
```
effectfully, line 275: Quadratic? Strange. I guess I need to keep reading.

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
```diff
@@ -238,6 +238,7 @@ instance KnownTypeAst tyname uni a => KnownTypeAst tyname uni (Emitter a) where
     toTypeAst _ = toTypeAst $ Proxy @a
     {-# INLINE toTypeAst #-}
 
+
```
effectfully, line 241: Please remove the newline.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Convert.hs
```diff
@@ -3,7 +3,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 
 -- | Implementations for conversion primops from 'Integer' to 'ByteString' and back again.
-module PlutusCore.Builtin.Convert (
+module PlutusCore.Bitwise.Convert (
```
effectfully, line 6: Thanks! `Builtin` is for general builtins matter, not for anything Cardano-specific, so this is a helpful change.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1800,20 +1812,31 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
+          {- The second argument is wrapped in a LiteralByteSize to allow us to interpret it as a size during
+             costing.  It appears as an integer in UPLC: see Note [Integral types as Integer]. -}
+          integerToByteStringDenotation b (LiteralByteSize w) n =
+              if w > integerToByteStringMaximumOutputLength ||
+                     (w == 0 &&
+                            8 * fromSatInt (sumCostStream . flattenCostRose . memoryUsage $ n)
+                                  > integerToByteStringMaximumOutputLength)
+              then do
+                emit "byteStringToInteger result too big"
+                evaluationFailure
```
effectfully, line 1827: Thank you for adapting to the new way of doing things.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1800,20 +1812,31 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
     toBuiltinMeaning _semvar IntegerToByteString =
-      let integerToByteStringDenotation :: Bool -> Integer -> Integer -> BuiltinResult BS.ByteString
-          integerToByteStringDenotation = integerToByteStringWrapper
+      let integerToByteStringDenotation :: Bool -> LiteralByteSize -> Integer -> BuiltinResult BS.ByteString
+          {- The second argument is wrapped in a LiteralByteSize to allow us to interpret it as a size during
+             costing.  It appears as an integer in UPLC: see Note [Integral types as Integer]. -}
+          integerToByteStringDenotation b (LiteralByteSize w) n =
+              if w > integerToByteStringMaximumOutputLength ||
+                     (w == 0 &&
+                            8 * fromSatInt (sumCostStream . flattenCostRose . memoryUsage $ n)
```
effectfully, line 1823: That is a really weird line... The denotation of a builtin depends on costing for one of the arguments? Can we not rely on costing here and instead having some direct byte-handling logic? What if we change the size metric or something.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -417,6 +418,21 @@ instance HasConstantIn DefaultUni term => ReadKnownIn DefaultUni term Word8 wher
                _       -> throwing_ _EvaluationFailure
     {-# INLINE readKnown #-}
 
+-- | This allows us to transparently use values of type LiteralByteSize as
+-- built-in Integers but with a different size measure.
+instance KnownTypeAst tyname DefaultUni LiteralByteSize where
+    toTypeAst _ = toTypeAst $ Proxy @Integer
+
+-- See Note [Integral types as Integer].
+instance HasConstantIn DefaultUni term => MakeKnownIn DefaultUni term LiteralByteSize where
```
effectfully, line 427: You could probably simply use newtype-deriving for all the three instances?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -417,6 +418,21 @@ instance HasConstantIn DefaultUni term => ReadKnownIn DefaultUni term Word8 wher
...
+    makeKnown = makeKnown . unLiteralByteSize
+    {-# INLINE makeKnown #-}
+
+instance HasConstantIn DefaultUni term => ReadKnownIn DefaultUni term LiteralByteSize where
+    readKnown term = LiteralByteSize <$> readKnown term
+    {-# INLINE readKnown #-}
+
+
```
effectfully, line 435: Nit: two newlines suck.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -397,17 +437,23 @@ runTwoArgumentModel
                 if size1 < size2
                     then CostLast c
                     else run (CostLast size1) (CostLast size2)
+runTwoArgumentModel
+    (ModelTwoArgumentsQuadraticInY f) =
+        lazy $ \_ costs2 ->
+            CostLast $ evaluateOneVariableQuadraticFunction f $ sumCostStream costs2
```
effectfully, line 443: This needs a comment that costing here isn't lazy, so if lazy costing is required for some future quadratic builtin, this piece of code needs to be rewritten.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -437,6 +483,21 @@ runThreeArgumentModel
     (ModelThreeArgumentsLinearInZ (OneVariableLinearFunction intercept slope)) =
         lazy $ \_ _ costs3 ->
             scaleLinearly intercept slope costs3
+runThreeArgumentModel
+    (ModelThreeArgumentsQuadraticInZ f) =
+        lazy $ \_ _ costs3 -> CostLast $ evaluateOneVariableQuadraticFunction f $ sumCostStream costs3
```
effectfully, line 488: Same. I guess we need a Note. And we should reference it in old costing functions too. I should've done that myself when I introduced lazy costing. Feel free to do it here or create a ticket for me or yourself.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -437,6 +483,21 @@ runThreeArgumentModel
...
+{- Either a literal number of bytes or a linear function.  This is for
+   `integerToByteString`, where if the second argument is zero, the output
+   bytestring has the minimum length required to contain the converted integer,
+   but if the second argument is nonzero it specifies the exact length of the
+   output bytestring. -}
+runThreeArgumentModel
+    (ModelThreeArgumentsLiteralInYOrLinearInZ (OneVariableLinearFunction intercept slope)) =
+        lazy $ \_ costs2 costs3 ->
+            let width = sumCostStream costs2
+            in if width == 0
```
effectfully, line 498: Measuring `width` in costing units is extremely weird in my opinion and further reinforces the fact that our current approach needs revising.

plutus-ledger-api/test/Spec/CostModelParams.hs
```diff
@@ -32,8 +32,8 @@ tests =
             166 @=? length V1.costModelParamsForTesting
             175 @=? length v2_ParamNames
             175 @=? length V2.costModelParamsForTesting
-            223 @=? length v3_ParamNames
-            223 @=? length V3.costModelParamsForTesting
+            233 @=? length v3_ParamNames
+            233 @=? length V3.costModelParamsForTesting
```
effectfully, line 36: Should be golden tests, I think.

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
```diff
@@ -452,6 +455,9 @@ defineBuiltinTerms = do
             PLC.Bls12_381_mulMlResult -> defineBuiltinInl 'Builtins.bls12_381_mulMlResult
             PLC.Bls12_381_finalVerify -> defineBuiltinInl 'Builtins.bls12_381_finalVerify
 
+    -- Bitwise operations
+    defineBuiltinTerm annMayInline 'Builtins.integerToByteString               $ mkBuiltin PLC.IntegerToByteString
+    defineBuiltinTerm annMayInline 'Builtins.byteStringToInteger               $ mkBuiltin PLC.ByteStringToInteger
```
effectfully, line 460: Should be moved under the pattern matching right above, following the pattern.

effectfully (commented):
Haven't reviewed fully yet, will return shortly. 

Looks good to me, but I don't like how the costing function leaks into the denotation and the denotation leaks into the costing function. We should either embrace that the two are intertwined in a way that does not allow us to separate them meaningfully or do something else, the current approach is really weird in my opinion. Though, I don't mind to merge it first before we figure out a better design (if it's possible).

effectfully (commented):
Finishing the review with no new comments. 

I'm puzzled as to why the new builtins are quadratic, do you happen to have a rendered plot with data points?

plutus-core/cost-model/data/builtinCostModel.json
kwxm: That's unrelated to this PR.  I updated the benchmarking results for the bls12-381 functions at some point and they've only just made it into `builtinCostModel.json`.  In retrospect I should probably have updated those in a separate PR to avoid confusion.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
kwxm: > Measuring width in costing units is extremely weird in my opinion

Yes, I hate this.  Sizes in bytes would be much better (but it's tricky to change that at the moment).

kwxm:
> I'm puzzled as to why the new builtins are quadratic, do you happen to have a rendered plot with data points?

That was discussed on Slack.  I've sent you some links.

plutus-ledger-api/test/Spec/CostModelParams.hs
kwxm: > Should be golden tests, I think.

Hmmm.  I just extended what was there already.  @bezirg ?

plutus-core/plutus-core/src/PlutusCore/Bitwise/Convert.hs
james-iohk, line 1: Line 41 has a `-- TODO: Cost this builtin.` Can this comment be removed now?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
kwxm: That was probably just accidental, but sometimes I find it useful to have a bit more space to break the code into blocks of related functionality. Maybe in that case I should use a comment rather than blank lines.

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
kwxm: Thanks.  I'm not sure how that ended up there.

kwxm:
> I don't like how the costing function leaks into the denotation and the denotation leaks into the costing function.

Neither do I.  It occurs to me now that instead of 
```
  (w == 0 &&
        8 * fromSatInt (sumCostStream . flattenCostRose . memoryUsage $ n)
             > integerToByteStringMaximumOutputLength)
```
I could say 
```
 (w == 0 && 
    n >= 173766203193809456599982445949435627061939786100117250547173286503262376022458008465094333630120854338003194362163007597987225472483598640843335685441710193966274131338557192586399006789292714554767500194796127964596906605976605873665859580600161998556511368530960400907199253450604168622770350228527124626728538626805418833470107651091641919900725415994689920112219170907023561354484047025713734651608777544579846111001059482132180956689444108315785401642188044178788629853592228467331730519810763559577944882016286493908631503101121166109571682295769470379514531105239965209245314082665518579335511291525230373316486697786532335206274149240813489201828773854353041855598709390675430960381072270432383913542702130202430186637321862331068861776780211082856984506050024895394320139435868484643843368002496089956046419964019877586845530207748994394501505588146979082629871366088121763790555364513243984244004147636040219136443410377798011608722717131323621700159335786445601947601694025107888293017058178562647175461026384343438874861406516767158373279032321096262126551620255666605185789463207944391905756886829667520553014724372245300878786091700563444079107099009003380230356461989260377273986023281444076082783406824471703499844642915587790146384758051663547775336021829171033411043796977042190519657861762804226147480755555085278062866268677842432851421790544407006581148631979148571299417963950579210719961422405768071335213324842709316205032078384168750091017964584060285240107161561019930505687950233196051962261970932008838279760834318101044311710769457048672103958655016388894770892065267451228938951370237422841366052736174160431593023473217066764172949768821843606479073866252864377064398085101223216558344281956767163876579889759124956035672317578122141070933058555310274598884089982879647974020264495921703064439532898207943134374576254840272047075633856749514044298135927611328433323640657533550512376900773273703275329924651465759145114579174356770593439987135755889403613364529029604049868233807295134382284730745937309910703657676103447124097631074153287120040247837143656624045055614076111832245239612708339272798262887437416818440064925049838443370805645609424314780108030016683461562597569371539974003402697903023830108053034645133078208043917492087248958344081026378788915528519967248989338592027124423914083391771884524464968645052058218151010508471258285907685355807229880747677634789376
```
The number there is 2^8000, so maybe we could just write that instead.  I think I was a bit worried about having such a big number in the source, but maybe it doesn't matter so much.  We could also use `integerLog2 n > 8000`: maybe that'd be better.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Convert.hs
kwxm: Yes, thanks.

plutus-benchmark/bls12-381-costs/test/9.6/bls12-381-costs.golden
kwxm: > 1.9% to 0.1%? Does something get optimized away? Or does it now fail? Or why such a drastic difference?

Good question.  You had me worried, but it's because the computation's chaining together a bunch of bls12-381 builtins and the memory costs of those are pretty small:
```
bls12_381_G1_add                : 18
bls12_381_G1_compress           : 6
bls12_381_G1_equal              : 1
bls12_381_G1_hashToGroup        : 18
bls12_381_G1_neg                : 18
bls12_381_G1_scalarMul          : 18
bls12_381_G1_uncompress         : 18
bls12_381_G2_add                : 36
bls12_381_G2_compress           : 12
bls12_381_G2_equal              : 1
bls12_381_G2_hashToGroup        : 36
bls12_381_G2_neg                : 36
bls12_381_G2_scalarMul          : 36
bls12_381_G2_uncompress         : 36
bls12_381_finalVerify           : 1
bls12_381_millerLoop            : 72
bls12_381_mulMlResult           : 72
```
The CPU costs are a lot bigger, so the proportional change in the total CPU cost is much smaller.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
kwxm: I think I considered that and then decided that it would be easier just to mimic what it does for the other integral types.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
kwxm: For [other reviewers](https://github.com/IntersectMBO/plutus/pull/5733#issuecomment-1903468249), see this comment later.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
kwxm: Since we're talking about laziness, I made a ticket for you to do it at PLT-9312.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -417,6 +418,21 @@ instance HasConstantIn DefaultUni term => ReadKnownIn DefaultUni term Word8 wher
                _       -> throwing_ _EvaluationFailure
     {-# INLINE readKnown #-}
 
+-- | This allows us to transparently use values of type LiteralByteSize as
+-- built-in Integers but with a different size measure.
+instance KnownTypeAst tyname DefaultUni LiteralByteSize where
+    toTypeAst _ = toTypeAst $ Proxy @Integer
+
+-- See Note [Integral types as Integer].
```
michaelpj, line 426: I think this note is wrong. It might need its own note!

plutus-benchmark/bls12-381-costs/test/9.6/bls12-381-costs.golden
michaelpj: I mean, this is what we'd expect since the DIY version is awful?

plutus-core/cost-model/print-cost-model/Main.hs
```diff
@@ -131,35 +151,44 @@ renderLinearFunction (LinearFunction intercept slope) var =
...
+     MaxSize               f   -> [ renderLinearFunction f "max(x,y)" ]
+     LinearCost            f   -> [ renderLinearFunction f "x" ]
+     LinearInX             f   -> [ renderLinearFunction f "x" ]
+     LinearInY             f   -> [ renderLinearFunction f "y" ]
+     LinearInZ             f   -> [ renderLinearFunction f "z" ]
+     QuadraticInY          f   -> [ renderQuadraticFunction f "y" ]
+     QuadraticInZ          f   -> [ renderQuadraticFunction f "z" ]
+     LiteralInYOrLinearInZ f -> [ "if y==0"
+                                  , printf "then %s" $ renderLinearFunction f "z"
+                                  , printf "else y bytes"
```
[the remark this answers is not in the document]
kwxm: Oh, that's true.  In principle you could use it for CPU usage too, but at the moment it's just for dealing with the awkward memory costs for `integerToByteString`.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
kwxm: That's @effectfully 's note, so he should take a look.  Possibly I've added something that means that it no longer makes sense.

plutus-core/cost-model/data/builtinCostModel.json
```diff
@@ -128,7 +128,7 @@
     },
     "bls12_381_G1_neg": {
         "cpu": {
-            "arguments": 292890,
+            "arguments": 114242,
```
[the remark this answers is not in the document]
kwxm: Yes.  I think the earlier figures for these were perhaps generated before we got the new machine, and then at some point I re-generated them but didn't commit the results.  We should re-run the entire thing before the next HF anyway though.  See [this comment](https://github.com/IntersectMBO/plutus/pull/5733#discussion_r1460246687) too.

plutus-core/cost-model/budgeting-bench/Main.hs
```diff
@@ -1,8 +1,10 @@
+{-# OPTIONS_GHC -fno-warn-unused-imports #-}
```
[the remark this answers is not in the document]
kwxm: Left over from when I was experimenting.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: I usually use a comment indeed. Rarely I use three newline. Two newlines look like accidental noise to me (not enough signal like with three lines), but that's clearly just me being crazy.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
effectfully: Thanks!

kwxm:
I'll move the size checks back into the implementation so that the denotation doesn't have to care.  If we decide to remove the limit at a later date then we'll have to introduce two variants in `Default.Builtin` to preserve the current behaviour for old scripts.

kwxm:
@michaelpj @effectfully I've now moved the size checks out of `Default.Builtins` and back into `Bitwise.Conversions" and added some tests, so you might want to take another look.

kwxm:
The maximum size for outputs of `integerToByteString` is 8000, but maybe 8192 would be better.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Convert.hs
```diff
@@ -25,23 +29,62 @@ import Data.ByteString qualified as BS
...
+   number of different machines show that the CPU time increases smoothly for
+   inputs up to about 8K then increases sharply, becoming chaotic after about
+   14K).  This restriction may be removed once a more efficient implementation
+   becomes available, which may happen when we no longer have to support GHC
+   8.10. -}
+{- NB: if we do relax the length restriction then we will need two variants of
+   integerToByteString in Plutus Core so that we can continue to support the
+   current behaviour for old scripts.-}
+integerToByteStringMaximumOutputLength :: Integer
+integerToByteStringMaximumOutputLength = 8000
```
kwxm, line 48: Maybe 8192 would be better.

plutus-benchmark/bls12-381-costs/test/9.6/bls12-381-costs.golden
effectfully: Great to know, thanks!

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: I think Michael probably means that the Note does not apply here, not that it is wrong per se? The Note talks about how we encode integral types as `Integer` to get automatic bound checking, but here you do something very different: you need the wrapper for costing reasons and all those instances are literally the same as ones for `Integer`. I would therefore `newtype`-derive them and simply avoid referencing any Notes, `newtype`-deriving is enough of a stress that from the builtins point of view `LiteralByteSize` is indistinguishable from `Integer`.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Convert.hs
effectfully: I'd prefer that indeed, because 8000 feels entirely arbitrary and 8K somehow feels less arbitrary, but it's again my craziness is talking. `8192` also gives the reader a hint of what this is about.

plutus-conformance/agda/Spec.hs
```diff
@@ -37,8 +38,57 @@ agdaEvalUplcProg (UPLC.Program () version tmU) =
...
+-- The entries of the list should be paths from the root of plutus-conformance
+-- to the directory containing the test, eg
+--   "test-cases/uplc/evaluation/builtin/semantics/addInteger/addInteger1"
+
 failingTests :: [FilePath]
-failingTests = [
+failingTests =
+    [
+     --- byteStringToInteger
+      "test-cases/uplc/evaluation/builtin/semantics/byteStringToInteger/big-endian/all-zeros"
```
effectfully, line 49: It says `failingTests`, but `all-zeros` isn't a failing test? I'm confused.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/integerToByteString/little-endian/bounded/maximum-width-zero/maximum-width-zero.uplc
```diff
@@ -0,0 +1,4 @@
+-- Check that we can encode zero using the maximum width (8000).
+(program 1.0.1
+ [(builtin integerToByteString) (con bool False) (con integer 8000) (con integer 0)]
```
effectfully, line 3: If you're going to make it 8192, don't forget to update this test. And I guess plenty of other ones. So maybe it's easier to just keep 8000.

effectfully (approved):
Great tests.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Convert.hs
```diff
@@ -25,23 +28,62 @@ import Data.ByteString qualified as BS
...
-  -- costed, this won't be a problem.
-  --
-  -- TODO: Cost this builtin.
-  | lengthArg > 10240 = do
-      emit "integerToByteString: padding argument too large"
-      emit "If you are seeing this, it is a bug: please report this!"
+  -- Check that the requested length does not exceed the limit.  *NB*: if we
+  -- remove the limit we'll still have to make sure that the length fits into an
+  -- Int (so it should be at most 2^29, which is the largest bound that GHC
+  -- guarantees).
```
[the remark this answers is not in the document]
zliu41: Yeah, I don't think we need to worry about 32bit platforms.

zliu41 (approved):
Great work.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Convert.hs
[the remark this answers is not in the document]
michaelpj: Yep, this is right.

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
```diff
@@ -331,6 +334,10 @@ defineBuiltinTerms = do
     defineBuiltinTerm annMayInline 'Builtins.bls12_381_G2_compressed_zero $
         PIR.mkConstant annMayInline BLS12_381.G2.compressed_zero
 
+    -- Bitwise operations
+    defineBuiltinTerm annMayInline 'Builtins.integerToByteString $ mkBuiltin PLC.IntegerToByteString
+    defineBuiltinTerm annMayInline 'Builtins.byteStringToInteger $ mkBuiltin PLC.ByteStringToInteger
```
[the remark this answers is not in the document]
michaelpj: I think we could just make it throw if you redefine the same name. There's really no reason to do that.

plutus-tx/src/PlutusTx/Builtins.hs
```diff
@@ -603,13 +603,13 @@ bls12_381_finalVerify a b = fromBuiltin (BI.bls12_381_finalVerify a b)
 
 -- | Convert a 'BuiltinInteger' into a 'BuiltinByteString', as described in
 -- [CIP-0087](https://github.com/mlabs-haskell/CIPs/tree/koz/to-from-bytestring/CIP-XXXX).
-{-# INLINEABLE integerToByteString #-}
+{-# INLINABLE integerToByteString #-}
 integerToByteString :: Bool -> Integer -> Integer -> BuiltinByteString
```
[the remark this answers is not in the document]
michaelpj: This is a good point, for the "exposed" version of the builtins we can indeed give a nicer interface again.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Convert.hs
kwxm: > I'd prefer that indeed, because 8000 feels entirely arbitrary

Oh well, I suppose so.  I'll have to change a lot of  the conformance tests now too [goes off grumbling].

plutus-core/plutus-core/src/PlutusCore/Bitwise/Convert.hs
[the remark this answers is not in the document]
kwxm: OK, I've removed that and just said it should fit into an `Int`.

kwxm:
Oh, I seem to have accidentally merged the conformance tests with this PR when I was trying to do the opposite. That's a bit confusing.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Conversion.hs
```diff
@@ -403,6 +403,71 @@ i2bCipExamples = [
...
+                         then "Big-endian, "
+                         else "Little-endian, "
+            in [
+             -- integerToByteString 0 maxInput = 0xFF...FF
+             testCase (prefix ++ "maximum acceptable input, no length specified") $
+                      let actualExp = mkIterAppNoAnn (builtin () PLC.IntegerToByteString) [
+                                       mkConstant @Bool () endianness,
+                                       mkConstant @Integer () 0,
+                                       mkConstant @Integer () maxAcceptableInput
+                                      ]
```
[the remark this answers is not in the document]
kwxm: :+1: Good idea.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/byteStringToInteger/big-endian/all-zeros/all-zeros.uplc
```diff
@@ -0,0 +1,4 @@
+-- A bytestring consisting entirely of zeros decodes to 0.
+(program 1.0.1
```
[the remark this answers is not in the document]
kwxm: Oops.  I'll just make it 1.0.0 since that's what all the others have and this doesn't depend on 1.1.0 behaviour.

plutus-core/cost-model/budgeting-bench/Benchmarks/Bitwise.hs
```diff
@@ -0,0 +1,83 @@
...
+---------------- ByteString builtins ----------------
+
+-- Smallish bytestring inputs: 150 entries.  Note that the length of a
+-- bytestring is eight times the size.
+smallerByteStrings150 :: H.Seed -> [BS.ByteString]
+smallerByteStrings150 seed = makeSizedByteStrings seed [1..150]
+
+-- Make an integer of size n which encodes to 0xFF...FF
+allFF :: Int -> Integer
+allFF n = 256^(8*n) - 1
```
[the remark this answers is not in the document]
kwxm: > What is 256? Is it not `2^(8*n) - 1`?

No, that's right.  We're looking for _size_ n here, which means 8*n bytes (thanks again to the infuriating fact that we measure sizes in 8-byte words).  Since that's a number of bytes, the total number of _bits_ is 8*8*n, and 2^(8*8*n) = 256^(8*n) is 1000...000 with 64*n zero bits.  Subtracting 1 from that removes the 1 at the front and changes all the remaining bits to 1, so you've got 64*n 1 bits, or 8*n 0xFF bytes, which is what you want.  Thanks for making me check though: it's very easy to get this kind of thing wrong.

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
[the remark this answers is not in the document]
kwxm: > I don't think you need these three lines? 

Ooh, well-spotted.  Maybe we should have some tests that make sure that you can actually use all of the builtins in Haskell code: it's quite easy to forget to add something here or in plutus-tx (although here we've added something too often).  

What happens if you define something twice with different semantics here?  Does the later definition override the earlier one?

plutus-conformance/agda/Spec.hs
kwxm: > It says failingTests, but all-zeros isn't a failing test? I'm confused.

These do fail in the metatheory because in this branch the Agda code doesn't know about the new builtins yet.  That'll be fixed in a PR that'll appear shortly (and which I should have done before this one).  In that branch `failingTests` is empty.

kwxm: If you run the `agda-conformance` tests you get errors like
```
          big-endian
            all-zeros: FAIL (expected)
              Agda: unreachable code reached.
                error, called at /blah/blah/blah/plutus-metatheory-0.1.0.0/build/MAlonzo/RTE.hs:44:23 in plutus-metatheory-0.1.0.0-inplace:MAlonzo.RTE (expected failure)

plutus-tx/src/PlutusTx/Builtins.hs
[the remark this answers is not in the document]
kwxm: Yes, that'd be quite helpful.

plutus-tx-plugin/test/AsData/Budget/9.6/patternMatching.uplc.golden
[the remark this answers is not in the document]
kwxm: Oh!  Now it's gone back to what it was before.

plutus-tx/src/PlutusTx/Builtins.hs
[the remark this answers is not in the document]
kwxm: OK, I've made the plutus-tx versions take a `ByteOrder` argument.

plutus-conformance/agda/Spec.hs
effectfully: I see, thank you for the explanation.

kwxm:
Right, I'm going to merge this while the going's good.

PR #5740, opened 2024-01-25
Build all tasty tests with the threaded runtime and some capabilities

michaelpj:
This ensures that tasty actually runs tests in parallel. This makes a small but noticeable difference.


kwxm:
Yes, that seems to help.

PR #5741, opened 2024-01-25
PLT-9372  Extend Agda costing infrastructure to accommodate `integerToByteString`

mjaskelioff:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-metatheory/src/Cost/Size.lagda.md
+++ b/plutus-metatheory/src/Cost/Size.lagda.md
@@ -1,0 +1,114 @@
+# Size of Constants
+
+```
+module Cost.Size where 
+
+```
+
+# Imports
+
+```
+open import Data.Nat using (ℕ;zero;suc;_+_;_*_;_∸_)
+open import Data.Nat.DivMod using (_/_)
+open import Data.Integer using (ℤ;∣_∣)
+open import Data.String using (length)
+
+open import Utils using (_×_;_,_;_∷_;[];DATA;List;ByteString)
+open DATA
+
+open import Builtin using (lengthBS)
+open import Builtin.Signature using (_⊢♯)
+open _⊢♯
+
+open import Builtin.Constant.AtomicType using (AtomicTyCon)
+open AtomicTyCon
+open import Cost.Base 
+open import Untyped.CEK using (Value;V-con)
+open import RawU using (TmCon;tmCon) 
+
+open import Data.Vec using (Vec;[];_∷_) 
+                     renaming (lookup to lookupVec)
+
+```
+
+## Memory usage of type constants
+
+For each type constant we calculate its size, as a measure of memory usage.
+
+First we bring some functions from Haskell world.
+
+```
+postulate ℕtoWords : ℤ → CostingNat 
+postulate g1ElementCost : CostingNat
+postulate g2ElementCost : CostingNat
+postulate mlResultElementCost : CostingNat 
+
+{-# FOREIGN GHC {-# LANGUAGE MagicHash #-} #-}
+{-# FOREIGN GHC import GHC.Exts (Int (I#)) #-}
+{-# FOREIGN GHC import GHC.Integer  #-}
+{-# FOREIGN GHC import GHC.Integer.Logarithms  #-}
+{-# FOREIGN GHC import GHC.Prim #-}
+{-# FOREIGN GHC import PlutusCore.Crypto.BLS12_381.G1 as BLS12_381.G1 #-}
+{-# FOREIGN GHC import PlutusCore.Crypto.BLS12_381.G2 as BLS12_381.G2 #-}
+{-# FOREIGN GHC import PlutusCore.Crypto.BLS12_381.Pairing as BLS12_381.Pairing #-}
+
+{-# COMPILE GHC ℕtoWords = \i -> fromIntegral $ I# (integerLog2# (abs i) `quotInt#` integerToInt 64) + 1 #-}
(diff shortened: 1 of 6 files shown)


plutus-metatheory/src/Untyped/CEKWithCost.lagda.md
```diff
@@ -22,14 +22,15 @@ open import Untyped using (_⊢)
 open _⊢
 open import Utils using (Writer;_>>_;_>>=_;return;maybe;Either;inj₁;inj₂;RuntimeError;gasError;either;_∔_≣_)
 open Writer
-open import RawU using (tmCon)
+open import RawU using (tmCon;TmCon)
```
kwxm, line 25: Is `TmCon` ever used?

plutus-metatheory/src/Cost/Size.lagda.md
```diff
@@ -0,0 +1,114 @@
...
+{-# FOREIGN GHC {-# LANGUAGE MagicHash #-} #-}
+{-# FOREIGN GHC import GHC.Exts (Int (I#)) #-}
+{-# FOREIGN GHC import GHC.Integer  #-}
+{-# FOREIGN GHC import GHC.Integer.Logarithms  #-}
+{-# FOREIGN GHC import GHC.Prim #-}
+{-# FOREIGN GHC import PlutusCore.Crypto.BLS12_381.G1 as BLS12_381.G1 #-}
+{-# FOREIGN GHC import PlutusCore.Crypto.BLS12_381.G2 as BLS12_381.G2 #-}
+{-# FOREIGN GHC import PlutusCore.Crypto.BLS12_381.Pairing as BLS12_381.Pairing #-}
+
+{-# COMPILE GHC ℕtoWords = \i -> fromIntegral $ I# (integerLog2# (abs i) `quotInt#` integerToInt 64) + 1 #-}
```
kwxm, line 55: There may be a tiny danger of incompatibility for the size of 0 here because of [this](https://github.com/IntersectMBO/plutus/blob/4da3e0ad032209d6cb6e9aa4c0c9e2f2325d4763/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs#L189), which we only discovered a few days ago.  I don't think it's a big deal though.

plutus-metatheory/src/Cost/Size.lagda.md
```diff
@@ -0,0 +1,114 @@
...
+-- The following two functions are stupid but
+-- they make the termination checker happy
+sizeDataDataList [] = 0
+sizeDataDataList ((x , y) ∷ xs) = sizeData x + sizeData y + sizeDataDataList xs
+
+sizeDataList [] = 0
+sizeDataList (x ∷ xs) = sizeData x + sizeDataList xs
+
+defaultConstantMeasure : TmCon → CostingNat
+defaultConstantMeasure (tmCon (atomic aInteger) x) = ℕtoWords x
```
kwxm, line 91: The fact that these measures have to be the same as the Haskell versions is maybe a source of fragility since we might forget to keep them synchronised (although I suppose that budget tests might help with that).  Maybe it's unavoidable though.  If we ever want to prove something about costs we may need to be able to see exactly what the costing functions look like (so we can check that they're monotone or always positive or something),  and if we were somehow  accessing the Haskell versions directly instead then that might not be possible (I'm just thinking aloud here).

kwxm (approved):

kwxm:
This all looks good, and I tried it out and got sensible results.  The fact that the costing functions have to look up the sizes is perhaps a bit annoying, but I'm not sure that there's any sensible alternative.  I think we'll probably have to do something similar in the paper specification though, and we've talked about allowing multiple size measures for a single, which I think would definitely require a similar approach. 

Purely out of curiosity, could you have introduced a new type for the size argument of `integerToByteString` and then converted that to an integer at the point where the Agda builtin is wired up to the Haskell one?

mjaskelioff:
I think this approach is what is going to gives us more flexibility in the future. The fact that now we can compare the agda and the haskell costs will also give us security that the flexibility doesn't become a problem.

plutus-metatheory/src/Untyped/CEKWithCost.lagda.md
mjaskelioff: It was on a version, but it is not needed now.

plutus-metatheory/src/Cost/Size.lagda.md
mjaskelioff: Fixed. Now returns 1 for i == 0.

kwxm:
> I much prefer it this way, where the model decides how to use the argument.

Yes, I think that's sensible. It'd need some major refactoring to make the Haskell version behave like that, but that doesn't matter as long as the two versions behave identically when viewed from the outside.  I guess you could have a builtin which is linear in the value of some argument (rather than the size of the argument) and then you'd have to introduce a new model similar to `LinearInX` or something, but calculates the cost in a different way.  With luck there won't be too much of that kind of thing.

Anyway, this is passing CI now so I'll merge it and then open a PR on the other branch once the main costing branch is merged.  Thanks!

plutus-metatheory/src/Cost/Size.lagda.md
mjaskelioff: I think it's ok to abstract over constants. We probably don't need to know the precise value to reason over them (unless we want to calculate formally that something is an optimization).

PR #5745, opened 2024-01-27
[Test] Improve term shrinking

effectfully:
This improves term shrinking a bit and adds shrinking to a test.

### DIFF
--- a/plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
+++ b/plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
@@ -45,7 +45,6 @@ class ArbitraryBuiltin a where
 
 instance ArbitraryBuiltin ()
 instance ArbitraryBuiltin Bool
-instance ArbitraryBuiltin Data
 
 {- Note [QuickCheck and integral types]
 The 'Arbitrary' instances for 'Integer' and 'Int64' only generate small integers:
@@ -184,14 +183,39 @@ genDataFromSpine els = oneof
             ]
     ]
 
-instance Arbitrary Data where
-    arbitrary = arbitrary >>= genDataFromSpine
+pureIfNull :: (Foldable f, Applicative f) => a -> f a -> f a
+pureIfNull x xs = if null xs then pure x else xs
 
-    shrink (Constr i ds) = ds ++ map (Constr i) (shrink ds)
-    shrink (Map ps)      = map fst ps ++ map snd ps ++ map Map (shrink ps)
-    shrink (List ds)     = ds ++ map List (shrink ds)
-    shrink (I i)         = map I (shrink i)
-    shrink (B b)         = I 0 : map B (shrink b)
+instance ArbitraryBuiltin Data where
+    arbitraryBuiltin = arbitrary >>= genDataFromSpine
+
+    -- We arbitrarily assume that @I 0@ is the smallest 'Data' object just so that anything else in
+    -- a counterexample gives us a clue as to what the culprit may be. Hence @I 0@ needs to be
+    -- reachable from all nodes apart from @I 0@ itself. For all nodes but 'I' we achieve this by
+    -- returning @[I 0]@ if there's no other way to shrink the value, i.e. either shrinking keeps
+    -- going or it's time to return the smallest object. The clause for @I i@ doesn't require
+    -- mentioning @I 0@ explicitly, since we get it through general shrinking of @i@ (unless @i@
+    -- equals @0@, as desired).
+    shrinkBuiltin (Constr i ds) = pureIfNull (I 0) $ concat
+        [ ds
+        , map (Constr i) $ shrinkBuiltin ds
+        , map (flip Constr ds) $ shrinkBuiltin i
+        ]
+    shrinkBuiltin (Map ps) = pureIfNull (I 0) $ concat
+        [ map fst ps
+        , map snd ps
+        , map Map $ shrinkBuiltin ps
+        ]
+    shrinkBuiltin (List ds) = pureIfNull (I 0) $ concat
+        [ ds
+        , map List $ shrinkBuiltin ds
+        ]
+    shrinkBuiltin (B b) = pureIfNull (I 0) . map B $ shrinkBuiltin b
+    shrinkBuiltin (I i) = map I $ shrinkBuiltin i
+
+instance Arbitrary Data where
+    arbitrary = arbitraryBuiltin
+    shrink = shrinkBuiltin
 
 instance ArbitraryBuiltin BLS12_381.G1.Element where
     arbitraryBuiltin =
--- a/plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
+++ b/plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
@@ -9,9 +9,10 @@
 
 module PlutusIR.Generators.QuickCheck.ShrinkTerms where
 
+import PlutusPrelude
+
 import PlutusIR.Generators.QuickCheck.Common
 
-import PlutusCore.Generators.QuickCheck.Builtin
 import PlutusCore.Generators.QuickCheck.Common
 import PlutusCore.Generators.QuickCheck.ShrinkTypes
 import PlutusCore.Generators.QuickCheck.Substitutions
@@ -31,11 +32,8 @@ import PlutusIR
 import PlutusIR.Subst
 
 import Data.Bifunctor
-import Data.Either
-import Data.List.NonEmpty (NonEmpty (..))
 import Data.Map.Strict (Map)
 import Data.Map.Strict qualified as Map
-import Data.Proxy
 import Data.Set qualified as Set
 import Data.Set.Lens (setOf)
 import GHC.Stack
@@ -47,10 +45,6 @@ addTmBind :: Binding TyName Name DefaultUni DefaultFun ()
 addTmBind (TermBind _ _ (VarDecl _ x a) _) = Map.insert x a
 addTmBind (DatatypeBind _ dat)             = (Map.fromList (matchType dat : constrTypes dat) <>)
 addTmBind _                                = id
-
-shrinkConstant :: DefaultUni (Esc a) -> a -> [Term TyName Name DefaultUni DefaultFun ()]
-shrinkConstant uni x =
-    map (mkConstantOf () uni) $ bring (Proxy @ArbitraryBuiltin) uni $ shrinkBuiltin x
 
 scopeCheckTyVars :: TypeCtx
                  -> (Type TyName DefaultUni (), Term TyName Name DefaultUni DefaultFun ())
@@ -275,9 +269,19 @@ shrinkTypedTerm tyctx0 ctx0 (ty0, tm0) = concat
           | TyFun _ _ b <- [ty]
           ]
 
-        Apply _ fun arg | Right argTy <- inferTypeInContext tyctx ctx arg ->
-          -- Drop substerms
-          [(argTy, arg), (TyFun () argTy ty, fun)]
+        -- Drop substerms
+        Apply _ fun arg -> case inferTypeInContext tyctx ctx arg of
+          Right argTy ->
+            [ -- Prefer to keep the same type, so that we don't need to change types elsewhere,
+              -- which may remove or cover up the culprit.
+              (ty, fixupTerm tyctx ctx tyctx ctx ty arg)
+            , (ty, fixupTerm tyctx ctx tyctx ctx ty fun)
+            , -- But support type-changing shrinking as well in case it's 'fixupTerm' that removes
+              -- or covers up the culprit.
+              (argTy, arg)
+            , (TyFun () argTy ty, fun)
+            ]
+          Left err -> error $ displayPlcCondensedErrorClassic err
 
         TyAbs _ x _ body ->
           [ fixupTerm_ (Map.insert x k tyctx) ctx tyctx ctx tyInner' body
@@ -285,12 +289,16 @@ shrinkTypedTerm tyctx0 ctx0 (ty0, tm0) = concat
           , let tyInner' = typeSubstClosedType y (minimalType k) tyInner
           ]
 
-        -- Builtins can shrink to unit. More fine-grained shrinking is in `structural` below.
-        Constant _ val -> do
-            val'@(Some (ValueOf uni _)) <- shrink val
-            pure (mkTyBuiltinOf () uni, Constant () val')
-
-        _ -> []
+        -- TODO: allow non-structural shrinking for some of these.
+        Var{} -> []
+        Constant{} -> []
+        Builtin{} -> []
+        TyInst{} -> []
+        Error{} -> []
+        IWrap{} -> []
+        Unwrap{} -> []
+        PlutusIR.Constr{} -> []
+        Case{} -> []
 
     -- These are the structural (basically homomorphic) cases in shrinking.
     -- They all just try to shrink a single subterm at a time. We also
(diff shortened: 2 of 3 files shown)


plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
```diff
@@ -184,14 +183,18 @@ genDataFromSpine els = oneof
...
+instance ArbitraryBuiltin Data where
+    arbitraryBuiltin = arbitrary >>= genDataFromSpine
+
+    shrinkBuiltin (Constr i ds) = ds ++ map (Constr i) (shrinkBuiltin ds)
+    shrinkBuiltin (Map ps)      = map fst ps ++ map snd ps ++ map Map (shrinkBuiltin ps)
+    shrinkBuiltin (List ds)     = ds ++ map List (shrinkBuiltin ds)
+    shrinkBuiltin (I i)         = map I (shrinkBuiltin i)
+    shrinkBuiltin (B b)         = I 0 : map B (shrinkBuiltin b)
 
-    shrink (Constr i ds) = ds ++ map (Constr i) (shrink ds)
```
effectfully, line 190: Values of built-in types should be shrunk with `shrinkBuiltin`, not `shrink`, so this was a minor bug in the shrinking machinery (that I introduced one PR ago).

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
```diff
@@ -48,10 +46,6 @@ addTmBind (TermBind _ _ (VarDecl _ x a) _) = Map.insert x a
 addTmBind (DatatypeBind _ dat)             = (Map.fromList (matchType dat : constrTypes dat) <>)
 addTmBind _                                = id
 
-shrinkConstant :: DefaultUni (Esc a) -> a -> [Term TyName Name DefaultUni DefaultFun ()]
-shrinkConstant uni x =
-    map (mkConstantOf () uni) $ bring (Proxy @ArbitraryBuiltin) uni $ shrinkBuiltin x
```
effectfully, line 53: I somehow managed to integrate shrinking of constants into the term shrinker the wrong way. It's now fixed and we no longer need this definition.

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
```diff
@@ -275,22 +269,34 @@ shrinkTypedTerm tyctx0 ctx0 (ty0, tm0) = concat
...
 
-        Apply _ fun arg | Right argTy <- inferTypeInContext tyctx ctx arg ->
-          -- Drop substerms
-          [(argTy, arg), (TyFun () argTy ty, fun)]
+        Apply _ fun arg -> case inferTypeInContext tyctx ctx arg of
+          Right argTy ->
+            -- Drop substerms
+            [ -- Prefer to keep the same type, but...
+              (ty, fixupTerm tyctx ctx tyctx ctx ty arg)
+            , (ty, fixupTerm tyctx ctx tyctx ctx ty fun)
```
effectfully, line 277: These two do allow us to get better shrinks. Makes me wonder how many other such opportunities we have unrealized...

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
```diff
@@ -275,22 +269,34 @@ shrinkTypedTerm tyctx0 ctx0 (ty0, tm0) = concat
...
           ]
 
-        -- Builtins can shrink to unit. More fine-grained shrinking is in `structural` below.
-        Constant _ val -> do
-            val'@(Some (ValueOf uni _)) <- shrink val
-            pure (mkTyBuiltinOf () uni, Constant () val')
-
-        _ -> []
+        -- TODO: allow non-structural shrinking for some of these.
+        Var{} -> []
```
effectfully, line 291: I expanded the underscore just to make it clear what kind of shrinks we're missing.

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
```diff
@@ -320,13 +326,18 @@ shrinkTypedTerm tyctx0 ctx0 (ty0, tm0) = concat
                   tyctxInner = foldr addTyBind tyctx binds
                   ctxInner   = foldr addTmBind ctx binds
 
-        -- TODO: shrink the function too!
         TyInst _ fun argTy -> case inferTypeInContext tyctx ctx fun of
-          Right (TyForall _ x k tyInner) ->
+          Right funTy@(TyForall _ x k tyInner) ->
```
effectfully, line 330: Fixed the TODO.

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
effectfully, line 1: Note to self: create a ticket about breadth-first shrinking, I feel like there's an easy way to get much more efficient shrinking. And we can have a reasonable API like in [`Successors`](https://hackage.haskell.org/package/successors-0.1.0.1/docs/Control-Applicative-Successors.html) instead of the QuickCheck footgun.

plutus-core/testlib/PlutusIR/Pass/Test.hs
```diff
@@ -89,7 +89,7 @@ testPassProp' ::
       -> Pass m tyname name PLC.DefaultUni PLC.DefaultFun a)
   -> Property
 testPassProp' ann before after pass =
-  forAllDoc "ty,tm" genTypeAndTerm_ (const []) $ \ (_ty, tm) ->
+  forAllDoc "ty,tm" genTypeAndTerm_ shrinkClosedTypedTerm $ \ (_ty, tm) ->
```
effectfully, line 92: Shrinking was disabled for the passes, I enabled it as it helps a lot.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
```diff
@@ -184,14 +183,18 @@ genDataFromSpine els = oneof
...
-instance Arbitrary Data where
-    arbitrary = arbitrary >>= genDataFromSpine
+instance ArbitraryBuiltin Data where
+    arbitraryBuiltin = arbitrary >>= genDataFromSpine
+
+    shrinkBuiltin (Constr i ds) = ds ++ map (Constr i) (shrinkBuiltin ds)
+    shrinkBuiltin (Map ps)      = map fst ps ++ map snd ps ++ map Map (shrinkBuiltin ps)
+    shrinkBuiltin (List ds)     = ds ++ map List (shrinkBuiltin ds)
+    shrinkBuiltin (I i)         = map I (shrinkBuiltin i)
+    shrinkBuiltin (B b)         = I 0 : map B (shrinkBuiltin b)
```
zliu41, line 193: The `I 0` is bizarre. `Constr` only shrinks to `Constr`, `I` only shrinks to `I`, and so on, but why does `B` shrink to `I`? Worth a comment.

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
zliu41: I don't understand this - what do they do?

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
zliu41: Does this exist elsewhere, or are we no longer shrinking constants?

effectfully: We now shrink constants in a better way here:

```haskell
        Constant _ val ->
          shrink val <&> \val'@(Some (ValueOf uni _)) ->
            (mkTyBuiltinOf () uni, Constant () val')
```

The new code directly uses all the fancy constant-shrinking machinery that can change kinds, shrink to `unit` etc.

I.e. I got rid of garbage in favor of doing things how they were supposed to be done from the start.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
effectfully: `Constr` doesn't only shrink to `Constr`! It's 

```haskell
shrinkBuiltin (Constr i ds) = ds ++ map (Constr i) (shrinkBuiltin ds)
```

i.e. `Constr i ds` can shrink to any `d` from `ds`.

But good point, I'll add a comment.

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
effectfully: Made it

```
            [ -- Prefer to keep the same type, so that we don't need to change types elsewhere,
              -- which may remove or cover up the culprit.
              (ty, fixupTerm tyctx ctx tyctx ctx ty arg)
            , (ty, fixupTerm tyctx ctx tyctx ctx ty fun)
            , -- But support type-changing shrinking as well in case it's 'fixupTerm' that removes
              -- or covers up the culprit.
              (argTy, arg)
            , (TyFun () argTy ty, fun)
            ]
```

Hopefully that's clearer.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
effectfully: Added the following comment:

```
    -- We assume that @I 0@ is the smallest 'Data' object, hence it needs to be reachable from all
    -- nodes apart from @I 0@ itself. For recursive nodes ('Constr', 'Map', and 'List') we do not
    -- need to mention @I 0@, since all of those get shrunk to smaller nodes and therefore do not
    -- contribute any "unreachability". I.e. we only need to handle the base cases, 'I' and 'B'.
    -- The clause for @I i@ doesn't require mentioning @I 0@ explicitly, since we get it through
    -- general shrinking of @i@ (unless @i@ equals @0@, as desired). Hence the only places where
    -- @I 0@ has to appear explicitly is the clause for 'B'.
```

zliu41: > We assume that @I 0@ is the smallest 'Data' object, hence it needs to be reachable from all nodes apart from @I 0@ itself

Hmm, not sure why we assume that, since `B ""` or `List []` aren't any bigger than `I 0`. And you can't reach `I 0` from `List []`.

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
zliu41: I still don't understand what this is trying to do exactly. `ty` seems to be the type of `Apply fun arg`. Are we somehow "fixing up" `arg` so that it has type `ty`? How does that work?

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
effectfully: > Hmm, not sure why we assume that, since `B ""` or `List []` aren't any bigger than `I 0`.

Arbitrarily, just to have a final destination.

> `List []`

That's a very good point, thank you, my bad, will fix.

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
effectfully: > Are we somehow "fixing up" arg so that it has type ty?

That's the intention, yes.

> How does that work?

Badly it seems, I didn't get to improving it or even reviewing it properly, I think currently it just generates a minimal new term. There's also `fixupTerm_` that seems to be more clever, but I didn't review it either. Term generators and shrinkers were half-reviewed and almost zero-improved.

effectfully: Note that a function can be applied to an argument of the same type that the function returns. `f :: Integer -> Integer` for example. So if `arg` type checks against `ty`, then I think we'll get an ideal shrink. But again, I didn't even get to review the whole thing properly.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
effectfully: > Arbitrarily, just to have a final destination.

We don't need a single final destination for shrinking to work of course, it's just to have the least interesting `Data` object possible in the end and I arbitrarily decided that `I 0` is such. So that if you see anything else, you immediately start suspecting that this is where the problem is. It's not really important, just some icing on the cake.

zliu41 (approved):
Looks reasonable.

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
zliu41: Interesting. Are you sure the fixed up term isn't actually bigger than `Apply fun arg`? Otherwise we'd be shrinking in the wrong direction.

effectfully: > Interesting. Are you sure the fixed up term isn't actually bigger than `Apply fun arg`? Otherwise we'd be shrinking in the wrong direction.

No, but at least we have a test that is supposed to detect shrinking loops. We discussed it a lot back when PIR generators were getting reviewed and I did ask Max for a definitive Note multiple times (e.g. [here](https://github.com/IntersectMBO/plutus/pull/4559#issuecomment-1240746814)), but we decided to do it as a follow-up after [this](https://github.com/IntersectMBO/plutus/pull/4559#issuecomment-1241882441) response and as usual the follow-up never happened.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
effectfully: Made it that:

```haskell
    -- We arbitrarily assume that @I 0@ is the smallest 'Data' object just so that anything else in
    -- a counterexample gives us a clue as to what the culprit may be. Hence @I 0@ needs to be
    -- reachable from all nodes apart from @I 0@ itself. For all nodes but 'I' we achieve this by
    -- returning @[I 0]@ if there's no other way to shrink the value, i.e. either shrinking keeps
    -- going or it's time to return the smallest object. The clause for @I i@ doesn't require
    -- mentioning @I 0@ explicitly, since we get it through general shrinking of @i@ (unless @i@
    -- equals @0@, as desired).
    shrinkBuiltin (Constr i ds) = pureIfNull (I 0) $ concat
        [ ds
        , map (Constr i) $ shrinkBuiltin ds
        , map (flip Constr ds) $ shrinkBuiltin i
        ]
    shrinkBuiltin (Map ps) = pureIfNull (I 0) $ concat
        [ map fst ps
        , map snd ps
        , map Map $ shrinkBuiltin ps
        ]
    shrinkBuiltin (List ds) = pureIfNull (I 0) $ concat
        [ ds
        , map List $ shrinkBuiltin ds
        ]
    shrinkBuiltin (B b) = pureIfNull (I 0) . map B $ shrinkBuiltin b
    shrinkBuiltin (I i) = map I $ shrinkBuiltin i
```

Gonna merge the PR once CI is green.

effectfully: ^ which also added shrinking of `i` in `Constr`, 'cause why not.

PR #5730, opened 2024-01-18
Refactoring of Agda Main, and addition of Counting Mode.

mjaskelioff:
This PRs cleans up the `Main.lagda` file by moving functions into two new files: `Evaluator/Program.lagda.md` (used by `plc-agda` to run programs) and `Evaluator/Term.lagda.md`, used by Haskell programs to run terms in Agda. 

It adds a new function to be called from Haskell `runUCostingAgda`, which takes a `CostModel` and a UPLC term and returns a tuple consisting of the result of executing the CEK machine, the CPU cost (as an `Integer`) and memory costs (as an `Integer`). The cost model is a pair `(CekMachineCosts , BuiltinCostMap)` of machine parameters (for example `defaultCekMachineCosts`) and a map `[(Text, CpuAndMemoryModel)]` from builtin names to its cost model. This map can me obtained by calling the `getJSONModel` function in the `Cost/JSON.hs` file.

Because of the restructuring of the files, Haskell programs which wish to call `runUAgda` should now import `MAlonzo.Code.Evaluator.Term`.

PR #5747, opened 2024-01-28
Fix missing Agda import

kwxm:
This is just fixing an import that became outdated after merging #5730: see the comments there for more details.  I'll auto-merge it to get `master` working again.


kwxm:
Gah. This failed after merging with 
> error: some substitutes for the outputs of derivation '/nix/store/3zw2qwjvqh8a0s1cnzb6mvms7vk99y2p-hadrian-source.drv' failed (usually happens due to networking issues); try '--fallback' to build derivation from source

I think the original problem should be fixed though, so I won't try to fix that.

PR #5754, opened 2024-01-31
[PLT-9395] Bugfix "PlutusTx.Ratio.reduce allows negative denominator"

### LINKED ISSUE #5722: `PlutusTx.Ratio.reduce` allows negative integer in the denominator
### Summary

Documentation on `PlutusTx.Ratio.denominator` says that the "[denominator] will always be greater than, or equal to, 1, although the type does not describe this." However, using `PlutusTx.Ratio.reduce`, it is possible to build negative rational with negative integer on the denominator. 

```hs
λ> import PlutusTx.Ratio
λ> denominator $ reduce 1 (-1)

```

### Actual Result

```
-1
```

### Expected Result

```
1
```

`PlutusTx.Ratio.reduce` can be deprecated entirely since `PlutusTx.Ratio.unsafeRatio` serves identical purpose while correctly taking care of negative integer on denominator input. 

Also, adding invariant notes to `PlutusTx.Ratio.Rational`'s haddock documentation would be helpful. As it is currently not included in haddock: https://github.com/IntersectMBO/plutus/blob/master/plutus-tx/src/PlutusTx/Ratio.hs#L69-L72

### System info

```
Linux Mars 6.1.41 #1-NixOS SMP PREEMPT_DYNAMIC Mon Jul 24 16:55:35 UTC 2023 x86_64 GNU/Linux
NixOS 23.11.20230727.db8672b (Tapir) x86_64
plutus-tx: 1.7.0.0
```

Unisay:
Closes #5722

Pre-submit checklist:
- Branch
- PR
    - [x] (For external contributions) Corresponding issue exists and is linked in the description
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-tx-plugin/test/StdLib/Spec.hs
+++ b/plutus-tx-plugin/test/StdLib/Spec.hs
@@ -16,7 +16,6 @@ import Control.Exception (SomeException, evaluate, try)
 import Control.Monad.IO.Class (MonadIO (liftIO))
 import Data.Ratio ((%))
 import GHC.Exts (fromString)
-import GHC.Real (reduce)
 import Hedgehog (MonadGen, Property)
 import Hedgehog qualified
 import Hedgehog.Gen qualified as Gen
@@ -39,7 +38,6 @@ import PlutusTx.Plugin (plc)
 import PlutusCore.Data qualified as PLC
 
 import Data.Proxy (Proxy (Proxy))
-import Data.Ratio qualified as GHCRatio
 import PlutusPrelude (reoption)
 
 roundPlc :: CompiledCode (Ratio.Rational -> Integer)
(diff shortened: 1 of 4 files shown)


plutus-tx-plugin/test/StdLib/Spec.hs
```diff
@@ -100,28 +97,6 @@ testQuotRem = Hedgehog.property $ do
     Hedgehog.annotateShow plutusResult
     Hedgehog.assert (ghcResult == plutusResult)
 
-testReduce :: Property
```
michaelpj, line 103: was this test important?

michaelpj (approved):

plutus-tx-plugin/test/StdLib/Spec.hs
Unisay: It was important, but not anymore:
This test was checking that `PlutusTx.Ratio.reduce` returns the same results as `GHC.Real.reduce` for the same arguments. However, after removing the `PlutusTx.Ratio.reduce` there is nothing left to test, so I dropped the test.

PR #5756, opened 2024-01-31
PLT-9238 Export Simple JSON model from plutus-core

mjaskelioff:
Adds two modules to plutus-core so that it can export a simple model of the cost parameters of builtins.
This is used from Agda, and the `print-cost-model` executable.

Pre-submit checklist:
- Branch
    - [X] Tests are provided (if possible)
    - [X] Changelog fragments have been written (if appropriate)
    - [X] Relevant tickets are mentioned in commit messages
    - [X] Formatting, PNG optimization, etc. are updated
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested


plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/SimpleBuiltinCostModel.hs
```diff
@@ -0,0 +1,33 @@
...
+import Data.Aeson.THReader (readJSONFromFile)
+import Data.Bifunctor
+import Data.Text (Text, replace)
+import PlutusCore.DataFilePaths qualified as DFP
+import PlutusCore.Evaluation.Machine.CostingFun.SimpleJSON
+
+type BuiltinCostMap = [(Text, CpuAndMemoryModel)]
+
+-- | The default builtin cost map.
+builtinCostKeyMap :: KeyMap.KeyMap CpuAndMemoryModel
```
michaelpj, line 24: I wonder if it would be cleaner to instead create this from the read-in Haskell version of the cost model that we already have, rather than having two independent JSON readers?

michaelpj: That would also mean we could make it from a `CostModelParams` by going through the full version.

kwxm:
> possibly can be used from other tools, like the cost printing executable (not done in this PR).

We probably don't want to do that.  I use that for converting JSON files into an easily-readable format, for instance for comparing models when the costing changes.  Usually I have actual JSON files to hand and I don't necessarily want to look at the version that `plutus` is building with.  There's a `-d` option that goes and finds that if you really do want it.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/SimpleBuiltinCostModel.hs
mjaskelioff: If you are suggesting how to convert from the Haskell model to the model I need, I wouldn't know how to do it other than the naive way of going through every possible builtin explicitly.

mjaskelioff:
> Usually I have actual JSON files to hand and I don't necessarily want to look at the version that `plutus` is building with. There's a `-d` option that goes and finds that if you really do want it.

I meant to use the parser in `plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/SimpleJSON.hs`, not the whole thing (which you'd only use for the `-d` option.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/SimpleBuiltinCostModel.hs
michaelpj: Hmmm. I guess that is basically what you'd have to do, or do some Generics magic. Annoying. But it feels bad duplicating the JSON format...

mjaskelioff: The JSON format was already duplicated in the print-cost-model executable. So if we unify the JSON parser in that executable and the one used here, we have no more duplication than before.

mjaskelioff:
I updated `print-cost-model` to use the `SimpleJSON` Parser. Note that I only had to import a module and erase the corresponding code, as the parser in `SimpleJSON` was copied from there.

mjaskelioff:
So now this PR does not increase the number of JSON parsers in our codebase :)

plutus-metatheory/src/Opts.hs
```diff
@@ -3,8 +3,6 @@
 
 module Opts where
 
-import Paths_plutus_metatheory (getDataDir)
```
kwxm, line 6: Can we remove `Paths_plutus_metatheory` from the cabal file now?

plutus-core/cost-model/print-cost-model/Main.hs
kwxm, line 1: There's some extra space before an `=` on line 254. Could you change that into a single space while you're here please?

kwxm (approved):
OK, this all looks good. Thanks for tidying things up!

plutus-metatheory/src/Opts.hs
mjaskelioff: Yes. done

plutus-core/cost-model/print-cost-model/Main.hs
mjaskelioff: Done

PR #5761, opened 2024-02-05
[EXPERIMENT] Plutus Tx codebase re-formatted with fourmolu.

Unisay:
Plutus Tx (plutus-tx, plutus-tx-plugin) Haskell sources have been re-formatted with fourmolu.


fourmolu.yaml
```diff
@@ -0,0 +1,16 @@
+indentation: 2
+column-limit: none
+function-arrows: trailing
+comma-style: leading
+import-export-style: diff-friendly
+indent-wheres: true
+record-brace-space: false
```
michaelpj, line 7:
```suggestion
record-brace-space: true
```
IMO

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -893,6 +903,7 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
     GHC.Type _ -> throwPlain $ UnsupportedError "Types as standalone expressions"
     GHC.Coercion _ -> throwPlain $ UnsupportedError "Coercions as expressions"
 
+{- ORMOLU_DISABLE -}
```
michaelpj, line 906: why?

plutus-tx-plugin/src/PlutusTx/Compiler/Kind.hs
```diff
@@ -16,6 +16,7 @@ import GHC.Plugins qualified as GHC
 
 import PlutusCore qualified as PLC
 
+{- ORMOLU_DISABLE -}
```
michaelpj, line 19: why?

michaelpj (commented):
I generally like it, but
- I would like to see it on plutus-core
- I am suspicious of your ORMOLU_DISABLEs. Each of these is presumably an interesting case where it goes wrong, why not put a comment in explaining why (otherwise we're guessing, forever). Ideally with a link to the upstream ticket..

zliu41:
What flags did you use? I suppose at least `-o -XImportQualifiedPost -i --indentation=2`? I also find `--indent-wheres=true` to look better.

kwxm (commented):
This doesn't look too bad: having fewer spaces at the start of the lines would probably be an improvement.  I've added a couple of comments about things I like about the way the code's formatted at the moment.  I couldn't see any good examples in these files, but another thing I quite like about the current style is that it allows you to format multi-line signatures like [this](https://github.com/IntersectMBO/plutus/blob/9f6b72f3ea674fa033e6fc296a5d9336f272341f/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs#L599):
```
runCekM
    :: forall a cost uni fun ann
    . ThrowableBuiltins uni fun
    => MachineParameters CekMachineCosts fun (CekValue uni fun ann)
    -> ExBudgetMode cost uni fun
    -> EmitterMode uni fun
    -> (forall s. GivenCekReqs uni fun ann s => CekM uni fun s a)
    -> (Either (CekEvaluationException NamedDeBruijn uni fun) a, cost, [Text])
    ```
That keeps all of the noise at the start of the lines, where you can  pretty much ignore it.  It looks like fourmolou will move the arrows to the ends of the lines, which I think might be less readable.  Maybe that's configurable though.
    
I don't have any particularly strong opinions though: changes like this can be a bit annoying initially, but you get used to them.

File: plutus-tx/src/PlutusTx/Lift/TH.hs [hunk 4]
```diff
@@ -64,29 +65,30 @@ import Prelude as Haskell
...
+  | LiftMissingVar !TH.Name
+  deriving anyclass (Prelude.Exception)
 
 instance PP.Pretty LiftError where
-    pretty (UnsupportedLiftType t) = "Unsupported lift type: " PP.<+> PP.viaShow t
-    pretty (UnsupportedLiftKind t) = "Unsupported lift kind: " PP.<+> PP.viaShow t
-    pretty (UserLiftError t)       = PP.pretty t
-    pretty (LiftMissingDataCons n) = "Constructors not created for type: " PP.<+> PP.viaShow n
-    pretty (LiftMissingVar n)      = "Unknown local variable: " PP.<+> PP.viaShow n
+  pretty (UnsupportedLiftType t) = "Unsupported lift type: " PP.<+> PP.viaShow t
```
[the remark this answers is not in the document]
michaelpj: Yeah, being inconsistent about alignment is one of the annoyances with our style. I think ormolu/fourmolu tends to avoid alignment in the interest of diff-friendliness

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
```diff
@@ -1,22 +1,23 @@
 -- editorconfig-checker-disable-file
-{-# LANGUAGE ConstraintKinds       #-}
-{-# LANGUAGE DataKinds             #-}
-{-# LANGUAGE FlexibleContexts      #-}
-{-# LANGUAGE LambdaCase            #-}
-{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE ConstraintKinds #-}
```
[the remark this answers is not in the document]
Unisay: While having them aligned is more aesthetically pleasing, its also less diff-friendly: if I add a new LANGUAGE pragma that differs in length then it triggers the change in other lines too as they have to be adjusted.

plutus-tx-plugin/src/PlutusTx/Compiler/Type.hs
```diff
@@ -74,6 +74,7 @@ Generally, we need to call this whenever we are compiling a "new" type from the
 If we are compiling a part of a type we are already processing then it has likely been
 normalized and we can just use 'compileType'
 -}
+{- ORMOLU_DISABLE -}
```
effectfully, line 77: Why?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
Unisay: It was my quick way of working around this error:
```
$ fourmolu -i plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
Loaded config from: /home/yura/projects/cardano/plutus/fourmolu.yaml
plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs:992:25-26
  The GHC parser (in Haddock mode) failed:
  [GHC-58481] parse error on input `<-'
```
When I removed the CPP conditional compilation block the problem went away. 
I'll try to update `fourmolu` to the latest version (14.1) to see if the problem is already solved.

effectfully (commented):
I was half-way through, then the diff changed and now I'm lost.

plutus-tx-plugin/test/Plugin/Data/Spec.hs
```diff
@@ -1,20 +1,70 @@
...
+of them.
+-}
+-- See Note [Non-regular data types in tests].
+-- TODO: replace this with 'first' when we have working recursive functions
+-- See Note [Non-regular data types in tests].
+-- The type signatures are needed due to a bug (see 'emptyRoseNewId')
+-- See Note [Non-regular data types in tests].
+-- Note that the parameters get swapped.
+-- Despite the type family being applied to a parameterized type we can still reduce it
+-- Here we cannot reduce the type family
```
effectfully, line 58: Lol, did it just move all the comments in this module to the beginning of the module? Wonderful.

plutus-tx-plugin/test/Plugin/Data/Spec.hs
```diff
@@ -370,31 +400,28 @@ basicOpen :: CompiledCode (BasicOpen Bool -> BasicOpen Bool)
...
-paramId :: forall a . Param a -> AType (Param a) -> AType (Param a)
+paramId :: forall a. Param a -> AType (Param a) -> AType (Param a)
 paramId _ x = x
 
 associatedParam :: CompiledCode Integer
 associatedParam = plc (Proxy @"associatedParam") (paramId (Param 1) 1)
 
--- Here we cannot reduce the type family
 {-# NOINLINE tfId #-}
-{-# OPTIONS_GHC -Wno-incomplete-patterns #-}
```
effectfully, line 396: I guess this is what confused it.

effectfully (commented):
All the same bugs that we've discussed over the years, pretty much nothing got fixed, maybe even a few new ones got added. Or is it just the old version of the tool?

Anyway, I hate it, it's stupid, disgusting and broken, but I can still tolerate it, particularly if I'm allowed to sprinkle `ORMOLU_DISABLE` whenever it screws something up unintentionally (as opposed to doing that intentionally).

plutus-tx-plugin/test/Plugin/Data/Spec.hs
kwxm: > Lol, did it just move all the comments in this module to the beginning of the module? Wonderful.

:laughing: That's actually pretty impressive.  It's taken most (but not quite all) of the comments and put them [between two `OPTIONS_GHC` pragmas](https://github.com/IntersectMBO/plutus/blob/2dfc62146e729d24d6abe6d729cbc00468df5b5a/plutus-tx-plugin/test/Plugin/Data/Spec.hs#L9) at the top of the file!

kwxm:
Here's an [interesting comment](https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) about code formatting in Python where they quote Donald Knuth on why it makes sense to put symbols at the start of a line (from [this discussion](https://www.reddit.com/r/haskell/comments/khqs1f/silly_haskell_code_formatting_is_prevalent/) on Haskell formatting). It hadn't occurred to me before, but that's how things get formatted in displayed mathematical text.  I don't mind too much in this case though.

plutus-tx/src/PlutusTx/Lift/TH.hs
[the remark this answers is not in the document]
kwxm: > I think ormolu/fourmolu tends to avoid alignment in the interest of diff-friendliness

I think we had a discussion like this before and I accused you of putting the convenience of machines before the convenience of human beings.  However, I suppose that if I really want to align thing to look at them I can get emacs to do it temporarily and let the formatter change it back if it wants to.

Unisay:
The experiment is over

PR #5763, opened 2024-02-07
[Bug] [TypeCheck] Fix an off-by-maxBound error

effectfully:
This fixes the bug and adds a simple golden test for it, I want to have a better testing strategy, I'll create an issue about it, but for now let's just fix the bug.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/TypeCheck/Internal.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/TypeCheck/Internal.hs
@@ -523,11 +523,13 @@ inferTypeM t@(Constr ann resTy i args) = do
 
     -- We don't know exactly what to expect, we only know what the i-th sum should look like, so we
     -- assert that we should have some types in the sum up to there, and then the known product type.
-    let prodPrefix  = map (\j -> "prod_" <> Text.pack (show j)) [0 .. i - 1]
+    let -- 'toInteger' is necessary, because @i@ is a @Word64@ and therefore @i - 1@ would be
+        -- @maxBound :: Word64@ for @i = 0@ if we didn't have 'toInteger'.
+        prodPrefix  = map (\j -> "prod_" <> Text.pack (show j)) [0 .. toInteger i - 1]
         fields      = map (\k -> "field_" <> Text.pack (show k)) [0 .. length args - 1]
         prod_i      = "[" <> Text.intercalate " " fields <> "]"
-        shape       = "sop " <> foldMap (<> " ") prodPrefix <> prod_i <> " ... prod_n"
-        vars        = prodPrefix ++ fields ++ ["prod_n"]
+        shape       = "sop " <> foldMap (<> " ") prodPrefix <> prod_i <> " ..."
+        vars        = prodPrefix ++ fields
         expectedSop = ExpectedShape shape vars
     case unNormalized vResTy of
         TySOP _ vSTys -> case vSTys ^? wix i of
--- a/plutus-core/prelude/PlutusPrelude.hs
+++ b/plutus-core/prelude/PlutusPrelude.hs
@@ -239,7 +239,6 @@ showText = T.pack . show
 -- the same length.
 zipExact :: [a] -> [b] -> Maybe [(a,b)]
 zipExact [] []         = Just []
-zipExact [a] [b]       = Just [(a,b)]
 zipExact (a:as) (b:bs) = (:) (a, b) <$> zipExact as bs
 zipExact _ _           = Nothing
 
--- a/plutus-core/plutus-core/test/type-errors/constrTooFew.plc
+++ b/plutus-core/plutus-core/test/type-errors/constrTooFew.plc
@@ -1,0 +1,3 @@
+(program 1.1.0
+   (constr (sop [(con integer)] []) 0)
+)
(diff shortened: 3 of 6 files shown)


plutus-core/prelude/PlutusPrelude.hs
```diff
@@ -239,7 +239,6 @@ showText = T.pack . show
 -- the same length.
 zipExact :: [a] -> [b] -> Maybe [(a,b)]
 zipExact [] []         = Just []
-zipExact [a] [b]       = Just [(a,b)]
 zipExact (a:as) (b:bs) = (:) (a, b) <$> zipExact as bs
 zipExact _ _           = Nothing
```
effectfully, line 243: While I was here.

michaelpj (approved):
Good find!

plutus-core/plutus-core/src/PlutusCore/TypeCheck/Internal.hs
```diff
@@ -523,7 +523,9 @@ inferTypeM t@(Constr ann resTy i args) = do
 
     -- We don't know exactly what to expect, we only know what the i-th sum should look like, so we
     -- assert that we should have some types in the sum up to there, and then the known product type.
-    let prodPrefix  = map (\j -> "prod_" <> Text.pack (show j)) [0 .. i - 1]
+    let -- 'toInteger' is necessary, because @i@ is a @Word64@ and therefore @i - 1@ would be
+        -- @maxBound :: Word64@ for @i = 0@ if we didn't have 'toInteger'.
+        prodPrefix  = map (\j -> "prod_" <> Text.pack (show j)) [0 .. toInteger i - 1]
```
kwxm, line 528: Oh, that's pretty impressive.  I was expecting some subtle error in the typechecking algorithm, but it's just the _error message_.  That's an easy mistake to make: I'd forgotten that Word64 wraps round at the bounds.

kwxm (approved):
Well done for finding this so quickly.

kwxm:
While I'm here, I'll do some nitpicking about the rendering of the types and the error messages.  I was experimenting with sops and found them quite confusing.

1.  You get error messages like
```
Expected a term of type
  '(sop prod_0 prod_1 [field_0 field_1 field_2] ... prod_n

  )' for some 'prod_0', 'prod_1', 'field_0', 'field_1', 'field_2' and 'prod_n'
```
That kind of looks as if it should carry on with more `field` stuff after the first one, so maybe something like 
```
sop prod_0 prod_1 [field_0 field_1 field_2] prod_3... prod_n
```
would be more informative.  But then maybe that looks as if you're _required_ to have more prods afterwards, which isn't the case.

2.  The error messages are maybe a bit  misleading: 
```
   (constr (sop [(con bool)]) 1 (con unit ()))
```
gives you
```plc: Type mismatch at ()
Expected a term of type
  '(sop prod_0 [field_0] ... prod_n)' for some 'prod_0', 'field_0' and 'prod_n'
But found one of type
  '(sop [(con bool)])'
Namely,
  '(constr (sop [(con bool)]) 1 (con unit ()))'
```
That's a bit misleading: it says the term has a certain type, but it's malformed so maybe it doesn't have a type at all.

3.  There are types that get rendered like this
```
    (sop [(con bool)] [(con integer) (con string)])
```
It took me a while to work out what that meant because elsewhere `[(con integer) (con string)]` would represent a type application.  I suppose that's to fit in with the syntax for the arguments of `constr` though.

I'm just complaining for the sake of it here (although I'm maybe a bit worried because at some point we'll have to specify these types).  The types are rather complicated here and I don't have any sensible suggestions about how to clarify things.

effectfully:
> `sop prod_0 prod_1 [field_0 field_1 field_2] prod_3... prod_n`
>
> would be more informative. But then maybe that looks as if you're _required_ to have more prods afterwards, which isn't the case.

Yeah, I think that would make it look as if `prod_3` was required. To me the current error message is OK, but maybe we could change the type from

```
sop prod_0 prod_1 [field_0 field_1 field_2] ... prod_n
```

to e.g.

```
sop [...] [...] [field_0 field_1 field_2] ... [...]
```

> That's a bit strange: it says the term has a certain type, but it's malformed so maybe it doesn't have a type at all.

Agree, this one is confusing. I don't feel like worrying about it too much, but I agree it would be great to improve it.

> There are types that get rendered like this

I find the syntax for SOPs confusing too.

> The fact that you can have things with totally different types in TPLC which erase to the same `constr` in UPLC is a bit unpleasant as well.

I think it's by design? Same with lambdas for example.

kwxm:
```
sop [...] [...] [field_0 field_1 field_2] ... [...]
```

OK, that's quite a bit simpler: maybe that's a good idea.  Anyway, don't worry too much.  I was just complaining while I had the chance because I had trouble understanding what was going on.  What I should really have done was to look at the original PR properly.

kwxm:
> > > `sop [...] [...] [field_0 field_1 field_2] ... [...]`
> > > OK, that's quite a bit simpler: maybe that's a good idea.
> 
> Just tried it and I think while it's OK for this example, on a smaller one it looks really weird: `sop [] ... [...]`. I decided to remove `prod_n` / `[...]` entirely and keep the old formatting so that I can refer to the prefix `[...]` as actual (meta)variables that need to be instantiated.
> 
> I agree that the error messages are weird and cryptic and annoying, but I'm going to leave them at that, otherwise I'll spend much more time on the issue without making any meaningful progress.

:+1:  Yes, please just go ahead and merge it.

PR #5766, opened 2024-02-09
Inline constants by default (turned off in conservative mode)

ana-pantilie:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -107,7 +109,10 @@ applyAndBetaReduce rhs args0 = do
         -- the body `a` will be beta reduced in
         Term tyname name uni fun ann ->
         InlineM tyname name uni fun ann Bool
-      safeToBetaReduce = shouldUnconditionallyInline Strict
+      safeToBetaReduce arg term =
+        -- Does not inline constants because term size might increase
+        -- and other inlining optimisations will not get applied.
+        local (iiInlineConstants .~ False) . shouldUnconditionallyInline Strict arg term
```
ana-pantilie, line 115: Turning it back to `True` will actually _decrease_ the size of the term in all the tests which fail.

plutus-benchmark/nofib/test/9.6/queens4-bt.eval.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 8028768773
-| mem: 28840930})
\ No newline at end of file
+({cpu: 8028699773
+| mem: 28840630})
```
ana-pantilie, line 2: I've looked at all of these and it _seems_ like these all decrease. Unfortunately dealing with this sort of output is very error-prone, so I opened https://input-output.atlassian.net/browse/PLT-9459?atlOrigin=eyJpIjoiNGM1YTdlNWUyNjhhNGE4OGE3MTUzNmI4OTYwMzQ2MzgiLCJwIjoiaiJ9.

plutus-core/plutus-ir/test/PlutusIR/Transform/Inline/nameCapture.golden
```diff
@@ -1,15 +1,15 @@
...
+    (vardecl c (con integer))
+    [ [ (builtin addInteger) (con integer 10) ] (con integer 10) ]
   )
+  [
+    [
+      (builtin addInteger)
+      [ [ (builtin addInteger) (con integer 10) ] (con integer 7) ]
+    ]
+    (con integer 7)
+  ]
```
ana-pantilie, line 14: How was this right before this change? I'm looking at the test description and code in `testCapture` and the previous output doesn't seem right, wouldn't it evaluate to `7+7+7`?

plutus-tx-plugin/test/Budget/9.6/show.pir.golden
```diff
@@ -324,16 +318,13 @@ let
...
   !c : integer
     = trace
         {integer}
         (concatBuiltinStrings
-           ((let
-                !len : integer = lengthOfByteString s
-              in
-              go (`$fEnumBool_$cenumFromTo` 0 (subtractInteger len 1)))
-              (Nil {string})))
+           (go (`$fEnumBool_$cenumFromTo` 0 17) (Nil {string})))
```
ana-pantilie, line 327: This seems right, can we assume that the (huge) `uplc` version below is fine as well? :smile:

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
ana-pantilie: Actually, this doesn't work, and I can't think of a nice and quick way to implement this. The idea was that since callsite inlining _requires_ that the resulting term is smaller in size, having constant inlining turned on by default might not allow the callsite inlining to take place (if my reading of the code is correct).

Now, it seems like our test suites don't contain any examples where constant inlining increases program size. Can we be sure that the test suite reflects actual real world behavior? Should we try to add some more test cases?

ana-pantilie:
It seems like our test suites don't contain any examples where constant inlining increases program size. Can we be sure that the test suite reflects actual real world behavior?

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
zliu41: > having constant inlining turned on by default might not allow the callsite inlining to take place

I don't think it can. The size is measured by `termSize`, which is the number of AST nodes, and is only an approximation of the actual script size. In particular, a constant and a variable are both one node, even if it's a big constant. While it is not precise, it means inlining a constant cannot increase `termSize`.

zliu41 (commented):
> It seems like our test suites don't contain any examples where constant inlining increases program size. Can we be sure that the test suite reflects actual real world behavior?

Unfortunately we can't currently measure the precise effect of this, because the size measure is `termSize` which just returns the number of AST nodes, and inlining constant cannot increase `termSize`. So this is expected.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
ana-pantilie: Right, then the results make sense. Thank you.

ana-pantilie:
@zliu41 this is ready for review.

plutus-tx-plugin/src/PlutusTx/Options.hs
```diff
@@ -169,6 +170,7 @@ pluginOptions =
                     , Implication (== False) posRelaxedFloatin True
                     , Implication (== False) posPreserveLogging False
                     , Implication (== False) posCaseOfCaseConservative False
+                    , Implication (== False) posInlineConstants False
```
zliu41, line 173: This is "conservative-optimization == False implies inline-constants == False", but it should be the other way around. Also there should be another entry for `Implication (== True)`.

zliu41 (approved):
Just need a changelog entry

ana-pantilie:
> Just need a changelog entry

Oops, forgot to push that last night. Thanks for approving!

plutus-tx-plugin/test/Budget/9.6/toFromData.pir.golden
```diff
@@ -68,8 +58,8 @@ let
                         (Bool_match
                            arg
                            {all dead. data}
-                           (/\dead -> constrData 1 (mkNilData unitval))
-                           (/\dead -> constrData 0 (mkNilData unitval))
+                           (/\dead -> Constr 1 [])
```
michaelpj, line 61: nice

PR #5767, opened 2024-02-10
Add Note [Short-circuit evaluation for `&&` and `||`]

zliu41:
The Note below it, `Note [Lazy patterns on function parameters]`, talks about this a little bit, but it focuses on explaining the lazy pattern. The short-circuiting itself is important enough and deserves a separate Note that explains it in more details.


plutus-tx/src/PlutusTx/Bool.hs
```diff
@@ -38,6 +38,42 @@ infixr 2 ||
 not :: Bool -> Bool
 not a = if a then False else True
 
+{- Note [Short-circuit evaluation for `&&` and `||`]
+TLDR: as it stands, the behaviors of `&&` and `||` are inconsistent: they sometimes
+short-circuit, sometimes don't.
+
+The Plutus Tx compiler treats `if-then-else` and `case` specially, and ensures that
+the branches aren’t strictly evaluated. For example, `if b then exprT else exprF` is
+compiled to `(if b then (/\a -> exprT) (/\a -> exprF)) @(forall a. a)`. The type
```
kwxm, line 47: I've forgotten why we do this.  Is it because we can't use `()` because that doesn't necessarily exist in the current universe of built-in types and doing it with Scott units would introduce extra expense?  I was thinking that this seems like a bit of a hack and I don't recall having seen `/\` used like this before, but now I remember that we spent a long time (weeks/months) wondering whether `/\`should block evaluation or not and we eventually decided that it should.  I don't think there's consensus in the literature about whether evaluation should be allowed under unapplied `/\`.  [That doesn't have a lot to do with this PR: I'm just trying to refresh my memory about this stuff.]

kwxm (approved):
I'm surprised we didn't have a note about this before!  

I'm sure we must have talked about this before, but if we have I don't remember.  Can we discourage the use of `&&` and `||`?    Does the pre-inliner pay special attention to `&&` and `||` because they're very basic operations, and could we introduce replacements (`andalso`and `orelse` or something: see SML and C#, and apparently also VB) that are guaranteed to short-circuit in all cases because the pre-inliner doesn't know about them?  Maybe that wouldn't be too helpful: it'd be difficult to remember not to use the standard operations.

plutus-tx/src/PlutusTx/Bool.hs
```diff
@@ -38,6 +38,42 @@ infixr 2 ||
...
+
+If GHC's pre-inliner inlines `&&` in `exprL && exprR`, it produces
+`if exprL then exprR else False`, enabling short-circuit evaluation (the second
+argument of `&&` must not be strict for this to happen - see
+Note [Lazy patterns on function parameters]).
+
+On the other hand, if the pre-inliner does not inline `&&`, then it does not
+short-circuit. Usually, the pre-inliner inlines `&&` if it occurs once, and
+does not if it occurs multiple times, such as in a program like
+`\x y z -> x < 3 && y < 4 && z < 5`.
```
ana-pantilie, line 74: Nice write-up, I agree it makes practical sense to want `&&` and `||` to have short-circuiting semantics.

How is `if .. then .. else` treated specially? I understand how `case`  is treated specially, because `case` is part of the language, but I haven't seen `if .. then .. else`. Is it a builtin? Does it make sense to define a builtin without strict evaluation of its arguments? In any case, maybe we could we treat `&&` and `||` in the same way?

Sorry for all the questions, I just want to make sure I understand the details here.

ana-pantilie: I started working on adding tests for this, and I think that this doesn't work as expected.

If `&&` short-circuits, then `False && error` should behave the same as `case False of {False -> False} {_ -> error}`, right?

```
case False of
  False -> False
  _ -> error ()

||
PIR
||
v

let
  data Bool | Bool_match where
    True : Bool
    False : Bool
in
False

||
UPLC
||
v

program 1.1.0 (constr 1 [])

||
eval
||
v

(constr 1 [])
```

The `case` evaluates, as expected, to `False`, and the error branch isn't evaluated. However, the `&&` program:

```
False && error ()

||
PIR
||
v

let
  data Bool | Bool_match where
    True : Bool
    False : Bool
  data Unit | Unit_match where
    Unit : Unit
  !x : Bool = error {Bool}  -- note that we get this binding here; why??
in
False

||
UPLC
||
v

program 1.1.0 ((\x -> constr 1 []) error)

||
eval
||
v

error
```

As noted above, the branch we expect to get short-circuited gets bound to a variable `x`, which eventually forces its evaluation.

zliu41: > How is if .. then .. else treated specially? I understand how case is treated specially, because case is part of the language, but I haven't seen if .. then .. else. 

Very good question. I don't think we actually treat `if-then-else` differently (at least not any more), since it is nothing but `case` on `Bool`. I'll update the Note.

kwxm: > How is `if .. then .. else` treated specially? I understand how `case` is treated specially, because `case` is part of the language, but I haven't seen `if .. then .. else`. Is it a builtin?

I was wondering about this myself, and it turns out to be surprisingly complicated.  We do have a built-in `ifThenElse` function, but that doesn't always get used: sometimes the standard Church lambda-calculus encoding is used instead.  For example, if you have 
```
fff :: Integer -> Integer -> Integer -> Integer
fff a b c = if a == 0 then b else c
```
then it gets compiled to
```
(lam a (lam b (lam c
 (force
  [
   [
    [(force (builtin ifThenElse)) [ [ (builtin equalsInteger) (con integer 0) ] a ] ]
    (delay b)
   ]
  (delay c)
  ]
))))
```

but
```
ggg :: Bool -> Integer -> Integer -> Integer
ggg a b c = if a then b else c
```
gets compiled to
```
(lam a (lam b (lam c (force (case a (delay b) (delay c))))))
```

The Haskell `if ... then ... else ...` gets desugared to `case ... of True -> ... False -> ...` in [GHC Core](https://downloads.haskell.org/ghc/9.0.1/docs/html/libraries/ghc-9.0.1/GHC-Core.html), so the plugin never sees it.  I think that what's happening is that everything initially gets converted into PIR using the Church version and then that sometimes gets converted into the form using `ifThenElse` when the PIR is simplified.  I'm having some difficulty working out  where that happens though.

zliu41:
> Can we discourage the use of && and ||?

Yes, I think we should

> Does the pre-inliner pay special attention to && and || because they're very basic operations

They are likely inlined if occurred once, but the pre-inliner won't inline them if they occur multiple times

> could we introduce replacements (andalsoand orelse or something: see SML and C#, and apparently also VB) that are guaranteed to short-circuit in all cases because the pre-inliner doesn't know about them?

Potentially yes. We can either treat `&&` and `||` as magic functions, or use some kind of source annotations to tell the compiler that certain functions should be handled specially. The latter method is more general since users would be able to define their own versions of `&&` and `||`.

plutus-tx/src/PlutusTx/Bool.hs
zliu41: I don't remember exactly either; it could be that this is just easier to implement - all type abstractions and instantiations are simply compiled to `delay` and `force`. If we use `()` then it would have to be treated specially.

kwxm:
> I think that what's happening is that everything initially gets converted into PIR using the Church version and then that sometimes gets converted into the form using ifThenElse when the PIR is simplified. I'm having some difficulty working out where that happens though.

Nope, I was completely wrong about that.  In
```
fff :: Integer -> Integer -> Integer -> Integer
fff a b c = if a == 0 then b else c
```
then condition `a==0` gets compiled to a call to the builtin `equalsInteger`, which returns a value `x` of type [BuiltinBool](https://github.com/IntersectMBO/plutus/blob/08fe5f3f75f5046add34aa20e1b4bc6f0337c6c9/plutus-tx/src/PlutusTx/Builtins/Internal.hs#L99), and that gets [converted](https://github.com/IntersectMBO/plutus/blob/08fe5f3f75f5046add34aa20e1b4bc6f0337c6c9/plutus-tx/src/PlutusTx/Builtins/Class.hs#L49) into `ifThenElse x True False` in the PIR, where `True` and `False` are now PIR variables, and then there's a`match` round those, and it's eventually simplified away to a fairly direct call to `ifThenElse`.

 In the case of
 ```
ggg :: Bool -> Integer -> Integer -> Integer
ggg a b c = if a then b else c
```
`a` is a value of_Haskell's_ `Bool` type, which is a datatype with two constructors.  Datatypes get compiled using the Scott encoding, and `if` now gets converted into the standard Church/Scott thing.


Thanks to @zliu41 for the explanation.

File: plutus-tx/src/PlutusTx/Bool.hs (hunk 1 above)
Comment (effectfully) [reply plutus-tx/src/PlutusTx/Bool.hs:47]: > I've forgotten why we do this. Is it because we can't use `()` because that doesn't necessarily exist in the current universe of built-in types and doing it with Scott units would introduce extra expense?

`()` would require us to use a `\` which binds a useless name, which is less efficient than `/\` that compiles to a `delay`. I.e. basically we prefer `delay`-`force` to `lambda`-`apply`, since the former is both simpler and faster.

> `/\` should block evaluation or not and we eventually decided that it should.

I think that happened in [this](https://groups.google.com/a/iohk.io/g/plutus/c/GXPwl6o3xKE/m/HlxkO3hSBQAJ) Plutus mailing list thread.

File: plutus-tx/src/PlutusTx/Bool.hs (hunk 2 above)
Comment (effectfully) [reply plutus-tx/src/PlutusTx/Bool.hs:74]: Man this is a mess. Perhaps worth reflecting this discussion in the Note itself.

**Review by effectfully — COMMENTED** [review]
Great stuff, thanks for writing it up.

**Comment by effectfully** [comment]
> Datatypes get compiled using the Scott encoding

Why not SOP? Does it use an old version of the compiler or do we have Scott encoding still floating around in places? (we do: in the compilation code for recursive functions for example).

**Comment by kwxm** [comment]
> > Datatypes get compiled using the Scott encoding
> 
> Why not SOP? Does it use an old version of the compiler or do we have Scott encoding still floating around in places? (we do: in the compilation code for recursive functions for example).

I don't know.  Maybe I have to use a pragma to tell it whether the version is 1.0.0 or 1.1.0.  I'll look into that.

**Comment by kwxm** [comment]
> Maybe I have to use a pragma to tell it whether the version is 1.0.0 or 1.1.0. I'll look into that.

Yep.  If I compile 
```
ggg :: Bool -> Integer -> Integer -> Integer
ggg a b c = if a then b else c
```
then with `{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:target-version=1.0.0 #-}` (or nothing) I get
```
(lam a (lam b (lam c (force [ [ (force a) (delay b) ] (delay c) ])))
)
```
but with `1.1.0` I get
```
(lam a (lam b (lam c (force (case a (delay b) (delay c))))))
```

This probably means that all of the `nofib`benchmarks are compiled with 1.0.0.  Maybe we want to change that?

effectfully:
> This probably means that all of the `nofib` benchmarks are compiled with 1.0.0. Maybe we want to change that?

Oh god, I believe we certainly do. Thank you for digging into it.

zliu41:
Oh boy, I realized that I was totally wrong in saying that the behaviors of `&&` and `||` are at the mercy of the GHC pre-inliner. It doesn't concern the pre-inliner at all since `&&` and `||` are imported identifiers.

The real problem is with how the module using `&&` is compiled:
- If the module using `&&` is compiled with GHC optimizations, and it is an imported module (i.e., not the same module where we call `compile`), then GHC will compile this module first, and inlines `&&` (by the regular inliner, not pre-inliner). Therefore `&&` short-circuits.
- On the other hand, if the module using `&&` is compiled with `-O0 -fmax-simplifier-iterations=0 -fno-omit-interface-pragmas -fno-ignore-interface-pragmas`, or it is the same module where we call `compile`, then `&&` does _not_ get inlined, and it doesn't short-circuit.

This also gets to another question of whether we should compile any module with optimizations - because GHC will inline lots of stuff. Quite a mess.

@ana-pantilie ignore what I said in https://github.com/IntersectMBO/plutus/pull/5767#discussion_r1486221319.

plutus-tx/src/PlutusTx/Bool.hs
bezirg: Yes, i agree it's a mess. We have to trace where this exactly happens.

bezirg (approved):

bezirg:
I am relying a lot on chains of && . What shall I use instead?

ana-pantilie:
I added some tests as we discussed: https://github.com/IntersectMBO/plutus/pull/5774

I believe the results @zliu41 expected are visible in the tests.

ana-pantilie:
> This also gets to another question of whether we should compile any module with optimizations - because GHC will inline lots of stuff. Quite a mess.

Do we have any control over which GHC optimisation passes we allow? We should make sure that the transformations they apply are semantics-preserving from `plutus-core`'s point of view.

zliu41:
> I am relying a lot on chains of && . What shall I use instead?

zliu41:
> I am relying a lot on chains of && . What shall I use instead?

if-then-else

> Do we have any control over which GHC optimisation passes we allow? We should make sure that the transformations they apply are semantics-preserving from plutus-core's point of view.

We can (and I think we should) turn off most GHC optimizations, via `-O0 -fmax-simplifier-iterations=0`. However, not everything can be turned off, e.g., GHC runs a "simple optimizer" at the end of the desugaring phase, which I don't think can be turned off. Also, we currently rely on GHC's pre-inliner; see Note [GHC.sm_pre_inline].

PR #5772, opened 2024-02-14
A Plutus Blueprint data model

Unisay:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-tx/src/PlutusTx/Blueprint/Schema.hs
```diff
@@ -0,0 +1,417 @@
...
+import Data.Text.Encoding qualified as Text
+import Data.Typeable (Typeable, typeRep)
+import GHC.TypeLits qualified as GHC
+import Numeric.Natural (Natural)
+import Prelude hiding (max, min)
+
+data Schema
+  = SchemaInteger
+      -- | Title
+      (Maybe String)
```
zliu41, line 34: Suggest using a record type for each variant, and using newtypes when appropriate.

plutus-tx/src/PlutusTx/Blueprint/Schema.hs
```diff
@@ -0,0 +1,417 @@
...
+      -- | Description
+      (Maybe String)
+      -- | Comment
+      (Maybe String)
+      -- | Element schema
+      Schema
+  | SchemaOneOf (NonEmpty Schema)
+  | SchemaAnyOf (NonEmpty Schema)
```
zliu41, line 166: Please add Haddock explaining the difference between OneOf and AnyOf.

plutus-tx/src/PlutusTx/Blueprint/Schema.hs
```diff
@@ -0,0 +1,417 @@
...
+  | SchemaAllOf (NonEmpty Schema)
+  | SchemaNot Schema
```
zliu41, line 168: What does `SchemaNot` mean?

plutus-tx/src/PlutusTx/Blueprint/Schema.hs
```diff
@@ -0,0 +1,417 @@
...
+      Aeson.object
+        $ catMaybes
+          [ Just $ "dataType" .= ty
+          , fmap ("title" .=) title
+          , fmap ("description" .=) description
+          , fmap ("$comment" .=) comment
+          ]
+
+type HasSchemaDefinition :: forall a. a -> [a] -> Constraint
+type family HasSchemaDefinition n xs where
```
zliu41, line 292: This is another thing that could live in a different PR. Without seeing how it is used, it's difficult to tell whether this is worth having.

Is the idea to first extract all the definitions into a type-level list, before beginning to generate the blueprint? An alternative would be to use a writer monad that collects the `[Definition]` as you go, then populating the definition section with the `[Definition]` collected. It's unclear whether this is better.

plutus-tx/src/PlutusTx/Blueprint/Schema.hs
```diff
@@ -0,0 +1,417 @@
...
+  HasSchemaDefinition x (_ ': xs) = HasSchemaDefinition x xs
+
+schemaRef :: forall typ schemas. (Typeable typ, HasSchemaDefinition typ schemas) => Schema
+schemaRef = SchemaRef $ show $ typeRep $ Proxy @typ
+
+----------------------------------------------------------------------------------------------------
+-- Modifiers ---------------------------------------------------------------------------------------
+
+setTitle :: String -> Schema -> Schema
+setTitle title = \case
```
zliu41, line 308: Similar point - it would be better to include this for review when there exists code that uses it. The usefulness of this is questionable since one should set their blueprint title when creating the blueprint, and rarely need to reset it afterwards.

plutus-tx/src/PlutusTx/Blueprint/TH.hs
```diff
@@ -0,0 +1,90 @@
...
+import Language.Haskell.TH.Datatype qualified as TH
+import PlutusTx.Blueprint.Class (HasDataSchema (..))
+import PlutusTx.Blueprint.Schema (HasSchemaDefinition, Schema (..))
+import PlutusTx.IsData.TH (makeIsDataIndexed)
+
+{- | Generate a 'ToData', 'FromData', 'UnsafeFromData', 'HasDataSchema' instances for a type,
+using an explicit mapping of constructor names to indices.
+Use this for types where you need to keep the representation stable.
+-}
+makeIsDataSchemaIndexed :: TH.Name -> [(TH.Name, Natural)] -> TH.Q [TH.InstanceDec]
```
zliu41, line 28: Some comments in this module would be helpful. TH can otherwise be quite hard to read.

And is there tests for this? If not then I'd again suggest that this be included for review only when it is used/tested.

plutus-tx/src/PlutusTx/Blueprint/Argument.hs
```diff
@@ -0,0 +1,42 @@
...
+  }
+  deriving stock (Show)
+
+instance ToJSON ArgumentBlueprint where
+  toJSON MkArgumentBlueprint{..} =
+    Aeson.object
+      $ catMaybes
+        [ fmap ("title" .=) argumentTitle
+        , fmap ("description" .=) argumentDescription
+        , case Set.size argumentPurpose of
```
michaelpj, line 37: You do this pattern many times, worth abstracting IMO. Maybe a newtype around `Set` that encodes like this, and then you can just do `"purpose" .?= (MaybeMultiple argumentPurpose)` or something.

plutus-tx/src/PlutusTx/Blueprint/Argument.hs
```diff
@@ -0,0 +1,42 @@
...
+  -- ^ An informative description of the redeemer or datum.
+  , argumentPurpose     :: Set Purpose
+  -- ^ A possibly empty set of purposes for the redeemer or datum.
+  , argumentSchema      :: Schema
+  -- ^ A Plutus Data Schema using the core vocabulary defined below,
+  -- or a oneOf applicator of Plutus Data Schemas
+  }
+  deriving stock (Show)
+
+instance ToJSON ArgumentBlueprint where
```
michaelpj, line 31: surely we're going to want the other direction too? Another point in favour of generating it if we can.

plutus-tx/src/PlutusTx/Blueprint/Class.hs
```diff
@@ -0,0 +1,84 @@
...
+
+module PlutusTx.Blueprint.Class where
+
+import Prelude
+
+import Data.Kind (Type)
+import PlutusTx.Blueprint.Schema (Schema (..))
+import PlutusTx.Builtins (BuiltinByteString, BuiltinData, BuiltinString)
+
+class HasDataSchema (t :: Type) (canReferTypes :: [Type]) where
```
michaelpj, line 16: haddock? what is this? I can't read your mind :)

plutus-tx/src/PlutusTx/Blueprint/Contract.hs
```diff
@@ -0,0 +1,54 @@
...
+import Data.Map (Map)
+import Data.Map qualified as Map
+import Data.Maybe (catMaybes)
+import Data.Text (Text)
+import PlutusTx.Blueprint.Definition (DefinitionId)
+import PlutusTx.Blueprint.Preamble (Preamble)
+import PlutusTx.Blueprint.Schema (Schema)
+import PlutusTx.Blueprint.Validator (ValidatorBlueprint)
+
+data ContractBlueprint = MkContractBlueprint
```
michaelpj, line 20: haddock

plutus-tx/src/PlutusTx/Blueprint/Contract.hs
```diff
@@ -0,0 +1,54 @@
...
+            $ "$vocabulary"
+            .= Aeson.object
+              [ "https://json-schema.org/draft/2020-12/vocab/core" .= True
+              , "https://json-schema.org/draft/2020-12/vocab/applicator" .= True
+              , "https://json-schema.org/draft/2020-12/vocab/validation" .= True
+              , "https://cips.cardano.org/cips/cip57" .= True
+              ]
+        , Just $ "preamble" .= contractPreamble
+        , Just $ "validators" .= contractValidators
+        , if Map.null contractDefinitions
```
michaelpj, line 48: again, if we wanted to be generic we could write a newtype wrapper around Map that omits it when it is empty.

plutus-tx/src/PlutusTx/Blueprint/Parameter.hs
```diff
@@ -0,0 +1,42 @@
...
+import Data.Aeson (ToJSON (..), (.=))
+import Data.Aeson qualified as Aeson
+import Data.Maybe (catMaybes)
+import Data.Set (Set)
+import Data.Set qualified as Set
+import Data.Text (Text)
+import PlutusTx.Blueprint.Purpose (Purpose)
+import PlutusTx.Blueprint.Schema (Schema)
+
+data ParameterBlueprint = MkParameterBlueprint
```
michaelpj, line 18: haddock

plutus-tx/src/PlutusTx/Blueprint/TH.hs
```diff
@@ -0,0 +1,90 @@
...
+ where
+#if MIN_VERSION_template_haskell(2,17,0)
+  tyvarbndrName (TH.PlainTV n _)    = n
+  tyvarbndrName (TH.KindedTV n _ _) = n
+#else
+  tyvarbndrName (TH.PlainTV n)      = n
+  tyvarbndrName (TH.KindedTV n _)   = n
+#endif
+
+toDataClause :: TH.Type -> [(TH.ConstructorInfo, Natural)] -> TH.ClauseQ
```
michaelpj, line 73: misnamed

plutus-tx/src/PlutusTx/IsData/TH.hs
```diff
@@ -4,28 +4,29 @@
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE ViewPatterns    #-}
 module PlutusTx.IsData.TH (
```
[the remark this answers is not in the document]
zliu41: When I do reformatting I do it just before merging!

kwxm:
@Unisay Sorry for the delay, but I'll try to review this today.

plutus-tx/src/PlutusTx/Blueprint/Argument.hs
```diff
@@ -0,0 +1,42 @@
...
+import Data.Aeson (ToJSON (..), (.=))
+import Data.Aeson qualified as Aeson
+import Data.Maybe (catMaybes)
+import Data.Set (Set)
+import Data.Set qualified as Set
+import Data.Text (Text)
+import PlutusTx.Blueprint.Purpose (Purpose)
+import PlutusTx.Blueprint.Schema (Schema)
+
+data ArgumentBlueprint = MkArgumentBlueprint
```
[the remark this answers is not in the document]
Unisay: :ok-hand:

plutus-tx/src/PlutusTx/Blueprint/Argument.hs
```diff
@@ -0,0 +1,42 @@
...
+  { argumentTitle       :: Maybe Text
+  -- ^ A short and descriptive name for the redeemer or datum.
+  , argumentDescription :: Maybe Text
+  -- ^ An informative description of the redeemer or datum.
+  , argumentPurpose     :: Set Purpose
+  -- ^ A possibly empty set of purposes for the redeemer or datum.
+  , argumentSchema      :: Schema
+  -- ^ A Plutus Data Schema using the core vocabulary defined below,
```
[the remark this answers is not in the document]
Unisay: You're right.

plutus-tx-plugin/test/Blueprint/Spec.hs
```diff
@@ -0,0 +1,102 @@
...
+        [ MkValidatorBlueprint
+            { validatorTitle = "Acme Validator"
+            , validatorDescription = Just "A validator that does something awesome"
+            , validatorParameters =
+                Just
+                  $ pure
+                    MkParameterBlueprint
+                      { parameterTitle = Just "Acme Parameter"
+                      , parameterDescription = Just "A parameter that does something awesome"
+                      , parameterPurpose = Set.singleton Purpose.Spend
```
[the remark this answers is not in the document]
Unisay:
```haskell
[Purpose.Spend]
```

is indeed more concise, but I am afraid that its also more confusing, as one needs to take into consideration a non-local code fact that `OverloadedLists` are enabled.

Alternatively: 
```haskell
pure Purpose.Spend
```

plutus-tx/src/PlutusTx/Blueprint/Argument.hs
```diff
@@ -0,0 +1,42 @@
...
+  -- or a oneOf applicator of Plutus Data Schemas
+  }
+  deriving stock (Show)
+
+instance ToJSON ArgumentBlueprint where
+  toJSON MkArgumentBlueprint{..} =
+    Aeson.object
+      $ catMaybes
+        [ fmap ("title" .=) argumentTitle
```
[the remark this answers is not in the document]
Unisay: Good point, but we can't use newer aeson version yet, due to the transitive[ dependencies conflicts](https://github.com/tweag/HaskellR/blob/63afa4bc78d90f93630127003e25b0f53fb0815a/inline-r/inline-r.cabal#L82).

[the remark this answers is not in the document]
michaelpj: We could probably make the existing `allow-newer: inline-r:aeson` unconditional, since we don't set the `with-inline-r` flag except locally, so it won't cause distribution problems.

plutus-tx/src/PlutusTx/Blueprint/Argument.hs
Unisay: Actually I am not sure if we're going to ever implement parsing blueprints so I have intentionally omitted `FromJSON` instances following the [YAGNI principle](https://martinfowler.com/bliki/Yagni.html).

plutus-tx/src/PlutusTx/Blueprint/Class.hs
Unisay: Sure, will add.

plutus-tx/src/PlutusTx/Blueprint/Contract.hs
Unisay: 👌🏼

plutus-tx/src/PlutusTx/Blueprint/Argument.hs
michaelpj: If anyone was going to _use_ blueprints written by someone else to operate a contract they would want this. It seems bizarre to me for us to provide a library for working with these and to not provide a way to parse them.

michaelpj: The only reason it costs us anything to do is because we're not deriving all these instances ;)

plutus-tx-plugin/test/Blueprint/Acme.golden.json
```diff
@@ -0,0 +1,119 @@
...
+    },
+    "Bytes": {
+      "title": "SchemaBytes",
+      "dataType": "bytes"
+    },
+    "Datum": {
+      "oneOf": [
+        {
+          "$comment": "DatumLeft",
+          "constructor": 0,
```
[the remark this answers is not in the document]
Unisay: Good catch and good point! I don't think that we need to test the golden schema every CI build, it should suffice to test it once [a change];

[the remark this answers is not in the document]
michaelpj: Why not? it should be fast and easy, right? We just need to write a little check that uses a schema checker from nixpkgs to check the schema against this output, and add it to our nix checks. I think that's probably the best way since there isn't a convenient pure-Haskell schema checker as far as I can tell :thinking:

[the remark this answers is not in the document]
Unisay: We can, and I like that such a schema check could be shared, as it turns out that doing schema validation isn't that straightforward, this way I can set it up once and then anybody who's changing the golden file can use it. OTOH, golden file is our expectation, its a form of a spec. It feels a bit wasteful to test if our expectation is actually a valid one for every PR. But maybe its me thinking of a premature optimisation.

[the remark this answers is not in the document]
michaelpj: - If it isn't checked, it's broken
- Checking it periodically is harder than checking it always

:shrug:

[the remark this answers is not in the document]
Unisay: You convinced me, will do.

Unisay:
> * I definitely want us to validate the output against the schema! I found at least one bug, there might be more.

I researched `JSON Schema` validator [tools](https://json-schema.org/implementations) that:

1. Support [Draft 2020-12](https://json-schema.org/draft/2020-12/json-schema-core.html)
2. Is available in the nix packages (or flakes) and are installable/runnable with nix;

and the result is Ø

Do you think we can relax the requirement 2 and if yes then how would we make such a tool available for CI?

michaelpj:
> Do you think we can relax the requirement 2 and if yes then how would we make such a tool available for CI?

It will make our life much harder. What tools are there that satisfy requirement 1? We might just be able to package one.

Unisay:
> What tools are there that satisfy requirement 1?

https://ajv.js.org/packages/ajv-cli.html

I use it this way atm:

```
nix-shell -p nodejs_21
```
then `npx` or `npm install`:

```
npx ajv-cli test \
  --strict=log \
  --spec=draft2020 \
  -s plutus-blueprint \
  -r plutus-blueprint-argument \
  -r plutus-data \
  -r plutus-blueprint-parameter \
  -r plutus-builtin \
  -d Acme.golden \
  --valid
```
(it does catch some invalid schemas, but it doesn't catch the error you found `s/constructor/index/` 🤷🏼‍♂️ )

plutus-tx-plugin/test/Blueprint/Acme.golden.json
```diff
@@ -0,0 +1,119 @@
...
+            "$ref": "#/definitions/Params"
+          }
+        }
+      ],
+      "compiledCode": "01010032222323232300349103505435003232325333573466e1d200000218000a999ab9a3370e90010010c00cc8c8c94ccd5cd19b874800000860026eb4d5d0800cdd71aba13574400213008491035054310035573c0046aae74004dd51aba100109802a481035054310035573c0046aae74004dd50029919192999ab9a3370e90000010c0004c0112401035054310035573c0046aae74004dd5001119319ab9c00180019999a891001111400400e00500110948014004900126101030048810001",
+      "hash": "82c3f1e456b3856514af88ec975abe5e105d673ca51bf36101d7bd76"
+    }
+  ],
+  "definitions": {
+    "BuiltinByteString": {
```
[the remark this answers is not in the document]
Unisay: Will strip'em

plutus-tx/src/PlutusTx/Blueprint/Schema.hs
Unisay: I will remove these modifiers as they aren't used in the test and arguably are premature.

plutus-tx/src/PlutusTx/Blueprint/Argument.hs
Unisay: This particular instance has custom encoding logic for the `purpose` field, which I don't know how to implement using an instance derivation without leaking a derivation-specific newtypes into the user space, so I decide to leave it manual.

plutus-tx/src/PlutusTx/Blueprint/Contract.hs
Unisay: In which case Blueprint API users will have to know about such a map wrapper and will have to deal with it. I think this price for the automatic derivation is not justified.

plutus-tx-plugin/test/Blueprint/Spec.hs
```diff
@@ -0,0 +1,102 @@
...
+import Test.Tasty (TestName)
+import Test.Tasty.Extras (TestNested, testNested)
+import Test.Tasty.Golden (goldenVsFile)
+
+goldenTests :: TestNested
+goldenTests = testNested "Blueprint" [goldenBlueprint "Acme" acmeContractBlueprint]
+
+goldenBlueprint :: TestName -> ContractBlueprint -> TestNested
+goldenBlueprint name blueprint = do
+  goldenPath <- asks $ foldr (</>) name
```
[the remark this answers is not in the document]
Unisay: This one is specific for `ContractBlueprint`s and there wasn't a one using `goldenVsFile`.

plutus-tx/src/PlutusTx/Blueprint/Definition.hs
```diff
@@ -0,0 +1,38 @@
...
+import Data.Typeable (Proxy (..), Typeable, typeRep)
+import PlutusTx.Blueprint.Class (HasDataSchema, dataSchema)
+import PlutusTx.Blueprint.Schema (Schema)
+
+-- | A reference to a Schema definition.
+newtype DefinitionId = MkDefinitionId Text
+  deriving stock (Show)
+  deriving newtype (Eq, Ord, ToJSON, ToJSONKey)
+
+definitionId :: forall a. (Typeable a) => DefinitionId
```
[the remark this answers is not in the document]
Unisay: Good point, will do.

plutus-tx/src/PlutusTx/Blueprint/Validator.hs
```diff
@@ -0,0 +1,58 @@
...
+        , ("description" .=) <$> validatorDescription
+        , Just $ "redeemer" .= validatorRedeemer
+        , ("datum" .=) <$> validatorDatum
+        , ("parameters" .=) <$> validatorParameters
+        , ("compiledCode" .=) . toHex . compiledCodeBytes <$> validatorCompiledCode
+        , ("hash" .=) . toHex . compiledCodeHash <$> validatorCompiledCode
+        ]
+   where
+    toHex :: ByteString -> Text
+    toHex = Text.decodeUtf8 . Base16.encode
```
kwxm, line 59: I always panic when text encodings make  an appearance,  but I did some experiments in ghci and this looks as if it's doing the right thing (going by the example in the CIP).  However, all that the CIP says about `compiledCode`  is that it should contain "The full compiled and cbor-encoded serialized flat script".  The compiled script is a binary object, so it has to be converted into some textual form for inclusion in JSON and I don't think they ever say exactly how that conversion is done (although maybe there's some standard JSON way of doing this that I'm not aware of).  Perhaps we should ask the CIP authors to be more explicit about the format they expect.

plutus-tx/test/Blueprint/Definition/Fixture.hs
```diff
@@ -0,0 +1,25 @@
...
+{-# LANGUAGE DeriveAnyClass        #-}
+{-# LANGUAGE DerivingStrategies    #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE TypeApplications      #-}
+{-# LANGUAGE UndecidableInstances  #-}
+{-# LANGUAGE ViewPatterns          #-}
+
+module Blueprint.Definition.Fixture where
```
kwxm, line 11: I'm not really sure what "Fixture" means here and elsewhere. Maybe it's some standard terminology that I'm unaware of?

plutus-tx/src/PlutusTx/Blueprint/Argument.hs
kwxm: > It seems bizarre to me for us to provide a library for working with these and to not provide a way to parse them.

Yes, that seems a bit strange.  It would make testing a lot easier if we had a parser.  It might be a lot of work if we're not generating instances automatically though.

plutus-tx-plugin/test/Blueprint/Spec.hs
```diff
@@ -0,0 +1,102 @@
...
+            { validatorTitle = "Acme Validator"
+            , validatorDescription = Just "A validator that does something awesome"
+            , validatorParameters =
+                Just
+                  $ pure
+                    MkParameterBlueprint
+                      { parameterTitle = Just "Acme Parameter"
+                      , parameterDescription = Just "A parameter that does something awesome"
+                      , parameterPurpose = Set.singleton Purpose.Spend
+                      , parameterSchema = schemaRef @Fixture.Params @SchemaDefinitions
```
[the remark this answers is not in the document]
Unisay: Nice, will do.

kwxm (commented):
This looks pretty sensible.  I've never used  JSON Schema though, so I found it difficult to form an opinion about whether everything was being done in the "right" way. Nothing jumped out at me as looking weird though, so I'm not too worried.  I did wonder if it would be easier to derive the `ToJSON` (and ideally also `FromJSON`) instances automatically, but it's hard to tell how easy/difficult that would be without having tried to do it myself.

There were quite a few places where things seemed a bit ambiguous, but I think that was because the CIP isn't always precise about what it means, not because there's anything wrong with the code.  If you agree with that (and you'll have a better idea since you've had to implement it) then we should probably try to get them to fix the CIP.

plutus-tx-plugin/test/Blueprint/Acme.golden.json
```diff
@@ -0,0 +1,131 @@
...
+        {
+          "title": "Acme Parameter",
+          "description": "A parameter that does something awesome",
+          "purpose": "spend",
+          "schema": {
+            "$ref": "#/definitions/Params"
+          }
+        }
+      ],
+      "compiledCode": "01010032222323232300349103505435003232325333573466e1d200000218000a999ab9a3370e90010010c00cc8c8c94ccd5cd19b874800000860026eb4d5d0800cdd71aba13574400213008491035054310035573c0046aae74004dd51aba100109802a481035054310035573c0046aae74004dd50029919192999ab9a3370e90000010c0004c0112401035054310035573c0046aae74004dd5001119319ab9c001800199999a8911199a891199a89100111111400401600900380140044252005001001400084a400a0020038004008848a400e0050012410101010101010101000498101030048810001",
```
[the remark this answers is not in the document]
kwxm: Also, the example they give has the extra CBOR header at the front.

plutus-tx-plugin/test/Blueprint/Tests.hs
```diff
@@ -0,0 +1,104 @@
...
+import PlutusCore.Crypto.Hash (blake2b_224)
+import PlutusTx.Blueprint.Purpose qualified as Purpose
+import PlutusTx.Builtins (BuiltinByteString, BuiltinData)
+import System.FilePath ((</>))
+import Test.Tasty (TestName)
+import Test.Tasty.Extras (TestNested, testNested)
+import Test.Tasty.Golden (goldenVsFile)
+
+goldenTests :: TestNested
+goldenTests = testNested "Blueprint" [goldenBlueprint "Acme" acmeContractBlueprint]
```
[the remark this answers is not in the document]
Unisay: Aiken's HelloWorld example isn't the best fit for purpose, as it doesn't include compile time Params and uses a small subset of possible types, while I am trying to test as much types/features as possible here.

plutus-tx/test/Blueprint/Definition/Fixture.hs
Unisay: Yep, https://en.wikipedia.org/wiki/Test_fixture

plutus-tx/src/PlutusTx/Blueprint/Validator.hs
Unisay: Agree

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
```diff
@@ -0,0 +1,108 @@
...
+{-# OPTIONS_GHC -fno-full-laziness #-}
+{-# OPTIONS_GHC -fno-ignore-interface-pragmas #-}
+{-# OPTIONS_GHC -fno-omit-interface-pragmas #-}
+{-# OPTIONS_GHC -fno-spec-constr #-}
+{-# OPTIONS_GHC -fno-specialise #-}
+{-# OPTIONS_GHC -fno-strictness #-}
+{-# OPTIONS_GHC -fno-unbox-small-strict-fields #-}
+{-# OPTIONS_GHC -fno-unbox-strict-fields #-}
+{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:target-version=1.1.0 #-}
+
```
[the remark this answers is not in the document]
Unisay: I'll remove them

plutus-tx-plugin/test/Blueprint/Acme.golden.json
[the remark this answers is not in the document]
Unisay: Done.

plutus-tx/src/PlutusTx/Blueprint/Schema.hs
```diff
@@ -0,0 +1,235 @@
...
+      Aeson.Object $ dataType info "#boolean"
+    SchemaBuiltInInteger info ->
+      Aeson.Object $ dataType info "#integer"
+    SchemaBuiltInBytes info ->
+      Aeson.Object $ dataType info "#bytes"
+    SchemaBuiltInString info ->
+      Aeson.Object $ dataType info "#string"
+    SchemaBuiltInPair info MkPairSchema{left, right} ->
+      dataType info "#pair"
+        & requiredField "left" left
```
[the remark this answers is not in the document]
Unisay: There is a meta-schema attached to the CIP and it requires attributes `left` and `right` 
https://github.com/cardano-foundation/CIPs/blob/b99df6288fe5c391124073d439b99379532f7e89/CIP-0057/schemas/plutus-builtin.json#L131

plutus-tx/src/PlutusTx/Blueprint/Schema.hs
```diff
@@ -0,0 +1,235 @@
...
+emptyBytesSchema = MkBytesSchema{enum = [], minLength = Nothing, maxLength = Nothing}
+
+data ListSchema = MkListSchema
+  { schema   :: Schema
+  -- ^ Element schema
+  , minItems :: Maybe Natural
+  -- ^ An array instance is valid if its size is greater than, or equal to, this value.
+  , maxItems :: Maybe Natural
+  -- ^ An array instance is valid if its size is less than, or equal to, this value.
+  , unique   :: Maybe Bool
```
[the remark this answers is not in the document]
Unisay: I'll rename `unique` to `uniqueItems` to match the CIP name.

plutus-tx/src/PlutusTx/Blueprint/Schema.hs
```diff
@@ -0,0 +1,235 @@
...
+        & optionalField "enum" (fmap toHex <$> nonEmpty enum)
+        & optionalField "maxLength" maxLength
+        & optionalField "minLength" minLength
+        & Aeson.Object
+     where
+      toHex :: ByteString -> Text
+      toHex = Text.decodeUtf8 . Base16.encode
+    SchemaList info MkListSchema{..} ->
+      dataType info "list"
+        & requiredField "items" schema
```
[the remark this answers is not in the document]
Unisay: It is https://github.com/cardano-foundation/CIPs/blob/b99df6288fe5c391124073d439b99379532f7e89/CIP-0057/schemas/plutus-builtin.json#L115

plutus-tx/src/PlutusTx/Blueprint/Schema.hs
```diff
@@ -0,0 +1,235 @@
...
+        & optionalField "minItems" minItems
+        & optionalField "maxItems" maxItems
+        & optionalField "uniqueItems" unique
+        & Aeson.Object
+    SchemaMap info MkMapSchema{..} ->
+      dataType info "map"
+        & requiredField "keys" keySchema
```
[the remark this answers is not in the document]
Unisay: https://github.com/cardano-foundation/CIPs/blob/b99df6288fe5c391124073d439b99379532f7e89/CIP-0057/schemas/plutus-data.json#L97

plutus-tx/src/PlutusTx/Blueprint/Schema.hs
```diff
@@ -0,0 +1,235 @@
...
+        & requiredField "values" valueSchema
+        & optionalField "minItems" minItems
+        & optionalField "maxItems" maxItems
+        & Aeson.Object
+    SchemaConstructor info MkConstructorSchema{..} ->
+      dataType info "constructor"
+        & requiredField "index" index
+        & optionalField "fields" (nonEmpty fieldSchemas)
```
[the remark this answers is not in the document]
Unisay: Good catch!

plutus-tx/src/PlutusTx/Blueprint/Schema.hs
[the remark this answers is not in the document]
kwxm: Ah, I didn't see that: I just looked at the CIP.  Thanks for pointing that out.  Maybe what the CIP says should agree with the schema though.

plutus-tx-plugin/test/Blueprint/Spec.hs
```diff
@@ -0,0 +1,102 @@
...
+                    { argumentTitle = Just "Acme Datum"
+                    , argumentDescription = Just "A datum that contains something awesome"
+                    , argumentPurpose = Set.singleton Purpose.Spend
+                    , argumentSchema = schemaRef @Fixture.Datum @SchemaDefinitions
+                    }
+            , validatorCompiledCode =
+                Just
+                  MkCompiledCode
+                    { compiledCodeBytes = Fixture.serialisedScript
+                    , compiledCodeHash = blake2b_224 Fixture.serialisedScript
```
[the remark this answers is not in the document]
Unisay: Good suggestion, implemented.

plutus-tx/src/PlutusTx/Blueprint/PlutusVersion.hs
```diff
@@ -0,0 +1,18 @@
...
+{-# LANGUAGE LambdaCase         #-}
+{-# LANGUAGE OverloadedStrings  #-}
+
+module PlutusTx.Blueprint.PlutusVersion where
+
+import Prelude
+
+import Data.Aeson (ToJSON (..))
+
+data PlutusVersion = PlutusV1 | PlutusV2 | PlutusV3
```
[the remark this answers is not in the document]
Unisay: You're right on many points here, 
I've added a haddock comment clarifying these concerns.

plutus-tx/src/PlutusTx/Blueprint/TH.hs
Unisay: Yes, there is a Blueprint test that includes this functionality: `Blueprint.Tests` + `Blueprint.Tests.Lib`.

I'll add comments to explain TH functionality as you suggested.

Unisay:
> Does this PR include the function for generating the blueprint? I was expecting to see something like `mkBlueprint :: CompiledCode (Datum -> Redeemer -> ScriptContext -> Bool) -> ContractBlueprint` but I'm not seeing it. 

Yes, see the `PlutusTx.Blueprint.Write` module, its called `writeBlueprint`. There is also a test case that uses it.

plutus-tx-plugin/test/Blueprint/Acme.golden.json
```diff
@@ -0,0 +1,119 @@
+{
+  "$schema": "https://cips.cardano.org/cips/cip57/schemas/plutus-blueprint.json",
```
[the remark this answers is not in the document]
Unisay: I've added it `scripts/blueprints/check-json-schemas.js` (this Javascript requires `nodejs` to run). 
I haven't added this check to CI, though.

plutus-tx/src/PlutusTx/Blueprint/Definition.hs
```diff
@@ -0,0 +1,78 @@
...
+  forall (types :: [Type]).
+  (AsDefinitionsEntries types types) =>
+  Map DefinitionId (Schema types)
+deriveSchemaDefinitions = Map.fromList (definitionEntries @types @types)
+
+{- |
+  A class of types that can be converted to a list of schema definition entries.
+  It is used internally to derive a map of schema definitions from a list of types.
+-}
+class AsDefinitionsEntries (allTypes :: [Type]) (remainingTypes :: [Type]) where
```
zliu41, line 63: What does `remainingTypes` mean? "remaining" in what sense?

plutus-tx/src/PlutusTx/Blueprint/Parameter.hs
```diff
@@ -0,0 +1,58 @@
...
+  The 'referencedTypes' phantom type parameter is used to track the types used in the contract
+  making sure their schemas are included in the blueprint and that they are referenced
+  in a type-safe way.
+-}
+data ParameterBlueprint (referencedTypes :: [Type]) = MkParameterBlueprint
+  { parameterTitle       :: Maybe Text
+  -- ^ A short and descriptive name for the parameter.
+  , parameterDescription :: Maybe Text
+  -- ^ An informative description of the parameter.
+  , parameterPurpose     :: Set Purpose
```
zliu41, line 34: Why is this a set?

zliu41 (commented):
> Yes, see the PlutusTx.Blueprint.Write module, its called writeBlueprint.

But the type of `writeBlueprint` is `FilePath -> ContractBlueprint types -> IO ()`. To use it, one must instantiate `types` with a type-level list, which they have to construct manually.

Is it not possible to instead provide the following function?

```haskell
writeSpendingBlueprint ::
  forall f datum redeemer scriptContext.
  ( HasSchema datum
  , HasSchema redeemer
  , IsScriptContext scriptContext
  ) =>
  FilePath ->
  BlueprintSettings ->
  f (datum -> redeemer -> scriptContext -> Bool) ->
  IO ()
```

This would be much more convenient to use, since one will be able to just pass their compiled code, `CompiledCode (MyDatum -> MyRedeemer -> ScriptContext -> Bool)` to it.

Note that this `HasSchema` would have to be different from the one you defined. With the correct definition of `HasSchema`, this should be able to automatically calculate all types referenced, and should also be able to obtain the Plutus version automatically from the `scriptContext` - e.g., if it is a V1 ScriptContext then it is for Plutus V1.

zliu41:
By the way, (to be done separately) we'd also need to think about how this will interact with `AsData` - @michaelpj

plutus-tx/src/PlutusTx/Blueprint/Parameter.hs
Unisay: To rule out an invalid state of having same purpose more than one time, e.g.
```haskell
parameterPurpose = [Purpose.Spend, Purpose.Spend, Purpose.Mint]
```

plutus-tx/src/PlutusTx/Blueprint/Definition.hs
Unisay: Maybe you can suggest a better name, the idea is the following:

2 instances (one for a recursive step and one for a termination) iterate a type level list, taking a head and tail at each step,  checking if the head is in the `allTypes` and recursing on tail `remainingTypes`. This way in the beginning of iteration `allTypes` == `remainingTypes` and then `allTypes` stays the same list, while `remainingTypes` is shrinking until empty.

Here is an analogy at the value level, where `remainingTypes` serves a similar purpose:
```haskell

type Typ = String
type DefinitionId = String
type Schema = String

asDefinitionEntries :: [Typ] -> [(DefinitionId, Schema)]
asDefinitionEntries allTypes = go allTypes allTypes
  where
    go :: [Typ] -> [Typ] -> [(DefinitionId, Schema)]
    go allTypes remainingTypes = 
      case remainingTypes of
        [] -> []
        (t : tail) -> 
          let defId = lookupDefinitionId t allTypes
              schema = lookupSchema t allTypes
          in (defId, schema) : go allTypes tail

lookupDefinitionId :: Typ -> [Typ] -> DefinitionId
lookupDefinitionId t allTypes | t `elem` allTypes = "DefinitionId for " ++ t
lookupDefinitionId t _ = error $ "Type " ++ show t ++ " not found"

lookupSchema :: Typ -> [Typ] -> Schema
lookupSchema t allTypes | t `elem` allTypes = "Schema for " ++ t
lookupSchema t _ = error $ "Type " ++ show t ++ " not found"
```

Unisay:
> To use it, one must instantiate `types` with a type-level list, which they have to construct manually.

The idea is that in the minimal implementation users supply a list of types involved into a contract API manually,
such that a convenience function (which derives such a list of types from a typed validator function type) could be provided on top of it .

Anyway, only providing `CompiledCode (Datum -> Redeemer -> ScriptContext -> Bool)` is not enough for a blueprint:

1. its missing all the meta information: Preamble, titles, descriptions, comments.
2. CIP-57 describes a contract blueprint that is comprised of potentially more than one validator.

michaelpj:
> By the way, (to be done separately) we'd also need to think about how this will interact with AsData

It's a bit awkward. In particular, you would like to use a schema definition based on the "real" datatype definition, which doesn't even exist with `AsData` Perhaps we can still emit it with a slightly different name or something :thinking:

plutus-tx/src/PlutusTx/Blueprint/Parameter.hs
zliu41: What I meant to ask is why does a parameter have multiple purposes? Isn't it a oneOf? In fact I don't understand why a parameter should have any purpose at all, I thought purpose is something for a script, not for an individual parameter.

zliu41 (commented):
> a convenience function (which derives such a list of types from a typed validator function type) could be provided on top of it .

Ok, that's fine as long as we plan to do that.

> its missing all the meta information: Preamble, titles, descriptions, comments.

That's what `BlueprintSettings` is for.

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
```diff
@@ -0,0 +1,106 @@
...
+$(makeIsDataSchemaIndexed ''Datum [('DatumLeft, 0), ('DatumRight, 1)])
+
+type Redeemer = BuiltinString
+
+type ScriptContext = ()
+
+type Validator = Params -> Datum -> Redeemer -> ScriptContext -> Bool
+
+serialisedScript :: ByteString
```
zliu41, line 73: This looks useful. Put it in a non-test module?

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
```diff
@@ -0,0 +1,106 @@
...
+  {-# INLINEABLE untypedValidator #-}
+  untypedValidator :: Params -> BuiltinData -> BuiltinString -> BuiltinData -> ()
+  untypedValidator params datum redeemer ctx =
+    PlutusTx.check $ typedValidator params acmeDatum acmeRedeemer scriptContext
+   where
+    acmeDatum :: Datum = PlutusTx.unsafeFromBuiltinData datum
+    acmeRedeemer :: Redeemer = redeemer
+    scriptContext :: ScriptContext = PlutusTx.unsafeFromBuiltinData ctx
+
+  validatorScript :: PlutusTx.CompiledCode (BuiltinData -> BuiltinString -> BuiltinData -> ())
```
zliu41, line 95: (To be done separately) You can get the untyped validator from the typed validator:
- First, compiled the typed validator, and obtain `CompiledCode (MyParams -> MyDatum -> MyRedeemer -> ScriptContext ->  Bool)`
- `unsafeApplyCode` liftCodeDef myParams`, which produces `CompiledCode (MyDatum -> MyRedeemer -> ScriptContext -> Bool`
- `unsafeApplyCode f`, where `f :: (d -> r -> sc -> Bool) -> (BuiltinData -> BuiltinData -> BuiltinData -> ())`, producing the untyped validator.

zliu41 (approved):
Please make some tasks for the follow up work, e.g.,
- The type level list should not need to contain nested types
  - or better yet, no type level list at all
- Make it work with `AsData`
- Obtain metadata from annotations
- The comment below

PR #5758, opened 2024-02-02
Adding a Costing Evaluator to agda-conformance

mjaskelioff:
This PR adds an agda evaluator which return costing. 

It currently doesn't use the provided `CostModelParams`, but rather default ones. Since default parameters are the ones used for creating the `upc.budget.expected` files this is not a problem right now, and all test pass. 

A final implementation should really get the parameters from the provided `CostModelParams`, by correctly implementing the function
```haskell
processModelParams  :: CostModelParams -> (CekMachineCosts ,  BuiltinCostMap)
```
to actually use its argument. 
 

Pre-submit checklist:
- Branch
    - [X] Tests are provided (if possible)
    - [X] Commit sequence broadly makes sense
    - [X] Key commits have useful messages
    - [X] Changelog fragments have been written (if appropriate)
    - [X] Relevant tickets are mentioned in commit messages
    - [X] Formatting, PNG optimization, etc. are updated
- PR
    - [ ] (For external contributions) Corresponding issue exists and is linked in the description
    - [X] Targeting master unless this is a cherry-pick backport
    - [X] Self-reviewed the diff
    - [X] Useful pull request description
    - [X] Reviewer requested

PR #5773, opened 2024-02-14
Kwxm/agda conformance/better cost model (PLT-9449)

kwxm:
#5758 ("Add a costing evaluator to agda-conformance") contained a comment saying
```
TODO: In processModelParams construct the pair in the result from the argument.
Here, we ignore the argument and just return a default set.
```

This PR fixes that. The problem is that the Agda evualator uses a simplified cost model representation and we have to convert a set of `CostModelParams` in to that format.  We have [multiple cost model representations](https://github.com/IntersectMBO/plutus/tree/master/doc/notes/cost-model-representations) and it was a little difficult to find the best way of doing the conversion.  There's a comment explaining what I eventually did, and why.

### DIFF
--- a/plutus-conformance/agda/Spec.hs
+++ b/plutus-conformance/agda/Spec.hs
@@ -5,8 +5,7 @@ module Main (main) where
 
 import Control.Monad.Trans.Except (ExceptT, runExceptT, withExceptT)
 
-import MAlonzo.Code.Evaluator.Term (runUCountingAgda)
--- import MAlonzo.Code.Cost.Raw (HRawCostModel)
+import MAlonzo.Code.Evaluator.Term (runUAgda, runUCountingAgda)
 
 import PlutusConformance.Common (UplcEvaluator (..), runUplcEvalTests)
 import PlutusCore (Error (..))
@@ -19,14 +18,13 @@ import PlutusCore.Evaluation.Machine.ExMemory (ExCPU (..), ExMemory (..))
 import PlutusCore.Evaluation.Machine.MachineParameters (CostModel (..))
 import PlutusCore.Evaluation.Machine.SimpleBuiltinCostModel (BuiltinCostKeyMap, BuiltinCostMap,
                                                              toSimpleBuiltinCostModel)
-
 import PlutusCore.Quote
 import UntypedPlutusCore qualified as UPLC
 import UntypedPlutusCore.DeBruijn
 
 import Data.Aeson (Result (Error, Success), fromJSON, toJSON)
 
--- This type corresponds to Cost.Raw.RawCostModel from the Agda code.
+-- This type corresponds to Cost.Raw.RawCostModel in plutus-metaheory.
 type RawCostModel = (CekMachineCosts, BuiltinCostMap)
 
 {- Convert a set of `CostModelParams` into a `RawCostModel` suitable for use with
@@ -55,9 +53,18 @@ toRawCostModel params =
 
     in (machineCosts, toSimpleBuiltinCostModel costKeyMap)
 
+{- Note [Evaluation with and without costing] We provide two evaluators, one with
+   costing and one without: normally the tests should be run with costing
+   enabled.  It may occasionally be necessary to turn the costing off, for
+   example if the Haskell costing implementation has changed and the Agda
+   implementation has not yet caught up: to do this, change `WithCosting`
+   to `WithoutCosting` in `main`.
+-}
+data CostOrNot = WithCosting | WithoutCosting
+
 -- Evaluate a UPLC program using the Agda CEK machine
-agdaEvalUplcProg :: UplcEvaluator
-agdaEvalUplcProg =
+agdaEvalUplcProg :: CostOrNot -> UplcEvaluator
+agdaEvalUplcProg WithCosting =
     UplcEvaluatorWithCosting $ \modelParams (UPLC.Program () version tmU) ->
         let
             -- turn the body of the program into an untyped de Bruijn term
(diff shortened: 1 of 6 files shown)


plutus-conformance/plutus-conformance.cabal
```diff
@@ -70,8 +70,9 @@ test-suite haskell-conformance
   hs-source-dirs: haskell test-cases
   other-modules:
   build-depends:
-    , base                >=4.9 && <5
+    , base                >=4.9   && <5
```
kwxm, line 73: I think that must have been `stylish-haskell`.

plutus-conformance/src/PlutusConformance/Common.hs
```diff
@@ -211,23 +211,6 @@ updateGoldenFile ::
 updateGoldenFile goldenPath (Left txt) = T.writeFile goldenPath txt
 updateGoldenFile goldenPath (Right p)  = T.writeFile goldenPath (display p)
 
--- | Our `evaluator` for the Haskell UPLC tests is the CEK machine.
```
kwxm, line 214: This didn't seem very common, so I moved it to the file where it's used.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1299,6 +1299,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
             -- execution times, so it's safe to use the same costing function for
             -- both.
             (runCostingFunThreeArguments . paramVerifyEd25519Signature)
+
```
kwxm, line 1302: Random tidying up.

plutus-conformance/agda/Spec.hs
```diff
@@ -3,91 +3,92 @@
...
 
--- | Our `evaluator` for the Agda UPLC tests is the CEK machine with costs.
-agdaEvalUplcProg :: UplcEvaluator
-agdaEvalUplcProg = UplcEvaluatorWithCosting $ \ modelParams (UPLC.Program () version tmU) ->
-    let
-        -- turn it into an untyped de Bruijn term
-        tmUDB :: ExceptT FreeVariableError Quote (UPLC.Term NamedDeBruijn DefaultUni DefaultFun ())
-        tmUDB = deBruijnTerm tmU
+-- This type corresponds to Cost.Raw.RawCostModel from the Agda code.
+type RawCostModel = (CekMachineCosts, BuiltinCostMap)
```
kwxm, line 30: I don't know if there's a better way to link this up with the corresponding type from Agda.  I tried experimenting with `COMPILE GHC` but didn't get anywhere.

mjaskelioff: if you are referring to the type synonym, you could import it from the `Opts.hs` file in the metatheory.

plutus-conformance/agda/Spec.hs
```diff
@@ -3,91 +3,92 @@
...
---             case runUAgda tmUDBSuccess of
---                 Left _ -> Nothing
---                 Right tmEvaluated ->
---                     let tmNamed = runQuote $ runExceptT $
---                             withExceptT FreeVariableErrorE $ unDeBruijnTerm tmEvaluated
---                     in
---                     -- turn it back into a named term
---                     case tmNamed of
---                         Left _          -> Nothing
---                         Right namedTerm -> Just $ UPLC.Program () version namedTerm
```
mjaskelioff, line 43: I had left the evaluator without costing just in case in the future we need to run momentarily the conformance test without costing (for example because the cost model in the metatheory is not in synch with the Haskell implementation).

plutus-conformance/agda/Spec.hs
```diff
@@ -3,91 +3,92 @@
...
+              Left e  -> error $ show e
+              Right r -> r
 
-        {- TODO: In processModelParams construct the pair in the result from the argument.
-          Here, we ignore the argument and just return a default set.
-        -}
-        processModelParams  :: CostModelParams -> (CekMachineCosts ,  BuiltinCostMap)
-        processModelParams _ = (defaultCekMachineCosts , defaultSimpleBuiltinCostModel)
+        costKeyMap =
+            case fromJSON @BuiltinCostKeyMap $ toJSON builtinCosts of
```
mjaskelioff, line 52: Clever! Go back to JSON where you came from!

mjaskelioff (approved):
LGTM. Moving the data back to JSON might no be the most efficient but it is an incredibly succinct solution. 
Great work Kenneth!

plutus-conformance/agda/Spec.hs
kwxm: Oh, I didn't notice the difference!  Now that you've mentioned it, it's tempting to try and make a single function that does both, but on closer examination that's maybe not so easy.

kwxm: @mjaskelioff I've put the evaluator without costing back and added a switch to let you turn it on and off easily.

plutus-conformance/agda/Spec.hs
kwxm: I'm still not sure about this.  The type `RawCostModel = (CekMachineCosts, BuiltinCostMap)` here corresponds to `RawCostModel = HCekMachineCosts × BuiltinCostMap` in the Agda code and I was wondering if there was some more "correct" way to say that they're really the same.  However, it works as it is, so maybe that's not necessary.

kwxm:
OK, I think I'll merge this.  It's doing what it's supposed to.

PR #5774, opened 2024-02-14
Demonstrate && evaluation inconsistencies

ana-pantilie:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-tx-plugin/test/Budget/9.6/andWithoutGHCOpts.pir.golden
```diff
@@ -0,0 +1,89 @@
...
+                (v : a -> a -> Bool)
+                (v : a -> a -> Bool)
+                (v : a -> a -> a)
+                (v : a -> a -> a) ->
+                 v)
+  in
+  \(x : integer) (y : integer) ->
+    let
+      !l : Bool = `<` {integer} `$fOrdInteger` x 3
+      !r : Bool = `<` {integer} `$fOrdInteger` y 3
```
ana-pantilie, line 80: Same as in the "local" case above.

plutus-tx-plugin/test/Budget/9.6/andWithLocal.uplc.golden
```diff
@@ -0,0 +1,12 @@
+program
+  1.1.0
+  ((\x y ->
+      (\lessThanInteger ->
+         (\l ->
+            (\r -> force (case l [(delay r), (delay (constr 1 []))]))
+              (lessThanInteger y 3))
+           (lessThanInteger x 3))
+        (\x y ->
+           force ifThenElse (lessThanInteger x y) (constr 0 []) (constr 1 [])))
```
ana-pantilie, line 10: I was checking if this term and the "local" one are equivalent, and noticed something weird. It looks like we're overloading `lessThanInteger` here:
- on line 4 it's a bound variable
- on line 10 (outside ^'s scope) it's, I would guess, a builtin

I think the two terms (this and the "local" one) are alpha-equivalent, though.

plutus-tx-plugin/test/Budget/9.6/andWithoutGHCOpts.eval.golden
```diff
@@ -0,0 +1,2 @@
+({cpu: 1822026
+| mem: 5504})
```
ana-pantilie, line 2: This is equal to the "local" result. Shouldn't turning off all of the GHC optimisations have some impact as well?

plutus-tx-plugin/test/Budget/9.6/andWithLocal.uplc.golden
zliu41: They are not equivalent. The one introduced in line 4 corresponds to the one in `andWithLocal.pir.golden`, line 5. It returns the non-builtin `Bool` (which in PIR is just `Bool`, and in UPLC it is `constr 0 []` for True and `constr 1 []` for False). And it comes from `PlutusTx.Builtins.lessThanInteger`. In the `WithGhcOpts` test, it is inlined by GHC.

The `lessThanInteger` in line 10 is indeed the builtin function, and it returns the builtin `Bool`. The condition in the builtin `ifThenElse` requires the builtin `Bool`.

plutus-tx-plugin/test/Budget/9.6/andWithoutGHCOpts.eval.golden
zliu41: The PIRs are different but after the UPLC optimizations they become the same UPLC program.

The local one doesn't run much GHC optimization either. Even though the module is compiled with optimization, the plugin is installed quite early in the core-to-core pipeline, so it runs before most optimizations.

zliu41 (approved):
Looks great.

plutus-tx-plugin/test/Budget/9.6/andWithLocal.uplc.golden
michaelpj: Yes, there is name shadowing here, unfortunately.

plutus-tx-plugin/test/Budget/9.6/andWithoutGHCOpts.pir.golden
```diff
@@ -0,0 +1,89 @@
...
+              {a -> a -> Bool}
+              (\(v : (\a -> a -> a -> Bool) a)
+                (v : a -> a -> Ordering)
+                (v : a -> a -> Bool)
+                (v : a -> a -> Bool)
+                (v : a -> a -> Bool)
+                (v : a -> a -> Bool)
+                (v : a -> a -> a)
+                (v : a -> a -> a) ->
+                 v)
```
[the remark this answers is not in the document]
michaelpj: I think you mean specialization, but yeah, typeclass specialization is one of the complicated GHC optimizations that we would really benefit from but usually can't use.

plutus-tx-plugin/test/Budget/WithoutGHCOptimisations.hs
```diff
@@ -0,0 +1,11 @@
+{-# OPTIONS_GHC
+ -O0 -fmax-simplifier-iterations=0
+ -fno-omit-interface-pragmas
+ -fno-ignore-interface-pragmas
+#-}
```
michaelpj, line 5: nit for the future, we generally avoid multi-line pragmas, it's a bit less diff-friendly than doing one pragma per line with one option each

Issue #5711, opened 2024-01-10
Faster pattern matching for built-in types


effectfully:
This issue is for discussing approaches to improving performance of pattern matching for built-in types.

The current approach is inefficient. Details can be found in [this](https://github.com/IntersectMBO/plutus/blob/81ac3ed283e2928faa51185eb6c29696f9d3af39/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs#L902) Note, the gist is that currently pattern matching over lists is implemented as (pseudocode)

```haskell
matchList :: [a] -> b -> (a -> [a] -> b) -> b
matchList xs z f = chooseList xs (\_ -> z) (\_ -> f (head xs) (tail xs)) ()
```

where `chooseList`, `head` and `tail` are all built-in functions (`chooseList` returns either its second or its third argument depending on whether its first argument is null or not, respectively). Therefore in case of `nil` we end up performing one builtin call and in case of `cons` we end up performing three builtin calls. Note how we also have to pass a unit argument around in order not to always evaluate all branches: Plutus is strict, hence `chooseList xs z (f (head xs) (tail xs))` would have the wrong semantics.

There have been two proposals on how to implement faster pattern matching for built-in types:

1. allow expressing pattern matching built-in functions directly, implemented in #5486 
2. piggy-back on the pattern matching machinery that we use for sums-of-products, implemented in #5704

Both PRs are documented, see the initial comment on each PR if you want to understand the details of design and implementation.

Both approaches work and give us comparable speedup on built-in-list-specific benchmarks. 

This is what we get for (1):

![Screenshot from 2024-01-12 00-41-30](https://github.com/IntersectMBO/plutus/assets/10480926/ac4642ef-9ad9-4d07-a3ae-4030a978efb8)

![Screenshot from 2024-01-12 00-45-13](https://github.com/IntersectMBO/plutus/assets/10480926/3912e777-c036-4ad6-b62c-7d725e18c37a)

This is what we get for (2):

![Screenshot from 2024-01-12 00-27-27](https://github.com/IntersectMBO/plutus/assets/10480926/54d35ee5-b5b3-4c98-a783-1bf582c1bb99)

![Screenshot from 2024-01-12 00-27-41](https://github.com/IntersectMBO/plutus/assets/10480926/8308a400-3262-4bd1-8703-e4a5a1568a52)

[...]

PR #5776, opened 2024-02-15
[Builtins] add 'constrTermFromConstrData'

effectfully:
An experimental and very incomplete implementation of the `constrTermFromConstrData` builtin converting `Data.Constr` to a value that can be `case`d on.


effectfully:
We aren't doing it as per #5777 and #5711, hence closing.

PR #5779, opened 2024-02-15
Use specialised `ByteStringMap`

ana-pantilie:
As can be seen by the updated benchmarks, this approach is actually worse.

@zliu41 figured out that it must be because having two map libraries somehow doesn't enable the compiler to optimise as much as it could before. He found that by turning off all optimisations with:
```
{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:max-simplifier-iterations-pir=0 #-}
{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:max-simplifier-iterations-uplc=0 #-}
{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:max-cse-iterations=0 #-}
```
we revert back to similar performance results.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


michaelpj:
It's interesting that this is slower. It seems like it should benefit at least a bit from not having to pass the `Eq` dictionaries around.

Note that I think the main benefit of using a map with monomorphic keys is that then we _can_ implement that as a builtin type because we don't have to pass in the dictionaries.

zliu41:
> It's interesting that this is slower.

It makes things slower with optimizations. E.g., using two different Maps can make CSE less effective. CSE is not the only one though: I tried turning off CSE only, and there are still some test cases that got slower.

PR #5784, opened 2024-02-19
Fix the Notes (PLT-9513)

kwxm:
I was trying to tidy up the Notes and resolve some missing references and it turned out that there was a lot of inconsistency.  I've made up a bunch of rules (see below) and written a script (`check-Notes.sh`) that will scan the codebase and report violations: we could potentially use this in CI to check that the Notes are consistent.  The output of this when applied to the current state of `master` is attached later.  I've fixed most of the problems but there were some missing and duplicate references that I wasn't able to resolve.  I encourage other people to fix these.  I've also added a `list-Notes.sh` script that will print out a list of the titles of Notes and their locations: this was useful for finding Notes whose references were slightly wrong.

The rules that `check-Notes.sh` checks for are:

 * Notes are comments beginning with `{- Note [<title>]` or `-- Note [<title>]`.  The `{-` form is preferred.

 * A reference to a Note is of the form `Note [<title>]`.

 * The title of a note can be any sequence of printable characters, excluding ']' but including spaces

 * In both definitions and references any number of spaces is allowed before "Note" and between "Note" and "[".

 * No characters except spaces are allowed between "Note" and "["; in particular `Note: [...]` is prohibited.

 * References must refer to the exact title of the relevant Note, including spaces and capitalisation.

 * No two Notes should have the same title.

 * All references must refer to Notes which are defined somewhere in the codebase.

 * In definitions the title is allowed to contain line breaks but there must be no line breaks between the start of the comment and "Note" or between "Note" and "[".

 * In references, line breaks are allowed in the title and after "Note".  This is allowed because code formatters may insert line breaks in these positions automatically.

 * `NOTE [` and `note [` are not allowed in either definitions or references.

 * Definitions of Notes must be ordinary comments: Haddock comments are not allowed.

 * Plural references are not allowed: "See Notes [A], [B], and [C]" is prohibited.  To avoid this, use a single reference to each note: "See Note [A], Note [B], and Note [C]".

Here is the output of `check-Notes.sh` on `master`.
[note-problems-master.txt](https://github.com/IntersectMBO/plutus/files/14328670/note-problems-master.txt)


plutus-benchmark/nofib/src/PlutusBenchmark/NoFib/Knights/ChessSetList.hs
```diff
@@ -35,7 +35,7 @@ type Tile     = (Integer,Integer)
 data ChessSet = Board
                 Integer      -- % Size of board (along edge)
                 Integer      -- % Current move number
-                (Maybe Tile) -- % Initial square: see Note [deleteFirst] below
+                (Maybe Tile) -- % Initial square: see Note (deleteFirst) below
```
kwxm, line 38: A comment from the original `nofib` source.

kwxm:
Here are the problems that are currently unresolved.

```
Duplicate Note [Serialization of PIR] at ./plutus-core/plutus-ir/src/PlutusIR/Core/Instance/Flat.hs:14 and ./plutus-core/plutus-ir/src/PlutusIR/Core/Type.hs:56
Duplicate Note [Global uniqueness] at ./plutus-core/plutus-core/src/PlutusCore/Normalize/Internal.hs:29 and ./plutus-core/plutus-core/src/PlutusCore/TypeCheck/Internal.hs:44
Duplicate Note [Stable encoding of PLC] at ./plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Flat.hs:28 and ./plutus-core/plutus-core/src/PlutusCore/Flat.hs:34
Duplicate Note [Encoding/decoding constructor tags using Flat] at ./plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Flat.hs:49 and ./plutus-core/plutus-core/src/PlutusCore/Flat.hs:55
Missing Note [Default pretty instances for PLC]            ./plutus-core/plutus-ir/src/PlutusIR/Core/Instance/Pretty.hs:139
Missing Note [Parsing horribly broken]                     ./plutus-core/testlib/PlutusCore/Generators/Hedgehog/AST.hs:110
Missing Note [Inlining of fully applied functions]         ./plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs:9
Missing Note [Serialising unit annotations]                ./plutus-core/plutus-core/src/PlutusCore/Flat.hs:13
Missing Note [No catch-all]                                ./plutus-core/plutus-core/src/PlutusCore/Core/Instance/Eq.hs:113
Missing Note [Annotation types]                            ./plutus-core/executables/src/PlutusCore/Executable/Common.hs:318
Missing Note [Unboxed tuples]                              ./plutus-tx-plugin/src/PlutusTx/Compiler/Type.hs:346
Missing Note [Elaboration of higher-kinded type variables] ./plutus-core/plutus-core/src/PlutusCore/Builtin/Elaborate.hs:112
Missing Note [Serialising Scripts]                         ./plutus-core/plutus-core/src/PlutusCore/Flat.hs:13
```

zliu41 (approved):
Great, thanks for doing this! I'm shocked that you can write this much awk without gouging out your eyeballs.

Would it make sense to allow each Note to have at least one reference?

> In references, line breaks are allowed in the title and after "Note". This is allowed because code formatters may insert line breaks in these positions automatically.

Does stylish-haskell/fourmolu insert line breaks in comments? I didn't know that they do.

> In both definitions and references any number of spaces is allowed before "Note" and between "Note" and "[".

> In definitions the title is allowed to contain line breaks but there must be no line breaks between the start of the comment and "Note" or between "Note" and "[".

Unless this has to do with code formatters, I'd just do the simple thing - requiring that there be exactly one space between "Note" and "[", and the title be on a single line.

> "See Note [A], Note [B], and Note [C]"

I suppose this can spread across multiple lines?

kwxm:
> I'm shocked that you can write this much awk

It started off small but then I kept discovering (literal) edge cases.

> Does stylish-haskell/fourmolu insert line breaks in comments? I didn't know that they do.

Actually, maybe not.  Editors definitely will though.

> Unless this has to do with code formatters, I'd just do the simple thing - requiring that there be exactly one space between "Note" and "[", and the title be on a single line.

It's partly to do with automatic formatting. especially at the end of lines: I definitely saw a case that looked like
```
{- ...................................................... See Note
   [Some Note].  ..............................
-}
and you either have to accept that or complain, and it requires some work in either case.  Maybe I'll say that you should have exactly one space, but secretly accept more. We have to allow the title to be split over more than one line in a reference (like [here](https://github.com/IntersectMBO/plutus/blob/760a6dadc758c32399eb6b89bf84fb29f4f97a79/plutus-core/cost-model/budgeting-bench/Benchmarks/Nops.hs#L144), for example).  It uses the same code to extract titles in both definitions and references, so it's probably easier to allow line breaks in titles in definitions as well (but hope that people don't do it).

> > "See Note [A], Note [B], and Note [C]"
> I suppose this can spread across multiple lines?

Yes, that's fine.  It just treats each one as a separate reference. It seemed hard to parse mulitple references in general though (what about "See Notes [A] for more information about blah blah blah and [B] for the case of ranges like [1..10]"?) so just banning "Notes" seemed the simplest thing to do.

**Comment by kwxm** [comment]
Incidentally, please don't merge this yet: I'd like to get the missing and duplicated notes fixed first.

File: plutus-core/plutus-core/stdlib/PlutusCore/StdLib/Type.hs [hunk 2]
```diff
@@ -283,13 +283,13 @@ We've elaborated the encoding on example, but there is a problem to consider her
...
 -}
 
-{- Note [Denormalization]
-Originally, we were binding 'withSpine' and 'patN' (taken from the end of
-Note [Packing n-ary functors]) on the Plutus Core side and this resulted in huge unreadable types
-being produced. Now we bind 'withSpine', 'patN' and what 'withSpine' receives on the Haskell side,
-i.e. we use Haskell lambdas to bind variables and regular function application to eliminate those
-lambdas which allows us to defer type reduction business to Haskell.
-Here is how the definition of 'list' looks like:
+{- Note [Denormalization] Originally, we were binding 'withSpine' and 'patN'
```
effectfully, line 286: Accidental?

scripts/check-Notes.awk
```diff
@@ -0,0 +1,198 @@
...
+        if (linecount++ >= 5) {              # Stop it from eating the entire file if the closing bracket is missing
+            printf ("Unterminated Note title at %s:%d?\n", FILENAME, FNR) > "/dev/stderr"
+            exit_now = 1                     # We need this because `exit` jumps to the END actions
+            exit 1
+        }
+    }
+
+    t = $0
+    sub(/].*/, "", t)                        # Delete first closing bracket and everything after it
+    title = title ? title " " t : t          # ... and append everything before it to the title
```
effectfully, line 100: image

effectfully (approved):
This is extremely helpful, thank you. I'm in favor of keeping your tool (and integrating it into CI) instead of using the other one.

plutus-core/plutus-core/stdlib/PlutusCore/StdLib/Type.hs
kwxm: Probably an Emacs accident.

kwxm:
All of the problems have now been resolved.  We might want to think about running the `check-Notes.sh` script as part of CI, which would allow us to spot things like duplicate Notes and references to nonexistent ones, but let's do that as a separate ticket.

PR #5786, opened 2024-02-19
Add note on constant inlining

ana-pantilie:
Adding @michaelpj 's comment from https://github.com/IntersectMBO/plutus/pull/5766/files/581699a2fdbdbf1730a483d91f41b05be4c1366d#r1494395447.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
@@ -320,6 +320,15 @@ and are used more than once, we are at risk of doing more work or making things 
 
 -}
 
+{- Note [Inlining constants]
+
+Constants can in principle be large, and hence inlining them can make the program bigger.
+However, usually they are not large, and inlining them is helpful, in particular it can expose
+more opportunities for evaluating builtins.
+
+Hence we inline constants by default, but refrain from doing so if we are trying to be conservative.
+-}
+
 -- See Note [Inlining criteria]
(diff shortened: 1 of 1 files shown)


michaelpj (approved):
LGTM! Assuming this is in fact your reasoning :)

Also, do we need to change the UPLC inliner?

ana-pantilie:
@michaelpj

> Assuming this is in fact your reasoning :)

Yep, that's the exact motivation for doing this. Thanks for suggesting the note!

> Also, do we need to change the UPLC inliner?

Good question. From the issue description:
> We have a PIR inliner and an UPLC inliner. Since constant folding is done in PIR (PlutusIR.Transform.EvaluateBuiltins), the inlining of constants also needs to be done in PIR.

If we don't do any constant folding in the UPLC inliner, would inlining in UPLC bring any benefits? If you think it's worth trying I can investigate. Anyway, I didn't look at the UPLC inliner initially and now I see that it's supposed to be in-sync with the PlutusIR implementation, so I'll add a note if we decide to not do constant inlining here.

michaelpj:
> Anyway, I didn't look at the UPLC inliner initially and now I see that it's supposed to be in-sync with the PlutusIR implementation, so I'll add a note if we decide to not do constant inlining here.

Keeping it fully in-sync is tricky, but I think we should try and minimize unnecessary deviations between the two. In this case I agree it probably won't make that much difference, but it's just easier to reason about what's going on if they behave as similarly as possible, so I'd be in favour of adding the behaviour there as well.

zliu41 (approved):

PR #5762, opened 2024-02-05
Benchmarking experiments

kwxm:
This just changes some indentation to allow us to open a PR for benchmarking experiments.

PR #5788, opened 2024-02-19
Minor edit for experimentation

kwxm:
More benchmark experiments, carrying on from #5762.


kwxm:
I'm going to close this because the base commit has disappeared again.

PR #5796, opened 2024-02-21
Revert #5677 (More efficient Value.geq and leq)

zliu41:
@paluh noticed that the change made in #5677 isn't correct, e.g., when the two Values are otherwise the same, except that there's an additional negative entry in the left Value.

Opened https://input-output.atlassian.net/browse/PLT-9525 to improve testing on Value.


effectfully (approved):
That was subtle. Great catch!

michaelpj:
Yeah, we need much better testing here, these are critical functions.

PR #5797, opened 2024-02-21
[Generators] Fix 'LamAbs' handling in 'fixupTerm_'

effectfully:
Before this PR running

```
cabal test plutus-ir-test --test-options "--pattern shrinkTermSound --quickcheck-replay=21373"
```

produces

```
plutus-ir/test/Driver.hs
  PlutusIR
    Generators.QuickCheck.Tests
      shrinkTermSound: FAIL (0.23s)
        *** Failed! Falsified (after 6 tests and 7 shrinks):
        ty,tm = ( all a5_17. a5_17 -> (all a4_19. a5_17)
        , let
          data (d8_8 :: * -> *) a10_10 | m6_6 where
            
        in
        /\a5_8 -> \(x13_13 : a5_8) -> /\a4_11 -> x13_13 )
        [ ( ( a5_8 -> (all a4_19. unit)
        , let
          data (d8_8 :: * -> *) a10_10 | m6_6 where
            
        in
        \(x13_13 : a5_8) -> error {all a4_19. unit} )
        , Error during PIR typechecking:
        Type-level name mismatch at (): d8 is in scope, but a5 having the same Unique 20 is attempted to be referenced ) ]
        Use --quickcheck-replay=21373 to reproduce.
```

After:

```
plutus-ir/test/Driver.hs
  PlutusIR
    Generators.QuickCheck.Tests
      shrinkTermSound: OK (0.08s)
        +++ OK, passed 10 tests.
```

### DIFF
--- a/plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
+++ b/plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
@@ -82,7 +82,9 @@ fixupTerm_ :: TypeCtx
 fixupTerm_ tyctxOld ctxOld tyctxNew ctxNew tyNew tm0 =
   case inferTypeInContext tyctxNew ctxNew tm0 of
     Left _ -> case tm0 of
-      LamAbs _ x a tm | TyFun () _ b <- tyNew -> bimap (TyFun () a) (LamAbs () x a)
+      -- Make @a@ the new type of @x@. We can't take the old type of @x@, because it may reference
+      -- a removed binding. And we're trying to change the type of @tm0@ to @tyNew@ anyway.
+      LamAbs _ x _ tm | TyFun () a b <- tyNew -> bimap (TyFun () a) (LamAbs () x a)
                                               $ fixupTerm_ tyctxOld (Map.insert x a ctxOld)
                                                            tyctxNew (Map.insert x a ctxNew) b tm
       Apply _ (Apply _ (TyInst _ (Builtin _ Trace) _) s) tm ->


plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
```diff
@@ -82,7 +82,9 @@ fixupTerm_ :: TypeCtx
 fixupTerm_ tyctxOld ctxOld tyctxNew ctxNew tyNew tm0 =
```
effectfully, line 82: Note to self: `tyctxNew` and `ctxNew` aren't used, are they? Try simplifying this whole thing.

plutus-core/testlib/PlutusIR/Generators/QuickCheck/ShrinkTerms.hs
```diff
@@ -82,7 +82,9 @@ fixupTerm_ :: TypeCtx
...
   case inferTypeInContext tyctxNew ctxNew tm0 of
     Left _ -> case tm0 of
-      LamAbs _ x a tm | TyFun () _ b <- tyNew -> bimap (TyFun () a) (LamAbs () x a)
+      -- Make @a@ the new type of @x@. We can't take the old type of @x@, because it may reference
+      -- a removed binding. And we're trying to change the type of @tm0@ to @tyNew@ anyway.
+      LamAbs _ x _ tm | TyFun () a b <- tyNew -> bimap (TyFun () a) (LamAbs () x a)
```
effectfully, line 87: Everyone's favorite x-hours-investigation-and-one-line-fix.

effectfully (commented):
What bothers me here is how come we didn't run into this earlier? What's so unique about generating a type abstraction that binds a variable clashing with some existing one? That's most of the setup, the only remaining part is referencing that variable in a lambda. Although, it's not clear whether the lambda has to come right after the type abstraction. Well, maybe it's indeed not a trivial thing to generate, although it surely is very basic still.

kwxm:
I don't really understand what's going on here, but I'm willing to believe that you've fixed it.

Question: it says 
```
d8 is in scope, but a5 having the same Unique 20 is attempted 
``` 
In the code that it shows us it has things like
```
 all a5_17. a5_17 -> (all a4_19. a5_17)
```
What are the numbers after the underscores? I assumed they were the uniques, but I don't see 20 there anywhere (although it does have `a5_8` later, under a `/\`).

kwxm (approved):

effectfully:
> I assumed they were the uniques, but I don't see 20 there anywhere (although it does have `a5_8` later, under a `/\`).

They are uniques, it's just that the type checker performs renaming as a prepass and then we end up seeing the wrong uniques in the error message. In Zonk the renamer would allow the caller to unrename the thing back, which would allow us to get the original uniques, but we never bothered implementing anything like that in Plutus. We could simply preserve the old uniques in annotations or whatever, but again we never bothered to.

PR #5799, opened 2024-02-22
Generalised force-delay optimisation

ana-pantilie:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
```diff
@@ -226,6 +227,40 @@ multiApp = runQuote $ do
...
+      t = Delay () (LamAbs () x (Delay () (LamAbs () y (Delay () (LamAbs () z (Var () z))))))
+      app = Apply () (Force () (Apply () (Force () (Apply () (Force () t) one)) two)) three
+  pure app
+
+-- | The UPLC term in this test should come from the following TPLC term after erasing its types:
+-- @ (/\(p :: *) -> \(x : p) -> /\(q :: *) -> \(y1 : q) (y2 : String) (y3 : String) -> /\(r :: *) -> \(z : r) -> z) Int 1 Int 2 "foo" "bar" Int 3 @
+-- Note that compared to the 'forceApply' test above, this term has multiple term applications nested inside a single
+-- type abstraction/instantiation.
+forceMultipleApply :: Term Name PLC.DefaultUni PLC.DefaultFun ()
+forceMultipleApply = runQuote $ do
```
ana-pantilie, line 249: WIP: this test isn't good enough because the term inside the body doesn't contain references to all the bound vars and the inlining optimisation simplifies the term

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,186 @@
...
+
+   force ((\x -> delay b) a1 a2) => force ((delay b[x1/a1]) a2)
+ @
+ The latter cannot be reduced further according to the reducation semantics of UPLC.
+
+ To generalise, we consider the family of terms above applied to a family of types and
+ terms:
+ @
+   (/\T1 -> \X1 ->  ... -> /\Tn -> \Xn -> body)
+     T1 X1 ... Tn Xn
```
effectfully, line 39: Overloading of the names is kinda unfortunate. E.g. `T1` means both an expression and a variable mapping to that expression.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,186 @@
...
+ @
+   (force^|Tn| (... (force^|T3| ((force^|T2| ((force^|T1| original) X1)) X2)) X3 ...)) Xn
+ @
+ After inlining @original@ we get:
+ @
+   (force^|Tn|
+     (...
+       (force^|T3|
+         ((force^|T2|
+           ((force^|T1|
```
effectfully, line 60: Why do we jump from one opening paren to two opening parens and then stay at two opening parens? I tried writing down the expression myself and only got one pair of parens at each node, but I'm known for messing up such things.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,186 @@
...
+             (delay^|T1|
+               (\X1 ->
+                  delay^|T2| (\X2 ->
+                    delay^|T3| (\X3 ->
+                      ... ->
+                        delay^|Tn| (\Xn -> body)))))
+            ) X1)) X2)) X3 ...)) Xn
+ @
+ In the end, after applying the "UntypedPlutusCore.Transform.ForceDelay" optimisation:
```
effectfully, line 69: So I had to do the whole thing manually using my own small non-general example, because getting through this doubly-generalized (`n` lists of terms interleaved with `n` lists of types) is very non-trivial, to me at least. Maybe add a simple example for demonstration purposes? Like an `apply` followed by a `force`, followed by an `apply`, potentially followed by one more `force` or something of this sort -- whatever demonstrates the generality without any iterated notation.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,186 @@
...
+ @
+ where @T1 ... Tn@ are lists of type variables (e.g. @T1@ could be @[t, q, p]@)
+ and @X1 ... Xn@ are lists of term variables. Of course, each @/\@ and @\@ here would
+ desugar to a sequence of type/term abstractions.
+
+ In order to reason about the proposed optimisation we need to consider such terms in
+ the context of them being applied to some sequence of terms.
+
+ One important observation is that this transformation requires that the underlying
+ (term)-lambda abstraction will be fully reduced by the applications.
```
effectfully, line 21: Not just fully reduced, it should be exactly reduced, right? I.e. `force ((\x y -> delay b) e)` is as erroneous as `force ((\x -> delay b) e1 e2)`, right? Or am I being silly?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,186 @@
...
+
+{- | Converts the subterm of a 'Force' into specialised types for representing
+ multiple applications on top of multiple abstractions. Checks whether the lambda
+ will eventually get "fully reduced" and applies the optimisation.
+ Returns 'Nothing' if the optimisation cannot be applied.
+-}
+optimisationProcedure :: Term name uni fun a -> Maybe (Term name uni fun a)
+optimisationProcedure term = do
+    asMultiApply <- toMultiApply term
+    innerMultiAbs <- toMultiAbs . toApply $ asMultiApply
```
effectfully, line 138: It's rather weird that we have force-delay as a separate optimization given that `toMultiApply` detecting no applications would give pretty much precisely force-delay, wouldn't it?

effectfully (commented):
Partial review, need to think about it for a bit.

Docs are great, but having a small example would help a lot I think.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,186 @@
...
+    guard $ length (applyToIter asMultiApply) == length (absVars innerMultiAbs)
+    case rhs innerMultiAbs of
+        Delay _ subTerm ->
```
effectfully, line 141: So since we have a separate force-delay pass anyway, why do we insist on finding `Delay` here? Why not simply turn

```
force (apply (apply (\x1 x2 -> body) e1) e2)
```

into

```
(apply (apply (\x1 x2 -> force body) e1) e2)
```

and let force-delay clean `force body` up if `body` is a `delay`? This transformation is correct even if `body` isn't a `delay`, right?

I'm not sure if it's helpful to implement things this way, but maybe we can implement an additional optimization that remove an outer `delay` from a binding if that binding is always accompanied by a `force` at all call sites within a program. But even aside from that, not requiring a `Delay` to appear there is simpler, so why not lift the restriction?

We could also generalize that to for example transform

```
force (apply (apply (\x1 -> body) e1) e2)
```

into, erm, this:

```
apply (\x1 -> force (apply body e2)) e1
``` 

This has the advantage that now `body` is directly applied to `e2` instead of appearing under a lambda, so maybe that would help the callsite inliner if we had one for UPLC and not just PIR?

Speaking of which, @zliu41 @michaelpj does the callsite inliner recognize cases where a function is fully applied but is also eta-expanded? E.g. `(\x1 -> f x1) e1 e2` for `f` whose arity equals `2`?

Lots of questions...

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,186 @@
...
+
+ Since UPLC programs are created from erasing the types of TPLC programs (see
+ "PlutusCore.Compiler.Erase") we will consider TPLC terms of the following structure:
+
+ @
+   /\T1 -> \X1 -> /\T2 -> \X2 -> /\T3 -> \X3 -> ... -> /\Tn -> \Xn -> body
+ @
+ where @T1 ... Tn@ are lists of type variables (e.g. @T1@ could be @[t, q, p]@)
+ and @X1 ... Xn@ are lists of term variables. Of course, each @/\@ and @\@ here would
+ desugar to a sequence of type/term abstractions.
```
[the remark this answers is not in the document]
ana-pantilie: Good point, thanks, I wasn't aware of the `Readble` format.

These terms are already pretty hard to read and using a different notation would make it worse. I'm mentioning, as you suggested, that it's pseudo-code.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
michaelpj: > So since we have a separate force-delay pass anyway, why do we insist on finding Delay here?

I agree that in general our approach is for a pass to do One Thing and let other passes do their One Thing. So I would indeed be inclined to let `ForceDelay` do the cancellation.

> This transformation is correct even if body isn't a delay, right?

I think so. 

> maybe we can implement an additional optimization that removes an outer delay from a binding if that binding is always accompanied by a force at all call sites within a program.

That would also probably be useful, although a bit fiddly perhaps.

> We could also generalize that...

I think there's a simpler way of looking at this family of optimizations more generally. They're exactly let-floating.

```
force (apply (apply (\x1 -> body) e1) e2)
===
force (apply (let x1 = e1 in body) e2)
--->
let x1 = e1 
in force (apply body e2)
```
which makes it more obvious that you can float the "let" out of all sorts of contexts, not just forces but also applications etc. As usual, you have to worry about effects, but we already figured all this out for the PIR let-floater.

Perhaps it would be better to port the PIR let-floater instead of thinking of this as a special case where we "push in forces"?

> Speaking of which, @zliu41 @michaelpj does the callsite inliner recognize cases where a function is fully applied but is also eta-expanded? E.g. (\x1 -> f x1) e1 e2 for f whose arity equals 2?

I don't think so, no.

michaelpj:
Haven't reviewed the code yet because I just had the realisation that this optimization is basically just let-floating and I think we should discuss whether to implement it like that.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,186 @@
...
+ the context of them being applied to some sequence of terms.
+
+ One important observation is that this transformation requires that the underlying
+ (term)-lambda abstraction will be fully reduced by the applications.
+ For UPLC, this can happen only when the number of lambda abstracted variables is equal
+ to the number of terms to which it will be applied.
+ For example, @force (\x -> delay b) => (\x -> b)@ is invalid, since the former is @error@.
+ The other case, we can see that applying the optimisation modifies the end result:
+ @
+   force ((\x -> delay b) a1 a2) => (\x -> b) a1 a2 => b[x1/a1] a2
```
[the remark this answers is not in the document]
ana-pantilie: I don't feel strongly about this either, switched to `b[x1:=a1]`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
ana-pantilie: Sorry, I think I messed this up. Will fix, thanks.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
ana-pantilie: The tests I added should be pretty good examples. I can add a link to them, if you think that's suitable.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
ana-pantilie: Oh that's a good point, yes, "exactly reduced" is what I meant. I'll fix.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
ana-pantilie: I would agree with @michaelpj on this. In my experience, simplifiers can get _very_ tricky to understand if you're not careful. Keeping concerns as separate as possible is one way to make sure that each pass does it's own explicitly defined thing without any hidden complexity.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
ana-pantilie: > I think there's a simpler way of looking at this family of optimizations more generally. They're exactly let-floating.

Oh, that's a very good point which I didn't consider, thanks for the example. I'll look into the PIR let-floater approach.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
ana-pantilie: Actually I don't think I messed this up. The reason there are two parenthesis before each `force` is that we have `(force (something something))` applied to `Xi` so `((force (something something)) Xi)`. 
I omitted one of the parenthesis after, respectively before, the two `...`.

ana-pantilie: So, there should be two parenthesis before each `force`. Notice that in my fix `X3` is followed by a single parenthesis because we're not explicitly showing the application to `X4`. I hope this makes sense, it's confusing for me as well so I'm not ruling out being wrong about this as well.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
effectfully: > Hmm, do you think it would be better to switch to using unicode symbols? Haddock supports unicode, right?

I'd say let's not do unicode in non-agda files. You can simply use lowercase `t` for type variables and uppercase `T` for types for example. Anyway, I'm just nitpicking, feel free to ignore.

effectfully:
Before I review the code, let me summarize what I was saying about "viewing an interleaved application as an iterated `let` is the very algorithm that we need to implement here".

Say we have the following typed expression:

```
(/\a -> \(x : a) -> /\b -> \(y : b) -> body) @A X @B Y
```

where type and term applications are interleaved and type application is denoted as `@` (Haskell-style).

We can turn that into an iterated `let` assuming that no typing problems exist (i.e. `let`s are transparent and `a` means the same thing as `A` given `type a = A`) and making up some syntax for the occasion:

```
let
  type a = A
  term x : a = X
  type b = B
  term y : b = Y
in body
```

Now what will those `type` bindings become if we move from the typed world to the untyped one? Well, we'll simply eliminate them, i.e. we'll get

```
let
  term x = X
  term y = Y
in body
```

Now we expand this back into a lambda expression:

```
(\x -> \y -> body) X Y
```

and we're done.

That's the whole idea. Implement the `force-apply` for UPLC the same way you'd implement turning interleaved application into an iterated `let`, except drop type bindings along the way.

Does that make sense? Is it similar to what the code does already? I'll review the code, just can't get to it, sorry.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,189 @@
...
+    t -> t
+
+{- | Converts the subterm of a 'Force' into specialised types for representing
+ multiple applications on top of multiple abstractions. Checks whether the lambda
+ will eventually get "exactly reduced" and applies the optimisation.
+ Returns 'Nothing' if the optimisation cannot be applied.
+-}
+optimisationProcedure :: Term name uni fun a -> Maybe (Term name uni fun a)
+optimisationProcedure term = do
+    asMultiApply <- toMultiApply term
```
effectfully, line 140: Given my reasoning above, do you need that `toMultiApply` logic? Can you just match `Apply`-`LamAbs` and `Force`-`Delay` as they appear in the AST?

plutus-tx-plugin/test/AsData/Budget/9.6/patternMatching.uplc.golden
```diff
@@ -35,10 +35,10 @@ program
                                   (addInteger cse))
                                (addInteger cse))
                             (addInteger cse))
-                         (case cse [(\x y z w -> w)]))
+                         (case cse [(\x y z w -> x)]))
                       (case cse [(\x y z w -> y)]))
                    (case cse [(\x y z w -> z)]))
-                (case cse [(\x y z w -> x)]))
+                (case cse [(\x y z w -> w)]))
```
effectfully, line 41: This reordering bothers me a bit. Are we sure it's safe? I think it is, but it would be good to make sure. @zliu41 what do you think?

effectfully (commented):
Lots of nitpicking, sorry. There's a couple of more important points though:

1. this isn't as general as it could be and I feel like the more general version wouldn't be any harder
2. this doesn't remove `Force`-`Delay` when they're next to each other, but removes them when they're apart and I feel like that only complicates code and that merging the force-delay pass into this one will allow us to have one fewer pass while at the same time simplifying this one

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,189 @@
...
+        Delay _ subTerm ->
+            let optimisedInnerMultiAbs = innerMultiAbs { rhs = subTerm}
+                optimisedMultiApply =
+                    asMultiApply { toApply = fromMultiAbs optimisedInnerMultiAbs }
+            in pure . fromMultiApply $ optimisedMultiApply
+        _               -> Nothing
+
+data MultiApply name uni fun a = MultiApply
+    { toApply     :: Term name uni fun a
+    , applyToIter :: [(a, Term name uni fun a)]
```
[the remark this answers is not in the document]
ana-pantilie: I hated the original name, thanks for the good idea!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,189 @@
...
+    guard $ length (applyToIter asMultiApply) == length (absVars innerMultiAbs)
+    case rhs innerMultiAbs of
+        Delay _ subTerm ->
+            let optimisedInnerMultiAbs = innerMultiAbs { rhs = subTerm}
+                optimisedMultiApply =
+                    asMultiApply { toApply = fromMultiAbs optimisedInnerMultiAbs }
+            in pure . fromMultiApply $ optimisedMultiApply
+        _               -> Nothing
+
+data MultiApply name uni fun a = MultiApply
```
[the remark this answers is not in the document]
ana-pantilie: Hmmm, I think `MultiApply` is clearer. `HeadSpineRev` makes me ask "what is this the head-spine-rev of?" :slightly_smiling_face: But I do agree that the field names should change as you suggested.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceApply.hs
```diff
@@ -0,0 +1,189 @@
...
+ will eventually get "exactly reduced" and applies the optimisation.
+ Returns 'Nothing' if the optimisation cannot be applied.
+-}
+optimisationProcedure :: Term name uni fun a -> Maybe (Term name uni fun a)
+optimisationProcedure term = do
+    asMultiApply <- toMultiApply term
+    innerMultiAbs <- toMultiAbs . toApply $ asMultiApply
+    guard $ length (applyToIter asMultiApply) == length (absVars innerMultiAbs)
+    case rhs innerMultiAbs of
+        Delay _ subTerm ->
```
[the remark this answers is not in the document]
ana-pantilie: > We still insist on encountering a Delay here even though it's not really necessary? Or do you disagree that it's not necessary?

For this branch in the `case` it is necessary because we're removing the `Force`.

You do have a point I didn't consider: what happens if instead of returning `Nothing` in the non-delay case we would push the `Force` down? (Which I think is what you were trying to tell me but I thought it meant "keep the algorithm the same but don't match on `Delay`") .

A quick look at this, unfortunately, makes a few of the benchmark test outputs a bit worse. For example, I am getting the following diff for `checkScriptContextEqualityTerm-20`:
```
-({cpu: 304395100
 -| mem: 1215868})

+({cpu: 304533100
+| mem: 1216468})
```
No changes to any golden term files.

I think, though, that what you suggest here is part of a different class of optimisations. As we discussed during the technical meeting, I think the "pushing T down without eliminating it" where T is either Force, Constr, or something else for which it's sound to "float lets out of" in UPLC, is a different optimisation altogether. I think we need a spike for investigating them. After that I will be much more confident to say if we can somehow merge all of them into a single optimisation.

> This pass doesn't do the job of eliminating a Force-Delay pair if they're next to each other, but eliminates them if they're a few Apply-LamAbs pairs apart. 

I can move the force-delay cancelling into this module, that's fine and after thinking about it a good idea because it's essentially the case where the apply is empty and the lambda is empty. :+1: 

> I'd say pick one: either you always eliminate Force-Delay pairs or you let the other pass clean them up
Could doing the job of the force-delay pass here make not only allow us to remove the latter, but also simplify this pass as well? You have that Maybe indirection in toMultiApply and toMultiAbs -- what if you simply removed the Maybe, would it still result in correct logic despite providing more functionality and requiring less code?

That wouldn't work. You're saying that this pass should only push forces down, correct? Say you have: 
```
force (force [(\x1 -> delay [(\x2 -> delay [(\x3 -> body) 5]) 7]) 9])
==> (push inner force)
force [(\x1 -> force (delay [(\x2 -> delay [(\x3 -> body) 5]) 7]) 9]
==> (push outer force)
[(\x1 -> force (force (delay [(\x2 -> delay [(\x3 -> body) 5]) 7])) 9]
```
...and you get stuck, because you no longer have a `force` applied to an application of a lambda.

Let's say that then you apply a different pass which removes forces immediately followed by delays, you get:
```
[(\x1 -> force [(\x2 -> delay [(\x3 -> body) 5]) 7]) 9]
```
Now you need to run the "push" pass again!

**My point is that being able to push multiple forces in _one_ pass requires that you also remove the force-delay pairs along the way.**

Otherwise you have to do an arbitrary number of traversals of the term, which we agreed is not ideal. This is the reason why we can't generalise this optimisation further into some general "let-floating" algorithm, and why we _need_ to have something separate for `force` and `delay` interleaving.

I hope this makes sense, let me know what you think.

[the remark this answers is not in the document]
effectfully: > A quick look at this, unfortunately, makes a few of the benchmark test outputs a bit worse. For example, I am getting the following diff for `checkScriptContextEqualityTerm-20`:

That is weird... Thanks a lot for doing the analysis, could you please mention this in the comments in the code, so that we don't forget why we didn't implement the more general algorithm?

> I think, though, that what you suggest here is part of a different class of optimisations. As we discussed during the technical meeting, I think the "pushing T down without eliminating it" where T is either Force, Constr, or something else for which it's sound to "float lets out of" in UPLC, is a different optimisation altogether.

I think it's the same optimization, but YMMV.

> My point is that being able to push multiple forces in one pass requires that you also remove the force-delay pairs along the way.

Great analysis, thank you for elaborating, I now see that what I proposed doesn't make much sense.

effectfully (approved):
Great work! Sorry for bothering you for so long with a lot of stupid comments.

plutus-tx-plugin/test/AsData/Budget/9.6/patternMatching.uplc.golden
zliu41: It is safe in this instance, but it is weird, because the new transformation has no other effect on this test case except the reordering. If there's no opportunity for `ForceApply`, shouldn't it be a no-op?

@ana-pantilie do you have a guess for the reason of the reordering? Maybe you can print the programs before and after the ForceApply transformation and see how it changed.

zliu41 (commented):
Good work!

plutus-tx-plugin/test/AsData/Budget/9.6/patternMatching.uplc.golden
ana-pantilie: @zliu41 @effectfully I've pinned it down to being an interaction between `floatDelay`, `inline` and this optimisation. What I mean is that I have these results:
```
only old forceDelay
vs.
only new forceDelay
==> no change

old forceDelay && inline
vs.
new forceDelay && inline
==> no change

old forceDelay && floatDelay
vs.
new forceDelay && floatDelay
==> no change

old forceDelay && floatDelay && inline
vs
new forceDelay && floatDelay && inline
==> the exact change you see in this PR's diff
```

I didn't mess around with the order, it's always `floatDelay >=> forceDelay >=> ... >=> inline`.

I don't understand why, though, do you have any idea what might be happening? In any case, I can look into this more deeply tomorrow.

zliu41: Good investigation. Digging in further would be optional. I can imagine that `forceDelay` modifies the program in some way, but after `floatDelay` and `inline`, all that's left to be seen is the reordering.

If you do want to investigate further, you can try printing the program before and after `forceDelay`, and see what exactly is changed by `forceDelay`. One easy way to make a `term :: Term name uni fun ann` printable is to do `unsafeCoerce (void term) :: Term Name DefaultUni DefaultFun ()`, which can then be printed via `display . AsReadable`.

ana-pantilie: Cool, thanks for the trick! I am having some trouble setting things up to just run my test, though, because this code runs at compile time and I get tons of unrelated output from other tests I believe. I'm going to merge this in the meantime.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceDelay.hs
```diff
@@ -1,16 +1,215 @@
+{- | The 'ForceDelay' optimisation pushes 'Force' inside its direct 'Apply' subterms,
```
michaelpj, line 1: Nit: we usually write long explanatory comments as Notes, rather than Haddock. I guess my general feeling is that Haddock should be useful for someone trying to use the module or functions, whereas Notes are for the implementors.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceDelay.hs
```diff
@@ -1,16 +1,215 @@
...
 
+{- | Checks whether the term is of the right form, and "pushes"
+ the 'Force' down into the underlying lambda abstractions.
+-}
 processTerm :: Term name uni fun a -> Term name uni fun a
 processTerm = \case
     Force _ (Delay _ t) -> t
-    t                   -> t
+    original@(Force _ subTerm) ->
+        fromMaybe original (optimisationProcedure subTerm)
```
michaelpj, line 156: https://github.com/IntersectMBO/plutus/blob/master/STYLEGUIDE.adoc#prefer-pattern-matching

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/ForceDelay.hs
```diff
@@ -1,16 +1,215 @@
...
+                    asMultiApply { appHead = fromMultiAbs optimisedInnerMultiAbs }
+            in pure . fromMultiApply $ optimisedMultiApply
+        _               -> Nothing
+
+data MultiApply name uni fun a = MultiApply
+    { appHead     :: Term name uni fun a
+    , appSpineRev :: [(a, Term name uni fun a)]
+    }
+
+toMultiApply :: Term name uni fun a -> Maybe (MultiApply name uni fun a)
```
michaelpj, line 182: I'm surprised we don't have this somewhere else already :thinking:

michaelpj (commented):
sorry for the late review, looks good!

PR #5800, opened 2024-02-22
fix: another haddock parse error in GHC 8.10

Unisay:
Fixes this failure https://github.com/IntersectMBO/cardano-haskell-packages/actions/runs/8004482180/job/21862427753


michaelpj (approved):
oof

PR #5802, opened 2024-02-23
V3.TxInfo uses V3.TxId consistently.

Unisay:
in particular, `GovernanceActionId` newtype now wraps around the `V3.TxId` (used to wrap around the `V2.TxId`)


michaelpj:
This affects some benchmark output, need to update it.

Issue #4306, opened 2021-12-31
Internals of builtins, roadmap


effectfully:
This issue is for me to dump all the plans regarding builtins in a largely unstructured way. It is expected that properly structured documentation will come out of this eventually.

effectfully:
This comment outlines the general structure of the implementation of builtins. Copied directly from a slack discussion.

The builtins machinery consists of these pieces:

1. a Plutus-unspecific [library for packing up types together into universes](https://github.com/input-output-hk/plutus/blob/73595eb828c1af7ae103b1e0c8a94b9062c8019f/plutus-core/plutus-core/src/Universe/Core.hs) and doing stuff with them. It does not depend on anything Plutus-related, but it assumes that the user is going to need polymorphic built-in types (otherwise it would be simpler)
2. [`TypeScheme`](https://github.com/input-output-hk/plutus/blob/73595eb828c1af7ae103b1e0c8a94b9062c8019f/plutus-core/plutus-core/src/PlutusCore/Constant/Typed.hs#L84) in `plutus-core/plutus-core/src/PlutusCore/Constant/Typed.hs` -- this specified the signature of a builtin. The signature can be used for getting the Plutus type of the builtin (used during type checking of Plutus programs) and it can be used for evaluation ("does a partially applied builtin expect a type or a term argument or is the builtin fully applied?")
3. [`KnownTypeAst`](https://github.com/input-output-hk/plutus/blob/73595eb828c1af7ae103b1e0c8a94b9062c8019f/plutus-core/plutus-core/src/PlutusCore/Constant/Typed.hs#L477) in `plutus-core/plutus-core/src/PlutusCore/Constant/Typed.hs` -- allows you to convert a Haskell type to the corresponding Plutus one, e.g. `Integer` ~> `con integer`, `EvaluationResult Bool` ~> `con bool` (note how `EvaluationResult` has disappeared -- the `Error` effect is implicit in Plutus!). This allows us to only specify the Haskell type of a builtin and the Plutus one will be inferred automatically and used during type checking of Plutus programs
[...]

PR #5806, opened 2024-02-27
[Builtins] Store 'BuiltinRuntime' lazily explicitly

effectfully:
This allows us to be more precise regarding how `BuiltinRuntime`s are stored and it allows us to implement a proper `NoThunks` instance for `BuiltinsRuntime` at the cost of introducing an indirection that slightly slows things down at runtime (may be undetectable, but let's see).

This is an EXPERIMENT, do not merge.


effectfully:
^ -3.15% on average lol.

effectfully:
^ -2.18%. I don't believe it.

image

effectfully:
^ -2.84%. Ok, I'll look into it. I don't trust it still.

effectfully:
^ +1.76% on average.

kwxm:
> ^ -2.18%. I don't believe it.

Quick!  Merge it!

effectfully:
I thought this wouldn't be compatible with extensible cost models in the end, but it is compatible. Still GHC-8.10 doesn't like something. I'm closing this for now and I've added an entry about this issue to #4306 in case we want to return to it in future.

PR #5808, opened 2024-02-27
Update iohk-nix to include the new blst version.

angerman:


michaelpj:
Windows cross seems to be broken? Perhaps we need a newer haskell.nix?

michaelpj:
I made https://github.com/IntersectMBO/plutus/pull/5809 to bump everything

kwxm:
Thanks.  I think this'll fix the problem that I had earlier.  It'll also let us get up-to-date costs for the BLS operations.

angerman:
fixed in #5809 instead.

PR #5811, opened 2024-02-28
Yet more benchmarking experiments

kwxm:
More experiments, carrying on from #5788.

DON'T MERGE!!


kwxm:
I've probably got enough data now.  The base branch for  this has vanished as well.

PR #5812, opened 2024-02-28
Added PlutusTx.SortedMap

bezirg:
DO NOT LOOK HERE YET @paluh
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


colll78:
Please add a `tryLookup` function. Typically, the use of optional types (i.e. Maybe and similar constructs) is considered an anti-pattern in most production onchain codebases. Often you want fast fail semantics and even in cases that you don't it is common to handle the not-found path with a continuation functions. In general, please expose 'try' variants for any functions which return optional types in the standard library because in practice the optional variants are almost never used.

Also please add the same for all other iterations of `Map` or structures with a `lookup` or `find` function. If you look at any production smart contract codebase you will almost never see the `Maybe` or sum type variants of such functions used. In practice, x-units optimization is simply too important.

PR #5813, opened 2024-03-01
[Evaluation] Use 'lookupInline'

effectfully:
Don't look here yet.


kwxm:
^ That's kind of what I was expecting, but I thought it was worth checking.  Most of the work in these is spent on CPU-intensive library functions.

effectfully:

PR #5816, opened 2024-03-04
Vectors in UPLC 

michaelpj:
Don't look here

### DIFF
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Eq.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Eq.hs
@@ -21,6 +21,7 @@ import Universe
 
 import Data.Foldable (for_)
 import Data.Hashable
+import Data.Vector qualified as V
 
 instance (GEq uni, Closed uni, uni `Everywhere` Eq, Eq fun, Eq ann) =>
             Eq (Term Name uni fun ann) where
@@ -34,6 +35,10 @@ type HashableTermConstraints uni fun ann =
   , Hashable ann
   , Hashable fun
   )
+
+-- TODO: dubious
+instance Hashable a => Hashable (V.Vector a) where
+  hashWithSalt s = hashWithSalt s . toList
 
 instance HashableTermConstraints uni fun ann => Hashable (Term Name uni fun ann)
 
@@ -94,13 +99,13 @@ eqTermM (Error ann1) (Error ann2) = eqM ann1 ann2
 eqTermM (Constr ann1 i1 args1) (Constr ann2 i2 args2) = do
     eqM ann1 ann2
     eqM i1 i2
-    case zipExact args1 args2 of
+    case zipExact (toList args1) (toList args2) of
         Just ps -> for_ ps $ \(t1, t2) -> eqTermM t1 t2
         Nothing -> empty
 eqTermM (Case ann1 a1 cs1) (Case ann2 a2 cs2) = do
     eqM ann1 ann2
     eqTermM a1 a2
-    case zipExact cs1 cs2 of
+    case zipExact (toList cs1) (toList cs2) of
         Just ps -> for_ ps $ \(t1, t2) -> eqTermM t1 t2
         Nothing -> empty
 eqTermM Constant{} _ = empty
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Eq.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Eq.hs
@@ -36,7 +36,8 @@ type HashableTermConstraints uni fun ann =
   , Hashable fun
   )
 
--- TODO: dubious
+-- This instance is the only logical one, and exists also in the package `vector-instances`.
+-- Since this is the same implementation as that one, there isn't even much risk of incoherence.
 instance Hashable a => Hashable (V.Vector a) where
   hashWithSalt s = hashWithSalt s . toList
 
@@ -99,7 +100,7 @@ eqTermM (Error ann1) (Error ann2) = eqM ann1 ann2
 eqTermM (Constr ann1 i1 args1) (Constr ann2 i2 args2) = do
     eqM ann1 ann2
     eqM i1 i2
-    case zipExact (toList args1) (toList args2) of
+    case zipExact args1 args2 of
         Just ps -> for_ ps $ \(t1, t2) -> eqTermM t1 t2
         Nothing -> empty
 eqTermM (Case ann1 a1 cs1) (Case ann2 a2 cs2) = do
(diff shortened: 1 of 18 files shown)


effectfully:
@michaelpj I don't think we need vectors in `VConstr` (as opposed to `FrameCases`), but removing them doesn't help performance according to the number above.

michaelpj:
> I don't think we need vectors in VConstr (as opposed to FrameCases)

Yes, you're right, I was just being overly zealous there.

michaelpj:
We also don't need them in `Constr`. But I don't think it costs us much, we still have O(1) performance on `V.uncons`.

michaelpj:
Still looks a bit better.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Eq.hs
```diff
@@ -35,6 +36,10 @@ type HashableTermConstraints uni fun ann =
   , Hashable fun
   )
 
+-- TODO: dubious
```
michaelpj, line 39: Not sure what to do about this. I could remove it, but then I would need to manually write an instance for `Term`, which would be okay but kind of annoying.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -763,12 +762,13 @@ enterComputeCek = computeCek
...
-        case todo of
-          (next : todo') -> computeCek (FrameConstr env i todo' done' ctx) env next
-          _              -> returnCek ctx $ VConstr i done'
+        case V.uncons todo of
+          Just (next, todo') -> computeCek (FrameConstr env i todo' done' ctx) env next
+          Nothing            -> returnCek ctx $ VConstr i done'
     -- s , case _ (C0 ... CN, ρ) ◅ constr i V1 .. Vm  ↦  s , [_ V1 ... Vm] ; ρ ▻ Ci
     returnCek (FrameCases env cs ctx) e = case e of
-        (VConstr i args) -> case cs ^? wix i of
+        -- TODO: handle word/int conversion better
```
michaelpj, line 770: Also unsure what to do here. Check if the `Integer` is in range, otherwise assume it's a miss?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Eq.hs
effectfully: Could use [vector-instances](https://hackage.haskell.org/package/vector-instances). Although given that it implements the instance the same way, I think simply keeping the instance is fine. Worth a comment though.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -572,9 +571,9 @@ data Context uni fun ann
     | FrameForce !(Context uni fun ann)
     -- ^ @(force _)@
     -- See Note [Accumulators for terms]
-    | FrameConstr !(CekValEnv uni fun ann) {-# UNPACK #-} !Word64 ![NTerm uni fun ann] !(ArgStack uni fun ann) !(Context uni fun ann)
+    | FrameConstr !(CekValEnv uni fun ann) {-# UNPACK #-} !Word64 !(V.Vector (NTerm uni fun ann)) !(ArgStack uni fun ann) !(Context uni fun ann)
```
effectfully, line 574: I still don't like having `Vector` here. It doesn't seem to affect performance in a detectable way and is used entirely as `[]`, so why not just `[]`. I think it's a pointless complication.
I'm not going to die on this hill though.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: It's _probably_ fine, because you're only going to get a negative integer if an overflow occurs and that will be caught by `(V.?!)`, but this really does need some extensive manual testing.

You could also do what I did in the `SmallArray` PR:

```
        VConstr i args
            | i < unsafeCoerce (sizeofSmallArray cs) ->
                computeCek (transferArgStack args ctx) env . indexSmallArray cs $ unsafeCoerce i
```

where `indexSmallArray` is unsafe lookup and `sizeofSmallArray` stupidly returns the length of the array rather than its byte size (man it pisses me off that they literally gave up on the usual camelCasing and stole the name of a C primitive just to assign it a completely different meaning). The logic here is

- we know that the length of the array is a non-negative `Int`, so it's safe to `unsafeCoerce` it to a `Word64`
- once we know that `i` is within the array bounds, it's safe to `unsafeCoerce` it to an `Int`, unless the array is somehow larger than `fromIntegral (maxBound :: Int) :: Word64`, which isn't something we could deserialize

Please note that I absolutely do not propose to have `unsafeCoerce` in there, just explaining the reasoning. Replacing both with `fromIntegral` should do the job just fine.

effectfully (commented):
It's beyond my understanding how this can be faster than `SmallArray`.

effectfully:
> It's beyond my understanding how this can be faster than `SmallArray`.

Hm, maybe I should try `Array`-based solution then. Maybe something's wrong with `SmallArray`. Anyway, feel free to merge this one. Still don't like `Vector` in `Constr` though.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
michaelpj: > It's probably fine, because you're only going to get a negative integer if an overflow occurs and that will be caught by (V.?!), but this really does need some extensive manual testing.

Yeah no, you can still wrap around to positive again.

I'll add some kind of bounds check.

michaelpj: Okay, so it can in fact never trigger, because `i` is a `Word64` and the largest `Word64` wraps to only `-1` as an `Int64`, so can't produce an "apparently good" value. I've left the check in out of caution.

I wanted to write a conformance test for this, but I was blocked by the fact that the parser doesn't do range checks, so in fact already wraps a "too large" number around problematically :joy: but that's a parser bug.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
michaelpj: Reverted to only do this for case.

michaelpj:
I'm unsure if that's a real result or if the changes to `Constr` actually did help!

michaelpj:
Okay, close enough that I do doubt that the constr change was adding anything. We can always do it on its own later.

PR #5817, opened 2024-03-04
CIP-0057 Howto: exporting a Plutus contract blueprint

Unisay:
This document describes functionality implemented in the adjacent PR https://github.com/IntersectMBO/plutus/pull/5772


zliu41 (changes requested):
This will need to be updated to account for https://github.com/IntersectMBO/plutus/pull/5820 and https://github.com/IntersectMBO/plutus/pull/5824

doc/read-the-docs-site/howtos/exporting-a-blueprint.rst
```diff
@@ -0,0 +1,307 @@
...
+Haskell type from which its derived with a corresponding annotation:
+
+* ``SchemaTitle`` 
+* ``SchemaDescription``
+* ``SchemaComment``
+
+For example, to add a title and description to the ``MyParams`` type,
+we can use the ``SchemaTitle`` and ``SchemaDescription`` annotations::
+
+  {-# ANN type MyParams (SchemaTitle "Title for the MyParams definition") #-}
```
zliu41, line 256: If you put the title in an annotation, then what do you put in the `parameterDescription` field? `Nothing`?

zliu41 (commented):
The code should be checked by CI. You can put the code in a Haskell module, like what `simple-example.rst` does.

Also we still don't have a changelog - let's add one in this PR.

doc/read-the-docs-site/howtos/exporting-a-blueprint.rst
Unisay: Yes

zliu41: Ok, we should explain this in the doc (with an example would be great)

Unisay: I have misunderstood your question originally. The correct answer is that 
annotations are attached to constructors and then appear in the comment/titile/description fields in the schema definitions. The contract parameter and argument blocks have their own comment/title/description fields and those could be specified as literal values.

zliu41: I'm not sure I understand. Let's just use your example, in which you have:

```haskell
    , validatorParameters =
        [ MkParameterBlueprint
            { parameterTitle = Just "My Validator Parameters"
            , parameterDescription = Just "Compile-time validator parameters"
            , parameterPurpose = Set.singleton Spend
            , parameterSchema = definitionRef @MyParams
            }
        ]
```

But if you also have

```haskell
{-# ANN type MyParams (SchemaTitle "Title for the MyParams definition") #-}
```

then are the two titles, "My Validator Parameters" and "Title for the MyParams definition", both in the blueprint but in different places, or only one of them is chosen?

Unisay: Added paragraph about using `deriveParameterBlueprint` and `deriveArgumentBlueprint` TH functions.

doc/read-the-docs-site/howtos/exporting-a-blueprint.rst
```diff
@@ -0,0 +1,403 @@
...
+
+.. literalinclude:: Cip57Blueprint.hs
+   :start-after: BEGIN interface types
+   :end-before:  END interface types
+
+.. literalinclude:: Cip57Blueprint.hs
+   :start-after: BEGIN validator
+   :end-before:  END validator
+
+First of all we need to import required functionality:
```
michaelpj, line 33: personally I try to go through the blocks in an order that it is at least possible to put them in a file in. Yes, anyone who knows Haskell will no think that they can put the imports in after some data declarations... but I still think it reads a little confusingly.

doc/read-the-docs-site/howtos/exporting-a-blueprint.rst
```diff
@@ -0,0 +1,403 @@
...
+
+.. literalinclude:: Cip57Blueprint.hs
+   :start-after: BEGIN imports
+   :end-before:  END imports
+
+Next we define a contract blueprint value of the following type:
```
michaelpj, line 39: sigh, this should all just be haddock links, but we still have no good haddock hosting solution :(

doc/read-the-docs-site/howtos/exporting-a-blueprint.rst
```diff
@@ -0,0 +1,403 @@
...
+      , validatorRedeemer =
+          $(deriveArgumentBlueprint ''MyRedeemer (Set.fromList [Purpose.Spend, Purpose.Mint]))
+      , ... elided
+      }
+
+
+Result
+------
+
+Here is the full `CIP-0057`_ blueprint produced by this "howto" example:
```
michaelpj, line 300: you could check this in as a standalone file and have a test that checks that it's actually the blueprint you get from the haskell file.

michaelpj: it looks like it's already checked in separately?

michaelpj (approved):
Nice!

doc/read-the-docs-site/howtos/exporting-a-blueprint.rst
Unisay: Actually, we have this hosted https://intersectmbo.github.io/plutus/master/plutus-tx/html/PlutusTx-Blueprint-Contract.html#t:ContractBlueprint

Should I link it instead of embedding?

doc/read-the-docs-site/howtos/exporting-a-blueprint.rst
Unisay: Last time I tried to do this I hit this https://github.com/input-output-hk/iogx/pull/45/files

but now as it has been merged I'll try it again (after updating iogx)

zliu41 (approved):

PR #5818, opened 2024-03-04
[AST] Add 'SixList'

effectfully:
Another one of those.


effectfully:
I hate programming.

effectfully:
^ about the same as #5816.

effectfully:

PR #5820, opened 2024-03-07
CIP-0057 Annotations for the PlutusTx.Blueprint.Schema

Unisay:
This pull request adds annotations to the PlutusTx.Blueprint.Schema module to provide additional information about schema definitions.

Supersedes https://github.com/IntersectMBO/plutus/pull/5819


plutus-tx/src/PlutusTx/Blueprint/Schema.hs
```diff
@@ -136,22 +140,11 @@ instance ToJSON (Schema referencedTypes) where
...
     infoFields :: SchemaInfo -> Aeson.Object
-    infoFields MkSchemaInfo{title, description, comment} =
+    infoFields (Set.toList -> info) =
       KeyMap.empty
-        & optionalField "title" title
-        & optionalField "description" description
-        & optionalField "$comment" comment
-
--- | Additional information optionally attached to any datatype schema definition.
-data SchemaInfo = MkSchemaInfo
```
michaelpj, line 146: I'd have been tempted to keep this and write a `Set SchemaAnn -> SchemaInfo` function. In particular, it seems to me that that function should fail if you have duplicate annotations, whereas currently it will arbitrarily pick one, which seems bad.

plutus-tx/src/PlutusTx/Blueprint/TH.hs
```diff
@@ -40,10 +45,15 @@ makeHasSchemaInstance dataTypeName indices = do
...
-      Nothing    -> fail $ "No index given for constructor " ++ show (TH.constructorName ctorInfo)
+  indexedCons :: [(TH.ConstructorInfo, SchemaInfo, Natural)] <- do
+    typeSchemaInfo <- lookupSchemaInfo dataTypeName
+    for (TH.datatypeCons dataTypeInfo) $ \ctorInfo -> do
+      let ctorName = TH.constructorName ctorInfo
+      case lookup ctorName indices of
+        Nothing -> fail $ "No index given for constructor " ++ show (TH.constructorName ctorInfo)
+        Just index -> do
+          ctorSchemaInfo <- lookupSchemaInfo ctorName
+          pure (ctorInfo, Set.union ctorSchemaInfo typeSchemaInfo, index)
```
michaelpj, line 56: Is this right? You inherit the annotations from the type onto the constructors? That doesn't seem right to me, and isn't obviously documented. It also IMO behaves wrongly if so: your handling of duplicated annotations is to pick one randomly, but here if anything the constructor annotation should override the type one.

plutus-tx/src/PlutusTx/Blueprint/TH.hs
```diff
@@ -83,12 +110,7 @@ mkSchemaClause ts ctorIndexes =
     let whereDecls = []
     TH.clause patterns (TH.normalB body) whereDecls
 
-  mkSchemaConstructor :: (TH.ConstructorInfo, Natural) -> TH.ExpQ
-  mkSchemaConstructor (TH.ConstructorInfo{..}, naturalToInteger -> ctorIndex) = do
+  mkSchemaConstructor :: (TH.ConstructorInfo, SchemaInfo, Natural) -> TH.ExpQ
```
michaelpj, line 113: where does the info for the type go? does it do nothing but apply to the constructors? write it down!

plutus-tx-plugin/test/Blueprint/Acme.golden.json
```diff
@@ -67,13 +67,17 @@
     "Datum": {
       "oneOf": [
         {
-          "$comment": "DatumLeft",
+          "title": "Datum",
```
michaelpj, line 70: seems wrong

plutus-tx-plugin/test/Blueprint/Acme.golden.json
```diff
@@ -67,13 +67,17 @@
...
+          "description": "DatumLeft",
```
michaelpj, line 71: seems wrong

michaelpj: I would expect something like
```
"title": "DatumLeft",
"description": "Description about the constructor, basically it's haddock"
```

michaelpj: But I can't really engage with this because you haven't written down your reasoning for assigning annotations in this way. Write it down!

plutus-tx/src/PlutusTx/Blueprint/TH.hs
Unisay: I have added a comment that explains this logic.

plutus-tx/src/PlutusTx/Blueprint/TH.hs
Unisay: Done.

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
```diff
@@ -49,27 +51,33 @@ newtype Bytes (phantom :: Type) = MkAcmeBytes BuiltinByteString
   deriving anyclass (AsDefinitionId)
 
 instance HasSchema (Bytes phantom) ts where
-  schema = SchemaBytes emptySchemaInfo{title = Just "SchemaBytes"} emptyBytesSchema
+  schema = SchemaBytes emptySchemaInfo { title = Just "SchemaBytes" } emptyBytesSchema
 
+{-# ANN type DatumPayload (SchemaComment "MkDatumPayload") #-}
```
zliu41, line 56: I think the name of this module is a bit confusing. We usually say "test lib" when we mean a module that provides helper functions for writing tests. E.g., we have `plutus-core-testlib`, `plutus-ledger-api-testlib` and `plutus-tx-testlib`. But this is just a specific test case.

On the other hand, things such as `goldenBlueprint` should belong to a testlib package.

plutus-tx/src/PlutusTx/Blueprint/TH.hs
```diff
@@ -40,10 +44,90 @@ makeHasSchemaInstance dataTypeName indices = do
...
-      Just index -> pure (ctorInfo, index)
-      Nothing    -> fail $ "No index given for constructor " ++ show (TH.constructorName ctorInfo)
+  indexedCons :: [(TH.ConstructorInfo, SchemaInfo, Natural)] <- do
+    typeSchemaAnns <- lookupSchemaAnns dataTypeName
+    for (TH.datatypeCons dataTypeInfo) $ \ctorInfo -> do
+      let ctorName = TH.constructorName ctorInfo
+      case lookup ctorName indices of
+        Nothing -> fail $ "No index given for constructor " ++ show (TH.constructorName ctorInfo)
+        Just index -> do
+          {- Its possible to annotate a type as well as its constructors with schema annotations.
```
zliu41, line 54: This is too much inline comment. We use Notes for long comments. Can you convert it to a Note?

zliu41 (approved):
I realized that https://github.com/IntersectMBO/plutus/pull/5772 was merged without a changelog, yet it should have one. Can you add the changelog in this PR?

plutus-tx/src/PlutusTx/Blueprint/TH.hs
```diff
@@ -40,10 +44,90 @@ makeHasSchemaInstance dataTypeName indices = do
...
+                  }
+                ]
+              },
+            ...
+            }
+
+            Such schema lists all constructors of a type with their respective annotations,
+            however, there is no place to put annotations for the type itself.
+
+            For this reason, we merge annotations for the type and its constructor:
```
michaelpj, line 86: What reason? the previous sentence doesn't justify _merging_ them. In particular, I don't see from this what's wrong with my suggestion: for _all_ of the type and the constructors, use the "title" field for the name, and the "description" field for the documentation. No merging.

plutus-tx/src/PlutusTx/Blueprint/TH.hs
```diff
@@ -40,10 +44,90 @@ makeHasSchemaInstance dataTypeName indices = do
...
+            data Or = MkLeft | MkRight
+
+            becomes:
+
+            { ...
+              "Or": {
+                "oneOf": [
+                  {
+                    "title": "Type: Or",
+                    "description": "Constructor: MkLeft",
```
michaelpj, line 100: I'm sorry, this still reads like nonsense to me. To me, "title/description" means "name/description". In what way is the "name" of the constructor "Or"? It just isn't! It's "MkLeft"!

plutus-tx/src/PlutusTx/Blueprint/TH.hs
```diff
@@ -40,10 +44,90 @@ makeHasSchemaInstance dataTypeName indices = do
...
+                  }
+                ]
+              },
+            ...
+            }
+
+            Merging annotations for the type and its constructor also allows to keep things DRY
+            if the same attribute needs to be present in all 'oneOf' schemas:
+            Such annotation is attached to the type only once and in the schema
+            a corresponding attribute will be repeated for every constructor.
```
michaelpj, line 120: okay, this is fine I guess (what's the usecase?)

plutus-tx/src/PlutusTx/Blueprint/TH.hs
```diff
@@ -40,10 +44,90 @@ makeHasSchemaInstance dataTypeName indices = do
...
+
+            Beware: if the same annotation is attached to the type and its constructor,
+            then an error is reported in compile time, so this is wrong:
```
michaelpj, line 123: why not do overriding?

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
Unisay: I have extracted `goldenJson` function out of the `goldenBlueprint`  and moved it to the `Tests/Lib` module, where it gets higher chance of re-use.

plutus-tx/src/PlutusTx/Blueprint/TH.hs
Unisay: Ok

plutus-tx/src/PlutusTx/Blueprint/TH.hs
Unisay: > What reason? the previous sentence doesn't justify merging them. 

I didn't expand on the reason enough:
By "there is no place to put annotations for the type itself."  I meant "there is no place to put annotations for the type itself, and one would need to repeat a type-specific annotation for every constructor, which is counter-ergonomic and should be avoided".

Unisay: After discussing this more with @michaelpj we came to an agreement  to go forward without such functionality to merge type-level/ctor-level annotations. This was an attempt to work around the blueprint-schema limitation that "oneOf" can't have adjacent attributes: "title", "description", etc.

zliu41:
@Unisay We still need a changelog entry

PR #5824, opened 2024-03-08
CIP-0057: ContractBlueprint with derived schema definitions and safe schema refs

Unisay:
The idea is that user specifies at most 3 types: Param, Datum, Redeemer and all the types nested withing them are extracted automatically (un-rolled).

### DIFF
--- a/plutus-tx-plugin/test/Blueprint/Tests.hs
+++ b/plutus-tx-plugin/test/Blueprint/Tests.hs
@@ -1,20 +1,21 @@
 {-# LANGUAGE DataKinds         #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell   #-}
 {-# LANGUAGE TypeApplications  #-}
 
 module Blueprint.Tests where
 
 import Prelude
 
-import Blueprint.Tests.Lib (Datum, Params, Redeemer, goldenJson, serialisedScript)
+import Blueprint.Tests.Lib (Datum, Datum2, Param2a, Param2b, Params, Redeemer, Redeemer2,
+                            goldenJson, serialisedScript, validatorScript1, validatorScript2)
 import Data.Set qualified as Set
-import PlutusTx.Blueprint.Argument (ArgumentBlueprint (..))
 import PlutusTx.Blueprint.Contract (ContractBlueprint (..))
 import PlutusTx.Blueprint.Definition (definitionRef, deriveDefinitions)
-import PlutusTx.Blueprint.Parameter (ParameterBlueprint (..))
 import PlutusTx.Blueprint.PlutusVersion (PlutusVersion (PlutusV3))
 import PlutusTx.Blueprint.Preamble (Preamble (..))
 import PlutusTx.Blueprint.Purpose qualified as Purpose
+import PlutusTx.Blueprint.TH (deriveArgumentBlueprint, deriveParameterBlueprint)
 import PlutusTx.Blueprint.Validator (ValidatorBlueprint (..))
 import PlutusTx.Blueprint.Write (writeBlueprint)
 import Test.Tasty.Extras (TestNested, testNested)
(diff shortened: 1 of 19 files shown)


michaelpj:
I feel like I've lost the context again. Can we have a Note explaining what the deal is with nested types inside a schema and why we need to handle them in a special way?

plutus-tx/src/PlutusTx/Blueprint/Contract.hs
```diff
@@ -41,23 +41,65 @@ data ContractBlueprint (referencedTypes :: [Type]) = MkContractBlueprint
...
+This way blueprint authors can specify the top-level types used in a contract and the UnrollAll
+type family will take care of discovering all the nested types:
+
+  Blueprint '[Redeemer, Datum]
+
+  is equivalent to
+
+  ContractBlueprint '[Redeemer, MyStruct, Integer, Bool, ()]
+
+-}
```
Unisay, line 81: @michaelpj I hope this note helps to acquire a context.

plutus-tx-plugin/test/Blueprint/Tests.hs
```diff
@@ -10,92 +11,63 @@
...
-  , BuiltinData
-  , BuiltinByteString
-  ]
-
-contractBlueprint :: ContractBlueprint ValidatorTypes
+-- | All the data types exposed (directly or indirectly) by the type signature of the validator
+-- This type level list is used to:
+-- 1. derive the schema definitions for the contract.
+-- 2. make "safe" references to the [derived] schema definitions.
+contractBlueprint :: Blueprint [Fixture.Params, Fixture.Redeemer, Fixture.Datum]
```
zliu41, line 29: This is better than before but there's still repetition: you are specifying the param, redeemer and datum types here, but then have to repeat them in `parameterSchema = definitionRef @Fixture.Params`, `argumentSchema = definitionRef @Fixture.Redeemer` and `argumentSchema = definitionRef @Fixture.Datum`. It should be possible to require writing the only once.

plutus-tx/src/PlutusTx/Builtins/Internal.hs
```diff
@@ -10,6 +10,9 @@
 -- they're NOINLINE!
 {-# OPTIONS_GHC -O0 #-}
 
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
```
zliu41, line 13: What is this for?

Unisay: Without this option the pragma below is considered "unrecognized" and HLINT emits a warning:

image

plutus-tx-plugin/test/Blueprint/Tests.hs
Unisay: The thing is that one contract blueprint can describe more than one validator. Imagine the following contract, for example:
```haskell
data V1Params = MkV1Params Integer
type V1Redeemer = ()
type V1Datum = Integer
data V2Params = MkV2Params Bool
data V2Redeemer = RedeemThis | RedeemThat
data V2Datum = Bool

validator1 :: V1Params -> V1Redeemer -> V1Datum -> ScriptContext -> Bool
validator2 :: V2Params -> V2Redeemer -> V2Datum -> ScriptContext -> Bool
```

Current implementation allows to define

```haskell
contractBlueprint :: Blueprint [V1Params, V1Redeemer, V1Datum, V2Params, V2Redeemer, V2Datum]
```
and then reference corresponding schema from a validator blueprint, e.g. first `ValidatorBlueprint` references `V1*` and second one references `V2*`.

With such case I don't see how it could be possible to write top-level types (Param/Redeemer/Datum) once, as there will be a need to specify which top level type belongs to which validator. 

I can think of something along these lines:

```haskell
multipleValidatorContract :: Blueprint 
    [ [V1Params, V1Redeemer, V1Datum] -- first validator
    , [V2Params, V2Redeemer, V2Datum] -- second validator
    ]
    
type Blueprint1 p r d = Blueprint '[[p, r, d]]
singleValidatorContract :: Blueprint1 MyParams MyRedeemer MyDatum    
```
but such solution is more complex. 

@zliu41 @michaelpj wdyt, which design is better?

zliu41: Is it possible to make it so that writing `definitionRef @V2Params` etc. is sufficient to register `V2Params` as a used type, and avoid needing a type-level list at the top? You can do this separately.

Unisay: This is a good idea: a definition is registered by mere fact of referencing it - this is another way of making sure there are no broken references!

Unfortunately I don't yet know how an implementation of such feature could look like. What comes to mind is that every definition reference contributes a constraint, something like:
```haskell
definitionRef :: forall t ts. HasDefinition t ts => Schema ts


```
and then all such constraints are accumulated at the top level

```haskell
( HasDefinition Params ts
, HasDefinition Datum ts
, HasDefinition Redeemer ts
, HasDefinition Integer ts
, HasDefinition Bool ts
) => ...
```
and this set of constraints then drives the `deriveSchemaDefinitions` ...

I'd need to give this approach a "spike" try as I am not sure it can work.

Unisay: >  I'm suggesting to replace deriveSchemaDefinitions with something like [ deriveDefinition @Foo, deriveDefinition @Bar ].

With this approach I don't see how to make references that lack a definition entry fail in compile time.

Unisay: I found a way to make type inference work the way @michaelpj has described:
1. User lists top-level contract types;
2. Users's type is un-rolled to include all nested types;
3. Schema definitions are derived from the un-rolled list;
4. Schema references are checked against the un-rolled list;
5. The `referencedTypes` type variable in the `ContractBlueprint referencedTypes` which holds un-rolled list is existentially hidden using a wrapper data type;

Unisay:
I am going to merge this PR as it contains pieces that were already reviewed in other PRs and merged here. I believe that I have addressed all the expressed concerns. If any concerns remain, though, I'd be happy to address them in next PRs.

PR #5827, opened 2024-03-11
Agda html

ramsay-t:
This is a small PR to convert all of the metatheory files to Markdown - specifically, everything should now be `.lagda.md` and should be Markdown style Literate Agda. The files have consistent Markdown style code blocks and headings, but not a huge amount of actual text yet. 

The command `nix build ".#plutus-metatheory-site"` should convert the Agda code to HTML and then the markdown to HTML, resulting in a set of linked and formatted HTML documents. The auto-generated CSS is pretty basic, so this could be augmented before this gets integrated with a release process? The current script uses Jekyll but I have a working Pandoc based bash script if that is preferred.


michaelpj:
The build of the metatheory Haskell package (`cabal build plutus-metatheory`) is broken. I suspect it's this line: https://github.com/IntersectMBO/plutus/blob/master/plutus-metatheory/Setup.hs#L81, the suffix needs to change.

michaelpj:
Did you try it? It still doesn't work and I'm not sure why. It doesn't seem to be running the preprocessor at all for some reason.

michaelpj:
Ping @zeme-wana . I've tried various options for the preprocessor extension and it just mysteriously seems to never trigger? I'm very unsure what's going on here.

ramsay-t:
> Did you try it? It still doesn't work and I'm not sure why. It doesn't seem to be running the preprocessor at all for some reason.

You were right about the extension, but it now builds fine in nix on my mac... These hydra test builds all seem to just not bother with the preprocessor - it is producing dozens of errors, each of which is "Can't find file: " and then a MAlonzo file that should be built by the preprocessor.

michaelpj:
> You were right about the extension, but it now builds fine in nix on my mac... These hydra test builds all seem to just not bother with the preprocessor - it is producing dozens of errors, each of which is "Can't find file: " and then a MAlonzo file that should be built by the preprocessor.

Yes, the preprocessor is not running. I think you may find it doesn't work if you `cabal clean` first, you probably just have cached versions of the MAlonzo files.

michaelpj (approved):

PR #5831, opened 2024-03-13
CIP-0057: TH to derive argument and parameter blueprints.

Unisay:
- [x] Extend blueprint test fixture to use 2 validators.
- [x] Derive parameter blueprint using `deriveParameterBlueprint :: TH.Name -> Set Purpose -> TH.ExpQ`
- [x] Derive argument blueprint using `deriveArgumentBlueprint :: TH.Name -> Set Purpose -> TH.ExpQ`
- [x] Compile typed validators directly, without wrapping them as untyped.


plutus-tx-plugin/test/Blueprint/Tests.hs
```diff
@@ -39,35 +51,42 @@ contractBlueprint =
...
-            },
+        Set.fromList
+          [ MkValidatorBlueprint
+              { validatorTitle =
+                  "Acme Validator #1",
+                validatorDescription =
+                  Just "A validator that does something awesome",
+                validatorParameters =
+                  Just . pure $
+                    $(deriveParameterBlueprint ''Fixture.Params (Set.singleton Purpose.Spend)),
```
michaelpj, line 62: I'm unsure if this is good. People generally dislike TH and IME they're even less comfortable with it in expression positions. I guess it's cheap enough to have as an option.

plutus-tx-plugin/test/Blueprint/Tests.hs
```diff
@@ -39,35 +51,42 @@ contractBlueprint =
...
+              },
+            MkValidatorBlueprint
+              { validatorTitle =
+                  "Acme Validator #2",
+                validatorDescription =
+                  Just "Another validator that does something awesome",
+                validatorParameters =
+                  Just $
+                    $(deriveParameterBlueprint ''Fixture.Param2a (Set.singleton Purpose.Mint))
+                      :| [ $( deriveParameterBlueprint
```
michaelpj, line 78: Wait, is this a non-empty list? That seems wrong?

plutus-tx/src/PlutusTx/Blueprint/TH.hs
```diff
@@ -117,3 +120,33 @@ mkSchemaClause ts ctorIndexes =
   mkSchemaConstructor (TH.ConstructorInfo{..}, info, naturalToInteger -> ctorIndex) = do
     fields <- for constructorFields $ \t -> [|definitionRef @($(pure t)) @($(pure ts))|]
     [|SchemaConstructor info (MkConstructorSchema ctorIndex $(pure (TH.ListE fields)))|]
+
+deriveParameterBlueprint :: TH.Name -> Set Purpose -> TH.ExpQ
+deriveParameterBlueprint tyName purpose = do
+  title <- lookupAnn @SchemaTitle tyName
+  description <- lookupAnn @SchemaDescription tyName
+  [| MkParameterBlueprint
+      { parameterTitle = listToMaybe (Text.pack . schemaTitleToString <$> title)
```
michaelpj, line 129: Hang on, I though we agreed we were going to fail in general if we had several annotations? Again we are using `listToMaybe`?

michaelpj (approved):
Seems fine

plutus-tx/src/PlutusTx/Blueprint/TH.hs
Unisay: Good catch, this one slipped out of my attention (it was written before we agreed and git-stashed away).

plutus-tx-plugin/test/Blueprint/Tests.hs
Unisay: Fixed

plutus-tx/src/PlutusTx/Blueprint/Validator.hs
```diff
@@ -35,7 +35,7 @@ data ValidatorBlueprint (referencedTypes :: [Type]) = MkValidatorBlueprint
   -- ^ A description of the redeemer format expected by this validator.
   , validatorDatum        :: Maybe (ArgumentBlueprint referencedTypes)
   -- ^ A description of the datum format expected by this validator.
-  , validatorParameters   :: Maybe (NonEmpty (ParameterBlueprint referencedTypes))
+  , validatorParameters   :: [ParameterBlueprint referencedTypes]
```
michaelpj, line 38: We discussed this, but indeed I think this is a case where the `Nothing` case is not meaningfully different. Having no parameters and having an empty list of parameters is the same.

PR #5837, opened 2024-03-14
CIP-0057: Extend the Blueprint test case with the `AsData` example.

Unisay:
`AsData` types like this example one

```haskell
$( PlutusTx.asData
    [d|
      data Datum2 = MkDatum2
        { datum2integer :: Integer
        , datum2bool :: Bool
        }
        deriving stock (Generic)
        deriving newtype (AsDefinitionId)
      |]
 )
```
can't have their `HasSchema Datum2` instance derived using TH:

```haskell
$(makeIsDataSchemaIndexed ''Datum2 [('MkDatum2 0)])
-- Error: No index given for constructorBlueprint.Tests.Lib.Datum2_6989586621683919396typec
```

I have extended the test for Blueprint generation with the `AsData` type and showed how a manual `HasSchema` instance could be provided in such case.


Unisay:
I have an idea to try extending the `asData` TH macros to also generate an `HasSchema` instance (analogous to how `makeIsDataSchemaIndexed` adds a `HasData` instance on top of what `makeIsDataIndexed` does)
but I am not sure how important is it at this time to implement,  @zliu41 wdyt?

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
```diff
@@ -135,7 +137,36 @@ newtype Param2b = MkParam2b Bool
...
+        , datum2bool :: Bool
+        }
+        deriving stock (Generic)
+        deriving newtype (AsDefinitionId)
+      |]
+ )
+
+-- The TH splice above generates a 'Datum2' newtype wrapper around the 'BuiltinData'.
+-- Generically derived 'HasSchema Datum2' instance isn't useful as it describes such type
+-- as an opaque 'BuiltinData', so we provide a custom instance to show its structure.
```
michaelpj, line 153: What if we exposed a version of the TH machinery that took a declaration quote (like `asData` does)? Then if you _have_ a declaration quote you could pass it to both, i.e.

```
let d = [d| ... |]

$(PlutusTx.asData d)
$(Blueprint.mkSchema d)
```
or something?

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
```diff
@@ -135,7 +137,36 @@ newtype Param2b = MkParam2b Bool
...
+instance
+  ( HasSchemaDefinition Integer ts
+  , HasSchemaDefinition Bool ts
+  ) =>
+  HasSchema Datum2 ts
```
michaelpj, line 158: Okay, so you're manually giving an instance here to make the schema machinery work? It does seem like it would be nice if you could just give an instance for the type family that tells you what the child types are, are you sure we can't make that work?

plutus-tx-plugin/test/Blueprint/Acme.golden.json
```diff
@@ -58,7 +58,7 @@
       "datum": {
         "purpose": "mint",
         "schema": {
-          "$ref": "#/definitions/Integer"
+          "$ref": "#/definitions/Data"
```
michaelpj, line 61: why did this change?

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
Unisay: Currently the implementation uses a closed type family **with an overlap:**
- Some specific type indexes that don't have a generic instance (`Integer`, `Text`, etc.) -> RHS is defined manually;
- All generic type indexes -> RHS is derived generically.

The moment I try to make it open/associated I hit the "Conflicting family instance declarations" error. 
Related: https://gitlab.haskell.org/ghc/ghc/-/issues/4259

plutus-tx-plugin/test/Blueprint/Acme.golden.json
Unisay: Mistake, fixed.

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
Unisay: Caveat: GHC stage restriction applies;

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
Unisay: Its possible to use an open type family without an overlap if we're willing to give up being able to un-roll any generic type by default. I have demonstrated such alternative in this PR https://github.com/IntersectMBO/plutus/pull/5845

michaelpj: Overlapping instances aren't _necessarily_ bad in this case. I think it might work fine if you have `{-# OVERLAPPABLE -#}` on the `Generic` instance, then I think the specific instance should take priority? Or maybe we need to use typeclasses with an associated type for that?

Using a typeclass also has the advantage that the user can then do something like `deriving Unrollable via (Generically MyType)` in the datatype declaration, which looks a little tidier. 

Also, if we're going to expose `Unroll` to the user I think we should give it a more descriptive name. How about `SchemaTypes` or something?

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
```diff
@@ -125,17 +131,17 @@ newtype Param2a = MkParam2a Bool
...
 newtype Param2b = MkParam2b Bool
   deriving stock (Generic)
   deriving anyclass (AsDefinitionId)
 
-$(PlutusTx.makeLift ''Param2b)
+$(makeLift ''Param2b)
 $(makeIsDataSchemaIndexed ''Param2b [('MkParam2b, 0)])
 
-type Datum2 = Integer
+$(asData datum2)
```
michaelpj, line 144: Could we define a schema instance here in terms of the one for the non-asData type definition? Then you could keep things local.

plutus-tx-plugin/test/Blueprint/Tests/Lib/AsData/Decls.hs
```diff
@@ -0,0 +1,23 @@
+{-# LANGUAGE TemplateHaskellQuotes #-}
+
+{- | This module contains TH data type declarations from which 'AsData' declarations are derived.
+
+These declarations are used for two purposes:
+1. To generate an 'AsData' type declaration to be used in real validators.
+2. To generate a regular data type declaration to be used in a blueprint.
+
+Because of the GHC stage restriction, we have to keep these TH declarations in a separate module.
```
michaelpj, line 9: bah

michaelpj (approved):

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
michaelpj: i.e. import it qualified, then do
```
instance ... where
   schemaFunctionWhatever = BlueprintTypes.schemFunctionWhatever
```

plutus-tx-plugin/test/Blueprint/Tests/Lib/AsData/Decls.hs
Unisay: 🤷🏼‍♂️

plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
Unisay: The `HasSchema` instance is codegenerated with TH; As it is implemented atm its not possible to import the schema definition only `BlueprintTypes.schemFunctionWhatever`.

michaelpj: I don't understand. Yes, the instance that you generate in `Blueprint.hs` is generated. That does not mean that you can't _use_ that instance to define the instance for `Datum2` in `Lib.hs`?

Unisay: The module `Blueprint.Tests.Lib.AsData.Blueprint` uses TH to generate
```
instance (HasSchemaDefinition Integer referencedTypes,
          HasSchemaDefinition Bool referencedTypes) =>
         HasSchema Datum2 referencedTypes where
  {-# INLINABLE schema #-}
  schema
    = SchemaConstructor
        (MkSchemaInfo Nothing Nothing Nothing)
        (MkConstructorSchema
           0
           [definitionRef @Integer @referencedTypes,
            definitionRef @Bool @referencedTypes])
```
and you propose to use this implementation of `schema` to define another instance for the same type in the `Blueprint.Tests.Lib`  ?

Please help me better understand the [locality] problem you're trying to solve and the solution you're suggesting.

michaelpj: So at the moment you have to import the "schema-only" type here: https://github.com/IntersectMBO/plutus/pull/5837/files#diff-310cdc5eee57d553b6ea5fdc8b1590aa1333b9c6d8ef15577b51522517953c77R13

So that you can refer to that type instead of the actual type you are actually using. I'm suggesting to avoid this by giving the _actual_ `Datum2` type a `HasSchema` definition like so:

```
import Blueprint.Tests.Lib.AsData.Blueprint qualified as SchemaTypes

instance (HasSchemaDefinition Integer referencedTypes,
          HasSchemaDefinition Bool referencedTypes) =>
         HasSchema Datum2 referencedTypes where

  schema = schema @SchemaTypes.Datum2
```

now you can use the _actual_ `Datum2` everywhere because it has the right instnace.

Unisay: Now I see what you meant! Done here https://github.com/IntersectMBO/plutus/pull/5877

PR #5839, opened 2024-03-15
CIP-0057: Alternative approach to deriving and referencing schema definitons.

Unisay:
This PR implements @michaelpj 's [suggestion](https://github.com/IntersectMBO/plutus/pull/5824#discussion_r1524647268) to derive schema definitions "bottom-up" in contrast with the "top-down" derivation approach from [this PR](https://github.com/IntersectMBO/plutus/pull/5824#):
1. User lists top-level contract types, e.g.:
   `[Params, Redeemer, Datum]`.
2. User uses this list to derive schema definitions:
   `deriveDefinitions @[Params, Redeemer, Datum]`:
    - before deriving, top-level types are un-rolled to include all nested types, e.g.:    
       `@[Params, Bool, Integer, Redeemer, Datum, (), ...]`.
4. Schema references checked against the un-rolled list, e.g.:  
   - `definitionRef @Datum`
   - `definitionRef @Bool`
5. The `referencedTypes` type variable in the `ContractBlueprint referencedTypes` which holds un-rolled list is existentially hidden using a wrapper data type;

Lets chose which one we like more and merge it.

### DIFF
--- a/plutus-tx/src/PlutusTx/Blueprint/Contract.hs
+++ b/plutus-tx/src/PlutusTx/Blueprint/Contract.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE DataKinds          #-}
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GADTs              #-}
 {-# LANGUAGE KindSignatures     #-}
 {-# LANGUAGE OverloadedStrings  #-}
 {-# LANGUAGE RecordWildCards    #-}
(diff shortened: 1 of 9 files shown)


plutus-tx/src/PlutusTx/Blueprint/Contract.hs
```diff
@@ -99,7 +97,14 @@ instance ToJSON (ContractBlueprint referencedTypes) where
...
-        . optionalField "definitions" (ensure (not . Map.null) contractDefinitions)
+        . optionalField "definitions" definitions
    where
     schemaUrl :: String
     schemaUrl = "https://cips.cardano.org/cips/cip57/schemas/plutus-blueprint.json"
+
+    definitions :: Maybe (Map DefinitionId Aeson.Value)
+    definitions = ensure (not . Map.null) (definitionsToMap contractDefinitions toJSON)
+
+-- | 'Blueprint' wrapper to hide the 'referencedTypes' parameter as an existential type.
```
michaelpj, line 107: why does this need to be a wrapper? Why not just change the definition of `ContractBlueprint`?

plutus-tx/src/PlutusTx/Blueprint/Definition.hs
```diff
@@ -36,11 +47,47 @@ import PlutusTx.Blueprint.Schema (Schema (..))
...
 
--- For more context see the note ["Unrolling" types]
+-- | A schema definition of a type @t@ with a list of referenced types @ts@.
+data Definition t ts = MkDefinition DefinitionId (Schema ts)
+  deriving stock (Show)
+
+-- | A registry of schema definitions.
+data Definitions (ts :: [Type]) where
+  NoDefinitions :: Definitions '[]
+  DCons :: SomeDefinition -> Definitions ts -> Definitions (t ': ts)
```
michaelpj, line 57: This seems wrong, it doesn't mention `t`!! You can always quantify over the variables in this constructor also, you don't need `SomeDefinition` AFAICT.

i.e.
```
DCons :: forall ts' . Definition t ts' -> Definitions ts -> Definitions (t ': ts)
```

michaelpj: TBH, I am also confused by the fact that `ts'` would be unused there...

michaelpj (approved):
Seems pretty reasonable!

michaelpj:
I think the note will be out of date once you merge this

Unisay:
> I think the note will be out of date once you merge this

Indeed, after removing an existential wrapper the note got orphaned, so I moved it to where the actual unrolling functionality is.

PR #5841, opened 2024-03-17
Kwxm/costing/read model type from r

kwxm:
This demonstrates how to simplify the cost modelling procedure somewhat by simplifying the interface between Haskell and the R code (it's a step towards PLT-408, amongst other things).  In particular, the "shapes" of the CPU costing functions used to be hard-coded into the Haskell source despite the fact that there's a tag in the JSON file that tells you what the shape is. This now gets the R code to return a similar tag to Haskell so the during cost model generation the Haskell code can look at the tag and parse the R model objects appropriately.  I'm also passing a much simpler structure from R to Haskell.

While I was at it I took the opportunity to improve a couple of other things. One noticeable change is that some functions (`equalsString` for example) are very fast if the argument have different sizes (since in that case they can't possibly be equal) so there's a linear costing function on the diagonal where the sizes of the two arguments are the same, and there's a constant cost off the diagonal.  Previously the off-diagonal constant costs were hard-coded into the Haskell code, but now they're passed in from the R code (and could be _inferred_ there, although I haven't gone that far yet).  They're also stored in `builtinCostModel.json`, which introduces some new cost model parameters that the ledger would have to know about.  I tidied up a couple of other things (and I'd like to do more along the same lines) which has led to the tags in the JSON file changing, and hence also the names of some of the cost model parameters.

### DIFF
--- a/plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
+++ b/plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
@@ -1,10 +1,8 @@
 -- editorconfig-checker-disable-file
-{-# LANGUAGE DataKinds           #-}
-{-# LANGUAGE DeriveGeneric       #-}
-{-# LANGUAGE OverloadedStrings   #-}
-{-# LANGUAGE QuasiQuotes         #-}
-{-# LANGUAGE RecordWildCards     #-}
-{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes       #-}
+{-# LANGUAGE RankNTypes        #-}
+{-# LANGUAGE RecordWildCards   #-}
 
 module CreateBuiltinCostModel where
 
@@ -19,18 +17,13 @@ import PlutusCore.Evaluation.Machine.ExMemoryUsage
 
 import Barbies (bmap, bsequence)
 import Control.Applicative (Const (Const, getConst))
-import Control.Exception (TypeError (TypeError))
-import Control.Monad.Catch (throwM)
-import Data.ByteString.Lazy qualified as BSL (fromStrict)
+-- import Control.Exception (TypeError (TypeError))
+-- import Control.Monad.Catch (throwM)
 import Data.Coerce (coerce)
-import Data.Csv (FromNamedRecord, FromRecord, HasHeader (HasHeader), decode, parseNamedRecord, (.:))
-import Data.Either.Extra (maybeToEither)
 import Data.Functor.Compose (Compose (Compose))
 import Data.SatInt
 import Data.Text (Text)
-import Data.Text.Encoding qualified as T (encodeUtf8)
-import Data.Vector (Vector, find)
-import GHC.Generics (Generic)
+import Text.Printf (printf)
 
 import H.Prelude (MonadR, Region)
 import Language.R (SomeSEXP, defaultConfig, fromSomeSEXP, runRegion, withEmbeddedR)
@@ -138,13 +131,13 @@ costModelsR bmfile rfile = do
   list <- [r|
              source(rfile_hs)
              modelFun(bmfile_hs)
-             |]
+           |]
   let makeCostModelEntry name =
           let n = getConst name
           in Compose $ fmap Const $ [r| list_hs [[n_hs]] |]
   bsequence $ bmap makeCostModelEntry builtinCostModelNames
 
--- ! Creates the cost model from a CSV benchmarking results file and a file
+-- | Creates the cost model from a CSV benchmarking results file and a file
 -- containing R modelling code.
 createBuiltinCostModel :: FilePath -> FilePath -> IO BuiltinCostModel
 createBuiltinCostModel bmfile rfile = do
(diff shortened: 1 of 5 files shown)


plutus-core/cost-model/data/builtinCostModel.json
```diff
@@ -357,16 +357,10 @@
...
-                        "intercept": 453240,
-                        "slope": 220
-                    },
-                    "type": "multiplied_sizes"
-                }
+                "intercept": 453240,
+                "slope": 220
             },
-            "type": "const_above_diagonal"
+            "type": "multiplied_sizes"
```
zliu41, line 363: Does changing the JSON file like this not break everything? I think we need to record both the old shape (for use by PlutusV1 and V2 before PV9) and the new shape?

zliu41 (commented):
> In particular, the "shapes" of the CPU costing functions used to be hard-coded into the Haskell source despite the fact that there's a tag in the JSON file that tells you what the shape is. This now gets the R code to return a similar tag to Haskell so the during cost model generation the Haskell code can look at the tag and parse the R model objects appropriately.

I'm not sure I totally understand. What exactly was hardcoded in Haskell before, that is un-hardcoded in this PR?

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
```diff
@@ -235,241 +228,217 @@ createBuiltinCostModel bmfile rfile = do
...
+-- function types which are currently only used for memory models (which are
+-- constructed directly, not via R), and those won't be handled here.
+
+loadOneVariableCostingFunction :: MonadR m => SomeSEXP (Region m) -> m ModelOneArgument
+loadOneVariableCostingFunction e = do
+  ty <- getType e
+  case ty of
+    "constant_cost" -> ModelOneArgumentConstantCost <$> getConstant e
+    "linear_cost"   -> ModelOneArgumentLinearCost <$> readOneVariableLinearFunction "x_mem" e
+    "linear_in_x"   -> ModelOneArgumentLinearCost <$> readOneVariableLinearFunction "x_mem" e  -- FIXME: duplicate
```
effectfully, line 316: Gonna do that in this PR?

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
```diff
@@ -235,241 +228,217 @@ createBuiltinCostModel bmfile rfile = do
...
+loadTwoVariableCostingFunction :: MonadR m => SomeSEXP (Region m) -> m ModelTwoArguments
+loadTwoVariableCostingFunction e = do
+  ty <- getType e
+  case ty of
+    "constant_cost"        -> ModelTwoArgumentsConstantCost       <$> getConstant e
+    "linear_in_x"          -> ModelTwoArgumentsLinearInX          <$> readOneVariableLinearFunction "x_mem" e
+    "linear_in_y"          -> ModelTwoArgumentsLinearInY          <$> readOneVariableLinearFunction "y_mem" e
+    "linear_in_x_and_y"    -> ModelTwoArgumentsLinearInXAndY      <$> readTwoVariableLinearFunction "x_mem" "y_mem" e
+    "added_sizes"          -> ModelTwoArgumentsAddedSizes         <$> readOneVariableLinearFunction "I(x_mem + y_mem)" e
+    "subtracted_sizes"     -> ModelTwoArgumentsSubtractedSizes    <$> error "subtracted sizes"
```
effectfully, line 328: Why do we need this line given the catch-all clause?

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
```diff
@@ -235,241 +228,217 @@ createBuiltinCostModel bmfile rfile = do
...
+loadThreeVariableCostingFunction :: MonadR m => SomeSEXP (Region m) -> m ModelThreeArguments
+loadThreeVariableCostingFunction e = do
+  ty <- getType e
+  case ty of
+    "constant_cost"               -> ModelThreeArgumentsConstantCost          <$> getConstant e
+    "linear_in_x"                 -> ModelThreeArgumentsLinearInX             <$> readOneVariableLinearFunction "x_mem" e
+    "linear_in_y"                 -> ModelThreeArgumentsLinearInY             <$> readOneVariableLinearFunction "y_mem" e
+    "linear_in_z"                 -> ModelThreeArgumentsLinearInZ             <$> readOneVariableLinearFunction "z_mem" e
+    "quadratic_in_z"              -> ModelThreeArgumentsQuadraticInZ          <$> readOneVariableQuadraticFunction "z_mem" e
+    "literal_in_y_or_linear_in_z" -> ModelThreeArgumentsLiteralInYOrLinearInZ <$> error "literal"
```
effectfully, line 347: And this one.

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
```diff
@@ -235,241 +228,217 @@ createBuiltinCostModel bmfile rfile = do
...
 
 boolMemModel :: ModelTwoArguments
 boolMemModel = ModelTwoArgumentsConstantCost 1
 
 memoryUsageAsCostingInteger :: ExMemoryUsage a => a -> CostingInteger
 memoryUsageAsCostingInteger = coerce . sumCostStream . flattenCostRose . memoryUsage
 
+-- | The types of functions which take an R SEXP and extract a CPU model from
+-- it, then pair it up with a memory costing function.
+type MakeCostingFun1 = forall m . MonadR m => SomeSEXP (Region m) -> m (CostingFun ModelOneArgument)
```
effectfully, line 365: Have a single

```
type MakeCostingFun model = forall m . MonadR m => SomeSEXP (Region m) -> m (CostingFun model)
```

instead and use it as `MakeCostingFun ModelTwoArguments`? Even if only for defining the same type synonyms.

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
```diff
@@ -235,241 +228,217 @@ createBuiltinCostModel bmfile rfile = do
...
+makeTwoVariableCostingFunction :: ModelTwoArguments -> MakeCostingFun2
+makeTwoVariableCostingFunction memModel e =
+  CostingFun <$> (loadTwoVariableCostingFunction e) <*> pure memModel
+
+makeThreeVariableCostingFunction :: ModelThreeArguments -> MakeCostingFun3
+makeThreeVariableCostingFunction memModel e =
+  CostingFun <$> (loadThreeVariableCostingFunction e) <*> pure memModel
+
+makeSixVariableCostingFunction :: ModelSixArguments -> MakeCostingFun6
+makeSixVariableCostingFunction memModel e =
```
effectfully, line 384: I'd make all these `loadOneVariableCostingFunction`, `loadTwoVariableCostingFunction` etc into a single method of a type class taking `ModelOneArgument`, `ModelTwoArguments` etc at the type-level.

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
```diff
@@ -235,241 +228,217 @@ createBuiltinCostModel bmfile rfile = do
...
-addInteger :: MonadR m => SomeSEXP (Region m) -> m (CostingFun ModelTwoArguments)
-addInteger cpuModelR = do
-  cpuModel <- ModelTwoArgumentsMaxSize <$> readModelMaxSize cpuModelR
-  -- The worst case is adding e.g. `maxBound :: Int` + `maxBound :: Int`, which increases the memory usage by one.
-  -- (max x y) + 1
-  let memModel = ModelTwoArgumentsMaxSize $ ModelMaxSize 1 1
-  pure $ CostingFun cpuModel memModel
+addInteger :: MakeCostingFun2
+addInteger =
+  makeTwoVariableCostingFunction $ ModelTwoArgumentsMaxSize $ OneVariableLinearFunction 1 1
```
effectfully, line 391: So we unhardcode parsing, but we still have to keep those models hardcoded here, because that's how we get the data and there's no other way we could infer them, right?

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
```diff
@@ -235,241 +228,217 @@ createBuiltinCostModel bmfile rfile = do
...
-  -- x - y
-  let memModel = ModelTwoArgumentsSubtractedSizes $ ModelSubtractedSizes 0 1 1
-  pure $ CostingFun cpuModel memModel
-
-remainderInteger :: MonadR m => SomeSEXP (Region m) -> m (CostingFun ModelTwoArguments)
+  makeTwoVariableCostingFunction $ ModelTwoArgumentsAddedSizes $ OneVariableLinearFunction 0 1
+
+divideInteger :: MakeCostingFun2
+divideInteger =
+  makeTwoVariableCostingFunction $ ModelTwoArgumentsSubtractedSizes $ ModelSubtractedSizes 0 1 1
```
effectfully, line 407: Wait, but we have 

```haskell
    "subtracted_sizes"     -> ModelTwoArgumentsSubtractedSizes    <$> error "subtracted sizes"
```

?

plutus-core/cost-model/data/builtinCostModel.json
```diff
@@ -789,8 +765,8 @@
     "sliceByteString": {
         "cpu": {
             "arguments": {
-                "intercept": 265318,
-                "slope": 0
+                "intercept": 266994,
+                "slope": 11
```
effectfully, line 769: Wait, it's a constant function, no? Where does the linearity come from? We have

```haskell
sliceByteStringDenotation start n xs = BS.take n (BS.drop start xs)
```

and both `BS.take` and `BS.drop` are O(1) according to their docs. As well as `BS.length` in the `ExMemoryUsage`.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -317,11 +317,11 @@ data ModelTwoArguments =
...
   | ModelTwoArgumentsLinearInXAndY      TwoVariableLinearFunction
-  | ModelTwoArgumentsAddedSizes         ModelAddedSizes
+  | ModelTwoArgumentsAddedSizes         OneVariableLinearFunction
   | ModelTwoArgumentsSubtractedSizes    ModelSubtractedSizes
-  | ModelTwoArgumentsMultipliedSizes    ModelMultipliedSizes
-  | ModelTwoArgumentsMinSize            ModelMinSize
-  | ModelTwoArgumentsMaxSize            ModelMaxSize
+  | ModelTwoArgumentsMultipliedSizes    OneVariableLinearFunction
+  | ModelTwoArgumentsMinSize            OneVariableLinearFunction
+  | ModelTwoArgumentsMaxSize            OneVariableLinearFunction
```
effectfully, line 324: Drop the definitions of `ModelAddedSizes`, `ModelMultipliedSizes`, `ModelMinSize` and `ModelMaxSize` then?

effectfully (commented):
I'll probably need to have another run, here's an initial review.

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
kwxm: That was left over from when I was trying to get it to work.  I've removed it now.  We don't use this function (or the other one later) for CPU usage, so we don't need a case for them.

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
kwxm: I thought there would probably be a more generic way of doing this but I was in a hurry!  I'll think about that again later.

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
kwxm: That's right. If we wanted to be totally general I suppose we could return the model formulas from R (things like [this](https://github.com/IntersectMBO/plutus/blob/a79131eecfec1292b37eb2cce1e6299b46f360cc/plutus-core/cost-model/data/models.R#L424)) and try to parse them and then we could deal with more or less arbitrary functions (and we'd also need some language for describing regions of the input space where different cost models apply).  That'd be the most generic solution to PLT-408, but it might be overkill for what we're doing.

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
kwxm: That's the memory usage: we don't need to parse that.

plutus-core/cost-model/data/builtinCostModel.json
kwxm: See [this comment](https://github.com/IntersectMBO/plutus/blob/a79131eecfec1292b37eb2cce1e6299b46f360cc/plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs#L475) in the old code and [this one](https://github.com/IntersectMBO/plutus/blob/fe35887112a04ca44f4318b1e526766c2717dad8/plutus-core/cost-model/data/models.R#L497) in the new code.  The slope here is negligible, so the function's essentially constant.  I did wonder whether just to make it actually constant, but that would make the costing function have a less general type than in previous versions.  Maybe that's OK.

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
effectfully: I'm not sure I understand, but I trust you!

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
effectfully: I think doing it generically isn't gonna be any better, my version of `MakeCostingFun` merely eliminates some of the boilerplate.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
kwxm: Yes, I want to rework some of this later.  It's a bit confusing because it'll change the names of some of the cost model parameters and it's kind of orthogonal to what I was trying to (although I did get carried away and make some changes along those lines that probably shouldn't be in this PR).

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
effectfully: Yeah, that does sound like an overkill. Thanks!

It would be really great if we could somehow have these models hardcoded here and reused within the definitions of builtins. No JSON for shapes, just taking the shapes directly from this place and using them where we need them. That would be the most reliable and ergonomic solution I think, but for that we also need to untangle shapes from constants (currently we put constants right into the shapes basically).

I feel like my approach would be better long-term. Short-term, not so sure, perhaps yours is better. Both are complicated though: with yours we need to have multiple JSON files or to add support for some form of versioning into the existing JSON file (right?) and with mine we need to rewrite half the costing code. Yours sounds annoying, mine sounds insufferable.

plutus-core/cost-model/data/builtinCostModel.json
kwxm: Oh, there's something wrong there. I'll check that later.

plutus-core/cost-model/data/builtinCostModel.json
effectfully: I see, thanks.

plutus-core/cost-model/data/builtinCostModel.json
effectfully: > `builtinCostModel.json`, one for each `PlutusV<n>`

Not just `PlutusV<n>`, though? We also need to worry about protocol versions. And I don't think we need to worry about all of `PlutusV<n>`. I think the end result is that for Conway we'll need 3 files (it's a coincidence that we have exactly that many Plutus versions). I'll write about that later.

plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs
kwxm: Yes. I forgot about that.

kwxm:
> > In particular, the "shapes" of the CPU costing functions used to be hard-coded into the Haskell source despite the fact that there's a tag in the JSON file that tells you what the shape is. This now gets the R code to return a similar tag to Haskell so the during cost model generation the Haskell code can look at the tag and parse the R model objects appropriately.
> 
> I'm not sure I totally understand. What exactly was hardcoded in Haskell before, that is un-hardcoded in this PR?

This is mostly just about improving the R/Haskell interface.  The Haskell code that generates the cost model from the R output currently has the "shapes" of the CPU costing functions hardcoded, like [this](https://github.com/IntersectMBO/plutus/blob/a79131eecfec1292b37eb2cce1e6299b46f360cc/plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs#L598):
```
 cpuModel <- ModelTwoArgumentsAddedSizes <$> readModelAddedSizes cpuModelR
```
It reads the numbers from R and then fits them into a `ModelTwoArgumentsAddedSizes` object.  However, the R code is actually what decides that the model's supposed to be added sizes, like [this ](https://github.com/IntersectMBO/plutus/blob/a79131eecfec1292b37eb2cce1e6299b46f360cc/plutus-core/cost-model/create-cost-model/CreateBuiltinCostModel.hs#L598)
```
lm(t ~ I(x_mem + y_mem)
```
(R has a rather [esoteric syntax](https://www.econometrics.blog/post/the-r-formula-cheatsheet/) for specifying regression formulae).  

This PR is mostly about completely removing the hardcoded model shapes from the Haskell code and getting R to tell it the shape instead (and it also moves some hardcoded numerical constants into the R code too).  Now the Haskell code knows nothing at all about the shapes of CPU costing functions for specific builtins, which should make it a bit more flexible; it still has all of the memory costing functions hardcoded, but we could put them in a JSON file too.

I don't know if we actually want to do this, but I wanted to check that it was possible.  It helped me to spot a few other things that could do with being tiidied up too.

plutus-core/cost-model/data/builtinCostModel.json
zliu41: We need one JSON per semantic variant, so 3 sounds right. Of course, we can merge them into a single JSON file, but all the information needs to be there.

kwxm:
Closed in  favour of #5857

PR #5845, opened 2024-03-19
Make Unroll an open type family

Unisay:
This PR contains an alternative approach to un-rolling `AsData`-types: it uses an open type family for it.

This has its own pro:
* Users are free to specify how to unroll any type (including custom `AsData` types) 
by provinding a TF instance, for example:  
`type instance Unroll AsDataFoo = '[AsDataFoo, Bool, Integer]` 

and con:

* There is no default un-rolling of generic types, instead users need to instantiate the `Unroll` TF for every custom data type:
`type instance Unroll MyGenericFoo = GenericUnroll Foo`


Unisay:
Decided not to implement it.

PR #5807, opened 2024-02-27
[Builtins] Add another variant

effectfully:
Don't look here yet.

PR #5846, opened 2024-03-19
[Builtins] Push 'MajorProtocolVersion' into 'MachineParameters'

effectfully:
This one competes with #5807.

Don't look here yet.


plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
```diff
@@ -399,22 +398,5 @@ toBuiltinsRuntime
...
-        -- lambda binding the @cost@ variable, which renders all the optimizations useless. By
-        -- using a @NOINLINE@ pragma we tell GHC to create a separate thunk, which it can properly
-        -- optimize, because the other bazillion things don't get in the way.
-        {-# NOINLINE runtime #-}
-    in
-        -- Force each 'BuiltinRuntime' to WHNF, so that the thunk is allocated and forced at
-        -- initialization time rather than at runtime. Not that we'd lose much by not forcing all
-        -- 'BuiltinRuntime's here, but why pay even very little if there's an easy way not to pay.
-        force runtime
+    BuiltinsRuntime $ toBuiltinRuntime cost . inline toBuiltinMeaning semvar
```
effectfully, line 401: `force` no longer makes sense with the addition of `MajorProtocolVersion` (as we can't reduce to WHNF all `BuiltinRuntime` thunks when constructing an `EvaluationContext` due to the new lambda getting in the way) and removing it causes the code to be compiled slightly differently making the `NOINLINE` pragma pretty much a no-op (it only adds a pointless join point that doesn't affect evaluation in any way). Hence we can simply remove all of that logic. 

I don't know if things work the same way for other GHC versions though.

effectfully:
This doesn't make any sense, because `machineCosts` depend on the protocol version too, not just `builtinsRuntime`.

PR #5847, opened 2024-03-20
Chore: formatting, explicit imports.

Unisay:
Housekeeping chore: format a few modules (`Purity`) before changing them.

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Purity.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Purity.hs
@@ -9,28 +9,30 @@
 {-# LANGUAGE ViewPatterns          #-}
 
 module PlutusIR.Purity
-    ( isPure
-    , isSaturated
-    , isWorkFree
-    , EvalOrder
-    , unEvalOrder
-    , EvalTerm (..)
-    , Purity (..)
-    , termEvaluationOrder
-    ) where
+  ( isPure
+  , isSaturated
+  , isWorkFree
+  , EvalOrder
+  , unEvalOrder
+  , EvalTerm (..)
+  , Purity (..)
+  , termEvaluationOrder
+  ) where
 
-import PlutusCore.Builtin
-import PlutusCore.Pretty
-import PlutusIR
-import PlutusIR.Contexts
-
-import Control.Lens hiding (Strict)
+import Control.Lens ((^.))
 import Data.DList qualified as DList
 import Data.List.NonEmpty qualified as NE
+import PlutusCore.Builtin (BuiltinMeaning (..), ToBuiltinMeaning (..), TypeScheme (..))
 import PlutusCore.Name.Unique qualified as PLC
-import PlutusIR.Analysis.Builtins
-import PlutusIR.Analysis.VarInfo
-import Prettyprinter
+import PlutusCore.Pretty (Pretty (pretty), PrettyBy (prettyBy))
+import PlutusIR (Binding (TermBind), Name, Recursivity (NonRec, Rec),
+                 Strictness (NonStrict, Strict), Term (..), TyName)
+import PlutusIR.Analysis.Builtins (BuiltinsInfo, biSemanticsVariant, builtinArityInfo)
+import PlutusIR.Analysis.VarInfo (VarInfo (DatatypeConstructor), VarsInfo, lookupVarInfo,
+                                  varInfoStrictness)
+import PlutusIR.Contexts (AppContext (..), Saturation (Oversaturated, Saturated, Undersaturated),
+                          fillAppContext, saturates, splitApplication)
+import Prettyprinter (vsep, (<+>))
 
 saturatesScheme :: AppContext tyname name uni fun a -> TypeScheme val args res -> Maybe Bool
 -- We've passed enough arguments that the builtin will reduce.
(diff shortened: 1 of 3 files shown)


plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -9,28 +9,30 @@
...
-import PlutusCore.Builtin
-import PlutusCore.Pretty
-import PlutusIR
-import PlutusIR.Contexts
-
-import Control.Lens hiding (Strict)
+import Control.Lens ((^.))
 import Data.DList qualified as DList
 import Data.List.NonEmpty qualified as NE
+import PlutusCore.Builtin (BuiltinMeaning (..), ToBuiltinMeaning (..), TypeScheme (..))
```
michaelpj, line 25: I don't think we reached any agreement on whether this is desirable or not. For me, this makes things worse, since these are project modules that I would expect people to be familiar with.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -75,41 +76,47 @@ instance Pretty WorkFreedom where
   pretty MaybeWork = "maybe work?"
   pretty WorkFree  = "work-free"
 
--- | Either the "next" term to be evaluated, along with its 'Purity' and 'WorkFreedom',
--- or we don't know what comes next.
-data EvalTerm tyname name uni fun a =
-  Unknown
+{- | Either the "next" term to be evaluated, along with its 'Purity' and 'WorkFreedom',
```
michaelpj, line 79: The more this happens the more I dislike our current approach of "people can just reformat things to their personal preference". It's terrible. _Please_ can we decide to either use `fourmolu` consistently or just not reformat things.

michaelpj (commented):
I am unhappy with ad-hoc decisions here.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
michaelpj: At the moment nothing stops someone just reformatting this back a different way with "Chore: reformatting" :(

Unisay:
> I am unhappy with ad-hoc decisions here.

Let me undo those decisions that make you unhappy. What is it in particular?

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
Unisay: > these are project modules that I would expect people to be familiar with.

Actually, in my case it is other way around: I might be acquainted with the `Prettyprinter` symbols from using the library in other projects, but I don't yet know all the symbols exported by our `PlutusTx` modules. 🤷🏼‍♂️ 

Anyway, I am not particularly hang up on explicit imports, I can undo those ones you prefer to leave wildcard.

michaelpj:
"unhappy" was meant to attach to "ad-hoc" rather than the specific decisions :) I just think it's past time we agreed on a policy.

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
ana-pantilie: I agree that we should have a policy. I personally prefer explicit export lists, and I _think_ this can be enforced somewhere either in the formatter (once we decide what we want) or in the commit pipeline. Maybe we should talk about this during the technical discussion?

plutus-core/plutus-ir/src/PlutusIR/Purity.hs
Unisay: > stylish-haskell and it seems to format new code in the same way the existing code in the module is formatted

Two reasons:
* we already enforce stylish-haskell (git hooks) .
* stylish-haskell is very liberal: it does change the imports and aligns vertical arrows, equals, etc. It won't change indentation.

By the way, I have configured local formatting such that  `fourmolu` -produced output is piped to the `stylish-haskell`: their combination leads to a least change.

michaelpj: > A thing that is decided by a fourmolu

Nothing in our style guide or contributing guide even says that we use fourmolu at all.

> By the way, I have configured local formatting such that fourmolu -produced output is piped to the stylish-haskell: their combination leads to a least change.

:scream: 

> did the team try to use fourmolu at some point? What didn't they like about it?

@effectfully is the main opponent, you can search in slack for some discussion.

effectfully: @ana-pantilie see [this](https://github.com/IntersectMBO/plutus/pull/5761#pullrequestreview-1863517330) for example.

ana-pantilie: > see https://github.com/IntersectMBO/plutus/pull/5761#pullrequestreview-1863517330 for example.

I initially read the first diff as fourmolu removing the `INLINEABLE` pragmas which got me very scared, but I see it's only adding a newline between the pragma and the type declaration, right? I looked up the docs and `INLINEABLE` or `INLINE` can be added anywhere in the module I believe, https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/pragmas.html#inlinable-pragma.
The other changes you pointed out aren't really dangerous, are they? If they're not, although I agree that they're ugly, I think they're not _that bad_ once you consider the bigger picture. Don't hate me, but I think that as long as these "formatting bugs" are cosmetic and relatively infrequent it's a small price to pay for having consistent automated formatting.

PR #5848, opened 2024-03-21
Introduce Data-backed basic types

ana-pantilie:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-tx/src/PlutusTx/DataList.hs
```diff
@@ -0,0 +1,71 @@
...
+) where
+
+import PlutusTx.Builtins qualified as P
+import PlutusTx.Builtins.Internal qualified as BI
+import PlutusTx.IsData qualified as P
+import PlutusTx.Prelude hiding (map)
+
+import GHC.Exts qualified as H
+
+newtype List a = List P.BuiltinData
```
zliu41, line 17: You can define a regular list, wrapped in `asData`:

```haskell
AsData.asData
  [d| data List a = Nil | Cons a (List a) |]
```

This should make defining the list operations much easier since you can use `Nil` and `Cons`.

plutus-tx/src/PlutusTx/DataList.hs
```diff
@@ -0,0 +1,59 @@
...
+  [d|
+    data List a = Nil | Cons a (List a)
+      deriving newtype (P.ToData, P.FromData, P.UnsafeFromData)
+  |]
+
+type DataElem a = (P.UnsafeFromData a, P.ToData a)
+
+instance (DataElem a) => H.IsList (List a) where
+    type Item (List a) = a
+    fromList [] = Nil
```
zliu41, line 29: If using the list syntax means calling `fromList` and `toList` all the time, that would be very expensive, since each call of `fromList` and `toList` goes through the whole list. You can verify that by writing a budget test. If that's the case we should just give up on the list syntax.

michaelpj: Yes, unfortunately the `fromList` calls are inserted into the program and are present at runtime. And they're likely to be recursive functions, I highly doubt our optimizer is going to be good enough to optimize them away.

ana-pantilie: Hmm, yeah, I guess what we actually want is to define `asData` on `[a]`. How about `asData [d| data List a = List [a] |]`? Then the `toList/fromList` transformations are whatever is implemented in the `pattern List` that the TH produces, right?

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,263 @@
...
+    MCons a (Map_ as) =
+      let aPair = Pair.toBuiltinPair a
+          builtinList = BI.unsafeDataAsMap as
+       in Map_ $ BI.mkMap $ BI.mkCons aPair builtinList
+
+{-# COMPLETE MNil, MCons #-}
+
+-- I think ^ doesn't work because the Data encoding doesn't know that
+-- BuiltinList (BuiltinPair BuiltinData BuiltinData) is an instance of BuiltinList BuiltinData
+-- also Map is a completely different constructor from List ( this is the actual reason)
```
ana-pantilie, line 84: Obsolete, TODO remove.

zliu41:
What I had in mind is to define DataList using `asData`:

```haskell
AsData.asData
  [d| data DataList a = DataNil | DataCons a (DataList a) |]
```

and DataMap as a newtype wrapper:

```haskell
newtype DataMap k v = DataMap (DataList (k, v))
```

Does that not work?

zliu41:
The Map operations look less efficient than the one in https://github.com/IntersectMBO/plutus/pull/5701. For example, `lookup`: every time `MCons` is used as a pattern or as a constructor involves calling `BI.unsafeDataAsMap` and `BI.mkMap`, so these end up being called repeatedly. But I'm not sure whether this is the main reason for the worse numbers.

In https://github.com/IntersectMBO/plutus/pull/5701, the `lookup` operation first converts the Map into a BuiltinList, and as a result, `BI.unsafeDataAsMap` is called only once, and `BI.mkMap` is not needed at all.

What you can try:
- Change the Marlowe validator incrementally. First you can make the changes similar to what I did in #5701 (i.e., leave `Value` alone, among other things), and see if you can get similar numbers
- Change the costs of `unsafeDataAsMap` and `mkMap` to 0, and see how it affects the numbers

plutus-benchmark/marlowe/src/PlutusBenchmark/Marlowe/Scripts/Semantics.hs
```diff
@@ -111,6 +122,15 @@ closeInterval (Interval (LowerBound (Finite (POSIXTime l)) lc) (UpperBound (Fini
     )
 closeInterval _ = Nothing
 
+-- TODO: abstract away
+toDataValue :: SOP.Value -> Data.Value
```
zliu41, line 126: This could be another reason for the much worse numbers, so again I'd suggest continuing to use the SOP value for now and change it later.

plutus-tx/src/PlutusTx/DataMap.hs
```diff
@@ -0,0 +1,254 @@
...
+instance (DataElem k, DataElem v, Eq k, Semigroup v) =>  Semigroup (Map k v) where
+  (<>) = unionWith (<>)
+
+instance (DataElem k, DataElem v, Eq k, Semigroup v) => Monoid (Map k v) where
+  mempty = empty
+
+empty :: Map k v
+empty = MNil
+
+lookup :: (DataElem k, DataElem v, Eq k) => k -> Map k v -> Maybe v
```
zliu41, line 97: All these `DataElem` constraints could also lead to higher costs, since sometimes only one of `ToData` and `UnsafeFromData` is required.

zliu41:
Another thing you can try: replace the functions in your `DataMap.hs` - insert, delete, lookup etc., one by one, with the implementation in #5701, and observe how it affects the Marlowe budget numbers.

zliu41:
As a starting point I added budget tests comparing `DataMap.all` against the implementation in #5701 (`all2`) in [8af9a78](https://github.com/IntersectMBO/plutus/pull/5848/commits/8af9a789c4d5adb770fb5ccd4bbbd3d580bb8ecc). As you can see the difference is quite big.

As far as I can tell there are a few kinds of overhead in `DataMap.all` compared to `all2`:
- Overhead from `DataElem` (all the `(%, %)` things in PIR). `all2` only requires `P.UnsafeFromData a`.
- Overhead from `Maybe`.
- This should be an easy win: `all` should be implemented with an inner function `go`, like `all2` does, which avoids threading `f` around.

Try and see if you can improve it to get similar performance as `all2` - if not you may have to switch to using `all2`.

PR #5849, opened 2024-03-22
Add simple force-delay tests + clean-up

ana-pantilie:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
```diff
@@ -227,6 +227,18 @@ multiApp = runQuote $ do
       app = mkIterAppNoAnn lam [mkConstant @Integer () 1, mkConstant @Integer () 2, mkConstant @Integer () 3]
   pure app
 
+forceDelayNoApps :: Term Name PLC.DefaultUni PLC.DefaultFun ()
```
michaelpj, line 230: btw you can write these with input files that have program syntax in them. That can be easier for simple programs like this! I'm not sure why this file has so many test cases defined by explicitly constructing ASTs...

ana-pantilie: Oh thanks for the tip! I can't seem to find any example, though, could you share one?

michaelpj: Hmm, weird, there just aren't any for UPLC? Not sure why. But look at e.g. `PlutusIR.Transform.Beta.Test`

zliu41: Yeah UPLC tests have always been done this way. The lack of types makes it easy to construct terms in Haskell. I find it at least as easy as writing UPLC by hand.

ana-pantilie:
Forgot to mark this as ready for review.

I tested the simple `force-delay` cancellation algorithm we had before my previous PR on the tests I've added, and it works. If someone remembers some test for which it failed when we took a look at this, please let me know. My suspicion is that we just misread the complicated nesting of some force-application-lambda-delay when we were looking at those examples related to my previous PR.

I also addressed some of the comments I got on my PRs after they were merged.

zliu41 (approved):

PR #5851, opened 2024-03-22
[Costing] Provide support for multiple 'CostModel's

effectfully:
This makes cost models configurable, so that the cost model can depend on the protocol version. All commentary is in the diff.


effectfully:
Uh-oh, up to 50% slowdown on `nofib` benchmarks.

effectfully:
Wwwhat? It runs locally!

effectfully:
^ shaky zero.

effectfully:
What in the name of fuck

plutus-benchmark/lists/bench/Bench.hs
```diff
@@ -55,5 +54,5 @@ main :: IO ()
 main = do
   -- Run each benchmark for at least 15 seconds.  Change this with -L or --timeout.
   config <- getConfig 15.0
-  evalCtx <- evaluate $ force mkEvalCtx
```
effectfully, line 58: I moved that `force` to the creation site, so that we can remove it from lots of call sites (which was risky to begin with, because we'd occasionally forget to add a `force`, thus distorting the benchmarking results).

plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
```diff
@@ -399,22 +398,15 @@ toBuiltinsRuntime
...
+    -- 'let'-binding. This makes it easier for GHC to optimize the internals of builtins, because
+    -- without a 'let'-binding GHC would sometimes refuse to cooperate and push 'toBuiltinRuntime'
+    -- to the inside of the inlined 'toBuiltinMeaning' call, creating lots of 'BuiltinMeaning's
+    -- instead of 'BuiltinRuntime's with the former hiding the costing optimizations behind a lambda
+    -- binding the @cost@ variable, which makes the optimizations useless.
+    -- By using 'lazy' we tell GHC to create a separate thunk, which it can properly optimize,
+    -- because the other bazillion things don't get in the way. We used to use an explicit
+    -- 'let'-binding marked with @NOINLINE@, but that turned out to be unreliable, because GHC
+    -- feels free to turn it into a join point instead of a proper thunk.
+    lazy . BuiltinsRuntime $ toBuiltinRuntime cost . inline toBuiltinMeaning semvar
```
effectfully, line 411: As I said during the GHC Core presentation, `lazy` is way more reliable here than `let` + `NOINLINE`.

plutus-ledger-api/plutus-ledger-api.cabal
```diff
@@ -151,7 +151,6 @@ test-suite plutus-ledger-api-test
     Spec.CostModelParams
     Spec.Eval
     Spec.Interval
-    Spec.NoThunks
```
effectfully, line 154: Incorporated it into `Spec.Eval` where it shares code with a test I added in this PR.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1287,7 +1291,8 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
                 :: BS.ByteString -> BS.ByteString -> BS.ByteString -> BuiltinResult Bool
             verifyEd25519SignatureDenotation =
                 case semvar of
-                  DefaultFunSemanticsVariant1 -> verifyEd25519Signature_V1
+                  DefaultFunSemanticsVariant0 -> verifyEd25519Signature_V1
+                  DefaultFunSemanticsVariant1 -> verifyEd25519Signature_V2
```
zliu41, line 1295: This mapping is potentially easy to get wrong. Is there any test for this?

zliu41 (approved):
Looks totally legit

plutus-ledger-api/test/Spec/Eval.hs
```diff
@@ -73,8 +84,48 @@ testUnlifting = testCase "check unlifting behaviour changes in Vasil" $ do
...
+        take (lengthParamNamesV lv) costParams & case lv of
+            PlutusV1 -> V1.mkEvaluationContext
+            PlutusV2 -> V2.mkEvaluationContext
+            PlutusV3 -> V3.mkEvaluationContext
+
+-- | Ensure that 'toMachineParameters' never throws for all language and protocol versions.
+evaluationContextCacheIsComplete :: TestTree
+evaluationContextCacheIsComplete =
+    testGroup "EvaluationContext has machine parameters for all protocol versions" $
+        enumerate <&> \lv -> testCase (show lv) $ do
```
michaelpj, line 107: could make it a property over language versions and protocol versions? probably not worth it

michaelpj (approved):
Looks good!

plutus-ledger-api/test/Spec/Eval.hs
effectfully: Why? It's computationally trivial to enumerate them all. Once we have 10k lvs * pvs, then maybe we should think about optimizing it, but right now I'm definitely not gonna introduce any unnecessary risk by reducing test coverage.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
effectfully: We're just replacing one implementation with another for maintenance reasons, so doing this matching here is only for the purpose of being paranoid about not breaking anything in the past in case of some unknown unknown. There's nothing additional to test therefore beyond what we already test, which is how we use `ed25519_Variant0Prop` to test all variants of this builtin.

Note that `consByteString` is untestable in the same way, but for a different reason: `Variant0` and `Variant1` versions of that builtin are completely identical (unlike with `verifyEd25519Signature`, whose variants behave identically but come from different libraries, i.e. are nominally different), so again there's nothing to test.

So for both the builtins we only make sure that whatever their variant is, it behaves as expected, we cannot test that `Variant0` behaves differently from `Variant1`, because it doesn't.

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
```diff
@@ -106,33 +106,71 @@ mkTermToEvaluate ll pv script args = do
...
+        Nothing -> error $ Prelude.concat
+            ["Internal error: ", show lv, " does not support protocol version ", show pv]
+        Just machPars -> machPars
 
 {-| An opaque type that contains all the static parameters that the evaluator needs to evaluate a
-script.  This is so that they can be computed once and cached, rather than being recomputed on every
+script. This is so that they can be computed once and cached, rather than being recomputed on every
 evaluation.
+
+Different protocol versions may require different bundles of machine parameters, which allows us for
```
kwxm, line 119: Maybe make this into a Note?  It seems  quite important.  It would seem to have some bearing on [this](https://github.com/IntersectMBO/plutus/pull/5851/files#r1565862928), for example.

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
```diff
@@ -106,33 +106,71 @@ mkTermToEvaluate ll pv script args = do
...
+example to tweak the shape of the costing function of a builtin, so that the builtin costs less.
+Currently this means that we have to create multiple 'DefaultMachineParameters' per language
+version, which we put into a cache (represented by an association list) in order to avoid costly
+recomputation of machine parameters.
+
+In order to get the appropriate 'DefaultMachineParameters' at validation time we look it up in the
+cache using a semantics variant as a key. We compute the semantics variant from the protocol
+version using the stored function. Note that the semantics variant depends on the language version
+too, but the latter is known statically (because each language version has its own evaluation
+context), hence there's no reason to require it to be provided at runtime.
```
kwxm, line 129: So in principle we have a matrix of semantics variants index by (LL,PV) pairs, but at this point we know the LL, so `machParsList` is essentially the row of that matrix corresponding to the given LL.  Is that a reasonable way to think of this?

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
```diff
@@ -106,33 +106,71 @@ mkTermToEvaluate ll pv script args = do
...
+   more semantically precise to associate bundles of machine parameters with semantics variants than
+   with protocol versions
 -}
-newtype EvaluationContext = EvaluationContext
-    { machineParameters :: DefaultMachineParameters
+data EvaluationContext = EvaluationContext
+    { _evalCtxLedgerLang    :: PlutusLedgerLanguage
+      -- ^ Specifies what language versions the 'EvaluationContext' is for.
+    , _evalCtxToSemVar      :: MajorProtocolVersion -> BuiltinSemanticsVariant DefaultFun
+      -- ^ Specifies how to get a semantics variant given a 'MajorProtocolVersion'.
```
kwxm, line 144: Should that really be "Specifies how to get a semantics variant _for this ledger language_ given a 'MajorProtocolVersion"?

Once again I'm becoming confused by all of the versions.  The PLC specification says (p 26)

>  To handle [changes in builtin semantics]  we define a series of built-in semantics variants, which indicate which
behaviour should be used. A fix will typically be deployed by defining a new semantics variant, and then
using that variant for future ledger languages (but not existing ones, since this is usually a backwards-
incompatible change).

That suggests that the semantics variant depends only on the ledger language, but I think we're now assuming that it depends on both the LL and the PV.  We should be careful to specify exactly what's going on.

kwxm (approved):
This looks good.  Inevitably once I started thinking about this stuff again I became confused by the versions and how they depend on each other.  I'm pretty convinced that this does what we want though.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1075,10 +1076,12 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
        possibly different semantics.  Note that DefaultFunSemanticsVariant1,
        DefaultFunSemanticsVariant1 etc. do not correspond directly to PlutusV1,
        PlutusV2 etc. in plutus-ledger-api: see Note [Builtin semantics variants]. -}
-    data BuiltinSemanticsVariant DefaultFun =
-              DefaultFunSemanticsVariant1
-            | DefaultFunSemanticsVariant2
-        deriving stock (Enum, Bounded, Show)
+    data BuiltinSemanticsVariant DefaultFun
+        = DefaultFunSemanticsVariant0
```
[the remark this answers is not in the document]
effectfully: I'm fine either way. I'm hopeful we'll be able to get rid of the semantics variants entirely and simply rely on the language and protocol versions directly (maybe in the "condensed" form).

plutus-ledger-api/src/PlutusLedgerApi/V3/EvaluationContext.hs
```diff
@@ -32,6 +32,10 @@ a protocol update with the updated cost model parameters.
...
-mkEvaluationContext = tagWithParamNames @V3.ParamName
-                    >=> pure . toCostModelParams
-                    >=> mkDynEvaluationContext Plutus.DefaultFunSemanticsVariant2
+mkEvaluationContext =
+    tagWithParamNames @V3.ParamName
+    >=> pure . toCostModelParams
+    >=> mkDynEvaluationContext
+        PlutusV3
+        [DefaultFunSemanticsVariant2]
+        (const Plutus.DefaultFunSemanticsVariant2)
```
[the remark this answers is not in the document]
effectfully: > Let me check: V1 and V2 have `if pv < conwayPV then ... else ...` here,  but it's safe to have `const` for V3 because we know that `PlutusV3` is going to be introduced at the same time as `conwayPV`, so a PlutusV3 script can never be executed with a lower PV, so there's no possiblity that `pv < conwayPV` here (but conceivably we might need `if pv < einsteinPV` or something in the future).  Is that right? 

I believe so, yes.

> Maybe we should try to explain that somewhere, since there's a LL/PV dependency that I don't think is captured in the code anywhere. Maybe we should have note containing a table like we do in the PLC specification (or a big table that contains all of the relevant information, since there are multiple tables in the spec).

Good point, I'll add a small table somewhere, but I'm not totally sure where given that the logic is spread across multiple files. I guess `PlutusLedgerApi.Common.Versions` indeed.

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
effectfully: > So in principle we have a matrix of semantics variants index by (LL,PV) pairs, but at this point we know the LL, so `machParsList` is essentially the row of that matrix corresponding to the given LL.  Is that a reasonable way to think about this?

Yes, exactly. I'll steal your wording for the comment I think. Thanks!

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
effectfully: > That suggests that the semantics variant depends only on the ledger language, but I think we're now assuming that it depends on both the LL and the PV. 

Yes, exactly.

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
effectfully: > Maybe make this into a Note?

I kinda want this to be the Haddock for extra transparency. I do reference the Haddock of a function instead of a Note sometimes.

plutus-ledger-api/src/PlutusLedgerApi/V2/EvaluationContext.hs
```diff
@@ -32,6 +34,12 @@ a protocol update with the updated cost model parameters.
...
-                    >=> pure . toCostModelParams
-                    >=> mkDynEvaluationContext Plutus.DefaultFunSemanticsVariant1
+mkEvaluationContext =
+    tagWithParamNames @V2.ParamName
+    >=> pure . toCostModelParams
+    >=> mkDynEvaluationContext
+        PlutusV2
+        [DefaultFunSemanticsVariant0, DefaultFunSemanticsVariant1]
+        (\pv -> if pv < conwayPV
+            then Plutus.DefaultFunSemanticsVariant0
```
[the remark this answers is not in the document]
effectfully: Yeah, just some leftovers, fixed, thank you.

PR #5854, opened 2024-03-25
Agda html

ramsay-t:
Added some text to the Agda files for the basic Type and Renaming/Substitution sections.  [PLT-1583](https://input-output.atlassian.net/browse/PLT-1583)

Most of this text is actually pulled from the "Fun and Profit" paper, but I have spent most of my time cross referencing between the two things, so it makes sense to have the fluent explanations on hand.

Some of the other text additions are just titles and  brief names to make the HTML more readable and easily navigated.


plutus-metatheory/src/Type/RenamingSubstitution.lagda.md
```diff
@@ -25,20 +25,35 @@ open _⊢⋆_
...
 ```
 variable
   ρ ρ' : Ren Φ Ψ
 ```
 
-Extending a type renaming — used when going under a binder.
+As we are going to push renamings through types we need to be able to push them
+under a binder. To do this safely the newly bound variable should remain untouched and
+other renamings should be shifted by one to accommodate this. (Note: this is
+called `lift⋆` in the paper).
```
ana-pantilie, line 56: In which paper?

ramsay-t: The answer is: the ["Fun and Profit" paper](https://ci.iog.io/build/1230848/download/1/paper.pdf) but I wasn't sure whether to reference it explicitly in the Agda, since I'd rather not have a big (and identical) bibliography in every file. I suppose it is markdown, I could do what I've just done here!

plutus-metatheory/src/Type.lagda.md
```diff
@@ -81,7 +81,10 @@ scope.
 A type is either a type variable, a pi type, a function type, a
 lambda, an application, an iso-recursive type `μ`, a size, or a type
 constant (base type). Note that recursive types range over an
-arbitrary kind `k` which goes beyond standard iso-recursive types.
+arbitrary kind `k` which goes beyond standard iso-recursive types. Also note that `Π`,
```
effectfully, line 84: I always thought `Π` was for dependent types and what we have in Plutus is `∀`.

plutus-metatheory/src/Type/Equality.lagda.md
```diff
@@ -33,6 +33,13 @@ open import Type.RenamingSubstitution
 
 This serves as a declaritive specification of the semantics of types.
 
+We define type equality as an intrinsically scoped and kinded relation. In particular,
+this means it is impossible to state an equation between types in different contexts, or of
+different kinds. The only interesting rule is the β-rule from the lambda calculus. We omit
+the η-rule as Plutus Core does not have it. The additional
+types (`⇒`, `∀`, and `µ`) do not have any computational behaviour, and are essentially inert.
```
effectfully, line 40: Wait, I'm confused, is it `Π` or `∀`?

plutus-metatheory/src/Type/Equality.lagda.md
```diff
@@ -33,6 +33,13 @@ open import Type.RenamingSubstitution
 
 This serves as a declaritive specification of the semantics of types.
 
+We define type equality as an intrinsically scoped and kinded relation. In particular,
+this means it is impossible to state an equation between types in different contexts, or of
+different kinds. The only interesting rule is the β-rule from the lambda calculus. We omit
```
effectfully, line 38: Is this just for convenience? Surely types can be equal in different contexts (but not of different kinds)?

plutus-metatheory/src/Type/RenamingSubstitution.lagda.md
```diff
@@ -25,20 +25,35 @@ open _⊢⋆_
 
 ## Type renaming
 
-A type renaming is a mapping of type variables to type variables.
+Types can contain functions and as such are subject to a nontrivial equality relation. To
+explain the computation equation (the β-rule) we need to define substitution for a single
+type variable in a type. A type renaming is a function from type variables in one context to
+type variables in another. We only need the ability to
```
effectfully, line 31: We don't use Conor's [Type-Preserving Renaming and Substitution](http://strictlypositive.org/ren-sub.pdf) trick to express both renaming and substitution using a single general thing, right? Would that trick complicate things for us or do we not use it for a different reason?

plutus-metatheory/src/Type/RenamingSubstitution.lagda.md
```diff
@@ -25,20 +25,35 @@ open _⊢⋆_
...
+function, and we inherit the properties of functions provided by η-equality, such as
+associativity of composition, for free. Note that even though
+renamings are functions we do not require our metatheory (Agda’s type system) to support
+functional extensionality. We only ever need to make use of an equation
+between renamings on a point (a variable) and therefore need only a pointwise version
+of equality on functions to work with equality of renamings and substitutions.
 
 ```
 Ren : Ctx⋆ → Ctx⋆ → Set
 Ren Φ Ψ = ∀ {J} → Φ ∋⋆ J → Ψ ∋⋆ J
```
effectfully, line 44: I learned the beauty of [order-preserving embeddings](https://fplab.bitbucket.io/posts/2008-03-07-order-preserving-embeddin.html) from James, I feel betrayed.

effectfully (approved):
LGTM

plutus-metatheory/src/Type.lagda.md
jmchapman: You're right. But, \forall is reserved so I think Phil chose Pi for this.

plutus-metatheory/src/Type/Equality.lagda.md
jmchapman: We're using the symbol Pi, so this should say Pi too

plutus-metatheory/src/Type/Equality.lagda.md
jmchapman: You could define it more heterogeneously and allow more degrees of freedom such as different contexts and kinds but it's not necessary.

plutus-metatheory/src/Type/RenamingSubstitution.lagda.md
jmchapman: Ha! We're doing stuff in a higher order way here by using NBE. Renaming and substitution as functions are quite nice to work with unless you run into issues with extensionality which we don't here. However you're right to notice that we don't need to full power of renamings but that doesn't bite us either. There is also a nice correspondence with relative monads which gives us a clear framework to work out what properties we need.

plutus-metatheory/src/Type/RenamingSubstitution.lagda.md
jmchapman: Right. That could save us some work, see later work by Allais et al which generalises renaming, substitution and evaluation. Note also you want to generalise the proofs not just the programs.  I didn't try this the first time around as I was more confident with ren+sub+evaluation defined and reasoned about separately. This has never got to the top of the list of priorities so far.

jmchapman: Hereditary substitutions are also worth considering at some point.

plutus-metatheory/src/Type.lagda.md
jmchapman: I suppose you could refer to it as "universal quantification `Π`" like we say "iso-recursive type `μ`"

plutus-metatheory/src/Type/RenamingSubstitution.lagda.md
jmchapman: Also amazed by the link!

plutus-metatheory/src/Type/RenamingSubstitution.lagda.md
```diff
@@ -229,7 +252,7 @@ variable
   σ σ' : Sub Φ Ψ
 ```
 
-Extending a type substitution — used when going under a binder.
+Extending a type substitution — used when going under a binder. (This is called `lifts` in the [paper](https://ci.iog.io/build/1230848/download/1/paper.pdf#page=8) ).
```
jmchapman, line 255: How do we keep this link up-to-date?

jmchapman (approved):
That's a nice improvement!

plutus-metatheory/src/Type/RenamingSubstitution.lagda.md
ramsay-t: Is there a DOI link somewhere?

jmchapman: Our local copy is the best version as it includes a fix

plutus-metatheory/src/Utils.lagda.md
```diff
@@ -201,6 +224,9 @@ infixr 5 _∷_
 
 {-# COMPILE GHC List = data [] ([] | (:)) #-}
 
+```
+## DATA
+```
```
kwxm, line 229: "kinds" near the bottom of the file  looks as if it should be a header.

plutus-metatheory/src/Utils.lagda.md
```diff
@@ -127,7 +138,9 @@ dec2Either : {A : Set} → Dec A → Either (¬ A) A
 dec2Either (yes p) = inj₂ p
 dec2Either (no ¬p) = inj₁ ¬p
 
--- writer monad
+```
+# Writer Monad
+```
```
kwxm, line 143: There's some weird alingment below: `where` is on  a new line for `WriterMonad`.

plutus-metatheory/src/Utils.lagda.md
```diff
@@ -57,6 +63,7 @@ cong₃ f refl refl refl = refl
                     subst P p z ≡ subst P q z
 ≡-subst-removable P refl refl z = refl 
  ```
+## Natural Sum Type
```
kwxm, line 66: "sum m" -> "sum to m" below?

plutus-metatheory/src/Utils.lagda.md
kwxm: I think the comment "We also have a kind called # which is used for sized integers and bytestrings" is outdated now: it's used for type variables which are polymorphic over built-in types now (* is used for polymorphism over general terms),

plutus-metatheory/src/Utils.lagda.md
ramsay-t: Good spot :)

plutus-metatheory/src/Type.lagda.md
kwxm, line 84: What's the little title/layout table at the start of the document?  It looks as if it's out of place.

Later: no, I see that they all have that.

kwxm (approved):
Looks good.  I found a couple of small things that have nothing to do with the PR though.

plutus-metatheory/src/Type.lagda.md
ramsay-t: It is supposed to set up stuff in the Markdown - presumably it makes the title tags for the HTML header? I suspect it is ignored by what we are currently using to make docs, but it is probably good to have it there.

PR #5855, opened 2024-03-25
Add plugin options: `preserve-logging` and `inline-constants`

### LINKED ISSUE #5853: There should be a plugin option for each behavior controlled by `conservative-optimisation`

The `conservative-optimisation` flag implies a few different things. When it is off:
- The float-in pass is more aggressive and may end up increasing the script cost.
- Constants are always inlined, which may end up increasing the script size.
- Builtin applications are always evaluated at compile time, which may cause fewer trace messages to be emitted at runtime.

Currently there's no knob to control any of these things, but there should be for each one of them.

_No response_

zliu41:
Currently, `conservative-optimisation` implies (or negates) `relaxed-float-in`, `inline-constants` and `preserve-logging`, but only `relaxed-float-in` is a plugin option by itself. This PR adds plugin options for the other two. Closes #5853


doc/read-the-docs-site/reference/writing-scripts/compiler-options-table.rst
```diff
@@ -73,6 +73,12 @@
      - Dump Untyped Plutus Core
 
 
+   * - ``inline-constants``
+     - Bool
+     - False
```
ana-pantilie, line 78: What does this `False` mean here?

effectfully: The default value perhaps?

effectfully (approved):
LGTM

doc/read-the-docs-site/reference/writing-scripts/compiler-options-table.rst
ana-pantilie: Shouldn't the default be true or am I confused?

zliu41: > Shouldn't the default be true or am I confused?

Right.. I don't know how we got `False` here. I'll need to take a look.

zliu41: @ana-pantilie Good catch. Fixed.

PR #5858, opened 2024-03-26
Update dev-onboarding-quick-setup.md

joseph-fajen:
Changing the name of the Haskell file in this contract example to reflect that we are now using a different example than before.


michaelpj (approved):

michaelpj:
Seems like you figured out the PR thing! :muscle:

PR #5863, opened 2024-03-28
More reliable short-cicruiting behavior for `||` and  `&&` operators.

### LINKED ISSUE #4114: overzZzealous: Non-lazy Semantics of || and && in Plutus Core
Hi there,

While developping tools with [Hachi][gh:hachi], we noticed something that we believe is a bug in the compiler from Plutus to Plutus Core.

Best,  
Nicolas

EDIT: It seems to not be considered a bug. Since we believe it will still come up regularly in issues here and there, we decided to give this peculiarity the sweet name of “overzZzealous”. See also [our blog post on the topic][blogpost].

## Area

- [x] *Plutus Foundation*
- [ ] *Plutus Application Framework*
- [ ] *Marlowe*
- [ ] *Other*

## Summary

The compiler from Plutus to Plutus Core seems to compile `||` and `&&` as non-lazy, that is where the right-hand side is evaluated regardless of what the left-hand side evaluates to.

## Steps to reproduce

In general:

1. Take any Plutus code that runs on-chain.
2. Introduce side-effects on the right of a `||` or `&&`. (eg. `if True || error () then do_something else True`)
3. Observe that the side-effect is triggered.

More guided, within the *Plutus Application Backend (PAB)* example:

1. Clone the [input-output-hk/plutus-starter][gh:iohk/ppsp] project.
2. Follow the README to setup the project and to play the game with the PAB. Check that everything works: you get a validation error if you guess wrong, and the transation is validated if you guess right.
3. Update `validateGuess` in `./examples/src/Plutus/Contracts/Game.hs` to introduce an `error ()` side-effect on the right of an `||`. We suggest to use: `validateGuess hs cs _ = if True || error () then isGoodGuess hs cs else True`.
4. Play the game again and observe that you get a validation error whether you guess right or wrong.

Check [our fork of the Plutus Platform starter project][gh:hachi/ppsp] for a more guided version, including Shell scripts to play and modify the code in your place.

## Expected behavior

We would expect `||` and `&&` to be lazy and to only evaluate their second argument only if the first one evaluates to `False` and `True` respectively. This expectation is based on the fact that it is the case in a lot of languages and therefore we believe programmers will expect this as well. Moreover, it seems to be the case both [in Haskell][haskell-or] and [in off-chain Plutus][plutus-or].

## System info

Tested on several machines, all with latest Plutus.

- 1:
  - OS: NixOS 21.05, up-to-date as of Friday, October 15th, 2021.
  - GHC: 8.10.4
- 2:
  - OS: MacOS 11.6
  - GHC: 8.10.4

## Attachments

See our blog post on the topic: [*overzZzealous*: the Peculiar Semantics of `||` and `&&` in Plutus Core][blogpost].

[blogpost]: https://blog.hachi.one/post/overzzzealous-peculiar-semantics-or-and-plutus-core

[gh:iohk/ppsp]:  https://github.com/input-output-hk/plutus-starter
[gh:hachi]:      https://github.com/HachiSecurity
[gh:hachi/ppsp]: https://github.com/HachiSecurity/overzzzealous-in-plutus-starter

[haskell-or]: https://hackage.haskell.org/package/base-4.15.0.0/docs/Prelude.html#v:-124--124-
[plutus-or]:https://github.com/input-output-hk/plutus/blob/75ccaf4/plutus-tx/src/PlutusTx/Bool.hs#L30

Unisay:
Before the fix:
```
  ShortCircuiting operators
    (&&) short-circuits with GHC optimisations:    OK
    (||) short-circuits with GHC optimisations:    OK
    (&&) short-circuits without GHC optimisations: FAIL
      test/ShortCircuit/Spec.hs:50:
      An error has occurred:  User error:
      Use -p '/ShortCircuiting/&&/(&&) short-circuits without GHC optimisations/' to rerun this test only.
    (||) short-circuits without GHC optimisations: FAIL
      test/ShortCircuit/Spec.hs:50:
      An error has occurred:  User error:
      Use -p '/ShortCircuiting/&&/(||) short-circuits without GHC optimisations/' to rerun this test only.

2 out of 4 tests failed (0.01s)
```
After the fix:
```
  ShortCircuiting operators
    (&&) short-circuits with GHC optimisations:    OK
    (||) short-circuits with GHC optimisations:    OK
    (&&) short-circuits without GHC optimisations: OK
    (||) short-circuits without GHC optimisations: OK

All 4 tests passed (0.01s)
```

The fix is implemented as a code rewrite when compiling an expression:
`a && b` is rewritten as `if a then b else False`
`a || b` is rewritten as `if a then True else b`

Based on [this](https://github.com/IntersectMBO/plutus/pull/4118) un-merged PR.

Closes #4114


plutus-tx-plugin/test/ShortCircuit/WithGHCOptimisations.hs
```diff
@@ -0,0 +1,16 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# OPTIONS_GHC -O1 #-}
+
+module ShortCircuit.WithGHCOptimisations
```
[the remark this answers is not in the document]
Unisay: Done

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -685,6 +685,18 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
     _                  -> throwPlain $ CompilationError "No info for ByteString builtin"
 
   case e of
+    {- Note [Lazy boolean operators]
+      (||) and (&&) have a special treatment: we want them lazy in the second argument,
+      as this is the behavior in Haskell and other PLs.
+      Covered by this spec: plutus-tx-plugin/test/ShortCircuit/Spec.hs
+    -}
+    -- Lazy ||
+    GHC.App (GHC.App (GHC.Var fid) a) b | GHC.getOccString fid == "||" -> do
```
zliu41, line 694: - Rather than adding two cases here, you can use a similar approach to how we deal with things in `PlutusTx.Builtins.Internal`. I don't know if the latter is necessarily better, but it's better to be consistent, unless there's a strong reason not to.

- Don't use string literals here. They are error prone, and you aren't even checking the module names, which means that "||" matches any `||`, not just `PlutusTx.Bool.||`. It's much better to use TH name here - `'PlutusTx.Bool.||`.

- We need a Note explaining why we don't simply make use of `PlutusIR.Transform.RewriteRules`, and add a rewrite rule. That seems a simpler solution.

zliu41 (commented):
@Unisay A few more comments.

`PlutusTx.Bool.&&` and `||` should now be marked `NOINLINE`, since we don't need their unfoldings.

> I don't really like special-casing certain functions in the compiler.

@michaelpj Yeah but if anything should be special-cased, it's `&&` and `||`, since they are special in almost all other strict languages.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
michaelpj: > Rather than adding two cases here, you can use a similar approach to how we deal with things in PlutusTx.Builtins.Internal. I don't know if the latter is necessarily better, but it's better to be consistent, unless there's a strong reason not to.

I think these are rather different to how the builtins work, so it's fine for them to be here. We do have various special cases for Haskell functions in this area, but they're more special stuff like `unpackCString` or whatever.

> We need a Note explaining why we don't simply make use of PlutusIR.Transform.RewriteRules, and add a rewrite rule. That seems a simpler solution.

I think that might be worse, since we _can't_ match on the TH name there!

michaelpj: Have a look at how the builtins do it. We already have a way of carrying around TH names

zliu41: > I think these are rather different to how the builtins work, so it's fine for them to be here.

I think they are exactly the same? When we see `PlutusTx.Builtins.Internal.addInteger`, we don't look at the unfolding, but directly interpret it as PIR term `(builtin addInteger)`. Here we directly interpret `||` as `if ... then ... else ...`.

> I think that might be worse, since we can't match on the TH name there!

Maybe but there still should be a Note

zliu41: Ok, I guess one difference than builtins is that we are rewriting the saturated applications of `||`, not `||` itself.

zliu41:
There's a Note [Short-circuit evaluation for && and ||] which is now out of date.

The Note should be rewritten to explain that saturated applications of `&&` and `||` always short-circuit, but unsaturated applications won't.

zliu41:
I wonder if we should raise an error if `||` and `&&` is used unsaturated. Or at least a warning, but I don't know that the Plutus Tx compiler has the ability to emit warnings.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
Unisay: Yes, this also works:
```haskell
  boolOperatorOr <- GHC.getName <$> getThing '(PlutusTx.Bool.||)
  boolOperatorAnd <- GHC.getName <$> getThing '(PlutusTx.Bool.&&)
  case e of
    {- Note [Lazy boolean operators]
      (||) and (&&) have a special treatment: we want them lazy in the second argument,
      as this is the behavior in Haskell and other PLs.
      Covered by this spec: plutus-tx-plugin/test/ShortCircuit/Spec.hs
    -}
    -- Lazy ||
    GHC.App (GHC.App (GHC.Var var) a) b | GHC.getName var == boolOperatorOr ->
      compileExpr $ GHC.mkIfThenElse a (GHC.Var GHC.trueDataConId) b
    -- Lazy &&
    GHC.App (GHC.App (GHC.Var var) a) b | GHC.getName var == boolOperatorAnd ->
      compileExpr $ GHC.mkIfThenElse a b (GHC.Var GHC.falseDataConId)
```

PR #5865, opened 2024-03-28
[AST] [Evaluation] Use 'SmallArray' in 'Case'

effectfully:
Another competitor. Previous in the series: #5818, #5813, #5816.


michaelpj:
Looks like `Vector` is faster? Could be measurement error, but the difference looks big and consistent enough to be real?

effectfully:
> Looks like `Vector` is faster? Could be measurement error, but the difference looks big and consistent enough to be real?

Yes, let me rerun and check the diff again, but otherwise `Vector` is clearly better, which I find really surprising. Maybe it's fusion that's kicking in somehow.

michaelpj:
I'll rebase my branch and rerun the benchmarks in case something has changed

effectfully:

PR #5868, opened 2024-03-29
review dev-onboarding-quick-setup.md

nstanford5:
Overall it looks great -- though I still have some questions..

1. Where is the Plutus compiler? I don't see instructions on installing it -- am I supposed to clone the Plutus repo and build something? Is it in Nix?

2. What if I'm not using Nix? It says you need 'libsodium' 'blst' and 'secp256k1' -- but how? Can they be installed at the command line? Do they need to be built from source?

3. Plutus formatting. This may be the standard -- but it made coding along with the demo extremely tedious in regards to whitespace characters. It seems like the formatting is to make things line up well, but actually spacing these out is a nightmare. Plus, the "Data.Time" program will yell at you for using tabs instead of spaces. I assumed this meant I needed to be diligent about my formatting only to come upon the Plutus code and be thrown for a loop. I stopped coding along and started copy/pasting, which I try to avoid in this demos so that I can better grasp what is happening in the code.

The VestingContract example is great for an intro -- useful, but simple. I also really like the conclusion that details: What are the steps to using this in the real world? I think it is one of the most common questions after an example like this.

This demo could benefit from an accompanying video -- most of the public videos related to "Intro to Plutus" still use the Plutus Playground, I couldn't find any that were relevant to getting started as reflected in this doc.

You may consider this two separate documents...
1. Quickstart (environment and versions only)
2. My First Plutus Contract (VestingContract)

Overall this is greatly improved!

### DIFF
--- a/doc/docs-revamp/dev-onboarding-quick-setup.md
+++ b/doc/docs-revamp/dev-onboarding-quick-setup.md
@@ -47,8 +47,8 @@ After installing, configure Nix to use IOHK’s binary caches to speed up the bu
 Stack is a cross-platform program for developing Haskell projects. It handles dependency management, building, and testing your Haskell code. 
 
 To install the Stack: 
-1. Visit the [official Haskell Stack website](https://docs.haskellstack.org/). 
-2. Follow the installation instructions for your operating system (Linux, macOS, or Windows).
+1. For installation on Mac and Linux `curl -sSL https://get.haskellstack.org/ | sh`
+2. For installation on Windows see the [official Haskell Stack website](https://docs.haskellstack.org/).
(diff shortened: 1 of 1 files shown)


effectfully:
@zliu41 I think we should reassign this pull request, if we still care about it?

nstanford5:
This PR was never meant to be merged, just used for feedback submission. The content is likely outdated and this branch should likely just be closed.

effectfully:
Thank you @nstanford5!

@zliu41 could you please still take a look?

zliu41:
Thanks for the PR. This page is no longer needed.

PR #5874, opened 2024-04-04
How to handle builtins before they're costed  (PLT-9686)

kwxm:
When you're implementing a new PLC builtin you have to add a costing function, but you can't determine exactly what that should be until the implementation is complete and you've benchmarked it.  This caused some confusion recently because for a while we had bitwise conversion functions in master whose costing functions were identically zero and these we deployed on a testnet where you could use them for free,  This PR's an attempt to make sure that doesn't happen again. I've

 * Changed the default costing functions so that they have a maximal cost and added a comment in `Builtins.hs` to try to make sure that implementers use those.
 * Added a note in `plutus-ledger-api` to try to make sure that new builtins are put in `futurePV` until they're completely implemented and we're sure that we want to deploy them.
 * Updated the document that tells you how to cost a new builtin.
 * Added a test that no builtin has a costing function that's identically zero.  This is very close to PLT-881, which is about making sure that nothing has a cost that's less than or equal to zero: we could  probably modify the test here to do that by getting it to check multiple random inputs instead of just a single small one, which is what it does at the moment.  I didn't do that here because I thought it might be getting away from the point a bit, but we might want to follow that up later.


plutus-core/plutus-core/test/CostModelSafety/Spec.hs
```diff
@@ -0,0 +1,209 @@
...
+    checkBudget cekDelayBudget
+    checkBudget cekForceBudget
+    checkBudget cekApplyBudget
+    checkBudget cekBuiltinBudget
+    checkBudget cekConstrBudget
+    checkBudget cekCaseBudget
+
+-- Builtin costs
+
+{- Much of the code here is based on Evaluation.Spec and
```
michaelpj, line 88: Could we instead just extend those tests?

michaelpj (approved):
This looks very helpful!

plutus-core/plutus-core/test/CostModelSafety/Spec.hs
kwxm: It already took me quite a long time to get what's there to work and I was running out of JIRA points so I thought I'd quit while I was ahead.  @effectfully already had an issue for improved tests (PLT-881), so I thought I'd let him do something more comprehensive later.

plutus-core/cost-model/CostModelGeneration.md
```diff
@@ -136,11 +136,12 @@ codebase.
...
     ```
     toBuiltinMeaning XorByteString =
-        makeBuiltinMeaning
-            xorByteString
-            mempty
-            where xorByteString a b =
+        let xorByteStringDenotation :: Data.ByteString.ByteString -> Data.ByteString.ByteString -> Data.ByteString.ByteString
+            xorByteStringDenotation a b =
                       Data.ByteString.pack $ zipWith (Data.Bits.xor) (Data.ByteString.unpack a) (Data.ByteString.unpack b)
+        makeBuiltinMeaning
```
effectfully, line 142: Nit: forgot an `in`.

plutus-core/cost-model/CostModelGeneration.md
```diff
@@ -150,9 +151,13 @@ codebase.
     a check that the inputs are the same length.  For more complicated functions
     one might also put the implementation in a separate file.
 
-    The final argument of `makeBuiltinMeaning` contains the costing functions for
-    the relevant builtin.  Initially this should be set to `mempty`; we'll come
-    back and fix it later.
+    The final argument of `makeBuiltinMeaning` contains the costing functions
+    for the relevant builtin.  Initially this should be set to
+    `runCostingFun<N>Arguments . def` (where `<N>` is the number of arguments
```
effectfully, line 156: I had no idea this would work! Just checked, it does work. I don't like it TBH, it looks too "normal", I'd rather have an explicit `maxBound` in there or something else that screams "this isn't implemented properly yet". Even just a dedicated type class would be helpful I think:

```haskell
class NotCostedYet model where
    notCostedYet :: BuiltinCostModel -> model
```

plutus-core/cost-model/CostModelGeneration.md
```diff
@@ -275,18 +281,19 @@ the Cost Model" note.
...
+`def` in the definition of the builtin with the appropriate
+`param<builtin-name>` function:
 
-```
+    ```
     toBuiltinMeaning XorByteString =
+        let xorByteStringDenotation :: Data.ByteString.ByteString -> Data.ByteString.ByteString -> Data.ByteString.ByteString
+            xorByteStringDenotation a b =
+                      Data.ByteString.pack $ zipWith (Data.Bits.xor) (Data.ByteString.unpack a) (Data.ByteString.unpack b)
         makeBuiltinMeaning
```
effectfully, line 293: Nit: `in`.

Thank you for adopting the style by the way.

plutus-core/plutus-core/test/CostModelSafety/Spec.hs
```diff
@@ -0,0 +1,209 @@
...
+   PlutusCore.Generators.Hedgehog.Builtin.  The tests are complicated by the
+   fact that We can't just read the models from `builtinCostModels.json` and
+   check them since the actual models are defined by calls to `toBuiltinMeaning`
+   in PlutusCore.Default.Builtins and those are free to supply their own costing
+   functions and ignore the definitions in the JSON file.  The costing functions
+   end up in the `BuiltinMeanings` and we can't access them directly and feed
+   them suitable input sizes: instead we have to manufacture terms containing
+   constants of the desired sizes and feed them to the denotation of the
```
effectfully, line 96: I took a quick look, I think I could expose the raw costing functions in `BulitinMeaning`. Feel free to create a GitHub issue for me to do that if you want that functionality.

plutus-core/plutus-core/test/CostModelSafety/Spec.hs
```diff
@@ -0,0 +1,209 @@
...
+
+data SomeConst uni = forall a. uni `HasTermLevel` a => SomeConst a
+
+smallConstant :: forall (a :: GHC.Type). TypeRep a -> SomeConst DefaultUni
+smallConstant tr
+    | Just HRefl <- eqTypeRep tr (typeRep @()) = SomeConst ()
+    | Just HRefl <- eqTypeRep tr (typeRep @Integer) = SomeConst (0 :: Integer)
+    | Just HRefl <- eqTypeRep tr (typeRep @Int) = SomeConst (0 :: Integer)
+    | Just HRefl <- eqTypeRep tr (typeRep @Word8) = SomeConst (0 :: Integer)
+    | Just HRefl <- eqTypeRep tr (typeRep @LiteralByteSize) = SomeConst (0 :: Integer)
```
effectfully, line 110: I still think "ewww" whenever I see this. We have really good `QuickCheck` generators for built-in types and their values and here we're doing this horrible nonsense with `hedgehog` again. I wish one day I'll end up getting stuck somewhere for a couple of weeks and write proper `QuickCheck` generators of arbitrary builtin instantiations to replace this abomination. But it does work for the time being just fine.

I just wish we never had `hedgehog` in our codebase. Not that it's a bad library, but it is unsuitable for what we do.

plutus-ledger-api/src/PlutusLedgerApi/Common/ProtocolVersions.hs
```diff
@@ -20,6 +20,15 @@ import Data.Set qualified as Set
 import GHC.Generics (Generic)
 import Prettyprinter
 
+{- Note [Adding new builtins: protocol versions]
+
+  *** ATTENTION! ***
+  New built-in functions must initially be added under `futurePV` and should
+  only be moved to an earlier MajorProtocolVersion once they have been fully
+  implemented and costed and their release under the relevant protocol version
+  has been officially approved.
```
effectfully, line 29: Sounds like this could be a test?

effectfully (approved):
Great stuff, thank you a ton for doing it.

plutus-core/plutus-core/test/CostModelSafety/Spec.hs
effectfully: I'm going to murder it all at some point and rewrite it completely with `QuickCheck` ('cause we have really good  generators for it), so let's not worry about polishing those tests.

plutus-ledger-api/src/PlutusLedgerApi/Common/ProtocolVersions.hs
kwxm: How would we test that?  I'm not averse to doing it, but it's not clear to me what we could do.

plutus-core/plutus-core/test/CostModelSafety/Spec.hs
kwxm: > I'm going to murder it all at some point and rewrite it completely with QuickCheck 

Good!  I look forward to that.

plutus-core/cost-model/CostModelGeneration.md
kwxm: OK, I'll think about that.

plutus-core/plutus-core/test/CostModelSafety/Spec.hs
kwxm: It might be useful to have, but maybe in fact doing the tests this way is kind of helpful, since it means that the costing functions always get the arguments they'll get in real life: integers can't have size 0 for example, so we may not want to check some costing functions at 0.  I'm not sure about that.

plutus-core/cost-model/CostModelGeneration.md
kwxm: >  I'd rather have an explicit maxBound in there or something else that screams "this isn't implemented properly yet"

I added a suitable typeclass with a method called `unimplementedCostingFun` that you're supposed to use until you've got a proper one.  Maybe I should have made it `costingFunUNIMPLEMENTED` or something.

kwxm: I couldn't make the cost of an "unimplemented" costing function `maxBound` because then the evaluator just fails all the time, so I made it ten times the current maximum on-chain CPU limit, which should be safe.  I tried to remove the `def` instances, but it turns out that without them you get an error [here](https://github.com/IntersectMBO/plutus/blob/9ec15b09238cb139b105a1c79f0f346a3a3350a9/plutus-core/testlib/PlutusIR/Test.hs#L191) because `compileWIthOpts` requires them (I'm not quite sure why).  It looks as if it copes OK if the default values are all `maxBound` (they were previously 0), so I've left them like that to be safe.

michaelpj: We could add a hlint warning for it!

effectfully: > We could add a hlint warning for it!

image

PR #5875, opened 2024-04-04
Remove plutus-metatheory from plutus-benchmark-common

kwxm:
`plutus-benchmarks` contains a library called `plutus-benchmark-common` that contains some shared code.  Some of the benchmarks are for the Agda evaluator in `plutus-metatheory`, so `plutus-benchmark-common` depended on that.  This sometimes caused `plutus-metatheory` to be built for things like the validation benchmarks, which don't care about the benchmarks: this takes a long time and is very annoying.  This PR removes the `plutus-metatheory` dependency and puts the relevant code into a library of its own, so the lengthy Agda build only happens when you really need it.

### DIFF
--- a/plutus-benchmark/plutus-benchmark.cabal
+++ b/plutus-benchmark/plutus-benchmark.cabal
@@ -93,7 +93,6 @@ library plutus-benchmark-common
     , flat               ^>=0.6
     , plutus-core        ^>=1.25
     , plutus-ledger-api  ^>=1.25
-    , plutus-metatheory
     , plutus-tx          ^>=1.25
     , tasty
     , tasty-golden
--- a/plutus-benchmark/plutus-benchmark.cabal
+++ b/plutus-benchmark/plutus-benchmark.cabal
@@ -37,7 +37,7 @@ common turn-off-plugin-warning
 -- transitively.  If you miss it out somewhere then nix will probably produce an
 -- error message saying "the component is not buildable in the current
 -- environment" if eg the nix shell supplies an unsupported GCH version.
--- See the section on GHC versions in CONTRIBUTING
+-- See the section on GHC versions in CONTRIBUTING.
 common ghc-version-support
   if (impl(ghc <9.6) || impl(ghc >=9.7))
     buildable: False
@@ -84,16 +84,17 @@ library plutus-benchmark-common
 
   other-modules:   Paths_plutus_benchmark
   build-depends:
-    , base          >=4.9   && <5
+    , base               >=4.9   && <5
     , bytestring
     , criterion
     , deepseq
     , directory
     , filepath
-    , flat          ^>=0.6
-    , plutus-core   ^>=1.25
+    , flat               ^>=0.6
+    , plutus-core        ^>=1.25
+    , plutus-ledger-api  ^>=1.25
     , plutus-metatheory
-    , plutus-tx     ^>=1.25
+    , plutus-tx          ^>=1.25
     , tasty
     , tasty-golden
     , temporary
--- a/plutus-benchmark/plutus-benchmark.cabal
+++ b/plutus-benchmark/plutus-benchmark.cabal
@@ -139,7 +139,7 @@ benchmark nofib
   hs-source-dirs: nofib/bench
   other-modules:  Shared
   build-depends:
-    , base                     >=4.9 && <5
+    , base                     >=4.9     && <5
     , criterion                >=1.5.9.0
     , nofib-internal
     , plutus-benchmark-common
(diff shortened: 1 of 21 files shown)


plutus-benchmark/plutus-benchmark.cabal
```diff
@@ -550,13 +550,25 @@ test-suite plutus-benchmark-marlowe-tests
 
 -- TODO: Add benchmarks for the executable semantics when we have a UPLC version
 
+library agda-internal
+  import:          lang, ghc-version-support
+  ghc-options:     -Wno-unused-packages
```
Unisay, line 555: I wonder why this is needed? Would be nice to have an answer in a comment.

Unisay (approved):

Unisay:
> Hmmm. I'm getting repeated CI failures. It all builds without problems locally, so I'm not sure if the CI problems are genuine or not.

I can reproduce the failure locally by running this command:
```
nix build --no-link .\#hydraJobs.x86_64-linux.required
```
Note:
```
cabal build all
```
works well in the default shell.

michaelpj (approved):

plutus-benchmark/plutus-benchmark.cabal
```diff
@@ -550,13 +550,25 @@ test-suite plutus-benchmark-marlowe-tests
 
 -- TODO: Add benchmarks for the executable semantics when we have a UPLC version
 
+library agda-internal
+  import:          lang, ghc-version-support
```
michaelpj, line 554: You included `ghc-version-support`, which marks things as unbuildable for some GHCs. That's not true here, and it's breaking things since the agda benchmarks are buildable on all GHCs.

plutus-benchmark/plutus-benchmark.cabal
kwxm: I'll take that out again: initially it wasn't there and when I started getting the failures I put it in to see if it helped since it was in some of the other things in the file.

plutus-benchmark/plutus-benchmark.cabal
kwxm: I have to say that the nix error message isn't very illuminating.

michaelpj: Well, it's a cabal error message, and it tells you that this component isn't buildable. I looked for reasons why it might not be buildable, that's how I found it :shrug:

kwxm: > Well, it's a cabal error message,

Oh. That's not what I was getting earlier, or if it was it was buried inside [2798 lines of nix crap](https://ci.iog.io/eval/39146#tabs-errors).

plutus-benchmark/plutus-benchmark.cabal
kwxm: @yura: `-Wno-unused-packages` was scattered about in various places in the cabal file but now I've worked out exactly where it's needed and introduced a sort of alias that makes it a bit clearer what's going on (although I'm still not totally sure what _is_going on).  I'd forgotten why we needed this, so thanks for making me think about it.

plutus-benchmark/plutus-benchmark.cabal
kwxm: OK, I think I understand what's happening now and I've managed to sort it all out.  Thanks for the pointers.

plutus-benchmark/plutus-benchmark.cabal
```diff
@@ -557,13 +541,25 @@ test-suite plutus-benchmark-marlowe-tests
 
 -- TODO: Add benchmarks for the executable semantics when we have a UPLC version
 
+library agda-internal
+  import:          lang, turn-off-plugin-warning, ghc-version-support
```
michaelpj, line 545: Why are you still including `ghc-version-support`? You've added it to the stuff that depends on this, but the right thing is to just _remove_ it: none of these depend on the plugin!

kwxm: > none of these depend on the plugin!

Yes they do! `nofib-agda-cek` and `marlowe-agda-cek` both depend on the plugin and they were getting it from `agda-internal`: they definitely don't work with 8.10.  I'm just about to push a fix that'll make them depend on the plugin directly and then `validation-agda-cek` should work with all versions.  I think there are no unnecessary plugin dependencies in the other tests.

michaelpj: It certainly sounds like those two components need to be marked as not buildable on earlier GHCs, but I don't see why they should need to depend on the plugin directly?

PR #5876, opened 2024-04-04
Refer to && and || by their qualified TH names.

Unisay:
- [x] Special (lazy) handling of `&&` and `||` is better done by referring to them using their fully-qualified `TH.Name`s.
- [x] `&&` and `||` should not be inlined as Plutus compiler handles them specially.
- [x] Remove outdated note.

Follow-up to [this thread](https://github.com/IntersectMBO/plutus/pull/5863#pullrequestreview-1975184625).

### DIFF
--- a/plutus-tx-plugin/test/Plugin/Coverage/9.6/coverageCode.pir.golden
+++ b/plutus-tx-plugin/test/Plugin/Coverage/9.6/coverageCode.pir.golden
@@ -104,6 +104,27 @@
     )
     (termbind
       (nonstrict)
+      (vardecl `&&` (fun Bool (fun Bool Bool)))
+      (lam
+        ds
+        Bool
+        (lam
+          x
+          Bool
+          {
+            [
+              [
+                { [ Bool_match ds ] (all dead (type) Bool) } (abs dead (type) x)
+              ]
+              (abs dead (type) False)
+            ]
+            (all dead (type) dead)
+          }
+        )
+      )
+    )
+    (termbind
+      (nonstrict)
       (vardecl
         `==`
         (all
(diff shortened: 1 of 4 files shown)


michaelpj (approved):

michaelpj:
Seems to have changed a bunch of tests, probably those `NOINLINE`s

Unisay:
> Seems to have changed a bunch of tests, probably those `NOINLINE`s

Indeed. What's peculiar is that cpu/mem results are slightly better when we let GHC inline those definitions. 
(both ways work correctly, though)

plutus-tx-plugin/test/Plugin/Coverage/9.6/coverageCode.pir.golden
```diff
@@ -102,6 +102,27 @@
       )
       equalsInteger
     )
+    (termbind
+      (nonstrict)
+      (vardecl `&&` (fun Bool (fun Bool Bool)))
```
michaelpj, line 107: This shouldn't exist, surely?

Unisay: Ok, so this solution never removes the function declaration itself, only rewrites its fully saturated applications...
M.b. there are partial applications left (in which case I'd expect this binding to remain)?

michaelpj: The application in this example looks saturated to me...

Unisay: Ok, this `&&` isn't the `PlutusTx.Bool.(&&)` (which is exactly the point of this PR - tell different `&&`s apart)

michaelpj: Haha, good spot. Arguably that's a bug in the test

PR #5877, opened 2024-04-04
AsData type delegates its schema instance implementation to its non-AsData counterpart.

Unisay:
Better way of defining `HasSchema` instance as explained [here](https://github.com/IntersectMBO/plutus/pull/5837#discussion_r1549451811).

### DIFF
--- a/plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
+++ b/plutus-tx-plugin/test/Blueprint/Tests/Lib.hs
@@ -19,6 +19,7 @@ module Blueprint.Tests.Lib
   , module AsData
   ) where
 
+import Blueprint.Tests.Lib.AsData.Blueprint qualified as SchemaTypes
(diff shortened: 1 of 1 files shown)


michaelpj:
And the test itself should change to no longer import the type from `Blueprints.Lib`, rigth?

michaelpj:
Ugh. Then this is rather pointless. I don't see why it can't be driven by the `HasSchema` instances - indeed it seems like the instance head in this case tells us the nested types, but whatever.

Unisay:
> Ugh. Then this is rather pointless. I don't see why it can't be driven by the `HasSchema` instances - indeed it seems like the instance head in this case tells us the nested types, but whatever.

If only I knew about this difficulty with `AsData` in the beginning I wouldn't use the `TF/Generic`-based derivation machinery, I'd employ heavier gun (Template Haskell) 🤷🏼‍♂️

PR #5879, opened 2024-04-05
Kwxm/flat/canonincal data encoding (PLT-9705)

kwxm:
As described in https://github.com/IntersectMBO/plutus/issues/5852, the `flat` encoding of `Data` objects did not always have the canonical format described in the Plutus Core specification, which is arguably a bug. This should fix the problem.  See the code comments for details.  The actual fix only requires one extra word, but I've added some tests too.

### DIFF
--- a/plutus-core/changelog.d/20240405_015609_kenneth.mackenzie_canonincal_data_encoding.md
+++ b/plutus-core/changelog.d/20240405_015609_kenneth.mackenzie_canonincal_data_encoding.md
@@ -1,0 +1,4 @@
+### Changed
+
+- The `flat` encoding of the `Data` type has been modified slightly to make sure that 
+  the result is always in the canonical format described in the Plutus Core specification.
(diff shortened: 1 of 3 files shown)


michaelpj (approved):
Good stuff, having the test is helpful.

michaelpj:
Do we need to check that we still deserialize non-canonical things correctly? I think surely we must, otherwise everything would be broken.

kwxm:
> Do we need to check that we still deserialize non-canonical things correctly? I think surely we must, otherwise everything would be broken.

Good point.  We certainly do deserialise non-canonical things correctly, but we don't have a test for that.  Maybe we could do that by serialising and deserialising some lazy bytestrings.  They definitely do get serialised non-canonically because if you take out  the `toStrict` added in this PR then the new canonicity test fails occasionally (I had to set the limit to 10000 to get that to happen consistently though).

PR #5881, opened 2024-04-05
Add some tests that non-canonically encoded bytestrings decode correctly

kwxm:
(PLT-9705 continued)

The Plutus Core specification recommends that bytestrings should be serialised in a certain canonical way, and #5879  made sure that our flat serialisation for Data  obeyed this recommendation.  However, when we're deserialising from flat we also accept _non-canonical_ encodings of bytestrings and Data, and this PR adds some tests to make sure that we really do accept those.

### DIFF
--- a/plutus-core/untyped-plutus-core/test/Flat/Spec.hs
+++ b/plutus-core/untyped-plutus-core/test/Flat/Spec.hs
@@ -54,12 +54,14 @@ single byte giving its length and the end of the bytestring is marked by a
 zero-length chunk.  The encoding is canonical if it consists of a (possibly
 empty) sequence of 255-byte chunks followed by an optional smaller block
 followed by an empty block.  See Note [Flat serialisation for strict and lazy
-bytestrings]. -}
+bytestrings].  See Appendix C of the Plutus Core specification for details of
+the `flat` encoding, in particular Section C.2.5. -}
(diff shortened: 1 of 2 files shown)


plutus-core/untyped-plutus-core/test/Flat/Spec.hs
```diff
@@ -85,6 +90,120 @@ test_canonicalByteString :: TestTree
...
+as well. -}
+test_nonCanonicalByteStringDecoding :: TestTree
+test_nonCanonicalByteStringDecoding =
+  let target = "This is a test." :: BS.ByteString
+
+      ch :: Char -> Word8
+      ch = fromIntegral . ord
+
+      input1 = BS.pack [ 0x01
+                       , 0x01, ch 'T'
```
zliu41, line 110: I assume 0x01 is the separator of the chunks, but can you add a comment pointing to where it comes from - e.g., a spec?

zliu41 (approved):

plutus-core/untyped-plutus-core/test/Flat/Spec.hs
kwxm: > I assume 0x01 is the separator of the chunks

That's the tag for a bytestring: it said that in one of the earlier tests, but I've repeated it here and added a reference to the Plutus  Core specification. The other 0x01 bytes here are the starts of chunks of length 1.

PR #5883, opened 2024-04-08
Improved error message for the `mustBeReplaced`

Unisay:
Before:
```
ghci> PlutusTx.List.head []
*** Exception: This must be replaced by the core-to-plc plugin during compilation: error
  error, called at src/PlutusTx/Utils.hs:8:26 in plutus-tx-1.25.0.0-inplace:PlutusTx.Utils
```

After:
```
PlutusTx.List.head []
*** Exception: The "error" placeholder must have been replaced by the core-to-plc plugin 
during compilation. The most likely reason the replacement didn't happen is that a PlutusTx 
function was evaluated by off-chain code.
  error, called at src/PlutusTx/Utils.hs:8:3 in plutus-tx-1.25.0.0-inplace:PlutusTx.Utils
```

### DIFF
--- a/plutus-tx/src/PlutusTx/Utils.hs
+++ b/plutus-tx/src/PlutusTx/Utils.hs
@@ -1,8 +1,12 @@
--- editorconfig-checker-disable-file
 module PlutusTx.Utils where
 
 -- We do not use qualified import because the whole module contains off-chain code
 import Prelude as Haskell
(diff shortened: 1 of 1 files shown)


michaelpj (approved):

michaelpj:
This change is fine though.

PR #5885, opened 2024-04-08
[Builtins] Make 'BuiltinSuccess' the first constructor

effectfully:
This makes `BuiltinSuccess` the first constructor of `BuiltinResult`, see the comment there of why we want that.

See [this](https://github.com/IntersectMBO/plutus/pull/5885#issuecomment-2151102873) comment for the very ambiguous benchmarking results.


effectfully:
@aadaa-fgtaa is telling me that simply making the constructor first isn't anywhere near being a reliable way to force GHC to branch-predict it and that for e.g.

```
data Foo = Bar | Baz

f Bar = 1#
f Baz = 2#
```

GHC would branch-predict the second one. Making the `BuiltinSuccess` case recursive would help, except... it already is! So why does reordering the constructors help? God knows.

@aadaa-fgtaa also referenced [this](https://gitlab.haskell.org/ghc/ghc/-/issues/849) GHC ticket.

We don't lose anything by reordering the constructors though, so if it helps some benchmarks, let's just do it and pray that GHC doesn't break things.

effectfully:
One run:

Screenshot from 2024-06-06 01-03-26

Another run, same benchmarks, same commits:

Screenshot from 2024-06-06 01-04-22

I'm not sure what I should take from that, I've now run benchmarks in this PR dozens of times and they're sometimes no change and sometimes a major speedup.

Let's just merge it and get over it I suppose.

plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
```diff
@@ -58,9 +58,19 @@ data BuiltinError
...
 -- we'd have to use @text-builder@ or @text-builder-linear@ or something of this sort.
 data BuiltinResult a
-    = BuiltinFailure (DList Text) BuiltinError
-    | BuiltinSuccess a
+    = -- 'BuiltinSuccess' is the first constructor to make it a bit more likely for GHC to
+      -- branch-predict it (which is something that we want, because most builtins return this
+      -- constructor). It is however not guaranteed that GHC will predict it, because even though
+      -- it's likely going to be a recursive case (it certainly is in the CEK machine) and thus the
+      -- constructor has precedence over 'BuiltinFailure', it doesn't have precedence over
+      -- 'BuiltinSuccessWithLogs', since that case is equally likely to be recursive.
```
Unisay, line 66: I wonder what happens if instead of 
```
 | BuiltinSuccess a
 | BuiltinSuccessWithLogs (DList Text) a
```
we have
```
 | BuiltinSuccess (DList Text) a
```
with an empty `DList Text` in case when there are no logs?

Unisay (approved):

plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
effectfully: We'll need to convert it to a list and either match on the list before doing logging to see if we need to do logging in the first place, which is quite a bit of unnecessary allocation (first the `DList` together with a pointer to it, then the list) plus we need to perform the match which will probably confuse the branch predictor as well. Or we could do logging without checking whether the list is empty, but that's calling a function passed dynamically, so another bunch of allocation and pointer chasing.

I.e. `BuiltinResult` is specifically designed to avoid all of that in the most common case of success with no logs.

Do feel free to try it out though if you're curious.

PR #5886, opened 2024-04-09
Update some more benchmarks to use correct evaluation context (PLT-9713)

kwxm:
This updates the BLS12-381 and Marlowe benchmarks to use the more precise evaluation strategy introduced in #5867.

Don't review it yet: I want to update some of the benchmarks but before I do that I want to get a fair comparison with the old versions.


plutus-benchmark/bls12-381-costs/test/9.6/bls12-381-costs.golden
kwxm, line 1: The costs in this file have changed because the affected examples are doing a fold that previously started off with a sort of random accumulator.  I've changed the initial values to something more sensible now but it means that a little more work is required at the start

kwxm:
I''m going to close this in favour of #5892 .  That makes the same changes but has less noise from the benchmarking. Also this one seems to be broken for some reason.

PR #5890, opened 2024-04-10
Refactor NoTrace spec

Unisay:
1.  Programmatically inspect UPLC term instead of using golden files.
 2. `no-conservative-optimizations`  otherwise some `trace`s are removed.
 3. Add negative test cases (no traces are removed when `no-remove-trace` flag is set)


plutus-tx-plugin/test/Plugin/NoTrace/WithTraces.hs
```diff
@@ -0,0 +1,40 @@
+{-# LANGUAGE BlockArguments    #-}
+{-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE TypeApplications  #-}
+{-# OPTIONS_GHC -fplugin PlutusTx.Plugin #-}
+-- The plugin uses non-conservative optimizations by default and they remove some traces.
+-- We disable them to make sure that no traces are removed.
+{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:conservative-optimisation #-}
```
michaelpj, line 9: the plugin options take effect in the module where you call `compile`, so if you wanted you could share the code between the two modules here, but not a big deal

PR #5892, opened 2024-04-11
Update some more benchmarks to use correct evaluation context (PLT-9713) [second attempt]

kwxm:
The previous PR seems to be broken.

This updates the BLS12-381 and Marlowe benchmarks to use the more precise evaluation strategy introduced in https://github.com/IntersectMBO/plutus/pull/5867.


kwxm:
Also I changed some of the bls12-381 benchmarks in a way that will slow them down a bit, or a lot in the n=0 case.

kwxm:
cabal bench bls12-381-benchmarks

kwxm:
The origin of this branch has vanished so I'm closing it in favour of #5903.

PR #5896, opened 2024-04-12
Make the case arguments of matching functions in PlutusTx.Bulitins non-strict

zliu41:
Dunno how this went unnoticed for so long, but obviously the case arguments shouldn't be strict.


michaelpj:
No, I don't agree with this. It has always been this way, and they just have exactly the caveat that if you want to delay evaluation of one of the cases you need to match on `() -> a` or whatever and apply it afterwards. These are just Normal Functions, and so they are strict. And anyway, I thought tildes on argument patterns didn't do anything unless GHC inlined the function?

bezirg:
We should perhaps write some comment that we should not use (~) in builtins and internals.
And that the user is responsible to non-strictify the arguments?

Unisay:
Regardless of the final decision I think that we should provide unit tests that assert effects of strictness annotations either way, analogously to how lazyness/strictness in boolean operator `False && error ()` influences evaluation results - this is testable (goldens aren't good for this purpose).

plutus-benchmark/ed25519-costs/test/9.6/ed25519-costs.golden
```diff
@@ -1,20 +1,20 @@
...
+   60   10893  (66.5%)      4227237512  (42.3%)         2393802  (17.1%) 
+   70   12620  (77.0%)      4930905912  (49.3%)         2789632  (19.9%) 
+   80   14346  (87.6%)      5634574312  (56.3%)         3185462  (22.8%) 
+   90   16073  (98.1%)      6338242712  (63.4%)         3581292  (25.6%) 
+  100   17800 (108.6%)      7041911112  (70.4%)         3977122  (28.4%) 
+  110   19527 (119.2%)      7745579512  (77.5%)         4372952  (31.2%) 
+  120   21253 (129.7%)      8449247912  (84.5%)         4768782  (34.1%) 
+  130   22980 (140.3%)      9152916312  (91.5%)         5164612  (36.9%) 
+  140   24707 (150.8%)      9856584712  (98.6%)         5560442  (39.7%) 
+  150   26434 (161.3%)     10560253112 (105.6%)         5956272  (42.5%)
```
effectfully, line 18: What??? This is a ridiculous optimization of memory usage.

plutus-tx-plugin/test/IsData/9.6/deconstructData.pir.golden
```diff
@@ -85,121 +31,141 @@ let
...
   ~`$fFromDataTuple2` :
      all a b.
        (\a -> data -> Maybe a) a ->
        (\a -> data -> Maybe a) b ->
        (\a -> data -> Maybe a) (Tuple2 a b)
     = `$fFromDataTuple2_$cfromBuiltinData`
+  !unsafeDataAsB : data -> bytestring = unBData
+  !unsafeDataAsI : data -> integer = unIData
+  !unsafeDataAsList : data -> list data = unListData
+  !unsafeDataAsMap : data -> list (pair data data) = unMapData
```
effectfully, line 132: Why aren't those getting inlined?

plutus-tx-plugin/test/IsData/9.6/deconstructData.pir.golden
```diff
@@ -85,121 +31,141 @@ let
...
+          (\(ds : Unit) ->
+             let
+               !ds : list (pair data data) = unsafeDataAsMap d
+             in
+             Nothing {integer})
+          (\(ds : Unit) ->
+             let
+               !ds : list data = unsafeDataAsList d
+             in
+             Nothing {integer})
```
effectfully, line 157: Good thing about pattern matching builtins is that they make this not only faster, but also quite shorter.

plutus-tx/src/PlutusTx/Builtins.hs
```diff
@@ -385,7 +385,7 @@ encodeUtf8 = BI.encodeUtf8
 
 {-# INLINABLE matchList #-}
 matchList :: forall a r . BI.BuiltinList a -> r -> (a -> BI.BuiltinList a -> r) -> r
-matchList l nilCase consCase = BI.chooseList l (const nilCase) (\_ -> consCase (BI.head l) (BI.tail l)) ()
+matchList l ~nilCase ~consCase = BI.chooseList l (const nilCase) (\_ -> consCase (BI.head l) (BI.tail l)) ()
```
effectfully, line 388: Here we explicitly rely on GHC inlining this definition for us. Can we instead make sure to inline it ourselves? Is there a general rule? How many other such cases are we missing?

It sucks that we don't want GHC to touch our code much (because it doesn't understand the semantics), while at the time we want to rely on it in some cases.

effectfully (approved):
Great catch. But this is one more argument in favor of Plutus Tx being weird.

effectfully:
> No, I don't agree with this. It has always been this way, and they just have exactly the caveat that if you want to delay evaluation of one of the cases you need to match on `() -> a` or whatever and apply it afterwards. These are just Normal Functions, and so they are strict. And anyway, I thought tildes on argument patterns didn't do anything unless GHC inlined the function?

I fully agree with you. But I agree with 53.2% -> 42.5% of max MEM even more.

michaelpj:
> I fully agree with you. But I agree with 53.2% -> 42.5% of max MEM even more.

I think this is just a path to making it totally impossible to reason about Plutus Tx code. I reluctantly agreed with the `&&` case, but this is IMO just terrible. _And_ it's unreliable since we're entirely at the mercy of GHC.

zliu41:
@michaelpj Right, adding `~` to arguments doesn't matter. I wrote the doc about this but I keep forgetting 🤦‍♂️
Clearly this needs to be documented in the Haddock of `matchList` and `matchData`

michaelpj:
>  Clearly this needs to be documented in the Haddock of matchList and matchData

And to the haddock of every single strict function we expose? Perhaps these are especially likely to trip people up, IDK :shrug:

effectfully:
@michaelpj 

> I think this is just a path to making it totally impossible to reason about Plutus Tx code. I reluctantly agreed with the `&&` case, but this is IMO just terrible. And it's unreliable since we're entirely at the mercy of GHC.

I agree with all of that. But:

1. I already can't really reason about Plutus Tx code. Maybe it's an old habit though, the behavior of `&&` was really confusing at times, but that is no longer the case
2. we're at the mercy of GHC here, but we're generally at its mercy, so why not benefit from the inevitable. GHC feels completely free to eta-expand functions for example and that changes the semantics for us. Should we insist on using `pedantic-bottoms` everywhere (perhaps yes)? Will it be enough (perhaps no)? Does GHC have a presimplifier or something of this sort? IIRC we cannot fully control GHC regardless, so if we can't, we don't have any option apart from trying to at least benefit from it some

If we can reliably reason about Plutus Tx code right now, then I'm not so sure we should do what we do here indeed.

zliu41:
@effectfully What you said is the imo the biggest and most annoying problem with Plutus Tx. And I don't see how we can fix it, outside of introducing lazy evaluation for Plutus Core, so that GHC's transformations become valid from Plutus Tx point of view.

That said, I think we can view Plutus Tx as two different subsets of Haskell:

1. The subset that the plugin can compile
2. The (smaller) subset that not only the plugin can compile, but the resulting Plutus Core has the exact same semantic as the original Haskell. This requires making everything strict, among other things.

I think we should try and make the second subset as large as possible and encourage people to write Plutus Tx using the second subset. What I did in this PR (adding `~`) makes the Haskell semantics deviate from the Plutus Core semantics which is not great.

PR #5897, opened 2024-04-12
Don't call `mustBeReplaced` in PlutusTx.Builtins.Internal

zliu41:
It makes no sense to say "placeholder must have been replaced by the core-to-plc plugin during compilation", when calling `PlutusTx.Builtins.Internal.error`. There's no placeholder - it's just an innocent use of `error`, and nothing needs to be replaced.

I was quite confused by this message when testing the constitution script.

### DIFF
--- a/plutus-tx/src/PlutusTx/Utils.hs
+++ b/plutus-tx/src/PlutusTx/Utils.hs
@@ -7,6 +7,4 @@ mustBeReplaced :: String -> a
 mustBeReplaced placeholder =
   error $
     "The " <> show placeholder <> " placeholder must have been replaced by the \
-      \core-to-plc plugin during compilation. \
-      \The most likely reason the replacement didn't happen is that a PlutusTx \
-      \function was evaluated by off-chain code."
+      \core-to-plc plugin during compilation."
--- a/plutus-tx/src/PlutusTx/Builtins/Internal.hs
+++ b/plutus-tx/src/PlutusTx/Builtins/Internal.hs
@@ -42,7 +42,6 @@ import PlutusCore.Crypto.Hash qualified as Hash
 import PlutusCore.Crypto.Secp256k1 qualified
 import PlutusCore.Data qualified as PLC
 import PlutusCore.Pretty (Pretty (..), display)
-import PlutusTx.Utils (mustBeReplaced)
 import Prettyprinter (viaShow)
 
 {-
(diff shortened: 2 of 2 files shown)


michaelpj (approved):
Yes, @Unisay and I were discussing this. I agree this is the wrong thing to put in a lot of cases.

effectfully (commented):
:+1:

plutus-tx/src/PlutusTx/Utils.hs
```diff
@@ -7,6 +7,4 @@ mustBeReplaced :: String -> a
 mustBeReplaced placeholder =
   error $
     "The " <> show placeholder <> " placeholder must have been replaced by the \
-      \core-to-plc plugin during compilation. \
-      \The most likely reason the replacement didn't happen is that a PlutusTx \
-      \function was evaluated by off-chain code."
+      \core-to-plc plugin during compilation."
```
Unisay, line 10: When I added this message:
_"The most likely reason the replacement didn't happen is that a PlutusTx function was evaluated by off-chain code."_ my motivation was to help users understand **why** the error happens, as in what causes it.

Otherwise, this message alone "_The 'foo' placeholder must have been replaced by the core-to-plc plugin during compilation._" could be frustrating as it doesn't help them to understand and  fix the cause.

zliu41: My understanding is that "placeholder" refers to `PlutusTx.Plugin.Utils.plc`. The plugin looks for `plc`, and replaces it with the actual compilation.

If the replacement does not happen, it is not because "a PlutusTx function was evaluated by off-chain code" (which one _can_ do - I don't think there's anything wrong with that). It is either because one uses `plc` directly but forgets to use the plugin, a bug in the plugin, or `plc` is used in a way that is not recognized by the plugin (the plugin requires `plc` to be applied to the right arguments to be triggered).

Unisay: Is seems to me that "a PlutusTx function was evaluated by off-chain code" happens when "one uses plc directly but forgets to use the plugin", right?

In either case, you list of reasons could be really useful for users to understand what they did wrong (modulo bug in the plugin), I'd extend the original message, not reduce it.

zliu41: > Is seems to me that "a PlutusTx function was evaluated by off-chain code" happens when "one uses plc directly but forgets to use the plugin", right?

I don't think so. "forgetting to use the plugin" means forgetting `{-# OPTIONS_GHC -fplugin PlutusTx.Plugin #-}`, which is totally different from evaluating Plutus Tx functions in off-chain code, which is a perfectly OK thing to do.

PR #5898, opened 2024-04-12
Add PlutusTx.List.indexBuiltinList

zliu41:
Is this useful? Let me know what you think.

### DIFF
--- a/plutus-tx/src/PlutusTx/ErrorCodes.hs
+++ b/plutus-tx/src/PlutusTx/ErrorCodes.hs
@@ -55,6 +55,8 @@ plutusPreludeErrorCodes = Map.fromDistinctAscList
   , ("PT18", "PlutusTx.Enum.Ordering.toEnum: bad argument")
   , ("PT19", "PlutusTx.List.last: empty list")
   , ("PT20", "PlutusTx.Ratio.recip: reciprocal of zero")
+  , ("PT21", "PlutusTx.List.indexBuiltinList: negative index")
+  , ("PT22", "PlutusTx.List.indexBuiltinList: index too large")
   ]
 
 -- | The error happens in TH generation of indexed data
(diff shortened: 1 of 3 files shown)


plutus-tx/src/PlutusTx/List.hs
```diff
@@ -246,6 +248,27 @@ xs0 !! n0 = go n0 xs0
...
+--
+--   >>> indexBuiltinList (toBuiltin [10, 11, 12]) 2
+--   12
+--
+indexBuiltinList :: forall a. BI.BuiltinList a -> Integer -> a
+indexBuiltinList xs0 i0
+  | i0 `Builtins.lessThanInteger` 0 = traceError builtinListNegativeIndexError
+  | otherwise = go xs0 i0 ()
+  where
+    go :: BI.BuiltinList a -> Integer -> () -> a
```
effectfully, line 261: I wonder if making it an argument instead of just inlining it actually helps performance.

effectfully (approved):

plutus-tx/src/PlutusTx/List.hs
zliu41: I'm not sure I understand - making what an argument?

michaelpj (approved):

plutus-tx/src/PlutusTx/List.hs
effectfully: `go :: BI.BuiltinList a -> Integer -> () -> a` vs `go :: BI.BuiltinList a -> Integer -> a`. Is that `()` needed? Does it help?

zliu41: Ah, I think you mean this:

```haskell
go :: BI.BuiltinList a -> Integer -> a
go xs i =
  Builtins.matchList
    xs
    (\_ -> traceError builtinListIndexTooLargeError)
    ( \hd tl _ ->
        if i `Builtins.equalsInteger` 0
          then hd
          else go tl (Builtins.subtractInteger i 1)
    )
    ()
```

I think they are the same but this does look prettier.

plutus-tx/src/PlutusTx/ErrorCodes.hs
```diff
@@ -55,6 +55,8 @@ plutusPreludeErrorCodes = Map.fromDistinctAscList
   , ("PT18", "PlutusTx.Enum.Ordering.toEnum: bad argument")
   , ("PT19", "PlutusTx.List.last: empty list")
   , ("PT20", "PlutusTx.Ratio.recip: reciprocal of zero")
+  , ("PT21", "PlutusTx.List.indexBuiltinList: negative index")
+  , ("PT22", "PlutusTx.List.indexBuiltinList: index too large")
```
Unisay, line 59: FYI this list is also [duplicated in docs](https://github.com/IntersectMBO/plutus/blob/27fd26250e1d45583fe95bf06b351f20357f85c1/doc/read-the-docs-site/troubleshooting.rst#L98), I suggest you update the docs too (as per instruction on line 21)

PR #5900, opened 2024-04-12
Add some more complex test cases for the remove-trace plugin option

Unisay:
- [x] Extract common code into a `Lib` module;
- [x] Test case for a non-constant trace.

### DIFF
--- a/plutus-tx-plugin/test/Plugin/NoTrace/Lib.hs
+++ b/plutus-tx-plugin/test/Plugin/NoTrace/Lib.hs
@@ -1,0 +1,64 @@
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE NoImplicitPrelude   #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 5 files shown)


plutus-tx-plugin/test/Plugin/NoTrace/Lib.hs
```diff
@@ -0,0 +1,64 @@
...
+-- Functions that contain traces -------------------------------------------------------------------
+
+traceArgument :: BuiltinString -> ()
+traceArgument x = trace x ()
+
+traceShow :: ()
+traceShow =
+  let f :: (Show s) => s -> ()
+      f s = trace (show s) ()
+   in noinline f MkArg
```
zliu41, line 45: What's the purpose of `noinline`?

zliu41 (approved):

plutus-tx-plugin/test/Plugin/NoTrace/Lib.hs
Unisay: My idea was to avoid GHC inlining `f` as much as possible in order to have an example of a function with constraint.

PR #5905, opened 2024-04-15
Update some more benchmarks to use correct evaluation context (PLT-9713) [fourth attempt]

kwxm:
This updates the BLS12-381 and Marlowe benchmarks to use the more precise evaluation strategy introduced in https://github.com/IntersectMBO/plutus/pull/5867.

Trying again without all of the benchmarking noise.


kwxm:
This seems to work OK now, so I'm going to merge it.  See issue #5906 though.

PR #5910, opened 2024-04-17
Add Delay type and use it

### LINKED ISSUE #5908: Add a `Lazy` type that compiles to delay/force

See this comment from Roman: https://github.com/IntersectMBO/plutus/pull/5901/files#r1563932610

_No response_

michaelpj:
Fixes #5908.

Unsure where to put this, it could go in its own module I guess.


plutus-tx/src/PlutusTx/Utils.hs
```diff
@@ -1,10 +1,21 @@
...
   error $
     "The " <> show placeholder <> " placeholder must have been replaced by the \
       \core-to-plc plugin during compilation."
+
+-- | Delay evalaution of the expression inside the 'Delay' constructor.
+newtype Delay a = Delay (forall b. a)
+
+-- | Force the evaluation of the expression delayed by the 'Delay'.
+force :: Delay a -> a
+force (Delay a) = a @BI.BuiltinUnit
```
zliu41, line 21: I guess the `BuiltinUnit` is necessary in order for it to be compiled into `delay` and `force`?

zliu41 (approved):

plutus-tx/src/PlutusTx/Utils.hs
michaelpj: Mysteriously it compiles without it, but I think it's better to be explicit about doing the type instantiation.

plutus-tx/src/PlutusTx/Utils.hs
```diff
@@ -1,10 +1,21 @@
...
 import Prelude as Haskell
 
 mustBeReplaced :: String -> a
 mustBeReplaced placeholder =
   error $
     "The " <> show placeholder <> " placeholder must have been replaced by the \
       \core-to-plc plugin during compilation."
+
+-- | Delay evalaution of the expression inside the 'Delay' constructor.
+newtype Delay a = Delay (forall b. a)
```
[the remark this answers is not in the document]
michaelpj: Sure.

michaelpj:
The problem with this is that the Haskell code you write using it is _not_ delayed when used _off-chain_ (and remember that we are using `Strict` in many places). So if you e.g. try to decode some `Data` object off-chain, you will end up forcing the error cases early. Annoying.

I'm not entirely sure how to fix this. If we actually want something that behaves differently on- and off-chain, then we probably need to bake it into the compiler.

michaelpj:
Yes, but that's the unit-lambda approach which is what we were trying to avoid here. That said, I think this would still be a handy library tool even if all it does is wrap up that pattern.

Unisay:
> That said, I think this would still be a handy library tool even if all it does is wrap up that pattern.

This is exactly how I was thinking about it. Same tool but more explicitly tagged.

PR #5911, opened 2024-04-18
Kwxm/plc spec/bitwise conversions

kwxm:
[PLT-8189]  This adds the two new integer/bytestring conversions to the specification. See p35-36 of the attached PDF for a readable version.
 
[plutus-core-specification.pdf](https://github.com/IntersectMBO/plutus/files/15024318/plutus-core-specification.pdf)

### DIFF
--- a/doc/plutus-core-spec/cardano/builtins4.tex
+++ b/doc/plutus-core-spec/cardano/builtins4.tex
@@ -69,8 +69,8 @@ The \texttt{integerToByteString} function converts non-negative integers to byte
 It takes three arguments:
 \begin{itemize}
 \item An endianness flag $e$.
-\item A width argument $w$.
-\item The integer $n$ to be converted.
+\item A width argument $w$ with $0 \leq w < 8192$.
+\item The integer $n$ to be converted: it is required that $0 \leq n < 256^{8192} = 2^{65536}$.
 \end{itemize}
 
 \noindent If $e$ is \texttt{True} then the conversion is big-endian, otherwise it is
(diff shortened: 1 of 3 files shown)


ramsay-t:
Are the subscripts LE and BE spelt out somewhere?

ramsay-t (approved):

doc/plutus-core-spec/cardano/builtins4.tex
```diff
@@ -37,13 +43,111 @@ \subsubsection{Miscellaneous built-in functions}
...
+\begin{cases}
+  \errorX & \text{if $n<0$ or $n \geq 2^{65536}$}\\
+  \errorX & \text{if $w<0$ or $w > 8192$}\\
+  [] & \text{if $n=0$ and $0 \leq w \leq 8192$}\\
+\end{cases}
+$$
+
+\noindent  Now assume that none of the conditions above hold, so $0 < n < 2^{65536}$ and
+$0 \leq w \leq 8192$.  Since $n>0$ we can write
+$n = \sum_{i=0}^{N-1}a_{i}256^i$ with $N \geq 1$ and $a_{N-1} \ne 0$.   We then  have
```
ana-pantilie, line 106: Sorry, maybe it's a stupid question, but I don't understand why we can write `n` as that sum. Can you provide some more explanations?

kwxm: I guess it's a base-256 expansion: I'll try to explain that a bit more clearly.

kwxm:
> Are the subscripts LE and BE spelt out somewhere?

I just had to introduce those because there wasn't room to get the full definition into the table.  I think I assumed it would be obvious what they stood for given the discussion of endianness, but it wouldn't do any harm to stick in a comment saying "(hence the subscripts LE and BE)" or something.

doc/plutus-core-spec/cardano/builtins4.tex
kwxm: Oh, I forgot to say that the a_i are supposed to lie between 0 and 255. I've fixed that now.

PR #5912, opened 2024-04-18
Change CMTooFewParamsError to a warning

zliu41:
This makes `mkEvaluationContext` succeed (with a warning) upon receiving too few parameters.

See https://github.com/IntersectMBO/cardano-ledger/issues/2902#issuecomment-2062528544, and also see the updated Note.


plutus-ledger-api/src/PlutusLedgerApi/Common/ParamName.hs
```diff
@@ -99,9 +100,11 @@ tagWithParamNames ledgerParams =
             tell [CMTooManyParamsWarn {cmTooManyExpected = lenExpected, cmTooManyActual = lenActual}]
             -- zip will truncate/ignore any extraneous parameter values
             pure $ zip paramNames ledgerParams
-        GT ->
+        GT -> do
+            -- Too few parameters - substitute a large number for the missing parameters
             -- See Note [Cost model parameters from the ledger's point of view]
-            throwError $ CMTooFewParamsError {cmTooFewExpected = lenExpected, cmTooFewActual = lenActual }
+            tell [CMTooFewParamsWarn {cmTooFewExpected = lenExpected, cmTooFewActual = lenActual}]
+            pure $ zip paramNames (ledgerParams ++ repeat (toInteger (maxBound :: Int64)))
```
lehins, line 107: Why don't we switch the type of `ledgerParams` to `Int64`? There is no point for CostModel parameter to be specified by an arbitrary precision `Integer`? `mkEvaluationContext` will fail for values outside of `Int64` range anyways.
```suggestion
            pure $ zip paramNames (ledgerParams ++ repeat (maxBound :: Int64))
```

lehins (approved):
Making no lower bound restriction on the number of parameters should work too.

michaelpj (commented):
I don't think we should do this until we've implemented some sensible behaviour for this case.

We should write a test on our side for what happens if we use the ledger interface this way with parameters missing for a given builtin and then try to evaluate a script that uses that builtin.

zliu41:
Yes I should add a test. By "until we've implemented some sensible behaviour", do you mean adding a test, or something else?

michaelpj:
I mean e.g. making it so that the missing parameters get set to MAX_VALUE or whatever. Something other than what currently happens, which is that the default values will leak out.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs
```diff
@@ -173,31 +186,28 @@ data CostModelApplyError =
...
 data CostModelApplyWarn =
-    CMTooManyParamsWarn { cmTooManyExpected :: !Int, cmTooManyActual :: !Int }
-    {- ^ More costmodel parameters given, than expected
-
-    See Note [Cost model parameters from the ledger's point of view]
-    -}
+      CMTooManyParamsWarn { cmTooManyExpected :: !Int, cmTooManyActual :: !Int }
+      -- ^ See Note [Cost model parameters from the ledger's point of view]
+    | CMTooFewParamsWarn { cmTooFewExpected :: !Int, cmTooFewActual :: !Int }
+      -- ^ See Note [Cost model parameters from the ledger's point of view]
```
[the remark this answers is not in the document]
zliu41: Good point on the renaming. Regarding record fields for sum types, it is handy for construction, but certainly not great for field access.

plutus-ledger-api/src/PlutusLedgerApi/Common/ParamName.hs
zliu41: > mkEvaluationContext will fail for values outside of Int64 range anyways.

How so? Where does it fail?

lehins: It'll fail here: https://github.com/IntersectMBO/plutus/blob/9d0a9cf5bffc78875c4b719d3a60aee4031a2e78/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs#L220-L222

With something like this:
```
       CMInternalWriteError "parsing Int failed, value is either floating or will cause over or underflow -1.8446744073709551615e19"
```

To be honest with you I am really surprised to see JSON and `aeson` having anything to do with Plutus evaluation context.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs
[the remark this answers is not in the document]
michaelpj: Also good for pattern matching. Ideally you just use `NoFieldSelectors` in this case, but we can't do that until we drop 8.10.

plutus-ledger-api/src/PlutusLedgerApi/Common/ParamName.hs
zliu41: Changing `ledgerParams` to `[Int64]` will also require some ledger changes, since the ledger currently uses `[Integer]` right? So this means the ledger will need to do the conversion and make sure it fails if there are overflows. Which may be fine, but I'd rather not do anything not strictly necessary right now.

effectfully: > To be honest with you I am really surprised to see JSON and `aeson` having anything to do with Plutus evaluation context.

Jesus Christ, what a horrible mess, yes please, let's make `mkEvaluationContext` accept `[Int64]`.

zliu41: See https://github.com/IntersectMBO/plutus/pull/5920

effectfully (commented):
LGTM

PR #5918, opened 2024-04-22
[Exe] Generalize 'withL' to 'withLangGeneral'

effectfully:
This adds a function encompassing all of `withFlatL`, `withShowL` and `withPrettyPlcL`. I'm not even sure that it's worth it at this point, but given that @bezirg asked for it and I liked the challenge, here we go.

### DIFF
--- a/plutus-core/executables/plutus/AnyProgram/With.hs
+++ b/plutus-core/executables/plutus/AnyProgram/With.hs
@@ -1,27 +1,40 @@
+{-# LANGUAGE AllowAmbiguousTypes      #-}
 {-# LANGUAGE DataKinds                #-}
 {-# LANGUAGE FlexibleInstances        #-}
 {-# LANGUAGE MultiParamTypeClasses    #-}
 {-# LANGUAGE PolyKinds                #-}
+{-# LANGUAGE QuantifiedConstraints    #-}
 {-# LANGUAGE RankNTypes               #-}
 {-# LANGUAGE StandaloneKindSignatures #-}
+{-# LANGUAGE TypeOperators            #-}
+{-# LANGUAGE UndecidableInstances     #-}
 {-# LANGUAGE UndecidableSuperClasses  #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 -- | BOILERPLATE needed to support Hasochism.
 -- See <https://homepages.inf.ed.ac.uk/slindley/papers/hasochism.pdf>
 module AnyProgram.With where
 
+import PlutusCore qualified as PLC
 import PlutusCore.Data qualified as PLC
 import PlutusCore.Pretty as PLC
+import PlutusIR qualified as PIR
 import UntypedPlutusCore qualified as UPLC
 
 import Data.Kind (Constraint)
-import Flat
 import Types
 
 -- for: (typeclass `compose` type)
 type ComposeC :: forall a b. (b -> Constraint) -> (a -> b) -> a -> Constraint
 class    constr (f x) => ComposeC constr f x
 instance constr (f x) => ComposeC constr f x
+
+type UnitC :: forall a. a -> Constraint
+class    UnitC x
+instance UnitC x
+
+type AndC :: forall a. (a -> Constraint) -> (a -> Constraint) -> a -> Constraint
+class    (constr1 x, constr2 x) => AndC constr1 constr2 x
+instance (constr1 x, constr2 x) => AndC constr1 constr2 x
 
 withN :: forall constr s r
       . ( constr (FromName 'Name)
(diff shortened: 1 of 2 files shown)


plutus-core/executables/plutus/AnyProgram/With.hs
```diff
@@ -52,34 +65,71 @@ withA s r = case s of
...
-    SPir sname sann -> withN @Show sname $ withNT @Show sname $ withA @Show sann r
-    SPlc sname sann -> withN @Show sname $ withNT @Show sname $ withA @Show sann r
-    SUplc sname sann -> withN @Show sname $ withN @(ComposeC Show UPLC.Binder) sname $
-                       withA @Show sann r
-    SData -> r
+withLang
+    :: forall constr s r.
+        ( constr (FromNameTy 'Name)
+        , constr (FromNameTy 'DeBruijn)
+        , constr (FromNameTy 'NamedDeBruijn)
```
effectfully, line 108: We could give these three constraints a single name and reuse it here, in `withLangGeneral` and in `withNT`, plus do the same with `FromName`, `FromAnn` etc constraints, but I'm not sure if it buys us much and we can always do that later.

plutus-core/executables/plutus/AnyProgram/With.hs
```diff
@@ -1,28 +1,41 @@
...
 
+type UnitC :: forall a. a -> Constraint
+class    UnitC x
+instance UnitC x
+
+type AndC :: forall a. (a -> Constraint) -> (a -> Constraint) -> a -> Constraint
+class    (constr1 x, constr2 x) => AndC constr1 constr2 x
+instance (constr1 x, constr2 x) => AndC constr1 constr2 x
+
 withN :: forall constr s r
```
effectfully, line 39: I hate these names BTW, `withName` wouldn't be much longer and it would be much more readable.

zliu41 (commented):
I'm not sure what this buys us? Now we have more lines of code, and it seems all it does is to allow us to write `withFlatL` instead of `withLang @Flat`?

effectfully:
> Now we have more lines of code

Yeah, but that's because one set of constraints is written down twice, we could `type`-synonym it and there would be less code I think. 

> all it does is to allow us to write `withFlatL` instead of `withLang @Flat`?

Yes, but maybe we'll need other constraints: `Show`, `Serialise`, `NFData` etc. With this machinery it's much easier to get those than duplicating the whole definition every time.

Anyways, I don't really mind this getting closed as "won't do".

zliu41 (approved):
> Yes, but maybe we'll need other constraints: Show, Serialise, NFData etc. With this machinery it's much easier to get those than duplicating the whole definition every time.

Ok, sounds reasonable.

plutus-core/executables/plutus/AnyProgram/With.hs
```diff
@@ -52,34 +65,71 @@ withA s r = case s of
...
-                      withA @Pretty sann r
-    SPlc sname sann -> withN @PrettyClassic sname $ withN @PrettyReadable sname $
-                      withNT @PrettyClassic sname $ withNT @PrettyReadable sname $
-                      withA @Pretty sann r
-    SUplc sname sann -> withN @PrettyClassic sname $ withN @PrettyReadable sname $
-                       withA @Pretty sann r
-    SData -> r
+withPrettyPlcL :: forall s r. SLang s -> (PrettyPlc (FromLang s) => r) -> r
+withPrettyPlcL = withLangGeneral
+    @(PrettyClassic `AndC` PrettyReadable)
```
bezirg, line 129: @effectfully : Just out of curiosity: is `AndC` strictly necessary here?

Could it alternatively be done by decomposing to two `withLangGeneral` calls? Something like:

```
withPrettyPlcL = withLangGeneral  @PrettyClassic ....   $ withLangGeneral @PrettyReadable ...
```
?

bezirg (approved):

bezirg:
> I love it , I am going to rebase and rename some stuff and push here again. Thank you so much @effectfully <3

Actually I am going to merge it as it is and do the renaming in another PR

plutus-core/executables/plutus/AnyProgram/With.hs
effectfully: > is `AndC` strictly necessary here?

I believe so, you have to satisfy a qualified constraint `(a, b, c) => d` and you can't do that unless `a`, `b` and `c` are indeed as expected by `d` and so you have to provide the right `a`, `b` and `c` which forces you to use `AndC`. In particular, `PrettyPlc` can only hold if both `PrettyClassic` and `PrettyReadable` hold, so you can't just split the constraint for individual parts. I think it might be possible in theory to make it so that `PrettyPlc` is implied by some product of constraints and then satisfy those individually, but that's pointless, writing `AndC` is simply easier.

PR #5920, opened 2024-04-23
Use Int64 in CostModelParams

zliu41:
See [this thread](https://github.com/IntersectMBO/plutus/pull/5912#discussion_r1571488080).


effectfully (approved):
Much better!

kwxm (commented):
This is a great improvement.  I was confused for a little while beacuse we use `Int64` in the cost model parameters but the actual cost model ultimately contains `SatInt`s disguised as `CostingInt`s and I couldn't see a conversion anywhere, but now I think that's secretly happening in the horrible [applyParameters](https://github.com/IntersectMBO/plutus/blob/b97e8be50470065f37948f62dcaa7d67f71b9a39/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs#L217) code where everything gets converted to JSON and back and all of our integral types get converted to/from [Scientific](https://hackage.haskell.org/package/scientific-0.3.7.0/docs/Data-Scientific.html).  We really should get rid of the Aeson translation and make it explicit what's going on.

PR #5923, opened 2024-04-24
CaseOfCase kind mismatch error fix

### LINKED ISSUE #5922: CaseOfCase kind mismatch error
```
cabal test plutus-ir-test --test-options "-p /caseOfCase/ --quickcheck-replay=753987"
```

```
          *** Failed! (after 685 tests and 6 shrinks):
          Error during PIR typechecking:
          Kind mismatch at ()
          Expected a type of kind
            '(fun (type) (type))'
          But found one of kind
            '(type)'
          Namely,
            '(con unit)'
          ty,tm = ( unit
          , let
            data (d12_12 :: (* -> *) -> * -> *) (a3_3 :: * -> *) a10_10 | m11_11 where
              c6_6 : unit -> d12_12 a3_3 a10_10
          in
          m11_11
            {list}
            {unit}
            (m11_11
               {list}
               {unit}
               (error {d12_12 list unit})
               {d12_12 list unit}
               (\(x23_23 : unit) -> error {d12_12 list unit}))
            {unit}
            (error {unit -> unit}) )
```

Unisay:
Closes #5922 by fixing order of type arguments as they are extracted.

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
@@ -8,6 +8,7 @@
 module PlutusIR.Contexts where
 
 import Control.Lens
+import Data.DList qualified as DList
 import Data.Functor (void)
 import PlutusCore.Arity
 import PlutusCore.Name.Unique qualified as PLC
@@ -132,13 +133,16 @@ data SplitMatchContext tyname name uni fun a = SplitMatchContext
   , smBranches  :: AppContext tyname name uni fun a
   }
 
+-- | Extract the type application arguments from an 'AppContext'.
+-- Returns 'Nothing' if the context contains a TermAppContext.
+-- See 'test_extractTyArgs'
 extractTyArgs :: AppContext tyname name uni fun a -> Maybe [Type tyname uni a]
-extractTyArgs = go id
+extractTyArgs = go DList.empty
   where
     go acc = \case
-      TypeAppContext ty _ ctx -> go (acc . (ty :)) ctx
-      TermAppContext{}        -> Nothing
-      AppContextEnd           -> Just (acc [])
+      TypeAppContext ty _ann ctx -> go (DList.snoc acc ty) ctx
+      TermAppContext{}           -> Nothing
+      AppContextEnd              -> Just (DList.toList acc)
 
 -- | Split a normal datatype 'match'.
 splitNormalDatatypeMatch
--- a/plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
@@ -133,10 +133,12 @@ data SplitMatchContext tyname name uni fun a = SplitMatchContext
   }
 
 extractTyArgs :: AppContext tyname name uni fun a -> Maybe [Type tyname uni a]
-extractTyArgs = go []
-  where go acc (TypeAppContext ty _ ctx) = go (ty:acc) ctx
-        go _ (TermAppContext{})          = Nothing
-        go acc AppContextEnd             = Just acc
+extractTyArgs = go id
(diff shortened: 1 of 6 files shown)


effectfully (changes requested):
That's not nearly enough:

- most importantly, this function is in some general module, how can you be sure you're not breaking something else by reversing the order of the result? There's no comment on what that function is supposed to return, no examples etc, maybe that's how it's supposed to work for some other use case that isn't case-of-case. I don't see any way around going through all transitive dependencies and ensuring you aren't breaking things elsewhere
- please document the offending function, so that it's clear what it returns to reduce the possibility of us screwing things up the same way again
- please add the test that caught the bug as a unit/golden test to make sure we never regress back

plutus-core/plutus-ir/src/PlutusIR/Contexts.hs
```diff
@@ -133,10 +133,12 @@ data SplitMatchContext tyname name uni fun a = SplitMatchContext
   }
 
 extractTyArgs :: AppContext tyname name uni fun a -> Maybe [Type tyname uni a]
-extractTyArgs = go []
-  where go acc (TypeAppContext ty _ ctx) = go (ty:acc) ctx
-        go _ (TermAppContext{})          = Nothing
-        go acc AppContextEnd             = Just acc
+extractTyArgs = go id
```
michaelpj, line 136: I think I'd prefer to use `DList` explicitly, this is not very obvious to people who aren't familiar with it. Or just reverse at the end. Either works.

Unisay: Looks like [I am not the first one](https://github.com/IntersectMBO/plutus/blob/b97e8be50470065f37948f62dcaa7d67f71b9a39/plutus-core/plutus-ir/src/PlutusIR/Contexts.hs#L73) to use this idiom ;-)

michaelpj: Yep, that person was wrong ;)

effectfully: I think all options are fine here, including the currently written one.

Unisay:
> most importantly, this function is in some general module, how can you be sure you're not breaking something else by reversing the order of the result? 
> I don't see any way around going through all transitive dependencies and ensuring you aren't breaking things elsewhere

Besides `CaseOfCase` transform I found only one place where this function `extractTyArgs` is used - in the `PlutusIR.Analysis.Builtins.defaultUniMatcherLike` and that one is only used in the `CaseOfCase`.

The tests for `CaseOfCase` transformation are all green so this gives me the confidence.

> There's no comment on what that function is supposed to return, no examples etc, maybe that's how it's supposed to work for some other use case that isn't case-of-case. 

I've added comment and test cases to explain the semantics as I understood them. I'd be grateful to @michaelpj  for checking those case cases making sure that I derived the intended behavior correctly.

> please document the offending function, so that it's clear what it returns to reduce the possibility of us screwing things up the same way again

Done.
 
> please add the test that caught the bug as a unit/golden test to make sure we never regress back

It was the property test that caught the bug, and its already there. I think that unit tests for the `extractTyApp` are enough to make sure we never have this bug again.

michaelpj:
So the analysis is that this would only happen when we had case-of-case on a type that had _two_ type parameters, is that right? That does indeed seem like something we might not have tripped up on until now. Hooray for the generators!

michaelpj (approved):

effectfully (approved):
As we discussed on the call I'd prefer to record the offender to increase coverage, but I'm not gonna be an ass about it, so here's your approval.

Unisay:
> As we discussed on the call I'd prefer to record the offender to increase coverage, but I'm not gonna be an ass about it, so here's your approval.

I have recorded the offender as a golden test ;-)

PR #5926, opened 2024-04-26
[Builtins] Replace 'EvaluationResult' with 'BuiltinResult'

effectfully:
This replaces several `Emitter (EvaluationResult a)` occurrences with `BuiltinResult`, something that I missed the last [time](https://github.com/IntersectMBO/plutus/pull/5728).

In addition to that, it also replaces `EvaluationResult` with `BuiltinResult` in general. It doesn't matter performance-wise (modulo a regression that we didn't notice some time ago), but `BuiltinResult`, unlike `EvaluationResult`, allows one to attach an error message to a failure, which we do in this PR as well, meaning we now get better error messages. And we also now respect the operational vs structural evaluation errors distinction.

The PR also replaces `Emitter` with `BuiltinResult`.


plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1407,85 +1424,93 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
+        let headListDenotation :: SomeConstant uni [a] -> BuiltinResult (Opaque val a)
             headListDenotation (SomeConstant (Some (ValueOf uniListA xs))) = do
-                DefaultUniList uniA <- pure uniListA
-                x : _ <- pure xs
-                pure $ fromValueOf uniA x
+                case uniListA of
+                    DefaultUniList uniA -> case xs of
+                        []    -> fail "Expected a non-empty list but got an empty one"
+                        x : _ -> pure $ fromValueOf uniA x
+                    _ -> throwing _StructuralUnliftingError "Expected a list but got something else"
```
effectfully, line 1488: It's awkward that the structural vs operational errors distinction leaks to the user space here, but that's inevitable, since "runtime type errors" are inherently structural and errors like "expected a non-empty list but got an empty one" are inherently operational. 

TODO: this needs documentation.

plutus-core/cost-model/budgeting-bench/Benchmarks/Nops.hs
```diff
@@ -132,12 +133,12 @@ nopCostParameters =
 infixr >:
 (>:) :: uni ~ DefaultUni
      => SomeConstant uni Integer
-     -> EvaluationResult Integer
-     -> EvaluationResult Integer
+     -> BuiltinResult Integer
+     -> BuiltinResult Integer
```
effectfully, line 137: We now recommend using`BuiltinResult` instead of `EvaluationResult` for all builtins. The latter doesn't allow for attaching any error messages to errors, so it simply shouldn't be used (it's now a type error to use `EvaluationResult` in a builtin).

plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
```diff
@@ -181,6 +188,10 @@ throwNotAConstant :: MonadError BuiltinError m => m void
 throwNotAConstant = throwing _StructuralUnliftingError "Not a constant"
 {-# INLINE throwNotAConstant #-}
```
kwxm, line 189: Does inlining things like this make any difference to general performance?  Presumably this is something that's only going to happen once, so is optimising it helpful?  I may well be missing something important though.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Error.hs
```diff
@@ -33,8 +33,8 @@ import Data.Bitraversable
    wrong. For example, 'Error' was executed, 'tailList' was applied to an empty list or evaluation
```
kwxm, line 33: `Error` -> `error`?

kwxm (approved):
I don't see anything to complain about.  I found a [comment](https://github.com/IntersectMBO/plutus/blob/bc8c3a765769d2c0cd41c43278f5954cfdfd9b15/plutus-core/plutus-core/src/PlutusCore/Crypto/Utils.hs#L15) that refers to `EvaluationResult` but should probably refer to `BuiltinResult` instead now.  I'm not sure if the rest of the comment's still accurate.

plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
effectfully: > Does inlining things like this make any difference to general performance?

Yes, it can make it worse :upside_down_face: 

I'm doing it, because

1. I read GHC Core for builtins a lot and I want everything in there look as bare-bones as possible, otherwise I start thinking "should this be optimized?" and get distracted. Without the pragma the function would likely still be inlined, but if it wasn't, we'd get a constraint dictionary in the middle of builtins and that's one thing that I don't want to see in there
2. it's much less likely that somebody forgets an `INLINE` pragma in an important place if we simply have them everywhere as that is an easier rule to follow than have `INLINE` here and `NOINLINE` here
3. if inlining this definition hurts performance, then it's a bug elsewhere where we don't have an appropriate `oneShot` call or something and that bug then can be triggered by other code, so exposing it's here with simple error throwing is kinda a feature

I'll turn this into a Note, thanks for the question.

effectfully: Turned out to be quite a rabbit hole. I've added this Note:

```
{- Note [INLINE and OPAQUE on error-related definitions]
We mark error-related definitions such as prisms like '_StructuralUnliftingError' and regular
functions like 'throwNotAConstant' with @INLINE@, because this produces significantly less cluttered
GHC Core. Not doing so results in 20+% larger Core for builtins.

However in a few specific cases we use @OPAQUE@ instead to get tighter Core. @OPAQUE@ is the same as
@NOINLINE@ except the former _actually_ prevents GHC from inlining the definition unlike the latter.
See this for details: https://github.com/ghc-proposals/ghc-proposals/blob/5577fd008924de8d89cfa9855fa454512e7dcc75/proposals/0415-opaque-pragma.rst

It's hard to predict where @OPAQUE@ instead of @INLINE@ will help to make GHC Core tidier, so it's
mostly just looking into the Core and seeing where there's obvious duplication that can be removed.
Such cases tend to be functions returning a value of a concrete error type (as opposed to a type
variable).
-}
```

If we use `NOINLINE` instead of `INLINE` in those places, then we'll get ~22k lines of Core instead of ~18k. But using `INLINE` and just a couple of well-placed `OPAQUE` calls we get the number of lines down to less than 15k.

I'll also bring that `OPAQUE` thing to the team w.r.t. Plutus Tx, since it's way more reliable than `NOINLINE` that we user currently.

effectfully:
> I don't see anything to complain about.  I found a [comment](https://github.com/IntersectMBO/plutus/blob/bc8c3a765769d2c0cd41c43278f5954cfdfd9b15/plutus-core/plutus-core/src/PlutusCore/Crypto/Utils.hs#L15) that refers to `EvaluationResult` but should probably refer to `BuiltinResult` instead now.  I'm not sure if the rest of the comment's still accurate.

I did my best to try to update all the existing comments, but I did miss that one. Thank you! Fixed now.

effectfully:
I'm merging this bad boi and if any more reviews come in I'll address them separately.

PR #5930, opened 2024-04-29
Fix some Notes so they pass the check

kwxm:
I ran the `scripts/check-Notes.sh` script and found a few references that said "note" instead of "Note" (see the rules [here](https://github.com/IntersectMBO/plutus/blob/0567ef7fc1efb4ec415fe882df9b313d8c3dd0df/scripts/check-Notes.awk#L3), which are supposed to make it easier to find Notes and references to them).  This fixes that.

(Also, see `find-Notes.sh` for a script to list/search the Notes.)

### DIFF
--- a/scripts/check-Notes.awk
+++ b/scripts/check-Notes.awk
@@ -120,41 +120,41 @@ maybeSplitReference && /^[- \t-]*\[/ {
 # Check for plural references
 /(NOTES|Notes|notes) *\[/ {
     printf ("Invalid note format (no plurals allowed) at %s:%d\n ", FILENAME, FNR)
-    printf ("> %s\n", $0)
+    printf ("> %s\n\n", $0)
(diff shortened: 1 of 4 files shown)


scripts/check-Notes.awk
kwxm, line 1: I added some extra newlines to make the output more readable.

Unisay (approved):
Ok, I didn't know that "note" word should be capitalized!

kwxm:
> Ok, I didn't know that "note" word should be capitalized!

We'd been a bit inconsistent about how we did stuff with notes and it was getting difficult to find things (plus there were references to things that didn't exist any more or had slightly different titles), so I made up some rules a few months ago and wrote a script to check them.  We were talking about adding the script to the CI checks but never got round to it. Maybe I'll being that up again.

PR #5931, opened 2024-04-29
[Experiment] Inline 'spendAccumulatedBudget'

effectfully:
Just an experiment to see how much worse I can make the CEK machine.


effectfully:
A 17% slowdown as a result of adding a single `INLINE` pragma, not bad!

PR #5934, opened 2024-04-29
PlutusV3: Datum/Redeemer in ScriptContext, Datum Optional, Unified Script Type

zliu41:
- The `ScriptContext` type for Plutus V3 is modified to contain redeemer and an optional
  datum for spending scripts, in the `scriptContextScriptInfo` field. As a result, all
  Plutus V3 scripts, regardless of the purpose, take a single argument: `ScriptContext`.
- Datum is now optional for Plutus V3 spending scripts.

See https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2078414218


lehins (changes requested):
These changes will require significant amount of work in Ledger. We do not have time to implement it for Conway and PlutusV3 without postponing the deadline. As was discussed on CIP-69

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -398,6 +400,42 @@ instance PlutusTx.Eq ScriptPurpose where
     a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
   _ == _ = Haskell.False
 
+-- | Like `ScriptPurpose` but also contains redeemers, and an optional datum
+-- for spending scripts.
+data ScriptInfo
+  = MintingScript V2.CurrencySymbol V2.Redeemer
```
michaelpj, line 406: It wasn't clear to me whether we were sure we wanted the redeemer in every constructor like this or as a separate product field.

michaelpj (approved):
Makes accessing the script context more expensive, but that's to be expected I guess. Obviously we need to decide modulo Alexey's comments.

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
lehins: @zliu41 Sorry to bother you, but I am working on integrating this into ledger and I remember we both agreed that it'll be a good idea to move `Redeemer` into `ScriptContext` instead of putting into every constructor of `ScriptInfo`

Whenever you get a chance could you make this change, if we are on the same page about this that is?

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -398,6 +400,42 @@ instance PlutusTx.Eq ScriptPurpose where
     a PlutusTx.== a' PlutusTx.&& b PlutusTx.== b'
   _ == _ = Haskell.False
 
+-- | Like `ScriptPurpose` but also contains redeemers, and an optional datum
+-- for spending scripts.
+data ScriptInfo
```
lehins, line 405: Just a suggestion about an alternative name. `Info` isn't very informative IMHO. Feel free to ignore this suggestion if you like the `ScriptInfo` name better.
```suggestion
data ScriptPurposeArgument
```

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
zliu41: Yes I'm going to do that

plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
zliu41: There have been quite a few similar names mentioned :smile: I'll think about it

PR #5935, opened 2024-04-30
Add 'Note [Table of all possible ledger's states w.r.t. cost model parameters update]'

effectfully:
As promised.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs
@@ -156,6 +156,66 @@ of these issues and the rationale for adopting the system described above.
 
 -}
 
+{- Note [Table of all possible ledger's states w.r.t. cost model parameters update]
+What Note [Cost model parameters from the ledger's point of view] says can be summarized in a table.
+We'll need to enumerate all possible combinations of
+
+    old_node / new_node
+    before_HF / after_HF
+    shorter_list / exact_list / longer_list
+
+where the last ones are about the list of cost model parameters supplied by the ledger, e.g.
+@shorter_list@ means that the ledger supplied a list of cost model parameters that is shorter than
+expected by the appropriate Plutus ledger language at the specific protocol version.
+
+Note that the exact same list may be @exact_list@ for @old_node@ and @shorter_list@ for @new_node@.
+And the exact same list may be @longer_list@ for @old_node@ and @exact_list@ for @new_node@.
+
+We'll also need to consider changing the cost model parameters of old builtins and adding cost model
+parameters for new builtins separately, because there are non-trivial differences. For old builtins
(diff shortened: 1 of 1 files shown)


plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs
```diff
@@ -156,6 +156,60 @@ of these issues and the rationale for adopting the system described above.
...
+worry about.
+
+We'll also only consider the case when the Plutus ledger language stays the same, because if it
+changes, then everything becomes trivial: we don't need to worry about any backward compatibility
+and can simply "start over" for a new ledger language.
+
+The table then looks like this:
+
+                                         old_builtin new_builtin
+    old_node / before_HF / shorter_list    OK [1]      OK [1]
```
zliu41, line 186: It would be helpful to define what "OK" means, because it's not obvious.

"OK" means: given the exact same list and the same protocol version (either before_HF or after_HF), old_node and new_node must have the exact same behavior: a script attempting to use the old/new builtin should both succeed, or both fail at the same phase (phase 1 or phase 2).

It may be helpful to further point out the exact same list may be `exact_list` for `old_node` and `shorter_list` for `new_node`, or `longer_list` for `old_node` and `exact_list` for `new_node`.

zliu41 (approved):

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostModelInterface.hs
effectfully: > It would be helpful to define what "OK" means, because it's not obvious.

It literally means "we handle this situation correctly", where "correctly" is defined by `Note [Cost model parameters from the ledger's point of view]`, I'm just summarizing that Note in a table. It's a great Note with clear explanations and examples, such as

> During PV9, both node-9.x and node-10.x must operate normally and agree on everything. This means
node-9.x must allow receiving more cost model parameters than it expects (since it may receive
120), and node-10.x must allow receiving fewer than it expects (since it may receive 100).
Node-10.x should fill in the missing parameters with a large enough number to prevent the new
builtins from being used, in case the hard fork to PV10 happens without first updating the number
of PlutusV3 cost model parameters to 120 (which is unlikely to happen, but just in case).

> "OK" means: given the exact same list and the same protocol version (either before_HF or after_HF), old_node and new_node must have the exact same behavior: a script attempting to use the old/new builtin should both succeed, or both fail at the same phase (phase 1 or phase 2).

No, it's not what OK means, because we consider the cases of old and new builtins separately in the table. OK there means "this follows the intended behavior as described by that other Note".

Well, I guess this discussion does prove it's not obvious. I'll add a comment.

> It may be helpful to further point out the exact same list may be `exact_list` for `old_node` and `shorter_list` for `new_node`, or `longer_list` for `old_node` and `exact_list` for `new_node`.

Right, good point, thank you, will do.

PR #5939, opened 2024-04-30
[StdLib] Strictify 'fix'

### LINKED ISSUE #5961: A recursive binding is compiled unexpectedly lazily
In the following PIR:

```
(program 1.1.0
(let
  (rec)
  (termbind
    (strict)
    (vardecl y (fun (con integer) (con integer)))
    [
      [
        { (builtin trace) (fun (con integer) (con integer)) }
        (con string "hello")
      ]
      (lam z (con integer) z)
    ]
  )
    (con integer 1)
)
)
```

the trace is not shown when executed as: `cabal run pir -- compile --dont-optimise --if textual -i $FILE| cabal run uplc -- evaluate --trace-mode Logs`.

effectfully:
This makes `fix` stricter so that the body of the function that it takes always gets forced regardless of whether the necessary argument is provided or not.

Resolves #5961.


plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
```diff
@@ -0,0 +1,117 @@
...
+parsePirProgram
+  :: FilePath
+  -> IO (PIR.Program PIR.TyName PIR.Name DefaultUni DefaultFun (Provenance ()))
+parsePirProgram file = do
+  res <- runExceptT @(PCE.Error DefaultUni DefaultFun ()) $ runQuoteT do
+    contents <- liftIO $ Text.readFile file
+    PC.parseGen pTermAsProg contents
+  case res of
+    Left err -> fail $ show err
+    Right x  -> pure $ Original () <$ x
```
effectfully, line 79: Why make every annotation `Original ()` here?

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
```diff
@@ -0,0 +1,117 @@
...
+makeCompilationCtx
+  :: ( Default (CostingPart uni fun)
+     , Default (BuiltinsInfo uni fun)
+     , Default (RewriteRules uni fun)
+     )
+  => PLC.TypeCheckConfig uni fun
+  -> PIR.CompilationCtx uni fun a
+makeCompilationCtx pirTcConfig =
+  toDefaultCompilationCtx pirTcConfig
+    & set (ccOpts . coPreserveLogging) True
```
effectfully, line 56: Seems oddly specific to deserve a separate definition? I'd just inline it at the call site.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
```diff
@@ -0,0 +1,117 @@
...
+  :: ReaderT
+      (PIR.CompilationCtx DefaultUni DefaultFun ())
+      (QuoteT (ExceptT (PIR.Error DefaultUni DefaultFun (Provenance ())) Identity))
+      a
+  -> a
+runCompilationM m =
+  unsafeFromRight @(PIR.Error DefaultUni DefaultFun (Provenance ())) . runExcept $
+    runQuoteT do
+      pirTcConfig <- PLC.getDefTypeCheckConfig noProvenance
+      runReaderT m $ makeCompilationCtx pirTcConfig
```
effectfully, line 68: And by extension this is oddly specific as well.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
```diff
@@ -0,0 +1,117 @@
...
+
+parsePirProgram
+  :: FilePath
+  -> IO (PIR.Program PIR.TyName PIR.Name DefaultUni DefaultFun (Provenance ()))
+parsePirProgram file = do
+  res <- runExceptT @(PCE.Error DefaultUni DefaultFun ()) $ runQuoteT do
+    contents <- liftIO $ Text.readFile file
+    PC.parseGen pTermAsProg contents
```
effectfully, line 76: Oddly specific as well: what if you want to read an actual program rather than a term as a program.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
```diff
@@ -0,0 +1,117 @@
...
+  -> a
+runCompilationM m =
+  unsafeFromRight @(PIR.Error DefaultUni DefaultFun (Provenance ())) . runExcept $
+    runQuoteT do
+      pirTcConfig <- PLC.getDefTypeCheckConfig noProvenance
+      runReaderT m $ makeCompilationCtx pirTcConfig
+
+parsePirProgram
+  :: FilePath
+  -> IO (PIR.Program PIR.TyName PIR.Name DefaultUni DefaultFun (Provenance ()))
```
effectfully, line 72: Given the signature it's more like `readPirProgram`, you aren't just parsing it, you're also reading it from the disk.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
```diff
@@ -0,0 +1,117 @@
...
+  case res of
+    Left err -> fail $ show err
+    Right x  -> pure $ Original () <$ x
+
+evaluatePirFromFile
+  :: (MonadIO m, MonadFail m)
+  => FilePath
+  -> m (EvaluationResult (UPLC.Term Name DefaultUni DefaultFun ()), [Text])
+evaluatePirFromFile fp = do
+  program <- liftIO $ parsePirProgram fp
```
effectfully, line 86: Oddly specific by extension as well.

effectfully (commented):
Looks good to me. It's oddly specific, but it's for testing a specific pass anyway, so it's perfectly fine. It would be great to try to abstract some of that stuff out, so that we can reuse the code in other tests, but I don't really have any suggestions on that. 

30 lines of imports is ridiculous, we should clean our APIs up.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/strictLetRec.golden
```diff
@@ -0,0 +1,329 @@
+(let
```
zliu41, line 1: Can we print out UPLC, instead of TPLC? The TPLC is 100x harder to read, and for no benefit.

plutus-core/plutus-core/stdlib/PlutusCore/StdLib/Data/Function.hs
```diff
@@ -107,7 +107,8 @@ unroll = runQuote $ do
 -- | 'fix' as a PLC term.
 --
 -- > /\(a b :: *) -> \(f : (a -> b) -> a -> b) ->
--- >    unroll {a -> b} (iwrap selfF (a -> b) \(s : self (a -> b)) \(x : a) -> f (unroll {a -> b} s) x)
+-- >     unroll {a -> b} (iwrap selfF (a -> b) \(s : self (a -> b)) ->
+-- >         f (\(x : a) -> unroll {a -> b} s x))
```
zliu41, line 111: Is `f (\(x : a) -> unroll {a -> b} s x))` the same as `f (unroll {a -> b} s))`?

Hmm I wonder why we defined `fix` that way before. Isn't this the more obvious definition? This is exactly the z combinator.

plutus-tx-plugin/test/Budget/9.6/andExpensive.uplc.golden
```diff
@@ -2,19 +2,15 @@ program
...
-          (s s))
+     (\s ds ->
+        force
+          (case
+             ds
+             [ (delay (constr 0 []))
+             , (\x xs ->
+                  delay
+                    (force
+                       (case x [(delay (s s xs)), (delay (constr 1 []))]))) ]))
```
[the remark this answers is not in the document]
zliu41: > Why does the optimizer not do that? Perhaps 'cause it's not semantics preserving, for all the optimizer knows evaluating s s (I just realized I should've named it differently...) may be impure.

That's exactly why.

What I'm wondering is how the cost improvement can be so significant. I'd expect a very tiny improvement, and if `and` is used more than once, I'd expect a regression :confused:

plutus-core/plutus-core/stdlib/PlutusCore/StdLib/Data/Function.hs
effectfully: > Is `f (\(x : a) -> unroll {a -> b} s x))` the same as `f (unroll {a -> b} s))`?

In a lazy language, yes. In a strict language that transformations turns the Z combinator into the Y combinator and the latter diverges.

> Hmm I wonder why we defined `fix` that way before. Isn't this the more obvious definition? This is exactly the z combinator.

I don't know! It was me who picked the old version and I have no idea why. Maybe because it's lazier and I wanted to avoid doing unnecessary work or something. Maybe it was completely accidental. I just read through all our discussions at the time and we always used the new version as the default. Somehow the old one still got merged. A mystery. Quite likely just a stupid mistake on my part.

plutus-tx-plugin/test/Budget/9.6/andExpensive.uplc.golden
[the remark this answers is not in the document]
effectfully: > I'd expect a very tiny improvement

It may simply enable us to do more optimizations, I'll take a closer look. But in any case not saving something into a name pointlessly and then needing to do a lookup certainly should indeed help.

> if `and` is used more than once, I'd expect a regression 😕

No, `and` is computed at every recursive step: the `\s -> s s` thing, i.e. the driver of recursion, is outside of it.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/strictLetRec.golden
effectfully: It's not just harder to read, it's also noisy. It has this huge `fixBy` definition that is generated "just in case", which is normally removed via dead code elimination, but not in this case because we turn optimizations off.

(Note that some work in this PR is done by me and some by Yura, so it's not immediately obvious who reviewers are talking to)

plutus-core/plutus-core/stdlib/PlutusCore/StdLib/Data/Function.hs
zliu41: > In a lazy language, yes. In a strict language that transformation turns the Z combinator into the Y combinator and the latter diverges.

Right, of course!

plutus-tx-plugin/test/Budget/9.6/andExpensive.uplc.golden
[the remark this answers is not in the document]
zliu41: > It may simply enable us to do more optimizations

But this is the final UPLC, after all optimizations. I struggle to see how the cost difference can be so big - I must be missing something.

> No, and is computed at every recursive step: the \s -> s s thing, i.e. the driver of recursion, is outside of it.

What I'm saying is that if `and` occurs twice, then inlining it causes `s s` to be evaluated twice at each recursive step, which should be worse. I'm surprised that there isn't a single test case that became worse.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/strictLetRec.golden
Unisay: > Can we print out UPLC, instead of TPLC? The TPLC is 100x harder to read, and for no benefit.

The reason I wrote PIR in this case is because I was following this advice by @michaelpj 
image

This test isolates just one pass PIR -> PIR, so that we can reason how PIR compilation treats Let(rec)s without taking into the equation possible effects added by compilation to UPLC.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
Unisay: Fixed

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
Unisay: Fixed

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
Unisay: Fixed

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
Unisay: Indeed, this function is specific to one test and its used only in it. I've moved it back, closer to where the test is.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
Unisay: Fixed

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
Unisay: Fixed, used `noProvenance` instead.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/strictLetRec.golden
zliu41: Right, this is PIR, not TPLC.. but still, this is too much to look at. At the very least you should print _readable_ PIR, but it will still be too big. I think the UPLC for this program should be quite small, probably 10 lines or so.

kwxm:
I ran some of the benchmarks in [this PR](https://github.com/IntersectMBO/plutus/pull/5953).  This seems to speed things up a bit, apart from some suspicious results for the list benchmarks.  It definitely doesn't make things worse.

plutus-tx-plugin/test/Budget/9.6/constAccL.uplc.golden
```diff
@@ -14,4 +14,4 @@ program
         (fix1
            (\go acc ds ->
               force (case ds [(delay acc), (\x xs -> delay (go acc xs))]))))
-     (\f -> (\s -> s s) (\s x -> f (s s) x)))
\ No newline at end of file
+     (\f -> (\s -> s s) (\s -> f (\x -> s s x))))
```
effectfully, line 17: For anybody wondering, this is the whole change.

plutus-tx-plugin/test/Budget/9.6/findIndexCheap.uplc.golden
```diff
@@ -2,20 +2,18 @@ program
...
+        force
+          (case
+             ds
+             [ (delay (constr 1 []))
+             , (\x xs ->
+                  delay
+                    (force
+                       (force ifThenElse
+                          (lessThanEqualsInteger 10 x)
+                          (delay ((\x -> s s x) (addInteger 1 i) xs))
```
effectfully, line 15: Optimization idea: we should eta-contract applied functions.

plutus-tx-plugin/test/Budget/9.6/findIndexExpensive.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 13963400
-| mem: 39440})
\ No newline at end of file
+({cpu: 13825400
+| mem: 38840})
```
effectfully, line 2: Barely any difference for `findIndex` and a big difference for `elem`:

```diff
- ({cpu: 10254990
- | mem: 32120})
+ ({cpu: 9426990
+ | mem: 28520})
```

and `all`:

```diff
- ({cpu: 10219630
- | mem: 32120})
+ ({cpu: 9391630
+ | mem: 28520})
```

The more I look at it, the more confused I become.

plutus-tx-plugin/test/Budget/9.6/notElemExpensive.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 10254990
-| mem: 32120})
\ No newline at end of file
+({cpu: 9426990
+| mem: 28520})
```
effectfully, line 2: No extra lam-apply => big difference.

effectfully:
@zliu41 I think the by far most likely explanation is simply removal of a single lam-apply pair. Because when we get a different one back, the numbers go roughly to where they were before. The simplest way to investigate it would be to eta-contract `\x -> s s x` to `s s` in `findIndex` and observe that it does indeed affect costs that much.

plutus-tx-plugin/test/Budget/9.6/andExpensive.uplc.golden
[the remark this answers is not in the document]
effectfully: > I now wonder if we can make it even more efficient by somehow caching `s s` once and for all.

I've looked at it as hard as I could and I failed to find any opportunity there, it probably doesn't exist.

zliu41 (commented):
> @zliu41 I think the by far most likely explanation is simply removal of a single lam-apply pair.

I think you are likely correct. For `any` or `elem`, the recursion is simply called on `xs` (i.e., `go xs`). The UPLC inliner can inline `xs`, yielding `s s xs`, which is efficient. 

For `findIndex`, on the other hand, we have `go (Builtins.addInteger i 1) xs`, and the inliner is unable to inline `Builtins.addInteger i 1` (because it's impure), leading to `(\x -> s s x) (addInteger 1 i)`.

There is a big cost difference for `findIndexEmptyList`, and I think that's because `s s` is evaluated once before, but not evaluated at all after. In general `s s` is evaluated one more time before vs. after.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests.hs
```diff
@@ -0,0 +1,53 @@
...
+          "strictLetRec"
+    , pure $ testCase "traces" do
+        (result, traces) <- do
+          pirTerm <- pirTermFromFile (joinPath path </> "StrictLetRec" </> "strictLetRec")
+          evalPirProgramWithTracesOrFail (pirTermAsProgram (void pirTerm))
+        case result of
+          EvaluationFailure ->
+            fail $ "Evaluation failed, available traces: " <> show traces
+          EvaluationSuccess term -> do
+            term @?= constant () (someValue (1 :: Integer))
```
effectfully, line 51: @Unisay just FYI, you can write `fromValue (1 :: Integer)` instead of `constant () (someValue (1 :: Integer)`. It's a bit confusing, because we have a `HasConstant` class that contains

```haskell
    -- | Wrap a Haskell value as a @term@.
    fromConstant :: Some (ValueOf (UniOf term)) -> term
```

which is a lot like `TermLike` containing

```haskell
    constant :: ann -> Some (ValueOf uni) -> term ann
```

but doesn't take an annotation. `fromValue` is defined in terms of the former and we use it during evaluation unlike anything from `TermLike`. And in general, evaluation stuff is separate (because it needs to be ultra-optimized), so it doesn't have great discoverability.

effectfully:
@Unisay has addressed comments, I've written a Note explaining what we compile (mutually) recursive functions to, let's merge this bad boi. There's still work to do here, I'll create some issues or something.

zliu41:
I still think `stirctLetRec.golden` should be UPLC. I don't think anyone is going to read it otherwise.

effectfully:
@zliu41 I think nobody is going to read it regardless. Without optimizations we produce this unreadable CPSed code and it's not much better in UPLC compared to TPLC/PIR, types even help reading it a bit. But if you have a strong opinion about it, I'll make it UPLC. Note that [this](https://github.com/IntersectMBO/plutus/pull/5954) PR reduces the size of the file ([new](https://github.com/IntersectMBO/plutus/blob/14a6ac9d8367375427d297be9b8f080bde357e1b/plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/strictLetRec.golden) version) by 2x.

zliu41:
> I think nobody is going to read it regardless.

We did read `andExpensive.uplc.golden` etc., and drew conclusions from them. I'm suggesting doing the same to `strictLetRec` - I don't see a reason not to. It will need to be read if we change something and the `strictLetRec` test fails.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests.hs
Unisay: Thanks, will use `fromValue` next time!

PR #5940, opened 2024-04-30
Do not build the plutus executable if GHC <9.6

zliu41:
It doesn't build with 9.2: https://github.com/IntersectMBO/cardano-haskell-packages/actions/runs/8902018953/job/24447188265


bezirg:
Thanks, I didn't know. I will look  if there is a solution to this, but most likely there isn't.

PR #5943, opened 2024-05-01
[Test] Do-notation for 'TestNested'

effectfully:
Don't look here yet.


plutus-core/plutus-core/test/Pretty/Readable.hs
```diff
@@ -31,8 +31,7 @@ test_PrettyReadable =
   where
     folder :: Pretty fun => PlcFolderContents DefaultUni fun -> TestTree
     folder
-        = runTestNestedIn ["plutus-core", "test", "Pretty", "Golden"]
-        . testNested "Readable"
+        = runTestNested ["plutus-core", "test", "Pretty", "Golden", "Readable"]
```
effectfully, line 34: I've realized that we do `testNested` after every call to `runTestNestedIn`, so we could as well just incorporate the former into the latter and save on typing and complexity.

plutus-core/plutus-core/test/TypeSynthesis/Spec.hs
```diff
@@ -128,27 +127,26 @@ test_typecheckIllTyped =
...
+    :: forall fun. (ToBuiltinMeaning DefaultUni fun, Show fun, Show (BuiltinSemanticsVariant fun))
     => String
     -> BuiltinSemanticsVariant fun
-    -> TestTree
-test_typecheckAllFun name semvar
-    = runTestNestedIn ["plutus-core", "test", "TypeSynthesis", "Golden"]
-    . testNested name
+    -> TestNested
+test_typecheckAllFun name semVar
+    = testNestedNamed name (show semVar)
```
effectfully, line 135: This now gives better output, before it would just print `DefaultFun` for every semantics variant of `Default`, now it prints the semantics variant instead.

effectfully:
Garbage. Closing in favor of #5948.

PR #5947, opened 2024-05-02
UCaseOfCase Translation Relation

### LINKED ISSUE #5916: Agda translation relation for CaseOfCase UPLC phase
The [CaseOfCase phase](https://github.com/IntersectMBO/plutus/blob/master/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs) in the UPLC compiler needs an Agda translation relation of the form defined in [the paper](https://iohk.io/en/research/library/papers/translation-certification-for-smart-contracts-scp/).

ramsay-t:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

Fixes #5916

### DIFF
--- a/doc/notes/verified-compilation/isabelle/Inline.thy
+++ b/doc/notes/verified-compilation/isabelle/Inline.thy
@@ -1,0 +1,109 @@
+section \<open>Outline\<close>
+text \<open>This is just an experiment with presenting the datatypes and relations from Jacco's
+      paper in Iasbelle and working out how well (or otherwise) sledgehamemr et al. can handle
+      generating proofs for them. \<close>
+
+theory Inline
+  imports 
+    Main
+    HOL.String
+    HOL.List
+begin
+
+type_synonym Name = string
+
+section \<open>Simple Lambda Calculus\<close>
+
+text \<open>A very cut down lambda calculus with just enough to demo the functions below. \<close>
+
+text \<open>I have commented out the Let definition because it produces some slightly complicated
+      variable name shadowing issues that wouldn't really help with this demo. \<close>
+
+datatype AST = Var Name
+  | Lam Name AST
+  | App AST AST
+(*   | Let Name AST AST *) (* The definition below does some name introduction without 
+                              checking things are free, which could cause shadowing, 
+                              which in turn makes the proofs harder...*)
+
+text \<open>Simple variable substitution is useful later.\<close>
+
+fun subst :: "AST \<Rightarrow> Name \<Rightarrow> AST \<Rightarrow> AST" where
+"subst (Var n) m e = (if (n = m) then e else Var n)"
+| "subst (Lam x e) m e' = Lam x (subst e m e')"
+| "subst (App a b) m e' = App (subst a m e') (subst b m e')"
+(* | "subst (Let x xv e) m e' = (Let x (subst xv m e') (subst e m e'))" (* how should this work if x = m? *) *)
+
+text \<open>Beta reduction is just substitution when done right. 
+      This isn't actually used in this demo, but it works...\<close>
+
+fun beta_reduce :: "AST \<Rightarrow> AST" where
+"beta_reduce (App (Lam n e) x) = (subst e n x)"
+| "beta_reduce e = e"
+
+section \<open>Translation Relation\<close>
+
+text \<open>In a context, it is valid to inline variable values.\<close>
+
+type_synonym Context = "Name \<Rightarrow> AST"
+
+fun extend :: "Context \<Rightarrow> (Name * AST) \<Rightarrow> Context" where
+"extend \<Gamma> (n, ast) = (\<lambda> x . if (x = n) then ast else \<Gamma> x)"
+
+text \<open>This defintion is from figure 2 of 
+      [the paper](https://iohk.io/en/research/library/papers/translation-certification-for-smart-contracts-scp/)\<close>
+
+inductive inline :: "Context \<Rightarrow> AST \<Rightarrow> AST \<Rightarrow> bool" ("_ \<turnstile> _ \<triangleright> _" 60) where
+"\<lbrakk>(\<Gamma> x) = y ; \<Gamma> \<turnstile> y \<triangleright> y' \<rbrakk> \<Longrightarrow> \<Gamma> \<turnstile> (Var x) \<triangleright> y'"
+| "\<Gamma> \<turnstile> (Var x) \<triangleright> (Var x)"
+(* | "\<lbrakk> \<Gamma> \<turnstile> t1 \<triangleright> t1' ; (extend \<Gamma> (x,t1)) \<turnstile> t2 \<triangleright> t2' \<rbrakk> \<Longrightarrow> \<Gamma> \<turnstile> (Let x t1 t2) \<triangleright> (Let x t1' t2')" *)
+| "\<lbrakk> \<Gamma> \<turnstile> t1 \<triangleright> t1' ; \<Gamma> \<turnstile> t2 \<triangleright> t2' \<rbrakk> \<Longrightarrow> \<Gamma> \<turnstile> App t1 t2 \<triangleright> App t1' t2'"
+| "\<Gamma> \<turnstile> t1 \<triangleright> t1' \<Longrightarrow> \<Gamma> \<turnstile> Lam x t1 \<triangleright> Lam x t1'"
+
+text \<open>Idempotency is useful for resolving inductive substitutions.\<close>
+
+lemma inline_idempotent : "\<Gamma> \<turnstile> x \<triangleright> x"
+proof (induct x)
+  case (Var x)
+  then show ?case by (rule inline.intros(2))
+next
+  case (Lam x1a x)
+  then show ?case by (rule inline.intros(4))
+next
+  case (App x1 x2)
+  then show ?case by (rule inline.intros(3))
+(* This requires x1a to be free in x1 and x2, or to shadow in a consistent way, which is a hassle to demonstrate...
+next
+  case (Let x1a x1 x2)
+  then show ?case
+    apply (rule_tac ?x="x1a" in VerifiedCompilation.inline.intros(3))
+     apply simp
+*) 
+qed
+
+section \<open>Experiments\<close>
+
+text \<open>We can present some simple pairs of ASTs and ask sledgehammer to produce proofs.
+      For all of these it just works with the simplifier, since they are just applications
+      of the definition of the translation. \<close>
+
+lemma demo_inline1 : "\<lbrakk> \<Gamma> x = (Var y) \<rbrakk> \<Longrightarrow> \<Gamma> \<turnstile> (Var x) \<triangleright> (Var y)"
+  by (simp add: inline.intros(1) inline_idempotent)
+
+lemma demo_inline2 : "\<lbrakk> \<Gamma> f = (Lam x (App g (Var x))) ; \<Gamma> y = (Var b) \<rbrakk> \<Longrightarrow> \<Gamma> \<turnstile> (App (Var f) (Var y)) \<triangleright> (App (Lam x (App g (Var x))) (Var b))"
+  by (simp add: inline.intros(1) inline.intros(3) inline_idempotent)
+
+lemma demo_inline_2step : "\<lbrakk> \<Gamma> f = (Lam x (Lam y (App (Var x) (Var y)))) ; \<Gamma> b = (Var z) \<rbrakk> \<Longrightarrow> \<Gamma> \<turnstile> (App (Var f) (Var b)) \<triangleright> (App (Lam x (Lam y (App (Var x) (Var y)))) (Var z))"
+  by (simp add: inline.intros(1) inline.intros(3) inline_idempotent)
+
+lemma demo_inline_4step : "\<lbrakk> \<Gamma> f = (Lam x (Lam y (App (Var x) (Var y)))) ; \<Gamma> b = (Var c) ; \<Gamma> c = (Var d) ; \<Gamma> d = (App (Var i) (Var j)) \<rbrakk> \<Longrightarrow> \<Gamma> \<turnstile> (App (Var f) (Var b)) \<triangleright> (App (Lam x (Lam y (App (Var x) (Var y)))) (App (Var i) (Var j)))"
+  using inline.intros(1) inline.intros(3) inline_idempotent by presburger
+
+text \<open>A false example to show verification actually happen! This reqriting isn't valid and 
+       nitpick can show you a counterexample pretty quickly (although it isn't very readable).\<close>
+lemma demo_wrong_inline : "\<lbrakk> \<Gamma> f = (Lam x (Lam y (App (Var x) (Var y)))) \<rbrakk> \<Longrightarrow> \<Gamma> \<turnstile> (App (Var f) (Var z)) \<triangleright> (App (Var x) (Var y))"
+  nitpick
+  sorry
+
+end
+
--- a/plutus-metatheory/src/Untyped.lagda.md
+++ b/plutus-metatheory/src/Untyped.lagda.md
@@ -40,6 +40,12 @@ open Untyped
 
 ## Well-scoped Syntax
 
+This defines the syntax for UPLC and requires that it be "well scoped", which
+is that only variables in the context are used. The context uses de Bruijn naming,
+so the variables are numbered. This numbering is provided by an indctivley defined
+natural number, which uses the Maybe type (so: `Nothing` = zero, `Just Just Nothing` = 2)
+to allow direct translation to Haskell. 
+
 ```
 data _⊢ (X : Set) : Set where
   `   : X → X ⊢
--- a/plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
+++ b/plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
@@ -38,7 +38,9 @@ data _⊢̂_⊳̂_ (X : Set) : (X ⊢) → (X ⊢) → Set where
                 → X ⊢̂ tt ⊳̂ tt' -- recursive translation for the true branch
                 → X ⊢̂ ft ⊳̂ ft' -- recursive translation for the false branch
                 ----------------------
-                → X ⊢̂ (case ((((force (builtin ifThenElse)) · b) · tt) · ft) alts) ⊳̂ ((((force (builtin ifThenElse)) · b) · (case tt' alts')) · (case ft' alts'))
+                → X ⊢̂
+                       (case ((((force (builtin ifThenElse)) · b) · tt) · ft) alts)
+                       ⊳̂ ((((force (builtin ifThenElse)) · b) · (case tt' alts')) · (case ft' alts'))
    var : ∀ {x : X} → X ⊢̂ (` x) ⊳̂ (` x) 
    ƛ   : ∀ {x x' : Maybe X ⊢}
            → Maybe X ⊢̂ x ⊳̂ x'
@@ -71,7 +73,9 @@ expect anything else to be unaltered.
 This can just traverse the ASTs applying the constructors from the transition relation.
 
 ```
---_⊢̂_⊳̂?_ : {X : Set} → (X ⊢) → (X ⊢) → Dec (_⊢̂_⊳̂_)
+--_⊢̂_⊳̂?_ : {X : Set} (ast ast' : X ⊢) → Dec (X ⊢̂ ast ⊳̂ ast')
+--_⊢̂_⊳̂?_ ast ast' = ?
+
 ```
 
 ## Semantic Equivalence
@@ -90,5 +94,8 @@ large gas budget and begin in an empty context (which assumes the term is closed
 ```
 --semantic_equivalence : ∀ {X set} {ast ast' : ⊥ ⊢}
  --                    → ⊥ ⊢ ast ⊳ ast'
+ -- <Some stuff about whether one runs out of gas -- as long as neither runs out of gas, _then_ they are equivilent> 
  --                    → (stepper maxsteps  (Stack.ϵ ; [] ▻ ast)) ≡ (stepper maxsteps (Stack.ε ; [] ▻ ast'))
+
+-- ∀ {s : ℕ} → stepper s ast ≡ <valid terminate state> ⇔ ∃ { s' : ℕ } [ stepper s' ast' ≡ <same valid terminate state> ] 
 ```
--- a/plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
+++ b/plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
@@ -18,6 +18,7 @@ open import Utils as U using (Maybe;nothing;just;Either)
 open import RawU using (TmCon)
 open import Data.List.Relation.Unary.All using (All; []; _∷_)
 open import Data.Product using (_×_; proj₁; proj₂)
+open import Relation.Nullary using (Dec; yes; no)
 ```
 
 ## Translation Relation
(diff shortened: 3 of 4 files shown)


plutus-metatheory/src/Untyped.lagda.md
```diff
@@ -40,6 +40,12 @@ open Untyped
 
 ## Well-scoped Syntax
 
+This defines the syntax for UPLC and requires that it be "well scoped", which
+is that only variables in the context are used. The context uses de Bruijn naming,
+so the variables are numbered. This numbering is provided by an indctivley defined
```
ana-pantilie, line 45:
```suggestion
so the variables are numbered. This numbering is provided by an inductively defined
```

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+
+This compiler stage only applies to the very specific case where an `IfThenElse` builtin exists in a `case` expression.
+It moves the `IfThenElse` outside and creates two `case` expressions with each of the possible lists of cases. 
+
+This translation relation has a constructor for the specific case, and then inductive constructors for everything else
+to traverse the ASTs.
+
+```
+data _⊢̂_⊳̂_ (X : Set) : (X ⊢) → (X ⊢) → Set where
+--   refl : ∀ {X : Set} {x : X ⊢} → X ⊢̂ x ⊳̂ x
```
ana-pantilie, line 34: Should this comment be removed?

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Decision Procedure
+
+Since this compiler phase is just a syntax re-arrangement in a very particular situation, we can
+match on that situation in the before and after ASTs and apply the translation rule for this, or
+expect anything else to be unaltered.
+
+This can just traverse the ASTs applying the constructors from the transition relation.
+
+```
```
ana-pantilie, line 74: Maybe add a TODO comment here?

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+isn't a refinement argument - the state before and after the operation is the same type, and is
+unaltered buy the syntax re-arrangement.
+
+This does rely on the encoding of the semantics of `IfThenElse` in the CEK module, since we
+need to show that the effective list of cases is the same as it would have been without the re-arrangement.
+
+The `stepper` function uses the CEK machine to evaluate a term. Here we call it with a very
+large gas budget and begin in an empty context (which assumes the term is closed).
+
+```
```
ana-pantilie, line 93: Add a TODO here as well?

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+data _⊢̂_⊳̂_ (X : Set) : (X ⊢) → (X ⊢) → Set where
+--   refl : ∀ {X : Set} {x : X ⊢} → X ⊢̂ x ⊳̂ x
+   caseofcase : ∀ {b : X ⊢} {tn tn' fn fn' : ℕ} {alts alts' tt ft tt' ft' : List (X ⊢)} 
+                → All (λ altpair → X ⊢̂ (proj₁ altpair) ⊳̂ (proj₂ altpair)) (zip alts alts') -- recursive translation for the other case patterns 
+                → All (λ altpair → X ⊢̂ (proj₁ altpair) ⊳̂ (proj₂ altpair)) (zip tt tt') -- recursive translation for the true branch
+                → All (λ altpair → X ⊢̂ (proj₁ altpair) ⊳̂ (proj₂ altpair)) (zip ft ft') -- recursive translation for the false branch
+                ----------------------
+                → X ⊢̂
+                       (case ((((force (builtin ifThenElse)) · b) · (constr tn tt)) · (constr fn ft)) alts)
+                       ⊳̂ ((((force (builtin ifThenElse)) · b) · (case (constr tn' tt') alts')) · (case (constr tn' tt') alts'))
```
ana-pantilie, line 42: Is there a way to format this on multiple lines to make it more readable?

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+   ƛ   : ∀ {x x' : Maybe X ⊢}
+           → Maybe X ⊢̂ x ⊳̂ x'
+           ----------------------
+           →  X ⊢̂ ƛ x ⊳̂ ƛ x' 
+   _·_ : ∀ {f t f' t' : X ⊢} → (X ⊢̂ f ⊳̂ f') → (X ⊢̂ t ⊳̂ t') → (X ⊢̂ (f · t) ⊳̂ (f' · t'))
+   force : ∀ {t t' : X ⊢} → X ⊢̂ t ⊳̂ t' → X ⊢̂ force t ⊳̂ force t'  
+   delay : ∀ {t t' : X ⊢} → X ⊢̂ t ⊳̂ t' → X ⊢̂ delay t ⊳̂ delay t'  
+   con : ∀ {tc : TmCon} → X ⊢̂ con tc ⊳̂ con tc
+   constr : ∀ {xs xs' : List (X ⊢)} { n : ℕ }
+                → All (λ altpair → X ⊢̂ (proj₁ altpair) ⊳̂ (proj₂ altpair)) (zip xs xs')
```
ana-pantilie, line 53: What does the `All (λ altpair → X ⊢̂ (proj₁ altpair) ⊳̂ (proj₂ altpair)) (zip xs xs')` mean?

doc/notes/verified-compilation/isabelle/Inline.thy
```diff
@@ -0,0 +1,109 @@
...
+(*   | Let Name AST AST *) (* The definition below does some name introduction without 
+                              checking things are free, which could cause shadowing, 
+                              which in turn makes the proofs harder...*)
+
+text \<open>Simple variable substitution is useful later.\<close>
+
+fun subst :: "AST \<Rightarrow> Name \<Rightarrow> AST \<Rightarrow> AST" where
+"subst (Var n) m e = (if (n = m) then e else Var n)"
+| "subst (Lam x e) m e' = Lam x (subst e m e')"
+| "subst (App a b) m e' = App (subst a m e') (subst b m e')"
```
effectfully, line 34: Not capture-avoiding?

doc/notes/verified-compilation/isabelle/Inline.thy
```diff
@@ -0,0 +1,109 @@
...
+
+fun extend :: "Context \<Rightarrow> (Name * AST) \<Rightarrow> Context" where
+"extend \<Gamma> (n, ast) = (\<lambda> x . if (x = n) then ast else \<Gamma> x)"
+
+text \<open>This defintion is from figure 2 of 
+      [the paper](https://iohk.io/en/research/library/papers/translation-certification-for-smart-contracts-scp/)\<close>
+
+inductive inline :: "Context \<Rightarrow> AST \<Rightarrow> AST \<Rightarrow> bool" ("_ \<turnstile> _ \<triangleright> _" 60) where
+"\<lbrakk>(\<Gamma> x) = y ; \<Gamma> \<turnstile> y \<triangleright> y' \<rbrakk> \<Longrightarrow> \<Gamma> \<turnstile> (Var x) \<triangleright> y'"
+| "\<Gamma> \<turnstile> (Var x) \<triangleright> (Var x)"
```
effectfully, line 58: I feel like this syntax exists solely for the purpose of messing with human beings.

doc/notes/verified-compilation/isabelle/Inline.thy
```diff
@@ -0,0 +1,109 @@
...
+(* | "\<lbrakk> \<Gamma> \<turnstile> t1 \<triangleright> t1' ; (extend \<Gamma> (x,t1)) \<turnstile> t2 \<triangleright> t2' \<rbrakk> \<Longrightarrow> \<Gamma> \<turnstile> (Let x t1 t2) \<triangleright> (Let x t1' t2')" *)
+| "\<lbrakk> \<Gamma> \<turnstile> t1 \<triangleright> t1' ; \<Gamma> \<turnstile> t2 \<triangleright> t2' \<rbrakk> \<Longrightarrow> \<Gamma> \<turnstile> App t1 t2 \<triangleright> App t1' t2'"
+| "\<Gamma> \<turnstile> t1 \<triangleright> t1' \<Longrightarrow> \<Gamma> \<turnstile> Lam x t1 \<triangleright> Lam x t1'"
+
+text \<open>Idempotency is useful for resolving inductive substitutions.\<close>
```
effectfully, line 63: What's an inductive substitution?

plutus-metatheory/src/Untyped.lagda.md
```diff
@@ -40,6 +40,12 @@ open Untyped
...
+natural number, which uses the Maybe type (so: `Nothing` = zero, `Just Just Nothing` = 2)
```
effectfully, line 46:
```suggestion
natural number, which uses the Maybe type (so: `Nothing` = zero, `Just (Just Nothing)` = 2)
```

Perhaps it's also worth referencing [de Bruijn notation as a nested datatype](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/D8BFA383FDA7EA3DC443B4C42A168F30/S0956796899003366a.pdf/de-bruijn-notation-as-a-nested-datatype.pdf) for a detailed explanation of the trick.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+It moves the `IfThenElse` outside and creates two `case` expressions with each of the possible lists of cases. 
+
+This translation relation has a constructor for the specific case, and then inductive constructors for everything else
+to traverse the ASTs.
+
+```
+data _⊢̂_⊳̂_ (X : Set) : (X ⊢) → (X ⊢) → Set where
+--   refl : ∀ {X : Set} {x : X ⊢} → X ⊢̂ x ⊳̂ x
+   caseofcase : ∀ {b : X ⊢} {tn tn' fn fn' : ℕ} {alts alts' tt ft tt' ft' : List (X ⊢)} 
+                → All (λ altpair → X ⊢̂ (proj₁ altpair) ⊳̂ (proj₂ altpair)) (zip alts alts') -- recursive translation for the other case patterns
```
effectfully, line 36: What if `alts` and `alt'` have different lengths? Same question about `tt`/`tt'` and `ft`/`ft'`.

Also I'm very much not a fan of the cryptic names.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+
+This translation relation has a constructor for the specific case, and then inductive constructors for everything else
+to traverse the ASTs.
+
+```
+data _⊢̂_⊳̂_ (X : Set) : (X ⊢) → (X ⊢) → Set where
+--   refl : ∀ {X : Set} {x : X ⊢} → X ⊢̂ x ⊳̂ x
+   caseofcase : ∀ {b : X ⊢} {tn tn' fn fn' : ℕ} {alts alts' tt ft tt' ft' : List (X ⊢)} 
+                → All (λ altpair → X ⊢̂ (proj₁ altpair) ⊳̂ (proj₂ altpair)) (zip alts alts') -- recursive translation for the other case patterns 
+                → All (λ altpair → X ⊢̂ (proj₁ altpair) ⊳̂ (proj₂ altpair)) (zip tt tt') -- recursive translation for the true branch
```
effectfully, line 37: Agda has pattern matching lambda, so you can write `λ { (x , y) -> <...> }` for readability. Probably doesn't matter though.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+                → All (λ altpair → X ⊢̂ (proj₁ altpair) ⊳̂ (proj₂ altpair)) (zip ft ft') -- recursive translation for the false branch
+                ----------------------
+                → X ⊢̂
+                       (case ((((force (builtin ifThenElse)) · b) · (constr tn tt)) · (constr fn ft)) alts)
```
effectfully, line 41: No translation for `b`? It can be an arbitrary expression.

Why does it have to be `constr` rather than an arbitrary expression?

Also `ifThenElse` is strict, since the language is strict and all builtins are strict, hence what we normally do is represent the branches of `ifThenElse` as functions or `delay`ed values and that is what ends up being in the actual validators, so as stated this translation relation isn't very useful.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully, line 53: Again, what if the lengths differ? This doesn't look correct.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+   _·_ : ∀ {f t f' t' : X ⊢} → (X ⊢̂ f ⊳̂ f') → (X ⊢̂ t ⊳̂ t') → (X ⊢̂ (f · t) ⊳̂ (f' · t'))
+   force : ∀ {t t' : X ⊢} → X ⊢̂ t ⊳̂ t' → X ⊢̂ force t ⊳̂ force t'  
+   delay : ∀ {t t' : X ⊢} → X ⊢̂ t ⊳̂ t' → X ⊢̂ delay t ⊳̂ delay t'  
+   con : ∀ {tc : TmCon} → X ⊢̂ con tc ⊳̂ con tc
+   constr : ∀ {xs xs' : List (X ⊢)} { n : ℕ }
+                → All (λ altpair → X ⊢̂ (proj₁ altpair) ⊳̂ (proj₂ altpair)) (zip xs xs')
+                ------------------------
+                → X ⊢̂ constr n xs ⊳̂ constr n xs' 
+   case :  ∀ {p p' : X ⊢} {alts alts' : List (X ⊢)}
+                → All (λ altpair → X ⊢̂ (proj₁ altpair) ⊳̂ (proj₂ altpair)) (zip alts alts') -- recursive translation for the other case patterns
```
effectfully, line 57: And here.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+                → X ⊢̂ p ⊳̂ p'
+                ------------------------
+                → X ⊢̂ case p alts ⊳̂ case p alts'
```
effectfully, line 60:
```suggestion
                → X ⊢̂ case p alts ⊳̂ case p' alts' 
```

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+## Decision Procedure
+
+Since this compiler phase is just a syntax re-arrangement in a very particular situation, we can
+match on that situation in the before and after ASTs and apply the translation rule for this, or
+expect anything else to be unaltered.
+
+This can just traverse the ASTs applying the constructors from the transition relation.
+
+```
+--_⊢̂_⊳̂?_ : {X : Set} (ast ast' : X ⊢) → Dec (X ⊢̂ ast ⊳̂ ast')
```
effectfully, line 75: I'd start with `Maybe`, `Dec` is probably going to be quite painful.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+--_⊢̂_⊳̂?_ ast ast' = ?
+
+```
+
+## Semantic Equivalence
+
+We can show that this translation never alters the semantics of the statement. This is shown
+in terms of the CEK machine evaluation. Since it is a simple re-arrangement of the syntax, it
+isn't a refinement argument - the state before and after the operation is the same type, and is
+unaltered buy the syntax re-arrangement.
```
effectfully, line 85:
```suggestion
unaltered by the syntax re-arrangement.
```

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully, line 42: Again, this will evaluate both the branches of `ifThenElse`, i.e. you're going to evaluate two branches of `alts'` in total in general (in a particular case they could both end up being the same branch), while the original term doesn't have behavior (it does evaluate both branches of `ifThenElse`, but there are no `alts` inside of them, just `constr`s).

I.e. I think this translation is not semantics preserving.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+
+This does rely on the encoding of the semantics of `IfThenElse` in the CEK module, since we
+need to show that the effective list of cases is the same as it would have been without the re-arrangement.
+
+The `stepper` function uses the CEK machine to evaluate a term. Here we call it with a very
+large gas budget and begin in an empty context (which assumes the term is closed).
+
+```
+--semantic_equivalence : ∀ {X set} {ast ast' : ⊥ ⊢}
```
effectfully, line 94: Bottom rather than top?

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+ --                    → ⊥ ⊢ ast ⊳ ast'
+ -- <Some stuff about whether one runs out of gas -- as long as neither runs out of gas, _then_ they are equivilent>
```
effectfully, line 96: Can we ignore gas problems? Why do we care about them? Some optimizations may slow the program down in edge cases while significantly improving performance on average and allowing other optimizations to fire.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+
+We can show that this translation never alters the semantics of the statement. This is shown
+in terms of the CEK machine evaluation. Since it is a simple re-arrangement of the syntax, it
+isn't a refinement argument - the state before and after the operation is the same type, and is
+unaltered buy the syntax re-arrangement.
+
+This does rely on the encoding of the semantics of `IfThenElse` in the CEK module, since we
+need to show that the effective list of cases is the same as it would have been without the re-arrangement.
+
+The `stepper` function uses the CEK machine to evaluate a term. Here we call it with a very
```
effectfully, line 90: I don't mind using the CEK machine, but why? Why not simply small-step/big-step semantics? And then test that those match with the behavior of the CEK machine if we want to ensure that this is the case.

effectfully (commented):
As I'm not a cylon, I cannot read the Isabelle code. The Agda one looks wrong to me in multiple places, but maybe I'm being silly.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: Yes ... unless it turns out that having a simple pattern for things that are identical is useful... It will certainly be quicker than needing to traverse subtrees that are completely identical.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: All of these are horrible and will probably be better with the Pointwise thing Mauro has shown me.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: This is the least of the problems with this :)

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: This is where I was wanting your input: the comments on the Haskell, and my understanding of the Haskell suggest that the expressions have to be `constr`. We (Ana and Jacco really) came up with a reasonable explanation as to why this was important, but perhaps you can explain it better (and we should update the comments)?

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: If the lengths differ then the two things shouldn't match, but the Pointwise approach probably handles this more gracefully.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: I was led to believe the CEK machine was closer to what "really happens"? It also already exists, rather than producing a new semantics and a new equivalence...

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: This is the Decision procedure, so needs to be `Dec`. Having `Set` for `X` instead of `Maybe` is proving quite painful, though.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: They ought to be closed, so they should work with nothing in scope outside of themselves. However, maybe they need the three contract parameters?

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: Jacco made the valid point that we will be doing inductive proofs so if the gas limit is finite we need to account for that or the induction won't work.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: This is also what the tests seem to document: https://github.com/IntersectMBO/plutus/blob/f1eae1c5c8def282d91cf38a58c1493e09e1fa70/plutus-core/untyped-plutus-core/test/Transform/Simplify.hs#L48

effectfully: Weird! I'll look into it.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully: What I'm trying to say is that you don't need to make it a decision procedure right away and can start with "it either translates or I'm not sure" a.k.a. `Maybe`. You can do `Dec` right away, but it's probably going to be much more painful than taking a smaller step first.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully: Right, silly me, it should be bottom of course. Sorry!

> However, maybe they need the three contract parameters?

I think not, but not really sure. IIRC some scripts take a different number of parameters, but don't quote me on that.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully: Gas is just a specific way to look the non-termination effect (see [Turing-Completeness Totally Free](https://strathprints.strath.ac.uk/60166/1/McBride_LNCS2015_Turing_completeness_totally_free.pdf)), so it's not that you have to have it to do proofs, but I guess it's the simplest way to look at non-termination.

I'm a bit confused: my original comment was about "gas" in Plutus, i.e. CPU and MEM units, is this comment in the Agda code about a completely unrelated notion of gas that we need just for easy termination argument in Agda? In that case I'm totally fine with it.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully: The CEK machine _is_ what really happens, but it's probably going to be harder for you to establish a connection between the translation and the CEK machine. The CEK machine should comply with some sort of small-step/big-step semantics and I'd guess that establishing the connections this way:

```
CEK <~> operational semantics <~> the translation relation
```

Also note that the CEK machine is an evaluator and not a normalizer, it's not going to look under unapplied lambdas while optimizations certainly will.

Anyway, if you can make the CEK machine work, then that's great, I just feel like you might be complicating your life.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully: They shouldn't match, but they will, will they not? This is how `zip` is defined:

```
zipWith : (A → B → C) → List A → List B → List C
zipWith f (x ∷ xs) (y ∷ ys) = f x y ∷ zipWith f xs ys
zipWith f _        _        = []
```

it drops all the extra elements of the longer list. And then `All` will not apply to them. And so you can introduce arbitrary garbage at the end of one of the lists and it will be ignored by the translation.

Or am I missing something?

ramsay-t: This is a good point - to be honest, I put the zip in as a placeholder because I didn't like it but hadn't thought of a better solution. I'll look further at the Pointwise stuff.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: This Agda is just suppose to represent what the Haskell does - if you think the Haskell is wrong then we should fix that first :)

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: Yes, James and I have discussed creating a cleaner Reduction semantics but then we need to show that the CEK machine implements that, which is an extra thing to do. However, possibly a worthwhile on if there are a lot of these phases to prove.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: My problem is simpler: the CEK semantics has clauses for `gasError` so we have to handle those somehow in the induction proofs :)

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: James made the good point that the before and after things should reduce in the same way when treated as close terms. If they subsequently get applied to stuff I don't think that matters to them being equivalent as-is?

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully: > the comments on the Haskell, and my understanding of the Haskell suggest that the expressions have to be `constr`

Yes, it turns out you're right. Was discussed [here](https://github.com/IntersectMBO/plutus/pull/5337#discussion_r1199215315).

> Also `ifThenElse` is strict, since the language is strict and all builtins are strict, hence what we normally do is represent the branches of `ifThenElse` as functions or delayed values and that is what ends up being in the actual validators, so as stated this translation relation isn't very useful.

This does look like a bug with the Haskell code, I've talked with Yura about it and he's going to take a look.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully: Yep, you're totally right, I completely missed the fact that it's the Haskell code that I find deeply suspicious.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully: > James made the good point that the before and after things should reduce in the same way when treated as close terms. If they subsequently get applied to stuff I don't think that matters to them being equivalent as-is?

I'd expect for some reasonable operational semantics to exist for functions that aren't applied. Some kind of bisimulation relation or behavioral equivalence or whatever. 

"For any given arguments these two functions return the same result" for example.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: This is what Formal Methods are supposed to do - force us to look at the details of stuff we haven't previously :) Let me know what the "right" answer is and I'll make the spec match _that_.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
ramsay-t: This is what a sensibly abstract reduction semantics would do. Apparently there was one but it was ... "lost"... I shall look behind the sofa, and then might end up making a new one.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully: > This is what Formal Methods are supposed to do - force us to look at the details of stuff we haven't previously :) 

We did look at it at review time! At the general rule and at examples. I have no idea how we missed it. But let's not get ahead of ourselves, maybe I'm talking gibberish and it's all fine. We'll get to know soon.

> Let me know what the "right" answer is and I'll make the spec match that.

Will do.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
effectfully: You could pretend it didn't exist if you wanted to, because the machine supports a budgeting mode that doesn't have any gas limitations.

plutus-metatheory/src/VerifiedCompilation/UCaseOfCase.lagda.md
kwxm: > I'm pretty sure that I heard some mention of a proposal to make sure that all scripts take the same number of parameters.

See https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2078414218.

ramsay-t:
Now fixes #5944

ramsay-t:
Now *actually* fixes #5944 and #5916

PR #5953, opened 2024-05-03
Kwxm/experiment/strictify fix nofib

kwxm:
I just want to run the nofib benchmarks on #5939 , but that's not working.  This is a copy of that branch rebased on a recent master.


kwxm:
Seems to speed things up a bit.

PR #5954, opened 2024-05-04
 [PIR] Don't generate 'fixBy' if you don't need to

effectfully:
Currently we generate `fixBy` (a combinator used for compiling mutual recursion) whenever we have any recursive functions in the PIR program, but most of those only need the basic Z combinator and no fancy `fixBy` stuff, so generating the latter is pointless and can be very confusing if you want to look at the output of the compiler without optimizations (in particular, DCE) turned on. Not generating `fixBy` if you don't need to is trivial, so let's do that.


plutus-core/plutus-ir/src/PlutusIR/Compiler/Recursion.hs
```diff
@@ -98,15 +98,16 @@ mkFixpoint bs = do
           name <- liftQuote $ toProgramName fixByKey
           let (fixByTerm, fixByType) = Function.fixByAndType
           pure (PLC.Def (PLC.VarDecl noProvenance name (noProvenance <$ fixByType)) (noProvenance <$ fixByTerm, Strict), mempty)
-    fixBy <- lookupOrDefineTerm p0 fixByKey mkFixByDef
```
Unisay, line 101: This change does make sense to me: the `fixBy` is generated only if arity /= 0.

Unisay (changes requested):
Unfortunately I can't see a change in a golden file that demonstrates how `fixBy` was generated before the change and how its not generated after. 

Do you mind adding a golden file demonstrating the desired effect?

effectfully:
> Unfortunately I can't see a change in a golden file that demonstrates how `fixBy` was generated before the change and how its not generated after.
>
> Do you mind adding a golden file demonstrating the desired effect?

Well, I can't do that in this PR, because this PR removes generation of `fixBy`, so any new golden file will not contain it. But anyway, the idea was that #5939 adds such a test, so we'll have it soon without me needing to add it separately. Let's wait for #5939 to get merged and then I'll update things here.

effectfully:
> But anyway, the idea was that https://github.com/IntersectMBO/plutus/pull/5939 adds such a test, so we'll have it soon without me needing to add it separately. Let's wait for https://github.com/IntersectMBO/plutus/pull/5939 to get merged and then I'll update things here.

I should've said explicitly when creating the PR though. Sorry!

plutus-core/plutus-ir/src/PlutusIR/Compiler/Recursion.hs
effectfully: > `if arity /= 0`

If `arity /= 1`, because `0` isn't supported at all (we're in a strict language) and `1` is handled by the Z combinator.

Unisay: you're right, I made a typo.

Unisay (approved):

plutus-core/plutus-core/stdlib/PlutusCore/StdLib/Data/Function.hs
```diff
@@ -79,6 +79,51 @@ applyFun = runQuote $ do
         . lamAbs () x (TyVar () a)
         $ apply () (var () f) (var () x)
 
+{- Note [Recursion combinators]
```
effectfully, line 82: Oops, I confused branches and added the Note in this one rather than the `strictify-fix` one. Doesn't matter much, I guess.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/strictLetRec.golden
```diff
@@ -3,222 +3,23 @@
...
-      (all
-        F
-        (fun (type) (type))
-        (fun
-          (fun (all Q (type) (fun [ F Q ] Q)) (all Q (type) (fun [ F Q ] Q)))
-          (fun
-            (all Q (type) (fun [ F Q ] [ F Q ])) (all Q (type) (fun [ F Q ] Q))
-          )
-        )
-      )
```
effectfully, line 16: `fixBy` is gone.

PR #5964, opened 2024-05-07
Minor change for experimentation

kwxm:
I just want to run some benchmarks.


kwxm:
These all look consistent with previous runs, which is what I was wondering about.  That's all good then.

PR #5970, opened 2024-05-09
Logical operations

kozross:
This provides implementations for the logical operations, as described in [this CIP](https://github.com/mlabs-haskell/CIPs/blob/koz/logic-ops/CIP-XXX/CIP-XXX.md). I have also corrected references to the newly-numbered CIP-121 for the conversion builtins.

Pre-submit checklist:
- Branch
    - [X] Tests are provided (if possible)
    - [X] Changelog fragments have been written (if appropriate)
    - [X] Relevant tickets are mentioned in commit messages
    - [X] Formatting, PNG optimization, etc. are updated
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested


Unisay (dismissed):
🚢

plutus-core/plutus-core/src/PlutusCore/Bitwise/Logical.hs
kwxm, line 1: There's lots of impressively fiddly code in here. I'm glad we have all those tests.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Logical.hs
```diff
@@ -0,0 +1,461 @@
...
+  This doesn't pose any problems from the point of view of Plutus Core, as this
+  exception cannot 'leak'; we handle it entirely within `writeBits`, and no
+  other Plutus Core code can ever see it.
+-}
+newtype WriteBitsException = WriteBitsException Integer
+  deriving stock (Eq, Show)
+
+instance Exception WriteBitsException
+
+{- Note [Loop sectioning]
```
kwxm, line 313: Good notes!  This one's particularly helpful when trying to understand the code.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Logical.hs
```diff
@@ -0,0 +1,461 @@
...
+import Data.Text (pack)
+import Data.Word (Word64, Word8)
+import Foreign.Marshal.Utils (copyBytes)
+import Foreign.Ptr (Ptr, castPtr, plusPtr)
+import Foreign.Storable (peekByteOff, peekElemOff, pokeByteOff, pokeElemOff)
+import PlutusCore.Builtin (BuiltinResult, emit)
+import PlutusCore.Evaluation.Result (evaluationFailure)
+import System.IO.Unsafe (unsafeDupablePerformIO)
+
+{- Note [Binary bitwise operation implementation and manual specialization]
```
kwxm, line 31: This is very illuminating.  I skipped over it at first and then wondered "Why all the code duplication?". I'm not going to claim I understand all of the details though.  @effectfully You're good at persuading GHC to do what you want: do you have any comments/suggestions?

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Laws.hs
```diff
@@ -0,0 +1,565 @@
...
+getSet :: TestTree
+getSet =
+  testPropertyNamed "get-set" "get_set" . property $ do
+    bs <- forAllByteString1
+    i <- forAllIndexOf bs
+    let lookupExp = mkIterAppNoAnn (builtin () PLC.ReadBit) [
+          mkConstant @ByteString () bs,
+          mkConstant @Integer () i
+          ]
+    case typecheckReadKnownCek def defaultBuiltinCostModel lookupExp of
```
kwxm, line 72: This'll also have to change

```suggestion
    case typecheckReadKnownCek def defaultBuiltinCostModelForTesting lookupExp of
```

kwxm (approved):
This all looks very good. I didn't spot anything major to complain about.  You may have a little trouble merging it because we've updated the way the cost models work recently and I think that'll affect some of the testing code.  I've added some comments about how to change it (I'm not totally sure they're right though: ask us if you run into any problems).

I also left some comments on the CIP, and may leave some more later.

plutus-tx/src/PlutusTx/Builtins.hs
```diff
@@ -639,11 +646,139 @@ integerToByteString :: ByteOrder -> Integer -> Integer -> BuiltinByteString
...
+-- if 'False', truncation semantics will be used instead.
+--
+-- = See also
+--
+-- * [Padding and truncation
+-- semantics](https://github.com/mlabs-haskell/CIPs/blob/koz/logic-ops/CIP-0122/CIP-0122.md#padding-versus-truncation-semantics)
+-- * [Bit indexing
+-- scheme](https://github.com/mlabs-haskell/CIPs/blob/koz/logic-ops/CIP-0122/CIP-0122.md#bit-indexing-scheme)
+{-# INLINEABLE bitwiseLogicalAnd #-}
+bitwiseLogicalAnd ::
```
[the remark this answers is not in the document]
kozross: Will fix. Weird that the formatter doesn't do this for us.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Logical.hs
```diff
@@ -0,0 +1,461 @@
...
+   in go toCopy toTraverse (BS.length shorter)
+  where
+    go :: ByteString -> ByteString -> Int -> ByteString
+    go toCopy toTraverse traverseLen =
+      unsafeDupablePerformIO . BS.useAsCStringLen toCopy $ \(copyPtr, copyLen) ->
+        BS.useAsCString toTraverse $ \traversePtr -> do
+          BSI.create copyLen $ \dstPtr -> do
+            copyBytes dstPtr (castPtr copyPtr) copyLen
+            let (bigStrides, littleStrides) = traverseLen `quotRem` 8
+            let offset = bigStrides * 8
```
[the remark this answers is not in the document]
kozross: Technically yes. Not sure if GHC optimizes this in any way. Probably doesn't really matter either way.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Logical.hs
kozross: It's the reason _why_ I have all these tests. I got it wrong enough times when I first wrote this that I needed the tests to make sure something wasn't subtly broken all the time.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Logical.hs
kozross: I would also love to know if this can be written with less duplicative code. I tried about every trick I know, but anything using HOFs seemed to run like treacle.

kwxm: @effectfully says:

I feel like helper1 would work with an explicit inline call. Is it worth it though?

so like this:
>   While in theory inlining could help with at least the `helper1` case (`helper2` is beyond that technique), it doesn't seem like GHC is able to figure this out, even with `INLINE` is placed on `helper1`.

but also make it `inline helper1` at the call site.

Maybe it's also possible to turn `helper1` into a type class with a default implementation that doesn't take the functions as arguments but instead uses another method of the type class added specifically so that GHC knows what it is.
Or helper1 could take an argument `whichOp` and match on it inside the body of `helper1` to pick either `.&.` or `.|.` or ... One extra pattern match that should be fairly cheap and no headache with GHC not inlining things.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -152,6 +153,14 @@ data DefaultFun
     -- Conversions
     | IntegerToByteString
     | ByteStringToInteger
+    -- Logical
+    | BitwiseLogicalAnd
```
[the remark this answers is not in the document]
kwxm: I see you renamed these: thanks!  In retrospect maybe we should have called the builtins things like `IntegerAdd` and `ByteStringConcat` at the start and then they're have been easier to organise.  The internal names don't really matter, so we could change them at some point in the future if we end up with so many that they're hard to keep track of.

plutus-core/plutus-core/src/PlutusCore/Bitwise/Logical.hs
kozross: I tried an explicit `INLINE` and it didn't help at all.

effectfully: `INLINE` is not enough in some cases, you may need to `GHC.Magic.inline` at the call site. I'd be surprised if it didn't make the call inlined, `inline` is pretty reliable (unlike `INLINE`).

PR #5971, opened 2024-05-10
[PlutusTx] [Builtins] Split 'FromBuiltin' and 'ToBuiltin'

effectfully:
This addresses the [inconsistencies](https://github.com/IntersectMBO/plutus/pull/5547#issuecomment-2101686530) that we got from using the now retired `FromBuiltin`/`ToBuiltin` classes for two mutually exclusive purposes. Now we have two versions of those classes: one is used for converting builtins to Plutus Tx types and back, while the other one is for making builtins transparent/opaque within Plutus Tx. These two purposes are very similar, but also subtly different and that is explained in the docs.


plutus-tx/src/PlutusTx/Lift/Class.hs
```diff
@@ -132,72 +133,71 @@ instance (TypeError ('Text "Int is not supported, use Integer instead"))
...
-instance uni `PLC.HasTypeLevel` BS.ByteString => Typeable uni BS.ByteString where
-    typeRep = typeRepBuiltin
-
--- See Note [Lift and Typeable instances for builtins]
-instance uni `PLC.HasTypeLevel` Data => Typeable uni BuiltinData where
-    typeRep _ = typeRepBuiltin (Proxy @Data)
-
--- See Note [Lift and Typeable instances for builtins]
-instance uni `PLC.HasTermLevel` Data => Lift uni BuiltinData where
-    lift = liftBuiltin . builtinDataToData
```
effectfully, line 152: Moved below to match the order of the constructors of `DefaultUni`.

plutus-tx-plugin/test/Plugin/Primitives/9.6/deconstructorData2.pir.golden
```diff
@@ -247,6 +247,14 @@
             )
           )
         )
+        (termbind
+          (nonstrict)
+          (vardecl
+            `$fHasFromOpaqueBuiltinDataBuiltinData_$cfromOpaque`
+            (fun (con data) (con data))
+          )
+          (lam eta (con data) eta)
```
effectfully, line 256: This is the `id` removed from the below.

plutus-core/plutus-core/src/PlutusCore/Builtin/Polymorphism.hs
```diff
@@ -214,7 +215,8 @@ type family AllElaboratedArgs constr x where
 -- built-in type.
 type AllBuiltinArgs
         :: forall a. (GHC.Type -> GHC.Type) -> (GHC.Type -> GHC.Constraint) -> a -> GHC.Constraint
-type AllBuiltinArgs uni constr x = AllElaboratedArgs constr (ElaborateBuiltin uni x)
+class    AllElaboratedArgs constr (ElaborateBuiltin uni x) => AllBuiltinArgs uni constr x
+instance AllElaboratedArgs constr (ElaborateBuiltin uni x) => AllBuiltinArgs uni constr x
```
effectfully, line 219: There was no reason to make it a class synonym before, but now it's convenient, since we partially apply it in the `TestAllBuiltinsSatisfy` instances below.

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -0,0 +1,243 @@
...
+-- the unfoldings from going in. So we just stick it here. Fiddly.
+instance IsString BuiltinString where
+    -- Try and make sure the dictionary selector goes away, it's simpler to match on
+    -- the application of 'stringToBuiltinString'
+    -- See Note [noinline hack]
+    fromString = Magic.noinline stringToBuiltinString
+    {-# INLINE fromString #-}
+
+{- Note [Built-in types and their Haskell versions]
+'HasToBuiltin' allows us to convert a value of a built-in type such as 'ByteString' to its Plutus
```
zliu41, line 79: What's the use case for this, I think it is only useful for lifting? Worth elaborating.

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -0,0 +1,243 @@
...
+Tx counterpart, 'BuiltinByteString' in this case. The idea is the same for all built-in types: just
+take the Haskell version and make it the Plutus Tx one.
+
+'HasToOpaque' is different, we use it for converting values of only those built-in types that exist
+in the Plutus Tx realm. I.e. we cannot convert a 'ByteString', since 'ByteString's don't exist in
+Plutus Tx, only 'BuiltinByteString's do.
+
+But consider, say, the built-in pair type. In Plutus Tx, we have an (opaque) type for this. It's
```
zliu41, line 87: > But

I don't think there's a contrast here? This is just an example of what the previous paragraph talks about.

plutus-tx/src/PlutusTx/Builtins/HasBuiltin.hs
```diff
@@ -0,0 +1,111 @@
...
+import PlutusCore.Default qualified as PLC
+import PlutusTx.Builtins.Internal
+
+import Data.ByteString (ByteString)
+import Data.Kind qualified as GHC
+import Data.Text (Text)
+
+-- | A class for converting values of Haskell-defined built-in types to their Plutus Tx
+-- counterparts.
+type HasToBuiltin :: GHC.Type -> GHC.Constraint
```
zliu41, line 24: Link to `Note [Built-in types and their Haskell versions]`?

zliu41 (approved):

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
effectfully: Lifting is the primary use case for `fromBuiltin`, yes, but we also have `toBuiltin` which is used a lot in benchmarks, for one example (there are many):

```haskell
benchPairing :: EvaluationContext -> Benchmark
benchPairing ctx =
    case listOfSizedByteStrings 4 4 of
      [b1, b2, b3, b4] ->
          let emptyDst = Tx.toBuiltin BS.empty
              p1 = Tx.bls12_381_G1_hashToGroup (Tx.toBuiltin b1) emptyDst
              p2 = Tx.bls12_381_G2_hashToGroup (Tx.toBuiltin b2) emptyDst
              q1 = Tx.bls12_381_G1_hashToGroup (Tx.toBuiltin b3) emptyDst
              q2 = Tx.bls12_381_G2_hashToGroup (Tx.toBuiltin b4) emptyDst
              prog = mkPairingScript p1 p2 q1 q2
          in bench "pairing" $ benchProgramCek ctx prog
      _ -> error "Unexpected list returned by listOfSizedByteStrings"
```

In `PlutusLedgerApi.V1.Bytes`:

```haskell
fromBytes :: BS.ByteString -> LedgerBytes
fromBytes = LedgerBytes . P.toBuiltin

fromHex :: BS.ByteString -> Either LedgerBytesError LedgerBytes
fromHex = fmap (LedgerBytes . P.toBuiltin) . asBSLiteral
```

In `PlutusLedgerApi.V1.Value`:

```haskell
currencySymbol :: BS.ByteString -> CurrencySymbol
currencySymbol = CurrencySymbol . PlutusTx.toBuiltin

tokenName :: BS.ByteString -> TokenName
tokenName = TokenName . PlutusTx.toBuiltin
```

I.e. `toBuiltin` is used for converting Haskell values into into Plutus Tx values.

Unsurprisingly, `fromBuiltin`  is used for converting Plutus Tx values into Haskell values, for example in `PlutusLedgerApi.V1.Bytes`

```haskell
bytes :: LedgerBytes -> BS.ByteString
bytes = P.fromBuiltin . getLedgerBytes
```

In `PlutusLedgerApi.V1.Value`:

```haskell
fromTokenName :: (BS.ByteString -> r) -> (Text -> r) -> TokenName -> r
fromTokenName handleBytestring handleText (TokenName bs) = either (\_ -> handleBytestring $ PlutusTx.fromBuiltin bs) handleText $ E.decodeUtf8' (PlutusTx.fromBuiltin bs)
```

In `PlutusBenchmark.Marlowe.Util`:

```haskell
makeBuiltinData
  :: LedgerBytes
  -> BuiltinData
makeBuiltinData =
  dataToBuiltinData
    . deserialise
    . LBS.fromStrict
    . fromBuiltin
    . getLedgerBytes
```

So I think the current description is as accurate as it gets? `HasFromOpaque`/`HasToOpaque` are trickier and we need to explain what they are for (hence the Note, most of which I took from the existing Michael's docs), but `HasFromBuiltin` and `HasToBuiltin` are just for converting values between Haskell and Plutus Tx.

plutus-tx/src/PlutusTx/Builtins.hs
```diff
@@ -184,22 +187,23 @@ verifyEd25519Signature
     -> BuiltinByteString  -- ^ Message    (arbirtary length)
     -> BuiltinByteString  -- ^ Signature  (64 bytes)
     -> Bool
-verifyEd25519Signature pubKey message signature = fromBuiltin (BI.verifyEd25519Signature pubKey message signature)
+verifyEd25519Signature pubKey message signature =
```
Unisay, line 190: Thanks for wrapping the long line, even though its unrelated to this PR.

plutus-tx/src/PlutusTx/Builtins/HasBuiltin.hs
```diff
@@ -0,0 +1,113 @@
...
+import PlutusCore.Crypto.BLS12_381.Pairing qualified as BLS12_381.Pairing (MlResult)
+import PlutusCore.Data (Data)
+import PlutusCore.Default qualified as PLC
+import PlutusTx.Builtins.Internal
+
+import Data.ByteString (ByteString)
+import Data.Kind qualified as GHC
+import Data.Text (Text)
+
+-- Also see Note [Built-in types and their Haskell versions].
```
Unisay, line 22:
```suggestion
-- Also see Note [Built-in types and their Haskell implementations].
```
if I understood the idea correctly...

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -0,0 +1,243 @@
...
+
+However, this is harder for the plugin to deal with. It's okay to have a type variable for the
+representation type that needs to be instantiated later, but it's *not* okay to have an irreducible
+type application on a type variable. So fundeps are much nicer here.
+-}
+
+-- See Note [Built-in types and their Haskell versions].
+-- See Note [HasFromOpaque/HasToOpaque instances for polymorphic builtin types].
+-- See Note [Fundeps versus type families in HasFromOpaque/HasToOpaque].
+-- | A class for converting values of transparent Haskell-defined built-in types (such as '()',
```
Unisay, line 139: So "transparent" means "pattern-matchable" here, right?

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -0,0 +1,243 @@
...
+-- 'Bool', '[]' etc) to their opaque Plutus Tx counterparts. Instances for built-in types that are
```
Unisay, line 140: Consequently, "opaque" should mean "non-pattern-matchable".

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -0,0 +1,243 @@
...
+Plutus Tx, only 'BuiltinByteString's do.
+
+Consider, say, the built-in pair type. In Plutus Tx, we have an (opaque) type for this. It's opaque
+because you can't actually pattern match on it, instead you can only in fact use the specific
+functions that are available as builtins.
+
+We _also_ have the normal Haskell pair type. This is very different: you can
+pattern match on it, and you can use whatever user-defined functions you like on it.
+
+Users would really like to use the latter, and not the former. So we often want
```
Unisay, line 94: > Users would really like to use the latter, and not the former.

It would be nice to expand on why users would like it.

plutus-tx/src/PlutusTx/Lift/Class.hs
```diff
@@ -132,70 +133,69 @@ instance (TypeError ('Text "Int is not supported, use Integer instead"))
...
-
 -- See Note [Lift and Typeable instances for builtins]
 instance uni `PLC.HasTypeLevel` BS.ByteString => Typeable uni BuiltinByteString where
-    typeRep _proxyByteString = typeRepBuiltin (Proxy @BS.ByteString)
+    typeRep _ = typeRepBuiltin (Proxy @BS.ByteString)
```
Unisay, line 145: I agree that `_proxyByteString` is an overkill, but IMO `_proxy` is better than `_` because the latter doesn't answer a question "What is being ignored?".

plutus-tx/src/PlutusTx/Builtins.hs
```diff
@@ -184,22 +187,23 @@ verifyEd25519Signature
...
+    fromOpaque (BI.verifyEd25519Signature pubKey message signature)
 
 {-# INLINABLE equalsByteString #-}
 -- | Check if two 'ByteString's are equal.
 equalsByteString :: BuiltinByteString -> BuiltinByteString -> Bool
-equalsByteString x y = fromBuiltin (BI.equalsByteString x y)
+equalsByteString x y = fromOpaque (BI.equalsByteString x y)
```
Unisay, line 196: I have to admit that I thought I understood the difference between `HasBuiltin` and `HasOpaque` but in this case I have no idea what to use and if `fromOpaque` should be used instead of the `fromBuiltin` 😕 

If "opaque" types are not-pattern-matchable does it mean that we convert from them using `fromOpaque` with the sole intention to then pattern match?

bezirg: I am also confused, but from what I understood is that

`from/toOpaque` is supposed to be sth like Coercible but for **coercing** (0-cost) between Haskell and Tx universe (only that part that coincides),

whereas `from/toBuiltin` is for **converting** (thus, potentially non-zero cost) from a Haskell ADT (e.g. List) to a Tx builtin (primitive) type (e.g. BuiltinList). A List adt is a container of any `a`, whereas a BuiltinList can only contain `forall a in uni`.

bezirg: Actually I take it back I am now more confused.

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
effectfully: That part was authored by @michaelpj. I assume because pattern matching is more ergonomic than calling builtins manually, although we could have pattern matching synonyms, although those would be slower and in a very obscure-to-the-user way.

effectfully: Ah, well, you also have all the library functions for transparent types, like `map`, `foldr`, `concat` etc for lists, none of the is available for `BuiltinList`.

plutus-tx/src/PlutusTx/Builtins.hs
effectfully: > I have to admit that I thought I understood the difference between HasBuiltin and HasOpaque but in this case I have no idea what to use and if fromOpaque should be used instead of the fromBuiltin 😕

The Note literally instructs you to use `fromOpaque` in that case:

> This is what the 'HasToOpaque' and 'HasFromOpaque' classes do. They let us write wrappers for
> builtins relatively consistently by just calling 'toOpaque' on their arguments and 'fromOpaque' on
> the result.

> If "opaque" types are not-pattern-matchable does it mean that we convert from them using fromOpaque with the sole intention to then pattern match?

With the intention to be able to do with them anything you can do with a regular Plutus Tx data type, in particular pattern matching and using library functions.

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
effectfully: I've added such a comment just in case.

effectfully:
I've reviewed all discussions that we had about this in Slack channels and here, tweaked the docs a bit and added the following paragraph to a Note:

```
Summarizing, 'toBuiltin'/'fromBuiltin' should be used to cross the boundary between Plutus Tx and
Haskell, while 'toOpaque'/'fromOpaque' should be used within Plutus Tx to convert values to/from
their @Builtin*@ representation, which we need because neither pattern matching nor standard library
functions are available for values of @Builtin*@ types that we get from built-in functions.
```

I don't think I'm able to express the idea of this PR any better, so I'm going to hit merge the moment CI is green, given that polishing the docs for these Plutus Tx internals isn't a particularly high-priority task.

PR #5972, opened 2024-05-10
All names are printed with their unique values by default.

### LINKED ISSUE #4808: Pretty-printing uniques explicitly

Unisay:
- [x] A test demonstrating that without a unique suffix a name doesn't survive printing/parsing.
- [x] Update printer and parser to support unique suffixes, e.g. `foo#42`.

Closes #4808


plutus-core/untyped-plutus-core/test/Evaluation/Machines/Budget/IfThenElse/0.uplc.golden
```diff
@@ -1,47 +1,47 @@
 ( (Right (delay
-            (\z
-              f ->
-               f
+            (\z#0
+              f#1 ->
+               f#1
                  (delay
-                    (\z
```
Unisay, line 6: This example demonstrates that without unique suffixes variables `f` and `z` were shadowed multiple times.

plutus-core/plutus-core/src/PlutusCore/Pretty/ConfigName.hs
```diff
@@ -1,125 +1,36 @@
...
-    toPrettyConfigName = coerce
+  toPrettyConfigName = coerce
 
 -- | The 'PrettyConfigName' used by default: don't print 'Unique's.
 defPrettyConfigName :: PrettyConfigName
-defPrettyConfigName = PrettyConfigName False
+defPrettyConfigName = PrettyConfigName True
 
 -- | The 'PrettyConfigName' used for debugging: print 'Unique's.
 debugPrettyConfigName :: PrettyConfigName
```
effectfully, line 36: This should get flipped as well then (otherwise we end up having two identical configs and not a single other one for pretty-printing names), so that one can pretty-print stuff without uniques. The name should be changed too as it's no longer relevant to any debugging.

plutus-core/plutus-core/test/Names/Spec.hs
```diff
@@ -177,6 +181,25 @@ test_rebindCapturedVariable = testCase "rebindCapturedVariable" do
...
+    "name_print_parse_roundtrip"
+    prop_printing_parsing_roundtrip
+
+prop_printing_parsing_roundtrip :: Property
+prop_printing_parsing_roundtrip = property $ generalizeT do
+  name <- forAllPretty $ runAstGen genName
+  tripping name printName parseName
+  where
+    printName :: Name -> String
+    printName = render . pretty
```
effectfully, line 197: `render . pretty` is `display`.

plutus-core/plutus-core/test/Pretty/Golden/Readable/Examples/Data/exampleData.golden
```diff
@@ -1,7 +1,7 @@
-\(x : unit)
- (x : unit)
- (x : unit)
- (x : unit) ->
+\(x#0 : unit)
+ (x#0 : unit)
+ (x#0 : unit)
+ (x#0 : unit) ->
```
effectfully, line 4: I wonder what's that about, but we'll never know, because apparently it was me who added it.

plutus-core/plutus-core/test/Pretty/Golden/Readable/Examples/InterList/InterCons.golden
```diff
@@ -1,39 +1,47 @@
...
-                   (\a b -> rec (\(dat :: * -> * -> *) -> dat a b))))
-           (\(dat :: * -> * -> *) -> dat a b))
-        b
-        a) ->
+           (\(rec#6 :: ((* -> * -> *) -> *) -> *)
+             (spine#7 :: (* -> * -> *) -> *) ->
+              spine#7
+                ((\(interlist#2 :: * -> * -> *) a#0 b#1 ->
+                    all r#3.
+                      r#3 -> (a#0 -> b#1 -> interlist#2 b#1 a#0 -> r#3) -> r#3)
```
effectfully, line 12: I feel like `#` is too noisy, I think `` ` `` would be perfect here, compare this

```
r#3 -> (a#0 -> b#1 -> interlist#2 b#1 a#0 -> r#3) -> r#3)
```

and this

```
r`3 -> (a`0 -> b`1 -> interlist`2 b`1 a`0 -> r`3) -> r`3)
```

Now we of course already use `` ` `` for quoted identifiers, but we could use something else, I think `@` might work, given that it's basic Haddock syntax and GHC uses the symbol for type applications which we render differently, so we won't clash with GHC here (still need escaping for non-GHC-produced code, but that's a given).

effectfully (commented):
I'd prefer to use `` ` `` instead of `#` and replace our current usage of `` ` `` with `@`, I think that would make dumped UPLC _much_ more readable, but maybe it's just me.

We need proper escaping of delimiters, but I don't think it should block this PR, I think it's best to keep this one minimal (it's only 714 files changed after all) and implement escaping separately.

plutus-core/plutus-ir/test/PlutusIR/Parser/Tests.hs
```diff
@@ -53,24 +53,30 @@ separator :: Char -> Bool
 separator c = c `elem` separators || isSpace c
 
 aroundSeparators :: MonadGen m => m String -> String -> m String
-aroundSeparators = go False
+aroundSeparators = go False False
```
effectfully, line 56: I made a strategic decision not to review this test, sorry.

plutus-core/plutus-core/src/PlutusCore/DeBruijn/Internal.hs
```diff
@@ -164,15 +164,14 @@ instance Wrapped TyDeBruijn
 
 instance (HasPrettyConfigName config) => PrettyBy config NamedDeBruijn where
   prettyBy config (NamedDeBruijn txt (Index ix))
-    -- See Note [Pretty-printing names with uniques]
-    | showsUnique = pretty . toPrintedName $ txt <> "_i" <> render (pretty ix)
+    | showsUnique = pretty $ toPrintedName txt <> "#" <> render (pretty ix)
```
effectfully, line 167: Oh, I really do believe we shouldn't use `#` for both uniques and De Bruijn indices. With the former `x#1` and `x#0` cannot be the same variable, but with the latter they can. We can have e.g. `#i` or something for De Bruijn indices.

effectfully:
Thank you a ton for this PR, you're doing God's work.

plutus-core/plutus-core/src/PlutusCore/Parser/ParserCommon.hs
```diff
@@ -75,22 +59,22 @@ whenVersion p act = do
...
-  Parser a ->
-  String ->
-  T.Text ->
-  m a
+parse
+  :: (AsParserErrorBundle e, MonadError e m, MonadQuote m)
+  => Parser a
+  -> String
+  -> Text
+  -> m a
```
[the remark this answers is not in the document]
Unisay: Yes, this is controlled by the `function-arrows` option in fourmolu:
image

plutus-core/plutus-core/src/PlutusCore/Pretty/ConfigName.hs
```diff
@@ -1,125 +1,36 @@
...
+  toPrettyConfigName :: config -> PrettyConfigName
 
 instance HasPrettyConfigName (Sole PrettyConfigName) where
-    toPrettyConfigName = coerce
+  toPrettyConfigName = coerce
 
 -- | The 'PrettyConfigName' used by default: don't print 'Unique's.
 defPrettyConfigName :: PrettyConfigName
-defPrettyConfigName = PrettyConfigName False
+defPrettyConfigName = PrettyConfigName True
```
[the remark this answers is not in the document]
Unisay: I've just realized the comment above needs to be changed. A second thought "if we rename the option itself to something like `appendUniquesToNames` then we won't need a comment explaining its meaning, its going to be self-explanatory." Wdyt?

plutus-core/plutus-core/src/PlutusCore/Pretty/ConfigName.hs
Unisay: 👌🏼

plutus-core/plutus-core/test/Pretty/Golden/Readable/Examples/InterList/InterCons.golden
Unisay: Indeed, `` ` `` is lighter which could be more aesthetically pleasing, depending on your taste. I wouldn't mind us using `` ` ``, however these reasons make me think its not the best option:
1. More often than not back-ticks are used in pairs; When singular back-ticks happen to match they can hurt readability. Consider ``expected`0 risk`100`` -- it reads as if `0 risk` is spliced inside `expected100`
2. As you've mentioned: we already use back-ticks for escaping.

Between `` ` `` and `@` I chose the latter.

Lets also consider these characters:
```
ghci> import PlutusCore.Name.Unique
ghci> isIdentifierChar '~'
False
ghci> isIdentifierChar '!'
False
ghci> isIdentifierChar '-'
False
ghci> isIdentifierChar '^'
False
```

Unisay: Demo:
```
r-3 -> (a-0 -> b-1 -> interlist-2 b-1 a-0 -> r-3) -> r-3)
```
```
r^3 -> (a^0 -> b^1 -> interlist^2 b^1 a^0 -> r^3) -> r^3)
```
```
r~3 -> (a~0 -> b~1 -> interlist~2 b~1 a~0 -> r~3) -> r~3)
```
```
r!3 -> (a!0 -> b!1 -> interlist!2 b!1 a!0 -> r!3) -> r!3)
```
```
r|3 -> (a|0 -> b|1 -> interlist|2 b|1 a|0 -> r|3) -> r|3)
```

Unisay: Personally, I like `!` most of all above.

plutus-core/plutus-core/src/PlutusCore/Pretty/ConfigName.hs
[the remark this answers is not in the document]
effectfully: Yeah, but what if we add more stuff to that config. Anyways, either option is fine, pick your poison.

plutus-core/plutus-core/test/Pretty/Golden/Readable/Examples/InterList/InterCons.golden
effectfully: > More often than not back-ticks are used in pairs; When singular back-ticks happen to match they can hurt readability. Consider ``expected`0 risk`100`` -- it reads as if `0 risk` is spliced inside `expected100`

Yeah, good point.

I just took a long look at the demo. I'm OK with any of those, but `-` seems very nice as well as `!` indeed (except I'd perhaps prefer the former over the latter).

Unisay: 👍🏼 I went with `-` for `Unique` indexes and `!` for De Bruijn (instead of `#` for all)

plutus-core/plutus-core/src/PlutusCore/Pretty/Utils.hs
```diff
@@ -19,4 +19,4 @@ asBytes x = Text 2 $ T.pack $ addLeadingZero $ showHex x mempty
               | otherwise = id
 
 prettyBytes :: BS.ByteString -> Doc ann
-prettyBytes b = "#" <> fold (asBytes <$> BS.unpack b)
+prettyBytes b = "#" <> foldMap asBytes (BS.unpack b)
```
effectfully, line 22: :+1:

plutus-core/plutus-ir/test/PlutusIR/Transform/Rename/Tests.hs
```diff
@@ -13,18 +13,26 @@ import PlutusIR.Transform.Rename ()
...
+        [ "allShadowedDataNonRec"
+        , "allShadowedDataRec"
+        , "paramShadowedDataNonRec"
+        , "paramShadowedDataRec"
+        ]
   where
-    debugConfig = PLC.PrettyConfigClassic PLC.debugPrettyConfigName False
+    runGoldenPir =
+      goldenPir
+        ( PLC.AttachPrettyConfig prettyConfig
```
effectfully, line 28: I think we probably don't need that here, it was needed to adjust the default and now we want the default behavior. Anyways, not important.

effectfully (approved):
LGTM, the only remaining parts are mine:

1. create an issue for escaping special characters
2. restore the parts of the removed Note that are still relevant

I'll do these tomorrow, for now let's ask others to review.

zliu41 (changes requested):
I don't agree that we should print uniques in the golden files by default.

Most of the time, the uniques don't matter. They not only add noise and make things harder to read, but also, sometimes when you make a small change, it can change all the uniques in the golden file, which makes the actual changes extremely hard to find.

I've encountered a few tests that can benefit from adding uniques to the output, but I think we should only add uniques for those tests.

bezirg:
I also disagree that it should be on by default. Don't we have some pretty debug mode for this kind of stuff? Then all the golden files could just use the debug pretty mode.

effectfully:
There are two different things here:

1. whether the default config should prescribe to print the uniques
2. whether the golden test machinery should print them by default

I very strongly believe that when it comes to 1, the current `master` is straight out buggy. Pretty-printing a program and parsing it back shouldn't give you a semantically different program, that's just ridiculous. If we want to, we may pretty-print only those uniques that disambiguate names while ignoring others, but I'm not a fan of that: GHC adds "uniques" by default when it produces Core and I've never found that to be of any annoyance and I've spent quite a lot of time staring at GHC Core. So not fixing `master` in that regard is a non-option in my opinion, the current situation actively confuses our users (the original issue to fix the semantics changing pretty-printing exists because Runtime Verification ran into it, got very confused and, when I explained what was going on, very unhappy).

As for the second option, I feel less strongly, but I do believe that we should dump uniques into golden files by default. Your golden test is too large to have uniques? Well, then it's too large to be read anyway. [Worst practices should be hard](https://www.haskellforall.com/2016/04/worst-practices-should-be-hard.html) and it's not like using the non-default config is that hard. So I do believe that we should generally disincentivize creation of large golden files and if we can at the same time get correct scoping, that's a double win.

bezirg:
> I very strongly believe that when it comes to 1, the current `master` is straight out buggy. Pretty-printing a program and parsing it back shouldn't give you a semantically different program, that's just ridiculous.

I see now your point and I think also have ran into this before. But when does this exactly happen? Is it when there are free variables in the program or when we have to introduce fresh variables during the compilation?

Unisay:
I agree with:
* Default pretty printing config should produce a representation that always parses back (roundtrip);
* There might be some cases where we explicitly opt-out of printing indexes, it should be possible;
* The usage of golden files should be minimised and having to make an extra step to opt-out of indexes could have a positive effect in this dimension (aka "[Worst practices should be hard](https://www.haskellforall.com/2016/04/worst-practices-should-be-hard.html)")
* The "fragility" concern is real: if only indexes have changed but program didn't, then we don't want to see such change in the diff. @effectfully suggested that `rename` pass addresses this.

zliu41:
I agree should fix 1. For 2, I really disagree - uniques in anything but the tiniest golden files are quite annoying. Added a let binding? Everything afterwards will have different uniques. Added a `rename` in one of the passes? Congratulations, every single unique in every single golden file will now be different. What benefit do we get for such trouble?

Unisay:
I am in the process of re-working this PR in order to have a cake and eat it too:
- Unique indexes should be printed by default;
- Golden tests explicitly opt-in to a simple representation that doesn't print unique indexes;

This entails quite some boring work: the change isn't hard but its a lot files to visit and decide. Stay tuned.

plutus-benchmark/nofib/test/9.6/clausify-F5.pir.golden
```diff
@@ -1,990 +1,1001 @@
...
-                   {all dead. Ordering}
-                   (/\dead -> EQ)
-                   (\(ds : a) (ds : List a) -> /\dead -> LT)
-                   {all dead. dead})
-              (\(ds : a) (ds : List a) ->
-                 /\dead ->
-                   List_match
-                     {a}
-                     ds
+((let
```
zliu41, line 1: Looks like this file now has an additional pair of parentheses (it's hard to tell whether that's the only difference, but seems so?), which is redundant

plutus-core/plutus-ir/test/PlutusIR/Transform/Inline/letNonPureMultiStrict.golden
```diff
@@ -1,11 +1,4 @@
...
-    (strict)
-    (vardecl y (con integer))
-    [
-      [ { (builtin trace) (con integer) } (con string "hello") ] (con integer 1)
-    ]
-  )
-  [ [ (builtin addInteger) y ] [ [ (builtin addInteger) y ] y ] ]
-)
\ No newline at end of file
+  !y-3 : integer = trace {integer} "hello" 1
```
zliu41, line 2: I don't think we need uniques in all these gold files. IMO there are only a handful of tests whose golden files need uniques for disambiguation.

zliu41 (changes requested):
There are a number of golden files that now have additional redundant parentheses which appears to be what makes the line count so large. See the comment above for an example.

effectfully:
> now have additional redundant parentheses which appears to be what makes the line count so large. 

I promised to take a look at it and got caught up with other stuff. Gonna do it now.

effectfully:
@Unisay you've changed the behavior of `prettyPirReadable` in `PlutusIR.Core.Instance.Pretty.Readable` by turning `botPrettyConfigReadable` into `topPrettyConfigReadable`. So making it 

```haskell
-- | Pretty-print something with the @PrettyConfigReadable@ config.
prettyPirReadable :: PrettyPir a => a -> Doc ann
prettyPirReadable = prettyBy (botPrettyConfigReadable prettyConfigName def)

-- | Pretty-print something with the @PrettyConfigReadableSimple@ config.
prettyPirReadableSimple :: PrettyPir a => a -> Doc ann
prettyPirReadableSimple = prettyBy (botPrettyConfigReadable prettyConfigNameSimple def)
```

will solve the extra parens issue. BTW, we don't have two sets of parens there, one of them are there because a `let`-expression is applied to a `let`-expression, i.e. `(let x = a in q) (let y = b in r)`, i.e. those parens are as expected. Only took me a few thousand pulled out hairs to figure out, phew.

zliu41 (approved):

PR #5973, opened 2024-05-13
Kwxm/mainnet script budgets

kwxm:
This is just a small PR to add an option to `analyse-script-events` to print out the CPU and memory costs of each script event.

### DIFF
--- a/plutus-ledger-api/exe/analyse-script-events/Main.hs
+++ b/plutus-ledger-api/exe/analyse-script-events/Main.hs
@@ -360,6 +360,8 @@ max_tx_ex_steps = 10_000_000_000
 max_tx_ex_mem :: Double
 max_tx_ex_mem = 14_000_000
 
+-- Print out the CPU and memory budgets of each script event.  These are the costs
+-- paid for by the submitters, not the actual costs consumed during execution.
 getBudgets :: EventAnalyser
 getBudgets _ctx _params ev =
   let printFractions d =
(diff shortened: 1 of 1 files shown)


kwxm:
Here's what the start of the output looks like.  It can be read directly into R (in the form of a data frame) for further analysis.
```
         cpu     cpuFraction             mem    memFraction
      512081024   0.05120810           1426968   0.10192629
      540867862   0.05408679           1488026   0.10628757
      957982513   0.09579825           2846016   0.20328686
      512081024   0.05120810           1426968   0.10192629
      540867862   0.05408679           1488026   0.10628757
      957982513   0.09579825           2846016   0.20328686
      523569632   0.05235696           1447790   0.10341357
      940684283   0.09406843           2805780   0.20041286
      523569632   0.05235696           1447790   0.10341357
      940684283   0.09406843           2805780   0.20041286
...
```

zliu41 (approved):
Are you sure `dataBudget` is the claimed budget, not the actual budget? If it is not the actual budget, then why do we care about it?

kwxm:
> Are you sure `dataBudget` is the claimed budget, not the actual budget? If it is not the actual budget, then why do we care about it?

I haven't checked the details in the script-dumping program, but I'm pretty sure that these are the claimed budgets (ie, the ones that the fees are based on).  For example, about 7% of the events have memory budgets that are exact multiples of 1,000,000, and I think it's highly unlikely (about one chance in a million!) that that would occur as an actual cost.

The basic reason that I did it this way was that I was in a hurry: it took 2½ hours just to look at the scripts, and to get the real costs we would have had to execute them all, which would have taken a lot longer. Arguably it's useful to know these numbers anyway, but I'm going to claim that the numbers which aren't obviously made up are probably the real ones, and those account for probably over 90% of the data produced by this so it's still giving us useful information.  I agree that it would be good to know the real costs as well and I may come back and do that later.  I've added a TODO for that.

PR #5974, opened 2024-05-14
[Build] Remove the doctests runner of 'prettyprinter-configurable'

effectfully:
The doctests runner of `prettyprinter-configurable` has been broken for years, so this PR removes it. If we want to run the doctests, we should use `cabal-docspec` anyway and the existing code doesn't help with it, so we can drop it as well.


Unisay (approved):
Sure, unused code is a liability.

PR #6015, opened 2024-05-16
Script to prepare binary executables for uploading to the github release page.

Unisay:
- [x] `./scripts/prepare-bins.sh`
- [x] Tweak RELEASE.adoc.

### DIFF
--- a/RELEASE.adoc
+++ b/RELEASE.adoc
@@ -75,19 +75,11 @@ This updates versions and version bounds, and assembles the changelogs.open a PR
 - Choose as git tag `x.y.z.0`
 - Choose as target the git commit hash which points to the release commit
 - Click `Generate release notes` to automatically fill in the details of what's changed
-- Create and attach pir&uplc executables to the release by running the following commands inside the repository where its HEAD is at the release commit:
-- 
-+
-[source,bash]
--------------
-nix build .#hydraJobs.x86_64-linux.musl64.ghc96.pir 
-cp ./result/bin/pir ./pir-x86_64-linux-ghc96
-
-nix build .#hydraJobs.x86_64-linux.musl64.ghc96.uplc
-cp ./result/bin/uplc ./uplc-x86_64-linux-ghc96
--------------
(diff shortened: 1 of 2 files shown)


RELEASE.adoc
```diff
@@ -75,19 +75,11 @@ This updates versions and version bounds, and assembles the changelogs.open a PR
...
 - Click `Publish release`.
-7. Open a PR in the https://github.com/IntersectMBO/cardano-haskell-packages[CHaP repository] for publishing the new version. Run `./scripts/add-from-github.sh "https://github.com/IntersectMBO/plutus" COMMIT-SHA LIST-OF-UPDATED-PACKAGES` (see https://github.com/IntersectMBO/cardano-haskell-packages#-from-github[the README on CHaP]). Example: https://github.com/IntersectMBO/cardano-haskell-packages/pull/394.
+7. Open a PR in the https://github.com/IntersectMBO/cardano-haskell-packages[CHaP repository] for publishing the new version. +
+If you are making PR from your own fork then don't forget to sync your fork with the upstream first. +
+Run `./scripts/add-from-github.sh "https://github.com/IntersectMBO/plutus" COMMIT-SHA LIST-OF-UPDATED-PACKAGES` (see https://github.com/IntersectMBO/cardano-haskell-packages#-from-github[the README on CHaP]). Example: https://github.com/IntersectMBO/cardano-haskell-packages/pull/764.
```
Unisay, line 82: I've sneakily changed the example PR to a newer one with a slightly better description.

zliu41 (approved):
:+1:

bezirg:
This would require that `upx` is already installed in the developer's environment (outside nix). In my case:

```
$upx
bash: upx: command not found
```

I think it would be better to do it inside nix itself. 
But this PR is already better than what we had befoer, so no objections to that.

Unisay:
> This would require that upx is already installed in the developer's environment (outside nix). 

I don't think so! In the shebang I used `nix-shell` to make sure the `upx` package is available, see `-p upx`.

bezirg:
> > This would require that upx is already installed in the developer's environment (outside nix).
> 
> I don't think so! In the shebang I used `nix-shell` to make sure the `upx` package is available, see `-p upx`.

Oups sorry, i didn't see the shebang :(

PR #6036, opened 2024-05-16
[Builtins] Polish handling of integral types

effectfully:
This

- adds support for throwing operational unlifting errors in the builtins machinery, which allows us to attach a message to an unlifting failure. Unfortunately this is still not enough to propagate proper error messages to the user, because they get erased by the evaluation machinery down the line, this is our 2018's legacy of the stupid idea to have raw evaluation failures that don't carry any content
- makes handling of integral types in `PlutusCore.Default.Universe` uniform (it was a mess before with a weird partial set of instances and morally the same instances being implemented differently)


effectfully:
> Unfortunately this is still not enough to propagate proper error messages to the user, because they get erased by the evaluation machinery down the line, this is our 2018's legacy of the stupid idea to have raw evaluation failures that don't carry any content

I'll either create a separate PR for it or open an issue (likely the latter).

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -339,90 +336,142 @@ instance TestTypesFromTheUniverseAreAllKnown DefaultUni
...
-        inline readKnownConstant term >>= oneShot \(i :: Integer) ->
-           case toIntegralSized i of
-               Just w8 -> pure w8
-               _       -> throwing_ _EvaluationFailure
+deriving via AsInteger Word instance
+    KnownTypeAst tyname DefaultUni Word
+deriving via AsInteger Word instance HasConstantIn DefaultUni term =>
+    MakeKnownIn DefaultUni term Word
+instance HasConstantIn DefaultUni term => ReadKnownIn DefaultUni term Word where
+    readKnown term = fromIntegral @Word64 @Word <$> readKnown term
```
bezirg, line 409: This I think should be inside the `#if WORD_SIZE_IN_BITS == 64  cpp macro` since you are assuming Word is 64bits?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -339,90 +337,142 @@ instance TestTypesFromTheUniverseAreAllKnown DefaultUni
...
+    MakeKnownIn DefaultUni term Word32
+deriving via AsInteger Word32 instance HasConstantIn DefaultUni term =>
+    ReadKnownIn DefaultUni term Word32
+
+deriving via AsInteger Word64 instance
+    KnownTypeAst tyname DefaultUni Word64
+deriving via AsInteger Word64 instance HasConstantIn DefaultUni term =>
+    MakeKnownIn DefaultUni term Word64
+deriving via AsInteger Word64 instance HasConstantIn DefaultUni term =>
+    ReadKnownIn DefaultUni term Word64
```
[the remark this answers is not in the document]
bezirg: Maybe it is better to have here only those types that we currently rely upon?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: It is, it's just that the diff in unwieldy.

Screenshot from 2024-05-30 13-22-32

bezirg: oups sry

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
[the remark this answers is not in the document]
effectfully: I want to have a consistent set. Why should we support `Word8` but not `Int8`? `Int64` but not `Word64`? Let's just add all of them and save the user from scratching their head upon getting an error message saying that `Int8` isn't support when they copied some code from a builtin using `Word8`.

Unisay (approved):

PR #6037, opened 2024-05-17
Update issue templates

zliu41:
Removed "security report" and made all fields optional.


Unisay (approved):
LGTM

.github/ISSUE_TEMPLATE/feature_request.yml
effectfully, line 1: Remove the `enhancement` label that this template adds?
Also, we discussed during the call that this probably should be more general than just a feature request. Something more general like "tell us what you think" or whatever.

zliu41: > Something more general like "tell us what you think" or whatever.

I think people can create a blank issue for that.

effectfully: It was rather non-obvious to me that the blank issue template existed. I know it's right on that page, but it's smaller than the templates, looks differently and there are links all over the place, so I did miss it a number of times before finally finding it hiding in plain sight. Dunno, maybe it's just me.

zliu41: Yeah but I don't see a way to remove the blank issue option, and having two options for blank issues is more confusing IMO

effectfully: Good point.

PR #6043, opened 2024-05-17
[Errors] polish evaluation errors

effectfully:
We have this distinction of "user" vs "internal" errors that made some sense when TPLC was what was evaluated, but now that it's UPLC, all of those are "user" errors. So I was going to remove the distinction entirely just to realize that it in fact still makes sense, it's just now "operational" vs "structural". This PR clarifies this.

I've also tweaked errors a bit while I was here, plus removed a bunch of unsafe functions from the CEK machine API (they were convenient when they were exposed from a single module, but now that we have several of them, it's no longer worth the trouble, plus we really should use some different approach to calling the CEK machine in the benchmarks, but that's not within the scope of this PR).


plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
```diff
@@ -66,7 +66,7 @@ compilePirProgramOrFail pirProgram = do
     & runExceptT
     >>= \case
       Left (er :: PIR.Error DefaultUni DefaultFun (Provenance ())) -> fail $ show er
-      Right p -> pure (void p)
+      Right p                                                      -> pure (void p)
```
effectfully, line 69: Seems like `stylish-haskell` conflicts with `fourmolu`, I promise I didn't add those spaces manually.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
```diff
@@ -218,8 +225,10 @@ test_FailingSucc =
                 apply () (builtin () $ Right FailingSucc) $
                     mkConstant @Integer @DefaultUni @DefaultFunExt () 0
         typeErrOrEvalExcOrRes :: Either _ (Either BuiltinErrorCall _) <-
-            -- Here we rely on 'typecheckAnd' lazily running the action after type checking the term.
-            traverse (try . evaluate) $ typecheckEvaluateCek def defaultBuiltinCostModelExt term
```
effectfully, line 222: Ironically enough, we were also relying on `typecheckAnd` running the action strictly. Yes, both lazily and strictly, just in different ways, and I got a failure when I made `typecheckEvaluateCek` lazier. I've changed this test so that it's only sensitive to laziness and not strictness.

plutus-core/untyped-plutus-core/test/Evaluation/Debug/ex3.golden
```diff
@@ -7,5 +7,5 @@ OldState: Computing NewState: Returning
 Driver is going to do a single step
 OldState: Returning NewState: Computing
 Driver is going to do a single step
-OldState: Computing NewState is Error: An error has occurred:  User error:
+OldState: Computing NewState is Error: An error has occurred:
```
effectfully, line 10: Removed some noise from errors messages.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Exception.hs
```diff
@@ -66,13 +66,30 @@ data MachineError fun
...
+incorrectly while an operational error may be entirely expected.
+
+In other words, operational errors are regular runtime errors and structural errors are \"runtime
+type errors\". Which means that evaluating an (erased) well-typed program should never produce a
+structural error, only an operational one. This creates a sort of \"runtime type system\" for UPLC
+and it would be great to stick to it and enforce in tests etc, but we currently don't. For example,
+a built-in function expecting a list but getting something else should throw a structural error,
+but currently it'll throw an operational one. This is something that we plan to improve upon in
+future.
+-}
```
effectfully, line 89: This is the only important part of this PR.

plutus-core/plutus-core/src/PlutusCore/TypeCheck.hs
```diff
@@ -36,9 +36,9 @@ import PlutusCore.Quote
 import PlutusCore.Rename
 import PlutusCore.TypeCheck.Internal
 
--- | The constraint for built-in types/functions are kind/type-checkable.
+-- | The constraint for built-in types\/functions are kind\/type-checkable.
```
effectfully, line 39: Here's how the previous version was rendered:

Screenshot from 2024-05-18 09-59-35

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Exception.hs
bezirg: How about `data RuntimeError = RuntimeEvalError | RuntimeTypeError` ?

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Exception.hs
```diff
@@ -66,13 +66,30 @@ data MachineError fun
...
--- | The type of errors (all of them) which can occur during evaluation
--- (some are used-caused, some are internal).
-data EvaluationError user internal
-    = InternalEvaluationError !internal
-      -- ^ Indicates bugs.
-    | UserEvaluationError !user
-      -- ^ Indicates user errors.
+{- | The type of errors that can occur during evaluation. There are two kinds of errors:
+
+1. Operational ones -- these are errors that are indicative of the _logic_ of the program being
```
bezirg, line 71: mention also 'exceeding the gas limit'

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Exception.hs
```diff
@@ -81,47 +98,53 @@ mtraverse makeClassyPrisms
...
 EvaluationFailure.  This means that, for example, if we use the `plc`
 command to execute a program containing a division by zero, plc exits
 silently without reporting that anything has gone wrong (but returning
 a non-zero exit code to the shell via `exitFailure`).  This is because
-UserEvaluationError is used in cases when a PLC program itself goes
+OperationalEvaluationError is used in cases when a PLC program itself goes
 wrong (for example, a failure due to `(error)`, a failure during
 builtin evaluation, or exceeding the gas limit).  This is used to
 signal unsuccessful in validation and so is not regarded as a real
 error; in contrast, machine errors, typechecking failures,
```
bezirg, line 127: > machine errors, typechecking failures,

structural evaluation errors

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Exception.hs
```diff
@@ -81,47 +98,53 @@ mtraverse makeClassyPrisms
...
 
-type EvaluationException user internal =
-    ErrorWithCause (EvaluationError user internal)
+type EvaluationException operational structural =
+    ErrorWithCause (EvaluationError operational structural)
 
-{- Note [Ignoring context in UserEvaluationError]
-The UserEvaluationError error has a term argument, but
+{- Note [Ignoring context in OperationalEvaluationError]
+The OperationalEvaluationError error has a term argument, but
```
bezirg, line 117: I get what is happening here, but what I don't understand is why we do all this effort to distinquish between operational vs structural errors at actual script validation, by using
`Either (ErrorWithCause structural term) (EvaluationResult a)` . 

From the user's point of view, validating a script would either EvaluationSuccess (validate) or EvaluationFailure (not validate). 

In what way does it matter if the non-validation was caused by operational or structural error?

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Exception.hs
```diff
@@ -148,13 +171,7 @@ instance (HasPrettyDefaults config ~ 'True, Pretty fun) =>
...
-        , prettyBy config err
-        ]
-    prettyBy _      (UserEvaluationError err) = fold
-        [ "User error:", hardline
-        , pretty err
-        ]
+        , Pretty operational, PrettyBy config structural
+        ) => PrettyBy config (EvaluationError operational structural) where
+    prettyBy _      (OperationalEvaluationError operational) = pretty operational
+    prettyBy config (StructuralEvaluationError structural)   = prettyBy config structural
```
bezirg, line 177: And indeed here you show that it does not matter if the error was operational or structural :P

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Common.hs
```diff
@@ -39,14 +41,12 @@ typecheckAnd
             UPLC.Term Name uni fun () -> a)
     -> CostingPart uni fun -> TPLC.Term TyName Name uni fun () -> m a
 typecheckAnd semvar action costingPart term = TPLC.runQuoteT $ do
-    -- Here we don't use `getDefTypeCheckConfig`, to cover the absurd case where
-    -- builtins can change their type according to their BuiltinSemanticsVariant
+    -- Here we don't use 'getDefTypeCheckConfig', to cover the absurd case where
+    -- builtins can change their type according to their 'BuiltinSemanticsVariant'.
```
bezirg, line 45: I do often markdown style backticks in comments. Should I stop doing that?

plutus-core/changelog.d/20240517_191403_effectfully_polish_evaluation_errors.md
```diff
@@ -0,0 +1,3 @@
+### Removed
+
+- `unsafeRunCekNoEmit` and all `unsafeEvaluate*` functions in #6043. To replace e.g. `unsafeEvaluateCek` you can use `evaluateCek` in combination with `unsafeToEvaluationResult`.
```
bezirg, line 3: Also Changed section , renamed unsafeExtractEvaluationResult   to unsafeToEvaluationResult

bezirg (approved):

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Exception.hs
effectfully: I'd like to keep it intentionally separate from the whole types realm, we're talking about something entirely different here. `RuntimeTypeError` suggests that it's something about types but

1. we don't have types in UPLC
2. a missing case branch isn't a type error by any measure, yet we consider it to be a structural error. Is that a good idea? Dunno, but that's what we do

So let's just introduce new terminology for this thing that doesn't seem to match any typery exactly.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Exception.hs
effectfully: It's explained in the Haddocks:

```
On the chain both of these are just regular failures and we don't distinguish between them there:
if a script fails, it fails, it doesn't matter what the reason was. However in the tests it does
matter why the failure occurred: a structural error may indicate that the test was written
incorrectly while an operational error may be entirely expected.
```

Basically, it's an extra level of safety for us. `unsafeToEvaluationResult` is used in quite a lot of places and some tests fail if you remove the distinction between operational and structural errors, in the same way that some tests will start to fail if you get typing wrong, even if operationally nothing changes. Just ensures that any evaluation failures are the expected operational ones, not somebody messing up the test.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Exception.hs
effectfully: It matters when `extractEvaluationResult` or `unsafeToEvaluationResult` are called and they do get called. I'm not sure if reporting whether the error was operational or structural would add meaningful context or just confuse the reader who may not be familiar with the distinction. I do want _us_ to be familiar with the distinction.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Common.hs
effectfully: Dunno, Haddock "prefers" `'`, so I use that everywhere in our codebase for consistency together with `@`, but I think Haddock is also fine with backticks as a replacement for `'` (but not `@`), so probably not a big deal. I'd prefer consistency, but it does sound like ensuring consistency would be more pain than just being inconsistent here.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Exception.hs
bezirg: If it is an artifact of Tests alone, then perhaps EvaluationResult should belong to a testlib.

Because taking into account other Plutus-based languages (Aiken,Plutarch,etc) there may not be a distinction between structural vs operational errors and instead they have just errors?

effectfully: > If it is an artifact of Tests alone, then perhaps EvaluationResult should belong to a testlib.

Maybe. I'm currently removing it from builtins where we have a much more appropriate `BuiltinResult`. I doubt I'm going to get motivated enough to see it through though.

> Because taking into account other Plutus-based languages (Aiken,Plutarch,etc) there may not be a distinction between structural vs operational errors and instead they have just errors?

Perhaps, but I don't really care, I just want us to have that distinction for the same reasons I want us to have a type checker instead of juggling raw `Data`.

effectfully: > I doubt I'm going to get motivated enough to see it through though.

I mean, removing `EvaluationResult` from all non-test code, not replace it with `BuiltinResult` in the builtins code -- that ship has sailed.

plutus-core/plutus-core/src/PlutusCore/Evaluation/ErrorWithCause.hs
```diff
@@ -37,11 +37,14 @@ instance (Pretty err, Pretty cause) => Pretty (ErrorWithCause err cause) where
...
 instance (PrettyBy config cause, PrettyBy config err) =>
             PrettyBy config (ErrorWithCause err cause) where
-    prettyBy config (ErrorWithCause err mayCause) =
-        "An error has occurred: " <+> prettyBy config err <>
-            case mayCause of
-                Nothing    -> mempty
-                Just cause -> hardline <> "Caused by:" <+> prettyBy config cause
+    prettyBy config (ErrorWithCause err mayCause) = fold
+        [ "An error has occurred:"
+        , hardline
```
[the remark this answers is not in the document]
Unisay: Will `hardline` be added twice in case of `Just cause`  ?

Unisay (approved):

plutus-core/plutus-core/src/PlutusCore/Evaluation/ErrorWithCause.hs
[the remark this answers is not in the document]
effectfully: Yes, but there will be pretty-printer `err` in-between, you can see how it gets rendered at the ~bottom of the diff. Anyways, I'm just preserving the old behavior newlines-wise.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Common.hs
kwxm: `find . -name "*.hs" -exec sed -i 's/`/\'/g'`

What could possibly go wrong?

PR #6061, opened 2024-05-20
Conditionally add 'triage' label to new issues

zeme-wana:
Add triage label only when the issue does not have any of the 'Internal' labels.

### DIFF
--- a/.github/workflows/add-triage-label.yml
+++ b/.github/workflows/add-triage-label.yml
@@ -43,9 +43,15 @@ jobs:
               });
             }
 
-            if (!await issueHasInternalLabels()) {
-              async addNeedsTriageLabelToIssue();
-            }
+            try {
+              if (!await issueHasInternalLabels()) {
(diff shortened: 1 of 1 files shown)


.github/workflows/add-triage-label.yml
```diff
@@ -13,10 +17,41 @@ jobs:
...
-          script: |
-            github.rest.issues.addLabels({
-              issue_number: context.issue.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              labels: ["status: needs triage"]
-            })
\ No newline at end of file
+          script: | 
+            const INTERNAL_LABELS = ["Internal"];
```
effectfully, line 21: Please add here `status: triaged` as well.

(it really should be `status: triaged` only, but since I can't convince folks to follow the process, we'll probably have to rework it, so let's keep `Internal` as well)

effectfully (approved):
No idea what the JS code does, so LGTM. Thank you!

PR #6063, opened 2024-05-20
Kwxm/multi cost model benchmark test

kwxm:
DON'T REVIEW THIS OR MERGE IT!

There's an imminent PR to update the cost modelling machinery to handle multiple cost models.  The default cost model used for testing/benchmarking is now the provisional Conway cost model, which updates the costs of many things.  I want to check some stuff using the old cost model with the new machinery though, and that's what this is for.


kwxm:
OK, that looks good.

PR #6079, opened 2024-05-22
Make it build with ghc 9.10

erikd:
~This is a WIP mainly intended to make the Plutus maintainers aware of the fact that I am working on this.~

~This cannot be merged until the (currently ~17~ 16) SRPs are fixed upstream or fully documented.~

~**The fact that the Nix evals are failing can be ignored for now.~
~The Nix failures are due to missing SHA hashes for the SRPs which need to be removed before this PR can be merged and are likely to be changing frequenty.**~

Ready for review. No Changelog Required.


plutus-core/plutus-core.cabal
```diff
@@ -286,7 +286,7 @@ library
     , bytestring
     , bytestring-strict-builder
     , cardano-crypto
-    , cardano-crypto-class        ^>=2.1.2
+    , cardano-crypto-class        ^>= 2.2
```
erikd, line 289: Any reason not to do this? Plutus seems to compile with this change.

erikd: Reason not to do this is because `cardano-ledger` does not compile with the new version of `cardano-crypto-class`.

erikd:
I have removed all the SRPs, but can not longer push to this branch.

erikd:
Please check the last two commits on this PR. One test was disabled and 3 golden output changes were accepted.

plutus-core/cost-model/budgeting-bench/Generators.hs
```diff
@@ -10,7 +11,9 @@ import Control.Monad
 import Data.Bits
 import Data.ByteString (ByteString)
 import Data.Int (Int64)
+#if !MIN_VERSION_base(4,20,0)
 import Data.List (foldl')
+#endif
```
effectfully, line 16: Any way we could do this without the CPP?

plutus-core/plutus-core/src/PlutusCore/Name/UniqueMap.hs
```diff
@@ -23,7 +24,9 @@ module PlutusCore.Name.UniqueMap (
 import Control.Lens (view)
 import Control.Lens.Getter ((^.))
 import Data.Coerce (Coercible, coerce)
-import Data.Foldable (foldl')
+#if ! MIN_VERSION_base(4,20,0)
+import Data.List (foldl')
+#endif
```
effectfully, line 29: Same question.

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -22,6 +23,10 @@ import Data.String (IsString (..))
 import Data.Text qualified as Text
 import GHC.Magic qualified as Magic
 import Prelude qualified as Haskell (String)
+#if MIN_VERSION_base(4,20,0)
+import Prelude (type (~))
+#endif
```
effectfully, line 28: God, why do they make us go through this.

effectfully (commented):
All the new `foldl'`-related CPP is pretty annoying, I'll take a look to see if we can avoid it. Maybe by pulling it into `PlutusPrelude`? I'll check it.

plutus-core/cost-model/budgeting-bench/Generators.hs
erikd: I removed it somewhere else so will do the same here and the other places.

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
erikd: I have removed all `CPP` that I added other than that one.

plutus-core/cost-model/budgeting-bench/Generators.hs
erikd: Fixed.

effectfully (approved):
LGTM

PR #6080, opened 2024-05-22
Remove obsolete defer-plugin-errors flag

zeme-wana:
This flag is obsolete and replaced by the `defer-errors` flag


bezirg (approved):

Unisay:
As this is a user-facing change I'd expect it to be reflected in the changelog.

Issue #6054, opened 2024-05-17
Add versioning to metatheory and add formalization of the all versions of the language and its builtins


bezirg:
Similar to what we do in plutus-core with builtinsemanticsvariant.

bezirg:
This is not yet implemented, right ? @ana-pantilie @ramsay-t

kwxm:
Closed in favour of #7881.

PR #6087, opened 2024-05-22
Kwxm/provisional conway cost model

kwxm:
This  PR introduces an updated cost model for PlutusV3, which we expect to become available in the Chang hard fork at the start of the Conway era.  The coefficients of the costing functions (the "cost model parameters") are provisional and may be subject to later adjustment.  The new cost model currently gives script costs which are 20-30% cheaper than the current cost model, but given the provisional nature of the cost model this cost reduction should not be relied upon.

Some remarks:
  * I've renamed DefaultSemanticsVariant0/1/2` to `DefaultSemanticsVariantA/B/C` in an attempt to avoid confusion with `PlutusV1/V2/V3` etc.  Each of these has a corresponding cost model.
  * I'm not especially happy with `ExBudgetingDefaults.hs`.  This exports a lot of functions, most of them only used for testing, and it's a bit messy.
  * Related to the previous point, we have to decide how what to do with budget tests, for example, in `plutus-conformance`.  Do we want to run all of the tests with all of the semantic variants, which would potentially require a separate golden file for each test?  For the time being I'm just using the latest cost mode for each test (apologies for all of the updated golden files :( ). I've opened a separate issue for this at #6088.
  * Many of the tests in `plutus-ledger-api` (for example, [here](https://github.com/IntersectMBO/plutus/blob/master/plutus-ledger-api/testlib/PlutusLedgerApi/Test/V1/EvaluationContext.hs) and [here](https://github.com/IntersectMBO/plutus/blob/master/plutus-ledger-api/test/Spec/CostModelParams.hs) are now pretty much meaningless.  I've patched them so that they work, but I think that all of this stuff will have to be reworked.  I've opened an issue for this at #6089.
  * We also need to decide what to do with the metatheory.  This PR makes it use the latest cost model exclusively, but we may want to rethink that: see #6054.
   * I fixed `uplc` and `pir` to use the appropriate cost model for a given semantics variant, but I've just realised that I forgot to do that for the new `plutus` executable.


plutus-benchmark/bls12-381-costs/test/9.6/bls12-381-costs.golden
```diff
@@ -2,144 +2,144 @@ Hash n bytestrings onto G1 and add points
...
-   70     884   (5.4%)      4749685419  (47.5%)          237438   (1.7%) 
-   80     994   (6.1%)      5425741309  (54.3%)          270798   (1.9%) 
-   90    1104   (6.7%)      6101797199  (61.0%)          304158   (2.2%) 
-  100    1214   (7.4%)      6777853089  (67.8%)          337518   (2.4%) 
-  110    1324   (8.1%)      7453908979  (74.5%)          370878   (2.6%) 
-  120    1434   (8.8%)      8129964869  (81.3%)          404238   (2.9%) 
-  130    1544   (9.4%)      8806020759  (88.1%)          437598   (3.1%) 
-  140    1654  (10.1%)      9482076649  (94.8%)          470958   (3.4%) 
-  150    1764  (10.8%)     10158132539 (101.6%)          504318   (3.6%) 
+    0     115   (0.7%)        53556222   (0.5%)            3918   (0.0%)
```
zliu41, line 5: What happened to this?

plutus-core/changelog.d/20240522_211822_kenneth.mackenzie_provisional_conway_cost_model.md
```diff
@@ -0,0 +1,12 @@
+<!--
```
zliu41, line 1: This should be uncommented

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -72,8 +72,8 @@ builtinMemoryModels = BuiltinCostModelBase
   , paramMultiplyInteger                 = Id $ ModelTwoArgumentsAddedSizes $ OneVariableLinearFunction 0 1
   , paramDivideInteger                   = Id $ ModelTwoArgumentsSubtractedSizes $ ModelSubtractedSizes 0 1 1
   , paramQuotientInteger                 = Id $ ModelTwoArgumentsSubtractedSizes $ ModelSubtractedSizes 0 1 1
-  , paramRemainderInteger                = Id $ ModelTwoArgumentsSubtractedSizes $ ModelSubtractedSizes 0 1 1
-  , paramModInteger                      = Id $ ModelTwoArgumentsSubtractedSizes $ ModelSubtractedSizes 0 1 1
+  , paramRemainderInteger                = Id $ ModelTwoArgumentsLinearInY $ OneVariableLinearFunction 0 1
```
zliu41, line 75: Does this apply to PlutusV3 only, or all versions? I'd expect it to be V3 specific but there's nothing in this module specific to V3.

plutus-core/cost-model/data/builtinCostModelB.json
```diff
@@ -0,0 +1,919 @@
...
+        },
+        "memory": {
+            "arguments": 4,
+            "type": "constant_cost"
+        }
+    },
+    "integerToByteString": {
+        "cpu": {
+            "arguments": {
+                "c0": 9223372036854775807,
```
zliu41, line 488: As discussed, `integerToByteString` and `byteStringToInteger` should have normal parameters.

plutus-ledger-api/test/Spec/CostModelParams.hs
```diff
@@ -29,11 +27,8 @@ tests =
     "costModelParams"
     [ testCase "length" $ do
             166 @=? length v1_ParamNames
-            166 @=? length V1.costModelParamsForTesting
```
zliu41, line 32: Is this no longer relevant? If so, why?

kwxm: Those tests don't really make sense any more.  This is the stuff that needs to be reworked in #6089.

plutus-core/cost-model/data/builtinCostModelB.json
kwxm: I'm not sure about that because this model will apply from PV9 but those builtins won't be available until PV10.

plutus-benchmark/bls12-381-costs/test/9.6/bls12-381-costs.golden
kwxm: > What happened to this?

Do you mean why did the number for 0 get bigger?  If so, it's probably because that's a case where it's processing an empty list.  Some of the builtins have got  cheaper, but others have got more expensive, at least for certain inputs.  My guess is that there's some extra constant cost that gets swallowed up when it actually uses the BLS operations, which _have_ got cheaper.

plutus-core/cost-model/data/builtinCostModelB.json
kwxm: > I'm not sure about that because this model will apply from PV9 but those builtins won't be available until PV10.

To expand on that, the actual numbers in the file don't matter because the node will overwrite by the protocol parameters (which in fact we'll probably provide, based on the contents of this file).  In PV9 these builtins won't be available but in PV10 we will have real values for these parameters.  As I mentioned earlier we might strictly want four cost models, with one (based on this one) for PV9 and an updated version for PV10 (although it's probably safe to have sensible numbers here because they won't be used anyway).  We'll maybe be adding more builtins for PV10, so we'll have to update this for that anyway.

zliu41: > which in fact we'll probably provide, based on the contents of this file

That would be a problem. When we provide the parameters, all parameters should be normal (non-max). Otherwise, if we enable a builtin at PV10, then before the PV9 -> PV10 transition, its parameters must be updated from max to normal, which is just an unnecessary burden.

Also, note that in the Conway era, it may not even be possible to update a parameter exactly at the same time of a HF, due to new governance mechanisms. If so, we'll have to update the parameters before the HF. But if we do that, we may as well just provide the normal parameters in the first place!

zliu41 (changes requested):
I think the use of maxBound parameters in JSON as a way to prevent certain builtins from being used is more hindering than helpful:
- It's not helpful because script deserialization already takes care of preventing unavailable builins from being used. And this the status quo - we are relying solely on deserialization right now.
- If we actually use maxBound parameter values on-chain, then we must change them to normal values when the corresponding builtin becomes available, which is a non-trivial and unnecessary thing to do.

I would thus suggest replacing all maxBound parameter values in JSON with regular values.

plutus-core/cost-model/data/builtinCostModelB.json
kwxm: > That would be a problem. When we provide the parameters, all parameters should be normal (non-max).

I thought we wouldn't provide those parameters until the buitlins were actually available, so the PV9 parameters for PlutusV1/2 wouldn't contain any entries for `integerToByteString/byteStringToInteger` but the PV10 ones would.  Maybe you're right though: the PV9 protocol parameters could contain cost model parameters for those bultins but they'd never be used in PV9 because the builtins are disabled.  I suppose we might need to recalibrate the cost model in between PV9 and PV10 though, in which case we'd need to update the parameters anyway.

The procedure for updating the cost model parameters is a bit informal: we just come up with some numbers and occasionally we provide updated values to the node people for incorporation in the protocol parameters; in some PV updates nothing changes but in others the cost model parameters really do change.  Maybe we need some more formal process for doing this.

The reason that I put in the maximum numbers is that (at least with the way we currently decode the cost model from JSON) the JSON files _have_ to contain entries for all of the builtins, even for ones which aren't available during the relevant PV.  Those numbers are really supposed to serve as a `Nothing` value, and also have the advantage of making the builtins unusable for our off-chain tools and tests, where code doesn't always go trough the deserialisation process that checks for invalid builtins.  I think I said this somewhere else, but maybe a better way to do that would be to  serialise and deserialise with the appropriate PV/LL combination when we're using code that's read in from textual source or compiled from Haskell. That might be something to look into in #6088.

kwxm:
Those benchmark results are a bit strange: I tried this on a branch which was largely identical to this one a few days ago and got no change at all: see #6063.

plutus-core/changelog.d/20240522_211822_kenneth.mackenzie_provisional_conway_cost_model.md
kwxm: Oops!  Fixed.

plutus-core/cost-model/data/builtinCostModelB.json
kwxm: Also, I really wish you could put comments in JSON files to explain what's going on.

kwxm:
> * If we actually use maxBound parameter values on-chain, then we must change them to normal values when the corresponding builtin becomes available, which is a non-trivial and unnecessary thing to do.

Could you expand on that?  The JSON files, and hence the cost model objects in Plutus, have to contain entries for all of the builtins, for all LL/PV combinations, and when the node calls `applyCostModelParameters` to update them, some (but not necessarily all) of them will be overwritten by protocol parameters.  I think we could do at least two things:
  * The PV9 protocol parameters for PlutusV1/V2 don't contain cost model parameters for `integerToByteString` and `byteStringToInteger`, so whatever values the JSON file contains for those functions don't get updated. The PV10 protocol parameters will contain suitable cost model parameters which will be used to update the Plutus Cost model
  * Both the PV9 and PV10 protocol parameters for PlutusV1/V2 contain sensible values (possibly the same for both PVs) and in both cases the Plutus Core cost model will be updated with those values.
 
Whichever of these we do you won't be able to use the new builtins in PV9.  Are you proposing that we do the second thing?  That would mean that the PV9 protocol parameters contain cost model parameters for builtins that you can't use.  That would be simpler from an administrative point of view but might be misleading to people looking at the protocol parameters.

plutus-core/executables/uplc/Main.hs
```diff
@@ -322,7 +325,7 @@ runEval (EvalOptions inp ifmt printMode budgetMode traceMode
     let term = void $ prog ^. UPLC.progTerm
         cekparams = case cekModel of
                     -- AST nodes are charged according to the default cost model
-                    Default -> mkMachineParameters semvar PLC.defaultCekCostModel
+                    Default -> PLC.defaultCekParametersForVariant semvar
```
effectfully, line 328: And here.

plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
```diff
@@ -109,8 +109,8 @@ throw an error instead of overflowing its first argument.
 One denotation from each builtin is grouped into a 'BuiltinSemanticsVariant'.
 Each Plutus Language version is linked to a specific 'BuiltinSemanticsVariant'
 (done by plutus-ledger-api); e.g. plutus-v1 and plutus-v2 are linked to
-'DefaultFunSemanticsVariant1', whereas plutus-v3 changes the set of denotations
-to 'DefaultFunSemanticsVariant2' (thus fixing 'ConsByteString').
+'DefaultFunSemanticsVariantA', whereas plutus-v3 changes the set of denotations
+to 'DefaultFunSemanticsVariantC' (thus fixing 'ConsByteString').
```
effectfully, line 113: The very first one should be `DefaultFunSemanticsVariant1` and `DefaultFunSemanticsVariantB` rather than just the former?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1840,7 +1840,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
     -}
 
 instance Default (BuiltinSemanticsVariant DefaultFun) where
-    def = DefaultFunSemanticsVariant2
+    def = DefaultFunSemanticsVariantC
```
effectfully, line 1843: Could've just probably made it `def = maxBound` BTW.

effectfully (commented):
The first part of my review. Lots of "and here" -- these are just so that we can later go through them and check if we want to use the right bundle of machine parameters there.

Looks good to me so far.

zliu41:
> Are you proposing that we should do the second thing?

Yes, but it's less about what we put in the JSON files, than what the node actually uses as protocol parameters. What I'm saying is: once we are in PV9, the actual protocol parameters should simply contain sensible values for `byteStringToInteger`. There is no need to use maxBound for actual protocol parameters, because it serves no purpose, and we'll still have to change them to sensible values before we get to PV10.

Regarding JSON: if the actual protocol parameters should have sensible values, why not make the JSON the same - i.e., also with sensible values.

plutus-ledger-api/changelog.d/20240522_212027_kenneth.mackenzie_provisional_conway_cost_model.md
```diff
@@ -0,0 +1,8 @@
+### Added
+
+- A new cost model for PlutusV3.
+
+### Changed
+
+- We now have configurable cost models which allow different costs for different Plutus language versions and protocol versions.
+- `plutus-ledger-api` now provides version-dependent `mkEvaluationContext` functions for on-chain configuration of the Plutus Core evaluator.
```
effectfully, line 8: Wait, but the type of `mkEvaluationContext` changed neither in your PR nor in my earlier one? I think last time it was changed (aside from `[Integer]` -> `[Int64]`) was in 2022? We used to have three different `mkEvaluationContext` (one per ledger language) and we still have them.

effectfully (approved):
Quite a Herculean effort, thank you for doing it. LGTM.

Comments are mostly me being confused or pointing out places where we probably shouldn't have `defaultCekParametersForTesting`.

Now that I see in how many legitimate places we have `defaultCekParametersForTesting`, I realize that I probably shouldn't have made it twice as slow, that's just wasteful and probably makes our tests a lot slower. I guess we should simply revert that change and hope that things don't change much so that we keep having good performance.

Or we could drop `defaultCekParametersForTesting` entirely and insist that `mkMachineParametersFor` has to be used. That would be quite a nuclear option, but it would much more reliable.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetingDefaults.hs
```diff
@@ -236,3 +322,9 @@ unitCostBuiltinCostModel = BuiltinCostModelBase
     , paramByteStringToInteger             = unitCostTwoArguments
     }
 
+unitCekParameters :: Typeable ann => MachineParameters CekMachineCosts DefaultFun (CekValue DefaultUni DefaultFun ann)
```
bezirg, line 325: Since this is only by the uplc executable, I would suggest to move it there (if it is possible)

plutus-core/plutus-core/test/CostModelInterface/defaultCostModelParams.json
```diff
@@ -160,7 +160,6 @@
   "modInteger-cpu-arguments-model-arguments-intercept": 453240,
   "modInteger-cpu-arguments-model-arguments-slope": 220,
   "modInteger-memory-arguments-intercept": 0,
-  "modInteger-memory-arguments-minimum": 1,
```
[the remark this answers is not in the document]
bezirg: I am also confused. I actually thought that we are going to have 3 separate versions of this file.

bezirg (changes requested):
Shouldn't we have 3 versions of defaultCostModelParams.json?

bezirg:
Ah I think I understand now, the defaultCostModelParams.json reflects only the latest default semvar C. And it is only used for testing the json param-updating functionality.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
bezirg: Y I start liking Bounded more than Default

plutus-core/plutus-core/test/CostModelInterface/defaultCostModelParams.json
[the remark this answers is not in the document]
bezirg: I guess the memory-minimum was removed in sem.var C for modInteger and remainderInteger **intentionally**?

bezirg (approved):
Shouldn't we have 3 versions of defaultCostModelParams.json?

plutus-core/plutus-core/test/CostModelInterface/defaultCostModelParams.json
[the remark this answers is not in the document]
kwxm: This should all be reworked in #6088 and/or #6089 (soon, I hope).

plutus-benchmark/common/PlutusBenchmark/Common.hs
```diff
@@ -118,7 +118,7 @@ unsafeRunTermCek :: Term -> EvaluationResult Term
 unsafeRunTermCek =
     unsafeExtractEvaluationResult
         . (\(res, _, _) -> res)
-        . runCekDeBruijn PLC.defaultCekParameters Cek.restrictingEnormous Cek.noEmitter
+        . runCekDeBruijn PLC.defaultCekParametersForTesting Cek.restrictingEnormous Cek.noEmitter
```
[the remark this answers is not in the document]
kwxm: We might need one version of this for every cost model, so the name(s) might become less helpful later.

plutus-benchmark/nofib/exe/Main.hs
```diff
@@ -202,7 +202,10 @@ options = hsubparser
 ---------------- Evaluation ----------------
 
 evaluateWithCek :: UPLC.Term UPLC.NamedDeBruijn DefaultUni DefaultFun () -> UPLC.EvaluationResult (UPLC.Term UPLC.NamedDeBruijn DefaultUni DefaultFun ())
-evaluateWithCek = UPLC.unsafeExtractEvaluationResult . (\(fstT,_,_) -> fstT) . UPLC.runCekDeBruijn PLC.defaultCekParameters UPLC.restrictingEnormous UPLC.noEmitter
+evaluateWithCek =
+  UPLC.unsafeExtractEvaluationResult
+  . (\(fstT,_,_) -> fstT)
+  . UPLC.runCekDeBruijn PLC.defaultCekParametersForTesting UPLC.restrictingEnormous UPLC.noEmitter
```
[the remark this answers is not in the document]
kwxm: The executable is more about extracting the UPLC and getting budgets and so on, but I'll change it just to be safe. Maybe eventually it should have a command-line option to select the Plutus version and the cost model to use.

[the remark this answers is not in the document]
kwxm: I'll fix it though, just to be on the safe side.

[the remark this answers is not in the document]
kwxm: > I'll fix it though, just to be on the safe side.

Actually I won't.  Let's do that in #6088.

plutus-core/executables/uplc/Main.hs
kwxm: I think that one's actually OK: there's a command-line option to tell it which semantic variant to use.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
kwxm: > Y I start liking Bounded more than Default

Good idea.  There are a lot of `def`s in various places and it can be quite hard to tell exactly what the default is.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetingDefaults.hs
```diff
@@ -83,53 +86,136 @@ defaultCekMachineCosts =
...
     -> BuiltinsRuntime DefaultFun term
 -- See Note [noinline for saving on ticks].
-defaultBuiltinsRuntimeForSemanticsVariant semvar = noinline toBuiltinsRuntime semvar defaultBuiltinCostModel
+defaultBuiltinsRuntimeForSemanticsVariant = \case
+  DefaultFunSemanticsVariantA ->
+    noinline toBuiltinsRuntime DefaultFunSemanticsVariantA builtinCostModelVariantA
+  DefaultFunSemanticsVariantB ->
+    noinline toBuiltinsRuntime DefaultFunSemanticsVariantB builtinCostModelVariantB
+  DefaultFunSemanticsVariantC ->
+    noinline toBuiltinsRuntime DefaultFunSemanticsVariantC builtinCostModelVariantC
```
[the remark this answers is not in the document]
kwxm: Hmmm. I think I was a bit nervous about disturbing the `noinline` when I did that.  I've moved the cost model variant selection into a `where` function now though.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExBudgetingDefaults.hs
```diff
@@ -83,53 +86,136 @@ defaultCekMachineCosts =
...
+   the codebase. For the time being we'll just use the most recent cost model
+   for all of these.  In fact we may want tests for each variant in some cases
+   TODO:
+      * Maybe export fewer things and extract the required component at the use site.
+      * Make the names more sensible: does "default" refer to the variant
+        or the set of builtins?
+      * Perhaps export functions like `defaultBuiltinCostModelForVariant
+        and then apply those to `def` where they're used.
+-}
+defaultBuiltinsRuntimeForTesting
```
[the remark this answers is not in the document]
kwxm: Let's maybe fix that in #6088.  There are a lot of `defaultXyzForTesting` things in this file and we can try to come up with better names for them when we decide if we want versions for the different variants.

plutus-ledger-api/changelog.d/20240522_212027_kenneth.mackenzie_provisional_conway_cost_model.md
kwxm: OK, let me rephrase that.

kwxm:
> Shouldn't we have 3 versions of defaultCostModelParams.json?

This is another thing I want to deal with in #6088.

kwxm:
> Now that I see in how many legitimate places we have `defaultCekParametersForTesting`, I realize that I probably shouldn't have made it twice as slow, that's just wasteful and probably makes our tests a lot slower. I guess we should simply revert that change and hope that things don't change much so that we keep having good performance.

I'll add a comment about that to issue #6088.

plutus-benchmark/common/PlutusBenchmark/Common.hs
```diff
@@ -158,14 +158,14 @@ cekResultMatchesHaskellValue term matches value =
...
+    case PLC.defaultCostModelParamsForTesting of
         Just p ->
             let errOrCtx =
                     -- The validation benchmarks were all created from PlutusV1 scripts
                     LedgerApi.mkDynEvaluationContext
                         LedgerApi.PlutusV1
-                        [DefaultFunSemanticsVariant1]
-                        (const DefaultFunSemanticsVariant1)
+                        [DefaultFunSemanticsVariantC]
+                        (const DefaultFunSemanticsVariantC)
```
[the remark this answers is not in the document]
kwxm: > Great rename as well, but why the semantics change in this case? `DefaultFunSemanticsVariant1` is `DefaultFunSemanticsVariantB`, right?

Oops! Thanks.

plutus-core/cost-model/budgeting-bench/Common.hs
```diff
@@ -86,8 +86,9 @@ benchWith params name term = bench name $ whnf (unsafeEvaluateCekNoEmit params)
    back a 'Data' value it'll traverse all of it.
 -}
 
+{- Benchmark with the most recent CekParameters -}
 benchDefault :: String -> PlainTerm DefaultUni DefaultFun -> Benchmark
-benchDefault = benchWith defaultCekParameters
+benchDefault = benchWith defaultCekParametersForTesting
```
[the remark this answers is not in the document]
kwxm: Actually no, or at least not yet.  I did try changing that a while ago and a cost model based on the results gave substantially less accurate predictions of execution times than one based on this,  I'll make an issue to look into that if I haven't already.

zliu41 (approved):

PR #6144, opened 2024-05-28
Refactored Serialise/Flat-Via. Fixes #6083

### LINKED ISSUE #6083: DeserialiseFailureInfo does not account for Flat
https://github.com/IntersectMBO/plutus/blob/0114f6d66cf2dd23debe1928a320235a2250b39d/plutus-core/plutus-core/src/Codec/CBOR/Extras.hs#L77-L86

Also we switched to new cborg 2.10 version

bezirg:
Some refactoring to make SerialiseViaFlat and FlatViaSerialise uniform.

Also removed some code not needed anymore.

### DIFF
--- a/plutus-core/changelog.d/20240528_112406_bezirg.md
+++ b/plutus-core/changelog.d/20240528_112406_bezirg.md
@@ -1,0 +1,4 @@
+### Changed
+
+- Renamed decodeViaFlat to decodeViaFlatWith
+- Renamed AsSerialize to FlatViaSerialise
--- a/plutus-core/executables/plutus/AnyProgram/IO.hs
+++ b/plutus-core/executables/plutus/AnyProgram/IO.hs
@@ -17,7 +17,7 @@ import PlutusCore.Pretty qualified as PP
 import PlutusPrelude hiding ((%~))
 import Types
 
-import Codec.CBOR.Extras
+import Codec.Extras.SerialiseViaFlat
 import Codec.Serialise (deserialiseOrFail, serialise)
 import Data.ByteString qualified as BS
 import Data.ByteString.Lazy qualified as BSL
--- a/plutus-core/plutus-core/src/Codec/Extras/SerialiseViaFlat.hs
+++ b/plutus-core/plutus-core/src/Codec/Extras/SerialiseViaFlat.hs
@@ -1,13 +1,12 @@
 {-# LANGUAGE LambdaCase        #-}
 {-# LANGUAGE OverloadedStrings #-}
-
-module Codec.CBOR.Extras (
-  SerialiseViaFlat (..),
-  decodeViaFlat,
-  DeserialiseFailureInfo (..),
-  DeserialiseFailureReason (..),
-  readDeserialiseFailureInfo,
-) where
+module Codec.Extras.SerialiseViaFlat
+    ( SerialiseViaFlat (..)
+    , decodeViaFlatWith
+    , DeserialiseFailureInfo (..)
+    , DeserialiseFailureReason (..)
+    , readDeserialiseFailureInfo
+    ) where
 
 import Codec.CBOR.Decoding qualified as CBOR
 import Codec.CBOR.Read qualified as CBOR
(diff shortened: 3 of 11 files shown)


plutus-core/changelog.d/20240528_112406_bezirg.md
```diff
@@ -0,0 +1,4 @@
+### Changed
+
+- Renamed decodeViaFlat to decodeViaFlatWIth
```
Unisay, line 3: nit: `s/I/i/`

plutus-core/executables/plutus/AnyProgram/IO.hs
```diff
@@ -17,7 +17,7 @@ import PlutusCore.Pretty qualified as PP
 import PlutusPrelude hiding ((%~))
 import Types
 
-import Codec.CBOR.Extras
+import Codec.Extras.SerialiseViaFlat
```
Unisay, line 20:
```suggestion
import Codec.Extras.SerialiseViaFlat (SerialiseViaFlat (..))
```

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Flat.hs
```diff
@@ -91,17 +92,6 @@ encoding of bytestrings is a sequence of 255-byte chunks. This is okay, since us
 be broken up by the chunk metadata.
 -}
 
--- TODO: This is present upstream in newer versions of flat, remove once we get there.
```
Unisay, line 94: 👍🏼

plutus-tx/src/PlutusTx/Coverage.hs
```diff
@@ -1,4 +1,3 @@
--- editorconfig-checker-disable-file
 {-# LANGUAGE DeriveAnyClass     #-}
 {-# LANGUAGE DerivingStrategies #-}
```
Unisay, line 2:
```suggestion
```

plutus-tx/src/PlutusTx/Coverage.hs
```diff
@@ -32,9 +31,9 @@ module PlutusTx.Coverage ( CoverageAnnotation(..)
 
 import Control.Lens
 
+import Codec.Extras.FlatViaSerialise
```
Unisay, line 34:
```suggestion
import Codec.Extras.FlatViaSerialise (FlatViaSerialise (..))
```

plutus-tx/src/PlutusTx/Coverage.hs
```diff
@@ -32,9 +31,9 @@ module PlutusTx.Coverage ( CoverageAnnotation(..)
...
 import Codec.Serialise
-
-import PlutusCore.Flat
+import Flat hiding (to)
```
Unisay, line 36:
```suggestion
import Flat (Flat, Generic)
```

Unisay (approved):
Looks good, I had only stylistic comments.

plutus-core/plutus-core/src/Codec/Extras/SerialiseViaFlat.hs
```diff
@@ -45,16 +44,16 @@ readDeserialiseFailureInfo (CBOR.DeserialiseFailure byteOffset reason) =
   DeserialiseFailureInfo byteOffset $ interpretReason reason
  where
   -- Note that this is subject to change if `cborg` dependency changes.
-  -- Currently: cborg-0.2.9.0
+  -- Currently: cborg-0.2.10.0
```
kwxm, line 47: I don't think I'd looked at this code before.  It'll probably be sheer luck if we notice that the `cborg` error strings change at some time in the future!  I suppose it'd be harmless though.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Instance/Flat.hs
kwxm: Well spotted!

kwxm (approved):
Looks good.

kwxm:
[I thought I'd  said this before, but it seems to have vanished.  Anyway, it's not strictly about this PR.]

I vaguely wondered if we should intercept the Flat error messages and replace them with something more meaningful.  We often get questions about errors like this:

```
Flat deserialisation failure: TooMuchSpace (0x00000042005e82c8,S {currPtr = 0x00000042005e701d, usedBits = 3})
```

This happens when the flat decoder is expecting the end of a buffer but it finds that there's still some input left (see [here](https://hackage.haskell.org/package/flat-0.6/docs/src/Flat.Decoder.Run.html#strictDecoder)).  Usually when people run into this it's because they're trying to flat-decode something that isn't flat-encoded, typically something that _is_ flat-encoded but with an extra layer of CBOR encoding.  Should we try to say something more sensible?  Maybe not: there could be something else wrong.  I'm not sure what would be best.

PR #6146, opened 2024-05-29
Combined Haddock Generation

zeme-wana:
This PR introduces a new workflow that builds the combined haddock (including internal libraries) for plutus and publishes it to the *haddock/master* and *haddock/release/X* folders in the `gh-pages` branch.
An initial version can be viewed [here](https://intersectmbo.github.io/plutus/haddock/good-haddock)

### DIFF
--- a/plutus-benchmark/lists/src/PlutusBenchmark/Lists/Sort/MergeSort.hs
+++ b/plutus-benchmark/lists/src/PlutusBenchmark/Lists/Sort/MergeSort.hs
@@ -2,6 +2,8 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE TemplateHaskell   #-}
 
+{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:defer-errors #-}
+
 {- | Simple merge sort implementation -}
 module PlutusBenchmark.Lists.Sort.MergeSort where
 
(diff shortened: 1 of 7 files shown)


plutus-benchmark/lists/src/PlutusBenchmark/Lists/Sort/MergeSort.hs
```diff
@@ -2,6 +2,8 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE TemplateHaskell   #-}
 
+{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:defer-errors #-}
```
zeme-wana, line 5: This is needed or haddock breaks

scripts/combined-haddock.sh
```diff
@@ -0,0 +1,115 @@
...
+  --prologue=$SRC/haddock.prologue
+
+# Setup destination directory 
+DST=combined_haddock_dst
+rm -rf $DST
+mkdir -p $DST
+
+# List of target haskell packages 
+# TODO use `find` to avoid hard-coding the list
+PACKAGE_NAMES=(
```
zliu41, line 40: I don't think this list is complete - e.g., it doesn't have `containers`.

It would be nice if we didn't have to build haddock for `base` and other third party libraries, but link to their haddock on Hackage. What would it take to do so?

zliu41 (approved):
Looks great!

scripts/combined-haddock.sh
zeme-wana: We are not exactly building haddock for base and other third party libraries, we are just making all hyperlinks point to our local copy. 
Unfortunately `cabal project-haddock` still has some limitations and this is as far as we can go. 
We can wait until better support is available in cabal.

As for the incomplete list -- I update the script

PR #6149, opened 2024-05-29
Remove reference to benching.csv in plutus-core.cabal

kwxm:
There was a reference to `benching.csv` in `plutus-core.cabal`.  That file has now been replaced by `benching-vasil.csv` and `benching-conway.csv`, but I don't think either of these is needed in a distribution so I've just removed the entry for the original file.

I'll set this to auto-merge since it's not a big change.

### DIFF
--- a/plutus-core/plutus-core.cabal
+++ b/plutus-core/plutus-core.cabal
@@ -1,22 +1,22 @@
-cabal-version:      3.0
-name:               plutus-core
-version:            1.28.0.0
-license:            Apache-2.0
+cabal-version:   3.0
+name:            plutus-core
+version:         1.28.0.0
+license:         Apache-2.0
 license-files:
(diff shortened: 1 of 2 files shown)


zliu41 (approved):

plutus-core/plutus-core.cabal
```diff
@@ -1,24 +1,23 @@
...
+author:          Plutus Core Team
+synopsis:        Language library for Plutus Core
+description:     Pretty-printer, parser, and typechecker for Plutus Core.
+category:        Language, Plutus
+build-type:      Simple
 extra-doc-files:
   CHANGELOG.md
   README.md
 
-extra-source-files:
```
zliu41, line 19: Are these not extra source files, at least the JSON files?

Issue #6088, opened 2024-05-22
Decide how to deal with multiple cost models in the tests 


kwxm:
PR #6087 introduces multiple cost models for Plutus Core.  There are many budget tests which currently only use the latest cost model.  We should decide what do do with these: for example, should we run the tests with every cost model and have a golden file for each?

kwxm:
Something else to consider is that not all builtins are available in all Plutus Language versions. The on-chain deserialisation process [checks for the use of unavailable builtiins](https://github.com/IntersectMBO/plutus/blob/5771700610196202058c9cacd327837c51765106/plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs#L204) and fails if any are found.  However many (most?) of our tests are either read from textual UPLC or compiled from PlutusTx and no check for availability is performed, so you can run a program containing a builtin which is actually illegal for the PV/LL in use.  We might want to serialise and then deserialise again before test execution in order to avoid this. The same could be done for `uplc` and so on.  In fact, unavailable builtins are currently detected in these situations by giving them costing functions which are maximally expensive, but this isn't ideal and may change.  Also this technique is slightly weaker than the deserialisation check because that will reject a script immediately if it contains a prohibited builtin whereas using an expensive costing function will only detect a prohibited builtin when it is called, which may never happen,

kwxm:
We may also want to think about nofib-exe. For instance how do we [evaluate terms](https://github.com/IntersectMBO/plutus/blob/85cf1edc5db0b3c02fdd59f84e4e599334ed62bb/plutus-benchmark/nofib/exe/Main.hs#L205)?  Do we want a flag saying which Plutus/PV/cost model to use?

kwxm:
Also decide what to do with [defaultCostModelParams.json.](https://github.com/IntersectMBO/plutus/blob/master/plutus-core/plutus-core/test/CostModelInterface/defaultCostModelParams.json)

kwxm:
Maybe get rid of `defaultCekParametersForTesting` altogether: see [this comment](https://github.com/IntersectMBO/plutus/pull/6087#pullrequestreview-2074735785) on PR #6087.

PR #6152, opened 2024-05-30
Use the most recent evaluation context for most of the benchmarks

kwxm:
I had some trouble doing some benchmarking yesterday and discovered that it was because `mkEvalCtx` was [using the most recent cost model parameters](https://github.com/IntersectMBO/plutus/blob/e675e654c5f8847bd9cf1c25aa6fa0985c8a5e8f/plutus-benchmark/common/PlutusBenchmark/Common.hs#L92) with PlutusV1 and my branch had added some parameters that weren't compatible with the old cost model.  This probably crept in when I was merging [the PR that added the Conway cost model](https://github.com/IntersectMBO/plutus/pull/6087).    I'm not entirely sure what we want to do (cf #6088) but this PR adds a `mkMostRecentEvalCtx` function that creates an evaluation context for the most recent ledger language and semantics variant, and that gets used for all of the benchmarks except for the validation benchmarks, which are run with the earliest possible context since they're very old.

I'm open to being convinced that we should do something different, so opinions are welcome.

### DIFF
--- a/plutus-benchmark/validation/bench/BenchFull.hs
+++ b/plutus-benchmark/validation/bench/BenchFull.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE BangPatterns #-}
 module Main where
 
+import PlutusCore.Default (BuiltinSemanticsVariant (DefaultFunSemanticsVariantA))
 import PlutusCore.Evaluation.Machine.ExBudget
 import PlutusLedgerApi.Common.Versions
 import PlutusLedgerApi.V1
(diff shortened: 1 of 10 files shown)


plutus-benchmark/validation/bench/BenchFull.hs
kwxm, line 1: I'm not sure why the formatting's changed in this file.  Emacs seems to think it should be different from what it was before.

effectfully (approved):
Seems fine, I don't know what the "right" approach would be either.

ana-pantilie (approved):

kwxm:
Some slightly random benchmark results, but basically it looks good.

PR #6159, opened 2024-05-30
Require PlutusV3 scripts to evaluate to BuiltinUnit

zliu41:
Implements [CIP-0117](https://github.com/cardano-foundation/CIPs/pull/747).

Tests are in Spec.ReturnUnit.V1/2/3


plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
```diff
@@ -250,9 +255,17 @@ evaluateScriptCounting ll pv verbose ectx p args = swap $ runWriter @LogOutput $
...
             evaluateTerm UPLC.counting pv verbose ectx appliedTerm
     tell logs
-    liftEither $ first CekError $ void res
+    case res of
+        Left e -> throwError (CekError e)
+        Right v -> unless (ll == PlutusV1 || ll == PlutusV2 || isUnit v) $ throwError NonUnitReturnValue
     pure final
 
+isUnit :: UPLC.Term UPLC.NamedDeBruijn DefaultUni DefaultFun () -> Bool
+isUnit t = case readKnownConstant t of
```
effectfully, line 264: I think you can simply use `readKnown`? It should figure out that you want a constant.

Also, call it `isBuiltinUnit`?

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
```diff
@@ -250,9 +255,17 @@ evaluateScriptCounting ll pv verbose ectx p args = swap $ runWriter @LogOutput $
     let (res, UPLC.CountingSt final, logs) =
             evaluateTerm UPLC.counting pv verbose ectx appliedTerm
     tell logs
-    liftEither $ first CekError $ void res
+    case res of
+        Left e -> throwError (CekError e)
+        Right v -> unless (ll == PlutusV1 || ll == PlutusV2 || isUnit v) $ throwError NonUnitReturnValue
```
effectfully, line 260: Given that this logic is repeated twice and `tell logs` is repeated as well, I'd abstract it into a separate function like we did with `evaluateTerm`. I can imagine we'll want it for tests, new benchmarks etc, but even if not, I'd rather have it in a single place to make sure it never runs out of sync.

plutus-ledger-api/test-plugin/Spec/ReturnUnit/V1.hs
```diff
@@ -0,0 +1,87 @@
...
+  CompiledCode (BuiltinData -> a) ->
+  -- | Script argument
+  Data ->
+  TestTree
+expectSuccess name code arg = testCase name $ case res of
+  Left _  -> assertFailure "fails"
+  Right _ -> pure ()
+  where
+    sScript = serialiseCompiledCode code
+    script = either (error . show) id $ V1.deserialiseScript conwayPV sScript
```
effectfully, line 56: Nit: is this serialization-deserialization roundtrip necessary?

plutus-ledger-api/test-plugin/Spec/ReturnUnit/V1.hs
```diff
@@ -0,0 +1,87 @@
...
+  $$( compile
+        [||
+        \d ->
+          let n = PlutusTx.unsafeFromBuiltinData d
+           in PlutusTx.greaterThanInteger n 0
+        ||]
+    )
+
+tooManyParameters :: CompiledCode (BuiltinData -> BuiltinData -> BuiltinUnit)
+tooManyParameters =
```
effectfully, line 80: I have no idea how you find the willpower to avoid naming those `good`, `bad` and `ugly`.

plutus-ledger-api/test-plugin/Spec/ReturnUnit/V2.hs
```diff
@@ -0,0 +1,87 @@
...
+    )
+
+tooManyParameters :: CompiledCode (BuiltinData -> BuiltinData -> BuiltinUnit)
+tooManyParameters =
+  $$( compile
+        [||
+        \d _ ->
+          let n = PlutusTx.unsafeFromBuiltinData d
+           in check (PlutusTx.greaterThanInteger n 0)
+        ||]
```
effectfully, line 86: Nit: I'm not a fan of that repetition, I'd abstract it into a general set of tests or something. At least the definitions of the three `CompiledCode`s could be shared? Not important though.

effectfully (approved):
I'm not a fan of the repetition, particularly given that we're likely to have more Plutus versions in future, so we'd have to copy-paste all those tests once again. But not a big deal. Otherwise LGTM.

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
zliu41: Factored the common code out into `processLogsAndErrors`

plutus-ledger-api/test-plugin/Spec/ReturnUnit/V1.hs
zliu41: Yes, `ScriptForEvaluation` can only be obtained via `deserialiseScript`, which is by design.

plutus-ledger-api/test-plugin/Spec/ReturnUnit/V1.hs
zliu41: It took a lot of willpower indeed!

plutus-ledger-api/test-plugin/Spec/ReturnUnit/V2.hs
zliu41: > At least the definitions of the three CompiledCodes could be shared?

We need different compiler flags for V1/V2 vs. V3. Perhaps we can find a way to share them, but I'd rather think about that later.

Unisay: `plutus-tx-plugin/test/Plugin/NoTrace/Lib.hs` is an example of a shared text fixture.

Unisay (approved):
LGTM

plutus-ledger-api/test-plugin/Spec/ReturnUnit/V1.hs
effectfully: Thanks for the explanation!

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
```diff
@@ -249,10 +251,31 @@ evaluateScriptCounting ll pv verbose ectx p args = swap $ runWriter @LogOutput $
...
+    Either
+        (UPLC.CekEvaluationException NamedDeBruijn DefaultUni DefaultFun)
+        (UPLC.Term UPLC.NamedDeBruijn DefaultUni DefaultFun ()) ->
+    m ()
+processLogsAndErrors ll logs res = do
+    tell logs
+    case res of
+        Left e -> throwError (CekError e)
+        Right v ->
+            unless (ll == PlutusV1 || ll == PlutusV2 || isBuiltinUnit v) $
```
michaelpj, line 271: this needs a Note! With a reference to the CIP

michaelpj: I'd also be tempted to extract this out to a top-level predicate with a descriptive name and the explanation attached to it. It's a significant thing, worth making it look more significant!

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
```diff
@@ -249,10 +251,31 @@ evaluateScriptCounting ll pv verbose ectx p args = swap $ runWriter @LogOutput $
...
+                throwError NonUnitReturnValue
```
michaelpj, line 272: Include the actual result here for a more informative error?

michaelpj (approved):
Looks fine to me.

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
```diff
@@ -53,6 +54,7 @@ data EvaluationError =
     | CodecError !ScriptDecodeError -- ^ A deserialisation error
     -- TODO: make this error more informative when we have more information about what went wrong
     | CostModelParameterMismatch -- ^ An error indicating that the cost model parameters didn't match what we expected
+    | NonUnitReturnValue -- ^ The script evaluated to a value that is not the builtin unit.
```
bezirg, line 57: Maybe a more descript name like `NonUnitReturnValueConway`, to remind ourselves that this can only happen with Conway onwards

zliu41: It's not about Conway but about V3. I changed the name to `InvalidReturnValue`, and made the `Pretty` instance more descriptive.

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
zliu41: We can `show` the result but it may be too big. How about taking the first x characters? x = 1000?

plutus-tx/src/PlutusTx/Prelude.hs
```diff
@@ -178,8 +181,8 @@ import Prelude qualified as Haskell (return, (=<<), (>>), (>>=))
 
 {-# INLINABLE check #-}
 -- | Checks a 'Bool' and aborts if it is false.
-check :: Bool -> ()
-check b = if b then () else traceError checkHasFailedError
+check :: Bool -> BI.BuiltinUnit
+check b = if b then BI.unitval else traceError checkHasFailedError
```
bezirg, line 185: This is big change. This means that we can never "backport" anymore the NonUnitReturnValue error to also PV1,PV2 (if we ever decide that we want that)

bezirg:
If we go with this, we should remember to update the spec pdf and RTD (in another PR)

plutus-tx/src/PlutusTx/Prelude.hs
bezirg: Not exactly, but expecting `BuiltinData` instead of `BuiltinUnit` as the final result value would be more uniform, because a function would be of type `BuiltinData -> BuiltinData` (for v3) . And it would still leave open the possibility in the future to store/cache on-chain  the result of the computations serialised (a feature that nobody wants I guess)

plutus-ledger-api/src/PlutusLedgerApi/Common/Eval.hs
```diff
@@ -249,10 +251,31 @@ evaluateScriptCounting ll pv verbose ectx p args = swap $ runWriter @LogOutput $
...
+    m ()
+processLogsAndErrors ll logs res = do
+    tell logs
+    case res of
+        Left e -> throwError (CekError e)
+        Right v ->
+            unless (ll == PlutusV1 || ll == PlutusV2 || isBuiltinUnit v) $
+                throwError NonUnitReturnValue
+    where
+        isBuiltinUnit t = case readKnown t of
```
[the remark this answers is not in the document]
zliu41: This is what the CIP proposed. To quote @michaelpj 's words:

> I don't think we should have multiple different kinds of accepted return value. unit would be okay, so long as then it has to be unit. The whole point of this CIP is to narrow down what scripts return as much as possible, so its very clear if you're doing it correctly or not. I don't really see the benefit to having a larger space of possibilities.

plutus-tx/src/PlutusTx/Prelude.hs
zliu41: > This is big change. This means that we can never "backport" anymore the InvlalidReturnValue error to also PV1,PV2 (if we ever decide that we want that)

I'm not sure I understand - this is just a Plutus Tx helper function, how is this a big change?

But in any case, I don't think we can ever backport this to V1 and V2 anyway, since there are already lots of scripts onchain that don't return BuiltinUnit, and those script must continue to work.

zliu41: > the serialised result (BuiltinData) of the computations (a feature that nobody wants I guess).

We can have more discussion on this for Plutus V4. The motivation is not super clear to me..

zliu41 (commented):
> If we go with this, we should remember to update the spec pdf and RTD (in another PR also ok)

Agreed!

plutus-tx/src/PlutusTx/Prelude.hs
bezirg: Not very important. I just assumed that  Tx.check was all these years using `BuiltinUnit` as the result already. So I had hopes that we could  backport this BuiltinUnit-result-restriction also to PV1,PV2 somewhere in the future. But it is clear to me now that it definitely cannot be backported.

bezirg (approved):

zliu41:
Merging to unblock the release. Will address the remaining comments (mostly documentation) separately.

kwxm (commented):
This all looks good.  How will we publicise it? I guess script authors will have to be aware of it.

Also, I don't think CIP-117 has been merged yet.  How will we deal with that, given that we've gone ahead and done it anyway?

Issue #6089, opened 2024-05-22
Rework plutus-ledger-api tests for new cost model machinery


kwxm:
PR #6087 introduces multiple cost models for Plutus Core and increases the configurability of cost models.  `plutus-ledger-api` contains many tests for the cost model parameters which are now not very meaningful (for example, they used to check that the V1 cost model parameters are a subset of the V2 parameters, and that those are a subset of the V3 parameters, which is no longer true).  These tests have been adjusted so that they pass for now, but the whole thing really needs to be completely reworked.

kwxm:
We're planning to make all builtins available in all Plutus LL versions, so many of the tests will no longer make sense.  Exactly what _should_ we be testing now?  The situation's quite difficult because of the evolving nature of the different versions: assumptions that held in earlier PVs may no longer hold in later PVs, and we're only testing for the current/upcoming PV.

PR #6161, opened 2024-05-31
Quadratic costing for integer division functions

kwxm:
This provides more precise costing for the Plutus Core integer division builtins, whose CPU costs are rather complicated and were generally being overestimated by our previous model.  

I've had to add a number of extra cost model parameters for this, and that's broken some of the tests in `plutus-ledger-api` which make assumptions that are no longer true (for example, that the cost model parameters for a given PV are always a subset of the parameters for later PVs).  I've commented those out for the time being:  I don't particularly like doing that, but those tests need substantial reworking anyway (see #6089) and I didn't want the problematic ones to just vanish.


effectfully (approved):
LGTM

But commenting out tests sucks. I understand the rationale though.

kwxm:
Benchmark results look good.  The first lot showed a slowdown, but I think that was just random.

PR #6162, opened 2024-05-31
Tests for the `preserve-logging` flag.

Unisay:
A [user has reported](https://github.com/IntersectMBO/plutus/issues/5949#issuecomment-2141488071) that `preserve-logging` flag is not recognized, so I decided to verify it by adding it to the spec which verifies behavior of the `remove-trace` flag (very related).

(I wasn't able to reproduce the misbehaviour, tests show that in a presence of this flag logs are preserved.)

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
@@ -28,10 +28,10 @@ evaluateBuiltinsPass :: (PLC.Typecheckable uni fun, PLC.GEq uni, Applicative m)
   -> BuiltinsInfo uni fun
   -> CostingPart uni fun
   -> Pass m TyName Name uni fun a
-evaluateBuiltinsPass tcconfig conservative binfo costModel =
+evaluateBuiltinsPass tcconfig preserveLogging binfo costModel =
   NamedPass "evaluate builtins" $
     Pass
-      (pure . evaluateBuiltins conservative binfo costModel)
+      (pure . evaluateBuiltins preserveLogging binfo costModel)
       [Typechecks tcconfig]
       [ConstCondition (Typechecks tcconfig)]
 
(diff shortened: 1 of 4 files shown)


plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
```diff
@@ -55,18 +55,18 @@ evaluateBuiltins conservative binfo costModel = transformOf termSubterms process
       -> Maybe (Term tyname name uni fun ())
     eval (BuiltinCostedResult _ getX) AppContextEnd =
         case getX of
-            BuiltinSuccess v           -> Just v
+            BuiltinSuccess term           -> Just term
             -- Evaluates successfully, but does logging. If we're being conservative
             -- then we should leave these in, so we don't remove people's logging!
             -- Otherwise `trace "hello" x` is a prime candidate for evaluation!
-            BuiltinSuccessWithLogs _ v -> if conservative then Nothing else Just v
+            BuiltinSuccessWithLogs _ term -> if preserveLogging then Nothing else Just term
```
effectfully, line 62: It perhaps should be `conservative` rather than `preserveLogging`, because we probably don't want to optimize `error` that is on the evaluation path for example. But maybe not, I asked something along these lines on Slack.

plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
```diff
@@ -55,18 +55,18 @@ evaluateBuiltins conservative binfo costModel = transformOf termSubterms process
       -> Maybe (Term tyname name uni fun ())
     eval (BuiltinCostedResult _ getX) AppContextEnd =
         case getX of
-            BuiltinSuccess v           -> Just v
+            BuiltinSuccess term           -> Just term
```
effectfully, line 58: :+1:

effectfully (approved):
Great stuff.

plutus-core/plutus-ir/src/PlutusIR/Transform/EvaluateBuiltins.hs
Unisay: The thing is that at the `Compiler` applies this `evaluateBuiltins` function like this:
```haskell
preserveLogging <- view (ccOpts . coPreserveLogging)
... $ EvaluateBuiltins.evaluateBuiltinsPass tcconfig preserveLogging binfo costModel
```

PR #6165, opened 2024-06-03
Fix "un-rolling" a list type.

### LINKED ISSUE #6163: `Unroll` gets into infinite loop on `[SomeType]`
### Summary

This seems like unintended behavior because:

1. AssocMap uses lists, so they probably should work
2. There is instance of `HasSchema` for `[x]` provided

Compile:

```
definitions :: Definitions (Unroll [Integer])
definitions = error "TODO"
```

Same for `deriveDefinitions @'[[Integer]]`

### Actual Result

Typechecking fails with error like ` Reduction stack overflow; ...`

### Expected Result

Type-checking does not fail.

_No response_

### System info

plutus-tx 1.27.0.0

GHC  9.6.3

Unisay:
This fix closes #6163 by adding a type family clause, instead of relying on the generic rep of `[]`

### DIFF
--- a/plutus-tx/src/PlutusTx/Blueprint/Definition/Unroll.hs
+++ b/plutus-tx/src/PlutusTx/Blueprint/Definition/Unroll.hs
@@ -26,8 +26,8 @@ import GHC.TypeLits qualified as GHC
 import PlutusTx.Blueprint.Class (HasSchema)
 import PlutusTx.Blueprint.Definition.Id as DefinitionId (AsDefinitionId (..))
 import PlutusTx.Blueprint.Definition.Internal (Definitions (..), addDefinition, definition)
-import PlutusTx.Builtins.Internal (BuiltinByteString, BuiltinData, BuiltinList, BuiltinString,
-                                   BuiltinUnit)
+import PlutusTx.Builtins.Internal (BuiltinByteString, BuiltinData, BuiltinList, BuiltinPair,
+                                   BuiltinString, BuiltinUnit)
 
 ----------------------------------------------------------------------------------------------------
 -- Functionality to "unroll" types. -- For more context see Note ["Unrolling" types] -----------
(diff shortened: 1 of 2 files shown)


plutus-tx/src/PlutusTx/Blueprint/Definition/Unroll.hs
```diff
@@ -91,6 +91,7 @@ type family Unroll (p :: Type) :: [Type] where
   Unroll BuiltinString = '[BuiltinString]
   Unroll (BuiltinList a) = Prepend (BuiltinList a) (GUnroll (Rep a))
   Unroll BuiltinByteString = '[BuiltinByteString]
+  Unroll [a] = Unroll a
```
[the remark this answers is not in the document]
Unisay: There is a test showing that original type is indeed contained in the un-rolling. The RHS `Unroll a` will make sure that `a` is added to the un-rolling, as long as its not a list itself.

plutus-tx/test/Blueprint/Spec.hs
```diff
@@ -91,3 +94,22 @@ definitions = unroll @(UnrollAll '[Foo])
...
+
+testUnrollPair :: Unroll (Integer, Bool) :~: '[Bool, Integer]
+testUnrollPair = Refl
+
+testUnrollBuiltinPair :: Unroll (BuiltinPair Integer Bool) :~: '[Bool, Integer]
+testUnrollBuiltinPair = Refl
+
+testUnrollBuiltinList
+  :: Unroll (BuiltinList (BuiltinPair Bool BuiltinUnit)) :~: '[BuiltinUnit, Bool]
+testUnrollBuiltinList = Refl
```
effectfully, line 112: :+1:

plutus-tx/src/PlutusTx/Blueprint/Definition/Unroll.hs
```diff
@@ -89,8 +89,12 @@ type family Unroll (p :: Type) :: [Type] where
   Unroll BuiltinData = '[BuiltinData]
   Unroll BuiltinUnit = '[BuiltinUnit]
   Unroll BuiltinString = '[BuiltinString]
-  Unroll (BuiltinList a) = Prepend (BuiltinList a) (GUnroll (Rep a))
+  Unroll (BuiltinList a) = Unroll a
+  Unroll (BuiltinPair a b) = Unroll a ++ Unroll b
   Unroll BuiltinByteString = '[BuiltinByteString]
+  Unroll [a] = Unroll a
+  Unroll (a, b) = Unroll a ++ Unroll b
+  Unroll (Maybe a) = Unroll a
```
effectfully, line 97: Well I still have no idea why this is desired rather than the `Prepend` thing. I do remember our discussion about lists, pairs etc being an inherent part of a schema, but what if it's a list of lists, is that now "user-space"? And I'm failing to understand how this logic maps onto the `#list`, `#pair` etc built-in types as per the [CIP](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0057).

Sorry for being useless! Maybe the reviewers of the original blueprint PR would know better, CC'ing @kwxm @zliu41.

effectfully:
(just testing if benchmarks work)

effectfully (approved):
:rocket:

PR #6166, opened 2024-06-03
Fix V2's costModelParamsForTesting

zliu41:
V2 and V3 use different parameter names for `divideInteger` etc.


kwxm (approved):
This seems to solve our immediate problem, but we should come back later and do #6089 properly.

PR #6178, opened 2024-06-05
 [PlutusLedgerApi] [Refactoring] Polish imports and exports

### LINKED ISSUE #6098: Export from `PlutusLedgerApi.Common`

This is a very low priority feature request:

Here are some types/classes/functions that are exported from individual plutus version in the `plutus-ledger-api`, but not from Common, despite that they are not tied to an explicit plutus version:

* `ToData(..)`
* `FromData(..)`

(I'll try to see if there are other types/functions/classes that could benefit from being exported from Common and add them to the above list)

_No response_

effectfully:
This

- resolves #6098
- moves a bunch of stuff shared among `V1`/`V2`/`V3` into `Common` (see the PR review)
- makes all imports in the `Common`/`V1`/`V2`/`V3` modules qualified so that it's clear where definitions come from and whether they are inherited from `Common` or an earlier ledger language
- fixes some formatting, adds a definition for consistency etc

### DIFF
--- a/plutus-ledger-api/src/PlutusLedgerApi/Common.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/Common.hs
@@ -1,73 +1,112 @@
+-- editorconfig-checker-disable-file
+
 -- | The types and functions that are common among all ledger Plutus versions.
-module PlutusLedgerApi.Common
-    ( -- * Script (de)serialization
-      SerialisedScript
-    , ScriptForEvaluation
-    , serialisedScript
-    , deserialisedScript
-    , serialiseCompiledCode
-    , serialiseUPLC
-    , deserialiseScript
-    , uncheckedDeserialiseUPLC
-    , ScriptDecodeError (..)
-    , ScriptNamedDeBruijn (..)
+module PlutusLedgerApi.Common (
+  -- * Script (de)serialization
+  SerialisedScript.SerialisedScript,
+  SerialisedScript.ScriptForEvaluation,
+  SerialisedScript.serialisedScript,
+  SerialisedScript.deserialisedScript,
+  SerialisedScript.serialiseCompiledCode,
+  SerialisedScript.serialiseUPLC,
+  SerialisedScript.deserialiseScript,
+  SerialisedScript.uncheckedDeserialiseUPLC,
+  SerialisedScript.ScriptDecodeError (..),
+  SerialisedScript.ScriptNamedDeBruijn (..),
 
-      -- * Script evaluation
-    , evaluateScriptCounting
-    , evaluateScriptRestricting
-    , evaluateTerm
-    , VerboseMode (..)
-    , LogOutput
-    , EvaluationError (..)
-    -- reexport Data & ExBudget for convenience to upstream users
-    , PlutusCore.Data (..)
-    , PlutusCore.ExBudget (..)
+  -- * Script evaluation
+  Eval.evaluateScriptCounting,
+  Eval.evaluateScriptRestricting,
+  Eval.evaluateTerm,
+  Eval.VerboseMode (..),
+  Eval.LogOutput,
+  Eval.EvaluationError (..),
 
-      -- * Network's versioning
-      {-| The network's behaviour (and plutus's by extension) can change via /hard forks/,
-      which directly correspond to major-number protocol version bumps.
-      -}
-    , MajorProtocolVersion (..)
-    , PlutusLedgerLanguage (..)
-    , Version (..)
-    , builtinsIntroducedIn
-    , builtinsAvailableIn
-    , ledgerLanguageIntroducedIn
-    , ledgerLanguagesAvailableIn
+  -- * Network's versioning
+  {-| The network's behaviour (and plutus's by extension) can change via /hard forks/,
+  which directly correspond to major-number protocol version bumps.
+  -}
+  Versions.MajorProtocolVersion (..),
+  Versions.PlutusLedgerLanguage (..),
+  Versions.Version (..),
+  Versions.builtinsIntroducedIn,
+  Versions.builtinsAvailableIn,
+  Versions.ledgerLanguageIntroducedIn,
+  Versions.ledgerLanguagesAvailableIn,
 
-      -- * Network's costing parameters
-      {-| A less drastic approach (that does not rely on a HF)
-      to affect the network's (and plutus's by extension) behaviour
-      is by tweaking the values of the cost model parameters.
+  -- * Costing-related types
+  PLC.ExBudget (..),
+  PLC.ExCPU (..),
+  PLC.ExMemory (..),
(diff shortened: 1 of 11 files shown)


plutus-ledger-api/src/PlutusLedgerApi/Common.hs
```diff
@@ -1,73 +1,138 @@
...
+  MonadError,
+) where
+
+import PlutusLedgerApi.Common.Eval qualified as Eval (EvaluationContext (..), EvaluationError (..),
+                                                      LogOutput, VerboseMode (..),
+                                                      assertWellFormedCostModelParams,
+                                                      evaluateScriptCounting,
+                                                      evaluateScriptRestricting, evaluateTerm,
+                                                      mkDynEvaluationContext, mkTermToEvaluate,
+                                                      toMachineParameters)
```
effectfully, line 103: Made import lists explicit in this module. I normally don't care much about readability of imports, but given that it's one of the ledger API modules and it's important to keep it reasonable and readable, I thought it might be worth going full verbosity. Now it always clear where a definition comes from, regardless of whether you look at imports (due to the explicit list) or exports (due to everything being imported `qualified` apart from entirely trivial things, of which we only have `MonadError`).

Please tell me if I should remove the explicit imports or add them to other modules or keep everything the same.

plutus-ledger-api/src/PlutusLedgerApi/V1.hs
```diff
@@ -3,99 +3,101 @@
...
-  -- *** ByteStrings
-  BuiltinByteString,
-  toBuiltin,
-  fromBuiltin,
+  -- *** Builtins
+  Common.BuiltinByteString,
+  Common.toBuiltin,
+  Common.fromBuiltin,
+  Common.toOpaque,
+  Common.fromOpaque,
```
effectfully, line 52: I've also reexported `toOpaque` and `fromOpaque` from here, since `toBuiltin` and `fromBuiltin` used to subsume those.

plutus-ledger-api/src/PlutusLedgerApi/Common.hs
```diff
@@ -1,73 +1,138 @@
...
+  IsData.FromData (..),
+  IsData.UnsafeFromData (..),
+  IsData.toData,
+  IsData.fromData,
+  IsData.unsafeFromData,
+  Builtins.dataToBuiltinData,
+  Builtins.builtinDataToData,
+
+  -- * Misc
+  MonadError,
```
effectfully, line 94: Exported it from here, because it's used in all `V*` modules.

plutus-ledger-api/src/PlutusLedgerApi/Common.hs
```diff
@@ -1,73 +1,138 @@
...
+  Prelude.toBuiltin,
+  Prelude.fromBuiltin,
+  Prelude.toOpaque,
+  Prelude.fromOpaque,
+
+  -- * Data
+  PLC.Data (..),
+  Builtins.BuiltinData (..),
+  IsData.ToData (..),
+  IsData.FromData (..),
```
effectfully, line 85: Fixes #6098.

plutus-ledger-api/src/PlutusLedgerApi/Common.hs
```diff
@@ -1,73 +1,138 @@
...
+  IsData.UnsafeFromData (..),
+  IsData.toData,
+  IsData.fromData,
+  IsData.unsafeFromData,
+  Builtins.dataToBuiltinData,
+  Builtins.builtinDataToData,
```
effectfully, line 91: Exported other `Data`-related stuff as well.

plutus-ledger-api/src/PlutusLedgerApi/V1.hs
```diff
@@ -105,52 +107,42 @@ module PlutusLedgerApi.V1 (
...
+import PlutusLedgerApi.V1.Interval qualified as Interval
+import PlutusLedgerApi.V1.ParamName qualified as ParamName
 import PlutusLedgerApi.V1.Scripts as Scripts
-import PlutusLedgerApi.V1.Time
-import PlutusLedgerApi.V1.Value
-import PlutusTx (FromData (..), ToData (..), UnsafeFromData (..), fromData, toData)
-import PlutusTx.Builtins.Internal (BuiltinData (..), builtinDataToData, dataToBuiltinData)
-import PlutusTx.Prelude (BuiltinByteString, fromBuiltin, toBuiltin)
-
-import Control.Monad.Except (MonadError)
```
effectfully, line 147: This module now contains only `PlutusLedgerApi` imports, no Plutus Tx or `base`.

plutus-ledger-api/src/PlutusLedgerApi/Common.hs
zliu41: > I thought it might be worth going full verbosity

I believe this can be enforced by HLint, so while you are at it you may want to add an HLint rule to ensure it stays this way.

> Please tell me if I should remove the explicit imports or add them to other modules or keep everything the same.

In general having _both_ qualified imports _and_ explicit imports is an overkill, so I wouldn't do this to other modules - I don't see the need. I'm in favor of using _either_ qualified imports _or_ explicit imports everywhere, but others even find this to be not worth it.

zliu41 (approved):

plutus-ledger-api/src/PlutusLedgerApi/Common.hs
Unisay: My take:
- Wildcard imports: mostly no, could be exceptions.
- Qualified imports: nice, should use more.
- Explicit imports: also nice, complements qualified:

```haskell

message :: Text.Text -- IMO this duplication is ugly
message = Text.toUpper "hello" -- this is readable.
```

the combo of qualified and explicit is what I'd use:

```haskell
import Data.Text (Text)

message :: Text -- The best 
message = Text.toUpper "hello" -- good
```

Unisay (approved):

plutus-ledger-api/src/PlutusLedgerApi/Common.hs
effectfully: > I believe this can be enforced by HLint, so while you are at it you may want to add an HLint rule to ensure it stays this way.

Sorry, I'm fully committed to never adding HLint anywhere. I was forced to use it for months and it damaged me badly.

> In general having both qualified imports and explicit imports is an overkill

I totally agree, it's just that these modules are very special in that everything exported by them must be scrutinized deeply. My usual standards for imports are lower than yours or Yura's.

@Unisay I agree with you, I'm just usually lazy when it comes to import management. Especially if it's within "standard" modules like `base` or `PlutusCore` or whatever. I'll try to do a better job.

effectfully:
I've made imports either qualified or explicit everywhere, since that what I gauge is preferable.

PR #6181, opened 2024-06-06
[Errors] Preserve operational unlifting errors

effectfully:
I've figured out a way to preserve operational unlifting errors. It's pretty ridiculous, but should work.


plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
```diff
@@ -262,12 +261,6 @@ typeMismatchError uniExp uniAct =
 -- | The monad that 'readKnown' runs in.
 type ReadKnownM = Either BuiltinError
 
--- | Lift a 'ReadKnownM' computation into 'BuiltinResult'.
-liftReadKnownM :: ReadKnownM a -> BuiltinResult a
-liftReadKnownM (Left err) = BuiltinFailure mempty err
-liftReadKnownM (Right x)  = BuiltinSuccess x
-{-# INLINE liftReadKnownM #-}
```
effectfully, line 269: This was a leftover from the immediate unlifting times, i.e. this function has been unused for like three years.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Golden/String/consByteString-out-of-range.err.golden
```diff
@@ -1,3 +1,5 @@
 An error has occurred:
 The machine terminated because of an error, either from a built-in function or from an explicit use of 'error'.
-Caused by: (consByteString 256 #68656c6c6f20776f726c64)
\ No newline at end of file
+Caused by: (consByteString 256 #68656c6c6f20776f726c64)
+Logs were:
+256 is not within the bounds of Word8
```
effectfully, line 5: Finally the actual error message is shown to the user.

effectfully:
^ benchmarks are noise as expected.

zliu41 (approved):
Is this only case where a log message (1) does not come from `trace`, and (2) is accumulated even in `Quiet` mode? If so, it's worth pointing this out somewhere; at least the Haddock of `VerboseMode` should be updated.

effectfully:
> Is this only case where a log message (1) does not come from `trace`, and (2) is accumulated even in `Quiet` mode? If so, it's worth pointing this out somewhere; at least the Haddock of `VerboseMode` should be updated.

Not quite. We already have plenty of cases when logs are _created_ in `Quiet` mode, in particular we do plenty of unconditional logging when failing a builtin call, for example

```
integerToByteStringWrapper :: Bool -> Integer -> Integer -> BuiltinResult ByteString
integerToByteStringWrapper endiannessArg lengthArg input
  -- Check that the length is non-negative.
  | lengthArg < 0 = do
      emit "integerToByteString: negative length argument"
      emit $ "Length requested: " <> (pack . show $ input)
      evaluationFailure
```

Using `emit` + `evaluationFailure` produces the same kind of unconditional logging + operational error in the end that this PR adds for a different kind of operational errors.

Those logs however are not _reported_ in `Quiet` mode, since we have the following line in `PlutusLedgerApi.Common.Eval.evaluateTerm`:

```haskell
(if verbose == Verbose then UPLC.logEmitter else UPLC.noEmitter)
```

and `noEmitter` simply erases all created logs.

Creating error messages just to erase them is perhaps not optimal, but who cares, given that we only do that when evaluation is about to fail anyway, i.e. we don't need to optimize that a lot unlike the happy path.

So overall, this PR minimally and consistently extends existing functionality, it doesn't add anything novel.

PR #6189, opened 2024-06-07
[Test] [Builtins] Add golden tests with unlifting errors

effectfully:
This refactors a file with tests so that unit tests that are supposed to result in evaluation failure also become golden tests with the error message printed to a golden file. Need it to demonstrate that #6181 does indeed allow us to preserve operational unlifting errors.

### DIFF
--- a/plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
+++ b/plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
@@ -1,6 +1,7 @@
 -- editorconfig-checker-disable-file
 -- | Tests for all kinds of built-in functions.
 
+{-# LANGUAGE LambdaCase            #-}
 {-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE PartialTypeSignatures #-}
 {-# LANGUAGE TypeApplications      #-}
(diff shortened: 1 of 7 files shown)


plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
```diff
@@ -458,107 +464,161 @@ test_SerialiseDataImpossible =
         typecheckAnd def evalRestricting defaultBuiltinCostModelForTesting dataLoop @?=
             Right EvaluationFailure
 
+stripParensIfAny :: String -> String
+stripParensIfAny ('(' : str) = init str
```
zliu41, line 468: So what this does is removing the first two characters if the first character is `(`. Worth a comment on why it does that.

zliu41 (approved):

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
effectfully: Made it

```haskell
-- | If the first char is an opening paren and the last chat is a closing paren, then remove them.
-- This is useful for rendering a term-as-a-test-name in CLI, since currently we wrap readably
-- pretty-printed terms in parens (which is to be fixed).
stripParensIfAny :: String -> String
stripParensIfAny str@('(' : str1) | last str == ')' = init str1
stripParensIfAny str                                = str
```

PR #6193, opened 2024-06-07
Feature/docusaurus post launch

joseph-fajen:
Adding styling updates from Andrew Larbi. 
Changing the term "Documentation" to "User guide" in the upper left area next to the wordmark logo "PLUTUS". 

@ana-pantilie and @effectfully  
I suspect that this branch here, intended only to push changes to the doc site, has somehow pulled in some of your code. If we need to avoid that, I want to find a fix rather than push your code through this branch pull request. I'll add you as a reviewer so that you can let me know if we need to stop this PR and approach it differently.


effectfully (commented):
I think the code changes are all from a branch by Ana, I don't see any of my code here.

I'm eager to see how you're going to resolve it, so that I know how to do it in future myself.

effectfully:
Well, actually, in that case it's probably simply `git revert 5e29c87`.

PR #6204, opened 2024-06-11
[Test] Add missing bitwise builtins golden files

effectfully:


kwxm (approved):
We might get more of these when the next lot of bytestring primitives get merged.  I'll try to head that off.

PR #6205, opened 2024-06-12
New `./scripts/combined-haddock.sh`

zeme-wana:
Complete re-write of `./scripts/combined-haddock.sh`, which also links to haddock and fixes most broken links.


nix/project.nix
```diff
@@ -80,16 +80,6 @@ let
         # Common 
         {
           packages = {
-            # Packages we just don't want docs for
```
zeme-wana, line 83: These were probably obsolete: we can build haddock successfully for all these

ana-pantilie (approved):

PR #6206, opened 2024-06-12
Update README.adoc

joseph-fajen:
Updated the link in README to point to the Plutus docs site on the Docusaurus platform instead of Read the docs.


effectfully (commented):
Sorry for the late review. LGTM.

PR #6213, opened 2024-06-14
Haskell Eq for AssocMap

ana-pantilie:
Ledger needs to be able to compare `ScriptContext`. This PR adds the Haskell instances back, this time they are based on the `Data.Map.Strict` implementation of `Map` as suggested by @lehins.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


Unisay (approved):
LGTM

plutus-tx/src/PlutusTx/AssocMap.hs
```diff
@@ -73,6 +75,14 @@ newtype Map k v = Map {unMap :: [(k, v)]}
   deriving stock (Generic, Haskell.Show, Data, TH.Lift)
   deriving newtype (NFData)
 
+instance (Haskell.Ord k, Haskell.Eq v) => Haskell.Eq (Map k v) where
+  Map l == Map r =
+    on (Haskell.==) HMap.fromList l r
```
lehins, line 80: So, this will work as long as there are no duplicates in the assoc map.

Because of conversion of types from ledger this is guaranteed, but I am not sure if this is sufficient for you. I am just not sure what are the semantic guarantees that you provide. So, if in presence of duplicate keys assoc map ignores them then we are fine with this implementation, but otherwise we either need to fallback to a slow O(n^2) implementation or implement something smarter

```suggestion
    let lm = HMap.fromList l
        rm = HMap.fromList r
        lLen = length l
        rLen = length r
     in if lLen == rLen && lLen == HMap.size l && HMap.size r
          then lm == rm
          else sortOn fst l == sortOn fst r
```

lehins (approved):
:+1:

PR #6236, opened 2024-06-24
Add workflows to check broken links and deploy papers to GH Pages

zeme-wana:


kwxm (approved):
I don't understand the details, but this looks as if it should do what's required.  Let's merge it and see if it works!

PR #6239, opened 2024-06-25
Cabal project to compile docusaurus examples.

Unisay:
**Before this PR:**

- docusaurus project contains Haskell source files that aren't typechecked and might contain compilation errors.
- A change in the Plutus plug-in or API might break examples without any CI failure.
-  CIP-57 Example validator script isn't compatible with the CIP-69 and CIP-117.

**After this PR:**

- Cabal package `docusaurus-examples` exists to build a single executable `example-cip57` (for now) which compiles a corresponding CIP-57 Blueprint example.
- A change in the Plutus plug-in or API that breaks examples will cause the whole project build to fail CI.
-  CIP-57 Example validator script is compatible with the CIP-69 and CIP-117.

### DIFF
--- a/doc/docusaurus/docusaurus-examples.cabal
+++ b/doc/docusaurus/docusaurus-examples.cabal
@@ -1,0 +1,37 @@
+cabal-version: 3.0
+name:          docusaurus-examples
+version:       0.1.0.0
+license:       Apache-2.0
+license-file:  LICENSE
+author:        Yura Lazaryev
+maintainer:    Yuriy.Lazaryev@iohk.io
+category:      Language
+build-type:    Simple
+
+source-repository head
+  type:     git
+  location: https://github.com/IntersectMBO/plutus
+
+common lang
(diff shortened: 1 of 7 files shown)


doc/docusaurus/docusaurus-examples.cabal
```diff
@@ -0,0 +1,27 @@
...
+build-type:    Simple
+
+common warnings
+  ghc-options: -Wall
+
+executable example-cip57
+  import:           warnings
+  main-is:          Example/Cip57/Blueprint/Main.hs
+  hs-source-dirs:   static/code
+  default-language: GHC2021
```
zeme-wana, line 18: GHC2021 is supported since 9.2.1.
In CI we are still supporting 8.10.7 so you'll have to wrap this in an `if()` or not use it at all I'm afraid.

zeme-wana (approved):

PR #6245, opened 2024-06-26
Cost model overview document

kwxm:
We don't currently have any documentation about the cost model.  This provides a high-level overview and explains by example what the cost model parameters mean.  Eventually we'll have to add the details of the cost model to the specification, but this should give a rough idea of how it works for the time being.

Here's a PDF: 

[cost-model-overview.pdf](https://github.com/user-attachments/files/15988812/cost-model-overview.pdf)


doc/cost-model-overview/cost-model-overview.tex
```diff
@@ -0,0 +1,386 @@
...
+runaway memory allocation, and we regard CPU costing as our primary measure of
+cost.
+
+\subsection{Cost models and cost model parameters}
+A Plutus Core \textit{cost model} consists of CPU and memory costs for all of
+the basic Plutus Core operations together with CPU and memory costing functions
+for all of the built-in functions.  The specific cost model used when a script
+is evaluated on the chain may depend on the Plutus language version and the
+protocol version.  This allows us to account for factors such as improvements in
+the efficiency of the evaluator, and also to maintain backwards compatiiblity
```
bezirg, line 126: typo: compatiiblity

doc/cost-model-overview/cost-model-overview.tex
```diff
@@ -0,0 +1,386 @@
...
+$$
+\text{CPU cost} = \begin{cases}
+  \const & \text{if $x<y$} \\
+  \max(d, c_{00} + c_{10}x + c_{01}y + c_{20}x^2 + c_{11}xy + c_{02}y^2) & \text{if $x \geq y$}.
+\end{cases}
+$$
+
+\subsection{Cost model parameters in the protocol parameters}
+\label{sec:cost-model-parameters}
+In the protocol parameters the cost model parameters take the form of a list of
```
bezirg, line 323: This sentence is not exactly correct, since in the "real" protocol parameters, there are no names, but just the coefficient values in a big array: https://github.com/perturbing/plutus-plonk-example/blob/1451b2a67b445089a579298195895915b76a523f/local-testnet/scripts/babbage/conway-babbage-test-genesis.json#L28

kwxm: > in the "real" protocol parameters, there are no names
I think that if you ask the node for the protocol parameters you get back a JSON object where all of the protocol parameters are tagged with their names.  At least that's what I got when I queried a node with cardano-cli a while ago.  I'll try that again and see if the same thing happens.  I think that the node maybe has different views of the protocol parameters, some with tags and some without, but the tags have always been removed from the cost model parameters when the node calls `mKEvaluationContext`.  This is all a bit unclear to me.

bezirg: For me too. If they do provide a JSON to the user, then disregard what I said

kwxm:
I'll close this and open a PR for an extended version later.

effectfully (commented):
Great stuff! Thank you for writing it down.

Issue #6183, opened 2024-06-06
PlutusTx PIR simplifier huge memory consumption


koslambrou:
### Summary

After updating to the latest `plutus-tx`, we noticed some significant regression in the instances generated by `PlutusTx.makeIsDataIndexed ''C [...]` .

Compilation times skyrocketed and memory usage is between 60-100GB of RAM (I have a lot of memory).

The issue seems to be the simplification process of the script compilation. For now, we reduced as much as possible the simplifier iterations with `PlutusTx.Plugin:max-simplifier-iterations-pir=1`. After that change, memory usage is back to normal.

I don't know exactly the issue, but I started by writing my own `UnsafeFromData` instance for `C` which resulted in the same issue. Then, I noticed that the amount of used memory seemed to be proportional to the number of constructors of C.

### Steps to reproduce the behavior

For a minimal reproductible example:

```
import PlutusTx qualified
import PlutusTx.Prelude hiding (Semigroup (..), unless)
import Prelude qualified as H

data C = A | B | C | D | E | F | G | H | I | J | K | L | M | N | O | P | Q | R

PlutusTx.makeIsDataIndexed ''C [ ('A, 0) , ('B, 1) , ('C, 2) , ('D, 3) , ('E, 4) , ('F, 5) , ('G, 6) , ('H, 7) , ('I, 8) , ('J, 9) , ('K, 10) , ('L, 11) , ('M, 12) , ('N, 13) , ('O, 14) , ('P, 15) , ('Q, 16) , ('R, 17) ]

{-# INLINABLE validator #-}
validator ::
  BuiltinData
  -> BuiltinData
  -> BuiltinData
  -> Bool
validator d _ _ =
  let !_ = PlutusTx.unsafeFromBuiltinData @C d
   in True

unappliedValidator ::
  PlutusTx.CompiledCode (PlutusTx.BuiltinData -> PlutusTx.BuiltinData -> PlutusTx.BuiltinData -> BuiltinUnit)
unappliedValidator =
   $$(PlutusTx.compile [|| \d r c -> check $ validator d r c ||])
```

### System info

Plutus: 88ad493ad94ecdb34196f7a9aef558613bdee8b2

effectfully:
There's no comment or any other indication on how this issue was triaged and why and I haven't seen any internal discussion on when we're going to address the issue and who's responsible for making sure we will, so I'm "untriaging" the issue. Sorry!

effectfully:
[...]

Issue #5966, opened 2024-05-08
UPLC case-of-case improvements


effectfully:
After #5960 we have some left-overs:

1. this example is incorrect and needs to be updated:

```
{- | A special case of case-of-case optimisation: transforms

@
    case ((force ifThenElse) b (constr t) (constr f)) alts
@

into

@
    force ifThenElse b (case (constr t) alts) (case (constr f) alts)
@

This is always an improvement.
-}
```

2. the PR added generation of extra `force`-`delay` pairs, why don't we see it them in the existing golden tests? Those that were affected by the original PR introducing case-of-case

zliu41:
> Plus, there's no explanation of what we do and why we do it.

The name of the transformation is self-explanatory: this is basically case-of-case transformation. The reason it is useful is because it can trigger case-of-known-constructor transformation on `case (constr t)` and `case (constr f)`.

zliu41:
> should we avoid generating force-delay pairs when branches are pure?

Yes, I think we already do that when compiling cases. Forces and delays are only added if at least one of the branches isn't pure.

effectfully:
> The name of the transformation is self-explanatory: this is basically case-of-case transformation. The reason it is useful is because it can trigger case-of-known-constructor transformation on `case (constr t)` and `case (constr f)`.

The name is self-explanatory, but the `force`-`delay` caveat was so obscure that if went through the review process without anybody noticing. This must be documented.

> Yes, I think we already do that when compiling cases.

In the UPLC case-of-case transformation? Where?

zliu41:
> In the UPLC case-of-case transformation? Where?

I think it's in the compilation of `Case` in GHC Core into PIR, but it's the same idea.

effectfully:
Ah, I misunderstood, I thought you were saying it was already implemented for UPLC case-of-case.

PR #6248, opened 2024-06-26
[Optimization] Remove case-of-case

effectfully:
This removes the case-of-case optimization entirely due to it having multiple sources of exponentiality OOming users' machines. See #6183


plutus-core/plutus-ir/src/PlutusIR/Compiler.hs
```diff
@@ -136,14 +135,12 @@ simplifierIteration suffix = do
...
   hints <- view (ccOpts . coInlineHints)
   preserveLogging <- view (ccOpts . coPreserveLogging)
-  cocConservative <- view (ccOpts . coCaseOfCaseConservative)
   rules <- view ccRewriteRules
   ic <- view (ccOpts . coInlineConstants)
 
   pure $ P.NamedPass ("simplifier" ++ suffix) $ fold
       [ mwhen (opts ^. coDoSimplifierUnwrapCancel) $ Unwrap.unwrapCancelPass tcconfig
       , mwhen (opts ^. coDoSimplifierCaseReduce) $ CaseReduce.caseReducePass tcconfig
-      , mwhen (opts ^. coDoSimplifierCaseReduce) $ CaseOfCase.caseOfCasePassSC tcconfig binfo cocConservative noProvenance
```
effectfully, line 146: This is the only Haskell file changed and this line is the idea of this PR.

effectfully:
I'll merge the PR just to unblock people, but please do still review it.

plutus-core/plutus-ir/src/PlutusIR/Compiler.hs
ana-pantilie: So only the PIR case-of-case has this problem, the UPLC one doesn't?

ana-pantilie (commented):
Sorry, I need to clean up my GitHub notifications so I just saw this one.

plutus-core/plutus-ir/src/PlutusIR/Compiler.hs
effectfully: That's a great question, thank you for asking it! Yes, I believe UPLC case-of-case has one of the sources of exponentiality of PIR case-of-case. I've amended #5966 to reflect that.

Unisay: What about the corresponding CLI flag - Should we remove it too?

PR #6272, opened 2024-07-01
[Bug] Fix 'isNormalType' and add 'prop_normalizedTypeIsNormal'

### LINKED ISSUE #6033: Fix `isNormalType` and add a test
`PlutusCore.Check.Normal` is buggy:

```

>>> :set -XTypeApplications
>>> import PlutusCore.MkPlc
>>> import PlutusCore.Default
>>> import PlutusCore.Check.Normal
>>> import PlutusCore.Normalize
>>> import PlutusCore
>>> isNormalType . unNormalized . runQuote . normalizeType $ mkTyBuiltin @_ @[Integer] @DefaultUni @() @TyName ()
False
```

We should fix the bug and make `isNormalType` respect normalization for built-in types which is defined as

```haskell
normalizeUni :: forall k (a :: k) uni tyname. (HasUniApply uni) => uni (Esc a) -> Type tyname uni ()
normalizeUni uni =
  matchUniApply
    uni
    -- If @uni@ is not an intra-universe application, then we're done.
    (mkTyBuiltinOf () uni)
    -- If it is, then we turn that application into normal type application and recurse
    -- into both the function and its argument.
    (\uniF uniA -> TyApp () (normalizeUni uniF) $ normalizeUni uniA)
```

Plus we should add a test that `normalizeType` returns a normal type.

effectfully:
This fixes a bug in `isNormalType` and adds a test. Without the bug fixed the test fails with

```
  PlutusCore.Generators.QuickCheck
    TypesTests
      normalizedTypeIsNormal: FAIL
        *** Failed! 'normalizeType' returned a non-normal type (after 8 tests and 1 shrink):
        k,ty = (*, list unit)
```

as expected.

Resolves #6033.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Check/Normal.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Check/Normal.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE PolyKinds             #-}
 {-# LANGUAGE RankNTypes            #-}
 
 -- | This module makes sure types are normalized inside programs.
@@ -14,22 +15,26 @@ import PlutusPrelude
 
 import PlutusCore.Core
 import PlutusCore.Error
+import PlutusCore.MkPlc (mkTyBuiltinOf)
 
 import Control.Monad.Except
+import Universe.Core (HasUniApply (matchUniApply), SomeTypeIn (..))
 
 -- | Ensure that all types in the 'Program' are normalized.
 checkProgram
-    :: (AsNormCheckError e tyname name uni fun ann, MonadError e m)
+    :: (AsNormCheckError e tyname name uni fun ann, HasUniApply uni, MonadError e m)
     => Program tyname name uni fun ann -> m ()
 checkProgram (Program _ _ t) = checkTerm t
 
 -- | Ensure that all types in the 'Term' are normalized.
 checkTerm
-    :: (AsNormCheckError e tyname name uni fun ann, MonadError e m)
+    :: (AsNormCheckError e tyname name uni fun ann, HasUniApply uni, MonadError e m)
     => Term tyname name uni fun ann -> m ()
 checkTerm p = throwingEither _NormCheckError $ check p
 
-check :: Term tyname name uni fun ann -> Either (NormCheckError tyname name uni fun ann) ()
+check
+    :: HasUniApply uni
+    => Term tyname name uni fun ann -> Either (NormCheckError tyname name uni fun ann) ()
 check (Error _ ty)           = normalType ty
 check (TyInst _ t ty)        = check t >> normalType ty
 check (IWrap _ pat arg term) = normalType pat >> normalType arg >> check term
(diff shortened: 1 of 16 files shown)


plutus-core/plutus-core/src/PlutusCore/Check/Normal.hs
```diff
@@ -43,20 +48,35 @@ check Var{}                  = pure ()
...
-neutralType ty                = Left (BadType (typeAnn ty) ty "neutral type")
+neutralType
+    :: HasUniApply uni
+    => Type tyname uni ann -> Either (NormCheckError tyname name uni fun ann) ()
+neutralType TyVar{}                 = pure ()
+neutralType (TyBuiltin ann someUni) = neutralUni ann someUni
+neutralType (TyApp _ ty1 ty2)       = neutralType ty1 >> normalType ty2
+neutralType ty                      = Left (BadType (typeAnn ty) ty "neutral type")
+
+-- See Note [Normalization of built-in types].
```
effectfully, line 72: The Note:

```
{- Note [Normalization of built-in types]
Instantiating a polymorphic built-in type amounts to applying it to some arguments. However,
the notion of "applying" is ambiguous, it can mean one of these two things:

1. lifting the built-in type to 'Type' and applying that via 'TyApp'
2. applying the built-in type right inside the universe to get a monomorphized type tag
   (e.g. the default universe has 'DefaultUniApply' for that purpose)

We need both of these things. The former allows us to assign types to polymorphic built-in functions
(otherwise applying a built-in type to a type variable would be unrepresentable), the latter is
used at runtime to juggle type tags so that we can avoid @unsafeCoerce@-ing, bring instances in
scope via 'bring' etc -- for all of that we have to have fully monomorphized type tags at runtime.

So in order for type checking to work we need to normalize polymorphic built-in types. For that
we simply turn intra-universe applications into regular type applications during type normalization.

We could go the other way around and "reduce" regular type applications into intra-universe ones,
however that would be harder to implement, because collapsing a general 'Type' into a 'SomeTypeIn'
is harder than expanding a 'SomeTypeIn' into a 'Type'. And it would also be impossible to do in the
general case, 'cause you can't collapse an application of a built-in type to, say, a type variable
into an intra-universe application as there are no type variables there. I guess we could "reduce"
application of built-in types in some cases and not reduce them in others and still make the whole
thing work, but that requires substantially more logic and is also a lot harder to get right.
Hence we do the opposite, which is straightforward.
-}
```

plutus-core/plutus-core/src/PlutusCore/Core/Instance/Eq.hs
```diff
@@ -106,9 +106,9 @@ eqTypeM (TyFun ann1 dom1 cod1) (TyFun ann2 dom2 cod2) = do
     eqM ann1 ann2
     eqTypeM dom1 dom2
     eqTypeM cod1 cod2
-eqTypeM (TyBuiltin ann1 bi1) (TyBuiltin ann2 bi2) = do
+eqTypeM (TyBuiltin ann1 someUni1) (TyBuiltin ann2 someUni2) = do
```
effectfully, line 109: I've also changed some naming and formatting for consistency.

Unisay (approved):

PR #6277, opened 2024-07-02
Add support for ghc910 in hydra

zeme-wana:


zeme-wana:
Re-opening to reset hydra CI

PR #6282, opened 2024-07-04
Fix CsvTable.tsx and LiteralInclde.tsx in docusaurus

zeme-wana:


doc/docusaurus/docusaurus.config.ts
```diff
@@ -9,8 +9,13 @@ const config: Config = {
 
   // Set the production url of your site here
   url: "https://intersectmbo.github.io",
```
zeme-wana, line 11: Should we change this to `https://plutus.cardano.intersectmbo.org`?

ianhanssoniohk (approved):
This looks fine to me, there may be some merit in making these use an ENV variable for the url but it wouldn't really matter practically unless there's intention on deploying in multiple locations or changing in the future.

PR #6285, opened 2024-07-04
[Costing] Increase the cost of constructors of '[]'

effectfully:
This increases the cost for the `[]` and `(:)` constructors of the `[]` data type. It shouldn't really affect anything since all relevant builtins are constant-time anyway, but making individual pieces of the costing machinery more accurate is always good.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
@@ -214,15 +214,25 @@ instance ExMemoryUsage BS.ByteString where
     {-# INLINE memoryUsage #-}
 
 instance ExMemoryUsage T.Text where
-    -- This is slow and inaccurate, but matches the version that was originally deployed.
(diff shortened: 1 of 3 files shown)


plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -253,29 +269,17 @@ instance ExMemoryUsage a => ExMemoryUsage [a] where
...
-        -- have any other cases.
-        combine (CostRose cost1 forest1) (CostRose cost2 forest2) =
-            CostRose (cost1 + cost2) (forest1 ++ forest2)
-        {-# INLINE combine #-}
-
-        sizeData d = combine nodeMem $ case d of
+        dataNodeRose = singletonRose 4
+        {-# INLINE dataNodeRose #-}
+
+        sizeData d = addConstantRose dataNodeRose $ case d of
```
kwxm, line 275: I think the issue mentioned in the next line is PLT-1176 now.

kwxm (approved):
This looks fine.

PR #6299, opened 2024-07-10
Move  cross-compiling logic from project.nix to cabal

zeme-wana:
In Plutus, several packages cannot be built on Windows (e.g., plutus-benchmark, plutus-tx-plugin, plutus-conformance). Currently, we mark these packages as not buildable in nix/project.nix using haskell.nix. However, this creates two inconsistent sources of truth: cabal.project and nix/project.nix. This PR addresses the issue by moving the relevant logic from nix/project.nix to the corresponding .cabal files.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -58,6 +58,17 @@ if impl(ghc < 9.0) || os(windows)
   -- won't actually build it.
   allow-older: plutus-cert:base
 
+if os(windows)
+   -- When cross compiling we don't have a `ghc` package, so use
+  -- the `plutus-ghc-stub` package instead.
+  package plutus-tx-plugin
+    flags: +use-ghc-stub
+
+  -- Exclude tests that use `doctest`.  They will not work for
+  -- cross compilation and `cabal` will not be able to make a plan.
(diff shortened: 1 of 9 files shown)


cabal.project
```diff
@@ -58,6 +58,17 @@ if impl(ghc < 9.0) || os(windows)
   -- won't actually build it.
   allow-older: plutus-cert:base
 
+if os(windows)
```
zeme-wana, line 61: Moved as-is from nix/project.nix

effectfully (approved):
No idea how it works, hence here's your approval.

plutus-tx-plugin/plutus-tx-plugin.cabal
```diff
@@ -100,7 +104,7 @@ library
     build-depends: ghc >=9.2
```
bezirg, line 104: This `build-depends: ghc >= 9.2` should be reduced to `build-depends: ghc` because we guard already the version using `ghc-version-support`

bezirg (approved):

PR #6300, opened 2024-07-10
Analyse script events supports PlutusLedgerLanguage V3

Unisay:
- [x] Added support for the `PlutusLedgerApi.V3` script events.
- [x] `PlutusEvent` is a product type with the `PlutusLedgerLanguage` as a parameter instead of a sum type.


zliu41 (approved):
Looks fine

PR #6301, opened 2024-07-11
Kwxm/costing/bitwise 4

kwxm:
This adds costing benchmarks for the new bitwise builtins and a cost model based on the benchmark results.  I ran a large number of exploratory benchmarks at various scales and with various configurations (eg, extension semantics v. truncation semantics for `andByteString` and different indices for `readBit`).  The final benchmarks are designed to measure the worst case behaviour and to be fairly accurate for "small" inputs (mostly up to 1200 bytes). The models based on these benchmarks generally extrapolate well to larger inputs.  The attached graph is fairly typical (the blue points are benchmark times for large inputs, the red points are the times for the final benchmarks, and the orange line is the line of best fit through the red points).

ComplementByteString


plutus-core/testlib/PlutusCore/Generators/Hedgehog/Builtin.hs
```diff
@@ -100,6 +103,10 @@ genConstant tr
     , Just HRefl <- eqTypeRep trList (typeRep @[]) =
         case genConstant trElem of
             SomeGen genElem -> SomeGen $ Gen.list (Range.linear 0 10) genElem
+    | trList' `App` trElem <- tr
```
kwxm, line 106: @effectfully Does this look right to you?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1887,89 +1893,90 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
-            (runCostingFunTwoArguments . unimplementedCostingFun)
-
-    -- Bitwise
+            (runCostingFunTwoArguments . paramReplicateByte)
 
     toBuiltinMeaning _semvar ShiftByteString =
-        let shiftByteStringDenotation :: BS.ByteString -> Int -> BS.ByteString
-            shiftByteStringDenotation = Bitwise.shiftByteString
+        let shiftByteStringDenotation :: BS.ByteString -> IntegerCostedLiterally -> BS.ByteString
+            shiftByteStringDenotation s (IntegerCostedLiterally n) = Bitwise.shiftByteString s (fromIntegral n)
```
effectfully, line 1951: This is a major bug I believe. `fromIntegral` wraps around:

```
>>> fromIntegral (toInteger (maxBound :: Int) + 1) :: Int
-9223372036854775808
```

and costing won't save you because the builtin is linear in its first argument and the second one is irrelevant costing-wise. Meaning this builtin now accepts large integers, runs and misbehaves without throwing, while previously it'd throw a runtime operational error along the lines of "the given `Integer` isn't in the bounds of `Int`".

And even if this isn't a bug, you can't use `fromIntegral`. Quoting the docs in this module

```
    YOU MUST NEVER USE 'fromIntegral' AND SIMILAR FUNCTIONS THAT CAN SILENTLY UNDER- OR OVERFLOW
    WHEN DEFINING A BUILT-IN FUNCTION
```

It's way too general and can easily result in overflow/underflow if `Int` is changed to `Integer` in the type signature of a builtin like you did with `replicateByte`.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1887,89 +1893,90 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
         in makeBuiltinMeaning
             shiftByteStringDenotation
-            (runCostingFunTwoArguments . unimplementedCostingFun)
+            (runCostingFunTwoArguments . paramShiftByteString)
 
     toBuiltinMeaning _semvar RotateByteString =
-        let rotateByteStringDenotation :: BS.ByteString -> Int -> BS.ByteString
-            rotateByteStringDenotation = Bitwise.rotateByteString
+        let rotateByteStringDenotation :: BS.ByteString -> IntegerCostedLiterally -> BS.ByteString
+            rotateByteStringDenotation s (IntegerCostedLiterally n) = Bitwise.rotateByteString s (fromIntegral n)
```
effectfully, line 1959: Same.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -168,19 +170,43 @@ instance ExMemoryUsage () where
...
+{- | A wrapper for Integers whose "memory usage" for costing purposes is the
+   absolute value of the integer.  This is used for costing built-in functions
+   such as `shiftByteString` and `rotateByteString`, where the cost may depend
+   on the actual value of the shift argument, not its size.  If this is used to
+   wrap an argument in the denotation of a builtin then it *MUST* also be used
+   to wrap the same argument in the relevant budgeting benchmark.
+-}
+newtype IntegerCostedLiterally = IntegerCostedLiterally { unIntegerCostedLiterally :: Integer }
+instance ExMemoryUsage IntegerCostedLiterally where
+    memoryUsage (IntegerCostedLiterally n) = singletonRose . fromIntegral $ abs n
```
effectfully, line 200: Handling of negative values is kinda weird. So it's not quite a `CostedLiterally` integer?

plutus-core/testlib/PlutusCore/Generators/Hedgehog/Builtin.hs
effectfully: Yeah, looks correct to me. Except the range is somewhat short, but maybe that's how it's supposed to be, don't know. In the bright future I'll kill all of it and create a proper `QuickCheck` generator for all this stuff.

plutus-ledger-api/test/Spec/CostModelParams.hs
```diff
@@ -36,8 +36,11 @@ tests =
                 assertBool "tripping v2 cm params failed" $ Just p == readParamName (showParamName p)
             for_ v3_ParamNames $ \ p ->
                 assertBool "tripping v3 cm params failed" $ Just p == readParamName (showParamName p)
-    , testCase "default values costmodelparamsfortesting" $ do
-            defaultCostModelParamsForTesting @=? Just (toCostModelParams V3.costModelParamsForTesting)
+-- *** FIXME !!! *** : The introduction of the new bitwise builtins has messed
+-- this up because defaultCostModelParamsForTesting is the cost model parameters
+-- for model C, which now includes the new bitwise builtins.
+--    , testCase "default values costmodelparamsfortesting" $ do
+--            defaultCostModelParamsForTesting @=? Just (toCostModelParams V3.costModelParamsForTesting)
```
effectfully, line 43: What's this about?

effectfully (changes requested):
Modulo what I think is a bug, this all is great. I didn't dive too deep into the costing stuff though, just a shallow review.

plutus-core/cost-model/print-cost-model/Main.hs
```diff
@@ -55,6 +69,16 @@ renderTwoVariableQuadraticFunction
...
+-- FIXME.  This is arguably slightly incorrect because some of the arguments are
+-- wrapped in newtypes that change the memory usage instance of their content
+-- and this isn't reflected in the output.  We're able to fix this for
+-- LiteralInYOrLinearInZ since we can tell from the constructor that the second
+-- argument is wrapped, but this doesn't work for `replicateByte`, where the
+-- replication count is wrapped in NumBytesCostedAsNumWords and we don't see that
+-- here.  We should really print "x bytes" instead of just "x", but to fix that
+-- we'd need access to the signatures of the builtins here as well.  Maybe it
+-- could be argued that the user should be aware of the wrappings and interpret
+-- the output accordingly, but it would be helpful to make it explicit.
```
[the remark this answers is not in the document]
kwxm: This isn't a terribly important program so I didn't put the effort in to fix it.  I'll think about what to do though.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -467,11 +470,40 @@ deriving via AsInteger Word64 instance HasConstantIn DefaultUni term =>
...
+  => KnownTypeAst tyname DefaultUni (ListCostedByLength a)
+deriving newtype instance
+  forall term tyname a .
+  ( Contains DefaultUni a
+  , HasConstantIn DefaultUni term
+  , KnownTypeAst tyname DefaultUni a
+  )
+  => MakeKnownIn DefaultUni term (ListCostedByLength a)
+deriving newtype instance
+  forall term tyname a .
```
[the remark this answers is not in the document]
kwxm: I think I just tried that randomly and it worked!  I was quite surprised when it did.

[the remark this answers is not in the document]
kwxm: No, wait.  I think that's why we need `AllowAmbiguousTypes`.  Without that, at line 492 I get 
```
plutus-core/src/PlutusCore/Default/Universe.hs:492:3: error: [GHC-39999]
    • Could not deduce ‘KnownTypeAst tyname0 DefaultUni a’
      from the context: (Contains DefaultUni a,
                         HasConstantIn DefaultUni term, KnownTypeAst tyname DefaultUni a)
        bound by a stand-alone deriving instance declaration:
                   forall term tyname a.
                   (Contains DefaultUni a, HasConstantIn DefaultUni term,
                    KnownTypeAst tyname DefaultUni a) =>
                   MakeKnownIn DefaultUni term (ListCostedByLength a)
        at plutus-core/src/PlutusCore/Default/Universe.hs:(492,3)-(497,55)
      The type variable ‘tyname0’ is ambiguous
       <etc. etc. etc.>
```

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
kwxm: > So it's not quite a CostedLiterally integer?
No, not exactly.  This gets used for `shiftByteString` where you are allowed to shift by negative amounts, and the amount of work you do depends on the size of the shift (there's less copying to do if a lot of bits fall off the end and most of the result is zeros).  I couldn't think of a better name though.

plutus-core/testlib/PlutusCore/Generators/Hedgehog/Builtin.hs
kwxm: > Except the range is somewhat short,
I just copied that from the case for `[]` without thinking.  I'm not sure exactly which tests these generators get used in, so it's not clear to me whether we need bigger lists.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -7,6 +7,7 @@
 -- to test that some constraints are solvable
 {-# OPTIONS -Wno-redundant-constraints #-}
 
+{-# LANGUAGE AllowAmbiguousTypes      #-}
```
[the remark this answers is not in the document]
kwxm: It seems that we do need that.  See [below](https://github.com/IntersectMBO/plutus/pull/6301/files/1c763cdb1e3d8894b1fff529e3c3f24e0e6602bf#r1677237769).

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
kwxm: Oops! Thanks for pointing that out.  I'll fix it.

plutus-core/plutus-core/src/PlutusCore/Bitwise.hs
```diff
@@ -590,12 +590,20 @@ writeBits bs changelist = case unsafeDupablePerformIO . try $ go of
...
       emit "replicateByte: negative length requested"
       evaluationFailure
-  | otherwise = pure . BS.replicate len $ w8
+  | len > integerToByteStringMaximumOutputLength = do
+      emit . pack $ "replicateByte: requested length is too long (maximum is "
+               ++ show integerToByteStringMaximumOutputLength
+               ++ " bytes)"
+      emit $ "Length requested: " <> (pack . show $ len)
+      evaluationFailure
+  | otherwise = pure . BS.replicate (fromIntegral len) $ w8
```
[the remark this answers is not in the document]
kwxm: > Why the change from Int to Integer and this fromIntegral call?

The reason for that is that we already had the `NumBytesCostedAsNumWords` wrapper for one of the arguments to `integerToByteString` and I needed it again here, but `integerToByteString` takes an `Integer` argument so that's what it was wrapping.   Here the underlying function took an `Int` so I would have needed a separate wrapper (I think).  Maybe we should introduce a version of `fromIntegral` that fails if the argument doesn't fit into an `Int` just to avoid any unpleasant surprises.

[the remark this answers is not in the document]
kwxm: I've put this back to the way it was before and changed the contents of `NumBytesCostedAsNumWords` to `Int`.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
kwxm: I think this should be fixed now, but we need to see what happens with #6309 before merging this PR.

plutus-ledger-api/test/Spec/CostModelParams.hs
kwxm: The increased flexibility of the cost models means that some invariants that were being tested aren't invariants any more (for example, it's no longer the case that the Vn cost model parameters are a subset of the V(n+1) ones).  There's an issue to fix this (#6089) but for the time being I've left some commented-out tests so that we know what to fix when we get round to it.

kwxm:
@effectfully I think I've addressed your earlier concerns about `fromIntegral`.  I've also updated this to incorporate some implementation changes from the last week, so it should be mergeable without any problems.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -168,19 +170,43 @@ instance ExMemoryUsage () where
...
-   contain `w` bytes, allowing its costing function to work properly.
+   contain `w` bytes, allowing its costing function to work properly.  We also
+   use this for `replicateByte`.  If this is used to wrap an argument in the
+   denotation of a builtin then it *MUST* also be used to wrap the same argument
+   in the relevant budgeting benchmark.
 -}
-newtype LiteralByteSize = LiteralByteSize { unLiteralByteSize :: Integer }
-instance ExMemoryUsage LiteralByteSize where
-    memoryUsage (LiteralByteSize n) = singletonRose . fromIntegral $ ((n-1) `div` 8) + 1
+newtype NumBytesCostedAsNumWords = NumBytesCostedAsNumWords { unNumBytesCostedAsNumWords :: Int }
```
effectfully, line 186: Sounds like we should do bounds checking manually, since we're interested in a smaller range than `Integer` for bitwise builtins, so we don't really need to convert stuff to `Int` at the interface level, so this probably should be `Integer`. I don't know though and I don't think it should block this PR.

plutus-core/plutus-core/test/CostModelSafety/Spec.hs
```diff
@@ -109,7 +110,8 @@ smallConstant tr
     | Just HRefl <- eqTypeRep tr (typeRep @Integer) = SomeConst (0 :: Integer)
     | Just HRefl <- eqTypeRep tr (typeRep @Int) = SomeConst (0 :: Integer)
     | Just HRefl <- eqTypeRep tr (typeRep @Word8) = SomeConst (0 :: Integer)
-    | Just HRefl <- eqTypeRep tr (typeRep @LiteralByteSize) = SomeConst (0 :: Integer)
+    | Just HRefl <- eqTypeRep tr (typeRep @NumBytesCostedAsNumWords) = SomeConst (0 :: Integer)
+    | Just HRefl <- eqTypeRep tr (typeRep @IntegerCostedLiterally) = SomeConst (0 :: Integer)
```
effectfully, line 114: Oh, I should update these tests in my PR as well...

effectfully (commented):
LGTM, but I didn't review the costing benchmarks properly.

Let me quickly take a look at the default universe module, I feel like it can be simplified.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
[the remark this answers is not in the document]
effectfully: Done in [this](https://github.com/IntersectMBO/plutus/pull/6301/commits/bb345fcb26df1079170ef9b7b766781854e1657e) commit.

effectfully (approved):

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
kwxm: I think I went back and forth on whether this should be `Integer` or `Int` and eventually came down on the side of `Int` because this currently gets used for `IntegerToByteString` and `replicateByte`, and in both of these it specifies a number of bytes (as the name suggests) and for that not to fit into an `Int` you'd be wanting more than something like 37 petabytes, which seems unlikely. If we made it into `Integer` then we'd have to add a range check and conversion to the implementation of `replicateByte`, but that's very fast anyway so the added cost wouldn't be important.  I'll merge this as is, but I may come back and change it to `Integer` for consistency.

PR #6302, opened 2024-07-11
fix: parameter 33 should be a Rational in the constitution script

RSoulatIOHK:
Changed parameter 33 to rational in accordance with the newer conway.cddl Updated the testing suite
Updated the documentation
Generated the new golden test numbers

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


zliu41 (approved):
This looks fine, but I think we should either add `nonnegative_interval`, or rename `unit_interval` to `interval` or `ratio`.

`unit_interval` is supposed to be in the range of [0,1], and if that's not true, then it doesn't make sense to call it _unit_ interval.

bezirg:
> This looks fine, but I think we should either add `nonnegative_interval`, or rename `unit_interval` to `interval` or `ratio`.
> 
> `unit_interval` is supposed to be in the range of [0,1], and if that's not true, then it doesn't make sense to call it _unit_ interval.

Yes, that is the most sensible. I will remain them to `interval` for now, and perhaps in a next iteration we can have two separate types: `unit_interval` and `nonnegative_interval`

PR #6305, opened 2024-07-12
Address guardrail script audit comments

zliu41:
Addresses

- 2.2.1.1■ Location of lookupUnsafe utility function
- 2.2.2.1■ Use traceError


cardano-constitution/src/Cardano/Constitution/Validator/Common.hs
```diff
@@ -37,7 +36,7 @@ withChangedParams fun (scriptContextToValidGovAction -> validGovAction) =
     case validGovAction of
         Just cparams ->  if fun cparams
             then BI.unitval
-            else BI.trace "ChangedParams failed to validate" (B.error ())
```
bezirg, line 40: Actually, come to realise, my old trace was wrong: the message would not be shown since trace's second argument (error) would run before trace could run. This is because of strict semantics.

The correct way would be to `error (trace msg ())` , aka traceError

PR #6307, opened 2024-07-12
fix: types in defaultConstitution

RSoulatIOHK:
More descriptive type names in the defaultConstitution. Some types had names that could imply some additional properties that were wrong. For example, type 33 was a `unit_interval` but is actually a nonnegative_interval as it can be over 1.
Added all the type names used in conway.cddl to be 1:1 with their naming convention.

No added checks on the types have been added to not change the logic of the script. The script generated with this fix and the one currently in `master` should be the same.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


cardano-constitution/src/Cardano/Constitution/Config/Instance/FromJSON.hs
```diff
@@ -118,13 +118,20 @@ parseParamValue = \case
       parseTypedParamValue = withObject "RawParamValue" $ \o -> do
           String ty <- o .: typeKey
           case ty of
-              "integer"       -> RawParamInteger <$> (o .: predicatesKey)
+              "integer"              -> RawParamInteger <$> (o .: predicatesKey)
```
bezirg, line 121:
```suggestion
      parseTypedParamValue :: Value -> Parser RawParamValue
      parseTypedParamValue = withObject "RawParamValue" $ \o -> do
          ty <- o .: typeKey
          parseSynonymType ty o

      -- the base types we support
      parseBaseType :: Key -> Object -> Parser RawParamValue
      parseBaseType ty o = case ty of
          "integer"              -> RawParamInteger <$> (o .: predicatesKey)
          -- NOTE: even if the Tx.Ratio.Rational constructor is not exposed, the 2 arguments to the constructor
          -- will be normalized (co-primed) when Tx.lift is called on them.
          -- SO there is no speed benefit to statically co-prime them ourselves for efficiency.
          "rational"             -> RawParamRational <$> (o .: predicatesKey)
          "any"                  -> pure RawParamAny
          _                      -> fail "invalid type tag"

      -- synonyms to ease the transition from cddl
      parseSynonymType = \case
          "coin"                 -> parseBaseType "integer"
          "uint.size4"           -> parseBaseType "integer"
          "uint.size2"           -> parseBaseType "integer"
          "uint"                 -> parseBaseType "integer" -- For ex units
          "epoch_interval"       -> parseBaseType "integer" -- Rename of uint.size4
          "unit_interval"        -> parseBaseType "rational"
          "nonnegative_interval" -> parseBaseType "rational"
          "costMdls"             -> parseBaseType "any"
          x -> parseBaseType x -- didn't find synonym, try as basetype
```

bezirg (approved):

bezirg:
I will add a changelog later

PR #6309, opened 2024-07-15
Fix overflow bug in shiftByteString, rotateByteString, add tests to ensure it stays fixed

kozross:
This repairs an issue spotted by @effectfully and @kwxm in both `shiftByteString` and `rotateByteString`. Specifically, this happened due to `abs minBound = minBound` for `Int`, which would lead to incorrect behaviour if `shiftByteString` or `rotateByteString` was given that value as a shift or rotation argument. More specifically, we added the following:

1. An additional check in `shiftByteString`, via `Integer`, to ensure that this case gets properly trapped;
2. An additional check in `rotateByteString`, via `Integer`, to do the same thing;
3. Test that verifies that any such shift or rotation does not crash, instead giving a sensible result.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Bitwise.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Bitwise.hs
@@ -11,6 +11,7 @@ module PlutusCore.Bitwise (
   integerToByteStringWrapper,
   byteStringToIntegerWrapper,
   shiftByteStringWrapper,
+  rotateByteStringWrapper,
   -- * Implementation details
   IntegerToByteStringError (..),
   integerToByteStringMaximumOutputLength,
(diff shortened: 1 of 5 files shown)


kwxm (approved):
This looks OK, but now I'm wondering if it could be simplified, especially `rotateByteString`.  It might be easier to pass in the size of the shift/rotation as an `Integer` and sidestep the automatic conversion to `Int`.  

* For `shiftByteString`, check if the absolute value of the shift is greater than the bitlength of the bytestring.  If it is, return 0x00...00.  If the shift is smaller than the bitlength then it's bound to fit into an `Int` since a bytestring with `maxBound::Int` bits would have to be something like 37 petabytes long (but I suppose we should check anyway), so you can cast it and proceed as before.
* For `rotateByteString`, suppose you're rotating the bytestring `s` by `b::Integer` bits .  This will be exactly  the same as rotating by `b' = b mod (8*length s)` and `b'` will never be negative (note `mod`,  not `rem`).  Again, `b'` will surely fit into an `Int` (otherwise `s` would be massive) so we can cast to an `Int` and proceed as before, but `negativeRotate` will no longer be needed since `b >= 0`.

I'm not going to demand that we do this, but I think it would completely avoid the problematic edge cases.

kwxm (approved):
This looks fine to me.  I'm going to merge it so that we can move on, but I may open a separate PR to add a couple of comments about why it's safe to use`fromIntegral`.

kwxm:
Oh, I can't merge it now because we apparently now require all commits to be signed.

PR #6311, opened 2024-07-17
Preliminary Milestone 3 benchmark

kozross:
Draft of the 8-queens benchmark using bitwise primops.


plutus-benchmark/bitwise/src/PlutusBenchmark/NQueens.hs
```diff
@@ -0,0 +1,77 @@
...
+{-# LANGUAGE MultiWayIf        #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module PlutusBenchmark.NQueens (nqueens) where
+
+import PlutusTx.Builtins (replicateByte)
+import PlutusTx.Prelude
+
+-- Based on Qiu, Zongyan (February 2002). "Bit-vector encoding of n-queen problem". ACM SIGPLAN Notices. 37 (2): 68–70
+-- For simplicity, this only accepts multiples of 8 for the dimension (so 8, 16,
```
kwxm, line 11: It might be an interesting exercise to extend this to work for arbitrary n.  Let's not do that now though.

kwxm (approved):
This all looks good, and it's a lot smaller than I was expecting.  It's entirely our fault that it can't be compiled to Plutus Core at the moment and I'm sure that this'll work in PLC when we fix that.

PR #6312, opened 2024-07-17
Release 1.31.0.0

Unisay:
Result of `./scripts/prepare-release.sh "1.31.0.0"`

* I had to remove the `create-json-envelope` executable as it depends on the `cardano-api` which in turn depends on the `plutus` making it a circular dependency hindering the plutus version upgrade.

### DIFF
--- a/cardano-constitution/cardano-constitution.cabal
+++ b/cardano-constitution/cardano-constitution.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-constitution
-version:            1.30.0.0
+version:            1.31.0.0
 license:            Apache-2.0
 license-files:
   LICENSE
@@ -77,10 +77,10 @@ library
     , base               >=4.9   && <5
     , containers
     , filepath
-    , plutus-core        ^>=1.30
-    , plutus-ledger-api  ^>=1.30
-    , plutus-tx          ^>=1.30
-    , plutus-tx-plugin   ^>=1.30
+    , plutus-core        ^>=1.31
+    , plutus-ledger-api  ^>=1.31
+    , plutus-tx          ^>=1.31
+    , plutus-tx-plugin   ^>=1.31
     , regex-tdfa
     , safe
     , template-haskell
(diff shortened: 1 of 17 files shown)


cardano-constitution/cardano-constitution.cabal
```diff
@@ -74,13 +74,13 @@ library
...
-    , plutus-ledger-api  ^>=1.30
-    , plutus-tx          ^>=1.30
-    , plutus-tx-plugin   ^>=1.30
+    , plutus-core
+    , plutus-ledger-api
+    , plutus-tx
+    , plutus-tx-plugin
     , regex-tdfa
     , safe
     , template-haskell
```
bezirg, line 86: Can u put back the version numbering?

cardano-constitution/cardano-constitution.cabal
```diff
@@ -110,15 +110,15 @@ test-suite cardano-constitution-test
...
+    , cardano-constitution
     , containers
     , directory
     , filepath
-    , plutus-core:{plutus-core, plutus-core-testlib}  ^>=1.30
-    , plutus-ledger-api                               ^>=1.30
-    , plutus-tx                                       ^>=1.30
+    , plutus-core:{plutus-core, plutus-core-testlib}
+    , plutus-ledger-api
+    , plutus-tx
```
bezirg, line 121: here also

cardano-constitution/cardano-constitution.cabal
bezirg: I guess the prepare-release script does pick and update those, right?

bezirg (changes requested):

cardano-constitution/cardano-constitution.cabal
Unisay: No, I believe that buy not putting an explicit version boundary we're saying: the current version from this cabal project.

cardano-constitution/cardano-constitution.cabal
```diff
@@ -127,13 +127,3 @@ test-suite cardano-constitution-test
...
-
-executable create-json-envelope
-  import:         lang, ghc-version-support, os-support
-  hs-source-dirs: create-json-envelope
-  main-is:        Main.hs
-  build-depends:
-    , base
-    , cardano-api           ^>=8.48
-    , cardano-constitution
-    , plutus-ledger-api
```
bezirg, line 139: Is it ok for you to resurrect this and its accompanying Main.hs file with a `buildable:False` temporary solution?

cardano-constitution/cardano-constitution.cabal
Unisay: CHaP might need the explicit bounds, restoring them.

cardano-constitution/cardano-constitution.cabal
Unisay: Trying if this works...

bezirg (approved):

PR #6313, opened 2024-07-17
Re-export bitwise builtins from PlutusTx.Prelude

kozross:
This re-exports CIP-121, CIP-122 and CIP-123 operations from `PlutusTx.Prelude`.


kwxm:
Thanks.

PR #6315, opened 2024-07-18
[Test] Improve distribution of generated integers

effectfully:
As the title says this improves distribution of generated integers, so that we more often hit important edge cases such as `2 ^ 16`, `2 ^ 32 - 1`, `2 ^ 32`, `2 ^ 64` etc. The whole approach isn't particularly scientific, I was just tweaking the magical numbers in the code until I was satisfied with the distribution.


plutus-core/plutus-ir/test/PlutusIR/Parser/Tests.hs
```diff
@@ -79,23 +79,46 @@ aroundSeparators = go False False
                 pure $ a : s1 ++ b : s2 ++ rest
             | otherwise -> (a :) <$> go inQuotedName inUnique splice (b : l)
 
+-- | Check whether the given constant can be scrambled (in the sense of 'genScrambledWith').
+isScramblable :: PLC.Some (PLC.ValueOf PLC.DefaultUni) -> Bool
+isScramblable (PLC.Some (PLC.ValueOf uni0 x0)) = go uni0 x0 where
```
effectfully, line 84: I hate the verbosity and awkwardness of this definition and the `isPrintable` one, but I'm going to refrain from spending time on trying to make it nicer. Ultimately, we should split the universe into the type-level and term-level ones and this will make this definition quite nicer.

plutus-core/testlib/PlutusCore/Generators/Hedgehog/AST.hs
```diff
@@ -101,20 +113,10 @@ genBuiltin :: (Bounded fun, Enum fun) => AstGen fun
 genBuiltin = Gen.element [minBound .. maxBound]
 
 genConstant :: AstGen (Some (ValueOf DefaultUni))
-genConstant = Gen.choice
-    [ pure (someValue ())
-    , someValue @Integer <$> Gen.integral_ (Range.linear (-10000000) 10000000)
-    , someValue <$> Gen.utf8 (Range.linear 0 40) Gen.unicode
-    ]
+genConstant = arbitrary
```
effectfully, line 116: This is the gist of the PR.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
```diff
@@ -55,44 +56,69 @@ The 'Arbitrary' instances for 'Integer' and 'Int64' only generate small integers
...
+of each of those ranges.
+-}
 
-    >>> import Data.Int
-    >>> fmap (any ((> 10000) . abs) . concat . concat . concat) . sample' $ arbitrary @[[[Int64]]]
-    True
+-- See Note [QuickCheck and integral types].
+nextInterestingBound :: Integer -> Integer
+nextInterestingBound 1 = 127
+nextInterestingBound x = (x + 1) ^ (2 :: Int) * 2 - 1
```
kwxm, line 68: This kind of works accidentally because of the way the sizes of the fixed-width integral types happen to grow. I was going to suggest having a function with an explicit case for each width, but having thought about it that would probably be really horrible.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
```diff
@@ -55,44 +56,69 @@ The 'Arbitrary' instances for 'Integer' and 'Int64' only generate small integers
     >>> fmap (any ((> 30) . abs) . concat . concat . concat) . sample' $ arbitrary @[[[Int]]]
     False
 
-We want to at least occasionally generate some larger ones, which is what the 'Arbitrary'
-instance for 'Int64' does:
+We want to generate larger ones, including converted-to-Integer 'minBound' and 'maxBound' of various
```
kwxm, line 59: Would it be feasible to get a sensible generator by gluing together the ones for `Int8`, `Int16`, ...?  I've no idea how those behave at the endpoints, so maybe that wouldn't work.

kwxm: Hmmm.  It says above that "The 'Arbitrary' instances for 'Integer' and 'Int64' only generate small integers", but the generators for `Int` and `Int64` seem to be different:
```
ghci> fmap (any ((> 30) . abs) . concat . concat . concat) . sample' $ arbitrary @[[[Int]]]
False
ghci> fmap (any ((> 30) . abs) . concat . concat . concat) . sample' $ arbitrary @[[[Int64]]]
True
```
However
```
ghci> fmap (any ((> 30000) . abs) . concat . concat . concat) . sample' $ arbitrary @[[[Int16]]]
False
```
`maxBound` is 32768, so it looks as if it's not reaching the top.  Trying to use that probably wouldn't be helpful then.

kwxm (approved):
* This seems to do what we want, although it's difficult to tell by looking at the generators: the test output's pretty convincing though.  It might be useful to see how often it comes up with 1 and -1 since they're  kind of edge cases too, but surely they're bound to turn up in the `Int8` numbers.

* Also, are we ever likely to want to do the same for `Word` values?  Probably not: I don't think we use them very much in situations where the ends of the ranges would be significant.

* I recommend waiting until #6350 is merged and then rebasing on top of that to reduce the danger of conflicts in the costing code.

effectfully:
>  It might be useful to see how often it comes up with 1 and -1 since they're kind of edge cases too, but surely they're bound to turn up in the `Int8` numbers.

They're at the bounds, so they generated often:

```
         2.44% -1
         2.38% 1
```

I've added generation of those stats to `test_IntegerDistribution`.

> Also, are we ever likely to want to do the same for Word values? Probably not: I don't think we use them very much in situations where the ends of the ranges would be significant.

Good points. Maybe. But I'm not gonna bother with it in this PR.

plutus-core/plutus-core/test/Spec.hs
```diff
@@ -222,7 +223,8 @@ text, hopefully returning the same thing.
 -}
 propParser :: Property
 propParser = property $ do
-  prog <- TextualProgram <$> forAllPretty (runAstGen genProgram)
+  prog <- TextualProgram <$>
+    forAllPretty (runAstGen $ discardIfAnyConstant (not . isPrintable) genProgram)
```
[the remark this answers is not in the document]
effectfully: Quite possibly, I don't know the numbers and I can't quite figure out how to make `hedgehog` tell them to me. I'll probably create a ticket to address this separately.

plutus-core/plutus-ir/src/PlutusIR/Core/Plated.hs
```diff
@@ -115,6 +118,23 @@ bindingIds f = \case
                     <.> PLC.theUnique f n
                     <.*> traverse1Maybe ((PLC.varDeclName . PLC.theUnique) f) vdecls)
 
+-- | Get all the direct constants of the given 'Term' from 'Constant's.
```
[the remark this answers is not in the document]
effectfully: It just focuses on the contents of the `Constant` constructor (if any) and allows you to either extract it or map it with a function. Actually, looking at it, it's a more general `Prism'` rather than merely a `Traversal'`, but I'm not gonna bother changing it, since a bunch of others need to be changed as well then. We have plenty of those traversals and if they didn't work properly, we'd know that for sure, even just this one would cause tests to fail very quickly, since none of the `isPrintable` logic would work.

plutus-core/plutus-ir/test/PlutusIR/Parser/Tests.hs
```diff
@@ -79,23 +79,46 @@ aroundSeparators = go False False
...
+    go (PLC.DefaultUniProtoPair `PLC.DefaultUniApply` uniA `PLC.DefaultUniApply` uniB) (x, y) =
+        go uniA x && go uniB y
+    go (f `PLC.DefaultUniApply` _ `PLC.DefaultUniApply` _ `PLC.DefaultUniApply` _) _  =
+        noMoreTypeFunctions f
+    go PLC.DefaultUniData _ = True
+    go PLC.DefaultUniBLS12_381_G1_Element _ = False
+    go PLC.DefaultUniBLS12_381_G2_Element _ = False
+    go PLC.DefaultUniBLS12_381_MlResult _ = False
+
 genScrambledWith :: MonadGen m => m String -> m (String, String)
```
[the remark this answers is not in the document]
effectfully: Inserts whitespace around things, I presume. I'm not really sure either, it was here before I started touching these things and I'm gonna resist disturbing it.

plutus-core/testlib/PlutusCore/Test.hs
```diff
@@ -139,6 +141,27 @@ checkFails :: Property -> IO ()
...
+    go TPLC.DefaultUniUnit _ = True
+    go TPLC.DefaultUniBool _ = True
+    go (TPLC.DefaultUniProtoList `TPLC.DefaultUniApply` uniA) xs =
+        all (go uniA) xs
+    go (TPLC.DefaultUniProtoPair `TPLC.DefaultUniApply` uniA `TPLC.DefaultUniApply` uniB) (x, y) =
+        go uniA x && go uniB y
+    go (f `TPLC.DefaultUniApply` _ `TPLC.DefaultUniApply` _ `TPLC.DefaultUniApply` _) _ =
+        noMoreTypeFunctions f
+    go TPLC.DefaultUniData _ = True
+    go TPLC.DefaultUniBLS12_381_G1_Element _ = False
```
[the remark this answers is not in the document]
effectfully: Right, I've changed the function to `isSerialisable` and decided to keep assuming that unserialisable implies unprintable even though it's not actually the case.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
```diff
@@ -86,6 +90,24 @@ defaultBuiltinCostModelExt = (defaultBuiltinCostModelForTesting, ())
    semantics variant?
 -}
 
+test_IntegerDistribution :: TestTree
+test_IntegerDistribution =
+    QC.testProperty "distribution of 'Integer' constants" . QC.withMaxSuccess 10000 $
+        \(AsArbitraryBuiltin i) ->
```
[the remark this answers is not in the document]
effectfully: Fair enough, added a type annotation.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
effectfully: > This kind of works accidentally because of the way the sizes of the fixed-width integral types happen to grow.

Probably not gonna change anytime soon! I'm keeping it as is for now, we can always change it later.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
effectfully: > "The 'Arbitrary' instances for 'Integer' and 'Int64' only generate small integers"

That's a typo. Thanks! Should be `'Integer' and 'Int'`, you can see that it's `Integer` and `Int` in the examples right below that line. Fixed the typo.

PR #6317, opened 2024-07-18
Change writeBits to take changelists as two separate lists

kozross:
This modifies `writeBits` to take two lists as arguments: one of indexes, another of values to set those indexes to.


kozross:
Not quite sure what happened there. I forgot to sign my commits, so I did a rebase with signing and ended up here.

kwxm:
Looks good.

PR #6318, opened 2024-07-18
[Costing] Remove the 'ExMemoryUsage' instance for 'SomeTypeIn'

effectfully:
The instance seems to be unused and it's wrong, so I'm removing it.


kwxm (approved):
Yeah, why would we ever want to know the size of a type in UPLC?

PR #6327, opened 2024-07-24
[Builtins] Split 'KnownBuiltinTypeIn' out of 'DefaultKnownBuiltinTypeIn'

effectfully:
This renames `KnownBuiltinTypeIn` to `DefaultKnownBuiltinTypeIn` (as it's only useful in the default methods of `MakeKnownIn` and `ReadKnownIn`) and defines a new `KnownBuiltinTypeIn` that forces dispatching of the outermost `HasTermLevel` instance, which ensures that GHC does not provide it as a dictionary to the builtins code, which would make it necessary to do the dispatching at runtime. I don't really know if it helps performance, but it's easy and I've definitely seen GHC not dispatching constraints when it can, thereby slowing things down.


effectfully:
Screenshot from 2024-07-30 10-28-44

A clear improvement and the Core looks perfect finally, but the improvement is only detectable in very specialized benchmarks and I'm not really sure if it's worth the additional boilerplate. We really should just split the universe into a type-level and a term-level and rip all these benefits and then quite some more, plus simplify the builtins machinery by a lot.

I'll think about it a bit more, but most likely I'm going to close this PR in the end.

effectfully:
OK, clearly not worth it.

PR #6328, opened 2024-07-24
[Benchmark] Reduce the number of benchmarks

effectfully:
Since we advise to run benchmarks multiple times anyway, there's no need in making them long-running, hence let's run less stuff. Plus, let's not run something that finishes way too quickly, since it's likely that those results are noise.

### DIFF
--- a/plutus-benchmark/lists/bench/Bench.hs
+++ b/plutus-benchmark/lists/bench/Bench.hs
@@ -44,11 +44,11 @@ benchmarks ctx =
       mkBMsForSort name f =
         bgroup name $ sizesForSort <&> \n ->
           bench (show n) $ benchTermCek ctx (f n)
-      sizesForSort = [10, 20..500]
+      sizesForSort = [50, 100..300]
       mkBMsForSum name f =
         bgroup name $ sizesForSum <&> \n ->
           bench (show n) $ benchTermCek ctx (f [1..n])
-      sizesForSum = [10, 50, 100, 500, 1000, 5000, 10000]
+      sizesForSum = [100, 500, 1000, 2500, 5000]
 
 main :: IO ()
 main = do
--- a/plutus-benchmark/nofib/bench/Shared.hs
+++ b/plutus-benchmark/nofib/bench/Shared.hs
@@ -38,11 +38,8 @@ mkBenchMarks (benchClausify, benchKnights, benchPrime, benchQueens) = [
                      , bench "8x8" $ benchKnights 100 8
                      ]
   , bgroup "primetest" [ bench "05digits" $ benchPrime Prime.P5
-                       , bench "08digits" $ benchPrime Prime.P8
                        , bench "10digits" $ benchPrime Prime.P10
-                       , bench "20digits" $ benchPrime Prime.P20
                        , bench "30digits" $ benchPrime Prime.P30
-                       , bench "40digits" $ benchPrime Prime.P40
                        , bench "50digits" $ benchPrime Prime.P50
                        -- Larger primes are available in Primes.hs, but may take a long time.
                        ]


plutus-benchmark/nofib/bench/Shared.hs
```diff
@@ -33,33 +33,30 @@ mkBenchMarks (benchClausify, benchKnights, benchPrime, benchQueens) = [
...
   , bgroup "primetest" [ bench "05digits" $ benchPrime Prime.P5
                        , bench "08digits" $ benchPrime Prime.P8
                        , bench "10digits" $ benchPrime Prime.P10
                        , bench "20digits" $ benchPrime Prime.P20
-                       , bench "30digits" $ benchPrime Prime.P30
-                       , bench "40digits" $ benchPrime Prime.P40
-                       , bench "50digits" $ benchPrime Prime.P50
                        -- Larger primes are available in Primes.hs, but may take a long time.
                        ]
+  , bgroup "queens3x3" [ -- N-queens problem on a 3x3 board
```
kwxm, line 46: I don't think using 3x3 instead of 5x5 will speed things up.  Criterion works by executing your program repeatedly until a time limit (by [default](https://github.com/haskell/criterion/blob/4ac74c9810790e3e83e38da4e7ff3bda15b55501/Criterion/Main/Options.hs#L82) 5 seconds) is exceeded and then it analyses the times that it's collected, so speeding up the program will  just mean that it'll get run more often but the total time will remain the same.  In fact, the time limit for `nofib` is set to [60 seconds](https://github.com/IntersectMBO/plutus/blob/bfac69f8681376b6f7e34df461efec0326b0363a/plutus-benchmark/nofib/bench/Shared.hs#L109) in order to make sure that the examples get run lots of times.  The longest-running nofib example in a [recent run](https://github.com/IntersectMBO/plutus/pull/6301#issuecomment-2222316780) of the nofib benchmarks was `queens5x5/fc`, which ran for 178ms, so it'll have been run about 337 times.  A better way to speed things up might be to leave the benchmarks unaltered but turn the limit down to 30s, or even 20s: 20s would make the entire suite run in one third of the time and still give you over 100 runs of the the slowest example, which would probably be enough to still get accurate results.

Note also that the time limits for `lists` is [15 seconds](https://github.com/IntersectMBO/plutus/blob/bfac69f8681376b6f7e34df461efec0326b0363a/plutus-benchmark/lists/bench/Bench.hs#L56) and for `validation` it's [20 seconds](https://github.com/IntersectMBO/plutus/blob/bfac69f8681376b6f7e34df461efec0326b0363a/plutus-benchmark/validation/bench/Common.hs#L107), so there may be scope for speeding things up a bit there as well.  If you alter the time limits there's maybe a danger that the results would become less reliable because we'd get fewer samples, so I'd recommend running the benchmarks a few times to check whether they become more variable: using `/benchmark` in this branch would be a good way to do that because in the base branch the benchmarks would still be run with the old limits.

plutus-benchmark/lists/bench/Bench.hs
```diff
@@ -44,11 +44,11 @@ benchmarks ctx =
       mkBMsForSort name f =
         bgroup name $ sizesForSort <&> \n ->
           bench (show n) $ benchTermCek ctx (f n)
-      sizesForSort = [10, 20..500]
+      sizesForSort = [50, 100..300]
```
kwxm, line 47: This seems like a good idea.  I think the `lists` benchmarks take a very long time (I'll run them to check), so reducing the number of cases might speed things up significantly.

kwxm (commented):
Try changing the time limits for `nofib` instead.

kwxm (changes requested):
Oops, I think I messed that up.  My overall comment was to try reducing the time limit for `nofib` instead.

kwxm:
Also, if you run a set benchmarks locally and then look at `plutus-benchmark/report.html` there's a lot of information about how long the benchmarks took to run and what the distribution of times was like.  Unfortunately we don't have access to that for benchmarks run on the benchmarking machine (although maybe that could be arranged...).

kwxm:
> /benchmark lists

Of course that won't work properly.  I'll try in a new PR.

kwxm:
I found some `lists` results [here](https://github.com/IntersectMBO/plutus/pull/6327#issuecomment-2249058393).  The slowest benchmark was about 200ms, which would be about 75 runs with the 15 second limit.  That's already quite low, so reducing the time might not be such a good idea. Cutting down the number of tests is probably more effective.

effectfully:
> I don't think using 3x3 instead of 5x5 will speed things up. Criterion works by executing your program repeatedly until a time limit (by [default](https://github.com/haskell/criterion/blob/4ac74c9810790e3e83e38da4e7ff3bda15b55501/Criterion/Main/Options.hs#L82) 5 seconds) is exceeded and then it analyses the times that it's collected

I'd totally forgotten about that! Thank you for spotting. I've restored the original alignment of the benchmarks and simply removed a few of them from `nofib` and a lot of them from `lists`. Please rereview and if you're fine with these changes, merge the PR.

kwxm:
I ran this on my laptop and the time for `nofib` went down from 25 minutes to 22 minutes, which is about what you'd expect given that the nofib benchmarks run for 60 seconds each and you've removed three of them. I think we could safely reduce the time limit to 30s and halve the total time (and remember that when we  run the benchmarks from GitHub they all get run twice).  I might try that in a separate PR.  

The `lists` benchmarks took 19 minutes,  which is a lot less than usual (I think it's about an hour).  Each case runs for 15 seconds, and the faster benchmarks got run tens of thousands of times each; the slowest running ones still got run 153 times, so we might be able to reduce the time limit to 10 seconds without losing too much accuracy.

kwxm (approved):
That's good.  I'll merge it.

PR #6329, opened 2024-07-25
[Plinth] [Builtins] Fix 'writeBits'

effectfully:
This will hopefully eventually fix the issue of `writeBits` being unusable in Plinth.


kwxm:
See #6331.

plutus-tx-plugin/test/Plugin/Profiling/9.6/argMismatch2.eval.golden
effectfully, line 1: Looks like maybe for `Profiling` tests we should turn off showing uniques.

effectfully:
I'm labeling this with "No Changelog Required" since I don't think our users are interested in `writeBits` being fixed, given that it's not quite released properly yet.

kwxm (approved):
Yep, this seems to work.

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -204,6 +204,18 @@ instance HasFromOpaque BuiltinBool Bool where
     fromOpaque b = ifThenElse b True False
     {-# INLINABLE fromOpaque #-}
 
+instance HasToOpaque [BuiltinInteger] (BuiltinList BuiltinInteger) where
```
zliu41, line 207: Why not `instance HasToOpaque a o => HasToOpaque [a] (BuiltinList o)`?

plutus-tx-plugin/test/Plugin/Debug/9.6/letFun.pir.golden
```diff
@@ -8,19 +8,19 @@
       { no-src-span }
       (datatype
         { no-src-span }
-        (tyvardecl { no-src-span } Bool-445 ({ no-src-span } type))
+        (tyvardecl { no-src-span } Bool-451 ({ no-src-span } type))
```
zliu41, line 11: I thought we didn't end up printing uniques for golden files, but apparently we did.

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
```diff
@@ -332,6 +334,12 @@ defineBuiltinTerms = do
     -- Text constant
     defineBuiltinTerm annMayInline 'Builtins.emptyString $ PIR.mkConstant annMayInline ("" :: Text)
 
+    -- List constants
+    defineBuiltinTerm annMayInline 'Builtins.mkNilInteger $
```
zliu41, line 338: `mkNilInteger` and `mkNilBool` are not builtins, so `defineBuiltinTerm` is both a misnomer and unnecessary. How about `` mkNil :: DefaultUni `Contains` a => Proxy a -> BuiltinList a ``, and handle it in the plugin similar to `&&` and `||`?

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
effectfully: I wanted to make it that, but it's pretty hard! We need to have the fully specialized `a` and compute the right type tag from it and it seems like it's above my Template Haskell skills currently. I can probably make it work without TH, but it won't be reliable. In any case, I do intend to create a follow-up polishing this mess a bit, this PR was just for fixing the immediate issue.

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
effectfully: I'm following the existing convention to a T, you can see that it's the same `defineBuiltinTerm` for the empty bytestring right above, I've only replicated that for the empty lists. 

> How about mkNil :: DefaultUni `Contains` a => Proxy a -> BuiltinList a, and handle it in the plugin similar to && and ||?

(see my other comment)

PR #6342, opened 2024-07-25
[Plinth] Ban using 'toBuiltin' and 'fromBuiltin'

effectfully:
It used to be possible to use `toBuiltin`/`fromBuiltin` within a smart contract, but this is no longer the case, but this isn't obvious to the users as they already have code with `toBuiltin`/`fromBuiltin` that now just misbehaves instead of throwing a type error or breaking compilation some other way. This fixes the problem by throwing on any usage of `toBuiltin`/`fromBuiltin` with a suggestion to use `toOpaque`/`fromOpaque` instead.

### DIFF
--- a/plutus-tx/src/PlutusTx/IsData/Class.hs
+++ b/plutus-tx/src/PlutusTx/IsData/Class.hs
@@ -11,10 +11,8 @@
 {-# OPTIONS_GHC -fno-omit-interface-pragmas #-}
 module PlutusTx.IsData.Class where
 
-import Prelude qualified as Haskell (Either (..), Int, error)
+import Prelude qualified as Haskell (Int, error)
 
-import PlutusCore.Crypto.BLS12_381.G1 qualified as BLS12_381.G1
-import PlutusCore.Crypto.BLS12_381.G2 qualified as BLS12_381.G2
 import PlutusCore.Data qualified as PLC
 import PlutusTx.Base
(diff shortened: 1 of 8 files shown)


plutus-tx/src/PlutusTx/IsData/Class.hs
```diff
@@ -159,10 +157,7 @@ instance FromData Builtins.BuiltinBLS12_381_G1_Element where
     fromBuiltinData d =
         case fromBuiltinData d of
           Nothing -> Nothing
-          Just (BI.BuiltinByteString bs) ->
-              case BLS12_381.G1.uncompress bs of
-                Haskell.Left _  -> Nothing
-                Haskell.Right g -> Just $ toBuiltin g
+          Just bs -> Just $ bls12_381_G1_uncompress bs
```
effectfully, line 160: @kwxm This was very wrong, we cannot use Haskell within Plinth and we have an actual Plinth builtin to call here anyway.

zliu41: How was this not noticed before - I guess nobody has been using it?

zliu41 (approved):

plutus-tx/src/PlutusTx/IsData/Class.hs
effectfully: Yeah, I can't think of another explanation.

PR #6344, opened 2024-07-25
Ed25519 benchmark

kozross:
This provides a benchmark for Ed25519 signature verification, as implemented using bitwise primops.

Due to the large volume of code involved, this compiles extremely slowly and uses a huge amount of RAM (~1h on my machine, and 20GiB peak load). However, the benchmark does run:

```
Benchmark bitwise-bench: RUNNING...
benchmarking 8-queens
time                 335.4 ms   (326.5 ms .. 344.2 ms)
                     1.000 R²   (NaN R² .. 1.000 R²)
mean                 342.9 ms   (339.1 ms .. 345.3 ms)
std dev              3.700 ms   (1.648 ms .. 4.737 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarking Ed25519
time                 1.593 s    (1.560 s .. 1.608 s)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.539 s    (1.483 s .. 1.559 s)
std dev              38.70 ms   (595.5 μs .. 47.58 ms)
variance introduced by outliers: 19% (moderately inflated)
```


plutus-benchmark/bitwise/bench/Main.hs
```diff
@@ -1,26 +1,33 @@
...
-main :: IO ()
-main = print "Pending"
-
-{- Currently not able to run, due to problems with writeBits compiling under PlutusTx
 
 main :: IO ()
 main = defaultMain [
-  bench "8-queens" . benchProgramCek mkMostRecentEvalCtx . getPlcNoAnn $ nqueensCompiled
+  bench "8-queens" . benchProgramCek mkMostRecentEvalCtx . getPlcNoAnn $ nqueensCompiled,
+  bench "Ed25519" . benchProgramCek mkMostRecentEvalCtx . getPlcNoAnn $ ed25519Compiled
```
zliu41, line 18: Can you add the golden tests for 8-queens and ed25519? [Here](https://github.com/IntersectMBO/plutus/blob/master/plutus-benchmark/nofib/test/Spec.hs) are some examples.

plutus-benchmark/bitwise/bench/Main.hs
```diff
@@ -1,26 +1,33 @@
...
   ]
 
+
 -- Helpers
 
+ed25519Compiled :: CompiledCode Bool
```
zliu41, line 24: It's better to make `CompiledCode (Signature -> Pk -> Bool)`, then use `unsafeApplyCode` to obtain the `CompiledCode Bool`. This ensures that nothing is evaluated at compile time. [This module](https://github.com/IntersectMBO/plutus/blob/master/plutus-tx-plugin/test/AsData/Budget/Spec.hs) has some examples.

kozross: I guess this also applies to the 8 queens benchmark?

zliu41: Yes that's correct

zliu41 (approved):
This looks good to me. @kwxm can you review `Ed25519.hs` and check if the calculation is done correctly?

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519/Compiled.hs
```diff
@@ -0,0 +1,34 @@
...
+checkValidCompiled = $$(compile [|| \signature msg pk -> checkValid (unsafeFromBuiltinData signature)
+                                                                    (unsafeFromBuiltinData msg)
+                                                                    (unsafeFromBuiltinData pk) ||])
+
+msgAsData :: CompiledCode BuiltinData
+msgAsData = $$(compile [|| toBuiltinData ("hello world" :: BuiltinByteString) ||])
+
+signatureAsData :: CompiledCode BuiltinData
+signatureAsData =
+  $$(compile [|| toBuiltinData ("\NUL\147!x\173\167\209z`\t\243|\195$X$\233\166\234\NUL\134\152l\DC4\243\&4\217\NAK\152\180{$M\227R\214\218%\241\157\ENQ\SO\ENQ\t\152\140\171\240\200f\184\133\203\227z\163\NUL\185\155Y\139\178\249\STX" :: BuiltinByteString) ||])
```
kwxm, line 30: I hate this literal syntax for bytestrings, but that's not your fault!  We sometimes use `pack [0x00, 0x67, ...]`, which is more readable but very fiddly

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519.hs
```diff
@@ -0,0 +1,147 @@
...
+    xA = x * i `modulo` q
+    xB :: Integer
+    xB = q - x
+    xAB :: Integer
+    xAB = q - xA
+    cond1 :: Bool
+    cond1 = (x * x - xx) `modulo` q /= 0 -- x here is always the input x
+    cond2 :: Bool
+    cond2 = if cond1 then odd xA else odd x
+
```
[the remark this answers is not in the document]
kwxm: However, it still seems that the calculation is wrong, although I'm confused by the way that this function interacts with `decodePoint`  For example the values x and q-x are two possible square roots of `xx` and the uses of `even` and `odd` are selecting one of them based on the lowest bit of x, but the selection should in fact be based on the value of one of the bits of the bytestring input to `decodePoint`.  I think that maybe this bit should be fed to `xRecover` and used to select which of x and q-x is returned, and similarly for x' and q-x' (although I see now that `decodePoint` sometimes replaces the x returned by `xRecover` with q-x, so there's some double negation going on which is maybe intended to deal with this:  if so, that's very confusing because the decoding logic is being split over two functions).

In addition, there are elements y of the base field such that there is no x with (x,y) lying on the curve, so given an encoded y-value it may not be possible to find a suitable x, and `xRecover` should fail: this possibility doesn't seem to be accounted for anywhere (I think that `cond` should really be `x * x = -xx modulo q` and `not cond` should be `x * x = xx modulo q` (or vice versa), which leaves cases where neither is true which should lead to failure; furthermore there's a special case when x=0 and the sign bit's set in the encoded bytestring input, which should also fail).

Are you assuming that this code is never applied to an invalid point encoding?  If so, that should be stated in a comment: the code's already difficult to understand and if there are some unstated preconditions that won't help.

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519.hs
```diff
@@ -0,0 +1,147 @@
...
+    x :: Integer
+    x = expMod xx ((q + 3) `divide` 8) q
+    xA :: Integer
+    xA = x * i `modulo` q
+    xB :: Integer
+    xB = q - x
+    xAB :: Integer
+    xAB = q - xA
+    cond1 :: Bool
+    cond1 = (x * x - xx) `modulo` q /= 0 -- x here is always the input x
```
kwxm, line 117: I don't know what the comment means.

kwxm (changes requested):
That was a rather gruelling PR to review.  I'm not going to claim that I checked every single byte of the SHA-512 code, but it looks basically correct and the tests help to convince me that it's OK.

However I found the `decodePoint` and `xRecover` functions in Ed25519.hs very unclear and hard to understand and I think they may in fact be incorrect in general.  It would be very helpful if you could clarify this code a bit.

The enormous compilation time is unfortunate as well, but that's probably not your fault.  At least it's given us a puzzle to solve.

plutus-benchmark/bitwise/test/Main.hs
```diff
@@ -1,12 +1,72 @@
...
+        ]
+      ]
+    ]
   ]
+
+-- Cases
+
+sha512Case :: IO ()
+sha512Case = do
+  let testBS = "hello world"
```
kwxm, line 55: A few more test cases for SHA-512 here wouldn't do any harm given the complexity of the implementation!

zliu41 (changes requested):
@kozross Replace all `INLINE` with `INLINEABLE` in SHA512.hs and the compilation will finish in a few seconds. Inlining everything in that module leads to massive GHC Core.

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519.hs
```diff
@@ -0,0 +1,147 @@
...
+
+newtype Point = Point (Integer, Integer)
+
+instance Eq Point where
+  {-# INLINE (==) #-}
+  Point (x1, y1) == Point (x2, y2) = x1 == x2 && y1 == y2
+
+{-# INLINE bPoint #-}
+bPoint :: Point
+bPoint = Point (bx `modulo` q, by `modulo` q)
```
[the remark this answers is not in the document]
kwxm: ... and infact `bPoint` could be inllined in `checkValid` as well.

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519/Compiled.hs
kozross: The main reason I ended up doing this was because the bytestrings in question were generated using a reference implementation, then 'exported' using `show`. I would have preferred to write them better, but the risk of copy-paste errors was too high.

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519.hs
```diff
@@ -0,0 +1,147 @@
...
+    xBS = integerToByteString LittleEndian 32 x
+    xLSBVal = readBit xBS 248
+
+{-# INLINE scalarMult #-}
+scalarMult :: Point -> Integer -> Point
+scalarMult p e =
+  if e == 0
+    then Point (0, 1)
+    else
+      let q' = scalarMult p (e `divide` 2)
```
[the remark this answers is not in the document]
kozross: This is because `q` is already defined as a top-level constant further down. The naming is unfortunate, but we basically followed the [reference implementation in Python](https://ed25519.cr.yp.to/python/ed25519.py) closely.

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519.hs
```diff
@@ -0,0 +1,147 @@
...
+    cond1 :: Bool
+    cond1 = (x * x - xx) `modulo` q /= 0 -- x here is always the input x
+    cond2 :: Bool
+    cond2 = if cond1 then odd xA else odd x
+
+{-# INLINE clearBit #-}
+clearBit :: Integer -> BuiltinByteString -> BuiltinByteString
+clearBit ix bs = writeBits bs [ix] [False]
+
+{-# INLINE inv #-}
```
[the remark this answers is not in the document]
kozross: The power-of-two thing won't help much: in my experience, GHC doesn't optimize `Integer` divisions (or multiplications, for that matter) into shifts, even where it theoretically could.

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519/Compiled.hs
```diff
@@ -0,0 +1,35 @@
...
+import PlutusTx.Prelude
+import PlutusTx.TH (compile)
+
+checkValidCompiled :: CompiledCode (BuiltinData -> BuiltinData -> BuiltinData -> Bool)
+checkValidCompiled = $$(compile [|| \signature msg pk -> checkValid (unsafeFromBuiltinData signature)
+                                                                    (unsafeFromBuiltinData msg)
+                                                                    (unsafeFromBuiltinData pk) ||])
+
+msgAsData :: CompiledCode BuiltinData
+msgAsData = $$(compile [|| toBuiltinData ("hello world" :: BuiltinByteString) ||])
```
zliu41, line 27: `BuiltinData` can be lifted directly into `CompiledCode`:

```haskell
msgAsData :: CompiledCode BuiltinData
msgAsData = liftCodeDef (toBuiltinData ("hello world" :: BuiltinByteString))
```

zliu41 (approved):
Looks good

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519/Compiled.hs
kozross: This replacement appears fragile. This change here:

```haskell
signatureAsData :: CompiledCode BuiltinData
signatureAsData = liftCodeDef (toBuiltinData ("\NUL\147!x\173\167\209z`\t\243|\195$X$\233\166\234\NUL\134\152l\DC4\243\&4\217\NAK\152\180{$M\227R\214\218%\241\157\ENQ\SO\ENQ\t\152\140\171\240\200f\184\133\203\227z\163\NUL\185\155Y\139\178\249\STX" :: BuiltinByteString))

{-
signatureAsData :: CompiledCode BuiltinData
signatureAsData =
  $$(compile [|| toBuiltinData ("\NUL\147!x\173\167\209z`\t\243|\195$X$\233\166\234\NUL\134\152l\DC4\243\&4\217\NAK\152\180{$M\227R\214\218%\241\157\ENQ\SO\ENQ\t\152\140\171\240\200f\184\133\203\227z\163\NUL\185\155Y\139\178\249\STX" :: BuiltinByteString) ||])
-}
```

produces this golden failure:

```
    test
      9.6
        Ed25519:         FAIL (0.01s)
          Test output was different from 'bitwise/test/9.6/Ed25519.pir.golden'. Output of ["diff","-u","bitwise/test/9.6/Ed25519.pir.golden","/tmp/nix-shell.YeLs4n/Ed25519.pir341253-4.actual"]:
          --- bitwise/test/9.6/Ed25519.pir.golden	2024-08-06 10:37:15.782140753 +1200
          +++ /tmp/nix-shell.YeLs4n/Ed25519.pir341253-4.actual	2024-08-06 10:43:20.606177116 +1200
          @@ -2319,6 +2319,6 @@
                           (/\dead -> ifThenElse {Bool} (equalsInteger y y) True False)
                           (/\dead -> False)
                           {all dead. dead})))
          -  (B #c080c2932178c2adc2a7c3917a6009c3b37cc383245824c3a9c2a6c3aac080c286c2986c14c3b334c39915c298c2b47b244dc3a352c396c39a25c3b1c29d050e0509c298c28cc2abc3b0c38866c2b8c285c38bc3a37ac2a3c080c2b9c29b59c28bc2b2c3b902)
          +  (B #00c2932178c2adc2a7c3917a6009c3b37cc383245824c3a9c2a6c3aa00c286c2986c14c3b334c39915c298c2b47b244dc3a352c396c39a25c3b1c29d050e0509c298c28cc2abc3b0c38866c2b8c285c38bc3a37ac2a300c2b9c29b59c28bc2b2c3b902)
             (B #68656c6c6f20776f726c64)
             (B #283ac3bfc3bbc281372d5e77c3bdc2910b681bc2ab72c2bdc39f2fc395517a62c3b9c2af247ac39371c38311c386)
          \ No newline at end of file
          
          Use -p '$0=="bitwise.Ed25519.test.9.6.Ed25519"' to rerun this test only.
        Ed25519:         OK
        Ed25519:         FAIL (0.02s)
          Test output was different from 'bitwise/test/9.6/Ed25519.budget.golden'. Output of ["diff","-u","bitwise/test/9.6/Ed25519.budget.golden","/tmp/nix-shell.YeLs4n/Ed25519.budget341253-6.actual"]:
          --- bitwise/test/9.6/Ed25519.budget.golden	2024-08-06 10:37:18.085499655 +1200
          +++ /tmp/nix-shell.YeLs4n/Ed25519.budget341253-6.actual	2024-08-06 10:43:20.629510707 +1200
          @@ -1,2 +1,10 @@
          -({cpu: 2137056248968
          -| mem: 7088454370})
          \ No newline at end of file
          +An error has occurred:
          +Caused by: integerToByteString
          +             False
          +             32
          +             -705236757155852610088934725679574536026126424420972475413373104382334263805059295045975968846120247953652472732782828952697926196694920784559513127758387
          +Final budget: ({cpu: 7896928821
          +| mem: 26930816})
          +Logs: integerToByteString: cannot convert negative Integer
          +Input: -705236757155852610088934725679574536026126424420972475413373104382334263805059295045975968846120247953652472732782828952697926196694920784559513127758387
          \ No newline at end of file
          
          Use -p '$0=="bitwise.Ed25519.test.9.6.Ed25519"' to rerun this test only.
        Ed25519:         FAIL (0.01s)
          Test output was different from 'bitwise/test/9.6/Ed25519.eval.golden'. Output of ["diff","-u","bitwise/test/9.6/Ed25519.eval.golden","/tmp/nix-shell.YeLs4n/Ed25519.eval341253-7.actual"]:
          --- bitwise/test/9.6/Ed25519.eval.golden	2024-08-06 10:37:20.305524297 +1200
          +++ /tmp/nix-shell.YeLs4n/Ed25519.eval341253-7.actual	2024-08-06 10:43:20.646177556 +1200
          @@ -1 +1,6 @@
          -(constr 1)
          \ No newline at end of file
          +An error has occurred:
          +Caused by: integerToByteString
          +             False
          +             32
          +             -705236757155852610088934725679574536026126424420972475413373104382334263805059295045975968846120247953652472732782828952697926196694920784559513127758387
          \ No newline at end of file
          
          Use -p '$0=="bitwise.Ed25519.test.9.6.Ed25519"' to rerun this test only.
```

It is specifically this change and no other. The other lifted arguments work fine.

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519.hs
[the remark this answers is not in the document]
kozross: I took this directly from the reference implementation. Due to its complexity, I was afraid to change it. You're likely right, but I'd need an expert opinion on this to be sure.

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519.hs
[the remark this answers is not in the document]
kwxm: Oh, of course!

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519.hs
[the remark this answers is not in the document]
kwxm: Thanks.  I'll take a look at the reference implementation.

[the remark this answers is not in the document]
kwxm: > Thanks. I'll take a look at the reference implementation.

Having looked, I think that what they're doing is returning a potentially invalid x-coordinate from `xRecover` but then later on they have an `isoncurve` check (omitted here) that will reject (x,y) pairs where the x value is wrong.  That still seems a bit suspect, but I should complain to them rather than you.

plutus-benchmark/bitwise/src/PlutusBenchmark/Ed25519/Compiled.hs
kwxm: @zliu41^ Do you have any comment on this?  I don't know much about this stuff.

kwxm (approved):

kwxm:
Oops, I forgot to merge this.

kwxm:
There are conflicts that I can't resolve here. I'll have to check out a new branch and do it there.

kwxm:
There's still a CI failure because of a cabal problem, so I'm going to close this in favour of# 6404 where that's fixed.

Issue #6333, opened 2024-07-25
Plutus Core builtin for modular exponentiation


zliu41:
We need to add the modular exponentiation primitive (CIP-0109) as a built-in function in Plutus Core.  This should be a fairly straightforward process once we have identified a suitable library function.  Note that the library function must be compatible with GHC 8.10 until we drop support for that.

PR #6346, opened 2024-07-26
[plc] Support for `Natural` numbers in the universe, backed by Integer

bezirg:
Prerequisite for introducing #6333 

One argument and the result type is going to be `Natural`
See https://hackage.haskell.org/package/ghc-bignum-1.3/docs/GHC-Num-Integer.html#v:integerPowMod-35-

This is not optimised and left as TODO.

Should work on GHC8.10 onwards. However, Idk yet if the actual builtin will work for GHC8.10.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
@@ -489,6 +489,26 @@ deriving newtype instance KnownBuiltinTypeIn DefaultUni term [a] =>
 deriving newtype instance KnownBuiltinTypeIn DefaultUni term [a] =>
     ReadKnownIn DefaultUni term (ListCostedByLength a)
 
+deriving via AsInteger Natural instance
+    KnownTypeAst tyname DefaultUni Natural
+deriving via AsInteger Natural instance HasConstantIn DefaultUni term =>
+    MakeKnownIn DefaultUni term Natural
+instance HasConstantIn DefaultUni term => ReadKnownIn DefaultUni term Natural where
+    readKnown term =
+        -- See Note [Performance of ReadKnownIn and MakeKnownIn instances].
+        -- Funnily, we don't need 'inline' here, unlike in the default implementation of 'readKnown'
+        -- (go figure why).
+        inline readKnownConstant term >>= oneShot \(i :: Integer) ->
+            -- TODO: benchmark alternatives:signumInteger,integerIsNegative,integerToNaturalThrow
+            if i >= 0
+            -- TODO: benchmark alternatives: ghc8.10 naturalFromInteger, ghc>=9 integerToNatural
+            then pure $ fromInteger i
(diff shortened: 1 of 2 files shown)


plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -489,6 +489,26 @@ deriving newtype instance KnownBuiltinTypeIn DefaultUni term [a] =>
...
+instance HasConstantIn DefaultUni term => ReadKnownIn DefaultUni term Natural where
+    readKnown term =
+        -- See Note [Performance of ReadKnownIn and MakeKnownIn instances].
+        -- Funnily, we don't need 'inline' here, unlike in the default implementation of 'readKnown'
+        -- (go figure why).
+        inline readKnownConstant term >>= oneShot \(i :: Integer) ->
+            -- TODO: benchmark alternatives:signumInteger,integerIsNegative,integerToNaturalThrow
+            if i >= 0
+            -- TODO: benchmark alternatives: ghc8.10 naturalFromInteger, ghc>=9 integerToNatural
+            then pure $ fromInteger i
```
effectfully, line 505: From a quick look `fromInteger` looks less efficient than `integerToNatural`. I doubt it's gonna make a measurable difference though.

effectfully (approved):
Makes sense, so I'm approving it, but given that it doesn't have any testing at all, my preference would be to do it at the same time as adding the builtin that will exercise this code path. It doesn't matter much though, so go ahead and merge.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -489,6 +489,26 @@ deriving newtype instance KnownBuiltinTypeIn DefaultUni term [a] =>
...
+deriving via AsInteger Natural instance
+    KnownTypeAst tyname DefaultUni Natural
+deriving via AsInteger Natural instance HasConstantIn DefaultUni term =>
+    MakeKnownIn DefaultUni term Natural
+instance HasConstantIn DefaultUni term => ReadKnownIn DefaultUni term Natural where
+    readKnown term =
+        -- See Note [Performance of ReadKnownIn and MakeKnownIn instances].
+        -- Funnily, we don't need 'inline' here, unlike in the default implementation of 'readKnown'
+        -- (go figure why).
+        inline readKnownConstant term >>= oneShot \(i :: Integer) ->
```
effectfully, line 501: The code duplication is unfortunate, but I don't have any better ideas either.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
bezirg: I can't use integerToNatural since it is not in GHC8.10 without some macro switching. I will optimise later.

PR #6347, opened 2024-07-26
[Plinth] [Builtins] Add a general 'mkNil'

effectfully:
This replaces the `mkNilInteger`, `mkNilData` etc boilerplate with a single `mkNil` function at the expense of moving the boilerplate to the compiler code. Overall, seems like a wash, but the new way of doing things is perhaps better and hopefully we'll be able to improve it further.


plutus-tx-plugin/test/Plugin/Debug/9.6/letFun.pir.golden
effectfully, line 1: @Unisay if you have some bandwidth, could you please remove printing of uniques frrom `Debug/*` modules?

plutus-tx-plugin/test/Plugin/Profiling/9.6/addInt.pir.golden
effectfully, line 1: @Unisay and `Profiling/*` ones.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -777,6 +797,17 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
+      | GHC.getName n == mkNilOpaqueName -> case ty of
+          GHC.TyConApp tyCon []
+              | tyCon == GHC.integerTyCon || tyCon == builtinIntegerTyCon ->
+                  pure $ PLC.mkConstant annMayInline ([] @Integer)
+              | tyCon == builtinBoolTyCon -> pure $ PLC.mkConstant annMayInline ([] @Bool)
+              | tyCon == builtinDataTyCon -> pure $ PLC.mkConstant annMayInline ([] @PLC.Data)
+          GHC.TyConApp tyCon [GHC.TyConApp tyArg1 [], GHC.TyConApp tyArg2 []]
+              | (tyCon, tyArg1, tyArg2) == (builtinPairTyCon, builtinDataTyCon, builtinDataTyCon) ->
+                  pure $ PLC.mkConstant annMayInline ([] @(PLC.Data, PLC.Data))
+          _ -> throwPlain $ CompilationError "'mkNil' applied to an unknown type"
```
zliu41, line 810: Print the type here?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -777,6 +797,17 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
     -- <error func> <overall type> <message>
     GHC.Var (isErrorId -> True) `GHC.App` GHC.Type t `GHC.App` _ ->
       PIR.TyInst annMayInline <$> errorFunc <*> compileTypeNorm t
+    GHC.Var n `GHC.App` GHC.Type ty
+      | GHC.getName n == mkNilOpaqueName -> case ty of
+          GHC.TyConApp tyCon []
```
zliu41, line 802: This can be made more general to allow, e.g., `BuiltinList (BuiltinList Integer)`, but it's not easy to do, and I'm fine with this

zliu41 (approved):

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
effectfully: I did think of that, but we don't do it in other places, e.g.

```haskell
  GHC.LitString _ -> throwPlain $ UnsupportedError "Literal string (maybe you need to use OverloadedStrings)"
```

```haskell
  index <- case elemIndex dc dcs of
    Just i -> pure i
    Nothing ->
      throwPlain $
        CompilationError "Data constructor not in the type constructor's list of constructors"
```

```haskell
    GHC.LitAlt _ -> throwPlain $ UnsupportedError "Literal case"
```

so I assumed that the context is getting printed anyway and we don't need to print the type manually.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
effectfully: Do you have any specific ideas on how to do that? I was thinking of something like having this:

```haskell
mkNilOpaque :: PLC.ValueOf PLC.DefaultUni [FromBuiltin arep] -> BuiltinList arep
mkNilOpaque _ = BuiltinList []
{-# OPAQUE mkNilOpaque #-}

class MkNil arep where
    mkNil :: BuiltinList arep
    default mkNil :: PLC.DefaultUni `PLC.HasTermLevel` (FromBuiltin arep) => BuiltinList arep
    mkNil = Magic.oneShot mkNilOpaque (PLC.ValueOf PLC.knownUni ([] @(FromBuiltin arep)))
instance MkNil BuiltinInteger
instance MkNil BuiltinBool
instance MkNil BuiltinData
instance MkNil (BuiltinPair BuiltinData BuiltinData)
```

and then spotting `mkNilOpaque`, checking that it's applied to a `ValueOf` and `unsafeCoerce`ing the whole thing to the type that it's supposed to have, i.e. `PLC.ValueOf PLC.DefaultUni [FromBuiltin arep]`, which we can trivially embed into a term as a constant. But that's probably not gonna work, since GHC may pull stuff out of that `ValueOf` value into variables. Except maybe it's somehow fine? 

The only more or less reliable way I could think of is serializing the type at the type level into a `Symbol` (to make sure it's fully done at compile time), requiring it to be `KnownSymbol`, then extracting the resulting `String` in the compiler and parsing it back to get a `DefaultUni` object. Sounds pretty crazy, but I can't think of another way of parsing a Haskell value back after embedding it into a `CoreExpr`. 

Or am I saying nonsense and it can be somehow done more or less reliably without such crazy workarounds?

zliu41: I don't understand your first approach: when the plugin spots `mkNilOpaque` it is applied to a `CoreExpr` - what do you use `unsafeCoerce` on?

What I had in mind was more similar to your second approach - via serialising to strings, but now I doubt it actually works, so never mind.

PR #6350, opened 2024-07-27
Make NumBytesCostedAsNumWords use Integer instead of Int

kwxm:
The `NumBytesCostedAsNumWords` wrapper contained an `Int`, but this changes it to `Integer` for consistency with the other wrappers.  This change also affects the type of `Bitwise.replicateByte`. 

@effectfully There were some conflicts between the recently-committed code for bitwise costing and your PR #6315.  I hope that this will make it easier to reconcile the two.

### DIFF
--- a/plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Conversion.hs
+++ b/plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Conversion.hs
@@ -407,7 +407,6 @@ i2bCipExamples = [
 i2bLimitTests ::[TestTree]
 i2bLimitTests =
     let maxAcceptableInput = 2 ^ (8*Bitwise.maximumOutputLength) - 1
-        maxAcceptableLength = Bitwise.maximumOutputLength -- Just for brevity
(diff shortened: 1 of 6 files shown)


plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Conversion.hs
```diff
@@ -47,7 +47,7 @@ i2bProperty1 = do
   e <- forAllWith ppShow Gen.bool
   -- We limit this temporarily due to the limit imposed on lengths for the
   -- conversion primitive.
-  d <- forAllWith ppShow $ Gen.integral (Range.constant 0 integerToByteStringMaximumOutputLength)
+  d <- forAllWith ppShow $ Gen.integral (Range.constant 0 Bitwise.maximumOutputLength)
```
effectfully, line 50: I wonder if we should be using the fancy `QuickCheck` constant generator in some of these tests.

effectfully (approved):
Thanks a lot for that!

PR #6351, opened 2024-07-30
Reduce benchmarking time limits

kwxm:
This is a followup to #6328.  Criterion runs each individual benchmark up to some time limit then analyses the times of the different runs.  Our benchmark suites currently get run with different time limits: 60s for `nofib`, 20 for `validation`, and 15 for `lists`.  There  should be scope for reducing these limits without reducing the accuracy of the results too much.  I'm going to do some experiments in this PR to see what happens.  Using `/benchmark`, the benchmarks will be run in the base branch with the old time limits and in this branch with the new ones, so we should be able to get a decent comparison.


kwxm:
This is inconclusive.  There's a lot of variation in the results but it's not clear if that's due to the reduction in benchmarking times or inconsistency in the machine.

PR #6352, opened 2024-07-31
[Builtins] Specialize '[Integer]' in 'geq'

effectfully:
This specializes `geqStep` a bit further, so that handling of `[Integer]` and `[Data]` doesn't require any recursive calls and can get inlined fully. I doubt it's going to make a difference, but I thought it'd be easy enough to check, so why not.


effectfully:
Not worth it.

PR #6353, opened 2024-07-31
[Refactoring] Use 'KnownBuiltinTypeIn' instead of 'HasConstant' consistently

effectfully:
This replaces `HasConstantIn` constraints with `KnownBuiltinTypeIn` ones as we use the latter everywhere else. Just a quick refactoring for consistency. I've also downscaled a few slow tests.


kwxm (approved):
Fine.

PR #6361, opened 2024-08-01
Conformance for replicateByte

kozross:
Conformance cases for `replicateByte`, as per CIP-122.


kwxm (changes requested):
These are affected by the same bug in our Agda code that's causing problems for [writeBits](https://github.com/IntersectMBO/plutus/pull/6362), and I think there's another bug to do with the way that the Agda evaluator handles fixed-width integral types.  Please run
```
cabal test haskell-conformance --test-options "-p /replicateByte/ --accept"
```
to update the golden files and then modify the lists [here](https://github.com/IntersectMBO/plutus/blob/8dcfa54136742a4e13b6db7f9ec461c753ecaaef/plutus-conformance/agda/Spec.hs#L111) to include
```
failingEvaluationTests = [
    "test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-3"
  , "test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-4"
  , "test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-5"
  , "test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-6"
  ]
```
and

```
failingBudgetTests =
  [ "test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-3"
  , "test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-4"
  , "test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-5"
  , "test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-6"
  , "test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-7"
  ]
```

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-5/case-5.uplc.expected
```diff
@@ -0,0 +1 @@
+(program 1.0.0 (con bytestring #ffffffff))
```
kwxm, line 1: This definitely shouldn't be happening, but I think that's because our Agda evaluator is dealing with `Word8` arguments incorrectly.  Updating with the `haskell-conformance` results should turn this into a failure.

kwxm (changes requested):
We have a limitation that you're not allowed a replication count >= 8192.  Maybe we should have a test for that too.

kwxm:
Closed in favour of #6392.

PR #6365, opened 2024-08-01
Conformance for countSetBits

kozross:
Conformance cases for `countSetBits`, as per CIP-123.


kwxm (changes requested):
Maybe add a test or two where the bytestring has more than one bit set.

kwxm:
> Maybe add a test or two where the bytestring has more than one bit set.

Actually I'll just do that myself. I might add a few examples with big inputs to and/or/xor as well.

kwxm (approved):
Looks OK.

PR #6375, opened 2024-08-02
Add a page about different notions of version

zliu41:
This replaces the outdated "Plutus language versions" page, which was written 2 years ago.

Reviewers: ignore the diff outside of `versions.md`.


doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -0,0 +1,68 @@
...
+1.1.0 adds sums-of-products to the language by introducing two new AST node types: `Case` and `Constr`.
+See CIP-85 for more details.
+
+Note that adding new builtin functions does not require a new Plutus Core version.
+Once a new builtin function is added, one can simply start using the new builtin function with an existing Plutus Core version.
+
+## Builtin Semantics Variant
+
+Depending on the protocol version or the ledger language version, we may want to have different behavior of a particular builtin function.
+For example, we might want to fix a bug but need to retain the old buggy behavior for old evaluations, so we have to have a way of knowing which regime we are in.
```
bezirg, line 56: suggestion: "for old evaluations of already-submitted scripts"

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -0,0 +1,68 @@
...
+
+One example is the builtin function `consByteString`.
+If the first argument is outside of the range of `Word8`, the original behavior is to wrap it around.
+The new behavior starting at Plutus V3 is to throw an exception.
+
+When evaluating a standalone program using the `uplc` executable, flag `-S` or `--builtin-semantics-variant` can be used to inform the evaluator which semantics variant to use.
```
bezirg, line 62: This slightly contradicts the earlier:

> There is no need (and nowhere) to specify whether you are writing a Plutus V1, or Plutus V2, or Plutus V3 program

Since indirectly the ledger language version is linked to a builtin-semantics-variant, which is observable even when you run offline programs

bezirg: Suggestion to sth like: "omitting the flag defaults to latest/newest builtin semantics variant for each builtin"

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Plutus Haskell Package Version
+
+Several Plutus components are regularly released as libraries, such as `plutus-core`, `plutus-ledger-api` and `plutus-tx-plugin`.
+These packages are published on the [Cardano Haskell Package repository](https://github.com/IntersectMBO/cardano-haskell-packages) (CHaP), rather than Hackage, Haskell's default package archive.
+Each release has a version, and this is completely orthogonal and irrelevant to all other aformentioned notions of version.
```
bezirg, line 68: suggestion:

"Each release has a version followign a release versioning scheme, completely orthogonal and irrelevant to all other aformentioned notions of version."

bezirg (approved):
LGTM

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -0,0 +1,68 @@
...
+Previously it was also the case that introducing new builtin functions or new Plutus Core versions necessitate a new ledger language version, but this is no longer the case.
+Introducing a new ledger language version is very expensive as it must be maintained forever.
+Going forward we expect the launch of new ledger eras to be the only reason for introducing new ledger language versions.
+
+## Plutus Core Version
+
+Plutus Core version is the usual sense of version pertaining to programming languages - in this instance the Plutus Core language.
+So far there have been two Plutus Core versions: 1.0.0 and 1.1.0.
+1.1.0 adds sums-of-products to the language by introducing two new AST node types: `Case` and `Constr`.
+See CIP-85 for more details.
```
bezirg, line 48: Link to CIP-85

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -0,0 +1,68 @@
...
+
+Note that adding new builtin functions does not require a new Plutus Core version.
+Once a new builtin function is added, one can simply start using the new builtin function with an existing Plutus Core version.
+
+## Builtin Semantics Variant
+
+Depending on the protocol version or the ledger language version, we may want to have different behavior of a particular builtin function.
```
bezirg, line 55: Maybe you should explain also what protocol version is then?

bezirg, line 55: Suggestion:

 `[Ledger Language Version](#ledger-language-version)`

(i don't know if this is the right way to link in markdown)

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -0,0 +1,68 @@
...
+In essence, "Plutus V1", "Plutus V2" and "Plutus V3" are tags that the ledger attaches to validators in a transaction.
+
+From the ledger's perspective, they are in fact totally different languages:
+there is no requirement that they be similar or compatible in any way.
+However, the "V1", "V2" and "V3" naming scheme is practically useful because in reality they have a lot in common (e.g., the underlying language is Plutus Core and programs are executed by the Plutus Core evaluator, regardless of ledger language version); plus it makes it clear the order that they were introduced and the relationships among them.
+
+When simply writing a standalone program, say one that takes two integers and returns their sum, or one that implements a Sudoku solver, the notion of ledger language version is completely irrelevant.
+There is no need (and nowhere) to specify whether you are writing a Plutus V1, or Plutus V2, or Plutus V3 program - again, the concept of ledger language version does not apply to programs themselves.
+You simply write your code in Plutus Tx, compile it to UPLC, and run it with a UPLC evaluator.
```
ana-pantilie, line 21: I believe I understand what this is trying to say but I also think it's a bit ambiguous. It relies on "language in the sense of a programming language" having a clear definition when one is not given here and it's also not common knowledge what it might be.

How can we define what a "language version" is?
Would it make sense to say that the versioning `PlutusV1 ... PlutusVn` refers to the version of the UPLC evaluator? I believe that's not enough, because if `V2` added new constructors to the UPLC AST then it also extended the syntax.
So is it a new UPLC language? Where the syntax can only be extended from `Vn` to `Vn+1` and the semantics can be changed arbitrarily? Or can only the "builtin" semantics be changed arbitrarily? It's not clear to me what constraints we have here, and it would be very good to specify them.

All of these are independent of the ledger, so I'm not so sure of the "_ledger_ language version" naming. I would say that yes the ledger interacts with each language version differently, but the "languages" themselves are different in isolation of the ledger. And I believe we should (or, more appropriately, the ledger should) specify how it "consumes" each language version, since from "Depending on the ledger language version, the ledger will pass different arguments to the validator." I understand that the API differs from one version to the other.

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Plutus Core Version
+
+Plutus Core version is the usual sense of version pertaining to programming languages - in this instance the Plutus Core language.
+So far there have been two Plutus Core versions: 1.0.0 and 1.1.0.
+1.1.0 adds sums-of-products to the language by introducing two new AST node types: `Case` and `Constr`.
+See CIP-85 for more details.
+
+Note that adding new builtin functions does not require a new Plutus Core version.
+Once a new builtin function is added, one can simply start using the new builtin function with an existing Plutus Core version.
```
ana-pantilie, line 51: So extending the semantics of UPLC does not warrant a new language version. Would it be enough to say that any changes to UPLC which are not backwards compatible create a new language version? Then again, why did we need a new language version for SoPs? Doesn't that only extend the language? Or does it also make some backwards-incompatible changes?

doc/docusaurus/docs/essential-concepts/versions.md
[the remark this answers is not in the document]
zliu41: I'll do the renaming in one fell swoop.

zliu41: > Would it make sense to say that the versioning PlutusV1 ... PlutusVn refers to the version of the UPLC evaluator?

No, PlutusVn has nothing whatsoever to do with the UPLC evaluator.

> but the "languages" themselves are different in isolation of the ledger.

No, the languages themselves are not different, which is the entire point I'm trying to make here. How are they different in isolation of the ledger?

doc/docusaurus/docs/essential-concepts/versions.md
zliu41: > Would it be enough to say that any changes to UPLC which are not backwards compatible create a new language version?

What do you mean by "language version"? I'm using the terms "ledger language version" and "Plutus Core version" here. Any changes to UPLC itself, whether backwards compatible or not, creates a new Plutus Core version, but does not create a new ledger language version.

doc/docusaurus/docs/essential-concepts/versions.md
zliu41: Protocol version has nothing to do with Plutus. I will explain such concepts in a glossary page.

doc/docusaurus/docs/essential-concepts/versions.md
zliu41: I don't see a contradiction. When evaluating a standlone program, you don't specify the ledger language version (since there's no ledger), but you can specify the builtin semantics variant. The fact that a particular ledger language version uses a particular builtin semantics variant doesn't contradict anything; it just happens to be that way. It's just a random correlation.

doc/docusaurus/docs/essential-concepts/plutus-platform.mdx
```diff
@@ -61,31 +61,30 @@ graph TD;
...
+Nonetheless, writing the rest of the application contains plenty of complexity.
+Amongst other things, an application needs to deal with the software around the ledger (wallets, nodes, etc.); distributed systems issues such as settlement delays, inconsistent state between parties, and rollbacks; and simple user-experience issues like upgrades, state management and synchronization.
 Furthermore, while none of these are quite as security-critical as the trusted kernel, users certainly *can* be attacked through such applications, and even non-malicious bugs are likely to be quite upsetting when a user's money is at stake.
 
 Even simple applications must deal with this complexity, and for more advanced applications that deal with state across time, the difficulty is magnified.
 
 ## Additional resources
 
-- Michael Peyton-Jones and Jann Mueller introduce the Plutus platform in [this session](https://youtu.be/usMPt8KpBeI?si=4zkS3J7Bq8aFxWbU) from the Cardano 2020 event. 
+- Michael Peyton-Jones and Jann Mueller introduce the Plutus platform in [this session](https://youtu.be/usMPt8KpBeI?si=4zkS3J7Bq8aFxWbU) from the Cardano 2020 event.
```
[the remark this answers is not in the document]
zliu41: Don't bother reviewing anything other than versions.md. A lot of them will be removed soon.

doc/docusaurus/docs/essential-concepts/versions.md
zliu41: I don't find it a good reader experience to put a link in a page that points to a different part of the same page, unless that page is really long.

zliu41 (commented):
> Stylistic suggestion: having all versions (e.g. PlutusV1.., 1.0.0.) inside markdown `code` , PlutusV1, 1.0.0 etc

I disagree - version numbers are not code, and I think most people don't put them in code blocks.

doc/docusaurus/docs/essential-concepts/ledger.md
```diff
@@ -101,7 +101,7 @@ The Cardano ledger currently has a few different kinds of validator scripts:
 
 See [The EUTXO Handbook, A deep dive into Cardano's accounting model](https://www.essentialcardano.io/article/the-eutxo-handbook).
 
-For more help on how to actually implement interesting logic using the EUTXO model and scripts, see: 
+For more help on how to actually implement interesting logic using the EUTXO model and scripts, see:
 
 - [Using Plutus Tx](../category/using-plutus-tx)
 - [Working with scripts](../category/working-with-scripts)
```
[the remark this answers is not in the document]
zliu41: It does exist: https://plutus.cardano.intersectmbo.org/docs/category/working-with-scripts

doc/docusaurus/docs/essential-concepts/versions.md
ana-pantilie: > No, the languages themselves are not different, which is the entire point I'm trying to make here. How are they different in isolation of the ledger?

`PlutusV2` introduces new constructors to the UPLC language, right? What does this have to do with the ledger?

doc/docusaurus/docs/essential-concepts/versions.md
ana-pantilie: So what is the exact relationship between "ledger language" versions and "Plutus Core" versions?

doc/docusaurus/docs/essential-concepts/ledger.md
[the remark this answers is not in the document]
ana-pantilie: Clicking on the links in the rich diff sent me to a page displaying:
```
404 - page not found
The 
zliu41/doc-version

 branch of 
plutus

 does not contain the path 
doc/docusaurus/docs/category/using-plutus-tx.
```

doc/docusaurus/docs/essential-concepts/versions.md
zliu41: > PlutusV2 introduces new constructors to the UPLC language, right? What does this have to do with the ledger?

No it doesn't. New constructors were introduced in a new Plutus Core Version (1.1.0), not in PlutusV2.

doc/docusaurus/docs/essential-concepts/versions.md
zliu41: A new Plutus Core version changes the languages - e.g., Plutus Core 1.1.0 introduces `Case` and `Constr`.

A new ledger language version does not change the language. The main difference is that the scripts will receive different arguments from the ledger.

doc/docusaurus/docs/essential-concepts/ledger.md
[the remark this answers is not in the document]
zliu41: It does work on the doc site, which appears to uses a different path finding mechanism than GH.

doc/docusaurus/docs/essential-concepts/versions.md
ana-pantilie: I understand. I think it would be very helpful to add this to the document as an example. Thank you!

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -0,0 +1,68 @@
...
+This necessitates a new ledger language version, because the arguments passed to existing ledger language versions must stay exactly the same, so that existing validators continue to work.
+
+Previously it was also the case that introducing new builtin functions or new Plutus Core versions necessitate a new ledger language version, but this is no longer the case.
+Introducing a new ledger language version is very expensive as it must be maintained forever.
+Going forward we expect the launch of new ledger eras to be the only reason for introducing new ledger language versions.
+
+## Plutus Core Version
+
+Plutus Core version is the usual sense of version pertaining to programming languages - in this instance the Plutus Core language.
+So far there have been two Plutus Core versions: 1.0.0 and 1.1.0.
```
effectfully, line 46: Please let's call it AST version. Having "Plutus Core version" and "`plutus-core` version" to mean different things is kind of horrible.

doc/docusaurus/docs/essential-concepts/versions.md
effectfully: > Depending on the protocol version or the ledger language version

Might be worth adding "or both" explicitly.

doc/docusaurus/docs/essential-concepts/versions.md
effectfully: > It's just a random correlation.

That's a weird thing to say. The builtin semantics variant is solely determined by the protocol version and the ledger language version, it's basically like a unique hash of the two. I'm against viewing it as some abstract concept that just happens to be there (and I'm against its very existence in the first place).

effectfully (approved):
I disagree with this description a little, because I disagree with the current code structure a lot.

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -0,0 +1,68 @@
...
+---
+
+# Different Notions of Version
+
+There are several different notions of version that Cardano smart contract developers must distinguish.
+
+## Ledger Language Version
+
+This is what "Plutus V1", "Plutus V2", "Plutus V3" refer to.
+It is called "ledger" language version because they are _not_ really different versions of a language in the sense of programming languages, but different languages from the Cardano ledger's point of view; the ledger handles Plutus V1 vs. V2 vs. V3 differently.
```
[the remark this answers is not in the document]
zliu41: None of these are fundamentally why have multiple ledger language versions. E.g., currently they do have different features (different builtins, scott vs. sop), but in the near future those differences will go away. Same with the difference in costing.

The only real difference between V1, V2, V3: they receive different arguments from the ledger.

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -0,0 +1,68 @@
...
+In essence, "Plutus V1", "Plutus V2" and "Plutus V3" are tags that the ledger attaches to validators in a transaction.
+
+From the ledger's perspective, they are in fact totally different languages:
+there is no requirement that they be similar or compatible in any way.
+However, the "V1", "V2" and "V3" naming scheme is practically useful because in reality they have a lot in common (e.g., the underlying language is Plutus Core and programs are executed by the Plutus Core evaluator, regardless of ledger language version); plus it makes it clear the order that they were introduced and the relationships among them.
+
+When simply writing a standalone program, say one that takes two integers and returns their sum, or one that implements a Sudoku solver, the notion of ledger language version is completely irrelevant.
+There is no need (and nowhere) to specify whether you are writing a Plutus V1, or Plutus V2, or Plutus V3 program - again, the concept of ledger language version does not apply to programs themselves.
```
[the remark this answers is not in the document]
zliu41: As I said in my other comment - in the near future all other differences between V1, V2 and V3 will go away, and the only difference left would be: they receive different arguments from the ledger. There is no fundamental reason for there to be different programming language features among V1/V2/V3.

doc/docusaurus/docs/essential-concepts/versions.md
zliu41: > The ledger language version (e.g. PlutusV1) literally determines the AST version (1.0.0 for PlutusV1)

Again, this will no longer be true in the near future. There is no fundamental reason why 1.1.0 cannot be used for PlutusV1.

doc/docusaurus/docs/essential-concepts/versions.md
zliu41: The paragraph above does say "1.1.0 adds sums-of-products to the language by introducing two new AST node types: `Case` and `Constr`.".

doc/docusaurus/docs/essential-concepts/versions.md
zliu41: I think AST version is fine, but it is unusual - it is common for a programming language to have a version - e.g. Python 3.7

doc/docusaurus/docs/essential-concepts/versions.md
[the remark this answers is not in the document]
effectfully: > As I said in my other comment - in the near future all other differences between V1, V2 and V3 will go away, and the only difference left would be: they receive different arguments from the ledger. 

How can this possibly be true? We _compute_ the semantics variant from the ledger language version (and the protocol version), it cannot be true that the only difference is in arguments, since the semantics variant affects the behavior of programs. The semantics variant is a derivative and in my view is a vestige of the past. It's the ledger language version and the protocol version that are real, the semantics variant is just a made up hash of the two.

In particular, V1 (or V2) and V3 _behave differently_ for certain programs with a _fixed_ protocol version. How is that not a difference? Sure we arrange that through the semantics variant, but that's just a detail of the implementation and not at all how we actually look at versioning ("we'll change this behavior in V3").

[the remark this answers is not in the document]
zliu41: The Plutus Core language can exist as a standalone programming language, without the node, ledger, or anything Cardano. It is turing complete and you can use it to do anything. For that reason, builtin semantics variant should be the first-class concept. When we talk about the language without Cardano, there is no need to introduce the concept of PlutusV1, V2, V3 etc.

The overwhelmingly main reason to introduce a new PlutusVx is due to a new ledger era, necessitating changes to the script context or other script arguments or return types. Admittedly, there's another potential reason to introduce a new PlutusVx, without a new ledger era, namely changing the behavior of a builtin function. But crucially, the only reason a new PlutusVx is needed in this case, rather than just changing the behavior at the next Protocol Version, _is still Cardano-related_: because of the nature of the blockchain, we want scripts that are already submitted to continue to have the old behavior, thus new PlutusVx, rather than new Protocol Version.

Without Cardano in the picture, not only would we not need Plutus V1/V2/V3, but we certainly wouldn't name them Plutus V1/V2/V3. How is it a better name than Builtin Semantics Variant V1/V2/V3? The latter is much more specific and accurate, and less confusing.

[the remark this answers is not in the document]
zliu41: btw one more point: the fact that the ledger language version and protocol version must be used together to determine the behaviors of builtin functions is another reason why the ledger language version is irrelevant when we are talking about the language itself, independent of Cardano: protocol versions only exist in Cardano, and without protocol version, the ledger language version does not even determine anything by itself.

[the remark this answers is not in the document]
effectfully: I agree with everything you've said apart from

> Without Cardano in the picture, not only would we not need Plutus V1/V2/V3, but we certainly wouldn't name them Plutus V1/V2/V3. How is it a better name than Builtin Semantics Variant V1/V2/V3? The latter is much more specific and accurate, and less confusing.

(as I'm arguing in those threads that I've referenced earlier, making the semantics variant in Cardano a number/letter is a bad idea that we or our successors will very likely backtrack on. We think of it as a product type and it really should be a product).

I just do not find your point of view helpful for users, i.e. those people for whom these docs are written for. We're talking about Cardano here, why should we generalize anything beyond that?

> without protocol version, the ledger language version does not even determine anything by itself.

Technically,

1. `PlutusV3` solely determines the semantics variant
2. for earlier ledger language versions they determine the _set_ of available semantics variant

But I see your point.

> For that reason, builtin semantics variant should be the first-class concept. 

I agree. I don't think it's helpful for these docs though. In Cardano the semantics variant is a combination of the ledger language version and the protocol version, nothing more than that. Hashing them into that `A`/`B`/`C` thing is just counter-productive and so is saying things like

> When simply writing a standalone program, say one that takes two integers and returns their sum, or one that implements a Sudoku solver, the notion of ledger language version is completely irrelevant.

Replace `addInteger` in that example with `consByteString` (or even `multiplyInteger`!) and it simply falls apart. You're presuming that there's an `addInteger` operation, a very much Cardano-specific one, and then you're saying that the ledger language is completely irrelevant, which is simply not true, because it is relevant for the behavior of builtins. There's no Sudoku solver if you don't know how your builtins are supposed to behave, it's all one giant undefined behavior agent in that case.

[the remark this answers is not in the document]
zliu41: > I just do not find your point of view helpful for users, i.e. those people for whom these docs are written for. We're talking about Cardano here, why should we generalize anything beyond that?

I find it easier to think about this in terms of two layers: (1) what is the language itself and the notions related only to the language; and then (2) how it works with Cardano, which is in principle just one of the applications of Plutus Core. Mixing them together makes it harder to untangle the different concepts.

> Replace addInteger in that example with consByteString (or even multiplyInteger!) and it simply falls apart.

With `consByteString`, the ledger language version is still irrelevant in the sense that... it simply isn't a thing when talking about the language itself (layer 1). But I see your point: when running the script on Cardano, different ledger language versions do lead to different behaviors. My argument still holds but I could've phrase it more accurately.

doc/docusaurus/docs/essential-concepts/versions.md
effectfully: Python 3.7 corresponds to `plutus-core-1.32`, not `1.1.0`.

doc/docusaurus/docs/essential-concepts/versions.md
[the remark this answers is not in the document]
effectfully: > I find it easier to think about this in terms of two layers: (1) what is the language itself and the notions related only to the language; and then (2) how it works with Cardano, which is in principle just one of the applications of Plutus Core.

I agree that thinking in terms of two layers is best for _us_. I really disagree that it's helpful to users. In particular, ironically

> Mixing them together makes it harder to untangle the different concepts.

mixing things up is more likely when you try to untangle them. This shows w.r.t. my point about `consByteString`, but also more generally when you say things like "when simply writing a standalone program". There's no such thing as a program written in some abstract Plutus, it's always some concrete form of Plutus and in this case it's Cardano Plutus (I'm not even sure if other Plutuses exist). Plutus-the-library is a language constructor: you can have your own builtins, your own versioning for them, your own versioning for the AST etc. But the moment you start "writing a program", these become fixed and you can no longer untangle it from its context, i.e. Cardano in our case.

This also shows in

> You simply write your code in Plutus Tx, compile it to UPLC, and run it with a UPLC evaluator.

which, as I said earlier, is merely a bug in the implementation.

So no, I do not agree with

> My argument still holds

and I most certainly do not agree with

> I could've phrase it more accurately.

because these are supposed to be user-friendly docs, not the kind of docs that we write for ourselves with all the gory details and whatnot. All that phrased correctly context is going to be lost on folks who just want to know how to develop smart contracts. E.g. they will just think that the behavior of Plinth scripts is always the same regardless of the ledger language version, which is very much not the case.

Summarizing:

1. I believe that the ledger language version (likely it's "condensed" form) should be an explicit (rather than implicit) part of the semantics variant
2. Plinth users, i.e. Cardano users, should be able to control the ledger language version that they're using for their Plinth scripts right in those scripts
3. while the two-layered model that we've been discussing here is correct, it's counter-productive to put it in these docs as it adds a lot of unnecessary confusion. I really don't feel like this level of detail belongs to an "essential concepts" section

I think one way forward would be to ask the smart contract developers community to read the docs and tell us whether they find them helpful and what their takeaways are.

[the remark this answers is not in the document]
zliu41: >  I really disagree that it's helpful to users

I guess we'll agree to disagree. I definitely find discussing different things separately useful, whether I'm the developer or the user. Not dissimilar to modularity of code. Plus, this is something we are already doing: the documentation has a section talking about "using Plutus Tx", i.e., just language stuff, and another section for "working with scripts", i.e., Cardano-related. I think it's much better than merging them together - I fail to see how the latter improves readability.

> All that phrased correctly context is going to be lost on folks who just want to know how to develop smart contracts

It's IMO fairly straightforward: when you are writing a standlone program, you can use builtin semantics variant to determine the builtin behavior. When you are writing a script, this is determined by (protocol version, ledger language version). Not that hard!

PR #6376, opened 2024-08-05
Restore CI checks for doc site code

zliu41:
The checks were removed during the RTD -> Docusaurus migration.

### DIFF
--- a/doc/docusaurus/docusaurus-examples.cabal
+++ b/doc/docusaurus/docusaurus-examples.cabal
@@ -24,6 +24,28 @@ common ghc-version-support
   if (impl(ghc <9.6) || impl(ghc >=9.7))
     buildable: False
 
+library docusaurus-code
+  import:         lang, ghc-version-support
+  hs-source-dirs: static/code
+
+  if (impl(ghcjs) || os(windows))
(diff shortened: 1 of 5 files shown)


doc/docusaurus/docusaurus-examples.cabal
```diff
@@ -24,6 +24,28 @@ common ghc-version-support
...
+    AuctionValidator
+    BasicPlutusTx
+    BasicPolicies
+    BasicValidators
+
+  build-depends:
+    , base               >=4.9   && <5
+    , plutus-core        ^>=1.31
+    , plutus-ledger-api  ^>=1.31
+    , plutus-tx          ^>=1.31
```
effectfully, line 44: Does the release script know that these need to be updated?

effectfully (approved):

doc/docusaurus/docusaurus-examples.cabal
zliu41: Good point - I don't think so. I'll remove these bounds

zliu41: No I was wrong, it actually does know.

PR #6380, opened 2024-08-06
Add 'VerifiedCompilation' blob

effectfully:
This is the same as #5947 but doesn't change 2453 files.


ramsay-t:
Fixes #5916 and #5944 and #6383

PR #6386, opened 2024-08-07
Hide bitwise internals

kwxm:
I accidentally used the internal  versions of a couple of bitwise functions when I [added them](https://github.com/IntersectMBO/plutus/pull/6368) to `plutus-metatheory` and this led to a [segmentation fault](https://ci.iog.io/build/4981627/nixlog/2) in the conformance tests.  This PR hides the internal implementations and makes sure that we only export the safe versions which check the inputs carefully. I've also renamed some things to reduce the risk of confusion.  The problem only affected the Agda code: the functions that we expose in PLC and PlutusTx were safe.


plutus-core/plutus-core/src/PlutusCore/Bitwise.hs
```diff
@@ -705,8 +704,8 @@ of 8, we can be _much_ faster, as Step 2 becomes unnecessary in that case.
 -}
 
 -- | Shifts, as per [CIP-123](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0123).
-shiftByteString :: ByteString -> Int -> ByteString
-shiftByteString bs bitMove = unsafeDupablePerformIO . BS.useAsCString bs $ \srcPtr ->
+unsafeShiftByteString :: ByteString -> Int -> ByteString
```
zliu41, line 707: Much better naming!

zliu41 (approved):

PR #6392, opened 2024-08-08
Conformance tests for `replicateByte`

kwxm:
This is a continuation of #6361.  We couldn't merge that because the tests were failing because the Agda denotation of `replicateByte` was wrong.  This fixes that.


bezirg (approved):

plutus-metatheory/src/Builtin.lagda.md
```diff
@@ -494,6 +494,17 @@ postulate
 ### What builtin operations should be compiled to if we compile to Haskell
 
 ```
+-- NOTE.  Many of the denotations in PlutusCore.Default.Builtins involve
```
bezirg, line 497: Can you add a name to this? Like our NOTE convention, so I can refer to it also myself

PR #6407, opened 2024-08-13
Add sprint-issues-rolloever.yml workflow

zeme-wana:


.github/workflows/sprint-issues-rollover.yml
```diff
@@ -0,0 +1,24 @@
+# This workflows automatically moves issues and pull requests to the next 
+# iteration of our Plutus Backlog project when a new sprint begins.
+
+name: "🛼 Sprint Issues Rollover"
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 5 * * 1' # 5am on Monday
```
ana-pantilie, line 9: I think the new sprint always starts on Wednesday, every 2 weeks. The new one will start on the 7th. Unfortunately I have no idea what time zone GitHub Projects uses. I couldn't find anything [in our sprint settings](https://github.com/orgs/IntersectMBO/projects/18/settings/fields/109884532) or the [GitHub Projects docs](https://docs.github.com/en/issues/planning-and-tracking-with-projects/understanding-fields/about-iteration-fields).

zliu41:
Is there an easy way to manually trigger this? If so, we can run it during sprint planning.

.github/workflows/sprint-issues-rollover.yml
zeme-wana: Thanks for pointing this out.
I'm not sure it's possible to tell cron "on Wednesday, every *2* weeks", but I will change this to "on Wednesday, every week."
Running the workflow when a new sprint hasn't started yet should be a no-op, so this should be safe to run every Wednesday. 

As for timezone, GitHub primarily uses UTC. Se we can *assume* that both cronjobs and sprint iterations use UTC, and that sprint iterations trigger at midnight. This means that 5am should work fine.

zeme-wana:
> Is there an easy way to manually trigger this? If so, we can run it during sprint planning.

Yes I've added a `workflow_dispatch` so we can test it/run it whenever

Issue #6406, opened 2024-08-13
Fix metatheory/test2


kwxm:
The tests in `plutus-metatheory:test2` are broken (and turned off in CI) because they're comparing a Named AST with a de Bruijn-indexed one.  This should be easy to fix.

kwxm:
In fact part of the problem seems to be that some tests are attempting to use the Haskell parser to read de Bruijn terms.

kwxm:
Done in #6414.

PR #6414, opened 2024-08-14
Kwxm/metatheory/fix test2

kwxm:
`plutus-metatheory:test2` was disabled in CI and would fail if you ran it locally.  This should fix the tests, after a fashion.  Those tests (now renamed to `test-detailed`)  run various Agda and Haskell evaluators on some examples and check that the evaluators agree.  The problem was that the test was attempting to re-parse the output and compare ASTs, but the Agda evaluators can only output textual PLC/UPLC with names (often just `A` or something similar) annotated with de Bruijn indices, and the parsers can't handle those.  To work around this I've modified  the `plc` and `uplc` evaluators to be able to output textual code with de Bruijn indices as well and modified the tests to compare canonicalised versions of the textual output (squashing up all the whitespace and replacing variables with just their de Bruijn indices).  This is kind of horrible, but it does work and saves us from having to worry about alpha equivalence.To fix it properly I think we'd either have to make the Haskell parser accept de Bruijn indices or get the Agda code to output unique names, and neither of these seem especially easy.

I've also rewritten the tests to use Tasty.HUnit, which we use quite a lot elsewhere: previously they were using something called `detailed-0.9`, which seems to be an obscure type of test suite that cabal knows about.  I don't really know what that is, but I've  been able to replace it with `exitcode-stdio-1.0`, which we use everywhere else.

I messed about with `project.nix` to try to re-enable the tests in CI, but I'm not sure if I've done that correctly.

I also deleted some files in the top level of `plutus-metatheory/test` that appear to be unused.


nix/project.nix
```diff
@@ -48,24 +48,21 @@ let
...
-            plutus-metatheory.components.tests.test1.preCheck =
+            plutus-metatheory.components.tests.test-simple.preCheck =
               let
-                cmp = config.hsPkgs.plutus-metatheory.components.tests.test1;
+                cmp = config.hsPkgs.plutus-metatheory.components.tests.test-simple;
                 deps = cmp.executableToolDepends;
               in
               ''PATH=${lib.makeBinPath deps}:$PATH'';
 
-            # FIXME: Somehow this is broken even with setting the path up as above
```
kwxm, line 58: @zeme-wana I think this is what I have to remove to re-enable these tests in CI, but that's a guess.  I'm also not sure what the stuff above about `preCheck` is for or whether we still need it.

kwxm:
This is issue #6406.

nix/project.nix
zeme-wana: Yes, the test2->test-detailed came back and is passing.
As for the `preCheck`, it *might* be obsolete. 
I commented it out, let's see what CI thinks

zeme-wana (approved):

ana-pantilie:
> This is issue https://github.com/IntersectMBO/plutus/issues/6406.

@kwxm you can write "Fixes <link-to-issue/#issueno" to automatically close the issue when the PR gets merged

kwxm:
> > This is issue #6406.
> 
> @kwxm you can write "Fixes <link-to-issue/#issueno>" to automatically close the issue when the PR gets merged

Ah, thanks.  I wondered if there was some way to make a PR and an issue interact sensibly.

plutus-core/executables/plc/Main.hs
```diff
@@ -161,16 +175,20 @@ runApplyToData (ApplyOptions inputfiles ifmt outp ofmt mode) = do
...
       Left (e :: PLC.Error PLC.DefaultUni PLC.DefaultFun ()) ->
         errorWithoutStackTrace $ PP.displayPlc e
-      Right ty                                               ->
-        T.putStrLn (PP.displayPlc ty) >> exitSuccess
+      Right (PLC.Normalized ty)                                  ->
+        if printModeDeBruijn
+        then
+          let w = toDeBruijnTypePLC ty
+          in print (getPrintMethod printMode w) >> exitSuccess
+        else print (getPrintMethod printMode ty) >> exitSuccess
```
ana-pantilie, line 191: Does something like this make sense? Makes it a bit cleaner imo, but feel free to ignore if you don't agree/it doesn't work.

```suggestion
let transformNames = if printModeDeBruijn then toDeBruijnTypePLC else id
in print (getPrintMethod printMode (transformNames ty)) >> exitSuccess
```

plutus-core/executables/plc/Main.hs
```diff
@@ -184,12 +202,17 @@ runOptimisations (OptimiseOptions inp ifmt outp ofmt mode) = do
...
   let evaluate = Ck.evaluateCkNoEmit (PLC.defaultBuiltinsRuntimeForSemanticsVariant semvar)
       term = void $ prog ^. PLC.progTerm
   case evaluate term of
-    Right v  -> print (getPrintMethod printMode v) >> exitSuccess
+    Right v  ->
+      if printModeDeBruijn
+      then
+        let w = toDeBruijnTermPLC v
+        in print (getPrintMethod printMode w) >> exitSuccess
+      else print (getPrintMethod printMode v) >> exitSuccess
```
ana-pantilie, line 215: Same as above?

plutus-core/executables/src/PlutusCore/Executable/Parsers.hs
```diff
@@ -104,7 +104,7 @@ printmode = option auto
   <> showDefault
   <> help
     ("Print mode for textual output (ignored elsewhere): Classic -> plcPrettyClassicDef, "
-     <> "Debug -> plcPrettyClassicDebug, "
+     <> "Simple -> plcPrettyClassicSimple, "
```
ana-pantilie, line 107: Why the renaming?

plutus-core/executables/uplc/Main.hs
```diff
@@ -67,6 +68,7 @@ data EvalOptions =
       Input
       Format
       PrintMode
+      Bool -- Use de Bruijn indices in the output?
```
ana-pantilie, line 71: I personally think it's better to write a new datatype with 2 constructors something like  `ShowNames = UseDeBruijn | UseNames` to avoid "boolean blindness"; it makes it possible to figure out from just the type/constructors what semantics they have, if you use booleans you have to add comments to maintain readability.

plutus-core/executables/uplc/Main.hs
```diff
@@ -339,12 +350,16 @@ runEval (EvalOptions inp ifmt printMode budgetMode traceMode
...
             do
               let (res, budget, logs) = Cek.runCek cekparams bm emitM term
               case res of
                 Left err -> hPrint stderr err
-                Right v  -> writeToFileOrStd outputMode (show (getPrintMethod printMode v))
+                Right v  ->
+                  if printModeDeBruijn
+                  then let w = toDeBruijnTermUPLC v
+                       in writeToFileOrStd outputMode (show (getPrintMethod printMode w))
+                  else writeToFileOrStd outputMode (show (getPrintMethod printMode v))
```
ana-pantilie, line 362: Similar to the above?

ana-pantilie (approved):
Cool, thank you for fixing this! I only have some minor comments, feel free to ignore.

plutus-core/executables/uplc/Main.hs
kwxm: You caught me!  I did think about doing that but was too lazy, partly because I would have had to go and look up the right way to deal with the command-line options.  I'll change it.

plutus-core/executables/plc/Main.hs
kwxm: Unfortunately that doesn't work because `ty` and `toDeBruijnTy ty` have different types (the  type of PLC types is parametric over the type of names, and `toDebruijnType` changes that).  I think we'd have to introduce a new type class to get round that.

plutus-core/executables/src/PlutusCore/Executable/Parsers.hs
kwxm: The names changed in the prettyprinting code a while ago and `plcPrettyClassicDebug` no longer exists but the help message didn't get updated.

plutus-core/executables/plc/Main.hs
ana-pantilie: Looks great, thank you!

ramsay-t (approved):
Looks good to me - handling the parsing of the weird de brujin variables seems ok now, but we might want to figure out where they are coming from eventually.

PR #6434, opened 2024-09-02
[Builtins] Make unlifting lazy again

effectfully:
This is technically a regression performance-wise, but we may already be charging for it (depends on how budgeting calibration works), so let's see, maybe it's not detectable regardless.


effectfully:
Make your bets, is this going to happen again:

image

effectfully:
-2.3% in total. I don't trust it, probably just a glitch, let's see the next one.

effectfully:
-1.6% in total and I still don't believe it. Also note how the `primetest` ones are (insignificant) regressions, those are the most sensitive to changes in the builtins code.

effectfully:
@kwxm needed rebasing on `master`, I've now done it.

effectfully:
@kwxm it works! Thank you. Did you want to make it this though:


| Script                                   |  1d2ad5a   |  989b32c   |   Change   |
| :------| :------: | :------: | :------: |
| TOTAL                                    |  36.34 ms  |  36.04 ms  |    -0.8%   |

?

effectfully:
Interesting!

plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
```diff
@@ -294,13 +294,29 @@ instance
...
+        -- unlifting" (as opposed to "operationally immediate unlifting") or "call-by-name
+        -- unlifting" (as opposed to "call-by-value unlifting"). We do it this way to guarantee that
+        -- the cost of unlifting will be accounted for before unlifting is performed. If we did
+        -- unlifting eagerly here, this would make the node do work that is not accounted for until
+        -- full saturation is reached, which may never happen if the partial application is thrown
+        -- away.
+        --
+        -- The disadvantage of this approach is that @addInteger 42@ will always unlift @42@ upon
+        -- full saturation even if this partial application is saved to a variable. But the way
+        -- costing calibration benchmarks are setup, we always evaluate a single application, so
```
kwxm, line 308:
```suggestion
        -- costing calibration benchmarks are set up, we always evaluate a single application, so
```

plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
```diff
@@ -294,13 +294,29 @@ instance
     -- See Note [One-shotting runtime denotations].
     -- Grow the builtin application within the received action and recurse on the result.
     toMonoF getBoth = BuiltinExpectArgument . oneShot $ \arg ->
-        -- Ironically computing the unlifted value strictly is the best way of doing deferred
```
kwxm, line 297: > Ironically computing

This sounds like some kind of technique: "ironic computation".  We should invent something just so we can call it that.

kwxm (approved):
This seems perfectly reasonable.  It doesn't sound like it's going to impose a massive penalty on typical programs and it's probably better to avoid any possible surprises.

plutus-core/plutus-core/src/PlutusCore/Builtin/Meaning.hs
effectfully: Love the idea.

Unisay (approved):

PR #6439, opened 2024-09-03
[Errors] Swap operational and structural errors

effectfully:
This simply swaps around two type variables and two constructors, affecting `EvaluationError` and `EvaluationException`. Structural errors correspond to type errors and operational errors correspond to runtime errors, hence the former "precede" the latter. This PR finally reflects that perceived ordering. It's not really important, but I've got things mixed up before due to the previously counter-intuitive ordering, so I thought it'd be worth fixing it, given how easy that is.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Error.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Error.hs
@@ -29,29 +29,26 @@ import Data.Bitraversable
 
 {- | The type of errors that can occur during evaluation. There are two kinds of errors:
 
-1. Operational ones -- these are errors that are indicative of the _logic_ of the program being
-   wrong. For example, 'error' was executed, 'tailList' was applied to an empty list or evaluation
-   ran out of gas.
-2. Structural ones -- these are errors that are indicative of the _structure_ of the program being
+1. Structural ones -- these are errors that are indicative of the _structure_ of the program being
    wrong. For example, a free variable was encountered during evaluation, a non-function was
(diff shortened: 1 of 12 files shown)


plutus-core/plutus-core/src/PlutusCore/Evaluation/Error.hs
```diff
@@ -60,34 +60,34 @@ mtraverse makeClassyPrisms
...
-    bifoldMap _ g (StructuralEvaluationError err)  = g err
+    bifoldMap f _ (StructuralEvaluationError err)  = f err
+    bifoldMap _ g (OperationalEvaluationError err) = g err
     {-# INLINE bifoldMap #-}
 
 instance Bitraversable EvaluationError where
-    bitraverse f _ (OperationalEvaluationError err) = OperationalEvaluationError <$> f err
-    bitraverse _ g (StructuralEvaluationError err)  = StructuralEvaluationError <$> g err
+    bitraverse f _ (StructuralEvaluationError err)  = StructuralEvaluationError <$> f err
+    bitraverse _ g (OperationalEvaluationError err) = OperationalEvaluationError <$> g err
```
effectfully, line 74: I've checked, we don't need to update any call sites, because there's only one and `f` and `g` are the same there:

```haskell
        AsUnliftingEvaluationError (EvaluationError structural operational) where
        go =
            prism'
                (bimap
                    (review _UnliftingError)
                    (review _UnliftingError))
                (bitraverse
```

zliu41 (approved):
Also update Haddock?

effectfully:
> Also update Haddock?

Great point, thanks. Done.

PR #6443, opened 2024-09-04
[Errors] Test that evaluation of a well-typed term doesn't fail with a structural error

### LINKED ISSUE #6150: [Epic] Sort out structural vs operational evaluation errors
We have two kinds of evaluation errors: structural ones and operational ones. More on that [here](https://github.com/IntersectMBO/plutus/blob/b61cce8a4c0f598dacaf531a47c00bfc9ce33e72/plutus-core/plutus-core/src/PlutusCore/Evaluation/Error.hs#L30-L56). This issue is for dumping all the things that we could improve upon in this context.

effectfully:
What the title says.

In theory, this is supposed to conclude the investigation on the "runtime type system" for UPLC. In practice unfortunately, PIR generators aren't very helpful here as per the recent Slack discussion (I'll reflect that in some GitHub issue later), plus the compiler being too slow doesn't help either. So this PR is more of a proof of concept than an actually helpful test, but better than nothing I suppose.

Resolves #6150.

### DIFF
--- a/plutus-core/plutus-ir/test/PlutusIR/Generators/QuickCheck/Tests.hs
+++ b/plutus-core/plutus-ir/test/PlutusIR/Generators/QuickCheck/Tests.hs
@@ -7,24 +7,33 @@
 
 module PlutusIR.Generators.QuickCheck.Tests where
 
+import PlutusPrelude
+
 import PlutusCore.Generators.QuickCheck
 import PlutusIR.Generators.QuickCheck
 
+import PlutusCore.Builtin (fromValue)
 import PlutusCore.Default
+import PlutusCore.Evaluation.Machine.ExBudget
+import PlutusCore.Evaluation.Machine.ExBudgetingDefaults (defaultCekParametersForTesting)
 import PlutusCore.Name.Unique
 import PlutusCore.Quote
 import PlutusCore.Rename
+import PlutusCore.Test (toUPlc)
+import PlutusCore.Version (latestVersion)
 import PlutusIR
 import PlutusIR.Core.Instance.Pretty.Readable
+import PlutusIR.Test ()
+import UntypedPlutusCore qualified as UPLC
+import UntypedPlutusCore.Evaluation.Machine.Cek (restricting, runCekNoEmit,
+                                                 unsafeSplitStructuralOperational)
 
+import Control.Exception
 import Control.Monad.Reader
-import Data.Bifunctor
 import Data.Char
 import Data.Either
-import Data.Function
 import Data.Hashable
 import Data.HashMap.Strict qualified as HashMap
-import Data.List.NonEmpty (NonEmpty (..))
 import Data.Map.Strict qualified as Map
 import Test.QuickCheck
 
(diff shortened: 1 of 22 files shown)


plutus-core/plutus-ir/test/PlutusIR/Generators/QuickCheck/Tests.hs
```diff
@@ -184,3 +194,22 @@ prop_noTermShrinkLoops = withMaxSuccess 10 $
   forAllDoc "ty,tm" genTypeAndTerm_
     (\(ty', tm') -> filter ((/= tm') . snd) $ shrinkClosedTypedTerm (ty', tm')) $ \(ty, tm) ->
   tm `notElem` map snd (shrinkClosedTypedTerm (ty, tm))
+
+-- | Check that evaluation of the given term doesn't fail with a structural error.
+noStructuralErrors :: UPLC.Term Name DefaultUni DefaultFun () -> IO ()
+noStructuralErrors term =
+  void . evaluate . unsafeToEvaluationResult . fst $ do
+    let budgeting = restricting . ExRestrictingBudget $ ExBudget 1000000000 1000000000
```
kwxm, line 202: Can you use [restrictingEnormous](https://github.com/IntersectMBO/plutus/blob/200ad0ffe59f841b6cae82995d5f080ba71aab0d/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/ExBudgetMode.hs#L157) here?

plutus-core/plutus-ir/test/PlutusIR/Generators/QuickCheck/Tests.hs
```diff
@@ -184,3 +194,22 @@ prop_noTermShrinkLoops = withMaxSuccess 10 $
   forAllDoc "ty,tm" genTypeAndTerm_
     (\(ty', tm') -> filter ((/= tm') . snd) $ shrinkClosedTypedTerm (ty', tm')) $ \(ty, tm) ->
   tm `notElem` map snd (shrinkClosedTypedTerm (ty, tm))
+
+-- | Check that evaluation of the given term doesn't fail with a structural error.
```
kwxm, line 198: It took some digging through the code to discover how this is checking something, so more explanation might be helpful.  The logic of  [extractEvaluationResult](https://github.com/IntersectMBO/plutus/blob/200ad0ffe59f841b6cae82995d5f080ba71aab0d/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Exception.hs#L98) seems a bit obscure because it's somewhat unexpectedly using both `Left` and `Right` to represent failures and then `unsafeToEvaluationResult` is turning one of those into an actual exception using `unsafeFromEither`, and that's what's coming into play here.  It might be easier to understand if all of this was done in a more explicit way, or at least if there was an explanation here of what's going on.

kwxm (approved):
This looks OK, although it took me a little while to see how some of it worked.

plutus-core/plutus-ir/test/PlutusIR/Generators/QuickCheck/Tests.hs
effectfully: I've now made it

```haskell
-- See the Haddock of 'EvaluationError' for what structural and operational errors are.
-- See Note [Ignoring context in OperationalEvaluationError].
-- | Preserve the contents of an 'StructuralEvaluationError' as a 'Left' and turn an
-- 'OperationalEvaluationError' into a @Right EvaluationFailure@ (thus erasing the content of the
-- error in the latter case).
splitStructuralOperational
    :: Either (EvaluationException structural operational term) a
    -> Either (ErrorWithCause structural term) (EvaluationResult a)
```

and

```haskell
noStructuralErrors :: UPLC.Term Name DefaultUni DefaultFun () -> IO ()
noStructuralErrors term =
  -- Throw on a structural evaluation error and succeed on both an operational evaluation error and
  -- evaluation success.
  void . evaluate . unsafeSplitStructuralOperational . fst $ do
    <...>
```

I feel like this is sufficiently clear now. `extractEvaluationResult` was just some old name back from the days when the distinction between structural and operational errors wasn't really defined.

cabal.project
```diff
@@ -76,10 +76,6 @@ if os(windows)
 package bitvec
   flags: -simd
 
--- See the note on nix/agda-packages.nix for why this is here.
```
zeme-wana, line 79: Obsolete: `adga-pacakges.nix` uses `cabalProjectLocal` to add `ieee` and `filemanip`

cabal.project
```diff
@@ -76,10 +76,6 @@ if os(windows)
...
--- (NOTE this will change to ieee754 in newer versions of nixpkgs).
-extra-packages: ieee, filemanip
```
zeme-wana, line 81: Obsolete: `adga-pacakges.nix` uses `cabalProjectLocal` to add `ieee` and `filemanip`

PR #6449, opened 2024-09-06
Enable Chang+1 builtins at Chang+1 and add them to V3.ParamName

zliu41:


plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
```diff
@@ -120,11 +120,14 @@ builtinsIntroducedIn = Map.fromList [
...
-  ((PlutusV3, futurePV), Set.fromList [
+  ((PlutusV3, changPlus1PV), Set.fromList [
           AndByteString, OrByteString, XorByteString, ComplementByteString,
           ReadBit, WriteBits, ReplicateByte,
           ShiftByteString, RotateByteString, CountSetBits, FindFirstSetBit,
-          Ripemd_160, ExpModInteger
+          Ripemd_160
+          ]),
+  ((PlutusV3, futurePV), Set.fromList [
+          ExpModInteger
```
kwxm, line 130: @bezirg Note that this has changed.

kwxm (approved):
I think this looks OK.  I'm going to mention #6089 again though, just to remind us to look at it some time).

kwxm:
I guess this doesn't actually enable the new builtins: I think that doesn't happen until the appropriate cost model parameters appear in the protocol parameters.  It does enable enabling the builtins though.

PR #6450, opened 2024-09-06
Add a compiler flag for the evaluateBuiltins pass

zliu41:
This flag is currently missing, `_posDoSimplifierEvaluateBuiltins` is always true and cannot be turned off.

### DIFF
--- a/plutus-tx-plugin/src/PlutusTx/Options.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Options.hs
@@ -157,8 +157,8 @@ pluginOptions =
           desc =
             "When conservative optimisation is used, only the optimisations that \
             \never make the program worse (in terms of cost or size) are employed. \
-            \Implies `no-relaxed-float-in`, `no-inline-constants`, and \
-            \`preserve-logging`."
+            \Implies `no-relaxed-float-in`, `no-inline-constants`, \
+            \`no-simplifier-evaluate-builtins`, and `preserve-logging`."
        in ( k
           , PluginOption
               typeRep
--- a/plutus-tx-plugin/src/PlutusTx/Options.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Options.hs
@@ -171,10 +171,12 @@ pluginOptions =
               , Implication (== True) posPreserveLogging True
               , Implication (== True) posCaseOfCaseConservative True
               , Implication (== True) posInlineConstants False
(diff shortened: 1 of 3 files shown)


plutus-tx-plugin/src/PlutusTx/Options.hs
```diff
@@ -171,10 +171,12 @@ pluginOptions =
               , Implication (== True) posPreserveLogging True
               , Implication (== True) posCaseOfCaseConservative True
               , Implication (== True) posInlineConstants False
+              , Implication (== True) posDoSimplifierEvaluateBuiltins False
```
ana-pantilie, line 174: It gets turned off here because it may increase program size?

ana-pantilie (approved):

plutus-tx-plugin/src/PlutusTx/Options.hs
zliu41: Because it may change semantics: https://github.com/IntersectMBO/plutus/issues/6167

plutus-tx-plugin/src/PlutusTx/Options.hs
```diff
@@ -171,10 +171,12 @@ pluginOptions =
...
               , Implication (== False) posRelaxedFloatin True
               , Implication (== False) posPreserveLogging False
               , Implication (== False) posCaseOfCaseConservative False
               , Implication (== False) posInlineConstants True
+              , Implication (== False) posDoSimplifierEvaluateBuiltins True
```
effectfully, line 179: My OCD would prefer `id` over `(== True)` and `not` over `(== False)`, but I guess it's currently more readable.

plutus-tx-plugin/src/PlutusTx/Options.hs
```diff
@@ -230,6 +232,11 @@ pluginOptions =
     , let k = "simplifier-beta"
           desc = "Run a simplification pass that performs beta transformations"
        in (k, PluginOption typeRep (setTrue k) posDoSimplifierBeta desc [])
+    , let k = "simplifier-evaluate-builtins"
+          desc =
+            "Run a simplification pass that evaluates fully saturated builtin applications. \
```
effectfully, line 237: Please add some comments! To the pass itself and somewhere here, so that the user knows they are enabling a semantics-changing pass. At least the link to the issue.

effectfully (commented):
Sorry for the late review.

PR #6453, opened 2024-09-09
Kwxm/conformance/extra bitwise tests

kwxm:
The conformance tests for the bitwise builtins were mostly operating on very small inputs.  This adds some tests with larger inputs and a number of tests for bit indices near the boundaries of `Int64`, where we previously had a problem due to things  wrapping round (that's fixed now, but let's guard against it returning in the future).  I also removed trailing whitespace in some of the inputs to the conformance tests.

There appears to be a  bug in `findFirstSetBit` : the `case-7` test for that is marked as being expected to fail in the two `Spec.hs` files (for Haskell and Agda) pending a fix.


bezirg:
> There appears to be a bug in findFirstSetBit : the case-7 test for that is marked as being expected to fail in the two Spec.hs files (for Haskell and Agda) pending a fix.

Do u know what the bug is?

bezirg:
I took a look at most tests, excluding and/or/xor because that would be hard without a calculator

kwxm:
> > There appears to be a bug in findFirstSetBit : the case-7 test for that is marked as being expected to fail in the two Spec.hs files (for Haskell and Agda) pending a fix.
> 
> Do u know what the bug is?

Failure to pass on an accumulator.  It's fixed in [this PR.](url)

PR #6454, opened 2024-09-10
Fix broken tests

kozross:
This fixes the issues with `findFirstSetBit` found in the original PR. Furthermore, we add some unit tests to ensure this regression does not re-occur.


kwxm:
Hmmm. I had absolutely no intention of closing this.  I think it must have happened because I merged #6453 and the branch was automatically deleted when I did that.  Apologies for that: please reopen the PR!

PR #6459, opened 2024-09-10
Fix for findFirstSet operation

kozross:
Retargeting of accidentally-closed #6454.


kwxm:
Susperseded by #6469.

PR #6461, opened 2024-09-10
Fix regression, ensure it doesn't come back with a property test

kozross:
Supercedes #6459, with a cleaner history. Additionally, instead of unit tests for the regression cases, there's now a property test instead.


kwxm (approved):
This all looks good. Thanks for the quick fix!

PR #6462, opened 2024-09-10
 [Builtins] [Test] Dump denotation signatures to golden files

effectfully:
This dumps the type signature of the denotation of each of the builtins to a golden file. All changes in type signatures must be reflected explicitly whenever they're captured by semantics variants (see the test code). This is a very important test given how easy it is to change the behavior of a builtin just by slightly tweaking the signature of its denotation.


plutus-core/plutus-core/test/TypeSynthesis/Spec.hs
```diff
@@ -152,10 +156,105 @@ test_typecheckDefaultFuns =
...
+
+-- | Dump the type signature of the denotation of each of the built-in functions to a golden file.
+-- If the signature of the denotation of a built-in function has ever changed and that is reflected
+-- in the semantics variants, then the signatures are dumped to @n + 1@ files where @n@ is the
+-- number of changes (e.g. if there was only one, there'll be two golden files: the one for the
+-- original signature and the other for the updated one). The number of semantics variants doesn't
+-- matter, only the number of type signature changes (controlled by the second argument).
+--
+-- This design ensures that all type signature changes of denotations are explicitly reflected and
+-- the additiona of another semantics variant won't mask an unexpected change in the signature of a
```
Unisay, line 200:
```suggestion
-- the addition of another semantics variant won't mask an unexpected change in the signature of a
```

Unisay (approved):
LGTM

PR #6463, opened 2024-09-11
Fix printing of type variables in print-builtin-signatures

kwxm:
`uplc print-builtin-signatures` prints out a table of all of the builtins and their signatures.  After some changes in the prettyprinter a while ago this was printing out things like

```
ifThenElse                         : [ forall a, bool, a-0, a-0 ] -> a-0
```

This PR jsut makes it use a vesion of the prettyprinter that gives

```
ifThenElse                         : [ forall a, bool, a, a ] -> a
```
instead.


kwxm:
This is a pretty trivial change so I'll just merge it automatically.

PR #6468, opened 2024-09-11
[Builtins] Add the 'dropList' builtin

effectfully:
This experiment adds the `dropList` builtin so that folks can play with it. Not intended for merging.


plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1557,6 +1558,18 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
             nullListDenotation
             (runCostingFunOneArgument . paramNullList)
 
+    toBuiltinMeaning _semvar DropList =
+        let dropListDenotation :: Int -> SomeConstant uni [a] -> BuiltinResult (Opaque val [a])
```
kwxm, line 1562: The time taken for this will probably be linear in the value (not the size!) of the first argument.  That'll mean susing some newtype wrapper, like [this](https://github.com/IntersectMBO/plutus/blob/983eba57bb6c2240780d078fa4618307d96f72a8/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs#L1998) .  We already have [IntegerCostedLiterally](https://github.com/IntersectMBO/plutus/blob/983eba57bb6c2240780d078fa4618307d96f72a8/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs#L192), but this has an `Int` so that may need another wrapper.

plutus-tx/src/PlutusTx/Builtins.hs
```diff
@@ -82,6 +82,7 @@ module PlutusTx.Builtins (
                          , headMaybe
                          , BI.head
                          , BI.tail
+                         , BI.drop
```
kwxm, line 85: Should this also go in [PlutusTx.Prelude](https://github.com/IntersectMBO/plutus/blob/master/plutus-tx/src/PlutusTx/Prelude.hs)? I seem to remember that the contents of that file are kind of random though.

plutus-tx/src/PlutusTx/Builtins/Internal.hs
```diff
@@ -408,6 +409,10 @@ chooseList :: BuiltinList a -> b -> b -> b
 chooseList (BuiltinList [])    b1 _ = b1
 chooseList (BuiltinList (_:_)) _ b2 = b2
 
+{-# OPAQUE drop #-}
+drop :: Integer -> BuiltinList a -> BuiltinList a
```
kwxm, line 413: Why not `dropList`?

kwxm (commented):
Just some comments in passing.

plutus-tx/src/PlutusTx/Builtins/Internal.hs
effectfully: For consistency, we also have `head` (and not `headList`) and `tail` (and not `tailList`) in this file.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
effectfully: > The time taken for this will probably be linear in the value (not the size!) of the first argument. 

No, I think it's gonna be a minimum of

- linear in the value of the first argument
- linear in the length of the second argument

since `drop maxBound []` equals `[]`.

effectfully: > Actually, why did I even make it `Int`, what's the point? Looks like a thinko.

It's proper `Integer` in there now.

kwxm: > No, I think it's gonna be a minimum of
>    linear in the value of the first argument
>    linear in the length of the second argument
> since drop maxBound [] equals [].

I deliberately ignored that case!  If we include the length of the list in the cost calculation then we'll have to traverse the etire list to measure the length when we're about to calculate the cost, and that'll slow things down.  If we just base it on the number of things we want to drop then we'll still get a conservative estimate of the cost, and it'll be faster.  Also, if you drop the whole of the list then you've probably done something wrong, and people probably won't submit scripts where that actually happens (or at least not often).

effectfully: Oh I see. OK, that's clever. Yeah, I think I agree, although I'd perhaps prefer to store lists together with their length, although that'd perhaps slow everything else down. Fair enough!

@ramsay-t ignore what I said and only consider the original comment in this thread.

kwxm: > I'd perhaps prefer to store lists together with their length

I've wondered about doing that in the past (and maybe something similar with Data for instance), but what happens if someone lies about the length of a list in their script?  Maybe that could be checked during deserialisation ...

kwxm:
> Comparing benchmark results of 'validation' on 'ac8c225aeb' (base) and 'd36d0f87f4' (PR)
> Results table
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 			
> 	[ac8c225](https://github.com/IntersectMBO/plutus/commit/ac8c225aebaa74e3db10731fe26d8ce5950d89ea) 	[d36d0f8](https://github.com/IntersectMBO/plutus/commit/d36d0f87f4443c5746f9b73d758dd435d7377523) 	Change
> TOTAL 	46.71 ms 	42.41 ms 	-9.2%

Oh, come on!

plutus-core/cost-model/data/models.R
```diff
@@ -823,6 +824,7 @@ modelFun <- function(path) {
         headListModel                        = headListModel,
         tailListModel                        = tailListModel,
         nullListModel                        = nullListModel,
+        dropListModel                      = dropListModel,
```
Unisay, line 827: Non-stylish Haskell :)

.github/workflows/manual-benchmark.yml
kwxm, line 1: I think that the changes in this file should be discarded: it's trying to pin the benchmarking process to a particular core, but that code's been removed in master.

scripts/ci-plutus-benchmark.sh
kwxm, line 1: I think these changes should also be discarded.

scripts/ci-plutus-benchmark.sh
```diff
@@ -2,23 +2,24 @@
 
 # This script runs the given benchmark and compares the results against origin/master.
 #
-# USAGE: 
+# USAGE:
```
zeme-wana, line 5: This looks like a bad merge. Please use the one from master

kwxm:
This had become very difficult to merge, so it's been closed in favour of #6817.

PR #6483, opened 2024-09-14
[Test] Remove uniques from 'Profiling' tests

effectfully:
This removes uniques from some golden test files, because those add a lot of noise in unrelated PRs.


effectfully:
This is a trivial change, so I'm gonna merge it right away without waiting for a review.

PR #6486, opened 2024-09-16
Remove "Exporting scripts, datums and redeemers" from user guide

zliu41:
This is redundant given we now have blueprints.

Also moved the blueprint page from "Using Plutus Tx" to "Working with scripts", since it is not a language feature, but something specific to scripts.


effectfully (approved):

Unisay:
IIUC we want to discourage exporting script without producing a blueprint?

zliu41:
> Did I understand correctly that we want to discourage exporting scripts without producing a blueprint?

No but for the purpose of avoiding accidental hash change, the best thing to do is to generate a blueprint.

One may still want tol export scripts to a `.uplc` file for purposes such as debugging. This can be mentioned in the doc on the `plutus` executable.

PR #6489, opened 2024-09-16
Update plutus-tx-template.yml

zliu41:
The job is broken by https://github.com/IntersectMBO/plutus-tx-template/pull/20.

### DIFF
--- a/.github/workflows/plutus-tx-template.yml
+++ b/.github/workflows/plutus-tx-template.yml
@@ -1,7 +1,7 @@
-# This workflows ensures that the plutus-tx-template repository stays working 
-# even if there are changes in plutus. It checks out both the current commit of 
-# plutus and the master branch of plutus-tx-template. Then, it creates a 
-# cabal.project.local for plutus-tx-template that adjusts the plutus version. 
+# This workflows ensures that the plutus-tx-template repository stays working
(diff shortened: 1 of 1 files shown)


.github/workflows/plutus-tx-template.yml
```diff
@@ -17,40 +17,40 @@ jobs:
...
       - name: Build Project With Docker
         run: |
-          cd plutus-tx-template 
+          cd plutus-tx-template
           docker run \
             -v ./.:/workspaces/plutus-tx-template \
             -w /workspaces/plutus-tx-template \
             -i ghcr.io/input-output-hk/devx-devcontainer:x86_64-linux.ghc96-iog \
-            bash -ic "cabal update && cabal run plutus-tx-template && test -e validator.uplc"
+            bash -ic "cabal update && cabal build all"
```
effectfully, line 56: We don't want to run any test?

zliu41: There's no test to run at the moment. The old command simply runs the executable, and checks that the output file exists, which isn't testing much, and now it no longer works.

effectfully: Thanks!

effectfully (approved):

PR #6498, opened 2024-09-18
Cost model overview document

kwxm:
This adds a document giving an overview of the Plutus Core cost model.  It's not supposed to be a full specfication but it should help people to understand how it works.  It's currently a PDF but later we might think about converting it (or some of it) into something more readable that can be added to the rest of the documentation.  Maybe we want to automatically publish the PDF as well, like we do with the specification.

The PDF's not in the repository: here's a copy.
[cost-model-overview.pdf](https://github.com/user-attachments/files/17044150/cost-model-overview.pdf)


doc/cost-model-overview/cost-model-overview.tex
```diff
@@ -0,0 +1,418 @@
...
+
+%% This is what necessitates the whole apparatus of costing functions: we can't
+%% instrument all of our builtins to calculate costs as they proceed, and we can't
+%% monitor execution dynamically: once we hand over control to a library function
+%% we just have to wait until it returns, so we have to be confident in advance
+%% that it is safe to call it.
+
+There are per-script limits for CPU and memory usage of scripts on the chain.
+These are the \texttt{steps} and \texttt{memory} components of the
+\texttt{maxTxExecution\-Units} protocol parameter and in Protocol Version 9 they
```
zliu41, line 184: Say something like "As of this writing they have values of ...", since their values may change without changing the PV.

zliu41 (approved):
Looks great. Can you include the pdf in the repo, so I can link to it from the user guide?

zliu41:
An additional suggestion: clarify that "Plutus Core" is short for "Untyped Plutus Core", and use acronym UPLC instead of PLC, since the former is more commonly used and recognized.

kwxm:
> Looks great. Can you include the pdf in the repo, so I can link to it from the user guide?

Yes, but I wondered if we should publish it automatically, like the specfication. Maybe it's not worth the trouble since this won't change too often.

doc/cost-model-overview/cost-model-overview.tex
kwxm: Wait, is that right?  I thought it needed a protocol update to change the protocol parameters.

kwxm:
The table includes the bitwise operations, which won't be available until Chang+1.  I suppose it won't do any harm to leave them in though.

PR #6501, opened 2024-09-18
Use NOINLINE on stringToBuiltinByteString and stringToBuiltinString

zliu41:
The plugin does not need their unfoldings, and cannot process their unfoldings: neither can the plugin handle `Text.pack`, nor does it have a handler for `obfuscatedId`.

Thus `obfuscatedId` makes no sense whatsoever. And there is no benefit to `Magic.noInline` compared to `NOINLINE`.


plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -49,35 +48,26 @@ of 'stringToBuiltinString' (since the newtype constructor vanishes), so we have
...
-{-# INLINABLE stringToBuiltinByteString #-}
+{-# NOINLINE stringToBuiltinByteString #-}
 
 stringToBuiltinString :: Haskell.String -> BuiltinString
--- To explain why the obfuscatedId is here
--- See Note [noinline hack]
-stringToBuiltinString str = obfuscatedId (BuiltinString $ Text.pack str)
-{-# INLINABLE stringToBuiltinString #-}
+stringToBuiltinString str = BuiltinString (Text.pack str)
+{-# NOINLINE stringToBuiltinString #-}
```
effectfully, line 57: We [should](https://github.com/IntersectMBO/plutus/issues/6269) use `OPAQUE`, not `NOINLINE`.

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -49,35 +48,26 @@ of 'stringToBuiltinString' (since the newtype constructor vanishes), so we have
 some obfuscation to the body to prevent it inlining.
 -}
```
effectfully, line 49: Update that Note as well?

Unisay (approved):

zliu41 (commented):
> I can't really sign off on these changes, because I don't really understand all the moving parts. But I don't mind this getting merged (once comments are addressed).

Ok, I'll merge it. The plugin 100% does not need and cannot compile the unfoldings of `stringToBuiltinByteString` and `stringToBuiltinString`, so there's no reason not to mark the OPAQUE.

PR #6502, opened 2024-09-19
[EXPERIMENT] [Builtins] 10x more builtins

effectfully:
This replicates the existing builtins so that there are 10x of them now. This is just to check whether it results in any evaluation slowdown.


effectfully:
That doesn't make any sense:

image

effectfully:
> That doesn't make any sense:

Yup, the benchmarking machine is cooked as it turns out (confirmed elsewhere). Well, gotta wait for it to recover.

effectfully:
Experiment over, this weirdness needs to be investigated, maybe we already pay some unnecessary costs. Made a subticket in #4306:

> Increasing the number of builtins by 10x (by replicating each builtin 10 times) slows the evaluator down by 15% with GHC Core not containing any extra duplication (#6502), just a large pattern match on `DefaultFun`. Why? Can a large pattern match be that slow? Or is something not cached properly? We should try using arrays for `DefaultFun` lookup again

effectfully:
The conflicts don't seem too bad, we should rerun the benchmarks once the issue of misleading benchmarking is fixed before this branch bitrots any further.

PR #6507, opened 2024-09-19
[Refactoring] Replace 'NOINLINE' with 'OPAQUE'

### LINKED ISSUE #6269: Replace `NOINLINE` with `OPAQUE`
We should replace all occurrences of `NOINLINE` with `OPAQUE`, because the former [doesn't](https://gitlab.haskell.org/ghc/ghc/-/issues/19553) actually prevent inlining. See also the original [proposal](https://github.com/christiaanb/ghc-proposals/blob/master/proposals/0000-opaque-pragma.md) to introduce `OPAQUE`.

Once this is done we should check if that's enough to consider #4193 done.

effectfully:
This replaces all occurrences of `NOINLINE` with the more reliable `OPAQUE`. However unlike `NOINLINE` `OPAQUE` prevents specialization, so let's benchmark this to make sure there's no unexpected slowdown. 

Resolves #6269.

### DIFF
--- a/doc/adr/adr3.md
+++ b/doc/adr/adr3.md
@@ -73,7 +73,7 @@ enterComputeCek debug ctx (Closure term env) =
         -> Closure uni fun
         -> CekM uni fun s (CekState uni fun)
     computeCek = if debug then computeCekDebug else computeCekStep
-    {-# NOINLINE computeCek #-}  -- Making sure the `if` is only evaluated once.
+    {-# OPAQUE computeCek #-}  -- Making sure the `if` is only evaluated once.
 
     -- in debugging mode, immediately returns the current `CekState` and halts execution. Debugging mode details to be worked out.
     computeCekDebug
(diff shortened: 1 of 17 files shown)


zliu41:
OPAQUE is strictly better than NOINLINE for the plugin, but I don't think that is the case for non-Plinth code.

effectfully:
> OPAQUE is strictly better than NOINLINE for the plugin, but I don't think that is the case for non-Plinth code.

Sure, but usually when I write `NOINLINE` I do really mean "don't inline that thing". So I think `OPAQUE` is a more sensible default and we can replace it with `NOINLINE` in special cases.

effectfully:
Interesting!

effectfully:
It honestly looks like the benchmarking machine is broken. I'll test that separately.

effectfully:
The benchmarking machine is broken at the moment, so we can't get accurate benchmarking results. I think it's unlikely that we need `NOINLINE` somewhere, so let's just merge this and be done with it.

zliu41 (approved):

PR #6510, opened 2024-09-23
Mark && and || OPAQUE

zliu41:
I was pretty sure we did this before - perhaps I'm having a Mandela effect moment.

Since `&&` and `||` are specially handled in the plugin, there's no need to rely on GHC to inline them, which is a much less reliable way to make the short-circuit.


effectfully (commented):
I don't understand why we should do this. In this case both our compiler and GHC do the same thing, it's just that the heuristics are different, but the transformation is correct by our standards regardless. So why give up on GHC optimizing the expression if it does help a tiny bit as is visible from the golden files?

zliu41:
> if it does help a tiny bit as is visible from the golden files?

That's the problem - GHC must be doing something else, in addition to what we do, hence the cost difference. In general we shouldn't trust any GHC transformation to be semantic preserving, at least not until we understand what exactly it does.

Unisay (approved):
Agree with the idea because^

1. The principle "We shouldn't trust GHC transformations in general" makes sense to me.
2. There are less ways of doing the same thing (short-circuiting) now,  this change reduces complexity.

effectfully:
> That's the problem - GHC must be doing something else, in addition to what we do, hence the cost difference. In general we shouldn't trust any GHC transformation to be semantic preserving, at least not until we understand what exactly it does.

Well, it would be great to figure out what it does then. If this is affected by the `INLINE` pragma, then maybe we should do it in the inliner as well?

Just to clarify, I'm not opposed to merging this PR. I'm just not satisfied with the reasoning.

zliu41:
> Well, it would be great to figure out what it does then.

Yes definitely. It's one of the new tickets I'm about to create.

zliu41:
https://github.com/IntersectMBO/plutus-private/issues/1143 and https://github.com/IntersectMBO/plutus-private/issues/1144 for further investigation

PR #6513, opened 2024-09-24
This seems 'too easy' but to certify things I don't think it needs to be more complex?

### LINKED ISSUE #6370: Translation relation and decision procedure for UPLC CSE

ramsay-t:
A straightforward Translation Relation for CSE certification. Perhaps there are some more subtle things we want to check?

Fixes #6370


plutus-metatheory/src/VerifiedCompilation/UCSE.lagda.md
```diff
@@ -0,0 +1,63 @@
...
+
+```
+
+isUntypedCSE? : {X : Set} {{_ : DecEq X}} → Binary.Decidable (Translation UCSE {X})
+
+{-# TERMINATING #-}
+isUCSE? : {X : Set} {{_ : DecEq X}} → Binary.Decidable (UCSE {X})
+isUCSE? ast ast' with (isApp? (isLambda? isTerm?) isTerm?) ast
+... | no ¬match = no λ { (cse x) → ¬match (isapp (islambda (isterm _)) (isterm _)) }
+... | yes (isapp (islambda (isterm x)) (isterm e)) with isUntypedCSE? (x [ e ]) ast'
```
effectfully: I think the whole approach to certification is very suboptimal. I.e. this pass and the others (not sure if all of them). So what you do here is you check that the LHS has the right shape and then you proceed with substitution and recurse. Potentially just to figure out that you didn't need to do all this work, because the terms match exactly and there was no point in all these substitutions and non-deterministic digressing into CSE, float-delay and everything else at every lambda application node.

I guess this is a fine start, but I wouldn't be surprised if it eventually turns out that all the passes need to be rewritten, because they can't handle real-world terms fast enough.

plutus-metatheory/src/VerifiedCompilation/UCSE.lagda.md
```diff
@@ -0,0 +1,63 @@
...
+semantics; we are explicitly not evaluating whether the particular choice of
+sub-expression was a "good" choice.
+
+As such, this Translation Relation primarily checks that substituting the expression
+back in would yield the original expression.
+
+```
+data UCSE : Relation where
+  cse : {X : Set} {x : Maybe X ⊢} {x' e : X ⊢}
+    → Translation UCSE (x [ e ]) x'
```
effectfully: I'm perhaps nitpicking, but this isn't sufficient in the general case. CSE promises to pull some common subexpressions out, but it doesn't promise to pull all of them out. I.e. `x'` may still start with the same lambda and `e` may be substituted partially in `x`.

effectfully (approved):
I'm OK with merging it, but chances are, it'll need to be rewritten in future.

plutus-metatheory/src/VerifiedCompilation/UCSE.lagda.md
ramsay-t: If they are identical then that should be handled by `translation?` before it gets into this decision procedure. 

It isn't because this one runs first in `translation?` because if it didn't then we might get half way down the AST before realising we need to look at some of the higher structure for the specific relation, and I'm not sure how neat the backtracking is in `translation?`. That is what I'd look at first...

effectfully: > If they are identical then that should be handled by `translation?` before it gets into this decision procedure.

Nope, `translation?` invokes `isR?` before doing anything else.

ramsay-t: The word "should" was important in that sentence :)

PR #6514, opened 2024-09-24
Remove plutus-ghc-stub

zliu41:
It has been non-functional for a long time (it is based on a GHC version before 9), and we no longer cross compile the plugin to Windows.


effectfully (approved):
:+1:

bezirg (approved):

PR #6515, opened 2024-09-24
Fix some issues and inconsistencies related to pretty-printing

zliu41:
- Renamed `prettyPirReadable` to `prettyReadable`, since it is not specific to PIR.
- Renamed `prettyPirReadableSimple` to `prettyReadableSimple`, since it is not specific to PIR.
- Removed `PrettyPir`, since it is the same as `PrettyReadable`, and not specific to PIR.
- Fixed `PlutusTx.Test.prettyPir`: it is supposed to use classic syntax, not readable syntax - we have `prettyPirReadable` for the latter.
- `prettyPir` prints `program 1.1.0` while `prettyPirReadable` doesn't. Now neither of them does.


Unisay (approved):
LGTM

plutus-core/plutus-core/src/PlutusCore/Pretty/Readable.hs
```diff
@@ -272,3 +272,11 @@ iterInterAppPrettyM fun args =
...
         ppArg (Right term) = prettyArg term
     in prettyFun fun :| map ppArg args
+
+-- | Pretty-print something with the @PrettyConfigReadable@ config.
+prettyReadable :: (PrettyReadable a) => a -> Doc ann
+prettyReadable = prettyBy (botPrettyConfigReadable prettyConfigName def)
+
+-- | Pretty-print something with the @PrettyConfigReadableSimple@ config.
+prettyReadableSimple :: (PrettyReadable a) => a -> Doc ann
+prettyReadableSimple = prettyBy (botPrettyConfigReadable prettyConfigNameSimple def)
```
effectfully, line 282: I wonder how this wasn't done ages ago... I suppose because we used `prettyPlcReadable` instead, but in any case it's very sensible to have these functions here. Thanks!

effectfully (approved):

plutus-core/plutus-core/src/PlutusCore/Pretty/Readable.hs
zliu41: Right, we still have `prettyPlcReadable` - it should be no longer needed

Issue #5967, opened 2024-05-09
[Epic] AST optimizations


effectfully:
This issue is for dumping all the plans regarding AST optimizations (such as case-of-case, case-of-known-constructors etc) in a largely unstructured way. We'll have separate messages for Plutus Core and Plutus IR, please just edit those instead of creating more messages.

effectfully:
## Untyped Plutus Core

- [ ] regarding order of effects, we should figure out a strategy for when it's fine to change it and when it's not. It's worth handling logging and bottom separately, since messing with the former is OK (given it's locked behind a flag not enabled by default) and messing with the latter is a no-no
- [ ] I saw `((\cse -> case cse [cse]) error)`, we should rewrite this to just `error`
- [ ] we need a test that the inliner inline `blah` in `(\b -> force ifThenElse b branch1 branch2) blah`. If it doesn't, we should implement that functionality
- [ ] handle `chooseList` in case-of-case similarly to `ifThenElse`?
- [ ] Aiken folks gave us an idea: try replacing Scott-encoded tuples with SOP-encoded tuples in the recursive functions combinators, SOP-encoded tuples should be faster thus making mutual recursion faster. If there’s anything else Scott-encoded there, we should change it too. Perhaps we could update the entire `PlutusCore.StdLib.Meta.Data.Tuple`. We have to keep the old code too for older versions of Plutus
- [ ] both UPLC and PIR have `termEvaluationOrder` defined for them which goes into "known function body" in case of `Apply` (see the logic there) and goes into "known delayed term" for `Force`, but does not go into "known branch" for `Constr`
- [ ] both UPLC and PIR have `termEvaluationOrder` defined for them which does into "known delayed term" for `Force` and considers that to be `MaybeWork`. Is it work though? It literally has to be optimized away, do we want to prevent some optimization from firing just because there's an immediately `Force`d `Delay`? This should at least be discussed in a Note
- [ ] we sometimes generate things like `force (force (force ifThenElse b (delay (delay x)) (delay (delay y))))`. Surely we can turn double-force-delays into single force-delays?
[...]

PR #6517, opened 2024-09-25
[Test] [Bug] Trigger the overpruned dependency bug

effectfully:
The added test fails with

```
test
  Plugin
    Data
      9.6
        families
          stakingCredential: FAIL
            Exception: Error: Error from the PIR compiler:
                   Error from the PLC compiler:
                   Free type variable at  Ann {annInline = MayInline, annSrcSpans = { no-src-span }} :  Credential
```

The bug is in the dead code eliminator. To be precise, it's this part that it doesn't handle correctly despite promising to in `PlutusIR.Analysis.Dependencies`:

```
{- Note [Dependencies for datatype bindings, and pruning them]
At face value, all the names introduced by datatype bindings should depend on each other.
Given our meaning of "A depends on B", since we cannot remove any part of the datatype
binding without removing the whole thing, they all depend on each other

However, there are some circumstances in which we *can* prune datatype bindings.

In particular, if the datatype is only used at the type-level (i.e. all the term-level parts
(constructors and destructor) are dead), then we are free to completely replace the binding
with one for a trivial type with the same kind.

This is because there are *no* term-level effects, and types are erased in the end, so
in this case rest of the datatype binding really is superfluous.

But how do we represent this in the dependency graph? We still need to have proper dependencies
so that we don't make the wrong decisions wrt transitively used values, e.g.

let U :: * = ...
let datatype T = T1 | T2 U
in T1

Here we need to not delete U, even though T2 is "dead"!

The solution is to focus on the meaning of "dependency": with the pruning that we can do, we *can*
remove all the term level bits en masse, but only en-mass. So we need to make *them* into a clique,
so that this is visible to the dependency analysis.
-}
```

I don't know what went wrong there.


effectfully:
@michaelpj any idea?

michaelpj:
What happens if you get rid of the bit where it tries to prune dead constructors?

effectfully:
> What happens if you get rid of the bit where it tries to prune dead constructors?

The test becomes green, see 229b6903412086ed05b468a65e751fb59dd2fb08.

michaelpj:
Since you have a small test case I would dump the output from just after the bad dead code pass and see what's wrong. It *looks* like a simple case we should handle properly...

michaelpj:
From the error it seems like we get rid of `Credential` entirely, which seems odd.

effectfully:
> Since you have a small test case I would dump the output from just after the bad dead code pass and see what's wrong. It looks like a simple case we should handle properly...

Yes, I've done exactly that when investigating and you're right that this is what's happening:

> From the error it seems like we get rid of `Credential` entirely, which seems odd.

michaelpj:
I don't see why from just looking at it, some debugging is going to be required I think!

effectfully:
> I don't see why from just looking at it, some debugging is going to be required I think!

Wanna do some debugging on a weekend? :slightly_smiling_face:

plutus-benchmark/script-contexts/test/9.6/checkScriptContextEqualityOverhead-20.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 34304100
-| mem: 214500})
\ No newline at end of file
+({cpu: 37152100
+| mem: 232300})
```
effectfully, line 2: @zliu41 is this large enough difference to keep the bug in the dead code eliminator that causes necessary constructors to get eliminated?

plutus-core/plutus-ir/src/PlutusIR/Analysis/Dependencies.hs
```diff
@@ -172,23 +172,10 @@ bindingDeps b = case b of
...
-    -- term-level parts depend on each other later, but it's good practice and will be
-    -- useful if we ever stop doing that.
-    destrDeps <-
-      G.overlays
-        <$> (withCurrent destr $ traverse (typeDeps . _varDeclType) constrs)
-    let tus = fmap (view PLC.theUnique) (destr : fmap _varDeclName constrs)
-    -- See Note [Dependencies for datatype bindings, and pruning them]
-    let nonDatatypeClique = G.clique (fmap Variable tus)
-    pure $ G.overlays $ [vDeps] ++ tvDeps ++ cstrDeps ++ [destrDeps] ++ [nonDatatypeClique]
+    let tyus = fmap (view PLC.theUnique) $ _tyVarDeclName d : fmap _tyVarDeclName tvs
```
zliu41, line 175: I'm not quite sure what this does; worth some comment. There should also be some explanation or reference to the bug here.

plutus-benchmark/script-contexts/test/9.6/checkScriptContextEqualityOverhead-20.budget.golden
zliu41: 10% is bad but only a small number of golden files are affected - I'm fine either way. What's your preference?

plutus-core/plutus-ir/src/PlutusIR/Analysis/Dependencies.hs
effectfully: I have no idea either, that's Michael code and as you can see above Michael isn't particularly keen on dealing with it :)

plutus-benchmark/script-contexts/test/9.6/checkScriptContextEqualityOverhead-20.budget.golden
effectfully: I guess let's keep the bug in then, I think it's quite low priority compared to other things and I don't feel like studying the DCE algo at the moment (or ever haha). I'll reflect this problem in an issue.

zliu41: Sounds good.

effectfully:
I've added an entry to #5967, so that we don't forget about this bug, and let's close this PR for now.

zliu41:
Fixed in https://github.com/IntersectMBO/plutus/pull/6806

PR #6540, opened 2024-10-04
Add simplifier tags in UPLC simplifier

ana-pantilie:
This cleans up the UPLC tracing infrastructure as requested by @effectfully and also provides some "compilations hints" to the certifier in the form of simplification stage tags.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Simplifier.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Simplifier.hs
@@ -1,19 +1,59 @@
 {-# LANGUAGE NamedFieldPuns #-}
 
 module UntypedPlutusCore.Transform.Simplifier (
-    Simplifier,
+    SimplifierT (..),
     SimplifierTrace (..),
     SimplifierStage (..),
     Simplification (..),
+    runSimplifierT,
+    evalSimplifierT,
+    execSimplifierT,
+    Simplifier,
+    runSimplifier,
+    evalSimplifier,
+    execSimplifier,
     initSimplifierTrace,
     recordSimplification,
 ) where
 
-import Control.Monad.State (StateT)
+import Control.Monad.State (MonadTrans, StateT)
 import Control.Monad.State qualified as State
+
+import Control.Monad.Identity (Identity, runIdentity)
+import PlutusCore.Quote (MonadQuote)
 import UntypedPlutusCore.Core.Type (Term)
 
-type Simplifier name uni fun a m = StateT (SimplifierTrace name uni fun a) m
+newtype SimplifierT name uni fun ann m a =
+  SimplifierT
+    { getSimplifierT :: StateT (SimplifierTrace name uni fun ann) m a
+    }
+  deriving newtype (Functor, Applicative, Monad, MonadTrans)
+
+instance MonadQuote m => MonadQuote (SimplifierT name uni fun ann m)
+
+runSimplifierT
+  :: SimplifierT name uni fun ann m a
+  -> m (a, SimplifierTrace name uni fun ann)
+runSimplifierT = flip State.runStateT initSimplifierTrace . getSimplifierT
+
+evalSimplifierT
+  :: Monad m => SimplifierT name uni fun ann m a -> m a
+evalSimplifierT = flip State.evalStateT initSimplifierTrace . getSimplifierT
+
+execSimplifierT
+  :: Monad m => SimplifierT name uni fun ann m a -> m (SimplifierTrace name uni fun ann)
+execSimplifierT = flip State.execStateT initSimplifierTrace . getSimplifierT
+
+type Simplifier name uni fun ann = SimplifierT name uni fun ann Identity
+
+runSimplifier :: Simplifier name uni fun ann a -> (a, SimplifierTrace name uni fun ann)
+runSimplifier = runIdentity . runSimplifierT
+
+evalSimplifier :: Simplifier name uni fun ann a -> a
+evalSimplifier = runIdentity . evalSimplifierT
+
+execSimplifier :: Simplifier name uni fun ann a -> SimplifierTrace name uni fun ann
+execSimplifier = runIdentity . execSimplifierT
 
 data SimplifierStage
   = FloatDelay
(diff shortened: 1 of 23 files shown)


ramsay-t (approved):
LGTM - I drove myself mad trying to understand the serialisation, but it turned out I still had one of my hacky files floating around :)

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Simplifier.hs
```diff
@@ -0,0 +1,96 @@
...
+    Simplifier,
+    runSimplifier,
+    evalSimplifier,
+    execSimplifier,
+    initSimplifierTrace,
+    recordSimplification,
+) where
+
+import Control.Monad.State (MonadTrans, StateT)
+import Control.Monad.State qualified as State
```
effectfully, line 20: Please always either use `Control.Monad.State.Strict` or explain in a comment why the lazy version is appropriate.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Simplifier.hs
```diff
@@ -0,0 +1,96 @@
...
+  | CaseOfCase
+  | CaseReduce
+  | Inline
+  | CSE
+
+data Simplification name uni fun a =
+  Simplification
+    { beforeAST :: Term name uni fun a
+    , stage     :: SimplifierStage
+    , afterAST  :: Term name uni fun a
```
effectfully, line 70: This ordering makes my OCD happy.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Simplifier.hs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+-- TODO2: we probably don't want this in memory so after MVP
+-- we should consider serializing this to disk
```
effectfully, line 74: It looks to me like we should use a streaming lib instead. I like `streaming`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Simplifier.hs
```diff
@@ -0,0 +1,96 @@
...
+  :: Monad m
+  => Term name uni fun a
+  -> SimplifierStage
+  -> Term name uni fun a
+  -> SimplifierT name uni fun a m ()
+recordSimplification beforeAST stage afterAST =
+  let simplification = Simplification { beforeAST, stage, afterAST }
+    in
+      modify $ \st ->
+        st { simplifierTrace = simplifierTrace st ++ [simplification] }
```
effectfully, line 94: Why not store the list in reversed form? I'm not a fan of easily avoidable quadratic complexity, although maybe it's fine in this case.

effectfully (approved):
Haskell code looks good to me, I didn't bother reviewing Agda code.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Simplifier.hs
ramsay-t: This is one of those things that is easy to see in a review, but not while writing it :) No one's sanity will cope with this list being reversed in the Agda, and that traverses it "forwards" but we could build it in reverse and reverse it at the last minute before it is passed to Agda?

effectfully: Yes.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Simplifier.hs
ana-pantilie: @effectfully Good point, I forgot to discuss this.

The reason I use the "lazy" version here (see below why "lazy" is in quotes) is that https://github.com/IntersectMBO/plutus/blob/bb33f082d26f8b6576d3f0d423be53eddfb6abd8/plutus-core/plutus-core/src/PlutusCore/Quote.hs#L89 is the one from the lazy version, and I didn't want to mess with it. However, notice that when updating the state I'm using the strict `modify'`.

The strange thing is that the [lazy StateT](https://hackage.haskell.org/package/transformers-0.6.1.2/docs/src/Control.Monad.Trans.State.Lazy.html#StateT) and the [strict StateT](https://hackage.haskell.org/package/transformers-0.6.1.2/docs/src/Control.Monad.Trans.State.Strict.html#StateT) have the exact same definition! There's no language strict pragma at the top, `(,)` is the regular (lazy) pair type, so unless there's something else I'm not seeing it seems like the types themselves are the same, and the only strictness/laziness changes are in the functions. So the `blabla.Strict` type is actually lazy! 😄 

Do you know if there was any intention to use the "lazy" version of `StateT` for the `MonadQuote` instance? And second of all, should we do anything about this given the above?

effectfully: > The strange thing is that the [lazy StateT](https://hackage.haskell.org/package/transformers-0.6.1.2/docs/src/Control.Monad.Trans.State.Lazy.html#StateT) and the [strict StateT](https://hackage.haskell.org/package/transformers-0.6.1.2/docs/src/Control.Monad.Trans.State.Strict.html#StateT) have the exact same definition!

That's a common thing, it's the same with `transformers` for example.

> and the only strictness/laziness changes are in the functions

Yes, but one of those functions is `>>=` and the lazy state leaks space horribly unlike the strict version.

> Do you know if there was any intention to use the "lazy" version of `StateT` for the `MonadQuote` instance?

https://getyarn.io/yarn-clip/f237528f-efda-453a-a007-8f8a7e75746d

> And second of all, should we do anything about this given the above?

Yes, let's merge your PR as it is and then try to replace all lazy state with strict state at ones. There's a chance that'll break something, but that's somewhat unlikely.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Simplifier.hs
ana-pantilie: Oops, thanks for noticing. Yeah, this is a remnant of the "let's get something working quickly" implementation. I'll fix it asap.

Issue #6528, opened 2024-10-01
writeBits design is extremely inefficient for practical use


colll78:
### Describe the feature you'd like

Currently the signature of `writeBits` is:
```haskell
-- | Given a 'BuiltinByteString', a list of indexes to change, and a list of values to change those
-- indexes to, set the /bit/ at each of the specified index as follows:
--
-- * If the corresponding entry in the list of values is 'True', set that bit;
-- * Otherwise, clear that bit.
writeBits ::
  BuiltinByteString ->
  BuiltinList BuiltinInteger ->
  BuiltinList BuiltinBool ->
  BuiltinByteString
```

Instead it should be:
```haskell
-- | Given a 'BuiltinByteString', a list of indexes to change, and a boolean set/clear flag, set the /bit/ at each of the specified indices
-- if the boolean is true or clear the /bit/ at each of the specified indices if the boolean is false.
writeBits ::
  BuiltinByteString ->
  BuiltinList BuiltinInteger ->
  BuiltinBool ->
  BuiltinByteString
```

The issue with the current implementation is that any non-trivial use requires a huge amount of wasted work in constructing the `BuiltinList BuiltinBool` argument. If the caller wants to set and unset bits, they can simply call `writeBits` twice once with the indices they want to set and the flag set to true, and the second time with the indices they want to clear and the flag set to false. 

Consider the current practical use-case:

```haskell
-- | Return true iff the first 256 integers in the list are all unique.
isUniqueSet :: BuiltinList BuiltinInteger -> Bool
isUniqueSet xs = 
  let byteArray = replicateByte 32 0
       numEles = builtinListLength xs
       flagUniqueBytes = writeBits byteArray xs (replicateBuiltinTrue numEles)
   in (countSetBits flagUniqueBytes) == numEles
 
 replicateBuiltinTrue :: Integer -> BuiltinList BuiltinBool
 replicateBuiltinTrue n = go
   where 
     go n = 
       if (n == 0) 
       then BI.mkNilData BI.unitval
       else BI.mkCons (toOpaque True) $ go (n - 1)
 
 builtinListLength :: BuiltinList a -> Integer
 builtinListLength xs = go 0 xs 
   where 
     go acc l = matchList l (\_ -> acc) (\_ ys -> go (acc + 1) ys) 
```
[...]

PR #6547, opened 2024-10-08
Kwxm/write bits/new semantics

kwxm:
This implements the suggestion in #6528 , changing the type of the `writeBits` PLC builtin from 
```
[bytestring, list integer, list bool] -> bytestring
```
to
```
[bytestring, list integer, bool] -> bytestring
```

It's currently awaiting the results of some costing benchmarks to update the cost model, but I don't anticipate a massive change.


cardano-constitution/test/Cardano/Constitution/Validator/GoldenTests/sorted.uplc.golden
kwxm, line 1: This is presumably a random CSE-related rearrangement of the code.

plutus-core/cost-model/data/models.R
```diff
@@ -773,10 +773,7 @@ modelFun <- function(path) {
     readBitModel              <- constantModel ("ReadBit")
     writeBitsModel            <- linearInY ("WriteBits")
     ## ^ The Y value here is the length of the list of positions because we use ListCostedByLength
-    ## in the relevant costing benchmark.  The time actually depends on the minimum of the lengths
-    ## of the second and third arguments of `writeBits`, but that will be at most Y, so using
-    ## linearInY is conservatively safe.  If `writeBits` is used correctly then the lengths of the
-    ## second and third arguments will always be the same anyway.
+    ## in the relevant costing benchmark.
```
kwxm, line 776: ^ Conveniently, the thing mentioned in the deleted comment allows us to change to the new implementation without any signifcant changes to the costing code. The cost is determined by the length of the second argument: the third one (which is what we've changed) wasn't involved in the cost anyway.

plutus-benchmark/bitwise/test/9.6/8 queens.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 448084881560
-| mem: 1786827931})
\ No newline at end of file
+({cpu: 409142294016
+| mem: 1575213647})
```
effectfully, line 2: Not bad at all.

plutus-benchmark/bitwise/test/9.6/8 queens.pir.golden
```diff
@@ -66,9 +47,17 @@
...
+      = \(bs : bytestring) (ixes : List integer) (bit : Bool) ->
+          writeBits
+            bs
+            (goList ixes)
+            (Bool_match
+               bit
+               {all dead. bool}
+               (/\dead -> True)
+               (/\dead -> False)
+               {all dead. dead})
```
effectfully, line 60: It looks like maybe we should have `writeBitsTrue` and `writeBitsFalse` provided in the library, so that the `Bool_match` logic is reduced away statically. Maybe it doesn't matter much though. 

What sucks here is that we need to convert SOP lists to built-in lists, Aiken folks won't need that and so we're at a disadvantage here.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/writeBits/case-31/case-31.uplc
```diff
@@ -1,13 +1,20 @@
...
+     ]
+     (con (list integer) [340, 342, 343, 340, 342, 343])
+    ]
+    (con bool True)
    ]
-   (con (list integer) [340, 342, 343, 340, 342, 343])
+   (con (list integer) [340, 342, 343])
   ]
-  (con (list bool) [True, True, True, False, False, False])
+  (con bool False)
```
effectfully, line 18: Looking at this test I feel like maybe it's not the worst idea to have both versions of `writeBits`. Splitting the bits manually is error-prone and kinda annoying, plus it hurts performance when the list is known statically. We can of course express the old `writeBits` in terms of the new one, but this is going to be less efficient than just having the old one (in addition to the new one) and efficiency is what we care about here.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/writeBits/case-32/case-32.uplc
```diff
@@ -7,6 +7,6 @@
    ]
    (con (list integer) [340, 342, 343])
   ]
-  (con (list bool) [True, True, True, False, False, False]) -- Only the leftmost values are used if the lists are of different lengths.
+  (con bool False)
```
effectfully, line 10: This should `con bool True`, not `con bool False`, if we want to preserve the old behavior.

plutus-core/plutus-core/src/PlutusCore/Bitwise.hs
```diff
@@ -567,28 +567,28 @@ writeBits bs ixs bits = case unsafeDupablePerformIO . try $ go of
...
     len :: Int
     len = BS.length bs
     bitLen :: Integer
     bitLen = fromIntegral len * 8
-    setAtIx :: Ptr Word8 -> Integer -> Bool -> IO ()
-    setAtIx ptr i b
+    setOrClearAtIx :: Ptr Word8 -> Integer -> IO ()
+    setOrClearAtIx ptr i
       | i < 0 = throw $ WriteBitsException i
       | i >= bitLen = throw $ WriteBitsException i
```
effectfully, line 582: While you're here, please replace these two `throw` calls with `throwIO` ones. The Haddock of the latter:

```
-- | A variant of 'throw' that can only be used within the 'IO' monad.
--
-- Although 'throwIO' has a type that is an instance of the type of 'throw', the
-- two functions are subtly different:
--
-- > throw e   `seq` ()  ===> throw e
-- > throwIO e `seq` ()  ===> ()
--
-- The first example will cause the exception @e@ to be raised,
-- whereas the second one won\'t.  In fact, 'throwIO' will only cause
-- an exception to be raised when it is used within the 'IO' monad.
--
-- The 'throwIO' variant should be used in preference to 'throw' to
-- raise an exception within the 'IO' monad because it guarantees
-- ordering with respect to other operations, whereas 'throw'
-- does not. We say that 'throwIO' throws *precise* exceptions and
-- 'throw', 'error', etc. all throw *imprecise* exceptions.
-- For example
--
-- > throw e + error "boom" ===> error "boom"
-- > throw e + error "boom" ===> throw e
--
-- are both valid reductions and the compiler may pick any (loop, even), whereas
--
-- > throwIO e >> error "boom" ===> throwIO e
--
-- will always throw @e@ when executed.
--
-- See also the
-- [GHC wiki page on precise exceptions](https://gitlab.haskell.org/ghc/ghc/-/wikis/exceptions/precise-exceptions)
-- for a more technical introduction to how GHC optimises around precise vs.
-- imprecise exceptions.
```

Using `throw` here is asking for trouble.

effectfully (approved):
Approving, but see the comment about `throw` -> `throwIO` (not related to this PR).

Also, I'm close to convincing myself that we should have both versions of the builtin. But if we're only going to have one, then the one in this PR is a clear winner.

plutus-benchmark/bitwise/test/9.6/8 queens.budget.golden
effectfully, line 2: Not bad at all.

plutus-core/plutus-core/src/PlutusCore/Bitwise.hs
kwxm: > While you're here, please replace these two throw calls with throwIO ones.

OK, I done that.  Thanks for pointing that out.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/writeBits/case-32/case-32.uplc
kwxm: That would make it a duplicate of [case-30.uplc](https://github.com/IntersectMBO/plutus/pull/6547/files#diff-7a8d00cb3164bd4ea936f09ca30b732480c73e65ac09aef539878e5522a4b1cb).  This test was testing something that doesn't apply any more and instead of deleting it and renaming the others I was lazy and changed it to test something else that I don't think we were testing elsewhere (or at least not exactly).

effectfully: Fair enough.

kwxm:
The costing benchmark results for this version of `writeBits` aren't substantially different from the old version so I'm just going to keep the existing cost model parameters.

kwxm:
This fixes #6528.

kwxm:
The Darwin thing's taking forever, so I'm going to bypass it.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/writeBits/case-31/case-31.uplc
colll78: In any scenario where the list is known statically (ie are constants) you should be able to automatically split the bits (precompilation) at zero onchain cost.

PR #6549, opened 2024-10-08
Remove the common weaknesses page

zliu41:
The Plutus user guide does not serve as a resource for smart contract security, and the Plutus team members are not experts in this area. Also, this page is far from comprehensive - it would be better to link to resources like [Cardano Plutus Script Vulnerability Guide](https://library.mlabs.city/common-plutus-security-vulnerabilities), which we already do.


effectfully (approved):
Agreed.

PR #6550, opened 2024-10-09
Redraw contract lifecycle images, remove the rest of the page

zliu41:
Redrew the images to be more accurate and consistent with the auction example.

The content of the contract lifecycle page was removed, as it was mostly redundant given the end-to-end example. The new images are inserted into the example.


Unisay (approved):
Cool images!

PR #6558, opened 2024-10-10
Expand "Overview of Languages Compiling to UPLC"

zliu41:
Adding some more content and arguments.

### DIFF
--- a/doc/docusaurus/docs/delve-deeper/languages.md
+++ b/doc/docusaurus/docs/delve-deeper/languages.md
@@ -24,12 +24,21 @@ A standalone DSL is a new language with its own syntax and semantics.
 By crafting a new language from scratch, you avoid inheriting the limitations and complexities of existing languages, allowing you to tailor-make it to be as simple, intuitive and elegant as possible to program for the specific domain it targets.
 
 On the other hand, standalone DSLs have some disadvantages.
-First, designing and implementing them can be challenging.
-Not only must the syntax and semantics be created from scratch, but you also need to build all necessary compiler components, tooling, and a library ecosystem from the ground up.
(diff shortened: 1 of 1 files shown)


doc/docusaurus/docs/delve-deeper/languages.md
```diff
@@ -50,6 +59,9 @@ Another disadvantage of eDSLs is that it is harder, compared to the other two ap
 This stems from the nature of eDSLs, which are libraries that construct and manipulate ASTs.
 Since they do not have direct access to the host language's ASTs, it can be challenging to retrieve information related to the source code, such as variable names, module names and code locations.
 
+Lastly, eDSLs often cannot leverage the host language's existing library ecosystem.
+Again, using an eDSL involves constructing and manipulating ASTs rather than writing regular programs, which means many existing libraries would be inapplicable.
```
effectfully, line 63: But it's very similar to what we have in Plinth? We have to reimplement everything to make sure it's Plinth-compatible and not just Haskell.

effectfully (approved):

doc/docusaurus/docs/delve-deeper/languages.md
zliu41: Yes but this is more particular to Plinth/Haskell than a general limitation of that type of language. As I said in the footnote, "This statement is not entirely true for Plutus Tx, a subset of Haskell. Due to certain GHC-specific technical limitations, it can't easily reuse many functions and operations from the base library, so it ships with its own standard library instead."

effectfully: Yeah, fair enough, I just do what I do best: unproductive nitpicking. Sorry!

PR #6560, opened 2024-10-10
Explain Scott/Church/SOP/Data encoding in the user guide

zliu41:


doc/docusaurus/docs/delve-deeper/encoding.md
```diff
@@ -0,0 +1,162 @@
...
+`Data` makes it simpler to create datums and redeemers using various languages and tools, or even manually, which is much easier than constructing UPLC terms.
+
+`Data` can be used to encode data types (though not as expressive as Scott encoding or SOP, since `Data` cannot contain functions).
+However, as with other data serialization/interchange formats (e.g., JSON or Protobuf-generated types), writing business logic directly using `Data` is inefficient and cumbersome.
+Instead, the standard practice is to validate the incoming data, turn it into proper, efficient domain types, and then operate on those domain types.
+
+For UPLC, the proper domain type is Scott or SOP terms.
+Thus a standard way of writing Plutus Tx is to immediately validate the incoming `Data` objects and turning them into Scott or SOP terms via `unsafeFromBuiltinData` (see [Plutus Ledger Language Version](../working-with-scripts/ledger-language-version.md)), before working with them.
+
+However, when it comes to Plutus scripts, working directly with `Data` can be more efficient, especially for scripts with simpler logic.
```
colll78, line 139: This line suggests that it is more efficient to use either SOP/Data for a given script depending on the complexity of the script logic, when in practice it is most efficient to use a combination of the two within the same script. Instead of general statements like Data encoding is more efficient for simple scripts, it should provide more practical advice ie:

Must know:
1. Plutus scripts arguments are encoded as `Data`
2. The cost of parsing, validating, and converting a `Data` encoded value to a `SOP` term is expensive. 

Indications to use SOP:
If you have to construct an intermediate value in a Plutus script you might consider constructing that value as an SOP term. However, do note that this only applies if the constructed value is an intermediate result (i.e. a value that is used in non-trivial computation after construction). If the constructed value is not used in non-trivial computation, and instead is used in validation, then you will likely benefit from constructing it with data-encoding. 

Examples:
```haskell
-- | intermediate result, `PoolState` should use SOP
-- assume foldUTxOs is a function that folds a function over a list of inputs and their corresponding (satisfying) outputs with an accumulator. 
processOrders :: BuiltinData -> BuiltinList BuiltinTxInInfo -> BuiltinList BuiltinTxOut -> BuiltinData
processOrders oldPoolStateData inputs outputs =
  -- first convert the old pool state into SOP
  let oldPoolState = unsafeFromBuiltinData @PoolState oldPoolStateData
       expectedNewPoolState = foldUTxOs inputs outputs oldPoolState 
   in
     -- This is the newPoolState that we have computed and thus that we expect to be the datum of the 
     -- pool UTxO output. We encoded it as SOP while it was an intermediate result (there was still additional computation to be done)
     -- but we have to convert back to data because we need to enforce that the pool output UTxO datum equals this,
     -- and the pool `TxOut` is data encoded, so if we wanted to compare it directly we would need to convert it to SOP  
     -- which would be very expensive. Instead, we convert our intermediate result from SOP to Data and check for 
     -- equality using `equalsData` which allows us to completely avoid parsing the pool output datum. 
     (toBuiltinData newPoolState) 
```

```haskell
-- | Use data encoding
-- Here we do not use SOP encoding because there is no intermediate result, we construct the entire expected pool output datum all at once, which we then will compare against the actual pool output datum as `BuiltinData`. We do not benefit from SOP encoding since the constructed value is not used in any subsequent calculations (ie. it is not an intermediate result).
processOneOrder :: BuiltinData -> BuiltinTxInInfo -> BuiltinTxOut -> BuiltinData 
processOneOrder oldPoolState ownInput ownOutput = 
  -- compute expectedPoolState
  ... 
```

colll78 (commented):
Add information about how Scott / SOP are almost always more efficient for intermediate result computation. 

The documentation here makes it seem whenData/Scott/SOP 
Ex a pool state UTxO and a number of requests

effectfully (approved):
Great stuff. I agree with @colll78's comment though.

doc/docusaurus/docs/delve-deeper/encoding.md
zliu41: I don't think these examples are very convincing.

- In the first example, you are converting something from Data to SOP, do some computation, then converting back to Data. This does not necessarily mean SOP is better - it all depends on how expensive the computation is. If it is a cheap one, you'd better off sticking with Data.
- The second example is similar - whether to use Data or SOP really depends on how expensive "compute expectedPoolState" is.

I'll provide some additional details to make the paragraph more clear.

doc/docusaurus/docs/delve-deeper/encoding.md
```diff
@@ -0,0 +1,162 @@
...
+It is identical to Scott encoding for non-recursive data types, but differs for recursive data types.
+The church encoding of `xs` and `head` is:
+
+```
+xs = \c n -> c 1 (c 2 (c 3 n))
+
+head = \xs -> xs (\y ys -> y) 42
+```
+
+However, the Church encoding for `tail` is much more complex and takes linear time (with respect to the length of the list) to evaluate, whereas it is constant time in Scott encoding.
```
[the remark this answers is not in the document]
zliu41: Rephrased.

zliu41 (commented):
Expanded the data encoding section and rephrased some parts.

doc/docusaurus/docs/delve-deeper/encoding.md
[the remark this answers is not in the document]
effectfully: Just to be clear, please don't feel any pressure to address my nitpicks whatsoever, I'm only leaving them in case you or anybody is interested.

doc/docusaurus/docs/delve-deeper/encoding.md
colll78: The examples are of the batcher architecture for processing many orders against a pool.

In the first example it does not matter how expensive the computation is, what matters is that it is an intermediate value that is continuously accessed in every iteration of `foldUTxOs`. The only way the SOP variant of this would be more efficient is if for some reason it is known that the length of the `BuiltinList` will not exceed some very low constant. 

In the second example it doesn't matter how expensive the computation is, since it is not an intermediate value (not used in subsequent computation) the fields only need to be extracted once, no matter how expensive the computation with the underlying data is, it wouldn't benefit from SOP at all. 

Here is a more concrete implementation of processOrder:

```haskell
data PoolState = 
  PoolState
    { reserveA :: Integer -- balance of token A in the pool
    , reserveB :: Integer -- balance of token B in the pool
    , totalShares :: Integer -- total lp shares
    , assetACS :: CurrencySymbol 
    , assetATN :: TokenName 
    , assetBCS :: CurrencySymbol
    , assetBTN :: TokenName 
    }

processOneOrder :: BuiltinData -> BuiltinTxInInfo -> BuiltinTxOut -> BuiltinData 
processOneOrder oldPoolState ownInput ownOutput = 
  let fields = BI.snd ( BI.unsafeDataAsConstr b )
      reserveA = BI.unsafeDataAsI (BI.head fields)
      !fields_1 = (BI.tail fields)
      reserveB = BI.unsafeDataAsI (BI.head fields_1)
      !fields_2 = (BI.tail fields_1)
      totalShares = BI.unsafeDataAsI (BI.head fields_2)
      !fields_3 = (BI.tail fields_2)
      assetAC = BI.head fields_3
      !fields_4 = (BI.tail fields_3)
      assetATN = BI.head fields_4
      !fields_5 = (BI.tail fields_4) 
      assetBCSData = BI.head fields_5
      assetBTNData = BI.head (BI.tail fields_5)

      ownInputValue = getField @"value" ownInput 
      ownInputA = builtinValueOf assetACS assetATN ownInputValue  
      -- Calculate the amount of AssetB to be sent out
      newReserveA = reserveA + ownInputA
      newReserveB = (reserveA * reserveB) `divide` newReserveA
      amountBOut = reserveB - newReserveB

      expectedOutputValue = 
        if assetBCS == adaSymbolData 
        then 
          let innerMap = BI.mkCons (BI.mkPairData (BI.mkB "") (BI.mkI amountBOut)) (BI.mkNilPairData BI.unitval)
          in BI.mkMap (BI.mkCons (BI.mkPairData (BI.mkB "") (BI.mkMap innerMap)) (BI.mkNilPairData BI.unitval))
        else 
          let innerMapAda = BI.mkCons (BI.mkPairData (BI.mkB "") (BI.mkI 2_000_000)) (BI.mkNilPairData BI.unitval)
              innerMapAssetA = BI.mkCons (BI.mkPairData assetBTN (BI.mkI amountBOut)) (BI.mkNilPairData BI.unitval)
          in BI.mkMap (BI.mkCons (BI.mkPairData (BI.mkB "") (BI.mkMap innerMapAda)) $ BI.mkCons (BI.mkPairData assetBCS (BI.mkMap innerMap)) (BI.mkNilPairData BI.unitval))
      in 
        if (getFields @"value" ownOutput == expectedOutputValue)
        then 
          BI.mkConstr 0 
            BI.mkCons (BI.mkI newReserveA) $
            BI.mkCons (BI.mkI newReserveB) $
            BI.mkCons (BI.mkI totalShares) $
            fields_3
        else error "output did not contain expected value"
```

You could change the constant product formula in the above to some crazy expensive computation, but it wouldn't change the fact that data encoding should be preferred here. The real heuristic isn't the complexity of the computation, but instead it is the degree to which the value is intermediate vs immediate. For an immediate value, you destructure all the necessary fields once after which you don't continue to use encapsulating `BuiltinData`. For an intermediate value, you continue to use the encapsulating structure, and instead of simply destructuring it and extracting all the fields and discarding it, you continue to use it (ie. update certain fields and pass it to another function, or just pass it as is to another function that also interacts with it).

zliu41: I see your point. So rather than framing it as cheap vs. expensive computation, you are suggesting using "extracted once" vs. "extract many times". Perhaps, though I'm not convinced that the latter is necessarily clearer.

For example, if you write "5th field of `foo` <= 8th field of foo", you should only need these fields once, but you can easily end up extracting fields 1 through 5 twice, unless you carefully optimize the code.

colll78: > I see your point. So rather than framing it as cheap vs. expensive computation, you are suggesting using "extracted once" vs. "extract many times". Perhaps, though I'm not convinced that the latter is necessarily clearer.
Exactly. 

> For example, if you write "5th field of foo <= 8th field of foo", you should only need these fields once, but you can easily end up extracting fields 1 through 5 twice, unless you carefully optimize the code.

This extraction should be done via TH utils to make sure the fields are all extracted in one traversal. 
```haskell
data Example = 
  Example 
    { fieldOne :: Integer
    , fieldTwo :: Integer 
    , fieldThree :: Integer 
    , fieldFour :: Integer
    , fieldFive :: Integer 
    , fieldSix :: Integer 
    , fieldSeven :: Integer
    , fieldEight :: Integer
    } 

exampleFunction :: BuiltinData -> Bool
exampleFunction exampleAsData = 
  $(generateFieldBindsLam ''Example ["fieldFive", "fieldEight"]) exampleAsData $ \fieldFive fieldEight ->
  fieldFive <= fieldEight 

-- $(generateFieldBindsLam ''Example ["fieldFive", "fieldEight"]) exampleAsData f
-- generates:
  let !fields = BI.tail $ BI.tail $ BI.tail $ BI.tail BI.snd ( BI.unsafeDataAsConstr b)
       fieldFive = BI.head fields
       fieldEight = BI.head $ BI.tail $ BI.tail $ BI.tail fields
   in f fieldFive fieldEight 
```

PR #6590, opened 2024-10-22
[Plinth] Optimize 'unionValue'

effectfully:
Addresses:

> `unionValue` is inefficient, because `unionWith` is, because `Map.union` is, because of all the `These` extra work that it does. If `unionValue` is important enough to be [considered](https://github.com/IntersectMBO/plutus/issues/6585) as a builtin, it is important enough to be optimized in Plinth

from #5967.


plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -416,14 +417,7 @@ unionVal (Value l) (Value r) =
...
 unionWith :: (Integer -> Integer -> Integer) -> Value -> Value -> Value
-unionWith f ls rs =
-    let
-        combined = unionVal ls rs
-        unThese k' = case k' of
-            This a    -> f a 0
-            That b    -> f 0 b
-            These a b -> f a b
-    in Value (fmap (fmap unThese) combined)
+unionWith f = coerce (Map.unionWith @CurrencySymbol (Map.unionWith @TokenName f))
```
effectfully, line 420: @zliu41 I don't really know much about `Value` stuff, is this actually correct?

plutus-benchmark/marlowe/test/role-payout/9.6/0101000100000101010000010101000100010101000001000001000000010101.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 4700192221958
-| mem: 4700000967314})
\ No newline at end of file
+({cpu: 4700174302334
+| mem: 4700000857208})
```
effectfully, line 2: 9% CPU speedup and 12% MEM speedup for such a low-hanging fruit.

plutus-benchmark/marlowe/test/role-payout/9.6/211e1b6c10260c4620074d2e372c260d38643a3d605f63772524034f0a4a7632.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 3700173474723
-| mem: 3700000850386})
\ No newline at end of file
+({cpu: 3700153208683
+| mem: 3700000726876})
```
effectfully, line 2: 11.5% CPU speedup and 15% MEM speedup.

plutus-benchmark/marlowe/test/role-payout/9.6/bd79f4a84db23b7c4cd219d498bd581e085cbc3437957e74a8862281a700700b.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 4100193313124
-| mem: 4100000968548})
\ No newline at end of file
+({cpu: 4100167271084
+| mem: 4100000808938})
```
effectfully, line 2: 13.5% CPU speedup and 16.5% MEM speedup.

plutus-benchmark/marlowe/test/role-payout/9.6/ec4712ee820eb959a43ebedfab6735f2325fa52994747526ffd2a4f4f84dd58e.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 4100190778419
-| mem: 4100000946910})
\ No newline at end of file
+({cpu: 4100161979547
+| mem: 4100000772592})
```
effectfully, line 2: 15% CPU speedup and 18% MEM speedup.

zliu41 (approved):
Looks good.

ana-pantilie (changes requested):

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
effectfully: Yes, semantics is different. But isn't it correct _now_? It was quite weird before, I'd argue.

zliu41: Good point. It should be documented that it treats `[]` and `[(token, 0)]` as different Values (so does the old implementation, but in a slightly different way).

ana-pantilie: So what is the expected behaviour of `unionWith`? What does it mean that this version is correct?

effectfully: Well, I by default expect the standard behavior, i.e. the [one](https://hackage.haskell.org/package/containers-0.7/docs/Data-Map-Internal.html#v:unionWith) in containers.

On the other hand, maybe I'm wrong, maybe the old behavior of `unionValue (-)` is the correct one. But in that case we should introduce

```haskell
unionWithDef :: forall k a. Eq k => a -> (a -> a -> a) -> Map k a -> Map k a -> Map k a
```

which feed the first argument into the second whenever a key is missing in one of the maps.

So yeah, I think you've convinced me that the old behavior was correct and my new one isn't. @zliu41 opinion?

zliu41: The old behavior doesn't treat a token with zero quantity and a token that doesn't exist the same way either. e.g.,

```
unionWith (\_ _ -> 42) [] [] == []
unionWith (\_ _ -> 42) [(token, 0)], [] == [(token, 42)]
```

This doesn't look any more correct than the new behavior.

> So what is the expected behaviour of unionWith? What does it mean that this version is correct?

I think we can document that the expected behavior is exactly how the new implementation behaves, and move on. I don't think this really makes a difference on anyone, since nobody is going to actually use something like `\_ _ -> 42` as the combining function.

ana-pantilie: If possible, it's also good to add a test which demonstrates this corner case.

zliu41: Yes some tests would be nice.

effectfully:
I'm closing this PR in favor of the #6656 issue.

PR #6591, opened 2024-10-22
Ramsay t/u casereduce

### LINKED ISSUE #6369: Translation relation and decision procedure for UPLC CaseReduce

ramsay-t:
Translation Relation and Decision Procedure for CaseReduce. 

The translation relation seems valid based on the Haskell code. There is a small example at the bottom of the Agda file that demonstrates the process on a single instance. For a repeated application, the compiler applies some true beta-reduction in between applications, which (deliberately) isn't implemented in these translation relations.  It will be important to see if that occurs as separate phases in the implementation. 

Fixes #6369


effectfully (approved):
Seems like we finally have one translation relation that isn't tricky.

PR #6595, opened 2024-10-24
[Test] Get the seed in 'noStructuralErrors'

effectfully:
Copying a Slack comment:

Apparently with `QuickCheck` you can throw exceptions in `IO` and of type `IO smth`, but what you can't do is throw an exception of type `IOException`, i.e. you can't throw an exception while you're throwing an exception, because in that case you won't get a seed. It is of course entirely accidental that we throw an exception while throwing an exception, but I didn't expect that to break `QuickCheck`.

### DIFF
--- a/plutus-core/plutus-ir/test/PlutusIR/Generators/QuickCheck/Tests.hs
+++ b/plutus-core/plutus-ir/test/PlutusIR/Generators/QuickCheck/Tests.hs
@@ -210,7 +210,8 @@ noStructuralErrors term =
 prop_noStructuralErrors :: Property
 prop_noStructuralErrors = withMaxSuccess 99 $
   forAllDoc "ty,tm" genTypeAndTerm_ shrinkClosedTypedTerm $ \(_, termPir) -> ioProperty $ do
-    termUPlc <- fmap UPLC._progTerm . modifyError throw . toUPlc $ Program () latestVersion termPir
+    termUPlc <- fmap UPLC._progTerm . modifyError (userError . displayException) . toUPlc $
+        Program () latestVersion termPir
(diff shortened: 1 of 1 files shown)


plutus-core/plutus-ir/test/PlutusIR/Generators/QuickCheck/Tests.hs
```diff
@@ -210,7 +210,8 @@ noStructuralErrors term =
 prop_noStructuralErrors :: Property
 prop_noStructuralErrors = withMaxSuccess 99 $
   forAllDoc "ty,tm" genTypeAndTerm_ shrinkClosedTypedTerm $ \(_, termPir) -> ioProperty $ do
-    termUPlc <- fmap UPLC._progTerm . modifyError throw . toUPlc $ Program () latestVersion termPir
+    termUPlc <- fmap UPLC._progTerm . modifyError (userError . displayException) . toUPlc $
```
Unisay, line 213: I don't quite understand this solution.
`userError` throws [IOError](https://hackage.haskell.org/package/base/docs/Prelude.html#v:userError) - how exactly does it help us not to lose the seed?

effectfully: It doesn't throw! It _returns_ an error where previously we'd throw one. I did check that it didn't work before abd works now.

Unisay: Gotcha! Forgot that `userError` simply constructs `IOError`, doesn't throw it.

Unisay (approved):

PR #6600, opened 2024-10-28
Release 1.37.0.0

zliu41:


ana-pantilie (changes requested):
@kwxm we can't release this, can we? The costing is still messed up as we discovered in  https://input-output-rnd.slack.com/archives/G01EE4NQ9U0/p1729865299741289

PR #6627, opened 2024-11-04
Kwxm/conformance/two digit test names

kwxm:
For some of the builtins we have conformance tests with names like `case-1`, `case-2`, ... `case-9`, `case-10`, `case-11`, .... This is annoying because sometimes they get listed and/or processed in alphanumeric order, not numeric order: `case-1`, `case-11`, `case-12`, `case-2`, `case-3`, ... .  This PR renames all of the files to have two-digit numbers, so we get things in the right order:   `case-01`, `case-02`, ..., `case-09`, `case-10`, `case-11`, ... .  This is tedious but it should makes things easier in future.  I also tried to make the naming  more consistent in general.


effectfully (approved):
I agree that this is much better. How did you do it?

kwxm:
> I agree that this is much better. How did you do it?

I used shell scripts involving `find` and `sed` along with a lot of `git reset --hard` when I got it wrong.

PR #6631, opened 2024-11-04
MintValue properties

Unisay:
`MintValue` properties tested both in Haskell and in Plinth.

Closes https://github.com/IntersectMBO/plutus-private/issues/1230


plutus-ledger-api/test-plugin/Spec/MintValue/V3.hs
```diff
@@ -0,0 +1,153 @@
...
+    , test_Plinth_AssetClassIsEitherMintedOrBurned
+    , test_Plinth_MintValueMintedIsPositive
+    , test_Plinth_MintValueBurnedIsPositive
+    ]
+
+test_Plinth_MintValueBuiltinData :: TestTree
+test_Plinth_MintValueBuiltinData =
+  testProperty "Builtin data representation of MintValue and Value is the same"
+    $ cekProp
+    . \values ->
```
ana-pantilie, line 120: Does this use some `Arbitrary` instance for `Value`?

plutus-ledger-api/test-plugin/Spec/MintValue/V3.hs
```diff
@@ -0,0 +1,153 @@
...
+--------------------------------------------------------------------------------
+-- MintValue properties --------------------------------------------------------
+
+prop_MintValueBuiltinData :: Either Value MintValue -> Bool
+prop_MintValueBuiltinData values =
+  let (value, mintValue) =
+        case values of
+          Left v   -> (v, coerce v)
+          Right mv -> (coerce mv, mv)
+   in toBuiltinData mintValue == toBuiltinData value
```
ana-pantilie, line 49: Can you convert between Value and MintValue just like that? I thought Value was a particular case of MintValue, meaning some MintValues cannot be converted to Values.

plutus-ledger-api/test-plugin/Spec/MintValue/V3.hs
Unisay: Yes, it imports existing `instance Arbitrary Value`

plutus-ledger-api/test-plugin/Spec/MintValue/V3.hs
Unisay: I can't think of a `MintValue` that can't be a valid `Value`, e.g. `Value`s can contain negative quantities, just not in tx outputs.

Unisay: The scope of this PR is quite narrowly focused on the `mint` field and addresses the problem with Plinth developers [shooting themselves into the foot](https://github.com/IntersectMBO/plutus/pull/5781) with the `currencySymbolValueOf` function.

plutus-ledger-api/test-plugin/Spec/MintValue/V3.hs
```diff
@@ -0,0 +1,153 @@
...
+
+prop_AssetClassIsEitherMintedOrBurned :: MintValue -> Bool
+prop_AssetClassIsEitherMintedOrBurned (mint :: MintValue) =
+  let assetClasses v = [AssetClass (cs, tn) | (cs, tn, _q) <- flattenValue v]
+      minted = assetClasses (mintValueMinted mint)
+      burned = assetClasses (mintValueBurned mint)
+   in List.all (`List.notElem` burned) minted
```
effectfully, line 56: I'd probably express this as "the intersection of the two lists is empty", but this is fine too. Just checked out of curiosity and it's pretty much what `Data.List.intersect` does under the hood anyway.

plutus-ledger-api/test-plugin/Spec/MintValue/V3.hs
```diff
@@ -0,0 +1,153 @@
...
+  testProperty "MintValue burned is positive" $ cekProp . \mintValue ->
+    $$(compile [||prop_MintValueBurnedIsPositive||])
+      `unsafeApplyCode` liftCodeDef mintValue
+
+--------------------------------------------------------------------------------
+-- Helper functions ------------------------------------------------------------
+
+scaleTestsBy :: (QC.Testable prop) => Haskell.Int -> prop -> QC.Property
+scaleTestsBy factor =
+  QC.withMaxSuccess (100 Haskell.* factor) . QC.mapSize (Haskell.* factor)
```
effectfully, line 149: I guess we should move that to some common `QuickCheck` module. But we can do that in another PR.

effectfully (approved):

PR #6636, opened 2024-11-05
`PlutusLedgerApi.V1.withCurrencySymbol`

Unisay:
Closes https://github.com/IntersectMBO/plutus/issues/5992

### DIFF
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs
@@ -50,6 +50,7 @@ module PlutusLedgerApi.V1.Data.Value (
     , Value(..)
     , singleton
     , valueOf
+    , withCurrencySymbol
     , currencySymbolValueOf
     , lovelaceValue
     , lovelaceValueOf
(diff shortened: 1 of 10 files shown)


plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -8,26 +10,24 @@
...
+{-# OPTIONS_GHC -fexpose-all-unfoldings #-}
 {-# OPTIONS_GHC -fno-omit-interface-pragmas #-}
 {-# OPTIONS_GHC -fno-spec-constr #-}
 {-# OPTIONS_GHC -fno-specialise #-}
-{-# OPTIONS_GHC -fexpose-all-unfoldings #-}
-{-# LANGUAGE BlockArguments       #-}
-{-# LANGUAGE DataKinds            #-}
-{-# LANGUAGE TypeFamilies         #-}
-{-# LANGUAGE UndecidableInstances #-}
--- We need -fexpose-all-unfoldings to compile the Marlowe validator
```
zliu41, line 22: This comment should be preserved

zliu41 (commented):
Could you link to the issue? I don't remember why this is needed. Also please add tests.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
Unisay: My logic was this:
In the Plutus documentation we have a list of GHC options that should be enabled for Plugin to work. Many Plinth modules use them. We should use the recommended options everywhere and there is nothing that makes this particular instance special enough to require and extra comment.

zliu41: `-fexpose-all-unfoldings` is not an option that should be enabled everywhere, as explained in https://plutus.cardano.intersectmbo.org/docs/using-plutus-tx/extensions-flags-pragmas

Unisay: Ok, I made a mistake assuming that this option should be used everywhere in Plinth by default.

Unisay: Do you know why `{-# INLINEABLE #-}` is not enough for the Marlowe validator?

zliu41: Missing unfolding for something, I think - we never identified the root cause.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -393,7 +408,7 @@ lovelaceValueOf v = Lovelace (valueOf v adaSymbol adaToken)
 {-# INLINABLE assetClassValue #-}
 -- | A 'Value' containing the given amount of the asset class.
 assetClassValue :: AssetClass -> Integer -> Value
-assetClassValue (AssetClass (c, t)) i = singleton c t i
+assetClassValue (AssetClass (c, t)) = singleton c t
```
Unisay, line 411: Eta reduction fixes hlint warning

zliu41 (approved):

plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs
```diff
@@ -21,11 +21,15 @@
...
 {-# OPTIONS_GHC -fno-specialise #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+
 {-# OPTIONS_GHC -fexpose-all-unfoldings #-}
--- We need -fexpose-all-unfoldings to compile the Marlowe validator
--- with GHC 9.6.2.
+-- We need -fexpose-all-unfoldings to compile the Marlowe validator with GHC 9.6.2.
 -- TODO. Look into this more closely: see https://github.com/IntersectMBO/plutus/issues/6172.
 
+{-# HLINT ignore "Redundant if" #-}
```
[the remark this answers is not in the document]
Unisay: I run hlint locally and generally find it helpful. 

Having hlint warning in the codebase is a noise for me as it breaks my workflow when I need to "Jump to the next problem" so I'd really like to either address warnings or suppress them. 

Suppressing warnings in `hlint.yaml` works for me, even though its not as convenient as choosing a "Ignore the XXX" in vscode.

> which make the compilation slower.
Is this really the case?

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -8,21 +10,22 @@
...
 {-# OPTIONS_GHC -fexpose-all-unfoldings #-}
-{-# LANGUAGE BlockArguments       #-}
-{-# LANGUAGE DataKinds            #-}
-{-# LANGUAGE TypeFamilies         #-}
-{-# LANGUAGE UndecidableInstances #-}
 -- We need -fexpose-all-unfoldings to compile the Marlowe validator
--- with GHC 9.6.2.
 -- TODO. Look into this more closely: see https://github.com/IntersectMBO/plutus/issues/6172.
 
+{-# HLINT ignore "Redundant if" #-}
```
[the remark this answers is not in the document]
Unisay: moved to `hlint.yaml`

plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs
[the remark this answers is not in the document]
Unisay: moved to `hlint.yaml`

PR #6638, opened 2024-11-05
[Plinth] [Builtins] Remove 'CaseList' and 'CaseData'

effectfully:
This removes usages of `CaseList` and `CaseData` from Plinth by compiling those to `ChooseList`- and `ChooseData`-based functions. This is just so that we can test the constitution script, the intention is to revert this PR once `CaseList` and `CaseData` are costed.


zliu41 (commented):
> This is just so that we can test the constitution script, the intention is to revert this PR once CaseList and CaseData are costed.

We can't revert this until `caseList` etc. are enabled on mainnet.

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
```diff
@@ -409,7 +410,49 @@ defineBuiltinTerms = do
...
+                -- > /\a r ->
+                -- >   \(z : r) (f : a -> list a -> r) (xs : list a) ->
+                -- >     chooseList
+                -- >       {a}
+                -- >       {all dead. r}
+                -- >       xs
+                -- >       (/\dead -> z)
+                -- >       (/\dead -> f (headList {a} xs) (tailList {a} xs))
+                -- >       {r}
+                fmap (const annMayInline) . runQuote $ do
```
zliu41, line 423: Why not just revert the changes to `PlutusTx.Builtins` and `PlutusTx.Builtins.Internal`? This is more complex and harder to follow.

plutus-tx-plugin/test/IsData/9.6/deconstructData.pir.golden
effectfully, line 1: I took the comments from this golden file.

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
effectfully: BTW, note that I didn't just write it down from scratch like a maniac, I copy-pasted definitions from `PlutusCore.StdLib.Data.List` and `PlutusCore.StdLib.Data.Data` and tweaked them just a little to get the right argument ordering and `all dead. <...>` instead of `() -> <...>`.

zliu41 (approved):

PR #6640, opened 2024-11-06
Remove extra-doc-files:README.md from plutus-executables.cabal

zeme-wana:
This file does not exist and causes `cabal haddock` to fail.


ana-pantilie (commented):
Sorry, I missed this yesterday and apologies for introducing this issue! Thank you for finding and fixing it.

PR #6646, opened 2024-11-09
Use taskset in ci-plutu-benchmark.sh

zeme-wana:
You can now run 
`/benchmark BENCH_NAME cap=CAPABILITY_NUM` 
In addition to 
`/benchmark BENCH_NAME`  

The former will run BENCH_NAME on the given CPU core

### DIFF
--- a/scripts/ci-plutus-benchmark.sh
+++ b/scripts/ci-plutus-benchmark.sh
@@ -16,6 +16,9 @@
 # fetch/build this each time since we want to run this in a clean environment.
 # The `jq` invocation below is necessary because we have to POST the PR comment as JSON data 
 # (see the curl command) meaning the script output has to be escaped first before we can insert it.
+# Also note the use of the envvar CAPABILITY_NUM and `taskset -c` to limit 
+# the benchmark to a single core. Experiments have shown that this can lead to more stable results.
+# This is only available on linux.
 
 set -e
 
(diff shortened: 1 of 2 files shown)


scripts/ci-plutus-benchmark.sh
```diff
@@ -36,6 +39,13 @@ else
    git checkout "$PR_BRANCH"
 fi
 
+if [ -z "$CAPABILITY_NUM" ] ; then
```
kwxm, line 42: Is there any way of setting this from GitHub?

kwxm (approved):
Looks good! Thanks.

scripts/ci-plutus-benchmark.sh
zeme-wana: Easiest way would be to:
`/benchmark BENCH_NAME CAP_NUM` 
Instead of
`/benchmark BENCH_NAME`
Sounds good?

kwxm:
Hydra seems to be broken, so I'm going to ignore it and merge this anyway.

scripts/ci-plutus-benchmark.sh
```diff
@@ -55,8 +65,14 @@ cabal update
 echo "[ci-plutus-benchmark]: Clearing caches with cabal clean ..."
 cabal clean
 
+if [[ -z $(which taskset) ]]; then
```
kwxm, line 68: Oh wait: I think that should be `-n`.  I'll open a quick PR to fix that.

PR #6647, opened 2024-11-09
Remove doctest across whole repo

zeme-wana:
Too high maintenance cost for too little value (it's causing https://github.com/IntersectMBO/plutus/issues/6625 too)


effectfully (approved):
:+1: 

We should probably be using [`cabal-docspec`](https://github.com/phadej/cabal-extras/blob/68e436d06c0f00bcc2173c6536650d53c3953d07/cabal-docspec/MANUAL.md) anyway.

PR #6648, opened 2024-11-10
Testing PR #6646

kwxm:
This is just to check that #6646 works OK.

DO NOT MERGE!


kwxm:
These results are terrible, but fortunately  I think we were doing the wrong thing.  I'm going to try again.

kwxm:
Tying again in #6650.  It's impressive just how bad these results are though.

kwxm:
Reopening this so I can observe what's happening on the machine.

kwxm:
Nope.  The base branch has vanished, so this won't work.

PR #6655, opened 2024-11-12
BuiltinByteString literals aren't UTF8 encoded.

Unisay:
- [x] `BuiltinByteString` literals are utf-8 decoded in the plugin.
- [x] `Haskell.IsString BuiltinByteString` instance delegates to the Haskell `ByteString` instance.
- [x] Unit tests to verify that bytes in range from 0 to 255 (\x00 - \xFF) found in string literals are passed to the `BuiltinByteString` without the UTF8 conversion.

Closes https://github.com/IntersectMBO/plutus/issues/6505


plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/bls12_381_G1_scalarMul/mul-neg-one/mul-neg-one.uplc.expected
Unisay, line 1: Idk what this golden file was empty

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -699,13 +732,21 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
     _                             -> throwPlain $ CompilationError "No info for Pair builtin"
 
   -- TODO: Maybe share this to avoid repeated lookups. Probably cheap, though.
-  (stringTyName, sbsName) <- case (Map.lookup ''Builtins.BuiltinString nameInfo, Map.lookup 'Builtins.stringToBuiltinString nameInfo) of
+  (stringTyName, sbsName) <-
+    case
```
Unisay, line 736: Wrap long line

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/bls12_381_G1_scalarMul/mul-neg-one/mul-neg-one.uplc.expected
kwxm: That may well be my fault.  I [renamed a lot of the conformance inputs](https://github.com/IntersectMBO/plutus/pull/6627) recently ad I may have got something wrong.  I'll check.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-07/case-07.uplc.budget.expected
```diff
@@ -1,2 +1,2 @@
-({cpu: 260294
-| mem: 601})
\ No newline at end of file
+({cpu: 260453
```
kwxm, line 1: Why have all these budgets changed?  These are very simple  textual UPLC programs and I don't see why the changes in this PR would have affected their execution costs.

Unisay: I tried to figure that out but couldn't understand the tests. There is TH involved which makes it harder to reason statically. How can I see those UPLC programs and what has changed in them?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -136,23 +141,51 @@ compileLiteral = \case
...
--- | Get the bytestring content of a string expression, if possible. Follows (Haskell) variable references!
-stringExprContent :: GHC.CoreExpr -> Maybe BS.ByteString
-stringExprContent = \case
-  GHC.Lit (GHC.LitString bs) -> Just bs
-  -- unpackCString# / unpackCStringUtf8# are just wrappers around a literal
-  GHC.Var n `GHC.App` expr
-    | let name = GHC.getName n
-    , name == GHC.unpackCStringName || name == GHC.unpackCStringUtf8Name ->
-      stringExprContent expr
+data StringExprContentAs = AsBytes | AsText
```
Unisay, line 144: The `stringExprContent` function is used to extract bytes for the `BuiltinByteString` as well as for the `BuiltinString` literals; 
I have added this datatype as its parameter to distinguish between these scenarios as we want to utf-8 decode `BuiltinByteString` literals only.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -136,23 +141,51 @@ compileLiteral = \case
...
+  GHC.Var (GHC.maybeUnfoldingTemplate . GHC.realIdUnfolding -> Just unfolding) ->
+    stringExprContent contentAs unfolding
+  _ -> Nothing
+
+{- | Decoding that undoes GHC's UTF-8 encoding of bytestring literals:
+
+This isn't a full UTF-8 decoder: it only decodes the subset of UTF-8 that
+is expected to be found in bytestring literals: 0x00 - 0xFF
+-}
+fromUtf8 :: [Word8] -> Maybe [Word8]
```
Unisay, line 182: I tried using `Data.Text.Encoding.decodeUtf8` and it didn't work: the error was saying about invalid utf-8 bytes.
It turned out that decoding rules are not super complex so I have implemented them below.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -699,13 +732,21 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
+      ( Map.lookup ''Builtins.BuiltinString nameInfo
+      , Map.lookup 'Builtins.stringToBuiltinString nameInfo
+      ) of
     (Just t1, Just t2) -> pure (GHC.getName t1, GHC.getName t2)
     _                  -> throwPlain $ CompilationError "No info for String builtin"
 
-  (bsTyName, sbbsName) <- case (Map.lookup ''Builtins.BuiltinByteString nameInfo, Map.lookup 'Builtins.stringToBuiltinByteString nameInfo) of
-    (Just t1, Just t2) -> pure (GHC.getName t1, GHC.getName t2)
-    _                  -> throwPlain $ CompilationError "No info for ByteString builtin"
+  (builtinByteStringTyName, sbbsName) <-
```
Unisay, line 743: Wrap another long line

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -730,39 +771,54 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
-                        throwSd UnsupportedError $
-                          "Use of fromString on type other than builtin strings or bytestrings:" GHC.<+> GHC.ppr ty
-              Nothing ->
-                throwSd CompilationError $
-                  "Use of fromString with inscrutable content:" GHC.<+> GHC.ppr content
+          case GHC.tyConAppTyCon_maybe ty of -- extract Type constructor without arguments
+            Just tc ->
+              if
+                | GHC.getName tc == builtinByteStringTyName ->
+                    case stringExprContent AsBytes (strip content) of
```
Unisay, line 781: Extracting content as bytes undoes the UTF8 encoding for `BuiltinByteString`

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -730,39 +771,54 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
+                      Nothing ->
+                        throwSd CompilationError $
+                          "Use of fromString @BuiltinByteString with inscrutable content:"
+                            GHC.<+> GHC.ppr content
+                      Just bs ->
+                        pure $ PIR.Constant annMayInline $ PLC.someValue bs
+                | GHC.getName tc == stringTyName ->
+                    case stringExprContent AsText (strip content) of
```
Unisay, line 789: Extracting content as text leaves original UTF8 encoding for `BuiltinString`

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -730,39 +771,54 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
           pure $ PIR.Constant annMayInline $ PLC.someValue bs
     -- 'stringToBuiltinString' invocation
-    (strip -> GHC.Var n) `GHC.App` (strip -> stringExprContent -> Just bs) | GHC.getName n == sbsName ->
-      case TE.decodeUtf8' bs of
-        Right t -> pure $ PIR.Constant annMayInline $ PLC.someValue t
-        Left err ->
-          throwPlain $
-            CompilationError $
-              "Text literal with invalid UTF-8 content: " <> (T.pack $ show err)
+    (strip -> GHC.Var n) `GHC.App` (strip -> stringExprContent AsText -> Just bs)
```
Unisay, line 814: Wrapped long line

plutus-tx-plugin/test/AssocMap/Spec.hs
```diff
@@ -91,21 +91,23 @@ map2 =
         \n ->
           let m1 =
                 Data.AssocMap.unsafeFromList
-                  [ (n PlutusTx.+ 1, "one")
```
Unisay, line 94: The fact that this test relies on UTF-8 encoding for BuiltinByteStrings is accidental

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-07/case-07.uplc.budget.expected
kwxm: Nothing's changed!  The conformance test machinery [traverses the directory hierarchy](https://github.com/IntersectMBO/plutus/blob/e0b8db206b16f296529f5b45939523f4683f0f2e/plutus-conformance/src/PlutusConformance/Common.hs#L72) looking for directories with no subdirectories and when it finds one it looks for a `.uplc` file and if it finds one it loads and executes it and then checks the result and budget against golden files.  Look inside the directories [here](https://github.com/IntersectMBO/plutus/tree/master/plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/replicateByte) for example.  The input files haven't changed, so the results shouldn't change either.  The only thing that I can think of is that I messed something up when I renamed the files.  I'll see if I can work out what's going on.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/bls12_381_G1_scalarMul/mul-neg-one/mul-neg-one.uplc.expected
kwxm: @Unisay I forgot to rename some of the files last week: PR #6659 should fix that.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-07/case-07.uplc.budget.expected
kwxm: @Unisay I'm not sure what's happened, but you seem to have checked in golden files with incorrect CPU budgets but correct memory budgets.  I checked out a completely new copy of the repository and ran `cabal test  haskell-conformance --test-option "-p /writeBits/" and the tests all failed and said that the budget results should have been the old versions (ie,what's in `master`).  Running the test files manually with `uplc -c` also returns the budgets that are in `master`.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/bls12_381_G1_scalarMul/mul-neg-one/mul-neg-one.uplc.expected
Unisay: > @Unisay I forgot to rename some of the files last week: PR #6659 should fix that.

I've rebased this PR on top of that one already.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/replicateByte/case-07/case-07.uplc.budget.expected
Unisay: They fail locally and `--accept` flag doesn't seem to update golden files.

image

Unisay: What does `unexpected success` mean in the context of comparing execution budgets?

Unisay: Ok, I've reverted all the golden file updated uner the `agda-conformance` and tests pass now (locally)

kwxm: > What does `unexpected success` mean in the context of comparing execution budgets?

Some tests are expected to fail, and "unexpected success" means that they haven't.  

After some digginng it turns out that this comes from [Test.Tasty.ExpectedFailure](https://hackage.haskell.org/package/tasty-expected-failure-0.12.3/docs/Test-Tasty-ExpectedFailure.html).

plutus-tx-plugin/test/AssocMap/Spec.hs
Unisay: I've re-worked the test to use `Integer`s instead of `BuiltinByteString`s.

.hlint.yaml
```diff
@@ -27,3 +27,4 @@
 - ignore: {name: Use first, within: [UntypedPlutusCore.Evaluation.Machine.Cek]}
 - ignore: {name: Redundant if, within: [PlutusLedgerApi.V1.Value, PlutusLedgerApi.V1.Data.Value]}
 - ignore: {name: Replace case with maybe, within: [PlutusLedgerApi.V1.Value, PlutusLedgerApi.V1.Data.Value]}
+- ignore: {name: Use bimap, within: [PlutusTx.Builtins.HasOpaque]}
```
effectfully, line 30: Oh no no no no no, not this fuckery again. If we make it mandatory or it is already mandatory, I will not comply.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
effectfully: I'd probably break my forehead against the wall of trying to reuse GHC stuff, but you really did the right thing here. Great work!

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -730,39 +769,54 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
+                      Just bs ->
+                        pure $ PIR.Constant annMayInline $ PLC.someValue bs
+                | GHC.getName tc == stringTyName ->
+                    case stringExprContent AsText (strip content) of
+                      Nothing ->
+                        throwSd CompilationError $
+                          "Use of fromString @BuiltinString with inscrutable content:"
+                            GHC.<+> GHC.ppr content
+                      Just bs ->
+                        case TE.decodeUtf8' bs of
```
effectfully, line 793: Hm, `decodeUtf8'` didn't work for bytes, but works for text? What if text _is_ bytes though? Or am I misunderstanding this comment:

> I tried using Data.Text.Encoding.decodeUtf8 and it didn't work: the error was saying about invalid utf-8 bytes. Maybe the implementation in GHC is not compatible with the one in Text (Currently GHC ships with at least five UTF-8 implementations)

?

plutus-tx-plugin/test/ByteStringLiterals/Spec.hs
```diff
@@ -0,0 +1,114 @@
...
+
+test_CompileBuiltinByteStringLiteral_IsString :: TestTree
+test_CompileBuiltinByteStringLiteral_IsString =
+  testCase "OverloadedStrings" do
+    display (_progTerm (getPlcNoAnn compiledLiteral)) @?= expectedUplc
+  where
+    compiledLiteral :: CompiledCode BuiltinByteString =
+      $$( compile
+            [||
+            "\x00\x01\x02\x03\x04\x05\x06\x07\x08\x09\x0A\x0B\x0C\x0D\x0E\x0F\
```
effectfully, line 48: This is probably going to sound ridiculous, but can you add another such tests, but with bytes being in the opposite order? Just to make sure that the tests pass not because we got lucky with ordering of bytes. I'm perhaps saying nonsense, but it'll also perhaps be easier to add another couple of tests that bring me to my senses.

effectfully (approved):
Great stuff!

.hlint.yaml
Unisay: I don't like ignoring HLint warnings in a place that's far away from the actual warning, but this is better for me than not doing anything, as HLint warnings interfere with the legit warnings and break my workflow.
What's your take on HLint warnings?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
Unisay: I guess that I don't have unit test with the bytes that break `Text.decodeUtf8`, m.b. it was `\xFF` or `\x00` that `Text` choke on.

plutus-tx-plugin/test/ByteStringLiterals/Spec.hs
Unisay: I see your point! I had similar thought: what if the way bytes are ordered influences how they are decoded? While I don't expect this to be the case, adding a test case is cheap and easy so I'll simply do it.

Unisay: Done.

.hlint.yaml
zliu41: Can you instruct HLint to use your own config file instead of the one in the repo?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
zliu41: This is surprising. An example in the comment would be helpful.

Also, why not fail fast here? That seems better than returning a `Nothing`.

.hlint.yaml
effectfully: > What's your take on HLint warnings?

I hate HLint and don't use it. I do not mind you using it of course.

> Can you instruct HLint to use your own config file instead of the one in the repo?

Apparently there are two in the repo. One for CI and one for I don't really know what.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
Unisay: > Also, why not fail fast here? That seems better than returning a Nothing.

the `fromUtf8` function is used in the 
```haskell
stringExprContent :: StringExprContentAs -> GHC.CoreExpr -> Maybe BS.ByteString
```
which is not monadic and doesn support our `throwPlain` and `throwSd` functions.

`stringExprContent` is used in pattern positions (via `ViewPatterns`) so failing fast there is associated with the much larger code rewrite.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
Unisay: It was `\xC0`:
```
Cannot decode byte '\xc0': Data.Text.Encoding: Invalid UTF-8 stream
```

PR #6662, opened 2024-11-14
[Evaluation] [Performance] Remove scope checking

effectfully:
Just being curious.


effectfully:
So scope checking costs us about 3% of total time for the `validation-full` benchmarks and 20% of preparation time for the `validation` benchmarks.

PR #6663, opened 2024-11-14
[Evaluation] [Performance] Tweak 'safeIndexOne'

effectfully:
Just being curious.


effectfully:
^ looks amazing, but suspiciously heterogeneous within the same UPLC programs.

effectfully:
> looks amazing, but suspiciously heterogeneous within the same UPLC programs.

Although in most cases heterogeneity is consistent across similar programs, so that's definitely not noise:


Screenshot from 2024-11-15 00-59-03

Screenshot from 2024-11-15 00-59-20

effectfully:
^ these are results for the old commit, let's see if the new is detectably better (the Core is better, but not significantly so).

effectfully:
Well, seems like making it worker-wrapper-friendly allowed us to claw a few more percent out of cold GHC hands. Pretty great!

zliu41 (approved):

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
```diff
@@ -127,25 +131,47 @@ unsafeIndexOne (BHead w t ts) !i =
...
+thing into a beautiful recursive join point full of @Word64#@s, i.e. allocating very little if
+anything at all.
+-}
+
+-- See Note [Optimizations of safeIndexOneCont].
+safeIndexOneCont :: forall a b. b -> (a -> b) -> RAList a -> Word64 -> b
+safeIndexOneCont z f = findTree where
+    findTree :: RAList a -> Word64 -> b
+    -- See Note [Optimizations of safeIndexOneCont].
+    findTree Nil !_ = z
```
bezirg, line 151: Can you explain why you need to bang the wildcard here: `findTree Nil !_`

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -878,11 +878,11 @@ enterComputeCek = computeCek
...
-        case varEnv `Env.indexOne` coerce varIx of
-            Nothing  ->
-                throwingWithCause _MachineError OpenTermEvaluatedMachineError $
-                    Just $ Var () varName
-            Just val -> pure val
+        Env.safeIndexOneCont
+            (throwingWithCause _MachineError OpenTermEvaluatedMachineError . Just $ Var () varName)
+            pure
+            varEnv
+            (coerce varIx)
```
bezirg, line 885: Not that it matters, but place this change also on the SteppableCek (for trying to keep the code similar :D )

bezirg:
There is a criterion benchmark inside index-envs, I am curious of the results there as well, so I will give it a try.

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
effectfully: It's in the referenced Note:

> Bangs in the local definitions of 'safeIndexOneCont' are needed to tell GHC that the functions are
> strict in the 'Word64' argument, so that GHC produces workers operating on @Word64#@.

Otherwise GHC doesn't know that it can unbox the second argument.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: Good point, thank you.

effectfully:
> There is a criterion benchmark inside index-envs if you want to use it. I will try to run it myself to see the before and after numbers

I personally don't really care. The CEK machine is significantly faster and that's enough for me.

PR #6667, opened 2024-11-14
[Benchmark] Include 'mkTermToEvaluate' in 'benchmark-decode'

effectfully:
`mkTermToEvaluate` is a part of the script preparation phase. It looks rather inefficient (in two different ways) and we don't seem to have any dedicated benchmarks for catching that other than `validation-full`, but that's too insensitive due to it evaluating the scripts. I really don't believe there's a good reason for `mkTermToEvaluate` to be inefficient, so I think we should include it in the decoding benchmarks. 

Let's see how much overhead `mkTermToEvaluate` contributes.


effectfully:
Just checking that the benchmarking machine does not see no changes in performance as changes.

PR #6672, opened 2024-11-15
Refactor: reduce code duplication and make more uniform how plugin looks up GHC things (Names, Ids, etc.)

Unisay:
Some chores before adding new functionality:

Extracted 3 utility lookup functions:
  1. `lookupGhcId :: Compiling uni fun m ann => TH.Name -> m GHC.Id`
  2. `lookupGhcName :: Compiling uni fun m ann => TH.Name -> m GHC.Name`
  3. `lookupGhcTyCon :: Compiling uni fun m ann => TH.Name -> m GHC.TyCon`
  
and used them uniformly where appropriate. This allowed to reduce some code duplication in `compileExpr`

### DIFF
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
@@ -698,11 +698,27 @@ entryExitTracing lamName displayName e ty =
    when boolean coverage is turned on.
 
    The annotation `<expr evaluated to True>` is implemented by adding a `CoverBool location True`
-   coverage annotation with the head function in `expr` as metadata. This means that in an expression
-   like:
-   `foo x < bar y && all isGood xs`
-   We will get annotations for `&&`, `<`, `all`, and `isGood` (given that `isGood` is defined in a module
-   with coverage turned on).
+   coverage annotation with the head function in `expr` as metadata. This means that in an
+   expression like: `foo x < bar y && all isGood xs`
+
+   We will get annotations for `&&`, `<`, `all`, and `isGood` (given that `isGood` is defined in a
+   module with coverage turned on).
+-}
+
+{- Note [GHC.Magic.noinline]
+   For some functions we have two conflicting desires:
+   - We want to have the unfolding available for the plugin.
+   - We don't want the function to *actually* get inlined before the plugin runs, since we rely
+   on being able to see the original function for some reason.
+
+   'INLINABLE' achieves the first, but may cause the function to be inlined too soon.
+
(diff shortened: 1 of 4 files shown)


plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -30,15 +30,14 @@ import GHC.Plugins qualified as GHC
...
-import GHC.Types.TyThing qualified as GHC
 
 #if MIN_VERSION_ghc(9,6,0)
 import GHC.Tc.Utils.TcType qualified as GHC
 #endif
 
 import PlutusTx.Bool qualified
 import PlutusTx.Builtins qualified as Builtins
-import PlutusTx.Builtins.Internal qualified as Builtins
+import PlutusTx.Builtins.Internal qualified as BuiltinsI
```
zliu41, line 40: This is imported as `BI` everywhere else.

plutus-tx-plugin/src/PlutusTx/Compiler/Utils.hs
```diff
@@ -21,14 +22,32 @@ import Language.Haskell.TH.Syntax qualified as TH
...
-    CompileContext{ccNameInfo=names} <- ask
-    case Map.lookup name names of
-        Nothing    -> throwSd CompilationError $ "Missing name:" GHC.<+> (GHC.text $ show name)
-        Just thing -> pure thing
+lookupGhcTyCon :: Compiling uni fun m ann => TH.Name -> m GHC.TyCon
+lookupGhcTyCon thName = do
+  CompileContext { ccNameInfo } <- ask
+  case Map.lookup thName ccNameInfo of
+    Just (GHC.ATyCon tc) -> pure tc
+    _ -> throwPlain $ CompilationError $ "TyCon not found: " <> T.pack (show thName)
```
zliu41, line 32: Use `_` or `Nothing` consistently. Btw these three functions look very similar.

zliu41 (approved):

plutus-tx-plugin/src/PlutusTx/Compiler/Utils.hs
Unisay: In this case `Nothing` couldn't be used because the first pattern is refutable and `Nothing` makes this case non-exhaustive.

Unisay: > Btw these three functions look very similar.

Indeed, I also noticed similarities and considered extracting it, but then after some consideration I decided not to do it because the indirection added seemed not worth reduced repetition to me.

PR #6693, opened 2024-11-19
Re-format prettyprinter-configurable with Fourmolu

Unisay:
- [x] Apply fourmolu to the `prettyprinter-configurable` package.
- [x] Apply `stylish-haskell` on top of it.

Closes https://github.com/IntersectMBO/plutus/issues/6682

### DIFF
--- a/prettyprinter-configurable/src/Text/Fixity/Internal.hs
+++ b/prettyprinter-configurable/src/Text/Fixity/Internal.hs
@@ -1,15 +1,16 @@
 -- editorconfig-checker-disable-file
--- | Precedence-general machinery for deciding whether an expression needs to be wrapped in
--- parentheses or not. Source code has comments on the approach used and how it compares to some
--- other known approaches.
 
-module Text.Fixity.Internal
-    ( Associativity (..)
-    , FixityOver (..)
-    , Direction (..)
-    , RenderContextOver (..)
-    , encloseIn
-    ) where
+{-| Precedence-general machinery for deciding whether an expression needs to be wrapped in
+parentheses or not. Source code has comments on the approach used and how it compares to some
+other known approaches.
+-}
+module Text.Fixity.Internal (
+  Associativity (..),
+  FixityOver (..),
+  Direction (..),
+  RenderContextOver (..),
+  encloseIn,
+) where
 
 {- Note [Approaches to precedence-aware pretty-printing]
 It's not trivial to find papers on precedence-aware pretty-printing.
@@ -205,49 +206,57 @@ Implementing all of that is left as future work.
 
 -- It's not necessary to deal with associativity, see: https://stackoverflow.com/a/43639618
 -- But I find it easier and nicer than changing precedence on the fly.
+
 -- | Associativity of an operator.
 data Associativity
-    = LeftAssociative
-    | RightAssociative
-    | NonAssociative
-    deriving stock (Show, Eq)
+  = LeftAssociative
+  | RightAssociative
+  | NonAssociative
+  deriving stock (Show, Eq)
 
 -- See Note [Approaches to precedence-aware pretty-printing].
--- | Fixity of an operator.
---
--- We allow unary operators to have associativity, because it's useful to distinguish
--- between an expression like @-(-x)@ (unary minus, left-associative) and @~~b@
--- (boolean NOT, right-associative).
---
--- Associativity of unary operators also matters when pretty-printing expressions like @(-x) + y@,
--- which is pretty-printed as @-x + y@, assuming unary minus has the same fixity as @+@ (and both
--- the operators are left-associative). I.e. unary minus is handled just like the binary one:
--- @(0 - x) + y@ is pretty-printed as @0 - x + y@.
---
--- Postfix operators are handled similarly. E.g. if @!@ is left-associative, then @(x!)!@ is
--- pretty-printed as @x!!@ and if it's right-associative -- @(x!)!@.
---
--- The data type is parameterized, so that the user can choose precedence to be integer\/fractional,
--- bounded\/unbounded, etc (we could also allows operators to be partially or totally ordered, but
--- at the moment @prec@ is required to implement 'Ord', i.e. it has to be totally ordered).
--- By default we go with bounded fractional precedence, see the main "Text.Fixity" module.
-data FixityOver prec = Fixity
-    { _fixityAssociativity :: !Associativity
-    , _fixityPrecedence    :: !prec
-    } deriving stock (Show, Eq)
 
--- | Direction in which pretty-printing goes. For example in @x + y@ @x@ is pretty-printed to the
--- left of @+@ and @y@ is pretty-printed to the right of @+@.
+{-| Fixity of an operator.
+
+We allow unary operators to have associativity, because it's useful to distinguish
+between an expression like @-(-x)@ (unary minus, left-associative) and @~~b@
+(boolean NOT, right-associative).
+
+Associativity of unary operators also matters when pretty-printing expressions like @(-x) + y@,
+which is pretty-printed as @-x + y@, assuming unary minus has the same fixity as @+@ (and both
+the operators are left-associative). I.e. unary minus is handled just like the binary one:
+@(0 - x) + y@ is pretty-printed as @0 - x + y@.
+
+Postfix operators are handled similarly. E.g. if @!@ is left-associative, then @(x!)!@ is
+pretty-printed as @x!!@ and if it's right-associative -- @(x!)!@.
+
+The data type is parameterized, so that the user can choose precedence to be integer\/fractional,
+bounded\/unbounded, etc (we could also allows operators to be partially or totally ordered, but
+at the moment @prec@ is required to implement 'Ord', i.e. it has to be totally ordered).
+By default we go with bounded fractional precedence, see the main "Text.Fixity" module.
+-}
+data FixityOver prec = Fixity
+  { _fixityAssociativity :: !Associativity
+  , _fixityPrecedence    :: !prec
+  }
+  deriving stock (Show, Eq)
+
+{-| Direction in which pretty-printing goes. For example in @x + y@ @x@ is pretty-printed to the
+left of @+@ and @y@ is pretty-printed to the right of @+@.
+-}
 data Direction
-    = ToTheLeft
-    | ToTheRight
-    deriving stock (Show, Eq)
+  = ToTheLeft
+  | ToTheRight
+  deriving stock (Show, Eq)
 
 -- | A context that an expression is being rendered in.
 data RenderContextOver prec = RenderContext
-    { _renderContextDirection :: !Direction
-    , _renderContextFixity    :: !(FixityOver prec)
-    } deriving stock (Show, Eq)
+  { _renderContextDirection :: !Direction
+  , _renderContextFixity    :: !(FixityOver prec)
+  }
+  deriving stock (Show, Eq)
+
+{- FOURMOLU_DISABLE -}
 
 -- Instead of receiving a @a -> a@ this function could simply return a 'Bool'.
 -- | Enclose an @a@ (using the provided function) if required or leave it as is.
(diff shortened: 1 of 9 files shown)


prettyprinter-configurable/src/Text/Fixity/Internal.hs
```diff
@@ -205,74 +206,85 @@ Implementing all of that is left as future work.
...
+The need for enclosing is determined from an outer 'RenderContext' and the inner fixity.
+-}
 encloseIn
-    :: Ord prec
-    => (a -> a)                -- ^ Enclose a value of type @a@ in parens.
-    -> RenderContextOver prec  -- ^ An outer context.
-    -> FixityOver prec         -- ^ An inner fixity.
-    -> a
-    -> a
+  :: (Ord prec)
```
effectfully, line 265: Why didn't it move `::` to the previous line?

prettyprinter-configurable/src/Text/Fixity/Internal.hs
```diff
@@ -205,74 +206,85 @@ Implementing all of that is left as future work.
...
-            (LeftAssociative, ToTheLeft)   -> id      -- No need for parens in @Add (Add x y) z@
-                                                      -- which is rendered as @x + y + z@.
-            (RightAssociative, ToTheRight) -> id      -- No need for parens in @Concat xs (Concat ys zs)@
-                                                      -- which is rendered as @xs ++ ys ++ zs@.
-            _                              -> parens  -- Every other case requires parens.
+  case precOut `compare` precInn of
+    LT -> id -- If the outer precedence is lower than the inner, then
+    -- do not add parens. E.g. in @Add x (Mul y z)@ the precedence
+    -- of @Add@ is lower than the one of @Mul@, hence there is
+    -- no need for parens in @x + y * z@.
```
effectfully, line 279: It was nice and readable and now it's just utterly disgusting. Could you please return the old version (modulo indentation) and turn off `fourmolu` in this file? Oh, and please remove the stupid newline from the comment above.

prettyprinter-configurable/src/Text/Fixity/Internal.hs
```diff
@@ -205,74 +206,85 @@ Implementing all of that is left as future work.
...
+    GT -> parens -- If the outer precedence is greater than the inner, then
+    -- do add parens. E.g. in @Mul x (Add y z)@ the precedence
+    -- of @Mul@ is greater than the one of @Add@, hence
+    -- parens are needed in @x * (y + z)@.
+    EQ -> case (assocOut, dir) of -- If precedences are equal, then judge from associativity.
+      _ | assocOut /= assocInn       -> parens -- Associativities differ => parens are needed.
+      (LeftAssociative, ToTheLeft)   -> id -- No need for parens in @Add (Add x y) z@
+      -- which is rendered as @x + y + z@.
```
effectfully, line 287: Just look at this nonsense.

prettyprinter-configurable/src/Text/PrettyBy/Fixity.hs
```diff
@@ -29,112 +29,122 @@ import Lens.Micro
...
 
--- | Instantiate a supplied continuation with a pretty-printer specialized to supplied
--- 'Fixity' and 'Direction'.
+{-| Instantiate a supplied continuation with a pretty-printer specialized to supplied
+'Fixity' and 'Direction'.
+-}
 withPrettyAt
-    :: MonadPrettyContext config env m
-    => Direction -> Fixity -> (AnyToDoc config ann -> m r) -> m r
+  :: (MonadPrettyContext config env m)
```
effectfully, line 100: Is there any option to make it not add the redundant parens?

prettyprinter-configurable/src/Text/PrettyBy/Monad.hs
```diff
@@ -5,40 +5,44 @@
...
 class HasPrettyConfig env config | env -> config where
-    prettyConfig :: Lens' env config
+  prettyConfig :: Lens' env config
 
 -- @env@ is an artefact of the encoding, it shouldn't be necessary as @m@ determines what it is
 -- and we're not interested in reflecting @env@ explicitly (unlike @config@, which is also
 -- determined by @m@ through @env@, but is useful to have explicitly). But GHC does not like it
 -- when @env@ is left implicit, see https://gitlab.haskell.org/ghc/ghc/issues/3490
+
 -- | A constraint for \"@m@ is a monad that allows to pretty-print values in it by a @config@\".
```
effectfully, line 31: Hate. The. Newline.

Like it at least could make it `--\n` instead to connect the comments.

prettyprinter-configurable/src/Text/PrettyBy/Internal.hs
```diff
@@ -574,256 +617,339 @@ application is implemented.
 
 -- | Throw @err@ when @b@ is stuck.
 type family ThrowOnStuck err (b :: Bool) :: Bool where
-    ThrowOnStuck _   'True  = 'True
-    ThrowOnStuck _   'False = 'False
-    ThrowOnStuck err _      = err
+  ThrowOnStuck _ 'True = 'True
+  ThrowOnStuck _ 'False = 'False
+  ThrowOnStuck err _ = err
```
effectfully, line 622: Huh, this won't conflict with `stylish-haskell`?

prettyprinter-configurable/src/Text/PrettyBy/Internal.hs
```diff
@@ -574,256 +617,339 @@ application is implemented.
...
+    TypeError
+      ( 'Text "No ’HasPrettyDefaults’ is specified for " ':<>: 'ShowType config
+          ':$$: 'Text "Either you're trying to derive an instance, in which case you have to use"
+          ':$$: 'Text "  standalone deriving and need to explicitly put a ‘PrettyDefaultBy config’"
+          ':$$: 'Text "  constraint in the instance context for each type in your data type"
+          ':$$: 'Text "  that supports default pretty-printing"
+          ':$$: 'Text "Or you're trying to pretty-print a value of a type supporting default"
+          ':$$: 'Text
+                  "  pretty-printing using a config, \
+                  \for which ‘HasPrettyDefaults’ is not specified."
```
effectfully, line 639: I hate `fourmolu`. This way the error message code was formatted directly reflected how the error message would end up being shown to the user. Now it's just random fucking garbage.

prettyprinter-configurable/src/Text/PrettyBy/Internal.hs
```diff
@@ -70,167 +70,170 @@ import GHC.TypeLits
...
-    -- Defaulting to 'prettyList' would require the user to either provide a 'Pretty' instance for
-    -- each data type implementing 'PrettyBy' or to supply the sensible default implementation as
-    -- 'defaultPrettyFunctorBy' manually and both the options are silly.
-    -- | 'prettyListBy' is used to define the default 'PrettyBy' instance for @[a]@ and @NonEmpty a@.
-    -- In normal circumstances only the 'prettyBy' function is used.
-    -- The default implementation of 'prettyListBy' is in terms of 'defaultPrettyFunctorBy'.
-    prettyListBy :: config -> [a] -> Doc ann
-    default prettyListBy :: config -> [a] -> Doc ann
-    prettyListBy = defaultPrettyFunctorBy
+  -- | Pretty-print a value of type @a@ the way a @config@ specifies it.
```
effectfully, line 108: Stupid tool can't consistently replace `-- |` with `{-|`.

prettyprinter-configurable/src/Text/Fixity/Internal.hs
Unisay: @zliu41 agreed to `leading` if we do it fast :)

prettyprinter-configurable/src/Text/PrettyBy/Monad.hs
Unisay: One is internal comment, other is a public documentation. I want to be able to quickly and visually distinguish them. Newline is a nice way of separating them. Now I am not going to accidentally think this is a single logical description.

prettyprinter-configurable/src/Text/PrettyBy/Fixity.hs
Unisay: Yes:
```yaml
# Whether to put parentheses around a single constraint (choices: auto, always, or never)
single-constraint-parens: always
```

prettyprinter-configurable/src/Text/PrettyBy/Internal.hs
Unisay: Agree.

prettyprinter-configurable/src/Text/PrettyBy/Internal.hs
Unisay: It probably does. My understanding is that we're going to stop using stylish haskell once we have adopted fourmolu.

prettyprinter-configurable/src/Text/PrettyBy/Internal.hs
Unisay: This is not fourmolu, I wrapped this long line (>100) manually.

prettyprinter-configurable/src/Text/Fixity/Internal.hs
```diff
@@ -260,19 +269,21 @@ encloseIn
...
+                  -- parens are needed in @x * (y + z)@.
+    EQ ->
+      case (assocOut, dir) of  -- If precedences are equal, then judge from associativity.
+        _ | assocOut /= assocInn       -> parens  -- Associativities differ => parens are needed.
+        (LeftAssociative, ToTheLeft)   -> id      -- No need for parens in @Add (Add x y) z@
+                                                  -- which is rendered as @x + y + z@.
+        (RightAssociative, ToTheRight) -> id      -- No need for parens in
+                                                  -- @Concat xs (Concat ys zs)@
+                                                  -- which is rendered as @xs ++ ys ++ zs@.
+        _                              -> parens  -- Every other case requires parens.
```
Unisay, line 289: Fourmolu disabled. 
The trade-off here: current formatting isn't diff-friendly (comment blocks shift horizontally depending on the lhs expression length) but is fitted to this particular code width thus making it more compact.

My preference is in making it less fragile by placing comments above the expressions with the same indentation level, such that it doesn't depend on the expression length:

```suggestion
  case precOut `compare` precInn of
    LT -> 
      -- If the outer precedence is lower than the inner, then
      -- do not add parens. E.g. in @Add x (Mul y z)@ the precedence
      -- of @Add@ is lower than the one of @Mul@, hence there is
      -- no need for parens in @x + y * z@.
      id
    GT ->
      -- If the outer precedence is greater than the inner, then
      -- do add parens. E.g. in @Mul x (Add y z)@ the precedence
      -- of @Mul@ is greater than the one of @Add@, hence
      -- parens are needed in @x * (y + z)@.
      parens
    EQ ->
      -- If precedences are equal, then judge from associativity.
      case (assocOut, dir) of 
        _ | assocOut /= assocInn ->
          -- Associativities differ => parens are needed.
          parens 
        (LeftAssociative, ToTheLeft) ->
          -- No need for parens in @Add (Add x y) z@
          -- which is rendered as @x + y + z@.
          id 
        (RightAssociative, ToTheRight) -> 
          -- No need for parens in
          -- @Concat xs (Concat ys zs)@
          -- which is rendered as @xs ++ ys ++ zs@.
          id
        _ ->
          -- Every other case requires parens.
          parens 
```

Such code trades vertical space (which I have plenty of) for precious horizontal space (which I like to save).

effectfully:
> How about I instead resolve https://github.com/IntersectMBO/plutus/issues/4921 and we won't need to do anything here? This package hasn't been updated for a while anyway, so it's pretty stable.

So I did that instead: #6699. If that PR is merged, we'll close this one.

zliu41 (commented):
Looks tolerable.

Besides leading arrows, what are the other non-default config options this uses?

effectfully:
Closing in favor of #6699.

PR #6694, opened 2024-11-19
Move some pragmas to below functions

kwxm:
With luck, this moves all of the `OPAQUE`, `INLINE`, `INLINABLE`, and `INLINEABLE` pragmas to below the things that they apply to (for the benefit of fourmolu).  I did this using find together with an awk script that says

```
BEGIN {n=1; P="^{-# (OPAQUE|INLINE|INLINABLE|INLINEABLE)" }
$0 ~ P {pragma = $0; while ((n=getline) && NF>0) print; print pragma}
$0 !~ P {if (n) print}
```
[Later: it turns out that if there's a big comment just before the start of a function then that code can sometimes move a pragma inside it, so it probably needs to be improved.]

A few occurrences had to be fixed manually where , eg, `INLINE`  appeared in the middle of a function (I'm not totally sure what to do in this case) or where there were empty lines before or after it (not sure if I got all of these).  

The pragmas mentioned above seem to be the only ones that we use that apply to blocks of code in the body of a file.  I've ignored things like `HLINT`, `ANN`, `MINIMAL`, `COMPLETE` and `RULES`.  I don't guarantee that this has dealt with every single thing that we might care about.

### DIFF
--- a/doc/docusaurus/static/code/BasicPolicies.hs
+++ b/doc/docusaurus/static/code/BasicPolicies.hs
@@ -36,6 +36,7 @@ currencyValueOf :: Value -> CurrencySymbol -> Value
 currencyValueOf (Value m) c = case Map.lookup c m of
     Nothing -> mempty
     Just t  -> Value (Map.singleton c t)
+{-# INLINABLE currencyValueOf #-}
 -- BLOCK2
 -- The 'plutus-ledger' package from 'plutus-apps' provides helper functions to automate
 -- some of this boilerplate.
@@ -43,7 +44,6 @@ oneAtATimePolicyUntyped :: BuiltinData -> BuiltinData -> BuiltinUnit
 -- 'check' fails with 'error' if the argument is not 'True'.
 oneAtATimePolicyUntyped r c =
     check $ oneAtATimePolicy (unsafeFromBuiltinData r) (unsafeFromBuiltinData c)
-{-# INLINABLE currencyValueOf #-}
 
 -- We can use 'compile' to turn a minting policy into a compiled Plutus Core program,
 -- just as for validator scripts.
--- a/doc/docusaurus/static/code/BasicPolicies.hs
+++ b/doc/docusaurus/static/code/BasicPolicies.hs
@@ -31,7 +31,6 @@ oneAtATimePolicy _ ctx =
     -- will assume we've got from elsewhere for now.
     in currencyValueOf minted ownSymbol == singleton ownSymbol tname 1
 
-{-# INLINABLE currencyValueOf #-}
 -- | Get the quantities of just the given 'CurrencySymbol' in the 'Value'.
 currencyValueOf :: Value -> CurrencySymbol -> Value
 currencyValueOf (Value m) c = case Map.lookup c m of
@@ -44,6 +43,7 @@ oneAtATimePolicyUntyped :: BuiltinData -> BuiltinData -> BuiltinUnit
 -- 'check' fails with 'error' if the argument is not 'True'.
 oneAtATimePolicyUntyped r c =
     check $ oneAtATimePolicy (unsafeFromBuiltinData r) (unsafeFromBuiltinData c)
+{-# INLINABLE currencyValueOf #-}
 
 -- We can use 'compile' to turn a minting policy into a compiled Plutus Core program,
 -- just as for validator scripts.
--- a/plutus-benchmark/nofib/src/PlutusBenchmark/NoFib/Clausify.hs
+++ b/plutus-benchmark/nofib/src/PlutusBenchmark/NoFib/Clausify.hs
@@ -126,40 +126,46 @@ formula1 = Eqv (Eqv (Sym 1) (Sym 1))
                     (Eqv (Sym 1) (Sym 1)))
 {-# INLINABLE formula1 #-}
 
+-- % One execution takes about 0.35s and 300 MB
 formula2 :: Formula  -- (a = a = a) = (a = a = a)
 formula2 = Eqv (Eqv (Sym 1) (Eqv (Sym 1) (Sym 1)))
                (Eqv (Sym 1) (Eqv (Sym 1) (Sym 1)))
-{-# INLINABLE formula2 #-} -- % One execution takes about 0.35s and 300 MB
+{-# INLINABLE formula2 #-}
 
+-- % One execution takes about 1.5s and 660 MB
 formula3 :: Formula  -- (a = a = a) = (a = a) = (a = a)
 formula3 = Eqv (Eqv (Sym 1) (Eqv (Sym 1) (Sym 1)))
                (Eqv (Eqv (Sym 1) (Sym 1))
                     (Eqv (Sym 1) (Sym 1)))
-{-# INLINABLE formula3 #-}  -- % One execution takes about 1.5s and 660 MB
+{-# INLINABLE formula3 #-}
 
+-- % One execution takes about 2s and 1 GB
 formula4 :: Formula  -- (a = b = c) = (d = e) = (f = g)
 formula4 = Eqv (Eqv (Sym 1) (Eqv (Sym 2) (Sym 3)))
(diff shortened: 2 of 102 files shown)


plutus-tx-plugin/test/Plugin/Coverage/Spec.hs
```diff
@@ -36,11 +36,11 @@ boolTrueFalse = plc (Proxy @"boolTrueFalse") (\() -> True && False)
 boolOtherFunction :: CompiledCode (Maybe Integer -> Maybe Bool)
 boolOtherFunction = plc (Proxy @"boolOtherFunction") fun
 
-{-# INLINEABLE fun #-}
 fun :: Maybe Integer -> Maybe Bool
 fun x = case x of
   Just y | otherFun y -> Just False
   _                   -> Nothing
+{-# INLINEABLE fun #-}
```
kwxm, line 43: This change causes some of the tests to fail.  I suspect that the numbers in `coverage` need to be changed but I'm not quite sure what it does so I'll take a closer look tomorrow.

plutus-tx-plugin/test/Plugin/Coverage/9.6/coverageCode.pir.golden
kwxm, line 1: See `Spec.hs`: moving the pragma changed some line numbers.

doc/docusaurus/static/code/BasicPolicies.hs
```diff
@@ -44,6 +43,7 @@ oneAtATimePolicyUntyped :: BuiltinData -> BuiltinData -> BuiltinUnit
 -- 'check' fails with 'error' if the argument is not 'True'.
 oneAtATimePolicyUntyped r c =
     check $ oneAtATimePolicy (unsafeFromBuiltinData r) (unsafeFromBuiltinData c)
+{-# INLINABLE currencyValueOf #-}
```
Unisay, line 46: This is too far, the blank line between functions was missing.

plutus-benchmark/nofib/src/PlutusBenchmark/NoFib/Clausify.hs
```diff
@@ -63,108 +62,108 @@ disin (Dis p q) =
...
 formula1 = Eqv (Eqv (Sym 1) (Sym 1))
                (Eqv (Eqv (Sym 1) (Sym 1))
                     (Eqv (Sym 1) (Sym 1)))
+{-# INLINABLE formula1 #-}
 
-{-# INLINABLE formula2 #-} -- % One execution takes about 0.35s and 300 MB
 formula2 :: Formula  -- (a = a = a) = (a = a = a)
 formula2 = Eqv (Eqv (Sym 1) (Eqv (Sym 1) (Sym 1)))
                (Eqv (Sym 1) (Eqv (Sym 1) (Sym 1)))
+{-# INLINABLE formula2 #-} -- % One execution takes about 0.35s and 300 MB
```
Unisay, line 132: Its peculiar: the comment moved together with the pragma. I don't think this is what we want.

plutus-benchmark/nofib/src/PlutusBenchmark/NoFib/Clausify.hs
```diff
@@ -63,108 +62,108 @@ disin (Dis p q) =
...
 
-{-# INLINABLE formula3 #-}  -- % One execution takes about 1.5s and 660 MB
 formula3 :: Formula  -- (a = a = a) = (a = a) = (a = a)
 formula3 = Eqv (Eqv (Sym 1) (Eqv (Sym 1) (Sym 1)))
                (Eqv (Eqv (Sym 1) (Sym 1))
                     (Eqv (Sym 1) (Sym 1)))
+{-# INLINABLE formula3 #-}  -- % One execution takes about 1.5s and 660 MB
```
Unisay, line 138: ditto

plutus-benchmark/nofib/src/PlutusBenchmark/NoFib/Knights/Queue.hs
```diff
@@ -9,57 +9,55 @@ import PlutusTx.Prelude as Tx
...
 emptyQueue [] = True
 emptyQueue _  = False
-
+{-# INLINABLE emptyQueue #-}
 
 {-
-{-# INLINABLE sizeQueue #-}
 sizeQueue :: Queue b -> Integer
 sizeQueue xs = length' xs
+{-# INLINABLE sizeQueue #-}
```
Unisay, line 62: 👍🏼

plutus-benchmark/nofib/src/PlutusBenchmark/NoFib/LastPiece.hs
```diff
@@ -111,61 +110,61 @@ prune ss =
...
 next :: Square -> Square
 next (row,col) = (row,col+1)
+{-# INLINABLE next #-}
 
-{-# INLINABLE maxRow #-}
 {-# INLINABLE maxCol #-}
 maxRow,maxCol :: Integer
 maxRow = 8
 maxCol = 8
+{-# INLINABLE maxRow #-}
```
Unisay, line 148: Two annotations, one moved 🐛

plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs
```diff
@@ -363,6 +362,7 @@ valueOf value cur tn =
     case Map.lookup tn tokens of
             Nothing -> 0
             Just v  -> v
+{-# INLINABLE valueOf #-}
 
 {-# INLINEABLE withCurrencySymbol #-}
```
Unisay, line 367: This one should be moved

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -355,6 +354,7 @@ valueOf value cur tn =
     case Map.lookup tn tokens of
       Nothing -> 0
       Just v  -> v
+{-# INLINABLE valueOf #-}
 
 {-# INLINEABLE withCurrencySymbol #-}
```
Unisay, line 359: ⚡ should be moved down.

Unisay: I think that script doesn't take into account alternative spelling INLINEABLE

kwxm: > I think that script doesn't take into account alternative spelling INLINEABLE

It does, but I think the problem was that in this file (and sometimes in other places) there was a blank line between the pragma and the thing it applied to, and I didn't consider that case.

plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs
```diff
@@ -232,18 +232,18 @@ newtype AssetClass = AssetClass {unAssetClass :: (CurrencySymbol, TokenName)}
   deriving (Pretty) via (PrettyShow (CurrencySymbol, TokenName))
 
 instance HasBlueprintSchema AssetClass referencedTypes where
-  {-# INLINEABLE schema #-}
+  {-# INLINABLE schema #-}
```
Unisay, line 235: FYI: `fourmolu` also does this kind of a normalisation.

kwxm: > FYI: fourmolu also does this kind of a normalisation.

Good.  I don't have strong opinions about how it should be spelled, but I noticed that in this file both forms were used, and I felt compelled to make them all the same.

Unisay (approved):
Thanks @kwxm !
Its pretty cool 😎 that you were able to automate all this 👍🏼

kwxm:
> Thanks @kwxm ! Its pretty cool 😎 that you were able to automate all this 👍🏼

Well I managed about 90% automatically but it turned out that there were lots of edge-cases that I hadn't thought of.

Issue #6608, opened 2024-10-29
[Epic] Tests improvements


effectfully:
This issue is for dumping all the plans regarding tests improvements in a largely unstructured way.

effectfully:
- [x] `discardIfAnyConstant (not . isSerialisable) $ Generators.genProgram @DefaultFun` potentially wastes a lot of time discarding generated programs. `discardIfAnyConstant` is used in quite a few places
- [ ] `Evaluation.Builtins.Definition` should demand that each built-in function is unit-tested
- [ ] remove `ignoreTest` from Plutus IR tests by fixing the tests
- [ ] increase the number of test runs in Plutus IR from no to some (i.e. fix `withMaxSuccess 0`) for failing shrinking tests by figuring out why they are taking too long and fixing that
- [ ] can we somehow ban running 100 tests altogether? So that the number of tests to run is always explicitly specified

PR #6695, opened 2024-11-20
[Test] Disable 'prop_shrinkTermSound'

effectfully:


kwxm:
Ironically this PR now appears to be [failing](https://github.com/IntersectMBO/plutus/pull/6695/checks?check_run_id=33274541452) on `noTermShrinkLoops`.  I also got a very long run in `stats_numShrink`.

kwxm (approved):
This may still not be enough, but let's try it and see.  We should maybe make a issue to find out why this started happening and try to fix it properly if possible.

effectfully:
> I also got a very long run in `stats_numShrink`.

Disabled it as well.

> We should maybe make a issue to find out why this started happening and try to fix it properly if possible.

Added an entry to #6608.

PR #6696, opened 2024-11-20
[Experiment] Check how much scope checking costs us

effectfully:
This experiment will tell us exactly how much of the preparation stage is wasted on the useless scope check.


effectfully:
So scope checking costs us about 3% of total time for the `validation-full` benchmarks and 20% of preparation time for the `validation-decode` benchmarks.

PR #6697, opened 2024-11-20
[Experiment] Check how much no-op `mkTermToEvaluate` would cost us

effectfully:
This is just to ensure that the numbers computed in #6696 aren't skewed by something I failed to think of.


effectfully:
what lol

effectfully:
Has to be noise.

effectfully:
OK, yes, everything is as expected.

PR #6700, opened 2024-11-21
Fully Data-backed V3.ScriptContext

ana-pantilie:
Part of https://github.com/IntersectMBO/plutus-private/issues/58

This PR adds `asData` to the Plutus V3 version of the `ScriptContext` and all of its underlying types.

Judging by the code in `plutus-benchmark/script-contexts/test/Data/9.6/checkScriptContext1.pir.golden`, the only remaining SOPs are `Maybe`, `Bool` and `Solo` (where does this Solo type come from?).

Anyway, the generated code is _much much_ smaller and all our `ScriptContext` benchmarks have significantly improved performance :tada:. See also`checkScriptContext2.size.golden` which now is _tiny_ (this script just decodes the script context and does nothing with it).

**Question**: Should we add more benchmarks?

I'll open a separate PR for the V1 and V2 ScriptContexts once this one is merged.


plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Credential.hs
```diff
@@ -0,0 +1,105 @@
+-- editorconfig-checker-disable-file
```
Unisay, line 1: Would you be able to remove this disabler and also format new files with fourmolu?

plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Address.hs
```diff
@@ -0,0 +1,105 @@
...
+{-# INLINEABLE stakingCredential #-}
+
+-- | The staking credential of an address (if any)
+stakingCredential :: Address -> Maybe StakingCredential
+stakingCredential (Address _ s) = s
+
+----------------------------------------------------------------------------------------------------
+-- TH Splices --------------------------------------------------------------------------------------
+
+$(PlutusTx.makeLift ''Address)
```
Unisay, line 105: FYI @zliu41 prefers his TH splices served without the `$(...)` ceremony :) He has removed it in other places.

plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Credential.hs
```diff
@@ -0,0 +1,105 @@
...
+import PlutusTx.Eq qualified as PlutusTx
+import PlutusTx.Show (deriveShow)
+import Prettyprinter (Pretty (..), (<+>))
+
+-- | Credentials required to unlock a transaction output.
+PlutusTx.asData
+    [d|
+        data Credential
+          =
+            -- | The transaction that spends this output must be signed by the private key.
```
Unisay, line 43: I don't think that haddock syntax is useful inside the `[d|  |]`

plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Interval.hs
```diff
@@ -0,0 +1,582 @@
...
+{-# LANGUAGE TypeApplications     #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns         #-}
+
+{-# OPTIONS_GHC -fno-omit-interface-pragmas #-}
+{-# OPTIONS_GHC -fno-specialise #-}
+{-# OPTIONS_GHC -fno-ignore-interface-pragmas #-}
+{-# LANGUAGE DataKinds            #-}
+{-# LANGUAGE TypeFamilies         #-}
+{-# LANGUAGE TypeOperators        #-}
```
Unisay, line 15: Move to the section above or remove the gap?

zliu41 (approved):
Looks good!

> where does this Solo type come from?

My guess is it's how GHC compiles pattern synonyms. It might use some kind of generic approach involving tuples, and `Solo` is a 1-tuple.

> Should we add more benchmarks?

We should create new variants of the constitution script and the marlowe scripts to use the new ScriptContext. The constitution script is hand-optimized so we should target for a comparable (if not the same) performance. The marlowe scripts uses SOP ScriptContext so there's likely room for optimization.

PR #6702, opened 2024-11-21
[Evaluation] [Names] Define all lookups in terms of 'contIndexZero'

effectfully:
Instead of 4 different implementation this will make it 1 main and 3 derivative ones.

This appears to make the `validation` benchmarks faster by a percent and the `nofib` benchmarks faster by a couple of percent while simplifying the code, so a clear win.

Please review this one VERY carefully.


effectfully:
OK, this appears to make `validation` benchmarks faster by a percent and `nofib` benchmarks faster by a couple of percent while simplifying the code, so a clear win.

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
```diff
@@ -74,122 +76,94 @@ uncons = \case
...
     uncons = uncons
-    {-# INLINABLE length #-}
-    length Nil             = 0
-    length (BHead sz _ tl) = sz + RAL.length tl
-    {-# INLINABLE indexZero #-}
+    {-# INLINE uncons #-}
+
+    length = go 0 where
+        go acc Nil             = acc
+        go acc (BHead sz _ tl) = go (acc + sz) tl
```
effectfully: I don't know if it's important to make this one fast, but why not while we're here. It's more readable too, in my opinion.

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
```diff
@@ -74,122 +76,94 @@ uncons = \case
...
+safeIndexOne = contIndexOne Nothing Just
+{-# INLINE safeIndexOne #-}
 
 instance RAL.RandomAccessList (RAList a) where
     type Element (RAList a) = a
 
-    {-# INLINABLE empty #-}
     empty = Nil
-    {-# INLINABLE cons #-}
+    {-# INLINE empty #-}
```
effectfully, line 146: We really want all of that to inline, no point is chickening out and saying `INLINABLE`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -878,7 +878,7 @@ enterComputeCek = computeCek
     -- | Look up a variable name in the environment.
     lookupVarName :: NamedDeBruijn -> CekValEnv uni fun ann -> CekM uni fun s (CekValue uni fun ann)
     lookupVarName varName@(NamedDeBruijn _ varIx) varEnv =
-        Env.safeIndexOneCont
+        Env.contIndexOne
```
effectfully, line 881: This names better matches the other names in the module, should've gone with it originally.

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
```diff
@@ -74,122 +76,94 @@ uncons = \case
...
+{-# INLINE unsafeIndexZero #-}
+
+-- 0-based
+safeIndexZero :: RAList a -> Word64 -> Maybe a
+safeIndexZero = contIndexZero Nothing Just
+{-# INLINE safeIndexZero #-}
+
+-- 1-based
+unsafeIndexOne :: RAList a -> Word64 -> a
+unsafeIndexOne = contIndexOne (error "out of bounds") id
```
bezirg, line 134: here the error changes from "index zero" to "out of bounds", but it does not matter.

bezirg (approved):
LGTM

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
```diff
@@ -63,6 +64,7 @@ cons :: a -> RAList a -> RAList a
 cons x = \case
     (BHead w1 t1 (BHead w2 t2 ts')) | w1 == w2 -> BHead (2*w1+1) (Node x t1 t2) ts'
```
bezirg, line 65: Do you think using `unsafeShiftL` in place of `2*w1` multiplication is worth it? I have a feeling that GHC would already apply such transformation

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
```diff
@@ -74,122 +76,94 @@ uncons = \case
...
     indexTree treeSize offset (Node _ t1 t2) =
         let halfSize = unsafeShiftR treeSize 1 -- probably faster than `div w 2`
-            offset' = offset - 1
-        in if offset' <= halfSize
-           then indexTree halfSize offset' t1
-           else indexTree halfSize (offset' - halfSize) t2
-{-# INLINE safeIndexOneCont #-}
+        in if offset <= halfSize
+           then indexTree halfSize (offset - 1) t1
+           else indexTree halfSize (offset - 1 - halfSize) t2
```
bezirg, line 114: Out of curiosity: did you inline `offset'` here because it is such a trivial computation?

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
effectfully: IIRC GHC isn't actually smart enough to do that, but I may be lying to you. But it's certainly best not to rely on GHC when it's easy.

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
effectfully: It was inlined in the original `safeIndexZero` and not inlined in the original `safeIndexOne`, I think because in the latter `offset'` was used twice for every branch, while in the former only once. So I followed and picked the non-inlined version. It doesn't really matter.

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
effectfully: Yeah, I thought so too.

kwxm (approved):
This looks fine: even just expressing `safeIndexOne` in in terms of `safeIndexZero` instead of having a separate implementation is a huge improvement.  I didn't review the PR quite as carefully as I wanted to though: I haven't understood the workings of the data structure well enough to convince myself that the implementation is correct, and there's a lot of scope for things being off by one.  Maybe it'd be a good exercise to formalise this in Agda. 

------------

I'm wondering about the other implementations of `RandomAccessList` here.  I don't remember the details, but I think we probably tried a number of different implementations in order to see which was the best.  I ran `index-envs-bench` and looked at the html report and it looks as if the different versions win and lose in different places.  In particular `SkewBinarySlab` looks as if it performs quite similarly.  It mght be worth (a) refactoring `SkewBinarySlab` in a smilar way to `SkewBinary` (and having similar implementations might help with maintainability), and (b) running all of the benchmarks with CEK machines modified to use each of the different implementations in order to check that `SekwBinary` is still the best thing to use.

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
```diff
@@ -74,122 +76,94 @@ uncons = \case
...
+in the GHC Core, allowing GHC to inline the arguments of 'contIndexZero' and transform the whole
 thing into a beautiful recursive join point full of @Word64#@s, i.e. allocating very little if
 anything at all.
 -}
 
--- See Note [Optimizations of safeIndexOneCont].
-safeIndexOneCont :: forall a b. b -> (a -> b) -> RAList a -> Word64 -> b
-safeIndexOneCont z f = findTree where
+-- See Note [Optimizations of contIndexZero].
+contIndexZero :: forall a b. b -> (a -> b) -> RAList a -> Word64 -> b
```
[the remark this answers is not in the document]
effectfully: I think it's a fairly standard abbreviation, particular given that the continuation monad is called [`Cont`](https://hackage.haskell.org/package/mtl-2.3.1/docs/Control-Monad-Cont.html#t:Cont).

bezirg:
There is risk of underflow with this approach. But i cannot see how this can become a problem on the chain. Worth perhaps to put a note about it?

effectfully:
> There is risk of underflow with this approach. 

How come?

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
```diff
@@ -74,122 +76,94 @@ uncons = \case
...
-    indexTree _ 0 _ = z -- "index zero"
     indexTree _ _ (Leaf _) = z
     indexTree treeSize offset (Node _ t1 t2) =
         let halfSize = unsafeShiftR treeSize 1 -- probably faster than `div w 2`
-            offset' = offset - 1
-        in if offset' <= halfSize
-           then indexTree halfSize offset' t1
-           else indexTree halfSize (offset' - halfSize) t2
-{-# INLINE safeIndexOneCont #-}
+        in if offset <= halfSize
```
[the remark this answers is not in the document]
effectfully: Yeah, that's unfortunate, but I probably can't do anything about it.

plutus-core/index-envs/src/Data/RandomAccessList/SkewBinary.hs
effectfully: Just checked, yes, GHC doesn't optimize that one:

```
plusWord64# (timesWord64# 2#Word64 bx_danZ) 1#Word64
```

Maybe `timeWord64#` somehow does, but I kinda doubt.

@kwxm I'll try your version out.

bezirg:
> > There is risk of underflow with this approach. 
> 
> How come?

My bad, there is no underflow I missed the clause at L118 . Actually that clause is redundant if u assume that scopecheck pass is in place. But i guess it is better not to assume that.

effectfully:
> Actually that clause is redundant if u assume that scopecheck pass is in place.

I'm not making that assumption. We could make the `Flat` instance fail upon encountering a `0` index if it doesn't already, but I'd rather pay this tiny constant cost per lookup and not bother.

effectfully:
OK, something's wrong with the benchmarking machine, so I'll merge this PR without the latest performance improvements (because they aren't benchmarked yet) and create another PR for those.

PR #6705, opened 2024-11-22
[Evaluation] [Performance] Strictify 'spend'

effectfully:
I used to see this:

```haskell
              $j_sXeb (x#_aWMv :: Int#) (wild1_aWMu :: Int)
                = let! { I# ipv6_sXec ~ wild3_X8 <- wild1_aWMu } in
```

and now I see this:

```haskell
              $w$j_sum7 :: Int# -> Int# -> (# State# s_sumh, () #)
              $w$j_sum7 (x#_sum1 :: Int#) (ww2_sum4 :: Int#)
```

in the GHC Core of the `spend` local function in `UntypedPlutusCore.Evaluation.Machine.Cek.ExBudgetMode`. It's a tiny but welcome improvement.

### DIFF
--- a/plutus-core/changelog.d/20241126_024818_effectfully_strictify_spend.md
+++ b/plutus-core/changelog.d/20241126_024818_effectfully_strictify_spend.md
@@ -1,0 +1,3 @@
+### Changed
+
+- In #6705 made the local `spend` function faster increasing overall performance of the evaluator by 1.8%.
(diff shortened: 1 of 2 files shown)


effectfully:
OK, these ones are great across all the runs:

image

So this is definitely an improvement and an extremely cheap one. Let's call it -1.8% on average (we got results from -1.2% to -2.6% with most being around -1.8%).

Unisay (approved):

PR #6706, opened 2024-11-25
[Evaluation] [Performance] Ditch 'NamedDeBruijn' in favor of 'DeBruijn'

effectfully:
Let's see how much these nasty name conversions cost us.

### DIFF
--- a/plutus-benchmark/common/PlutusBenchmark/Common.hs
+++ b/plutus-benchmark/common/PlutusBenchmark/Common.hs
@@ -78,7 +78,7 @@ getConfig limit = do
               }
 
 -- | Evaluate a script and return the CPU and memory costs (according to the cost model)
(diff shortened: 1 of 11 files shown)


kwxm:
Those results look pretty conclusive!

effectfully:
OK, I'm a moron, `validation` benchmarks don't include the preparation costs. It's still wild that we get a slowdown here, as if `toNamedDeBruijn` was forcing something important that somehow isn't forced by `force`. Weird.

effectfully:
WTF is going on in this PR

image

effectfully:
> WTF is going on in this PR

And it is consistent, here's the second output:

Screenshot from 2024-11-28 10-46-51

and the third one:

Screenshot from 2024-11-28 10-47-18

It appears that this PR somehow slows down the evaluator, while speeding up preparation, hence why we see results all over the place. How can it possibly slow down the evaluator? I have no idea, but this is likely me screwing something up, because I can't think of a way to make things slower by removing redundant code.

PR #6707, opened 2024-11-25
[Evaluation] [Performance] Remove 'unsafeFreezePrimArray' from 'itrav…

effectfully:
`unsafeFreezePrimArray` seems entirely unnecessary there. Let's check if this is the case and if we can save any work by removing it. Quite possible it's the opposite and we'll get a slowdown here.


effectfully:
Meh, not worth working on it I believe.

PR #6709, opened 2024-11-26
instance IsString BuiltinByteStringUtf8

Unisay:
Closes https://github.com/IntersectMBO/plutus/issues/6671
Closes https://github.com/IntersectMBO/plutus/issues/1137

Supersedes https://github.com/IntersectMBO/plutus/pull/6701


plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -138,39 +140,70 @@ compileLiteral = \case
 -- do different things to the inner expression. This one assumes it's a literal, the other one keeps compiling
 -- through it.
 
-data StringExprContentAs = AsBytes | AsText
+data EncodingAction = UndoUtf8 | PreserveUtf8
```
zliu41, line 143: I'd still like a better name than UndoUtf8. The name should help answer the following question: Given a string in the source code, say "Ёжище", _what kind of ByteString does it produce_? If it does truncating (like Haskell's ByteString), call it `Truncate`, if it does something else, say something else. If the name itself is not enough, add Haddock.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -138,39 +140,70 @@ compileLiteral = \case
...
   -- Chase variable references! GHC likes to lift string constants to variables,
   -- that is not good for us!
   GHC.Var (GHC.maybeUnfoldingTemplate . GHC.realIdUnfolding -> Just unfolding) ->
-    stringExprContent contentAs unfolding
+    stringLiteralBytes encodingAction unfolding
   _ -> Nothing
 
+-- | Given a GHC Core expression representing a string literal
+-- extracts a ByteString from it.
+stringLiteralAsBytes
```
zliu41, line 178: Please think of better names for `stringLiteralBytes` and `stringLiteralAsBytes`. Their main difference is that one returns `Maybe` and the other throws an error. The presence or absence of "as" does not convey such information whatsoever.

plutus-tx-plugin/changelog.d/20241121_161607_Yuriy.Lazaryev_IsString_instances.md
```diff
@@ -0,0 +1,8 @@
+### Added
+
+- `instance IsString BuiltinByteStringUtf8` allows using string literals to construct UTF8-encoded `BuiltinByteString` values.
```
zliu41, line 3: This should also mention `BuiltinByteStringHex`.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -754,58 +790,66 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
+            -- because it won't succeed anyway:
+            -- 'fromString' function contains 'Data.Char' type in its definition
+            -- and plugin can't compile it.
+            _ -> throwUnsupported
 
     -- 'stringToBuiltinByteString' invocation
-    (strip -> GHC.Var n) `GHC.App` (strip -> stringExprContent AsBytes -> Just bs)
-      | GHC.getName n == sbbsName ->
-          pure $ PIR.Constant annMayInline $ PLC.someValue bs
+    (strip -> GHC.Var n) `GHC.App` (strip -> arg)
```
zliu41, line 837: You also need to intercept `stringToBuiltinByteStringHex`.

zliu41 (changes requested):
much better!

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
Unisay: When I conceived this data type I meant it to answer one question: 
**Should the function undo the UTF8 encoding already applied to the byte string or not?**
From this perspective I could have named it `Utf8PreservationBehavour`.  The idea is that if provides 1 bit of information **currently required** by the `stringLiteralBytes` function (I had a first impulse to use `Bool` but this would have been a [boolean blindness](https://luzkan.github.io/smells/boolean-blindness)) This way I am avoiding the premature generalisation. We can gradually make this data type more general as soon as such need arrives (if ever).

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
Unisay: Ok, you want to highlight this difference in behaviour by reflecting it in the name. 
How do you like 
* `tryStringLiteralBytes`  <- maybe-returning 
* `stringLiteralBytes` <- throwing
?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
Unisay: Ups, good catch, totally missed it.

Unisay:
> By the way, there needs to be comments on `fromUtf8` explaining why you implemented it yourself.

I gave `TE.encodeUtf8` another try and it worked. I must have been using it in a wrong way before. It could be due to the difference between `Data.ByteString.unpack` and `Data.ByteString.Char8.unpack`. Anyway - the latter has worked (tests are green) and I have removed the custom decoding 🥳

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
zliu41: I don't know what "undo the UTF8 encoding" means. For example, given `"\12345" :: String`, its UTF-8 encoding is `"\227\128\185"`. What's the output of "undo the UTF8 encoding"?

Unisay: "Undo the UTF8 encoding" is an approximation of the more intricate meaning, which is "Recover a 0 .. 255 **subset of codepoints**  after GHC has UTF8 encoded them"

This function is not defined (hence `Maybe` on the RHS) on the Unicode strings that didn't originally contains every code-point in the range 0-255.

I intent to document this function better. Maybe I should rename it to `fromUtf8EncodedBytes` ?

zliu41: So if I understand correctly, what it does is
- if the string contains only 0-255 codepoints, convert each char to a byte
- fail if it contains anything 256 or above

Then you can call it any of these: `SingleBytes`, `BytesOnly`, `SingleBytesOnly`, `ExtendedAscii`, all of which are better and more informative than `UndoUtf8`.

Also

> This function is not defined (hence Maybe on the RHS) on the Unicode strings that didn't originally contain every code-point in the range 0-255.

This needs to be put into a comment near the definition of this type (or you can add a Note so that you can refer to it in other places - you probably want to refer to it in the function that returns `Maybe`).

Unisay: The best name is the one you don't have to invent :) So I have refactored code not to require that parameter.

zliu41 (approved):
:+1:

Unisay:
@zeme-wana the PR merge is blocked by the failing `plutus-template` build. This is expected, as PR introduces backward-incompatible changes. I know how to fix it after this PR is merged.

PR #6712, opened 2024-11-27
Accept updated golden files from agda-conformance tests

Unisay:
```sh
cabal test agda-conformance --test-options '--accept'
```


Unisay:
The tests pass even without `--accept`ing new golden files. Its a bug in the `--accept` functionality.

PR #6720, opened 2024-11-28
Kwxm/conformance/fix data syntax

kwxm:
Many of the conformance tests involving `Data` used obsolete syntax and were failing with parse errors (which were in golden files, perhaps due to over-enthusiastic use of `--accept`).  We must have forgotten to fix these when we introduced the final form of the concrete syntax: this fixes that.

I also corrected a test name that didn't match its directory and added a few extra tests for the concrete syntax of other types.


plutus-conformance/test-cases/uplc/evaluation/builtin/constant/bool/Maybe/Maybe.uplc.budget.expected
kwxm, line 1: I think GitHub is confused about what's happened here, but it's harmless.

effectfully (approved):

PR #6727, opened 2024-12-02
CIP-0138 Builtin Array

Unisay:
Closes https://github.com/IntersectMBO/plutus/issues/6717

- [x] Built-in Array type itself, implemented by the `Data.Vector.Strict`
- [x] Builtin funtion `LengthArray`
- [x] Builtin funtion `ListToArray`
- [x] Tests

The PR is ready for review but build is not working until we update flake inputs (need vector 0.13.2)


plutus-core/cost-model/data/builtinCostModelA.json
```diff
@@ -698,6 +698,26 @@
             "type": "constant_cost"
         }
     },
+    "lengthArray" : {
+        "cpu": {
+            "arguments": 99999999999999,
```
Unisay, line 703: Temporary value (stub)

plutus-core/plutus-core.cabal
```diff
@@ -341,7 +342,7 @@ library
     , time
     , transformers
     , unordered-containers
-    , vector
+    , vector                      ^>=0.13.2
```
Unisay, line 345: `Data.Vector.Strict`  introduced in this version of the library.

plutus-core/plutus-core/src/PlutusCore/Crypto/BLS12_381/Pairing.hs
```diff
@@ -51,7 +51,7 @@ instance NFData MlResult where
     rnf _ = ()
 
 instance Hashable MlResult where
-    hashWithSalt salt = const salt
+    hashWithSalt salt _MlResult = salt
```
Unisay, line 54: HLint warning fixed along the way

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1225,7 +1230,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
             {-# INLINE costingFun #-}
             consByteStringMeaning_V1 =
                 let consByteStringDenotation :: Integer -> BS.ByteString -> BS.ByteString
-                    consByteStringDenotation n xs = BS.cons (fromIntegral n) xs
+                    consByteStringDenotation n = BS.cons (fromIntegral n)
```
Unisay, line 1233: HLint warning fixed

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1979,16 +1999,15 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
               -> ListCostedByLength Integer
               -> Bool
               -> BuiltinResult BS.ByteString
-            writeBitsDenotation s (ListCostedByLength ixs) bit =
-               Bitwise.writeBits s ixs bit
+            writeBitsDenotation s (ListCostedByLength ixs) = Bitwise.writeBits s ixs
```
Unisay, line 2002: HLint warning fixed

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1979,16 +1999,15 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
             {-# INLINE writeBitsDenotation #-}
         in makeBuiltinMeaning
             writeBitsDenotation
             (runCostingFunThreeArguments . paramWriteBits)
 
     toBuiltinMeaning _semvar ReplicateByte =
         let replicateByteDenotation :: NumBytesCostedAsNumWords -> Word8 -> BuiltinResult BS.ByteString
-            replicateByteDenotation (NumBytesCostedAsNumWords n) w = Bitwise.replicateByte n w
+            replicateByteDenotation (NumBytesCostedAsNumWords n) = Bitwise.replicateByte n
```
Unisay, line 2010: HLint warning fixed

plutus-core/plutus-core/src/PlutusCore/Pretty/PrettyConst.hs
```diff
@@ -134,7 +136,7 @@ asBytes x = Text 2 $ T.pack $ addLeadingZero $ showHex x mempty
               | otherwise = id
 
 toBytes :: BS.ByteString -> Doc ann
-toBytes b = fold (asBytes <$> BS.unpack b)
+toBytes  = foldMap asBytes . BS.unpack
```
Unisay, line 139: HLint warning fixed

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -110,6 +110,8 @@ builtinMemoryModels = BuiltinCostModelBase
   , paramHeadList                        = Id $ ModelOneArgumentConstantCost    32
   , paramTailList                        = Id $ ModelOneArgumentConstantCost    32
   , paramNullList                        = Id $ ModelOneArgumentConstantCost    32
+  , paramLengthArray                     = Id $ ModelOneArgumentConstantCost    99
+  , paramListToArray                     = Id $ ModelOneArgumentConstantCost    99
```
Unisay, line 114: Dummy values

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
```diff
@@ -294,6 +298,20 @@ instance ArbitraryBuiltin a => ArbitraryBuiltin [a] where
             scale (`div` len) . coerce $ arbitrary @(AsArbitraryBuiltin a)
     shrinkBuiltin = coerce $ shrink @[AsArbitraryBuiltin a]
 
+instance ArbitraryBuiltin a => ArbitraryBuiltin (Strict.Vector a) where
+  arbitraryBuiltin = do
+    spine <- Strict.fromLazy <$> arbitrary
+    let len = length spine
+    for spine $ \() ->
```
effectfully, line 305: Could you define in terms of the list instance instead of the `Vector` one? Just `arbitraryBuiltin = Strict.fromList <$> arbitraryBuiltin` or something. And the same for `shrink`.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
```diff
@@ -368,6 +371,30 @@ test_IdBuiltinList =
...
 
+test_BuiltinArray :: TestTree
+test_BuiltinArray =
+  testGroup "BuiltinArray" [
+    testCase "listToArray" do
+      let listOfInts = mkConstant @[Integer] @DefaultUni () [1..10]
+      let arrayOfInts = mkConstant @(Vector Integer) @DefaultUni () (Vector.fromList [1..10])
+      let term = apply () (tyInst () (builtin () ListToArray) integer) listOfInts
+      typecheckEvaluateCekNoEmit def defaultBuiltinCostModelForTesting term @?=
+          Right (EvaluationSuccess arrayOfInts)
```
effectfully, line 382: Some edge cases would be nice, like an empty array etc.

You should've probably used

```haskell
evals
    :: DefaultUni `HasTermLevel` a
    => a
    -> DefaultFun
    -> [Type TyName DefaultUni ()]
    -> [Term TyName Name DefaultUni DefaultFun ()]
    -> TestNested
```

from the below, we use it for unit tests. Not important though.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
```diff
@@ -368,6 +371,30 @@ test_IdBuiltinList =
...
+    , testCase "lengthArray" do
+      let arrayOfInts = mkConstant @(Vector Integer) @DefaultUni () (Vector.fromList [1..10])
+      let expectedLength = mkConstant @Integer @DefaultUni () 10
+          term = apply () (tyInst () (builtin () LengthArray) integer) arrayOfInts
+      typecheckEvaluateCekNoEmit def defaultBuiltinCostModelForTesting term @?=
+          Right (EvaluationSuccess expectedLength)
+    , testCase "indexArray" do
```
effectfully, line 389: We should also have some property tests, but feel free to add them in a follow-up.

plutus-tx/src/PlutusTx/Builtins/HasBuiltin.hs
effectfully, line 1: I think you need to amend `PlutusTx.Builtins.HasOpaque` as well?

effectfully (commented):
Is the lack of Plinth versions of the builtins intentional?

All looks great. Kinda unfortunate that more and more is needed to add a new builtin, maybe we could automate some more. Although seems not too bad still.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1546,17 +1551,48 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
             (runCostingFunOneArgument . paramNullList)
 
+    toBuiltinMeaning _semvar LengthArray =
+      let lengthArrayDenotation :: SomeConstant uni (Vector a) -> BuiltinResult Int
+          lengthArrayDenotation (SomeConstant (Some (ValueOf uni vec))) =
+            case uni of
+              DefaultUniArray _argUni -> pure $ Vector.length vec
+              _ -> throwing _StructuralUnliftingError "Expected an array but got something else"
+          {-# INLINE lengthArrayDenotation #-}
+        in makeBuiltinMeaning lengthArrayDenotation (runCostingFunOneArgument . paramLengthArray)
```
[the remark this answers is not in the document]
kwxm: I think this is what happens if you try to group related builtins together.  I've always been in the habit of adding new builtins at the very end of the datatype and adding code for dealing with them as the final clause of any function that has to be modified (although I see that this convention's already been violated for `caseData` and `caseList`), and that's why the comment about `unimplementedCostingFunction` is at the end, where I thought it would be visible when someone's adding a new builtin.   The flat tags certainly have to go at the end, and it seems reasonable to try and preserve the historical order that the functions have been added in elsewhere. There's also the tricky business of tagging the cost model parameter names that we have to do in `plutus-ledger-api` ([here](https://github.com/IntersectMBO/plutus/blob/e9b258319b09c2fa657b56841db6ab1cb9db4c6d/plutus-ledger-api/src/PlutusLedgerApi/V3/ParamName.hs#L13) for example) and it might be easier to keep track of that if we keep stuff in the same order everywhere.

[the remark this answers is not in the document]
effectfully: I wouldn't mind it we said that new builtins must go last and move `caseList` and `caseData` there.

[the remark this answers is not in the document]
kwxm: Mind you, in the version in master, `toBuiltinMeaning` and its documentation already take up 1804 lines, so it's hard to see where you could put a comment so that no-one would miss it.  The not-very-easy-to-find [document](https://github.com/IntersectMBO/plutus/blob/master/plutus-core/cost-model/CostModelGeneration.md#adding-a-new-function) about adding a new function and costing it also tell you to use `unimplementedCostingFunction`.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
kwxm: > @kwxm is this what should be done here before costing? Particularly given that listToArray isn't gonna be constant-cost.

Good question.  I think that this doesn't really matter **as long as you remember to fix it**.  The only time this gets used is when you run `generate-cost-model`, which uses this information to fill in the memory function in the JSON version of the cost model, so it has to be correct before you generate the final form of the cost model.

plutus-core/cost-model/data/builtinCostModelA.json
kwxm: >  @kwxm should we add this stuff when adding a new builtin before costing is done?

I think you have to add _something_ here once you've added `paramLengthArray` [here](https://github.com/IntersectMBO/plutus/blob/5de42d2aa3c1729c27540e9feced4f0408a5169e/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/BuiltinCostModel.hs#L131), and similarly for the other new functions.  However, if you use `unimplementedCostingFunction` in `toBuiltinMeaning`, these numbers won't be used: they only come into play when you tell the denotation to run the real costing function.  I'd normally add a new builtin without adding any costing code at all in order to test that it behaves as it's supposed to, and then do all of the costing at once. That's just to cut down on the initial amount of work, and the way all the costing code is plumbed in it's probably not obvious that you don't initially need it.  The important thing is that there's no link at all between the builtin and the costing code until you put `runCostingFunction<N>Arguments . param<MyNewBuiltin>` in `toBuiltinMeaning`.  Maybe we should try to make that fact clearer to implementers somewhere.

Unisay: Right, doing it first time it isn't 100% clear where to draw the line, so I followed the compiler and it wanted me to fix JSON as there is TH which reads it and fails to compile if the section is missing.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1546,17 +1551,48 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
 
     toBuiltinMeaning _semvar NullList =
         let nullListDenotation :: SomeConstant uni [a] -> BuiltinResult Bool
-            nullListDenotation (SomeConstant (Some (ValueOf uniListA xs))) = do
+            nullListDenotation (SomeConstant (Some (ValueOf uniListA xs))) =
                 case uniListA of
-                    DefaultUniList _ -> pure $ null xs
-                    _                ->
-                        -- See Note [Structural vs operational errors within builtins].
```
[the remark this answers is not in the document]
Unisay: This was a hot take from my side :) The reasoning is that repeating mechanically the note reference at each call site doesn't scale well (noisy, repetitive) and is better done at the definition site once. If you disagree I'll change it back.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1546,17 +1551,48 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
-                        throwing _StructuralUnliftingError "Expected a list but got something else"
+                    DefaultUniList _argUni -> pure $ null xs
```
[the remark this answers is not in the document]
Unisay: ... except that my personal best practice is to prefer `_` ("something is discarded") over the `_foo` ("foo is discarded"). This is because I've been bitten by discarding info in the past and decided to be more cautious about it.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -36,32 +34,35 @@
...
-import Data.Int
-import Data.Proxy
+import Data.Int (Int16, Int32, Int64, Int8)
+import Data.Proxy (Proxy (Proxy))
 import Data.Text (Text)
 import Data.Text qualified as Text
 import Data.Typeable (typeRep)
-import Data.Word
+import Data.Vector.Strict (Vector)
+import Data.Word (Word16, Word32, Word64)
```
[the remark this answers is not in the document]
Unisay: ouch

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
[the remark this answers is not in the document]
effectfully: Nah, I think you're right, keep it.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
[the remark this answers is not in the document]
effectfully: OK, it's fine, I will make a willful decision not to nitpick on that one.

plutus-tx/src/PlutusTx/Builtins/HasBuiltin.hs
Unisay: Yes, working on the PlutusTx side now.

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
```diff
@@ -78,7 +80,11 @@ conBool =
 -- | Parser for lists.
 conList :: DefaultUni (Esc a) -> Parser [a]
 conList uniA = trailingWhitespace . inBrackets $
-    constantOf ExpectParensNo uniA `sepBy` symbol ","
+  constantOf ExpectParensNo uniA `sepBy` symbol ","
+
+-- | Parser for arrays.
+conArray :: DefaultUni (Esc a) -> Parser (Vector a)
+conArray uniA = Vector.fromList <$> conList uniA
```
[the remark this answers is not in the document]
kwxm: Are there any tests that should have caught this?  It doesn't matter because we'll have to add textual conformance tests later that will definitely catch this, but I had a vague idea that there were some property tests that involved parsing prettyprinted random ASTs.  I haven't tried to find those though.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
kwxm: > Some edge cases would be nice, like an empty array etc.

Those will definitely get tested in the conformance tests later.  Maybe it's  OK to have some overlap here though.

kwxm (approved):
This looks OK modulo a few minor quibbles.  I'd change the costs in `Default.Builtins` to be `unimplementedCostingFunction` just to make sure that we don't get any surprises before the costing's done properly though.

kwxm:
I wonder if we have to worry about `plutus-metatheory` at this stage now.  Because of the ceritfication stuff, `plutus-executables` now depends on `plutus-metatheory` and when you try to build that you get warnings about miissing cases for the new builtins.  These used to  be totally independent so you could get away with updating the metaheory in a separate PR, but maybe now there'll be CI problems because of the warnings.  That'd be annoying.

plutus-tx/src/PlutusTx/Builtins/HasBuiltin.hs
```diff
@@ -91,6 +92,10 @@ instance HasFromBuiltin a => HasFromBuiltin (BuiltinList a) where
     type FromBuiltin (BuiltinList a) = [FromBuiltin a]
     fromBuiltin (BuiltinList xs) = map fromBuiltin xs
 
+instance HasToBuiltin a => HasToBuiltin (Vector a) where
+    type ToBuiltin (Vector a) = BuiltinArray (ToBuiltin a)
+    toBuiltin = useToOpaque (BuiltinArray . map toBuiltin)
```
kwxm, line 97: This is causing a CI failure because `BultinArray` doesn't exist yet in this PR.  Can we just remove this from the PR?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1546,17 +1551,49 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
-                        throwing _StructuralUnliftingError "Expected a list but got something else"
+                    DefaultUniList _uniA -> pure $ null xs
+                    _ -> throwing _StructuralUnliftingError "Expected a list but got something else"
             {-# INLINE nullListDenotation #-}
         in makeBuiltinMeaning
             nullListDenotation
             (runCostingFunOneArgument . paramNullList)
 
+    toBuiltinMeaning _semvar LengthArray =
+      let lengthArrayDenotation :: SomeConstant uni (Vector a) -> BuiltinResult Int
```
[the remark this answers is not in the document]
Unisay: I don't mind, @effectfully are you ok with it?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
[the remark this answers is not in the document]
Unisay: You're right, this is exactly what happened: I tried to add new builtins in a place where they (arguably) make sense  (e.g. arrays come after lists) and its for this reason I didn't spot the comment in the end of a very large instance. 

Now, as you explained the benefits (1. historical order 2. guidance comments waiting right where they can be noticed)  I'll move the array functions to the end. I can also move `caseList` and `caseData` but its not clear if they were added strictly before array builtins or earlier.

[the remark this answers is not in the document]
Unisay: P.S. there are code comments that categorize builtins, these add to the confusion as they imply semantic ordering, not temporal:
```
-- Bitwise operations
...
-- Conversions
...
-- Logical
```

plutus-tx/src/PlutusTx/Builtins/HasBuiltin.hs
Unisay: You're right, it fails atm. My idea was to first merge [adjacent PR ](https://github.com/IntersectMBO/plutus/pull/6749)(related changes in Plinth) into this one first, ok?

plutus-core/plutus-core/src/PlutusCore/Crypto/BLS12_381/Pairing.hs
Unisay: point taken!

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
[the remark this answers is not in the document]
Unisay: Thanks for bringing it up! I have updated the generator to also generate arrays, this caused 2 test failures, which I fixed exactly like you suggested: by updating the parser.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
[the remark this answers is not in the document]
effectfully: Yes, sorry, I think it was me who gave you the bad advice.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -301,6 +314,15 @@ instance ExMemoryUsage a => ExMemoryUsage [a] where
         {-# INLINE consRose #-}
     {-# INLINE memoryUsage #-}
 
+instance ExMemoryUsage a => ExMemoryUsage (Vector a) where
+    -- sizeof(Vector v) = (7 + N) words + N * sizeof(v)
+    memoryUsage v = CostRose arrayCost [ memoryUsage a | a <- Vector.toList v ]
```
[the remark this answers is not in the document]
effectfully: Nah, let's keep it as-is, but could you please add a short comment that we could use `singletonRose` here and it would be faster and likely still fine?

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Definition.hs
effectfully: Yeah, I think it's better to have some overlap here, since it's cheap.

plutus-tx/src/PlutusTx/Builtins/HasBuiltin.hs
Unisay: Merged.

Unisay:
CI is stuck @ "Waiting for status to be reported" but [I found it](https://ci.iog.io/jobset/IntersectMBO-plutus/pullrequest-6727) directly in Hydra UI and its green. Going to merge it manually.

kwxm:
Stuck on "Waiting for statusto be reported", so merging regardless.

PR #6728, opened 2024-12-02
Disambiguate uc/lc filenames in conformance tests.

kwxm:
PR #6720 added a couple of new tests for the concrete syntax of `bool`, including a check for `true` and `false` where `True` or `False` is expected.  This led to the tests containing directories called `True`, `true`, `False`, and `false`, but unfortunately it turns out that some filesystems are still case-insensitive, meaning that the new tests lead to apparently duplicated paths in the filesystem.  This PR prevents that by renaming the new files.


ramsay-t:
Seems to work for me. On the one hand, the fact that new Macs are case insensitive by default is silly, on the other hand we probably shouldn't break on them?

ramsay-t (approved):
This ought to fix the immediate problem, although some of the CI with "darwin" in the name is failing?...

kwxm:
> This ought to fix the immediate problem, although some of the CI with "darwin" in the name is failing?...

I think that's because of the issue that #6729's supposed to fix.

kwxm:
Yeah, those CI errors went away after rebasing on #6729.

Issue #6714, opened 2024-11-27
Fix behaviour of failing conformance tests with `--accept`.


kwxm:
The Agda evaluator isn't always completely up to date with the Haskell one, and because of this we have a feature that allows us to mark some conformance tests as being [expected to fail](https://github.com/IntersectMBO/plutus/blob/c469a93257044566dc3c454c73b9cad22a554ed0/plutus-conformance/agda/Spec.hs#L114).  This uses `expectFail` from [Test.Tasty.ExpectedFailure](https://hackage.haskell.org/package/tasty-expected-failure-0.12.3/docs/Test-Tasty-ExpectedFailure.html).  For example, some of the bitwise builtins use a special size measure for budgeting that the metatheory doesn't know about at the time of writing and when you run  `cabal test agda-conformance` you get things like this
```
Test suite agda-conformance: RUNNING...
UPLC evaluation tests
  evaluation
    builtin
      semantics
        writeBits
          case-32
            case-32 (evaluation): OK (0.07s)
            case-32 (budget):     FAIL (expected) (0.09s)
            [etc. etc. etc.] 
```
(the problem being that there are golden files containing the expected budget and the Agda evaluator currently produces results that disagree with these).  When you run the `haskell-confomance` tests these all pass.

However (as discovered by @Unisay), if you run `cabal test agda-conformance --test-option --accept` then _the incorrect budget results are accepted_ and the golden files are updated accordingly.  This is arguably a bug in `Test.Tasty.ExpectedFailure`.

Now `Test.Testy.ExpectedFailure` also provides `ignoreTest` and when you use that instead of `expectFail` the output changes to 
```
       ...
       writeBits
          case-32
            case-32 (evaluation): OK (0.06s)
            case-32 (budget):     IGNORED
```
Futhermore, if you use `--accept` then the golden files are _not_ accepted, which is what we want.  The disadvantage to doing this is that with `expectFailure`, when we fix the Agda evaluator to behave identically to the Haskell one, the tests which are expected to fail now start to pass, which causes an error and reminds us to remove the formerly problematic tests from the  list of tests which are expected to fail; with `ingoreTest` they just continue to be ignored, so we might forget to update the list of problematic tests.

We should work out whether the behaviour with `expectFail` and `--accept` really is a bug and report it if so.  In the meantime we should maybe use `ignoreTest` instead of `expectFail`.

kwxm:
[...]

PR #6730, opened 2024-12-03
Kwxm/conformance/improve filename check

### LINKED ISSUE #6660: Conformance machinery can handle test names a bit misleadingly
The conformance test machinery [traverses the directory hierarchy](https://github.com/IntersectMBO/plutus/blob/e0b8db206b16f296529f5b45939523f4683f0f2e/plutus-conformance/src/PlutusConformance/Common.hs#L72) looking for directories with no subdirectories and when it finds such a directory it checks that it contains exactly one `.uplc` file and loads it, executes it, and checks  the result and budget against golden files in the same directory.  However, it doesn't check that the name of the `.uplc`  file matches that of the directory, but it _does_ expect the golden files to have basenames that match the name of the  directory.  Thus if you have a directory called `test-123` and it contains `test-456.uplc` it expects the golden files to be in `test-123-uplc.expected` and `test-123.uplc.budget.expected`.  This can be confusing if you create a source file whose name doesn't match that of the directory containing it.

This issue emerged after PR #6627; see also #6659.

kwxm:
This fixes #6660.  See the issue for a description of the problem. The README file was a bit out of date so I updated that as well.  

A related issue is #6714: that'll be dealt with in a separate PR.


Unisay (approved):

plutus-conformance/src/PlutusConformance/Common.hs
```diff
@@ -146,25 +170,21 @@ expectedToProg txt
...
 getTestedValue ::
     UplcEvaluatorFun res
     -> FilePath
     -> IO (Either T.Text res)
-getTestedValue eval dir = do
-    inputFile <- findByExtension [".uplc"] dir
-    case inputFile of
-        [] -> error $ "Input file missing in " <> dir
-        _:_:_ -> error $ "More than 1 input files in " <> dir
-        [file] -> do
```
kwxm, line 160: At this point it checks that  there's exactly one `.uplc` file in the directory, but it doesn't check that the filename matches the name of the directory.

PR #6734, opened 2024-12-09
Modify certifier to produce new form of certificates

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1280

`uplc optimise --certify <cert-name>` now also runs the certifier and checks that the optimisation is sound. It produces a certificate `<cert-name>.agda` which can be succesfuly type-checked by Agda. See example below.


ana-pantilie:
[Certificate.agda.txt](https://github.com/user-attachments/files/18062612/Certificate.agda.txt)

^ the certificate produced for factorial

zeme-wana (approved):

PR #6741, opened 2024-12-11
Remove taskset usage from ci-plutus-benchmarks.sh

zeme-wana:
The benchmarking machine has been configured to use the "performance" frequency governor on all cores.
This means that we no longer need to specify which core to run on, nor do we need to set the governor each time we run the benchmarks.


kwxm:
Note: for reference, this appears to set the CPU frequency to about 4200 MHz on all of the cores.

PR #6746, opened 2024-12-12
Ramsay t/translation efficiency 2

ramsay-t:
I broke the other PR :( sorry. This one should be properly signed and mergable?


ana-pantilie:
@ramsay-t let me know when this is ready for review again.

ana-pantilie (commented):
@ramsay-t let me know when this is ready for review again.

ramsay-t:
@ana-pantilie @Unisay I've done some fairly horrible things with our golden testing functions but it seems to work :) There are two examples - one currently certifying and one not - with a simple process to add more. 

Yes, technically a test of the certifier ought to not invoke the optimiser and should use the AST stacks as input, rather than UPLC and calling the optimiser, but that would be really annoying to serialise and store. I would argue that if we change the optimiser enough to change the code it produces then we probably should re-read the certifier tests at least quickly to make we haven't changed anything the certifier *should* care about.

plutus-executables/test/certifier/Spec.hs
```diff
@@ -0,0 +1,62 @@
...
+makeTestTree :: [ String ] -> [ TestTree ]
+makeTestTree = map (\s -> makeGoldenUplcCert fixedPath s)
+
+main :: IO ()
+main = do
+  setLocaleEncoding utf8
+  defaultMain $
+    testGroup "Certification"
+    [ testGroup "simple certification"  $ makeTestTree simpleTests
+    , testGroup "failing certification"  $ makeTestTree failingTests
```
ana-pantilie, line 61: It's good that we can run tests on CI now, but an important aspect of testing is communicating intent: why do we expect the `len` test to fail?

ramsay-t: As the comment above the definition on line 48 says: they "fail" in the sense that the certifier says "no" (currently). We will need to look into why that is, but you explicitly asked me to include this so we know the semantics aren't changed (until we do change them).

Unisay:
Please re-format the `plutus-executables/test/certifier/Spec.hs` with fourmolu

plutus-executables/test/certifier/Spec.hs
```diff
@@ -0,0 +1,62 @@
+{-# LANGUAGE OverloadedStrings #-}
```
Unisay, line 1: This pragma is likely not needed as the language extension is already switched on by default in  the cabal file.

plutus-executables/test/certifier/Spec.hs
ana-pantilie: Usually you want to add tests which actually behave the way you expect them to. So this failing test should be transformed into a succeeding test, and commented out until it's fixed.

ana-pantilie:
Can you add the rest of the tests we were using to check the certifier manually? Like the ones generated by the `uplc` executable (factorial, fibonacci etc.).

effectfully (approved):
Fine by me, but to repeat my previous point I currently do not believe this is the right solution long-term despite it making things faster short-term. More on that [here](https://github.com/IntersectMBO/plutus/pull/6713#issuecomment-2505645505).

plutus-executables/test/certifier/Spec.hs
ramsay-t: Perhaps "failing" is the wrong name - we want some tests that _don't_ certify. Otherwise we are just demonstrating that the certifier says "yes", and not that it actually checks things. I'm not sure what English word you would prefer to "failing" in that case? Rejected, or Negative?

plutus-executables/test/certifier/Spec.hs
```diff
@@ -0,0 +1,62 @@
...
+
+{- | The tests in this file run tests of the uplc certifier. Various unoptimised UPLC is
```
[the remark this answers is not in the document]
ramsay-t: ...this is 89?

I'm all for the classic 80 character limit if you'd prefer, but then lets set the github commit hook to enforce that.

plutus-executables/test/certifier/Spec.hs
ramsay-t: Yes, I know really we should have tests that we _intend_ to be rejected, but I was hoping to avoid turning this into a solid week of test engineering ;) I shall do that though...

ramsay-t:
> Can you add the rest of the tests we were using to check the certifier manually? Like the ones generated by the `uplc` executable (factorial, fibonacci etc.).

The problem is they are randomly generated, so we can't do them as "golden" tests. Then we would have to decide what we are testing? Do we assume they all certify? Are we defining that as the last line of the response being "The compilation was successfully certified." - or do we actually set the exit code properly in the executable when certification fails?

ana-pantilie:
@ramsay-t 

> The problem is they are randomly generated, so we can't do them as "golden" tests. 

How are they randomly generated? They're just UPLC programs, we can add the UPLC code as you have with the other tests you added here. 

> Then we would have to decide what we are testing? Do we assume they all certify? 

It doesn't make sense to have integration tests which are expected to fail. That would mean that either the compiler or the certifier is broken in some way.

> Are we defining that as the last line of the response being "The compilation was successfully certified." - or do we actually set the exit code properly in the executable when certification fails?

I think just checking the text is fine for now, we can improve that in the future.

plutus-executables/plutus-executables.cabal
ramsay-t, line 1: @effectfully this appears to have changed nothing in the CI :( perhaps you meant something different?

effectfully: I think you need to import it, like it's done in this file with `os-support`.

ramsay-t: Well, that fails _fast_ at least :)

plutus-executables/test/certifier/Spec.hs
```diff
@@ -0,0 +1,105 @@
...
+  [ "inc"
+  -- , len
+  ]
+
+-- These run but the certifier says "no"
+rejectedTests :: [String]
+rejectedTests =
+  [
+
+  ]
```
ana-pantilie, line 89: It doesn't make sense to keep this. It's dead code.

plutus-executables/test/certifier/Spec.hs
```diff
@@ -0,0 +1,105 @@
...
+
+-- We were just calling the nested stuff with this constant, so it
+-- might as well be constant for now.
+fixedPath :: [ String ]
+fixedPath = ["test", "certifier"]
+
+srcTests :: [ String ]
+srcTests =
+  [ "inc"
+  -- , len
```
ana-pantilie, line 81: At least add a comment describing why this test is not enabled.

plutus-executables/test/certifier/Spec.hs
```diff
@@ -0,0 +1,105 @@
...
+    $ "The compilation was successfully certified." == lastLine
+
+-- We were just calling the nested stuff with this constant, so it
+-- might as well be constant for now.
+fixedPath :: [ String ]
+fixedPath = ["test", "certifier"]
+
+srcTests :: [ String ]
+srcTests =
+  [ "inc"
```
ana-pantilie, line 80: Can you please add the other tests as well as I asked? That should be very quick. I would normally be ok with you adding these in a separate PR, but since you decided to combine the testing framework with changes to the decision procedures as well, I will have to insist that you show that the changes to the decision procedures do not break any of the tests we have been running manually until now.

ana-pantilie (changes requested):

plutus-executables/test/certifier/Spec.hs
ramsay-t: No it isn't its called on line 104 😸

plutus-executables/test/certifier/Spec.hs
ramsay-t: They are in the makeExample list....? Since they are generated from the executable it made more sense to let it generate fresh versions each time and certify those. We could include hardcoded instances or each of them, but the fresh ones are maybe more interesting?

ana-pantilie (approved):

plutus-executables/test/certifier/Spec.hs
ramsay-t: Except: why do these never trigger the bug that the short “len” example triggers? They are probably already optimised and so are just returning an ID change all the time…. If we can make them unoptimised then they will be more useful but will currently fail (possibly randomly?)

ana-pantilie: I remember that the `uplc` examples are unoptimised. If I printed the `factorial` or whatever script, I remember seeing `force ... delays` that could definitely be reduced.

ramsay-t: Looking at the generated certificates, there does seem to be _some_ changes happening, so these are meaningful tests - although they might be more meaningful if we used the plc examples and then fed those in so there was more work happening...

ramsay-t: I think we were typing at the same time :) These are fine for now, although I could spend ages improving them!

PR #6749, opened 2024-12-13
CIP-0138 BuiltinArray for Plinth

Unisay:
Closes #https://github.com/IntersectMBO/plutus/issues/6731


plutus-ledger-api/test/Spec/CostModelParams.hs
```diff
@@ -28,7 +28,7 @@ tests =
     [ testCase "length" $ do
             166 @=? length v1_ParamNames
             185 @=? length v2_ParamNames
-            297 @=? length v3_ParamNames
+            303 @=? length v3_ParamNames
```
ana-pantilie, line 31: What does this number mean?

plutus-tx-plugin/test/Array/9.6/compiledIndexArray.uplc.golden
```diff
@@ -0,0 +1 @@
+(program 1.1.0 (force indexArray [I 1, I 2, I 3] 2))
```
ana-pantilie, line 1: Why do the elements inside the array have `I` constructors, while the integer argument to `indexArray` (`2`) doesn't?

ana-pantilie (approved):
This doesn't implement the builtin functions, right? It just defines them and the compiler recognizes them as builtins, but the evaluator doesn't yet know how to evaluate them. I assume that will be done in a separate commit/PR.

plutus-ledger-api/test/Spec/CostModelParams.hs
Unisay: It means that for each function added (3) there are 2 parameter names introduced to the cost model, making it 6 parameters overall.

plutus-tx-plugin/test/Array/9.6/compiledIndexArray.uplc.golden
Unisay: Idk, looks like a pretty-printing issue to me

plutus-ledger-api/src/PlutusLedgerApi/V3/ParamName.hs
```diff
@@ -97,6 +97,12 @@ data ParamName =
   | FstPair'memory'arguments
   | HeadList'cpu'arguments
   | HeadList'memory'arguments
+  | LengthArray'cpu'arguments
```
zliu41, line 100: New parameters must be appended, not inserted in the middle.

And these names don't seem accurate - e.g., `ListToArray` is linear, so there needs to be an intercept and a slope. If they aren't accurate, then I'm not sure what the point is adding them at all. @kwxm What do you think?

zliu41: Also these need to be added to V1 and V2 as well.

zliu41 (changes requested):

plutus-ledger-api/src/PlutusLedgerApi/V3/ParamName.hs
zliu41: > I wrote a golden test which will fail if this condition is violated

How do we update the golden file when new parameters are added?

> The point is to make build green. Some parameters have to be added, otherwise CI tests fail.

I'll let @kwxm check whether this is the right thing to do.

Unisay: > How do we update the golden file when new parameters are added?

`cabal test plutus-ledger-api-test`  fails if new parameter names were inserted in the middle, not appended.
`cabal test plutus-ledger-api-test --test-options --accept` updates the golden file.

kwxm: The easiest thing to do would have been to update `plutus-tx` and `plutus-tx-plugin` and deal with `plutus-ledger-api` in a separate PR after the costing gets done.  I think it'd all get through CI in that case, although I might be wrong.

plutus-tx-plugin/test/Array/9.6/compiledIndexArray.uplc.golden
kwxm: I think there's definitely something wrong there.  The UPLC syntax is compeltely wrong.  It should be something like
```

(program 1.1.0 [[(force (builtin indexArray) (con (array integer) [1,2,3])] (con integer 2)])`
```

Unisay: We looked at this example with @bezirg and came to the conclusion that you're expecting "Classical" representation, while golden file uses "Readable" one.

kwxm (changes requested):
The sutff for `plutus-tx` and `plutus-tx-plugin` all looks OK (although I think there is a problem with the UPLC syntax in some of the golden tests: I'm not sure why that is becuase it's OK elsewhere).  However I'd recommend not committing allof the changes to `plutus-ledger-api`: I think it's enough to add the new builtins to `futurePV` and leave everything else unchanged (but stash the updated tests somewhere in case we any them later!).  It's not really meaningful to extend the cost model parameters until we know what they're going to be, so it's probably simpler to leave them unchaged for now.  We also have to do `expModInteger`, `caseList` and `caseData` as well as the array builtins, and it'd probably be best to do everything at once when all of the costing is finished.

kwxm:
With hindsight I now idly wondering if we could have combined built-in lists and built-in arrays into a single structure like [C++'s vector](https://en.cppreference.com/w/cpp/container/vector) that would give you list-like `cons` but also constant-time access.  The trouble with that is that it's really an array with an index to the head element and it has to be reallocated to something bigger when you run out of space for `cons`ing stuff, which would have been probleamtic for costing.  It's way too late to do that anyway so this is really just an academic comment.

plutus-tx-plugin/test/Array/9.6/compiledIndexArray.uplc.golden
kwxm: > while golden file uses "Readable" one.

Why though?  I think all (?) of the other UPLC golden files in the repository are in "Classical" form.  It's not that important though.

plutus-ledger-api/src/PlutusLedgerApi/V3/ParamName.hs
kwxm: > The point is to make build green. Some parameters have to be added, otherwise CI tests fail.

Is that true? If you put new things in `futurePV` then I think they're not tested.  Maybe I'm missing something though: it's difficult to remember how all this works without going through the whole process of adding something.

PR #6751, opened 2024-12-16
Replace stylish-haskell with fourmolu

ana-pantilie:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -102,12 +102,13 @@ in {
   '';
 
   preCommit = {
-    stylish-haskell.enable = true;
+    stylish-haskell.enable = false;
     cabal-fmt.enable = true;
     shellcheck.enable = false;
     editorconfig-checker.enable = true;
     nixfmt-classic.enable = true;
     optipng.enable = true;
+    # TODO: enable fourmolu once migration from stylish-haskell is almost done
     # fourmolu.enable = true;
     hlint.enable = false;
   };
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -108,8 +108,7 @@ in {
     editorconfig-checker.enable = true;
     nixfmt-classic.enable = true;
     optipng.enable = true;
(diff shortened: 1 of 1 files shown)


Unisay (approved):
Thanks!

ana-pantilie:
@Unisay looks like we also have to disable the `fourmolu` part of the `pre-commit-check` which runs as a separate workflow on all the files

zliu41:
I don't think we are ready to do this, are we? This means if anyone opens a PR that touches an unformatted file, it will be reformatted, potentially making the code review much harder.

ana-pantilie:
@zliu41 I thought we wanted to do this incrementally meaning, if I am working on some part of the codebase I also format it. Yes, this will result in a lot of noise concerning the diff.

If we don't want noise in diffs, then we'll have to do the reformatting in one go, but we decided against that (I don't remember why, though).

zliu41:
I suggest we do neither. Instead, let's format 10 modules at a time, and try picking 10 modules that nobody is actively working on.

ana-pantilie:
Uff, @Unisay asked me to format my PR with fourmolu, or maybe I misunderstood. I thought this meant we decided to format things as we go.

ana-pantilie:
In any case, even picking 10 modules at a time in separate PRs won't work without also messing with the commit precheck. If we don't disable stylish-haskell then it will end up running over the files formatted with fourmolu. If we disable stylish and enable fourmolu, then there's this _other_ job which runs the precheck on the whole repo.

zliu41:
It shouldn't be very difficult
- add `{- FOURMOLU_DISABLE -}` to every `.hs` file
- disable stylish and enable fourmolu in precheck
- reformat 10 modules at a time and remove `{- FOURMOLU_DISABLE -}` after reformatting

PR #6752, opened 2024-12-17
Add fully data-backed V1 and V2 script contexts

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/58

- Makes both V1 and V2 ScriptContexts fully data-backed; V1 has an extra SoP type: `(,)` because it doesn't use `Map`, and I was unsure of whether we want map semantics there or list semantics
- Adds tests similar to the V3 ScriptContext benchmarks


plutus-benchmark/script-contexts/src/PlutusBenchmark/V1/ScriptContexts.hs
```diff
@@ -0,0 +1,138 @@
...
+-- outputs). This should be a somewhat realistic example where a reasonable chunk of work is
+-- done in addition to decoding.
+checkScriptContext1 :: PlutusTx.BuiltinData -> ()
+checkScriptContext1 d =
+  -- Bang pattern to ensure this is forced, probably not necesssary
+  -- since we do use it later
+  let !sc = PlutusTx.unsafeFromBuiltinData d
+      ScriptContext txi _  = sc
+  in
+  if PlutusTx.length (txInfoOutputs txi) `PlutusTx.modInteger` 2 PlutusTx.== 0
```
zliu41, line 61: Using field accessors is in general inefficient compared to record patterns, but probably fine if there's only a single usage in a script.

zliu41 (approved):
Looks good. It's odd that I don't see a changelog file, but CI is not complaining. Am I missing something? Either way, there should be a changelog.

PR #6760, opened 2024-12-20
Fix typos in Plutus Core Spec of Batch 5 builtins

nielstron:
The plutus core spec Builtins Batch 5 Note 1 mentions that for the extending case of and/xor/orByteString, the bytestrings are extended with 0/1/1 respectively. This contradicts conformance test andByteString/case-06 and the statement a few sentences later that "In both cases, the otuput is the same as the input when 𝑘 = 0". Moreover the formular later also contradicts this. I assume this is a simple case of flipped bit in the first mention.

Further:
- fix typo otuput -> output
- fix typo in return type of readBit: bytestring -> bool


kwxm:
Thanks for the corrections!

PR #6773, opened 2025-01-06
[UPLC] [Test] Add scoping tests

effectfully:
This adds scoping tests for UPLC the same way we have them for PIR. It also tweaks the way the scoping machinery handles case-expressions for PIR.

I didn't stumble upon any unexpected behavior.

### DIFF
--- a/plutus-core/plutus-core.cabal
+++ b/plutus-core/plutus-core.cabal
@@ -195,6 +195,7 @@ library
     UntypedPlutusCore.Check.Scope
     UntypedPlutusCore.Check.Uniques
     UntypedPlutusCore.Core
+    UntypedPlutusCore.Core.Instance.Scoping
     UntypedPlutusCore.Core.Plated
     UntypedPlutusCore.Core.Type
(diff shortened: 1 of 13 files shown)


zliu41 (approved):
This looks good but we have a `Pass` data type (though currently only for PIR) where a `Pass` is associated with pre- and post-conditions. Would it be better to express the scoping properties in `Pass` instead of using this more ad-hoc way of testing?

effectfully:
> This looks good but we have a `Pass` data type (though currently only for PIR) where a `Pass` is associated with pre- and post-conditions. Would it be better to express the scoping properties in `Pass` instead of using this more ad-hoc way of testing?

The scoping tests are already rather complex, I'd rather not add any more complexity unless it's needed for something. Currently we only distinguish between

- transformations that require renaming and those that don't
- transformations that remove bindings and those that don't

I don't feel like adding an abstraction on top of this will make anything simpler or more readable. Particularly given that the handling logic is deep within the scoping tests machinery, so abstracting it out of there won't be nice.

PR #6784, opened 2025-01-10
Data backed constitution script

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1286

The performance is slightly worse in all of our benchmarks. They should be compared with the results in https://github.com/IntersectMBO/plutus/pull/6785, because master is not up to date.


zliu41 (approved):
Looks great - the numbers are not bad at all!

Can you open an issue for reducing code duplication, especially in tests. It may be ok to duplicate the validator code for performance, but there's little reason to duplicate test code.

bezirg:
> Looks great - the numbers are not bad at all!
> 
Yes the results look very good and what I was expecting them to be.

cardano-constitution/src/Cardano/Constitution/Validator/Data/Common.hs
```diff
@@ -0,0 +1,91 @@
...
+withChangedParams :: (ChangedParams -> Bool) -> ConstitutionValidator
+withChangedParams fun (scriptContextToValidGovAction -> validGovAction) =
+    case validGovAction of
+        Just cparams ->  if fun cparams
+            then BI.unitval
+            else traceError "ChangedParams failed to validate"
+        Nothing -> BI.unitval -- this is a treasury withdrawal, we just accept it
+{-# INLINABLE withChangedParams #-}
+
+validateParamValue :: ParamValue -> BuiltinData -> Bool
```
bezirg, line 40: I think validateParamValue can be imported from NonData.Common since it seems to be the same

bezirg (approved):

PR #6785, opened 2025-01-10
Regenerate constitution golden tests

ana-pantilie:
These are the tests which have been broken on master for a while now. They are slightly less efficient now. We need to update them. AFAIK the pattern matching builtin stuff is what changed the performance here. I don't know how we can fix this performance regression, but until we figure out a way (or maybe @effectfully already has something in mind) we can't just leave these tests to silently fail on master.


zliu41 (approved):
I think this came up before but could you remind me why we don't run them in CI?

> AFAIK the pattern matching builtin stuff is what changed the performance here

This is surprising. Could you create an issue for investigation?

ana-pantilie:
> I think this came up before but could you remind me why we don't run them in CI?

They take too long, there should be a nightly job which runs them but I don't know which one it is and why it isn't failing. Or maybe it is failing and nobody is being notified of the failures.

bezirg:
> > I think this came up before but could you remind me why we don't run them in CI?
> 
> They take too long, there should be a nightly job which runs them but I don't know which one it is and why it isn't failing. Or maybe it is failing and nobody is being notified of the failures.

Yes they take too long because all tests (unit,random,golden) are bundled together.
I could separate the golden tests in another testsuite and make only that run by the CI.

zliu41:
Sounds good @bezirg please open an issue

PR #6793, opened 2025-01-15
[AST] Add 'Fix'

effectfully:
This is an experiment adding `Fix` to the AST to see how much faster we could recurse.

The main change is adding this constructor to the untyped AST:

```haskell
data Term name uni fun ann
    = <...>
    | Fix !ann !name !(Term name uni fun ann)
```

which in the typed ASTs becomes

```haskell
data Term tyname name uni fun ann
    = <...>
    | Fix ann name (Type tyname uni ann) (Term tyname name uni fun ann)
```

with the following typing:

```haskell
-- [check| G !- ty :: *]    ty ~> vTy    [check| G , rec : vTy !- body : vTy]
-- ----------------------------------------------------------------------------
-- [infer| G !- fix rec ty body : vTy]
inferTypeM (Fix ann rec ty body) = do
    checkKindM ann ty $ Type ()
    vTy <- normalizeTypeM $ void ty
    withVar rec vTy $ checkTypeM ann body vTy
    pure vTy
```

and the following operational semantics:

```
        body ~> bodyV
----------------------------- (1)
fix rec body ~> fix rec bodyV

                  <empty>
-------------------------------------------- (2)
fix rec bodyV ~> [fix rec bodyV / rec] bodyV
```

I.e. we make `fix` a binder so that we can evaluate the body of a `fix` once (the (1) rule) and then substitute it for the name of the recursive call without recomputing it each time (the (2) rule).

In the CK machine this looks as

```haskell
data Frame uni fun
    = <...>
    | FrameFix Name (Type TyName uni ())

-- By (1)
stack |> Fix _ recName ty body = FrameFix recName ty : stack |> body

-- By (2)
FrameFix recName ty : stack <| bodyV =
    let bodyTerm = ckValueToTerm bodyV
    in case bodyV of
        VLamAbs{} -> stack |> termSubstClosedTerm recName (Fix () recName ty bodyTerm) bodyTerm
        _         -> throwingWithCause _MachineError NonLambdaFixedMachineError $ Just bodyTerm
```

The reason why we match on `VLamAbs` is to make the semantics of the CK machine match the ones of the CEK machine where we have to match on `VLamAbs` in order to update its environment, this is how it looks:

```haskell
data Context uni fun ann
    = <...>
    | FrameFix {-# UNPACK #-} !Word64 !(Context uni fun ann)

-- By (1)
computeCek !ctx !env (Fix _ rec body) = do
    stepAndMaybeSpend BFix
    let !len' = Env.length env + 1
    computeCek (FrameFix len' ctx) (Env.cons (VBlackHole (ndbnString rec) len') env) body

-- By (2)
returnCek (FrameFix recIx ctx) bodyV =
    case bodyV of
        VLamAbs nameArg bodyLam env -> do
            let env' = Env.contUpdateZero (\_ -> bodyV') env (Env.length env - recIx)
                bodyV' = VLamAbs nameArg bodyLam env'
            returnCek ctx bodyV'
        VBlackHole{} -> throwingDischarged _MachineError FixLoopMachineError bodyV
        _ -> throwingDischarged _MachineError NonLambdaFixedMachineError bodyV
```

The tricky parts:

1. In `computeCek` we save the current size of the environment in `FrameFix` and proceed to evaluate the body in the current environment extended with a `VBlackHole`. `VBlackHole` is a new kind of `CekValue` that we add to provide a value for the recursive call while we're evaluating the body of a fixpoint to a value. If while evaluating the body we reach `VBlackHole`, we know that this is infinite recursion and terminate the program earlier. The name is inspired by black holes in GHC, which are thunks being evaluated (and evaluation of a thunk runs into a blackhole, that is also a loop and the Haskell program terminates).
2. Once we've evaluated the body, we need to update the `rec :=> VBlackHole` environment entry by replacing `VBlackHole` with an actual value for `rec`, for that I added `contUpdateZero` (similar to `contIndexZero`). At that point we might have added new entries to the environment and so we need to calculate the De Bruijn index of `rec` in the environment by subtracting the old size of the environment from the new one.
3. We completely discharge `FrameFix` in `returnCek` without creating any new frames. We're able to achieve that because we encode recursion through a circular object: `env'` refers to `bodyV'` and vice versa -- this is an actual loop in memory. The idea here is to make the environment of `bodyV'` contain `bodyV'` itself, including its environment containing `bodyV'`, including its environment...

This should be the fastest way of evaluating `Fix`, since it doesn't create any additional frames, we just literally evaluate the body of a fixpoint multiple times. But it comes with a big disadvantage of creating an infinite structure requiring very careful handling if we want to support it in `dischargeCekValue` or the `Show` instance (we'll need to either use `unsafePtrEquality` to check for pointer loops or make `VLamAbs` carry a different sort of name indicating that it's an infinitely unrolled fixpoint).

Yet, even with such a performance-focused design the results are very underwhelming: `nofib` [isn't](https://github.com/IntersectMBO/plutus/pull/6793#issuecomment-2623609948) detectably faster (do we have other benchmarks that get recompiled every time?). The `lists` benchmarks are [faster](https://github.com/IntersectMBO/plutus/pull/6793#issuecomment-2623671591) by up to 14.5%, which is normally a very welcome improvement, but given that we're adding an entire new AST node (with _very_ non-trivial evaluation rules), is it worth it?

Note that we don't save much size either, this is what replacing the Z combinator with native recursion in a script amounts to:
image


plutus-benchmark/lists/test/Sum/9.6/left-fold-scott.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 82856900
-| mem: 454900})
\ No newline at end of file
+({cpu: 77960900
+| mem: 424300})
```
effectfully, line 2: 5.9% for CPU and 6.7% for MEM is better, but still not a lot.

plutus-benchmark/nofib/test/9.6/queens5-fc.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 182013934621
-| mem: 1040710766})
\ No newline at end of file
+({cpu: 168637918621
+| mem: 957110666})
```
effectfully, line 2: 7.3% for CPU and 8% for MEM.

plutus-benchmark/bitwise/test/9.6/Ed25519.size.golden
```diff
@@ -1 +1 @@
-3417
\ No newline at end of file
+3410
```
effectfully, line 1: Size differences are all insignificant, unsurprisingly.

effectfully:
looooooooooooooool

effectfully:
> looooooooooooooool

OK, so that was noise. It's kinda hard to believe that the new way of doing recursion is as efficient as the old one... But at the same time I don't think that the benchmarks don't use the old Z combinator, because the budget golden tests _are_ updated. This is really funny, is the cost of doing recursion this insignificant for complex programs like `nofib`? Let's see what the `lists` benchmark say.

effectfully:
(I've benchmarked `validation` just to make sure that they're not slower, I didn't expect them to become faster, since they're the old scripts with the Z combinator)

plutus-benchmark/lists/test/Sum/9.6/right-fold-scott.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 87656900
-| mem: 484900})
\ No newline at end of file
+({cpu: 82760900
+| mem: 454300})
```
effectfully, line 2: 5.6% for CPU and 9.9% for MEM

michaelpj:
What happens if we evaluate a reference to `rec` before we have done a substitution? It will fail, right? That's a little odd. Should we really be requiring the body of the fix to be a syntactic value or something? 

> The reason why we match on VLamAbs is to make the semantics of the CK machine match the ones of the CEK machine where we have to match on VLamAbs in order to update its environment, this is how it looks

What if we changed the definition of `CekValue` so that values _always_ had an environment? In most cases that would be empty; in the case of a lambda/delay it would generally have something in it, but crucially it would let us add to the environment of a value regardless of what the value is. This is also what we'd want if we were going to add a native `let` to PLC (unsurprisingly!): you just want it to be the value of the inner term plus whatever new bindings you have.

> This should be the fastest way of evaluating Fix, since it doesn't create any additional frames, we just literally evaluate the body of a fixpoint multiple times. But it comes with a big disadvantage of creating an infinite structure requiring very careful handling

I agree, this is pretty tricky. I'd be tempted to at least try a version that doesn't rely on the in-memory loop. I'd be interested to see if there is much difference in performance.

---

Overall the benchmarks are a bit surprising. I was thinking there might be size benefits, but also no. Unless we can squeeze more performance out of it, I'm unsure if it's enough of a win. I guess we just don't do _enough_ recursion for it to really bite :thinking:

effectfully:
> What happens if we evaluate a reference to rec before we have done a substitution? It will fail, right? That's a little odd. 

Very good point, thank you. It is indeed odd. But a bottom is a bottom, I guess, we can always says that we don't distinguish one bottom from another like GHC does. Operationally it's kind of a mess though, I agree.

> What if we changed the definition of `CekValue` so that values always had an environment? In most cases that would be empty; in the case of a lambda/delay it would generally have something in it, but crucially it would let us add to the environment of a value regardless of what the value is.

It'd be a bunch of pointless overhead, as long as this specific PR is concerned. I think looking inside of the body of a `fix` once per `fix` node is totally fine. I didn't try thinking whether requiring the body to be a lambda changes any semantics or not, at least we know that we still have to evaluate the body first in order to only do it once.

> I'd be tempted to at least try a version that doesn't rely on the in-memory loop. 

Me too, but I couldn't come up with a solution quickly and ended up implementing a circular program instead 🙈

> I'd be interested to see if there is much difference in performance.

I'd also be curious, but if we agree that what we see doesn't warrant an extra AST node, then neither will another solution and we shouldn't waste time figuring it out.

> Unless we can squeeze more performance out of it, I'm unsure if it's enough of a win. I guess we just don't do enough recursion for it to really bite 🤔

Yeah, I feel the same way and @colll78 said something along these lines too.

michaelpj:
I guess overall I would expect this to have basically identical semantics to our existing recursion combinators, and if it doesn't then that's a bit suspicious.

Being strict does make the naive approach tricky. We can't just stick the recursively defined value into an environment, since it's not a value until we evaluate it, which we can't do until we've put it in the environment. Did we consider a version that is instead an encoding of Z? So we know that the body can be eta-expanded?

wadler:
Re: @michaelpj comment above. Yes, it is common to require the body of the fix to be a value. Here are the relevant lines from a denotational semantics I wrote recently. Might be worth checking whether requiring the fixpoint body to be a lambda abstraction makes a difference in the benchmarks.

Intrinsically-typed term constructor (where `Γ ⊢ A` is a term in context `Γ` of type `A`).
```
  μƛ_ :     Γ , A ⇒ B , A ⊢ B
            -----------------
          → Γ ⊢ A ⇒ B
```

Fixpoint operator.
```
fixpoint : {{_ : Pointed X}} → (X →ᵐ X) → X
fixpoint f = ⊔ (iterate f ⊥ least)
```

Denotational semantics
```
⟦ μƛ N ⟧ᵗ ρ        =  pure (fixpoint {{Pointed⇒}} (λ k v → ⟦ N ⟧ᵗ (ρ , k , v)))
```

effectfully:
> I think what we need to do in `computeCek` is to save the current environment in `FrameFix`, extend the environment with a variable referencing itself and then proceed to `returnCek`.

I did that back then, but forgot to report on it here. The new design is reflected in the PR description. The additional logic didn't have much impact on performance, since it only affects the very first evaluation step of a fixpoint. 

@michaelpj 

> I guess overall I would expect this to have basically identical semantics to our existing recursion combinators, and if it doesn't then that's a bit suspicious.

I do think it's generally OK to replace one bottom with another bottom, particularly when the latter makes evaluation fail earlier or with a more helpful error message.

@wadler 

> Yes, it is common to require the body of the fix to be a value. Here are the relevant lines from a denotational semantics I wrote recently. Might be worth checking whether requiring the fixpoint body to be a lambda abstraction makes a difference in the benchmarks.

Implementation-wise this would be simpler indeed (at the cost of pushing complexity into the compiler where we'd need to ensure that the body is a lambda in codegen and for every optimization), but performance-wise it would only affect the very first evaluation step without affecting all the subsequent ones, so performance gains would be fairly negligible.

I'm gonna close this PR and the whole objective as "won't do". It's way too complex (regardless of the design) for the little benefits that it brings us. There's quite a few things where our efforts are gonna be rewarded much better (such as built-in `Value`, casing over builtins or an eDSL for processing `Data` within a single builtin).

Maybe once we've exhausted all the better ideas on how to improve performance, we'll return to this one and reopen the PR.

PR #6795, opened 2025-01-16
Refactor Nix Code (remove iogx)

zeme-wana:
iogx was created as a top-level abstraction layer for Haskell projects at IOG. It wraps haskell.nix, simplifying the process of setting up a working Flake. It makes several assumptions and provides additional features like a more powerful development shell, Read the Docs integration, and pre-configured pre-commit hooks.

Since its integration into Plutus, some of these features have become obsolete (e.g., Read the Docs), and we’ve realized we can rely more directly on haskell.nix for the rest.


flake.nix
```diff
@@ -2,18 +2,10 @@
   description = "Plutus Core";
 
   inputs = {
-
-    iogx = {
```
Unisay, line 6: By the way, this PR has no explanation, and I am not clear on the following:
- what is the problem with using `iogx` ?
- why is it used in the first place? What value does it add?

flake.nix
```diff
@@ -28,16 +20,21 @@
...
     };
 
+    flake-utils.url = "github:numtide/flake-utils";
+
+    pre-commit-hooks.url = "github:cachix/git-hooks.nix";
+  };
+
+  outputs = inputs: inputs.flake-utils.lib.eachDefaultSystem (system:
+    import ./nix/outputs.nix { inherit inputs system; }
+  );
```
Unisay, line 36: What I understand here is that instead of using `iogx` to create flake outputs we're creating outputs ourselves.

flake.nix
```diff
@@ -28,16 +20,21 @@
...
-      systems = [ "x86_64-darwin" "x86_64-linux" "aarch64-darwin" ];
-      outputs = import ./nix/outputs.nix;
+    iohk-nix = {
+      url = "github:input-output-hk/iohk-nix";
+      inputs.nixpkgs.follows = "nixpkgs";
     };
 
+    flake-utils.url = "github:numtide/flake-utils";
+
+    pre-commit-hooks.url = "github:cachix/git-hooks.nix";
```
Unisay, line 31: IIUC before this output was added by `iogx.lib.mkFlake`, right? Are there any other outputs that were added by `iogx` and are not present now?

nix/build-latex.nix
```diff
@@ -1,12 +1,14 @@
-{ repoRoot, inputs, pkgs, system, lib }:
```
Unisay, line 1: Same here: not clear how these changes are related to the `iogx` removal.

nix/outputs.nix
```diff
@@ -1,84 +1,139 @@
...
+    inherit static-haskell-packages;
+    inherit exposed-haskell-packages;
+    inherit flattened-ci-jobs;
+    inherit nested-ci-jobs;
+  };
+
+in
+
+{
+  inherit __internal;
```
Unisay, line 133: Why is this `__internal` indirection needed? Can't you inherit what's int the `__internal` directly?

nix/outputs.nix
```diff
@@ -1,84 +1,139 @@
...
+    "aarch64-linux" =
+      { };
+    "aarch64-darwin" =
+      (project-variants-roots-and-plan-nix) //
+      { devShells.default = shell; };
+  };
+
+  flattened-ci-jobs = utils.flattenDerivationTree ":" nested-ci-jobs;
+
+  ciJobs = utils.flattenDerivationTree ":" nested-ci-jobs.${system};
```
Unisay, line 110: Can you refer to the `flattened-ci-jobs` above?

nix/project.nix
```diff
@@ -1,6 +1,5 @@
 # editorconfig-checker-disable-file
-
-{ repoRoot, inputs, pkgs, system, lib }:
+{ inputs, pkgs, lib, agda-with-stdlib, r-with-packages }:
```
Unisay, line 2: YaY! this way the dependencies are more explicit, liking it!

nix/project.nix
```diff
@@ -12,18 +11,9 @@ let
 
     src = ../.;
 
-    shell = {
```
Unisay, line 15: Why is this shell getting removed?

nix/r-with-packages.nix
```diff
@@ -1,4 +1,4 @@
-{ repoRoot, inputs, pkgs, system, lib }:
+{ pkgs }:
```
Unisay, line 1: its only the `rPackages` that's used, I suggest declaring it as a parameter directly, avoiding passing in all the unnecessary pkgs.

nix/utils.nix
```diff
@@ -0,0 +1,16 @@
+{ lib }:
+{
+  flattenDerivationTree = separator: set:
```
Unisay, line 3: This function is only used in the `outputs.nix`, I'd simply keep it there.

Unisay (approved):

flake.nix
```diff
@@ -2,18 +2,10 @@
...
-      inputs.hackage.follows = "hackage";
-      inputs.CHaP.follows = "CHaP";
-      inputs.haskell-nix.follows = "haskell-nix";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
-
-    nixpkgs.follows = "haskell-nix/nixpkgs-unstable";
     nixpkgs-2405.follows = "haskell-nix/nixpkgs-2405";
 
+    nixpkgs.follows = "haskell-nix/nixpkgs";
```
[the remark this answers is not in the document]
zeme-wana: If I recall correctly, we used to follow `nixpkgs-unstable` in order to access some packages not available in `nixpkgs`. 
But in general we want to follow a stable `nixpkgs` if possible.

flake.nix
zeme-wana: Good point. I added a comment, and a follow up here: 
I wanted to explore what the code would look like without iogx and whether the shell would be lighter as a result (it is).
At this point, iogx doesn’t seem to provide enough value to justify its (future) maintenance cost. Removing it also makes the codebase more accessible to future maintainers.

flake.nix
zeme-wana: Yes iogx worked some magic and included some outputs in addition to the one explicitly defined by the user.

flake.nix
zeme-wana: Yes this input was added by iogx. There are no other hidden inputs.

nix/agda-with-stdlib.nix
```diff
@@ -1,6 +1,6 @@
-{ repoRoot, inputs, pkgs, system, lib }:
```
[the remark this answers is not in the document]
zeme-wana: iogx expected each nix file to be a function from `{ repoRoot, inputs, pkgs, system, lib }:`.
This is no longer the case.

nix/build-latex.nix
zeme-wana: Same as above

nix/outputs.nix
zeme-wana: I don't want to pollute the top-level flake namespace with a bunch of things. 
This is only used for debugging. 
The top-level should only have `packages`, `devShells`, etc...

nix/outputs.nix
zeme-wana: I wish, but no, we'll have to live with a little repetition

nix/project.nix
zeme-wana: It has been moved to `nix/shell.nix:project.shellFor` which is now the only source of truth for all shell things

nix/r-with-packages.nix
zeme-wana: I'd rather draw the line at `pkgs` and let consumers pull from it.
Also it's both `rPackages` and `rWrapper`

nix/utils.nix
zeme-wana: True but I prefer it here. One day it could be used elsewhere, or nix/utils.nix could have more functions.

PR #6797, opened 2025-01-21
Add flag to dump PIR ASTs for certifier

basetunnel:
Context
---

For my certifier in Coq, I need to get my hands on all intermediate PIR ASTs. I've added a plugin flag `dump-cert-trace`, that dumps all PIR ASTs to a single text file during a compilation.

Dumping to file
---

To prevent keeping all intermediate ASTs in memory, I'm dumping ASTs directly after a pass has run (rather than collecting a pure list of ASTs and dumping them at the end of the pipeline). To that end, I've added a parameter `dumpCert :: Text -> m ()` to `runPass` (in the pass abstraction), which can be used to append text to the dump file. It needs to be of type `Text -> m ()` rather than `Text -> IO ()`, since the monad stack `m` does not necessarily have IO in it (the pass abstraction does not require it). The concrete monad stack in Compiler.hs does have `IO`, so it can be lifted into `m`.

As a consequence, this new argument also shows up in functions like `PIR.compileProgram`, `PIR.compileToReadable`, `PIR.compileReadableToPlc`. This feels very clumsy, but I couldn't find a nice way to do it. Here are two ideas:

- The dump function could go in `CompilationCtx` as a `Text -> IO ()`, but that requires a `MonadIO` constraint in `Compiling` type and the `runPass` function in order to actually run it. I'm not sure if that is desired.
- Similar to how logging works (which uses `traceM`), use unsafePerformIO and add a "pure" function `Monad m => Text -> m ()` in the `CompilationCtx`

Any other suggestions?

Pretty printing ASTs
---

The dumping format is textual, and easily parseable in the proof assistant. It's _almost_ `Show`, but without record syntax and uses `Text` instead of `String` for performance. I've added a new typeclass `SimpleShow` in module `Text.SimpleShow`, which has some basic instances and the possibility for deriving using GHC generics. I've also written/derived instances for all PIR AST types.

I've added the `SimpleShow` constraints to the `Compiling` type/constraint synonym, this required me turning on the following extensions in some places:

- `QuantifiedConstraints` (because of the constraint `forall t. SimpleShow (uni t)`)
- `ImpredicativeTypes`, to add the constraint `forall t. SimpleShow (uni t)` to the `Compiling` type synonym.

Pretty printing pass information
---

In addition to ASTs I'm also dumping information about which pass was run. Here I've added the `PassId` type, which is used as an additional field for every basic `Pass` constructor.

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Pass.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Pass.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE GADTs      #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
 
 module PlutusIR.Pass where
@@ -16,7 +17,10 @@ import Control.Monad.Except (ExceptT, MonadError, throwError)
 import Control.Monad.Trans.Class (lift)
 import Data.Foldable
 import Data.Text (Text)
+import Data.Text qualified as T
+import GHC.Generics (Generic)
 import PlutusCore.Quote
+import PlutusIR.Core.Instance.ShowRocq
 
 -- | A condition on a 'Term'.
 data Condition tyname name uni fun a where
@@ -67,12 +71,43 @@ checkBiCondition c t1 t2 = case c of
     Just (a, e) -> throwError $ CompilationError a e
     Nothing     -> pure ()
 
+data PassId
+  = PassRename
+  | PassDeadCode
+  | PassThunkRec
+  | PassRecSplit
+  | PassLetMerge
+  | PassFloatIn
+  | PassFloatOut
+  | PassCompileLetNonStrict
+  | PassCompileLetType
+  | PassCompileLetData
+  | PassCompileLetRec
+  | PassCompileLetNonRec
+
+  -- Simplifier
+  | PassInline
+  | PassUnwrapWrap
+  | PassCaseReduce
+  | PassCaseOfCase
+  | PassBeta
+  | PassKnownConstructor
+  | PassStrictifyBindings
+  | PassEvaluateBuiltins
+  | PassRewriteRules
+
+  -- Misc
+  | PassTypeCheck
+  | PassOther String
+  deriving stock (Show, Generic)
+
 -- | A pass over a term, with pre- and post-conditions.
 data Pass m tyname name uni fun a =
   -- | A basic pass. Has a function, which is the pass itself, a set of pre-conditions
   -- which are run on the input term, and a set of post-conditions which are run on the
   -- input and output terms (so can compare them).
   Pass
+    PassId
     (Term tyname name uni fun a -> m (Term tyname name uni fun a))
     [Condition tyname name uni fun a]
     [BiCondition tyname name uni fun a]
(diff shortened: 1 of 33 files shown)


effectfully (changes requested):
I don't like `SimpleShow`. I don't think we need field selectors in `Show` output, so let's just get rid of them. Which ones do we have?

If for some reason we do need field selectors, then let's use `TextShow.Generic` instead of rolling out our own inefficient way of doing the same. So we can always do `deriving Show Blah via AsTextNoFields Blah` for a custom `AsTextNoFields` newtype-wrapper whose `Show` instance uses `TextShow.Generic`.

jaccokrijnen:
Thanks @effectfully , I think it's a good idea and I've tried it out. Types that needed a custom `Show` instances are : `Unique`, `Name`, `TyName`, `VarDecl` and `TyVarDecl`.

A different problem I'm running into is that the Show instance for `Text` is not exactly suitable for my purposes in Coq, which parses string literals in a different way (e.g. `\` is not treated in a special way). So I'd prefer to produce a custom format that basically prints the UTF-8 byte sequence. For other parts of the AST I'd just reuse Show instances. Do you have any recommendations on how to do this?

effectfully:
> A different problem I'm running into is that the Show instance for `Text` is not exactly suitable for my purposes in Coq, which parses string literals in a different way (e.g. `\` is not treated in a special way). So I'd prefer to produce a custom format that basically prints the UTF-8 byte sequence. For other parts of the AST I'd just reuse Show instances. Do you have any recommendations on how to do this?
 
Well, that does sound like custom pretty-printing then, given that we can't change the `Show` instance of `Text`. So given that you wouldn't really benefit from `prettyprinter-configurable`, you can do something like this:

```
newtype AsCoq a = AsCoq a
newtype AsCoq1 f a = AsCoq1 (f a)

deriving newtype instance Show (AsCoq Integer)
instance Show (AsCoq Text) where <...>

instance Show (AsCoq (Term Name DefaultUni DefaultFun ann)) where
    showsPrec pr (AsCoq term) = showsPrec (coerce term :: Term Name (AsCoq1 DefaultUni) DefaultFun ann)

deriving anyclass instance Pretty (AsCoq (Term Name DefaultUni DefaultFun ann)) 
```

Hm, except propagating `AsCoq1` in there looks complicated... OK, I'll take a look.

effectfully:
You can take the code from #6815 or fix the `Show` instances at the bottom of `UntypedPlutusCore.Core.Instance.CoqShow` and we'll merge my PR first and then yours.

It's pretty horrible, but you don't need to understand it, just fix the straightforward instances at the bottom.

An example using the custom `Text` instance:

```haskell
example :: Term Name (AsCoqUni DefaultUni) DefaultFun ()
example = Constant () (Some (ValueOf (AsCoqUni DefaultUniString) (AsCoq (Text.pack "abc\n"))))

-- >>> show example
-- "Constant () (Some (ValueOf DefaultUniString abc\n))"
```

effectfully:
Damn I implemented UPLC instead of PIR. Anyway, adding support for PIR should be a matter of adding a few straightforward instances. I'll do it tomorrow.

jaccokrijnen:
Thanks, that's quite an interesting technique! It's the first time I see how the flexibility of `Term`'s type parameters can be useful. I'll use the code and adapt it to PIR and what I need for `Text`.

effectfully:
> It's the first time I see how the flexibility of Term's type parameters can be useful.

The `fun` parameter is very useful for tests and costing calibrations and I believe it was totally worth introducing it, but arguably the `uni` parameter doesn't have a high utility-to-cost ratio.

> I'll use the code and adapt it to PIR

You can just move the `CoqShow` module it to PIR without keeping it on the UPLC side, since I assume we aren't exporting UPLC into Coq quite yet. Please do keep the `mapUni` functions in UPLC and TPLC though, these can be useful for others things.

plutus-core/plutus-ir/src/PlutusIR/Pass.hs
```diff
@@ -88,64 +122,76 @@ instance Monoid (Pass m tyname name uni fun a) where
   mempty = NoOpPass
 
 hoistPass :: (forall v . m v -> n v) -> Pass m tyname name uni fun a -> Pass n tyname name uni fun a
-hoistPass f p = case p of
-  Pass mainPass pre post -> Pass (f . mainPass) pre post
-  CompoundPass p1 p2     -> CompoundPass (hoistPass f p1) (hoistPass f p2)
-  NamedPass n pass       -> NamedPass n (hoistPass f pass)
-  NoOpPass               -> NoOpPass
+hoistPass f = \case
+  Pass p mainPass pre post -> Pass p (f . mainPass) pre post
```
bezirg, line 126: Suggestin: use a more descriptive variable name, like passId or pId

plutus-tx-plugin/src/PlutusTx/Plugin.hs
```diff
@@ -550,19 +553,27 @@ runCompiler moduleName opts expr = do
     -- GHC.Core -> Pir translation.
     pirT <- original <$> (PIR.runDefT annMayInline $ compileExprWithDefs expr)
     let pirP = PIR.Program noProvenance plcVersion pirT
-    when (opts ^. posDumpPir) . liftIO $
+    when (opts ^. posDumpPir) . liftIO $ do
```
bezirg, line 556: nit: unnecessary change

plutus-tx-plugin/src/PlutusTx/Plugin.hs
```diff
@@ -550,19 +553,27 @@ runCompiler moduleName opts expr = do
...
         dumpFlat (void pirP) "initial PIR program" (moduleName ++ "_initial.pir-flat")
 
+    (dumpCert, closeCertHandle) <- liftIO getDumpCert
+    liftIO $ dumpCert (Text.pack (show . UPLC.AsCoq $ void pirT))
+
     -- Pir -> (Simplified) Pir pass. We can then dump/store a more legible PIR program.
-    spirP <- flip runReaderT pirCtx $ PIR.compileToReadable pirP
-    when (opts ^. posDumpPir) . liftIO $
+    spirP <- flip runReaderT pirCtx $ PIR.compileToReadable (liftIO . dumpCert) pirP
+    when (opts ^. posDumpPir) . liftIO $ do
```
bezirg, line 564: also I think unnecessary `do`

plutus-tx-plugin/src/PlutusTx/Plugin.hs
```diff
@@ -550,19 +553,27 @@ runCompiler moduleName opts expr = do
...
+
     -- (Simplified) Pir -> Plc translation.
-    plcP <- flip runReaderT pirCtx $ PIR.compileReadableToPlc spirP
+    plcP <- flip runReaderT pirCtx $ PIR.compileReadableToPlc (liftIO . dumpCert) spirP
     when (opts ^. posDumpPlc) . liftIO $
         dumpFlat (void plcP) "typed PLC program" (moduleName ++ ".tplc-flat")
 
+    -- For now, we only dump the PIR -> PLC ASTs for the certifier,
+    -- so the handle can be closed.
+    liftIO closeCertHandle
```
bezirg, line 575: Not extremely important, but it is better to use the `withHandle`-idiom here to make sure that the handle is closed in case of an exception thrown

bezirg (approved):

PR #6798, opened 2025-01-23
[Evaluation] Autocompute 'NumberOfStepCounters'

effectfully:
This makes it so that we no longer need to update `NumberOfStepCounters`, it'll be updated increased automatically when a new constructor is added to `StepKind`.


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -311,9 +313,20 @@ We keep the counters for each step in the first indices, so we can index them si
 the 'Enum' instance of 'StepKind', and the total counter in the last index.
 -}
 
+-- So that we don't need to update 'NumberOfStepCounters' manually, which would be extremely
+-- error-prone and has caused a bug in the past.
+type CountConstructorsEnum :: (GHC.Type -> GHC.Type) -> Nat
+type family CountConstructorsEnum rep where
```
kwxm, line 319: I was going to suggest that we should put this somewhere else since it's not very Cek-specific, but I don't know where would be a sensible place for it.  Maybe `StepCounter.hs`?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -311,9 +313,20 @@ We keep the counters for each step in the first indices, so we can index them si
...
+    CountConstructorsEnum U1         = 1
+    CountConstructorsEnum (M1 _ _ f) = CountConstructorsEnum f
+    CountConstructorsEnum (f :+: g)  = CountConstructorsEnum f + CountConstructorsEnum g
+    CountConstructorsEnum V1         = TypeError ('Text "Cannot be empty")
+    CountConstructorsEnum (K1 _ _)   = TypeError ('Text "Cannot be a non-enumeration type")
+    CountConstructorsEnum (f :*: g)  = TypeError ('Text "Cannot be a non-enumeration type")
+
```
kwxm, line 326: I don't understand generics or (especially) type families very well, but if I add a case for `Par1` here the whole thing compiles without error.  I can also remove the cases involving `TypeError` and it still all works OK, presumably because they don't come into play for `StepKind`.  I'm not sure exactly what I'm trying to say, but I guess I'm wondering if something could change that would make this stop working (although I suppose the test should detect that kind of thing).

plutus-core/untyped-plutus-core/test/Evaluation/Machines/NumberOfStepCounters.golden
```diff
@@ -0,0 +1 @@
+9
```
kwxm, line 1: We can also compute this number dynamically by looking at the enumeration of `StepKind`.  Would it be sensible to do that, or is this safer?

kwxm (approved):
This looks OK to me. It should certainly avoid the problem that we had before.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: I'd like to keep it in there in case the structure of `StepKind` somehow changes. Like, what if it becomes parameterized for whatever reason? Currently it's written specifically with the structure of the existing `StepKind` in mind.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: OK, I've added

```haskell
    CountConstructorsEnum (Rec1 _ _) = TypeError ('Text "Cannot be a non-enumeration type")
    CountConstructorsEnum (Par1 _ _) =
        TypeError ('Text "If you really want a parameterized type, handle this clause")
```

It doesn't really matter, if one of those isn't handled, at worst you'll get a less readable type checking error and that's it, we don't even need the test for that.

plutus-core/untyped-plutus-core/test/Evaluation/Machines/NumberOfStepCounters.golden
effectfully: Great idea, I've added that test alongside the other one. Thanks!

PR #6801, opened 2025-01-27
Update the auction example in the user guide to Plutus V3

Unisay:
Closed https://github.com/IntersectMBO/plutus/issues/6519

I've also added existing `QuickStart.hs` module as an `executable` to cabal project. 
It allowed me to `cabal run quickstart`  generating `validator.uplc` from `AuctionValidator.hs`.

### DIFF
--- a/doc/docusaurus/static/code/AuctionValidator.hs
+++ b/doc/docusaurus/static/code/AuctionValidator.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE ImportQualifiedPost        #-}
 {-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE NoImplicitPrelude          #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE PatternSynonyms            #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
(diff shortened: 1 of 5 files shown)


doc/docusaurus/static/code/AuctionValidator.hs
```diff
@@ -22,23 +23,30 @@
 {-# OPTIONS_GHC -fno-strictness #-}
 {-# OPTIONS_GHC -fno-unbox-small-strict-fields #-}
 {-# OPTIONS_GHC -fno-unbox-strict-fields #-}
-{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:target-version=1.0.0 #-}
+{-# OPTIONS_GHC -fplugin PlutusTx.Plugin #-}
```
Unisay, line 26: This line fixes the compilation warning incorrectly saying that `plutus-tx-plugin` package is unused.

doc/docusaurus/static/code/AuctionValidator.hs
```diff
@@ -235,34 +242,35 @@ auctionTypedValidator params (AuctionDatum highestBid) redeemer ctx@(ScriptConte
...
 {-# INLINEABLE auctionUntypedValidator #-}
 
-auctionValidatorScript ::
-  AuctionParams ->
-  CompiledCode (BuiltinData -> BuiltinData -> BuiltinData -> PlutusTx.BuiltinUnit)
+auctionValidatorScript :: AuctionParams -> CompiledCode (BuiltinData -> BuiltinUnit)
 auctionValidatorScript params =
   $$(PlutusTx.compile [||auctionUntypedValidator||])
-    `PlutusTx.unsafeApplyCode` PlutusTx.liftCode plcVersion100 params
+    `PlutusTx.unsafeApplyCode` PlutusTx.liftCode plcVersion110 params
```
zliu41, line 273: you can use `lifeCodeDef`

doc/docusaurus/static/code/QuickStart.hs
```diff
@@ -1,26 +1,31 @@
...
+  B.writeFile "validator.uplc.hex"
+    $ Base16.encode
+    $ BS.fromShort
+    $ serialiseCompiledCode
+    $ auctionValidatorScript AuctionParams
+      { apSeller = Crypto.PubKeyHash (stringToBuiltinByteStringHex "")
+      , apCurrencySymbol = Value.currencySymbol "MNFT"
+      , apTokenName = Value.tokenName "NFT"
+      , apMinBid = 100
+      , apEndTime = Time.fromMilliSeconds 1_000_000 -- Replace with your desired end time
```
zliu41, line 29: If you change this, you should also update [this](https://github.com/IntersectMBO/plutus/blob/5458fe526da99a2b03f217c8fa87b1b0791fede5/doc/docusaurus/docs/auction-smart-contract/end-to-end/mint.md?plain=1#L56). But there are other `error`s, so it would make more sense to either keep all of them or change all of them.

zliu41 (approved):
This looks great but can you also update the docs, so that they stay consistent? E.g., [this](https://github.com/IntersectMBO/plutus/blob/5458fe526da99a2b03f217c8fa87b1b0791fede5/doc/docusaurus/docs/auction-smart-contract/on-chain-code.md?plain=1#L65) needs to be updated.

`plutus-tx-template` also needs to be updated, e.g., `GenMintingPolicyBlueprint.hs`.

doc/docusaurus/static/code/AuctionValidator.hs
zliu41: I don't think it is worth doing this just to appease the warning: one must remember to remove this line if the plugin is no longer needed.

Unisay: Also, without this pragma the recompilation avoidance mechanism is not triggering recompilation in case of changes in the plugin/compiler.

doc/docusaurus/static/code/QuickStart.hs
Unisay: https://github.com/IntersectMBO/plutus-tx-template/pull/34

zliu41 (approved):

PR #6814, opened 2025-01-30
Preserve logging by default

zliu41:


effectfully (approved):
That'll be a relief, thank you.

Issue #6603, opened 2024-10-29
Costing for `dropList`


kwxm:
We'll need to do this as part of https://github.com/IntersectMBO/plutus/issues/6593.  That should be straightforward since the CPU time should be a linear function of the number of things you want to drop (and independent of the length of the list or the type of its contents).

robpangilinanioi:
Need to wait on updates from @effectfully before we proceed.

robpangilinanioi:
@kwxm to help out with the JSON script for costing

PR #6817, opened 2025-01-31
Add the 'dropList' builtin, take 2

kwxm:
This is a continuation of #6468.  That became a bit outdated and GitHub was getting confused about conflicts.


kwxm:
I think this causes some extra failures in `cardano-constitution-test`.  There are [already some failures](https://github.com/IntersectMBO/plutus/issues/6807), but there are some golden budget failures as well now.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1479,25 +1487,6 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
             chooseListDenotation
             (runCostingFunThreeArguments . paramChooseList)
 
-    toBuiltinMeaning _ver CaseList =
```
kwxm, line 1482: I moved these nearer the bottom to keep the buitlins in the same order as the tags (and hence implementation date).

ramsay-t:
Fixes #6603

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -2042,6 +2013,69 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
+                List ds     -> headSpine fList [fromValue ds]
+                I i         -> headSpine fI [fromValue i]
+                B b         -> headSpine fB [fromValue b]
+            {-# INLINE caseDataDenotation #-}
+        in makeBuiltinMeaning
+            caseDataDenotation
+            (runCostingFunSixArguments . unimplementedCostingFun)
+
+    toBuiltinMeaning _semvar DropList =
+        let dropListDenotation :: Integer -> SomeConstant uni [a] -> BuiltinResult (Opaque val [a])
```
effectfully, line 2056: This needs to be `IntegerCostedDirectly`.

PR #6818, opened 2025-02-01
Add apply-to-cbor-data option to uplc

kwxm:
On the chain scripts are encoded using `flat` but `data` objects (in redeemers etc) are encoded using CBOR.  This PR adds an `apply-to-cbor-data` option to the `uplc` command that lets you apply a script to a list of such CBOR-encoded `data` objects.

We should really be using the unified `plutus` command instead, but it was a lot quicker for me to add this to `uplc`.


Unisay (approved):
Nice!

bezirg:
I think the unified plutus exe already supports this but I haven't tried it out yet.. Maybe it is a good time to give it a try

bezirg (approved):

PR #6825, opened 2025-02-06
Lift `fix` from PIR/PLC into Plinth

zliu41:
Added `PlutusTx.Function.fix`, which compiles to PIR/PLC term `PlutusCore.StdLib.Data.Function.fix`.

This allows us to implement combinators like [punrollUnbound](https://github.com/Plutonomicon/plutarch-plutus/blob/dd95b2d00abad4e3283604d34bdfeb6479e20118/Plutarch/Unroll.hs#L40-L42), among other things, in Plinth.

There's one issue with this: if you both use `fix`, and have a recursive binding, it will create two separate definitions of `fix`. It is not trivial to combine them, because they are added in two different `runDefT` invocations (one in `runCompiler` and one in `compileLets`). CSE won't combine them since it doesn't check alpha equivalence. I'll address this in a separate PR.


plutus-tx-plugin/test/Recursion/Spec.hs
```diff
@@ -0,0 +1,65 @@
+-- editorconfig-checker-disable-file
```
Unisay, line 1:
```suggestion
```

Unisay (approved):
Got satisfaction seeing that "direct" implementation of `length`  spends the same budget as fix-based `length`!

plutus-tx/src/PlutusTx/Function.hs
```diff
@@ -0,0 +1,5 @@
+module PlutusTx.Function (fix) where
+
+fix :: forall a b. ((a -> b) -> a -> b) -> a -> b
+fix f = let ~x = f x in x
+{-# OPAQUE fix #-}
```
Unisay, line 5: Do I understand correctly that implementation of this function doesn't matter as it is going to be replaced with the `fix` from `PlutusCore.StdLib.Data.Function.fix` ?

effectfully (approved):
I don't understand how that helps us implement `punrollUnbound`, but I don't mind having it.

zliu41:
> I don't understand how that helps us implement punrollUnbound

Because it uses `pfix`?

plutus-tx/src/PlutusTx/Function.hs
zliu41: It doesn't matter for the plugin, but it matters in the sense that we should make the semantics consistent when running a Plinth program in Haskell vs. in the CEK machine.

effectfully:
> Because it uses `pfix`?

Ah, I see, thanks.

PR #6829, opened 2025-02-10
Spec compares results of inlining recursive invocations.

Unisay:
Work in progress.


SeungheonOh:
yay!

plutus-tx-plugin/test/TH/Unroll/Golden/comparison.txt
```diff
@@ -0,0 +1,9 @@
+
+┌────────────────────────────┬──────────────────────┬───────────────────┬─────────────┐
+│ Benchmark                  │ CPU units            │ Memory units      │ Size        │
+├────────────────────────────┼──────────────────────┼───────────────────┼─────────────┤
+│ reference                  │ 2182272722  100.00 % │ 8917554  100.00 % │ 68    100 % │
+│ fix-based                  │ 2159024722   98.93 % │ 8772254   98.37 % │ 68    100 % │
+│ manually unrolled 3        │ 2055872722   94.20 % │ 8127554   91.14 % │ 890  1309 % │
```
zliu41, line 7: I don't think these numbers are very meaningful by themselves, because they entirely depend on the recursion depth and the unrolling depth. If either of them changes then these numbers will also change.

plutus-tx-plugin/test/TH/Unroll/Spec/Lib.hs
```diff
@@ -0,0 +1,210 @@
...
+                                    else
+                                      fibonacciManuallyUnrolled3 (n''' - 1)
+                                        + fibonacciManuallyUnrolled3 (n''' - 2)
+{-# INLINEABLE fibonacciManuallyUnrolled3 #-}
+
+-- Fibonacci function with fix
+fibonacciFix :: Integer -> Integer
+fibonacciFix = fix $ \self n ->
+  if n == 0 || n == 1
+    then n
```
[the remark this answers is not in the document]
Unisay: Fixed, the different in cost increased, as the version with `if n == 0 || n == 1  then n` spends slightly higher budget.

plutus-tx-plugin/test/TH/Unroll/Golden/comparison.txt
Unisay: For me the meaning is in seeing that relative numbers (percentages) decrease and a magnitude of the % difference.

PR #6837, opened 2025-02-11
Add PlutusTx.Optimize.SpaceTime.peel

zliu41:
This is equivalent to Plutarch's [unrollUnbound](https://github.com/Plutonomicon/plutarch-plutus/blob/dd95b2d00abad4e3283604d34bdfeb6479e20118/Plutarch/Unroll.hs#L40-L42). I think "peel" is a better name, since it only peels `n` layers off of the recursion, whereas "unroll" would imply "peeling `n` layers at a time, until the recursion is done".


plutus-tx-plugin/test/Recursion/Spec.hs
```diff
@@ -27,38 +28,49 @@ tests =
...
   fix
-    ( \f xs -> case xs of
+    ( \f -> \case
         []       -> 0
-        (_ : ys) -> 1 PlutusTx.+ f ys
+        (_ : xs) -> 1 PlutusTx.+ f xs
     )
 
+lengthPeeled :: [Integer] -> Integer
+lengthPeeled = $$(peel 3 (\f -> [|| \case [] -> 0; (_ : xs) -> 1 PlutusTx.+ $$f xs ||]))
```
Unisay, line 67: Unfortunately this implementation of `lengthPeeled`  needs to re-implement the `lengthFix` explicitly repeating its step-function code inside the TH quotes.

While this is already a useful macro and for some developers/cases the code repetition is not a big deal, 
I think its possible to improve by implementing it in the plugin with its access to the unfolding of `lengthFix` by
extracting the step-function automatically in order to re-compose it the way `peel` does.

Unisay (approved):

PR #6839, opened 2025-02-12
[Optimization] Make Z inline better

effectfully:
This makes Z inline better, _if_ it's already getting inlined. I.e. if it's already getting inlined, we get a speedup with this PR (and a bit less size consumed), otherwise we get a slowdown (and a bit more size consumed).

If we're gonna start always inlining Z, this is likely beneficial, otherwise not.

The idea is simple: eta-expand Z, so that inlining the stepper function gives us back a value (rather than an application) which can be inlined further. The gist of the change:
image

This partially cancels what #5939 did and reintroduces the strictness issue that that PR fixed, which is probably not good, but I dunno.

It'd probably be best to just mark `fix` as inlinable, since we generate it ourselves anyway and can do it. Then we won't need to rely on it being defined a certain way.


plutus-benchmark/bitwise/test/9.6/8 queens.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 409142294016
-| mem: 1575213647})
\ No newline at end of file
+({cpu: 432732134016
+| mem: 1722650147})
```
effectfully, line 2: I'm rather puzzled as to how eta-expanding Z leads to such a massive slowdown.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/strictLetRec.golden
```diff
@@ -1 +1 @@
-(\xxx -> 1) ((\s -> s s) (\s -> force trace "hello" (\z -> s s z)))
\ No newline at end of file
+1
```
effectfully, line 1: The semantics change.

zliu41 (commented):
This doesn't seem to work as expected? Z is already (unconditionally) inlined if it is only used once, which should be the case in a number of test cases. But I'm not seeing any cost reduction.

plutus-tx-plugin/test/Budget/9.6/sum.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 5476180
-| mem: 28020})
\ No newline at end of file
+({cpu: 5428180
+| mem: 27720})
```
effectfully, line 2: small

plutus-tx-plugin/test/Budget/9.6/listIndexing.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 6684432
-| mem: 33922})
\ No newline at end of file
+({cpu: 6636432
+| mem: 33622})
```
effectfully, line 2: small

effectfully (commented):
<empty>

plutus-tx-plugin/test/Budget/9.6/sum.budget.golden
effectfully, line 2: small

plutus-tx-plugin/test/Budget/9.6/listIndexing.budget.golden
effectfully, line 2: small

effectfully (commented):
a couple of examples

zliu41 (approved):
The results in https://github.com/IntersectMBO/plutus/pull/6842 looks great. I've merged https://github.com/IntersectMBO/plutus/pull/6841, so you can rebase and merge this now.

PR #6840, opened 2025-02-12
`unroll` macros, similar to the `peel`.

Unisay:
unrolling recursion here means such transformation of a recursive function that 
 instead of doing a single function step before calling self it does `n` steps (batches) and then calls self.


plutus-tx/src/PlutusTx/Optimize/SpaceTime.hs
```diff
@@ -1,37 +1,99 @@
...
+   . Int
+  -- ^ How many recursion steps to perform inside the recursion loop.
+  -> (TH.SpliceQ (a -> b) -> TH.SpliceQ (a -> b))
+  {- ^ Function that given a continuation splice returns
+  a splice representing a single recursion step calling this continuation.
+  -}
+  -> TH.SpliceQ (a -> b)
+unroll n f = [||fix \self -> $$(unrollBound n f [||self||])||]
+
+unrollBound
```
zliu41, line 85: This just composes the function with itself `n` times, right? So perhaps call it `nTimes`? There's already a [similar function](https://hackage.haskell.org/package/ghc-9.12.1/docs/GHC-Utils-Misc.html#v:nTimes) (non-TH) in `GHC.Utils.Misc`.

zliu41 (approved):

plutus-tx/src/PlutusTx/Optimize/SpaceTime.hs
Unisay: Thanks, updated to code.

PR #6841, opened 2025-02-12
Inline fixed point combinators by default

zliu41:
I made an earlier attempt (https://github.com/IntersectMBO/plutus/commit/ab4f2853288dde8a7fbb695848ceb93550b8ed41) to allow configuring the aggressiveness of the UPLC inliner, but that didn't work too well, because the size increase is too unpredictable and erratic.

So instead this PR enables inlining just the fixed point combinators. This seems preferable in most cases, and when it does increase script size, the increase is quite mild. So it is on by default. Can be turned off via `no-inline-fix` or `conservative-optimisation`.


Unisay:
I wonder if we should make this mechanism more general, e.g.
UPLC inliner decisions are driven by annotations on individual functions, rather than flags/options. The reason being is that next time we want to instruct UPLC inliner on how to treat another function we won't have to add another flag.

Unisay (approved):

zliu41:
> I wonder if we should make this mechanism more general, e.g.
UPLC inliner decisions are driven by annotations on individual functions, rather than flags/options.

Yes in general, but the fixed point operators are generated by the compiler, not written by the user.

PR #6847, opened 2025-02-13
Add PlutusTx.Optimize.Inline.inline for callsite inlining

zliu41:
This works like `GHC.Magic.inline`, and can be used in the form of `inline f` or `inline (f args)`.


Unisay:
> This works like `GHC.Magic.inline`, and can be used in the form of `inline f` or `inline (f args)`.

Thanks for adding this! To help me understand the motivation behind this change, could you describe the issue it's designed to resolve?  A brief explanation of the problem would be really helpful.

zliu41:
> could you describe the issue it's designed to resolve?

See the tests - the more you inline, the cheaper.

Unisay:
> See the tests - the more you inline, the cheaper.

Its not clear to me why `GHC.inline` is not enough? Why having our own magic is better?

ana-pantilie (approved):

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -797,6 +799,18 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
     -- Lazy &&
     GHC.App (GHC.App (GHC.Var var) a) b | GHC.getName var == boolOperatorAnd ->
       compileExpr $ GHC.mkIfThenElse a b (GHC.Var GHC.falseDataConId)
+    -- `inline f` or `inline (f args)`
```
effectfully, line 802: Nit: I'd perhaps write `f args` as `f x1 ... xn` to make it a bit more clear that it's an iterated application.

zliu41:
@Unisay GHC's `inline` _may_ work, and probably does work most of the time, but there's no guarantee that GHC won't transform it away before it gets to the plugin.

effectfully (approved):
Great stuff.

PR #6848, opened 2025-02-14
Kwxm/crypto/remove ed25519-donna

kwxm:
This removes the `verifyEd25519Signature_V1` builtin variant, which in turn removes our dependency on the `crypton`/`cryptonite` library and the deprecated `cardano-crypto` package.

### DIFF
--- a/doc/notes/model/UTxO.hsproj/Examples/Keys.hs
+++ b/doc/notes/model/UTxO.hsproj/Examples/Keys.hs
@@ -4,10 +4,10 @@
 module Examples.Keys
 where
 
-import "cryptonite" Crypto.PubKey.ECC.ECDSA
+import "crypton" Crypto.PubKey.ECC.ECDSA
 import Crypto.PubKey.ECC.Generate
 import Crypto.PubKey.ECC.Types
-import "cryptonite" Crypto.Random
+import "crypton" Crypto.Random
 
 import Data.Map (Map)
 import Data.Map qualified as Map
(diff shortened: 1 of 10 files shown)


doc/notes/model/UTxO.hsproj/Examples/Keys.hs
kwxm, line 1: This is now the only code that mentions `crypton`.  This code isn't used anywhere and isn't mentioned in `cabal.project`, so presumably it doesn't cause a dependency.  It doesn't build if you do include it in `cabal.project`, but I've made a few small updates anyway.  Maybe we should just ignore it completely.

plutus-tx/src/PlutusTx/Builtins/Internal.hs
```diff
@@ -262,7 +262,7 @@ ripemd_160 (BuiltinByteString b) = BuiltinByteString $ Hash.ripemd_160 b
 
 verifyEd25519Signature :: BuiltinByteString -> BuiltinByteString -> BuiltinByteString -> BuiltinBool
 verifyEd25519Signature (BuiltinByteString vk) (BuiltinByteString msg) (BuiltinByteString sig) =
-  case PlutusCore.Crypto.Ed25519.verifyEd25519Signature_V1 vk msg sig of
+  case PlutusCore.Crypto.Ed25519.verifyEd25519Signature vk msg sig of
```
kwxm, line 265: Presumably this should have been using the V2 variant (I'm not sure what we do about variants in Plinth), but it doesn't matter now anyway.

zliu41: This only matters if one runs Plinth code as Haskell (rather than in the CEK machine).

zliu41 (approved):
:+1:

PR #6854, opened 2025-02-17
Optimise AsData for product types

ana-pantilie:
Part of https://github.com/IntersectMBO/plutus/issues/6770

Modifies `AsData` to match on product types similar to @zliu41 's idea in https://github.com/IntersectMBO/plutus/commit/b31eab9d28789a871ceee36ee2beaea04284c334.


zliu41 (changes requested):
Great stuff, but I think you should use `TH.newName` instead of `TH.mkName`. Otherwise, if the module happens to have a top-level name called `field0` etc. it will cause a name shadowing warning.

plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Interval.hs
```diff
@@ -12,6 +12,9 @@
 {-# OPTIONS_GHC -fno-ignore-interface-pragmas #-}
 {-# OPTIONS_GHC -fno-omit-interface-pragmas #-}
 {-# OPTIONS_GHC -fno-specialise #-}
+-- AsData generates constraints on the type variables for 'Interval'
+-- which in this case are redunant.
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
```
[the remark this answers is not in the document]
ana-pantilie: The new TH code needs to generate a type for each matching function.

In this case `UnsafeFromData a` is not necessary for the fields of `Interval` to be `UnsafeFromData`. The instances are defined for any unconstrained `a`.
Usually, this isn't the case, so the TH code generates the type of the matching function by adding `UnsafeFromData` to each type parameter, as this is usually required.

The other solution would have been to not generate a type at all, and suppress the "top-level binding doesn't have a type definition" warning everywhere `asData` is used, which seemed like a way worse solution.

Ideally, there'd be a way to use Haskell's type inferencer to generate the correct type definition (given some hints), since it already can do that, but there doesn't seem to be a way to access it cleanly from TH.

plutus-tx/src/PlutusTx/IsData/TH.hs
```diff
@@ -57,6 +58,74 @@ mkConstrPartsMatchPattern conIx extractFieldNames =
...
+            (TH.VarP asConstrN)
+            (TH.NormalB $ TH.AppE (TH.VarE 'BI.unsafeDataAsConstr) (TH.VarE builtinData))
+            []
+        , TH.ValD
+            (TH.VarP constrArgsN)
+            (TH.NormalB $ TH.AppE (TH.VarE 'BI.snd) (TH.VarE asConstrN))
+            []
+        ]
+        <> fieldDecs
+        <> restDecs
```
ana-pantilie, line 113: Another very strange thing: if I reverse the order here, the tests change and report worse performance! So the only difference is that the `rest...` variables are declared first, and I imagine the order of declarations should not affect performance or semantics. I didn't look into this too much, but it's something that we might want to concern ourselves with.

zliu41: That's very strange. We should investigate by shrinking the test case and finding a small reproducible example. Please make an issue.

zliu41 (approved):
:+1:

plutus-tx-plugin/test/AsData/Budget/Types.hs
```diff
@@ -9,6 +9,10 @@
 {-# LANGUAGE TypeApplications    #-}
 {-# LANGUAGE ViewPatterns        #-}
 
+{-# OPTIONS_GHC -ddump-splices #-}
+
+
+-- recomp
```
Unisay: Seems like debugging leftovers

ana-pantilie: Removed in the last commit, thanks!

PR #6862, opened 2025-02-19
[Builtins] [Bug] Tweak 'dropList' costing and else

effectfully:
This PR

- fixes the bug with the denotation of `dropList` taking an `Integer` rather than an `IntegerCostedLiterally`
- makes the cost model `constant_cost`
- adds some comments
- adds the missing golden files

### DIFF
--- a/plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
+++ b/plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
@@ -175,6 +175,6 @@ builtinMemoryModels = BuiltinCostModelBase
   , paramExpModInteger                   = Id $ ModelThreeArgumentsConstantCost 100000000000 -- FIXME: stub
   -- paramCaseList
   -- paramCaseData
-  , paramDropList                        = Id $ ModelTwoArgumentsLinearInY $ OneVariableLinearFunction 4 0
+  , paramDropList                        = Id $ ModelTwoArgumentsConstantCost 4
   }
   where identityFunction = OneVariableLinearFunction 0 1
--- a/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
@@ -2051,20 +2051,52 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
             (runCostingFunSixArguments . unimplementedCostingFun)
 
     toBuiltinMeaning _semvar DropList =
-        let dropListDenotation :: Integer -> SomeConstant uni [a] -> BuiltinResult (Opaque val [a])
+        let dropListDenotation
+                :: IntegerCostedLiterally -> SomeConstant uni [a] -> BuiltinResult (Opaque val [a])
             dropListDenotation i (SomeConstant (Some (ValueOf uniListA xs))) = do
                 -- See Note [Operational vs structural errors within builtins].
                 case uniListA of
                     DefaultUniList _ ->
+-- We only support @base-4.15@ and higher, because prior versions don't expose the same interface to
+-- 'Integer'.
 #if MIN_VERSION_base(4,15,0)
-                        fromValueOf uniListA <$> case i of
+                        -- The fastest way of dropping elements from a list is by operating on
+                        -- an unboxed int (i.e. an 'Int#'). We could implement that manually, but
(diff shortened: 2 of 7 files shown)


plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -175,6 +175,6 @@ builtinMemoryModels = BuiltinCostModelBase
   , paramExpModInteger                   = Id $ ModelThreeArgumentsConstantCost 100000000000 -- FIXME: stub
   -- paramCaseList
   -- paramCaseData
-  , paramDropList                        = Id $ ModelTwoArgumentsLinearInY $ OneVariableLinearFunction 4 0
+  , paramDropList                        = Id $ ModelTwoArgumentsConstantCost 4
```
kwxm, line 178: That's 32 bytes.  Are we happy with that?  I'm not sure how much room a pointer takes up, but it's probably 8 bytes at most.  I don't mind all that much though since we're a bit inconsistent about this anyway: many of the builtins have a constant memory cost of 32 `ExMemory` (256 bytes).

kwxm (approved):
This all looks good.  Thanks for reviewing the code and spotting the mistake!

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
effectfully: > That's 32 bytes. Are we happy with that?

I have absolutely no idea.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -2051,20 +2051,52 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
+                            IN _ -> pure xs
+                            -- If the given 'Integer' is higher than @maxBound :: Int@, then we
+                            -- call 'drop' over the latter instead to get the same performance as in
+                            -- the previous case. This will produce a difference result when the
+                            -- list is longer than @maxBound :: Int@, but in practice not only is
+                            -- the budget going to get exhausted long before a @maxBound@ number of
+                            -- elements is skipped, it's not even feasible to skip so many elements
+                            -- for 'Int64' (and we enforce @Int = Int64@ in the @Universe@ module)
+                            -- as that'll take ~3000 years assuming it takes a second to skip
+                            -- @10^8@ elements.
```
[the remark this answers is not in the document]
effectfully: Yeah, but it may end up being allocated lazily, so reasoning about time is more reliable.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -2051,20 +2051,52 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
                     DefaultUniList _ ->
+-- We only support @base-4.15@ and higher, because prior versions don't expose the same interface to
+-- 'Integer'.
 #if MIN_VERSION_base(4,15,0)
-                        fromValueOf uniListA <$> case i of
+                        -- The fastest way of dropping elements from a list is by operating on
+                        -- an unboxed int (i.e. an 'Int#'). We could implement that manually, but
+                        -- 'drop' in @Prelude@ already does that under the hood, so we just need to
+                        -- convert the given 'Integer' to an 'Int' and call 'drop' over that.
+                        fromValueOf uniListA <$> case unIntegerCostedLiterally i of
```
kwxm, line 2067: Is there any reason to prefer `unIntegerCostedLiterally` here rather  than doing 
````
dropListDenotation (IntegerCostedLiterally i) (SomeConstant (Some (ValueOf uniListA xs))) = ...
```
?

I'm not complaining, just wondering if there's some convention or something.  Presumably neither of these incurs any computational cost.

effectfully: I initially wrote the former, but it resulted in this clunky long line, so I just turned it into the latter. It doesn't matter either way.

effectfully:
@kwxm the PR seems to be stuck? If you have the power to observe whether it builds correctly, could you please force-merge it?

kwxm:
Stuck, so merging manually.

Subway2023:
Is there a proof of concept (PoC) for this bug?

effectfully:
No, but I could give you a short description. This bug never reached production, does that fit the scope of your research?

PR #6864, opened 2025-02-19
Test

kwxm:
I just want to see what happens in CI when you have an unknown pragma.


kwxm:
It causes a failure, which is what I was expecting.  However GHC8.12 doesn't know about `{-# OPAQUE #-}` and that seems to get through OK, so maybe the warnings are ingored for 8.12.

kwxm:
Closing this since it was just an experiment.

PR #6871, opened 2025-02-24
Inline bindings that are used once in AsData

zliu41:
This makes it cheaper and smaller.


ana-pantilie (approved):
Nice, thanks for demonstrating this. I'll give it a go on the sum type version.

PR #6876, opened 2025-02-25
[Evaluation] Catch more exceptions

effectfully:
I just suddenly got paranoid, hence this PR.

### DIFF
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
@@ -1,12 +1,14 @@
 -- editorconfig-checker-disable-file
 -- | The CEK machine.
 
+{-# LANGUAGE AllowAmbiguousTypes      #-}
 {-# LANGUAGE BangPatterns             #-}
 {-# LANGUAGE ConstraintKinds          #-}
 {-# LANGUAGE DataKinds                #-}
 {-# LANGUAGE DeriveAnyClass           #-}
 {-# LANGUAGE FlexibleInstances        #-}
 {-# LANGUAGE ImplicitParams           #-}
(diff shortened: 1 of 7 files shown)


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -457,20 +460,45 @@ throwingDischarged
...
+        -> (CekEvaluationException NamedDeBruijn uni fun -> CekM uni fun s a)
+        -> CekM uni fun s a
+    a `catchError` h =
+        -- Here in addition to catching 'CekEvaluationException' we also catch common GHC exceptions
+        -- in case one of them somehow gets triggered during script execution (which would be a bug
+        -- on our side). We could probably use @enclosed-exceptions@, but spawning a thread per
+        -- script may be considered expensive, so we do the simplest thing possible.
+        CekM . unsafeIOToST $ aIO `catches`
+            [ Handler hIO
+            , panicHandler @IOError
```
zliu41, line 481: Why these particular exceptions? Where do you get the list from?

effectfully: I just looked at all the instances of the `Exception` class and picked those that seemed like they would definitely be useful to catch. We could use [this](https://next.fpcomplete.com/blog/async-exception-handling-haskell/) instead to catch all exceptions whose type indicated that they're sync ones:

image

But the type may lie unfortunately.

I'll ask about `enclosed-exceptions` though, it's the right approach, it's just costly.

zliu41 (approved):

Unisay (approved):

effectfully:
I've amended a comment to reflect our discussion with Alexey and added a test, this should be mergeable except I can't merge it because of CI.

PR #6879, opened 2025-02-26
AsData: remove id check from product types on old matching function

ana-pantilie:
Part of https://github.com/IntersectMBO/plutus/issues/6770


ana-pantilie:
Shows that the new AsData encoding is better.

PR #6881, opened 2025-02-27
Improved budget reporting for marlowe validators

Unisay:
Work in progress


ana-pantilie (approved):
❤️

plutus-benchmark/marlowe/test/Lib.hs
```diff
@@ -0,0 +1,169 @@
...
+  the golden file contents
+  -}
+goldenUplcMeasurements name goldenPath outputPath act =
+  goldenTest2
+    name
+    (Text.IO.readFile goldenPath)
+    (withFile outputPath WriteMode act >> Text.IO.readFile outputPath)
+    reportDifference
+    (createDirectoriesAndWriteFile goldenPath)
+    (removeFile outputPath)
```
kwxm, line 62: This doesn't seem to be removing the file (hence my earlier comment about creating new files in the repository without warning).  The documentation for `goldenTest2` says
> The [DeleteOuputFile](https://hackage.haskell.org/package/tasty-golden-2.3.5/docs/Internal.html#v:DeleteOuputFile) option controls the circumstances in which the output file is to be deleted.

The link's broken, but the documentation is [here](https://hackage.haskell.org/package/tasty-golden-2.3.5/docs/Test-Tasty-Golden.html#t:DeleteOutputFile).  Maybe it should be set to `OnPass` or `Always` (although the latter will make it hard to work out what's going on if something does change).  I think you probably have to use something different from `defaultMain` to do that.

plutus-benchmark/marlowe/test/Lib.hs
```diff
@@ -0,0 +1,169 @@
...
+  delta :: Integer -> Integer -> Double
+  delta x y
+    | x > 0 = (fromIntegral (y - x) * 100) / fromIntegral x
+    | otherwise = 0
+
+formatUnits :: Integer -> Text
+formatUnits = F.sformat (F.groupInt 3 ' ')
+
+-- | Like 'writeFile', but also create parent dirs if they are missing.
+createDirectoriesAndWriteFile :: FilePath -> Text -> IO ()
```
kwxm, line 132: I see that Tasty.Golden already has a `createDirectoriesAndWriteFile` function.  Why do we need another?  Does it need a different output function for some reason?

plutus-benchmark/marlowe/test/Lib.hs
```diff
@@ -0,0 +1,169 @@
...
+parseLine ln =
+  case Text.splitOn "\t" ln of
+    [cpu, mem, size] ->
+      (,,)
+        <$> readCpu cpu
+        <*> readMem mem
+        <*> readSize size
+    _ -> throwIO . userError $ "Can't parse line: " <> show ln
+ where
+  readCpu :: Text -> IO ExCPU
```
kwxm, line 150: That's quite a lot of work just to read three numbers, but I don't know if refactoring it would make it any shorter.

plutus-benchmark/marlowe/test/Lib.hs
```diff
@@ -0,0 +1,169 @@
...
+    actualResults <- parse actual
+    if expectedResults == actualResults
+      then pure Nothing
+      else do
+        let (cpu0, mem0, size0) = aggregateResults expectedResults
+            (cpu1, mem1, size1) = aggregateResults actualResults
+
+        pure . Just $
+          "Actual execution budgets and sizes differ from expected ones:\n"
+            <> Tabular.render
```
kwxm, line 77: I generally just use `printf` for stuff like this.  It always makes me feel a bit guilty for some reason, but it has the merit of compactness.  I don't object to this though.

kwxm (approved):
This is a big improvment: those hundreds of tiny budget files always really annoyed me.  It might have been easier just to use a standard golden test with the big file of numbers, but I'm not going to demand that you change it!

PR #6882, opened 2025-02-27
Allow `PlutusTx.Optimize.Inline.inline` to inline local bindings

zliu41:
`inline` currently relies on the variable having an unfolding, but local variables don't have unfoldings, so it fails to inline them.

This PR makes it work better by checking whether the definition of the variable has already been compiled with the corresponding `PIRTerm` in `Scope`. If so, use the `PIRTerm` directly. One limitation is that it only supports `inline f`, not `inline f (x1 ... xn)`.

As an example, see the new test case, `squareSum`.


plutus-tx-plugin/test/Inline/Spec.hs
```diff
@@ -98,6 +102,17 @@ factorial x
   | otherwise = x `PlutusTx.multiplyInteger` factorial (x `PlutusTx.subtractInteger` 1)
 {-# INLINEABLE factorial #-}
 
+-- | This test case verifies that `inline` can inline local bindings
+-- (like `square`).
+--
+-- The third usage of `square` is inlined in PIR, but not in UPLC, since
+-- in UPLC the inlining is reversed by CSE.
```
Unisay, line 109: It seems to me that this test should test inlining behavior in isolation from CSE, for which we can turn off it using the compiler flag.

If you want to demonstrate an interplay between CSE/inlining - this could be done in a dedicated unit test.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -805,7 +806,13 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
         case GHC.collectArgs (strip e') of
           (strip -> GHC.Var f, args) ->
             case GHC.maybeUnfoldingTemplate (GHC.realIdUnfolding f) of
-              Nothing -> compileExpr e'
+              Nothing ->
+                case lookupName scope (GHC.getName f) of
+                  -- If `f` is locally bound, and its definition has already been compiled,
+                  -- we use it directly.
+                  -- This only supports `inline f`, not `inline (f x1 ... xn)`.
+                  Just (_var, Just def) | null args -> pure def
```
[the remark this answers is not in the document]
zliu41: Because it's inconvenient. `f` has already been compiled, but the arguments haven't. To compile the arguments, there are many cases to handle - an argument could be a term, or could be a type, or `f` could be `&&`, or `f` could be `fromString` etc etc.

plutus-tx-plugin/test/Inline/Spec.hs
zliu41: The pir already proves that the inlining works. Turning off CSE proves nothing further. Instead, this way, it shows that both inlining and CSE works as expected - two birds with one stone.

effectfully: I initially missed that part too, but now I see that @zliu41 is right.

effectfully (approved):
The whole design is kinda weird. Why can't `inline` just attach an annotation to the PIR term after compiling it, so that the PIR optimizer is left to worry about inlining? Here we do it in the plugin, which creates all the weirdness with `args` not being available and with the whole distinction of "unfolding available" vs "already compiled". If you just annotate the term and let the PIR inliner worry about it, you will not have any of these issues, no?

zliu41:
> Why can't inline just attach an annotation to the PIR term after compiling it, so that the PIR optimizer is left to worry about inlining?

This will require a redesign of how the inline hints work. Currently the hint is `ann -> name -> Bool` and looks at the annotation in the VarDecl to determine whether the var should be inlined. In other words, annotations and inline hints are not designed to make _callsite_ inlining decisions. But it can be done and I do agree it is probably a cleaner way to do this.

zliu41:
> I do agree it is probably a cleaner way to do this.

Having said that, I think it can be quite messy to preserve and propagate annotations on arbitrary terms, because the annotation will need to survive all the transformations and optimizations.

PR #6885, opened 2025-02-28
Fix AssocMap.Spec

zliu41:
map2 -> map3


zliu41:
merging since it is trivial

PR #6886, opened 2025-02-28
Propagate INLINE pragmas for local bindings

zliu41:
In https://github.com/IntersectMBO/plutus/pull/5183, the Plinth inliner was made to inline bindings with `INLINE` pragmas. However it only works for top-level bindings, not local bindings (at least not strict local bindings, which are compiled to `case` binders). This PR fixes it.

It basically comes down to setting and propagating the correct `binderAnn` in `compileCase`.

Also fixed a bug in the PIR inliner (PlutusIR.Transform.Inline.Inline): it was using the wrong `ann` for inline hints.


effectfully (approved):
Great stuff.

PR #6888, opened 2025-03-01
Avoid creating strict dead bindings for AsData

zliu41:
The AsData matchers, such as `AsData.Budget.Types.$mInts`, eagerly unpack all fields, even for unused fields. This leads to strict dead bindings. Since they are strict and the inliner doesn't know that they are effect-free, it can't remove them. It is not uncommon that only one or a small number of fields in `ScriptContext` are used, so this could cause severe performance penalties.

The matches are generated by GHC for pattern synonyms, not written by hand. As a result, they are unfortunately not customizable - for example, by attaching certain annotations to certain variables as signals for the Plinth compiler.

The workaround is the new wrapper functions in `PlutusTx.AsData.Internal`. `wrapUnsafeDataAsConstr` is marked NOINLINE so that GHC doesn't inline it, yet its unfolding is made available via `-fexpose-all-unfoldings`. `wrapUnsafeDataAsConstr` is supposed to be used by AsData only, and not elsewhere, and the Plinth compiler relies on it to identify which functions are AsData matchers.

An AsData matcher unpacks the fields and passes them to a continuation. What we need is thus to mark the variables bound in the continuation as safe to inline. For instance, if the continuation is `\x y z w -> f x` (indicating a data type with 4 fields, of which only the first is used), marking `x`, `y`, `z` and `w` as safe to inline allows the inliner to drop the unused ones. This is done in Expr.hs. A new annotation `SafeToInline` is introduced for this purpose.

This eliminates unused field accessors, but there are still the `tailList` dead bindings, which are intermediate artifacts in generating the field accessors. This is what `wrapTail` is for. `wrapTail` appears in `case` scrutinees, as in `case wrapTail @BuiltinData l of l' -> ...`. So the compiler detects such `case` expressions, and mark the binder, `l'`, as safe to inline.

This not only avoids creating dead bindings, but also allows the inliner to (unconditionally) inline the bindings that are used once.

As shown in the budget tests, `onlyUseFirstField-manual` is much cheaper than `onlyUseFirstField` (even with the additional `idx == 0` check!).

Note that this only works with the old code generator for `AsData`. The [new code generator](https://github.com/IntersectMBO/plutus/pull/6854) produces slightly more efficient code (before this optimization), but this optimization does not apply since it doesn't generate dead bindings (as can be seen in the current `onlyUseFirstField.pir.golden` on master). To make it work with the new code generator we'll need to first implement another optimization that floats bindings out of `case`. We should therefore revert to the old code generator (sans the index check for product types), until that optimization is implemented.


zliu41:
An example PIR. Before:

```
\(d-2066 : data) ->
  let
    !tup-2067 : pair integer (list data) = unConstrData d-2066
  in
  Bool_match-2062
    (ifThenElse
       {Bool-2059}
       (equalsInteger 0 (fstPair {integer} {list data} tup-2067))
       True-2060
       False-2061)
    {all dead-2068. integer}
    (/\dead-2069 ->
       let
         !l-2070 : list data = sndPair {integer} {list data} tup-2067
         !l-2071 : list data = tailList {data} l-2070
         !l-2072 : list data = tailList {data} l-2071
         !ds-2073 : integer = unIData (headList {data} l-2070)
         !ds-2074 : integer = unIData (headList {data} l-2071)
         !ds-2075 : integer = unIData (headList {data} l-2072)
         !ds-2076 : integer = unIData (headList {data} (tailList {data} l-2072))
       in
       ds-2073)
    (/\dead-2077 ->
       Unit_match-2065 (error {Unit-2063}) {integer} (error {integer}))
    {all dead-2078. dead-2078}
```

After:
```
\(d-1851 : data) ->
  let
    !tup-1852 : pair integer (list data) = unConstrData d-1851
  in
  Bool_match-1847
    (ifThenElse
       {Bool-1844}
       (equalsInteger 0 (fstPair {integer} {list data} tup-1852))
       True-1845
       False-1846)
    {all dead-1853. integer}
    (/\dead-1854 ->
       unIData (headList {data} (sndPair {integer} {list data} tup-1852)))
    (/\dead-1855 ->
       Unit_match-1850 (error {Unit-1848}) {integer} (error {integer}))
    {all dead-1856. dead-1856}
```

Unisay:
> The AsData matchers, such as AsData.Budget.Types.$mInts, eagerly unpack all fields, even for unused fields. This leads to strict dead bindings. Since they are strict and the inliner doesn't know that they are effect-free, it can't remove them.

I don't understand the connection between dead strict bindings and "inliner". I thought that dead bindings are supposed to be removed by `plutus-core/plutus-ir/src/PlutusIR/Transform/DeadCode.hs`.  Why are you saying that "inliner can't remove bindings"? Do you consider DCE as a part of inliner?

zliu41:
> Why are you saying that "inliner can't remove bindings"? Do you consider DCE as a part of inliner?

The inliners (both UPLC and PIR inliners) [do remove dead bindings](https://github.com/IntersectMBO/plutus/blob/master/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs#L345-L346) - since inlining a binding that is dead is equivalent to removing it.

But neither the inliners nor DCE can remove strict dead bindings that aren't known to be pure. E.g., it is wrong to remove `x` in `let !x = error in 42`.

zliu41:
@michaelpj can you also give this a review?

ana-pantilie (approved):
🙏

Unisay:
> The inliners (both UPLC and PIR inliners) [do remove dead bindings](https://github.com/IntersectMBO/plutus/blob/6411a87afb124e9d4ddc1af6da6f59d493de65fc/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs#L345-L346) - since inlining a binding that is dead is equivalent to removing it.

Good to know, thanks.

> But neither the inliners nor DCE can remove strict dead bindings that aren't known to be pure. E.g., it is wrong to remove `x` in `let !x = error in 42`.

So if I understand correctly, for you and me these accessors are known to be strict and pure but the Plutus compiler doesn't know about it. The solution is to add a purity hint (annotation) for compiler as soon as it is possible, such that DCE can safely remove them?

zliu41:
> So if I understand correctly, for you and me these accessors are known to be strict and pure but the Plutus compiler doesn't know about their purity (only strictness). 

Correct. `unsafeFromBuiltinData d` and `tailList xs` are in general not pure and can throw errors, but those generated from the `AsData` pattern synonyms are guaranteed to be pure. 

> Is the solution to add a purity hint (annotation) for compiler as soon as it is possible, such that DCE can safely remove them?

Yes - that's what the `SafeToInline` annotation does.

zliu41:
Merged to get this into the release - any further comments will be addressed in a follow-up.

plutus-tx-plugin/test/AsData/Budget/9.6/destructSum-manual.pir.golden
```diff
@@ -89,24 +89,21 @@ in
                           (y : integer)
                           (z : integer)
                           (w : integer) ->
-                           let
-                             !int1_ : integer = addInteger x x
```
michaelpj, line 93: Hmm, don't we know that this is pure anyway? I thought we knew that e.g. `addInteger` was pure?

michaelpj (commented):
Looks fine! The maching on the various bits of `AsData` is a bit clumsy, but I'm not sure of a nicer way to do it right now. Perhaps we could do something with custom GHC annotations? Not sure.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/Utils.hs
```diff
@@ -414,12 +416,15 @@ shouldUnconditionallyInline s n rhs body = preUnconditional ||^ postUnconditiona
     -- We actually also inline 'Dead' binders (i.e., remove dead code) here.
     preUnconditional = do
       isTermPure <- checkPurity rhs
-      nameUsedAtMostOnce n &&^ effectSafe body s n isTermPure
+      nameUsedAtMostOnce n &&^ (pure safe ||^ effectSafe body s n isTermPure)
```
effectfully, line 419: Hm, I don't think this is generally an improvement? `effectSafe` doesn't just check for effects, it also (somewhat implicitly) checks that the body "isn't doing any substantial work":

```haskell
    -- While we don't check for lambda etc like in the paper, `effectSafe` ensures that it
    -- isn't doing any substantial work.
```

With this change you may end up inlining something that does do substantial work. Although I guess not with this PR specifically. But it still may end up being disadvantageous.

plutus-tx-plugin/test/AsData/Budget/9.6/onlyUseFirstField.pir.golden
effectfully, line 1: Could you please add `onlyUseLastField` as well?

effectfully (commented):
That's one hell of an improvement.

PR #6893, opened 2025-03-03
Remove of irrefutable patterns makes script cheaper to evaluate.

Unisay:
Removing these lazy pattern modifiers leads to the lower execution budget.
Any reason to keep them? Why were they needed in the first place?


ana-pantilie:
> Removing these lazy pattern modifiers leads to the lower execution budget.
Any reason to keep them? Why were they needed in the first place?

The idea is that we want them to be inlined. If they're lazy, it's safe for the compiler to inline them because we're not changing the semantics of the program. If they were strict, the inliner couldn't safely inline them because it could potentially move the `error` to a different part of the program, maybe even one that doesn't get evaluated at all!

zliu41 (approved):
In this instance the `~` doesn't really matter - as you can see the two PIRs are almost identical - they just differ in the order of the bindings. The cost reduction here is more of a happenstance: the new order happens to allow the UPLC inliner to inline `ds` on line 92 (because although it is strict, it is `immediatelyEvaluated`).

We can merge this nonetheless.

PR #6894, opened 2025-03-03
filip(feat): apply new branding with regard to the new colouring/typography/logos discussed in branding guidelines

fstoqnov-iohk:


fstoqnov-iohk:
Note: as there was no official guideline for the shading colours - the footer colouring followed a similar pattern to the previous one - background is a couple of shades darker version of the primary brand colour and the links/text are in line with the new branding colouring as per the doc

PR #6902, opened 2025-03-04
Remove plugin support for GHC < 9.6

kwxm:
[CONTRIBUTING.adoc](https://github.com/IntersectMBO/plutus/blob/master/CONTRIBUTING.adoc#plugin-support) says

> We currently support the plugin on: - 9.6

This PR removes support for GHC versions < 9.6.

### DIFF
--- a/plutus-ledger-api/plutus-ledger-api.cabal
+++ b/plutus-ledger-api/plutus-ledger-api.cabal
@@ -47,6 +47,7 @@ common lang
     -fno-strictness
 
 common ghc-version-support
+  -- This is only used for the plugin tests.
   -- See the section on GHC versions in CONTRIBUTING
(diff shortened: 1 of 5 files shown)


plutus-ledger-api/plutus-ledger-api.cabal
```diff
@@ -47,6 +47,7 @@ common lang
     -fno-strictness
 
 common ghc-version-support
+  -- This is only used for the plugin tests.
```
kwxm, line 50: I was wondering why `plutus-ledger-api` depends on the plugin, but it's only for some tests, so I've added a reminder about that.

zliu41: Yes exactly.

zliu41 (approved):
Also remove the CPP extensions.

PR #6924, opened 2025-03-09
Fix typo on term flat decoder for builtin term

SeungheonOh:
Fixes typo on the spec, current spec doesn't wrap `b` with `\Builtin{}` from term flat decoder


kwxm:
Thanks @SeungheonOh.  Well spotted!

kwxm (approved):

kwxm:
I suspect there will be more things like that, so let us know if yoou notice anything else.

PR #6926, opened 2025-03-09
Reorganise bitwise tests a bit

kwxm:
There are two lots of tests for the bitwise operations (corresponding to two CIPs) and the naming was a bit confusing.  This just renames them.


plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Bitwise1.hs
```diff
@@ -2,7 +2,8 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TypeApplications  #-}
 
-module Evaluation.Builtins.Laws (
+-- | Tests for [CIP-0122](https://cips.cardano.org/cip/CIP-0122).
+module Evaluation.Builtins.Bitwise1 (
```
Unisay, line 6: Wouldn't it be more self-descriptive to name it `Evaluation.Builtins.Bitwise.CIP0122` and `Evaluation.Builtins.Bitwise.CIP0123` ?

kwxm: OK, I'll do that.  I thought about it and couldn't make up my mind, but you've persuaded me now.  I alos thought about putting them all in one file, but they're both pretty big already.

plutus-core/untyped-plutus-core/test/Evaluation/Builtins/Bitwise/CIP0123.hs
```diff
@@ -236,8 +236,8 @@ csbXor = property $ do
         ]
   evaluatesToConstant @Integer 0 rhs
 
--- | There should exist a monoid homomorphism between natural number addition and function composition for
--- shifts over a fixed bytestring argument.
+-- | There should exist a monoid homomorphism between natural number addition
+-- and function composition for shifts over a fixed bytestring argument.
```
Unisay, line 240: 🥇

PR #6939, opened 2025-03-11
Certifier: add the decision procedure for caseReduce into the certifier

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1472

The `inline` procedure is now commented out at the top-level because it's broken. https://github.com/IntersectMBO/plutus-private/issues/1475 tracks the work being done to fix it. After it's fixed, we can enable it as well.


zliu41 (approved):
Looks reasonable

PR #6941, opened 2025-03-11
Make `generate-cost-model` work

kwxm:
The`generate-cost-model` executable doesn't work in master because the costing code for the array builtins is incomplete.  If you try to runit you get this not-very-informative error:
```
generate-cost-model: cast: Dynamic type cast failed. Expected: String. Actual: Nil.
  error, called at src/Foreign/R/Internal.hs:130:5 in inline-r-1.0.1-IXV34TVMlU0ER3NHFf3CBO:Foreign.R.Internal
```
  
This PR adds enough code to make it work.  I had to makethe same changes in [this branch](https://github.com/IntersectMBO/plutus/tree/kwxm/costing/case-list-case-data) and [this branch](https://github.com/IntersectMBO/plutus/tree/kwxm/costing/expModInteger), and merging this PR will make it a lot easier to merge those without conflicts.  This will also form the basis for costing the array operations.  The steps that still need to be carried out are from [Step 5 of CostModelGeneration.md](https://github.com/IntersectMBO/plutus/blob/master/plutus-core/cost-model/CostModelGeneration.md#step-5-add-a-benchmark-for-the-new-builtin-and-run-it) onwards, although some of these have been partially done here to make the other branches work.


plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -176,5 +173,8 @@ builtinMemoryModels = BuiltinCostModelBase
   -- paramCaseList
   -- paramCaseData
   , paramDropList                        = Id $ ModelTwoArgumentsConstantCost 4
+  , paramLengthOfArray                   = Id $ ModelOneArgumentConstantCost    99
```
kwxm, line 176: I'm trying to keep everything in historical order here.  If stuff gets added in the middle of functions or lists it's easy to miss that it's happened.

Issue #6937, opened 2025-03-10
UPLC weird issue


nau:
### Summary

I have an old test that checks evaluation correspondance between Scalus and Plutus by evaluating the same UPLC via Scalus evaluator and `uplc evaluate`. 

While upgrading from 1.30.0.0 to 1.31.0.0 the test broke.
The issue is that 1.31.0.0 `uplc` tool evaluates the same script differently than 1.30.0.0 which should not be the case. Could you please help me figure out what's going on?

Here is the old Plutus V1 script:

```haskell
serialisedDeserialiserV1 = B16.encode . Serialise.serialise . fromCompiledCode $ $$(PlutusTx.compile [||unsafeFromBuiltinData @V1.ScriptContext||])
```

and then I apply a `toData` serialized `ScriptContext`. Here is a fully applied flat hex encoded script:
[...]

PR #6942, opened 2025-03-11
[Textual] Replace 'Simple' with 'Classic'

effectfully:
This PR replaces `Simple` with `Classic` in the executables, so that we print the correct program and people don't get horribly confused like in #6937.

Also removes a couple of spurious golden files and adds a couple of forgotten ones.


plutus-core/plutus-core/src/PlutusCore/Evaluation/ErrorWithCause.hs
```diff
@@ -54,7 +54,7 @@ instance (PrettyBy config cause, PrettyBy config err) =>
 
 instance (PrettyPlc cause, PrettyPlc err) =>
             Show (ErrorWithCause err cause) where
-    show = render . prettyPlcReadableSimple
+    show = render . prettyPlcReadable
```
effectfully, line 57: Not sure how useful this is, but I think it's potentially more useful than noisy.

plutus-core/executables/src/PlutusCore/Executable/Parsers.hs
```diff
@@ -100,7 +100,7 @@ printmode :: Parser PrintMode
 printmode = option auto
   (  long "print-mode"
   <> metavar "MODE"
-  <> value Simple
+  <> value Classic
   <> showDefault
   <> help
     ("Print mode for textual output (ignored elsewhere): Classic -> plcPrettyClassic, "
```
kwxm, line 106: I'm still confused about the names and I'm too lazy to look at the code.  I think that (maybe) `evaluate` always prints the AST that it gets back from the CEK machine, and that has de Bruijn indices (although it may or may not showthem to you depending on the print mode).  I think the confusing thing is that we also have `convert` which lets you change the name type, but you can only do that if you output flat, and then maybe you can convert that to textual form.  Do we have any way of converting the output of a program to a textual form that you can sucessfully parse again?

Also, did we ever talk about distinguishing indices from uniques in textual output?  Maybe part of the confusion is that they look exactly the same.

Unisay: > Also, did we ever talk about distinguishing indices from uniques in textual output? Maybe part of the confusion is that they look exactly the same.

DeBruijn indexes [use `!` as a separator,](https://github.com/IntersectMBO/plutus/blob/66b240239e1afd8ef8a72791c3233300745db0b8/plutus-core/plutus-core/src/PlutusCore/DeBruijn/Internal.hs#L174) while [uniques use `-`.](https://github.com/IntersectMBO/plutus/blob/66b240239e1afd8ef8a72791c3233300745db0b8/plutus-core/plutus-core/src/PlutusCore/Name/Unique.hs#L109-L110)

PR #6947, opened 2025-03-12
[Evaluation] [Performance] Use the strict 'List' for 'Constr'

effectfully:
Just checking.


effectfully:
That does look like an improvement:

image

effectfully:
OK, `mergeSort` is genuinely ~2% faster. `sum-left-Scott` appears to be faster too, but less than that.

effectfully:
Meh, don't wanna resolve these

image

PR #6948, opened 2025-03-13
Remove parameter names for array-related builtin functions

Unisay:
As per the specification:

> Currently, once we add a feature for any given protocol version/ledger language, we also make it
available for all subsequent protocol versions/ledger languages. For example, Batch 2 of builtins was
introduced in PlutusV2 at protocol version 7.0, so it is also available in PlutusV2 at protocol versions
after 7.0, and PlutusV3 at protocol versions after 9.0 (when PlutusV3 itself was first introduced).

### DIFF
--- a/plutus-ledger-api/src/PlutusLedgerApi/V2/ParamName.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V2/ParamName.hs
@@ -200,11 +200,5 @@ data ParamName =
   | ByteStringToInteger'cpu'arguments'c2
   | ByteStringToInteger'memory'arguments'intercept
   | ByteStringToInteger'memory'arguments'slope
-  | LengthOfArray'cpu'arguments
-  | LengthOfArray'memory'arguments
-  | ListToArray'cpu'arguments
-  | ListToArray'memory'arguments
-  | IndexArray'cpu'arguments
-  | IndexArray'memory'arguments
     deriving stock (Eq, Ord, Enum, Ix, Bounded, Generic)
     deriving IsParamName via (GenericParamName ParamName)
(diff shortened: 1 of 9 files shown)


plutus-ledger-api/src/PlutusLedgerApi/V2/ParamName.hs
```diff
@@ -200,11 +200,5 @@ data ParamName =
   | ByteStringToInteger'cpu'arguments'c2
   | ByteStringToInteger'memory'arguments'intercept
   | ByteStringToInteger'memory'arguments'slope
-  | LengthOfArray'cpu'arguments
```
kwxm, line 203: I can't comment in V3/ParamName.hs, but not that we do need these parameters there, and probably at least the entries for `ListToArray` will need to be changed once we konw what the costing function really is.

kwxm (approved):
This looks good.  I see that [here](https://github.com/IntersectMBO/plutus/blob/aef868cdee40687f7a0e3b5c5b27eb1220f3d12d/plutus-ledger-api/testlib/PlutusLedgerApi/Test/V3/EvaluationContext.hs#L96) and [here](https://github.com/IntersectMBO/plutus/blob/aef868cdee40687f7a0e3b5c5b27eb1220f3d12d/plutus-ledger-api/testlib/PlutusLedgerApi/Test/V3/Data/EvaluationContext.hs#L94) there are functions that delete new builtins from the cost model parameters, supposedly in order to make some tests work. However the tests pass even though they don't mention the array functions, so it's probably OK not to change those functions.  I wrote that code but I don't remember exactly why it was needed now.  The `plutus-ledger-api` tests [need to be fixed](https://github.com/IntersectMBO/plutus/issues/6089) anyway, since some of the invariants that they're checking are no longer valid.

Update: I think the parameters need to be removed from `V3.Paramname` as well.  The parameter names are used to update the cost model when the node supplies a new set of cost model parameters (in the fomr of a list of integers) and it doesn't know about the array operations yet.

kwxm (changes requested):
The array functions need to be removed from `V3.ParamNames`.

kwxm (approved):
Good, I think this is what we want.

PR #6949, opened 2025-03-13
Make the Git revision available to pir/uplc/plc programs

zeme-wana:
With this PR, the executables pir/uplc/plc/plutus get a new `--version` option which prints version information like so: 
```
> pir --version
1.42.0.0 - git rev 10aa180855d92307e16c2d3b5f868daf5b6310fc - 2025-03-13T11:50:30Z

> plc --version
1.42.0.0 - git rev 10aa180855d92307e16c2d3b5f868daf5b6310fc - 2025-03-13T11:50:30Z

> uplc --version
1.42.0.0 - git rev 10aa180855d92307e16c2d3b5f868daf5b6310fc - 2025-03-13T11:50:30Z

> plutus --version
1.42.0.0 - git rev 10aa180855d92307e16c2d3b5f868daf5b6310fc - 2025-03-13T11:50:30Z
```


plutus-core/plutus-core.cabal
```diff
@@ -88,6 +88,8 @@ library
     Data.Either.Extras
     Data.List.Extras
     Data.MultiSet.Lens
+    Data.Version.Extras
+    GitRev.Extras
```
zeme-wana, line 92: This seemed like the right place to place these modules?

plutus-executables/plutus-executables.cabal
```diff
@@ -76,6 +76,7 @@ library lib
 executable pir
   import:         lang
   main-is:        pir/Main.hs
+  other-modules:  Paths_plutus_executables
```
zeme-wana, line 79: Needed to get the version from the cabal file

plutus-core/plutus-core.cabal
```diff
@@ -311,6 +313,7 @@ library
     , flat                        ^>=0.6
     , free
     , ghc-prim
+    , gitrev
```
zeme-wana, line 316: This provides the `Development.GitRev` module

plutus-core/plutus-core.cabal
```diff
@@ -88,6 +88,8 @@ library
     Data.Either.Extras
     Data.List.Extras
     Data.MultiSet.Lens
+    Data.Version.Extras
+    Development.GitRev.Extras
```
zeme-wana, line 92: This seemed like the right place to place these modules?

Unisay: Maybe `plutus-executables.cabal` is more appropriate.

Unisay: It doesn't look like these modules have anything `plutus-core` specific, I'd create a separate package. This way it could be added as a dependency in other packages, that doesn't really need to depend on `plutus-core`.

nix/shell.nix
```diff
@@ -122,6 +122,11 @@ let
   quick-shell = project.shellFor {
     name = "plutus-shell-${project.args.compiler-nix-name}";
     tools = { cabal = "latest"; };
+    shellHook = ''
```
kwxm, line 125: What's this doing?  I tried restarting the git shell in this branch and got "🤟 Welcome to Plutus 🤟" but got my own shell prompt (which is what I want).  I had some trouble overriding a prompt setting before, but it seems to be OK in this case.  Can I just pretend that this doesn't change PS1?

plutus-core/plutus-core/src/Development/GitRev/Extras.hs
```diff
@@ -0,0 +1,122 @@
...
+Usage:
+
+@
+module MyPrograms where
+
+import Development.GitRev.Extras (gitHash, gitCommitDate)
+
+main :: IO ()
+main = do
+  -- Falls back to reading the GIT_BRANCH env var
```
kwxm, line 21: Where would these shell variables get set?  They don't exist in my shell.

plutus-executables/executables/pir/Main.hs
```diff
@@ -281,6 +284,10 @@ runPrint (PrintOptions inp outp mode) = do
             StdOutput       -> putStrLn printed
             NoOutput        -> pure ()
 
+
```
kwxm, line 287: The code for these programs is (rather oddly) split between `plutus-executables` and `plutus-core`.  Can we put a single version of this code in `plutus-core/executables/src/PlutusCore/Executable/Common.hs` and import it from there?  Maybe there are TH complications that would make that infeasible though.

kwxm (approved):
This is great, and it does exactly what we wanted, so thanks for doing it!  As @Unisay says though, it's way more complicated than I expected.

I wonder if we can also get version information into the plugin as well?  I don't know if that's possible, or how you'd get the plugin to report its version to you.  We could maybe think about that separately.  @zliu41 : any ideas?

nix/shell.nix
zeme-wana: this doen't change the default shell, it changes other shells (ghc910, ghc98, etc) You can pretend that it doesn't change PS1

plutus-core/plutus-core/src/Development/GitRev/Extras.hs
zeme-wana: Anywhere you want. We set them in Nix code, for example. 
Because they don't exist in your shell, `git` will be used instead

plutus-executables/executables/pir/Main.hs
zeme-wana: Sure, I feel like this belongs to a separate PR though

effectfully (approved):
Great work, thanks a lot for doing it.

effectfully:
Oh, and I'm very sorry for the overly late PR review.

PR #6950, opened 2025-03-13
Costing of BuiltinArray functions

Unisay:
Closes https://github.com/IntersectMBO/plutus-private/issues/1375
Benchmark run: https://github.com/IntersectMBO/plutus/actions/runs/13907860876

image
image
image 
(^ not zoomed out)
image


plutus-core/cost-model/budgeting-bench/Common.hs
```diff
@@ -84,7 +84,9 @@ benchWith
 -- the result, so e.g. 'evaluateCek' won't work properly because it returns a pair whose components
 -- won't be evaluated by 'whnf'. We can't use 'nf' because it does too much work: for instance if it
 -- gets back a 'Data' value it'll traverse all of it.
-benchWith params name term = bench name $ whnf (evaluateCekNoEmit params) term
+benchWith params name term =
+  bench name $
+    whnf (unsafeSplitStructuralOperational . evaluateCekNoEmit params) term
```
Unisay, line 89: This makes a benchmark fail if evaluation fails (wasn't the case before)

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -173,8 +173,8 @@ builtinMemoryModels = BuiltinCostModelBase
   -- paramCaseList
   -- paramCaseData
   , paramDropList                        = Id $ ModelTwoArgumentsConstantCost 4
-  , paramLengthOfArray                   = Id $ ModelOneArgumentConstantCost    99
-  , paramListToArray                     = Id $ ModelOneArgumentConstantCost    99
-  , paramIndexArray                      = Id $ ModelTwoArgumentsConstantCost   99
+  , paramLengthOfArray                   = Id $ ModelOneArgumentConstantCost 10
+  , paramListToArray                     = Id $ ModelOneArgumentLinearInX identityFunction
+  , paramIndexArray                      = Id $ ModelTwoArgumentsConstantCost 32
```
Unisay, line 178: @kwxm `10` and `32` here are `magic` numbers I chose because they were chose in similar builtin functions. This is not based on any measurements I did whatsoever.

kwxm: I'm not sure what we should put for the memory costs.  I guess for `ListToArray` we'll need something with a non-zero constant term to account for some header data and then one pointer's worth of memory for each entry.  @effectfully Do you remember how much room a pointer takes up?

For `LengthOfArray` we'll need enough memory to hold one integer, so maybe a constant cost of 1 would do, and for `IndexArray` it'll again be one pointer's worth.

But then again, the current numbers are pretty similar to those for other things like `LengthOfByteString`, so maybe these are OK.  We charge 100 memory units for each CEK step anyway, so that dominates the memory cost and and extra 10 or 20 bytes for a builtin won't make a lot of difference.

plutus-core/cost-model/budgeting-bench/Common.hs
kwxm: Maybe you could add a comment to remind us why `unsafeSplitStructuralOperational` is there, since it's not immediately obvious what it's doing.

plutus-core/cost-model/budgeting-bench/Benchmarks/Arrays.hs
```diff
@@ -0,0 +1,74 @@
...
+makeBenchmarks :: StdGen -> [Benchmark]
+makeBenchmarks gen =
+  [ benchLengthOfArray gen
+  , benchListToArray gen
+  , benchIndexArray gen
+  ]
+
+benchLengthOfArray :: StdGen -> Benchmark
+benchLengthOfArray gen =
+  createOneTermBuiltinBench LengthOfArray [tyArrayOfInteger] listOfArrays
```
kwxm, line 31: I think we need to use the `create<N>TermBudgetingBenchWithWrappers` here to wrap the array arguments, like [this](https://github.com/IntersectMBO/plutus/blob/a48ebd739c1da07bffa467c54445945e5412b472/plutus-core/cost-model/budgeting-bench/Benchmarks/Lists.hs#L119), although that will only really matter for `listToArray` since the others are constant time.  We have to use the same size measures here as are used in the denotations.

Try adding the wrappers and re-running the benchmarks and see if the cost of `listToArray` changes significantly.  If it does, you'll have to rebuild the cost model with the new results.

kwxm: I'm thinking that we should probably make the costed-by-length measure the default one and maybe do the same for lists (although maybe then we'll need a wrapper that measures the entire memory footprint for some of the other things involving lists).  I'll make an issue to look at that separately.

kwxm (changes requested):
This looks mostly OK, although I think that the costing benchmarks need to have `ArrayCostedByLength` wrappers.  Apart from that it looks fine.

plutus-core/cost-model/budgeting-bench/Benchmarks/Arrays.hs
Unisay: > Try adding the wrappers and re-running the benchmarks and see if the cost of `listToArray` changes significantly.

Done. 

Without the `ListCostedByLength` wrapper.
image

Wrapped
image

kwxm: This gives very similar results to the unwrapped version, but that's because the inputs are lists containing integers of size 1, so the "deep" size is pretty much the same as the length.  This should be more robust though, since the results shouldn't change if we were to use lists of `data` say, whereas the previous version would have made `listToArray` cheaper because you could handle apparently larger inputs in the same amount of time.

plutus-core/cost-model/budgeting-bench/Benchmarks/Arrays.hs
```diff
@@ -0,0 +1,81 @@
...
```
kwxm, line 33: This should use the same wrappper as the denotation.  It won't make any difference because it's constant time, but let's keep things consistent anyway.
```suggestion
  createOneTermBuiltinBenchWithWrapper
    ArrayCostedByLength
    LengthOfArray 
    [tyArrayOfInteger] 
    listOfArrays
```

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -2111,25 +2112,30 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
             (runCostingFunTwoArguments . paramDropList)
 
     toBuiltinMeaning _semvar LengthOfArray =
-      let lengthOfArrayDenotation :: SomeConstant uni (Vector a) -> BuiltinResult Int
+      let lengthOfArrayDenotation :: SomeConstant uni (ArrayCostedByLength a) -> BuiltinResult Int
```
kwxm, line 2115: This may look as if the costing function has to measure the length of the array in order to work out the cost of calling the builtin that measures the length of the array, which might seem a bit silly.  However, if the costing function is constant I think the size measure will never actually be evaluated so it's kind of sane after all.

kwxm (approved):
Looks good!  I've added a suggestion to make the costing benchmark for `LengthOfArray` use an `ArrayCostedByLength` wrapper, so please check that I got that right and if it is please commit the suggestion and merge the PR.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
effectfully: > @effectfully Do you remember how much room a pointer takes up?

One machine word IIRC.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
effectfully: Why make it `ArrayCostedByLength` then, why not just `Vector`?

kwxm:
@yura You forgot to update the conformance tests to have the correct budgets.  I've done that with `--accept` and pushed the changes.

kwxm (approved):
Success!  This all looks perfectly sensible.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -173,8 +173,8 @@ builtinMemoryModels = BuiltinCostModelBase
   -- paramCaseList
   -- paramCaseData
   , paramDropList                        = Id $ ModelTwoArgumentsConstantCost 4
-  , paramLengthOfArray                   = Id $ ModelOneArgumentConstantCost    99
-  , paramListToArray                     = Id $ ModelOneArgumentConstantCost    99
-  , paramIndexArray                      = Id $ ModelTwoArgumentsConstantCost   99
+  , paramLengthOfArray                   = Id $ ModelOneArgumentConstantCost 10
+  , paramListToArray                     = Id $ ModelOneArgumentLinearInX identityFunction
```
kwxm, line 177:
```suggestion
  , paramListToArray                     = Id $ ModelOneArgumentLinearInX $ OneVariableLinearFunction 7 1
```

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
kwxm: I found [this page](https://wiki.haskell.org/GHC/Memory_Footprint), which says that the memory footprint of a vector is (7+N) words (plus the size of the contents, which we don't care about since they're already on the heap.  Based on that I suggestthe following:

 * Change the memory cost of `listToArray` to `Id $ ModelOneArgumentLinearInX $ OneVariableLinearFunction 7 1`
  * In the three json files, change the  memory model for `listToArray` to match that (this'll save you from having to regenerate the entire cost model):
```
     "memory": {
            "arguments": {
                "intercept": 7,
                "slope": 1
            },
            "type": "linear_in_x"
        }
  ```
   * Update the conformance results.

kwxm (commented):
I had a closer look at the memory usage for `listToArray` and I've suggested some small chaages based on that.  If you fix those I think we should be ready to merge.

kwxm:
@Unisay As a precaution I changed the benchmarks for `listToArray` to use lists of various lengths just containing the integer 12 and re-ran them.  Here are the results (red points) compared with your results for lists of bytestrings from yesterday (blue points).  They look very similar, so the problem we had earlier has definitely been fixed by changing the defautl memory usage for lists to be their length.
ListToArray4

kwxm (approved):
Looks good.  Go for it!

PR #6953, opened 2025-03-14
Fix incomplete case pattern warning in plutus-metatheory/src/Raw.hs

zeme-wana:


plutus-metatheory/src/Raw.hs
```diff
@@ -98,7 +98,7 @@ convTyCon (SomeTypeIn DefaultUniBLS12_381_G2_Element) = RTyCon (RTyConAtom ATyCo
 convTyCon (SomeTypeIn DefaultUniBLS12_381_MlResult)   = RTyCon (RTyConAtom ATyConBLS12_381_MlResult)
 convTyCon (SomeTypeIn DefaultUniProtoList)            = RTyCon RTyConList
 convTyCon (SomeTypeIn DefaultUniProtoPair)            = RTyCon RTyConPair
-convTyCon (SomeTypeIn (DefaultUniApply _ _))          = error "unsupported builtin type application"
+convTyCon _                                           = error "unsupported builtin type application"
```
ana-pantilie, line 101:
```suggestion
convTyCon (SomeTypeIn (DefaultUniApply x _)) = error "unsupported builtin type application"
convTyCon (SomeTypeIn DefaultUniProtoArray) = error "unsupported builtin array"

```

ana-pantilie (approved):
I think we should explicitly enumerate the cases here, because this just showed that we haven't implemented the `Array` builtin type into the metatheory. @ramsay-t is there an issue for this?

plutus-metatheory/src/Raw.hs
ramsay-t: We should make an issue for it.

ramsay-t (approved):
I agree with @ana-pantilie - we should make sure the missing cases are added explicitly so it warns us if we miss more in the future.

PR #6951, opened 2025-03-13
Use `-Werror` in GitHub CI

neilmayhew:

PR #6955, opened 2025-03-15
[Build] Add '-Werror' and make it work in GHC-8.10

effectfully:
An alternative version of #6952 that replaces `OPAQUE` with `NOINLINE` where it matters and adds `{-# OPTIONS_GHC -Wwarn=unrecognised-pragmas #-}` where it doesn't.

Required for integration into `cardano-node-10`.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
@@ -164,8 +164,9 @@ We mark error-related definitions such as prisms like '_StructuralUnliftingError
 functions like 'throwNotAConstant' with @INLINE@, because this produces significantly less cluttered
 GHC Core. Not doing so results in 20+% larger Core for builtins.
 
-However in a few specific cases we use @NOINLINE@ instead to get tighter Core. @NOINLINE@ is the same as
-@NOINLINE@ except the former _actually_ prevents GHC from inlining the definition unlike the latter.
+However in a few specific cases we use @NOINLINE@ instead to get tighter Core. @NOINLINE@ is the
+same as @OPAQUE@ except the latter _actually_ prevents GHC from inlining the definition unlike the
+former (we use the former because we currently have to support GHC-8.10).
 See this for details: https://github.com/ghc-proposals/ghc-proposals/blob/5577fd008924de8d89cfa9855fa454512e7dcc75/proposals/0415-opaque-pragma.rst
 
 It's hard to predict where @NOINLINE@ instead of @INLINE@ will help to make GHC Core tidier, so it's
--- a/plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
@@ -159,16 +159,16 @@ instance Pretty BuiltinError where
     pretty (BuiltinUnliftingEvaluationError err) = "Builtin evaluation failure:" <+> pretty err
     pretty BuiltinEvaluationFailure              = "Builtin evaluation failure"
 
-{- Note [INLINE and OPAQUE on error-related definitions]
+{- Note [INLINE and NOINLINE on error-related definitions]
 We mark error-related definitions such as prisms like '_StructuralUnliftingError' and regular
 functions like 'throwNotAConstant' with @INLINE@, because this produces significantly less cluttered
 GHC Core. Not doing so results in 20+% larger Core for builtins.
 
-However in a few specific cases we use @OPAQUE@ instead to get tighter Core. @OPAQUE@ is the same as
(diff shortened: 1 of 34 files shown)


Unisay:
> replaces `OPAQUE` with `NOINLINE` where it matters

What is your idea of a package, where s/OPAQUE/NOINLINE/ matters?

I thought that it only matters for packages used by node that must be compilable by GHC 8.10.7 without warnings, but it doesn't look like plutus-benchmark is such a package...

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -1,15 +1,17 @@
...
 {-# LANGUAGE StandaloneKindSignatures #-}
 {-# LANGUAGE TypeFamilies             #-}
 {-# LANGUAGE TypeOperators            #-}
 {-# LANGUAGE UndecidableInstances     #-}
+
 {-# OPTIONS_GHC -Wno-orphans #-}
 {-# OPTIONS_GHC -fno-omit-interface-pragmas #-}
 {-# OPTIONS_GHC -fno-specialise #-}
-{-# LANGUAGE DerivingStrategies       #-}
+{-# OPTIONS_GHC -Wwarn=unrecognised-pragmas #-}
```
neilmayhew, line 14: Maybe wrap this in `#if !MIN_VERSION_base(4, 15, 0)` here and in other modules that already use `LANGUAGE CPP`.

That way, it's easier to spot what to delete when we drop support for ghc 8. (It could probably be automated with a `sed` script, in fact.)

neilmayhew (changes requested):

neilmayhew:
I suggest you add [this commit](https://github.com/IntersectMBO/plutus/pull/6951/commits/cb27f47a12226a3eb27bf97ce8009b9302e2ef87) from #6951 as well

plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
```diff
@@ -159,16 +159,16 @@ instance Pretty BuiltinError where
...
-{- Note [INLINE and OPAQUE on error-related definitions]
+{- Note [INLINE and NOINLINE on error-related definitions]
 We mark error-related definitions such as prisms like '_StructuralUnliftingError' and regular
 functions like 'throwNotAConstant' with @INLINE@, because this produces significantly less cluttered
 GHC Core. Not doing so results in 20+% larger Core for builtins.
 
-However in a few specific cases we use @OPAQUE@ instead to get tighter Core. @OPAQUE@ is the same as
+However in a few specific cases we use @NOINLINE@ instead to get tighter Core. @NOINLINE@ is the same as
 @NOINLINE@ except the former _actually_ prevents GHC from inlining the definition unlike the latter.
 See this for details: https://github.com/ghc-proposals/ghc-proposals/blob/5577fd008924de8d89cfa9855fa454512e7dcc75/proposals/0415-opaque-pragma.rst
```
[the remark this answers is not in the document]
effectfully: Fixed the comment, thanks.

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
effectfully: Done, thanks.

effectfully: (for all modules, not just those where we have `CPP` already, I added it where it was required)

effectfully:
> I suggest you add [this commit](https://github.com/IntersectMBO/plutus/pull/6951/commits/cb27f47a12226a3eb27bf97ce8009b9302e2ef87) from https://github.com/IntersectMBO/plutus/pull/6951 as well

Done, thanks.

effectfully:
@Unisay 

> What is your idea of a package, where s/OPAQUE/NOINLINE/ matters?

> I thought that it only matters for packages used by node that must be compilable by GHC 8.10.7 without warnings, but it doesn't look like plutus-benchmark is such a package...

It's ideal to build all the packages with `-Werror`, so I made them all buildable with it. I used `NOINLINE` whenever it's safe and it's only not safe to use it when the plugin looks for a very specific name, hence `OPAQUE` in Plinth code.

@zliu41 

> Can you verify this on 9.8? If it fails the build, consider removing core lint in this module.

I'm not sure if I'm doing something wrong, but I can't reproduce. Where did you get that error from?

effectfully:
> I'm not sure if I'm doing something wrong, but I can't reproduce. Where did you get that error from?

Also, this PR fixes issue for GHC-8.10. I'd suggest not blocking it on GHC-9.8, since we need this in for the release. I'm happy to create a follow-up if there's an issue with GHC-9.8.

zliu41 (approved):

PR #6959, opened 2025-03-17
Add ScriptContext test

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1479

The `AsData` version is actually slightly better (in script size) than the manually optimised version! I am not entirely sure why, since I tried multiple versions of the manually optimised one, and this was the best one.


plutus-benchmark/script-contexts/src/PlutusBenchmark/V2/Data/ScriptContexts.hs
```diff
@@ -1,7 +1,9 @@
 {-# LANGUAGE BangPatterns      #-}
 {-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE NamedFieldPuns    #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE TypeApplications  #-}
```
ana-pantilie, line 6: I don't understand why Data-backed Marlowe is slightly faster now. The only relationship I can think of is that maybe it uses code from this module and adding these two extensions somehow made the compiler generate code which is slightly more optimal?

ana-pantilie: After talking to @zliu41, and seeing that other tests have also been impacted, it's most likely from the two `member'` and `insert'` functions being exported, but it remains to be seen what exactly is going on with the Haskell Core. Anyway, we're lucky that the performance changes work in our favor.

plutus-benchmark/script-contexts/src/PlutusBenchmark/V2/ScriptContexts.hs
```diff
@@ -139,3 +192,34 @@ mkScriptContextEqualityOverheadCode sc =
   in $$(PlutusTx.compile [|| scriptContextEqualityOverhead ||])
     `PlutusTx.unsafeApplyCode` PlutusTx.liftCodeDef sc
     `PlutusTx.unsafeApplyCode` PlutusTx.liftCodeDef d
+
+forwardWithStakeTrick :: BuiltinData -> BuiltinData -> ()
```
zliu41, line 196: Add some comments - it's hard to tell from the name - forwardWithStakeTrick - what it does.

plutus-benchmark/script-contexts/test/V2/9.6/dataFwdStakeTrick.pir.golden
```diff
@@ -0,0 +1,231 @@
...
+             {all dead. Unit}
+             (/\dead -> Unit)
+             (/\dead ->
+                let
+                  !x : Unit = trace {Unit} "not found" Unit
+                in
+                error {Unit})
+             {all dead. dead})
+        {all dead. dead})
+  (Constr 0 [Constr 0 [B #736f6d6543726564656e7469616c]])
```
zliu41, line 93: Don't include the script context in the golden files - it serves no purpose.

plutus-benchmark/script-contexts/src/PlutusBenchmark/V2/Data/ScriptContexts.hs
```diff
@@ -140,3 +195,82 @@ mkScriptContextEqualityOverheadCode sc =
...
+      let txInfoWdrl' = Map.toBuiltinList txInfoWdrl
+          wdrlAtZero = BI.fst $ BI.head txInfoWdrl'
+          rest = BI.tail txInfoWdrl'
+          wdrlAtOne = BI.fst $ BI.head rest
+      in
+        if PlutusTx.equalsData obsScriptCred wdrlAtZero
+          || PlutusTx.equalsData obsScriptCred wdrlAtOne
+          then ()
+          else
+            if Map.member' obsScriptCred rest
```
zliu41, line 212: I don't think you need `member'` here - you can use `PlutusTx.Data.List.elem` (or something very similar).

Even if you do need to expose `member'`, it should be exposed from the List module (since it has nothing to do with Map), and it should have Haddock.

zliu41 (changes requested):

plutus-benchmark/script-contexts/src/PlutusBenchmark/V2/Data/ScriptContexts.hs
```diff
@@ -140,3 +195,90 @@ mkScriptContextEqualityOverheadCode sc =
...
+      let txInfoWdrl' = Map.toBuiltinList txInfoWdrl
+          wdrlAtZero = BI.fst $ BI.head txInfoWdrl'
+          rest = BI.tail txInfoWdrl'
+          wdrlAtOne = BI.fst $ BI.head rest
+      in
+        if PlutusTx.equalsData obsScriptCred wdrlAtZero
+          || PlutusTx.equalsData obsScriptCred wdrlAtOne
+          then ()
+          else
+            if Map.member obsScriptCred (Map.unsafeFromBuiltinList rest)
```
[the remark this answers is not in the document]
ana-pantilie: > Can we convert it into a PlutusTx.Data.List, without losing efficiency?

No. `Map` is `BuiltinList (BuiltinPair BuiltinData BuiltinData)` and `List` is `BuiltinList BuiltinData`. We would need to traverse the `BuiltinList` and transform each `BuiltinPair` into a `BuiltinData`.

> If BuiltinList is indeed more efficient, then it suggests that the Plinth library should have some utilities for working with BuiltinList - such as elem, find, lookup and so on.

Would we use `BuiltinList` in the `ScriptContext` directly?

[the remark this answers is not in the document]
zliu41: No, but once you convert Map to BuiltinList, having to convert it back to a Map in order to call `member` or whatever, is awkward. Not to mention after some operations on the BuiltinList, it may no longer be a Map (and indeed, converting it back requires an `unsafe` operation).

It would be much nicer if there's a `lookup` operation and similar operations on BuiltinList, so once you have the list you just work with the list.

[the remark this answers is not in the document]
zliu41: Could you open an issue for adding BuiltinList operations.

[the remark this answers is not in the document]
ana-pantilie: I think you're right. Opened an issue: https://github.com/IntersectMBO/plutus/issues/6977

zliu41 (approved):

PR #6963, opened 2025-03-20
Fix: `Use of fromString @BuiltinByteStringHex with inscrutable content`

Unisay:
Closes https://github.com/IntersectMBO/plutus-private/issues/1496

### DIFF
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
@@ -76,6 +76,7 @@ import Control.Monad.Reader (ask, asks, local)
 import Data.Array qualified as Array
 import Data.ByteString qualified as BS
 import Data.ByteString.Base16 qualified as Base16
+import Data.ByteString.Char8 qualified as BSC
 import Data.Generics.Uniplate.Data (transform, universeBi)
 import Data.List (elemIndex, isPrefixOf, isSuffixOf)
 import Data.Map qualified as Map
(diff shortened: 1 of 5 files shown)


plutus-tx-plugin/test/ByteStringLiterals/Lib.hs
```diff
@@ -0,0 +1,11 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -fplugin PlutusTx.Plugin #-}
```
Unisay, line 3: Peculiar fact: with this compiler option the error is:
```
    GHC Core to PLC plugin: Error: 
    Unexpected error during compilation, please report this to the Plutus team:
    Use of fromString @PlutusTx.Builtins.HasOpaque.BuiltinByteStringHex with inscrutable content: 
    GHC.Types.: @GHC.Types.Char 
      (GHC.Types.C# 'f'#) 
      (GHC.Base.build @GHC.Types.Char (\ (@b) -> GHC.CString.unpackFoldrCString# @b "0"#))
```
but without it the error is different (even more confusing):
```
GHC Core to PLC plugin: Error: Unsupported feature: 
Literal string (maybe you need to use OverloadedStrings)
Context: Compiling expr: "f0"#
Context: Compiling expr: GHC.Base.build
                           @GHC.Types.Char
                           (\ (@b) -> GHC.CString.unpackFoldrCString# @b "f0"#)
```

zliu41: There's no reason to put `{-# OPTIONS_GHC -fplugin PlutusTx.Plugin #-}` in this module. 
@Unisay can you post the entire GHC core for this module?

Unisay: Without this pragma the GHC's recompilation avoidance mechanism would not recompile this module whenever something in the plugin changes; For this reason I always add it in tests.

However, my point was different: the presence/absense of this pragrma does change the compilation error, while my expectation was that it shouldn't make any difference.

> can you post the entire GHC core for this module?
Sure.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -149,14 +167,44 @@ tryStringLiteralAsBytes coreExpr = case coreExpr of
...
+    `GHC.App` GHC.Lam _ (GHC.Var unpack `GHC.App` _ `GHC.App` expr)
+      | GHC.getName build == GHC.buildName
+      , GHC.getName unpack == GHC.unpackCStringFoldrName
+      , GHC.getName charTyCon == GHC.charTyConName ->
+        tryStringLiteralAsBytes expr
+  {-
+  Example GHC Core expr this pattern matches:
+    GHC.Types.: @GHC.Types.Char (GHC.Types.C# 'f'#) expr
+  -}
+  ( GHC.Var consId
```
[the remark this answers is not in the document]
Unisay: Well spotted! Fixed.

PR #6964, opened 2025-03-20
Updated Pure definition in the metatheory

ramsay-t:
The previous Pure definition was just a stub. This works, although there are some differences from the Haskell and some open questions about some of the details.


ramsay-t:
Currently this definition of Purity also causes the certifier to say "no" to some of the tests. We should, obviously, fix that before we merge this.

plutus-metatheory/src/VerifiedCompilation/Purity.lagda.md
```diff
@@ -12,14 +12,131 @@ module VerifiedCompilation.Purity where
...
+
+sat-det : {X : Set} {arity args arity₁ args₁ : ℕ} {t t₁ : X ⊢}
+        → saturation t ≡ just (arity , args)
+        → saturation t₁ ≡ just (arity₁ , args₁)
+        → t ≡ t₁
+        → just (arity , args) ≡ just (arity₁ , args₁)
+sat-det sat-t sat-t₁ refl = trans (sym sat-t) sat-t₁
+
+data Pure {X : Set} : (X ⊢) → Set where
+    force : {t : X ⊢} → Pure t → Pure (force t)
```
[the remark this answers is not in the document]
ramsay-t: Did we agree that we only force delays (and builtins), so we don't need to handle force applied to arbitrary things?

plutus-metatheory/src/Untyped/Purity.lagda.md
```diff
@@ -0,0 +1,146 @@
...
+           → lookup? i ts ≡ just t
+           → Pure (iterApp t vs)
+           → Pure (case (constr i vs) ts)
+    -- case applied to anything else is Unknown
+
+    builtin : {b : Builtin} → Pure (builtin b)
+
+    unsat-builtin : {t₁ t₂ : X ⊢} {arity args : ℕ}
+            → saturation t₁ ≡ just (arity , args)
+            → arity > (suc args)
```
effectfully, line 71: Still wrong as per the previous discussion. `Apply IfThenElse True` is impure, because it fails due to a missing `force`, but this logic here doesn't recognize that.

The implementation is being fixed.

plutus-metatheory/src/Untyped/Purity.lagda.md
```diff
@@ -0,0 +1,170 @@
...
+           → Pure (iterApp t vs)
+           → Pure (case (constr i vs) ts)
+    -- case applied to anything else is Unknown
+
+    builtin : {b : Builtin} → Pure (builtin b)
+
+    unsat-builtin₀ : {t : X ⊢} {a₀ a₁ : ℕ}
+            → sat t ≡ want (suc (suc a₀)) a₁
+            → Pure t
+            → Pure (force t)
```
effectfully, line 90: What if the builtin only takes one type argument?

plutus-metatheory/src/Untyped/Purity.lagda.md
```diff
@@ -0,0 +1,170 @@
...
+
+    unsat-builtin₁ : {t t₁ : X ⊢} {a₁ : ℕ}
+            → sat t ≡ want zero (suc (suc a₁))
+            → Pure t
+            → Pure t₁
+            → Pure (t · t₁)
```
effectfully, line 96: What if the builtin takes only one term argument?

plutus-metatheory/src/Untyped/Purity.lagda.md
```diff
@@ -0,0 +1,170 @@
...
+  interleave-error : ∀ {A : Set} → A
+
+sat : {X : Set} → X ⊢ → Arity
+sat (` x) = no-builtin
+sat (ƛ t) = no-builtin
+sat (t · t₁) with sat t
+... | no-builtin = no-builtin
+... | want zero zero = want zero zero -- This will reduce the left first...
+... | want zero (suc zero) = want zero zero
+... | want zero (suc (suc a₁)) = want zero (suc a₁)
```
effectfully, line 59: These two look like they can be a single `want zero (suc a1)` clause?

plutus-metatheory/src/Untyped/Purity.lagda.md
ramsay-t: Then nothing applied to it will be pure! :)

plutus-metatheory/src/Untyped/Purity.lagda.md
ramsay-t: This is a valid question...

PR #6965, opened 2025-03-20
Remove Data.AssocMap.toDataList

ana-pantilie:
It does not compile because of https://github.com/IntersectMBO/plutus/issues/6085. I have recorded this in https://github.com/IntersectMBO/plutus/issues/6904 so that we do not forget to bring it back once the bug is fixed.


effectfully (approved):
That's a pretty radical solution.

ana-pantilie:
I'm going to merge this, if anyone can suggest a workaround I'll open a PR with that.

PR #6968, opened 2025-03-21
Fix haddock docstring in Data.Version.Extras.hs

zeme-wana:


plutus-core/plutus-core/src/Data/Version/Extras.hs
```diff
@@ -11,8 +11,8 @@ import Language.Haskell.TH qualified as TH
 
 
 gitAwareVersionInfo
-  :: Data.Version.Version -- | The version, usually coming from Paths_<pkg>.version
-  -> TH.ExpQ -- | A string that includes, if available, the git hash and the git commit date
+  :: Data.Version.Version -- ^ The version, usually coming from the Paths_<pkg> module
```
zeme-wana, line 14: `-- |` breaks on ghc810

PR #6971, opened 2025-03-22
Fix a bug in PIR's callsite inliner that caused it to skip valid inlining

### LINKED ISSUE #6970: Callsite inliner does not inline enough
In my PR https://github.com/IntersectMBO/plutus/pull/6945, I noticed that the callsite inliner doesn't inline enough. In particular:

Image

It should inline tail-1935 everywhere.

zliu41:
Fixes #6970

The third argument to `safeToBetaReduce` should be the body of the `LamAbs`, not the `LamAbs` itself!

### DIFF
--- a/plutus-core/plutus-ir/test/PlutusIR/Transform/Inline/callsite-6970
+++ b/plutus-core/plutus-ir/test/PlutusIR/Transform/Inline/callsite-6970
@@ -1,0 +1,51 @@
+{-
+let
+  !f : all a. list integer -> list integer
+    = /\a -> \(xs : list integer) -> tailList {integer} xs
+in
+\(xs : list integer) ->
+  headList
+    {integer}
+    (tailList {integer} (f {integer} (f {integer} xs)))
+-}
+
+(let
+  (nonrec)
+  (termbind
+    (strict)
+    (vardecl
+      f
+      (all
+        a
+        (type)
+        (fun
+          [ (con list) (con integer) ]
+          [ (con list) (con integer) ]
+        )
+      )
+    )
+    (abs
+      a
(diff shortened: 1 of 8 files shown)


plutus-core/plutus-ir/test/PlutusIR/Transform/Inline/callsite-6970
```diff
@@ -0,0 +1,51 @@
...
+  !f : all a. list integer -> list integer
+    = /\a -> \(xs : list integer) -> tailList {integer} xs
+in
+\(xs : list integer) ->
+  headList
+    {integer}
+    (tailList {integer} (f {integer} (f {integer} xs)))
+-}
+
+(let
```
Unisay, line 12: If only this file had an extension `.uplc` then syntax highlighting would have worked out of the box.

Unisay (approved):
Nice catch!

plutus-core/plutus-ir/test/PlutusIR/Transform/Inline/callsite-6970
zliu41: This is PIR though.

Unisay: I am confused: this file contains both and IIUC PIR is inside a comment (I don't expect code in the comment to be syntax-highlighted) but UPLC is outside (I do expect it to be syntax-highlighted). 

Why don't you go with the regular approach: one file for PIR and a separate one for UPLC ?

Anyhow, I find a good practice to use file extensions that hint at a media-type.

PR #6974, opened 2025-03-24
Proper purity check for unsaturated builtins

Unisay:
Closes https://github.com/IntersectMBO/plutus-private/issues/1497

(You might find it easier to review commits individually)


Unisay:
Golden execution budgets have changed (disclaimer: the summary has been LLM-generated):
| File Name | CPU | MEM | Old CPU | Old MEM | % CPU Change | % MEM Change |
|-----------|-----|-----|---------|---------|-------------|-------------|
| checkScriptContext1-20 | 23044762 | 90251 | 23140762 | 90851 | -0.41% | -0.66% |
| checkScriptContext1-4 | 6420522 | 24683 | 6516522 | 25283 | -1.47% | -2.37% |
| checkScriptContext1-20 | 23174425 | 90583 | 23270425 | 91183 | -0.41% | -0.66% |
| checkScriptContext1-4 | 6550185 | 25015 | 6646185 | 25615 | -1.44% | -2.34% |
| checkScriptContext1-20 | 23174425 | 90583 | 23334425 | 91583 | -0.69% | -1.09% |
| checkScriptContext1-4 | 6550185 | 25015 | 6710185 | 26015 | -2.38% | -3.84% |
| geq1 | 646178560 | 2031650 | 647138560 | 2037650 | -0.15% | -0.29% |
| geq2 | 682378509 | 2158070 | 683338509 | 2164070 | -0.14% | -0.28% |
| geq3 | 712809054 | 2260656 | 713817054 | 2266956 | -0.14% | -0.28% |
| geq4 | 619870155 | 1917742 | 620830155 | 1923742 | -0.15% | -0.31% |
| geq5 | 669871047 | 2104358 | 670879047 | 2110658 | -0.15% | -0.30% |
| gt1 | 754528368 | 2405284 | 755488368 | 2411284 | -0.13% | -0.25% |
| gt2 | 681242509 | 2150970 | 682202509 | 2156970 | -0.14% | -0.28% |
| gt3 | 823593022 | 2642276 | 824601022 | 2648576 | -0.12% | -0.24% |
| gt4 | 619598155 | 1916042 | 620558155 | 1922042 | -0.15% | -0.31% |
| gt5 | 720831527 | 2281862 | 721839527 | 2288162 | -0.14% | -0.28% |
| destructSum-manual | 7401268 | 26925 | 7449268 | 27225 | -0.64% | -1.10% |

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -100,66 +114,80 @@ termEvaluationOrder
...
-            evalThis (EvalTerm Pure WorkFree t)
-          else -- If it's saturated or oversaturated, we might have an effect here
-            evalThis (EvalTerm MaybeImpure MaybeWork t)
+      parameters :: [Param] =
+        builtinArity @uni @fun (Proxy @uni) builtinSemanticsVariant fun
+
+      termEvalOrder :: [Param] -> [Arg name uni fun a] -> EvalOrder name uni fun a
+      termEvalOrder [] _ =
+        singletonEvalOrder (EvalTerm MaybeImpure MaybeWork t)
+      -- Builtin is not fully saturated with term arguments, thus impure.
```
zliu41, line 128: "thus impure" should be "thus pure"

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -100,66 +114,80 @@ termEvaluationOrder
...
+      termEvalOrder (TermParam : _params) [] =
+        singletonEvalOrder (EvalTerm Pure WorkFree t)
+      -- Strip applied term parameter
+      termEvalOrder (TermParam : params) (_arg : remainingArgs) =
+        termEvalOrder params remainingArgs
```
zliu41, line 133: You need to check whether `_arg` is `ApplyTerm` or `ForceTypeParam`. If it's `ForceTypeParam` then it is impure.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -100,66 +114,80 @@ termEvaluationOrder
...
+      -- Strip applied type parameter
+      termEvalOrder (TypeParam : params) (ForceTypeParam : remainingArgs) =
+        termEvalOrder params remainingArgs
+      -- Type parameter expected, non-Force argument applied. Error is impure.
```
zliu41, line 137: What if the argument is `[]`?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -100,66 +114,80 @@ termEvaluationOrder
   -> EvalOrder name uni fun a
 termEvaluationOrder builtinSemanticsVariant = goTerm
  where
+  goTerm :: Term name uni fun a -> EvalOrder name uni fun a
   goTerm = \case
-    t@(splitApplication -> (Builtin _ann fun, args)) ->
-      foldMap (goTerm . snd) args <> evalOrder
+    t@(splitArgs -> (Builtin _ann fun, args)) ->
+      foldMap goTerm [ a | ApplyTerm a <- args ] <> termEvalOrder parameters args
```
zliu41, line 120: I don't think this is correct. In `(force ifThenElse) arg1 arg2 arg3`, you first evaluate `force ifThenElse`, which is `Pure MaybeWork`, before evaluating the args.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -100,66 +114,80 @@ termEvaluationOrder
...
-      evalOrder =
-        if length args < length (builtinArity @uni @fun (Proxy @uni) builtinSemanticsVariant fun)
-          then -- If it's unsaturated, we definitely don't do any work
-            evalThis (EvalTerm Pure WorkFree t)
-          else -- If it's saturated or oversaturated, we might have an effect here
-            evalThis (EvalTerm MaybeImpure MaybeWork t)
+      parameters :: [Param] =
+        builtinArity @uni @fun (Proxy @uni) builtinSemanticsVariant fun
+
+      termEvalOrder :: [Param] -> [Arg name uni fun a] -> EvalOrder name uni fun a
```
zliu41, line 125: nit: again I like the original name (`evalOrder`) better, or just call it `go`. A local function like this doesn't need a long name; also, having both `termEvaluationOrder` and `termEvalOrder` is imo awkward.

zliu41 (changes requested):

zliu41:
btw this is worth putting in the changelog.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
effectfully: `force <builtin>` isn't pure, it fails if the builtin expects a term argument. We also consider it `WorkFree` in the PIR purity check IIRC, but that's arguably a misfeature.

I'd just put all of this into a single local recursive definition, but I guess this is fine too. Especially given that the ordering doesn't really matter purity- or work-wise. But then this whole definition is about ordering, so maybe it's worth getting it right.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -100,66 +114,80 @@ termEvaluationOrder
...
+      termEvalOrder [] _ =
+        singletonEvalOrder (EvalTerm MaybeImpure MaybeWork t)
+      -- Builtin is not fully saturated with term arguments, thus impure.
+      termEvalOrder (TermParam : _params) [] =
+        singletonEvalOrder (EvalTerm Pure WorkFree t)
+      -- Strip applied term parameter
+      termEvalOrder (TermParam : params) (_arg : remainingArgs) =
```
effectfully, line 132: It's weird that we handle `arg` elsewhere and skip it here when it could all be a single definition handling all of the parts at once.

effectfully (commented):
Ziyang is right, there's bugs.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -76,12 +75,27 @@ unEvalOrder (EvalOrder ts) =
   takeWhileInclusive :: (a -> Bool) -> [a] -> [a]
   takeWhileInclusive p = foldr (\x ys -> if p x then x : ys else [x]) []
 
-evalThis :: EvalTerm name uni fun a -> EvalOrder name uni fun a
-evalThis = EvalOrder . DList.singleton
+singletonEvalOrder :: EvalTerm name uni fun a -> EvalOrder name uni fun a
```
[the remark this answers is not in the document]
Unisay: My problem with this meaning is that I don't know what a "now" is. 
My intuition is: create an evaluation order from  a single term, aka "Singleton Evaluation Order".

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -76,12 +75,27 @@ unEvalOrder (EvalOrder ts) =
...
   prettyBy config eo = vsep $ fmap (prettyBy config) (unEvalOrder eo)
 
+-- Terms can have Type and Term parameters.
+-- In correct programs Type parameters are 'Force'd while
+-- term parameters are 'Apply'd.
+-- This data type represents this distinction.
+data Arg name uni fun a = ApplyTerm (Term name uni fun a) | ForceTypeParam
+
+-- | Strip off arguments
+splitArgs :: Term name uni fun a -> (Term name uni fun a, [Arg name uni fun a])
```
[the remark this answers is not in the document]
Unisay: > this seems useful enough to be placed in the same module as `splitApplication`. 

I approach it differently: 
First I ask myself: _is there an evidence that this functionality is useful elsewhere?_
If yes: move to a common place, make general enough.
If no: keep it local until such an evidence appears.
This is my way to avoid a [premature generalization.](https://wiki.c2.com/?PrematureGeneralization)

In this case `splitArgs` is used in one place only, hence the local definition.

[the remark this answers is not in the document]
Unisay: > And it should be [(a, Arg name uni fun a)], instead of [Arg name uni fun a], i.e., the annotation should be preserved.

What would be a reason to preserve the annotation given that its not needed in the only place this function is used?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
Unisay: My bad: this comment relates to the line below it.
First I had blank lines in between the statements so it was clear that comment relates to the "closest" line, which is the one below. Then I decided to make code more vertically compact and the relation got ambiguous.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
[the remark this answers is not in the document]
Unisay: Btw, at first I had the `splitArgs` underneath the "where" clause, as a local function definition. Unfortunately Haskell doesn't support local data type definitions so I had to define `Arg` at the top level. After that I decided to lift `splitArgs` closer to the data type (top level) to highlight the relationship between it and  the data type, but still consider it "local".

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
Unisay: I don't understand what you mean. Could you elaborate, please?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
Unisay: Let me write down what I understood in a more specific way:

For the `(force ifThenElse) arg1 arg2 arg3` term, current code produces this evaluation order:
```haskell
[ EvalTerm Pure WorkFree (Var () (Name{_nameText = "arg1", _nameUnique = Unique{unUnique = 1}}))
, EvalTerm Pure WorkFree (Var () (Name{_nameText = "arg2", _nameUnique = Unique{unUnique = 2}}))
, EvalTerm Pure WorkFree (Var () (Name{_nameText = "arg3", _nameUnique = Unique{unUnique = 3}}))
, EvalTerm
    MaybeImpure
    MaybeWork
    ( Apply
        ()
        ( Apply
            ()
            ( Apply
                ()
                (Force () (Builtin () IfThenElse))
                (Var () (Name{_nameText = "arg1", _nameUnique = Unique{unUnique = 1}}))
            )
            (Var () (Name{_nameText = "arg2", _nameUnique = Unique{unUnique = 2}}))
        )
        (Var () (Name{_nameText = "arg3", _nameUnique = Unique{unUnique = 3}}))
    )
]
```
First 3 elements come from 
```haskell
foldMap goTerm [ a | ApplyTerm a <- args ]
```

last element is from 
```haskell
termEvalOrder parameters args
```

 @zliu41 expects a different order, such that starts from `EvalTerm Pure MaybeWork (Force () (Builtin () IfThenElse)) `
 ```haskell
[ EvalTerm Pure MaybeWork (Force () (Builtin () IfThenElse)) 
, EvalTerm Pure WorkFree (Var () (Name{_nameText = "arg1", _nameUnique = Unique{unUnique = 1}}))
, EvalTerm Pure WorkFree (Var () (Name{_nameText = "arg2", _nameUnique = Unique{unUnique = 2}}))
, EvalTerm Pure WorkFree (Var () (Name{_nameText = "arg3", _nameUnique = Unique{unUnique = 3}}))
, EvalTerm
    MaybeImpure
    MaybeWork
    ( Apply
        ()
        ( Apply
            ()
            ( Apply
                ()
                (Force () (Builtin () IfThenElse))
                (Var () (Name{_nameText = "arg1", _nameUnique = Unique{unUnique = 1}}))
            )
            (Var () (Name{_nameText = "arg2", _nameUnique = Unique{unUnique = 2}}))
        )
        (Var () (Name{_nameText = "arg3", _nameUnique = Unique{unUnique = 3}}))
    )
]
 ```
 
 right?
 
 @effectfully corrects that it should start from `EvalTerm MaybeImpure MaybeWork (Force () (Builtin () IfThenElse)) `:

 ```haskell
[ EvalTerm MaybeImpure MaybeWork (Force () (Builtin () IfThenElse)) 
, EvalTerm Pure WorkFree (Var () (Name{_nameText = "arg1", _nameUnique = Unique{unUnique = 1}}))
, EvalTerm Pure WorkFree (Var () (Name{_nameText = "arg2", _nameUnique = Unique{unUnique = 2}}))
, EvalTerm Pure WorkFree (Var () (Name{_nameText = "arg3", _nameUnique = Unique{unUnique = 3}}))
, EvalTerm
    MaybeImpure
    MaybeWork
    ( Apply
        ()
        ( Apply
            ()
            ( Apply
                ()
                (Force () (Builtin () IfThenElse))
                (Var () (Name{_nameText = "arg1", _nameUnique = Unique{unUnique = 1}}))
            )
            (Var () (Name{_nameText = "arg2", _nameUnique = Unique{unUnique = 2}}))
        )
        (Var () (Name{_nameText = "arg3", _nameUnique = Unique{unUnique = 3}}))
    )
]
```
Did I sum it up correctly?

File: plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs (hunk 7 above)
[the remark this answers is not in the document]
Comment (zliu41) [reply plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs:78]: `eval` or `evalTerm` are also good names- both are better than `singletonEvalOrder`.

`singletonEvalOrder` says "I'm taking one element, and turning it into a list". But that's totally irrelevant - the list thing is purely implementation detail.

E.g., consider

```haskell
    t@(Case _ scrut _) ->
      -- first the scrutinee
      goTerm scrut
        -- then the whole term, which means finding the case so work
        <> eval (EvalTerm Pure MaybeWork t)
        -- then we go to an unknown scrutinee
        <> eval Unknown
```

This is quite clear and intuitive - first you process the scrutinee. Then you evaluate the whole term `t`. Then you evaluate some unknown branch.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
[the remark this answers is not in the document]
zliu41: This `splitArgs` is in fact the equivalent of PIR's `splitApplications`. So arguably, UPLC's `splitApplications` should be this - for consistency. And so this could be as useful as PIR's `splitAppliations` - it could be useful for other optimizations, and there probably should be some tests for this function.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
zliu41: > @zliu41 expects a different order, such that starts from EvalTerm Pure MaybeWork (Force () (Builtin () IfThenElse))

Yes - because like @effectfully said, if we try to evaluate something like `(force addInteger) arg1 arg2`, then `arg1` and `arg2` won't be evaluated - it will immediately fail.

Unisay: But in this case the `termEvaluationOrder` has access to the information how many type arguments this particular builtin function expects, so if it looks at the `force addInteger` then it can know for sure that its impure, and conversely, when it looks at `force ifThenElse` then it knows the term is pure. 

In other words:
- `force <builtin>` may be impure
- `force ifThenElse` is pure
- `force addInteger` is impure

Unisay: On the actual evaluation order:

I have instrumented the `computeCek` function to print out steps and then evaluated this test UPLC program
```
(program
   1.1.0
    [
      [ [ (force (builtin ifThenElse)) 
          (con bool True) 
        ] 
        (con integer 2) 
      ]
      (con integer 3)
    ]
)
```

Got the following output:
```
cabal run plutus -- --run test.uplc

Apply ([ [ (force (builtin ifThenElse)) (con bool True) ] (con integer 2) ]) ((con integer 3))
Step (BApply)

Apply ([ (force (builtin ifThenElse)) (con bool True) ]) ((con integer 2))
Step (BApply)

Apply ((force (builtin ifThenElse))) ((con bool True))
Step (BApply)

Force ((builtin ifThenElse))
Step (BForce)

Builtin (IfThenElse)
Step (BBuiltin)

Constant (Some (ValueOf DefaultUniBool True))
Step (BConst)

Constant (Some (ValueOf DefaultUniInteger 2))
Step (BConst)

Constant (Some (ValueOf DefaultUniInteger 3))
Step (BConst)

Execution succeeded. Final Term:
(con integer 2)

Used budget: ExBudget {exBudgetCPU = ExCPU 204149, exBudgetMemory = ExMemory 901}
```

Please help me to understand how this de-facto CEK stepping order relates to the idea of "EvalOrder" in the `Purity` module?

Unisay: I think I have figured it out: instead of looking at the order in which terms are traversed,  I should be looking at the order in which evaluated terms are returned, e.g. if I print terms in the `returnCek` the output is this:
```
(builtin ifThenElse)
(force (builtin ifThenElse))
(con bool True)
[ (force (builtin ifThenElse)) (con bool True) ]
(con integer 2)
[ [ (force (builtin ifThenElse)) (con bool True) ] (con integer 2) ]
(con integer 3)
(con integer 2)
```
is this the exact order that we want to simulate in the `Purity` module?

zliu41: Yes - basically, given an application, we first evaluate the fun, then evaluate the arg, then evaluate the whole application.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -100,66 +139,94 @@ termEvaluationOrder
...
  where
+  goTerm :: Term name uni fun a -> EvalOrder name uni fun a
   goTerm = \case
-    t@(splitApplication -> (Builtin _ann fun, args)) ->
-      foldMap (goTerm . snd) args <> evalOrder
+    t@(splitArgs -> (Builtin _ann fun, args)) ->
+      -- First we evaluate a term with its application "structure"
+      go (builtinArity @uni @fun (Proxy @uni) builtinSemanticsVariant fun) args
+      -- then its arguments, in left-to-right order
+      <> foldMap goTerm [ a | ApplyTerm a <- args ]
```
zliu41, line 148: I believe this is incorrect. You are now evaluating all arguments at the end. But the evaluation of the arguments should be interleaved with what you call "application structure".

Using your example - `force ifThenElse True 2 3` - this is an application so you first evaluation the function, `force ifThenElse True 2`. This is again an application so you first evaluate `force ifThenElse True`. Similarly, this first evaluates `force ifThenElse`. Then evaluate `True`. So in the end the evaluation order is
- force ifThenElse
- True
- force ifThenElse True
- 2
- force ifThenElse True 2
- 3
- force ifThenElse True 2 3

All but the last evaluation is pure.

I'd suggest mirroring the `goBuiltinApp` function in PIR.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -100,66 +139,94 @@ termEvaluationOrder
...
-      evalOrder =
-        if length args < length (builtinArity @uni @fun (Proxy @uni) builtinSemanticsVariant fun)
-          then -- If it's unsaturated, we definitely don't do any work
-            evalThis (EvalTerm Pure WorkFree t)
-          else -- If it's saturated or oversaturated, we might have an effect here
-            evalThis (EvalTerm MaybeImpure MaybeWork t)
+      go :: [Param] -> [Arg name uni fun a] -> EvalOrder name uni fun a
+      go parameters arguments =
+        case parameters of
+          -- Parameter-less builtin is counted as impure.
```
zliu41, line 153: This is not a parameter-less builtin - it is a fully saturated builtin.

zliu41 (changes requested):

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
zliu41: Specifically you can do this: instead of `go <> ...` or `... <> go`, just do everything in `go`.

```haskell
go (p:ps) (a:as)
  | p is term, and a is term = goTerm a <> go ps as  -- evaluate term arg
  | p is type, and a is type = go ps as  -- correct force
  | otherwise = maybeImpureWork  -- mismatch
go [] _ = maybeImpureWork -- saturated builtin app
go (_:_) [] = pureWorkFree -- unsaturated
```

That should be it. This is slightly different from `goBuiltinApp`, since `goBuiltinApp` doesn't seem to care about the "mismatch" scenario - I guess in PIR you can assume there's no mismatch because of types.

Unisay: I added an `ifThenElse` test to the `PlutusIR.Purity.Tests` and the produced order is (pretty printed)
```
pure work-free: True
pure work-free: 2
pure work-free: 3
impure? maybe work?: ifThenElse True 2 3
```

zliu41: The saturated `ifThenElse` obviously needs to be evaluated at the end, after everything else has been evaluated. It's impossible to evaluate `force ifThenElse b t f`, and reduce it to `t` or `f`, without first evaluating `b` to a constant - either true or false. That's the same with all builtins - it's impossible to evaluate `addInteger x y`, without first evaluating `x` and `y` to constants.

In the order you are seeing, step 1 is evaluating an application, not the builtin itself. In other words it is the `𝑠; 𝜌 ⊳ [𝑀 𝑁]` case in the spec. Evaluating the builtin function at the end is the `𝖤𝗏𝖺𝗅 𝖢𝖤𝖪 (𝑠, 𝑏, 𝑉 ⋅𝑉 ) if 𝜄 ∈ U# ∪ {∗}` step, and it needs to be included in `EvalOrder`.

The PIR implementation appears to omit the evaluation of applications, which is acceptable - it is minimal work (evaluating an application just incurs a `BApply` step, which is minimal work, before immediately evaluating the function). You can do the same.

In my comment, when I say evaluating `force ifThenElse True 2`, it corresponds to the `if 𝜄 ∈ U# ∪ {∗}` and `if 𝜄 ∈ Q` part of the spec - i.e., checking whether the argument kind (term or type) is correct. This is not really an evaluation, just a check, which you can do in the `go` function as I suggested.

The entire CEK machine specification is in a single page in the spec (page 23), and I think it is better to refer to it directly, rather than adding printing or digging into the implementation.

Unisay: Thanks @zliu41 for taking time to dissect it for me. You're right that spec is a valuable source of truth that I should be taking into account. 

Looking at the figure 2.10 "A CEK Machine for Plutus Core" I have traced evaluation of the term  `(force (builtin ifThenElse) (con bool True) (con integer 2)) (con integer 3)` manually:

```
Stack: []
Control: ((force (builtin ifThenElse) (con bool True) (con integer 2)) (con integer 3))

--> Application rule: 𝑠; 𝜌 ⊳ [𝑀 𝑁] ↦ [_ (𝑁, 𝜌)]⋅𝑠; 𝜌 ⊳ 𝑀

Stack: [_ (con integer 3)]
Control: (force (builtin ifThenElse) (con bool True) (con integer 2))

--> Application rule: 𝑠; 𝜌 ⊳ [𝑀 𝑁] ↦ [_ (𝑁, 𝜌)]⋅𝑠; 𝜌 ⊳ 𝑀

Stack: [_ (con integer 2), _ (con integer 3)]
Control: (force (builtin ifThenElse) (con bool True))

--> Application rule: 𝑠; 𝜌 ⊳ [𝑀 𝑁] ↦ [_ (𝑁, 𝜌)]⋅𝑠; 𝜌 ⊳ 𝑀

Stack: [_ (con bool True), _ (con integer 2), _ (con integer 3)]
Control: (force (builtin ifThenElse))

--> Force rule: 𝑠; 𝜌 ⊳ (force 𝑀) ↦ (force _)⋅𝑠; 𝜌 ⊳ 𝑀

Stack: [(force _), _ (con bool True), _ (con integer 2), _ (con integer 3)]
Control: (builtin ifThenElse)

--> Built-in rule: 𝑠; 𝜌 ⊳ (builtin 𝑏) ↦ 𝑠 ⊲ 〈builtin 𝑏 [] 𝛼(𝑏)〉

Stack: [(force _), _ (con bool True), _ (con integer 2), _ (con integer 3)]
Control: 〈builtin ifThenElse [] ([∀∗, 𝚋𝚘𝚘𝚕, ∗, ∗] → ∗)〉

--> Force built-in rule: 
    (force _)⋅𝑠 ⊲ 〈builtin 𝑏 𝑉 (𝜄⋅𝜂)〉 ↦ 𝑠 ⊲ 〈builtin 𝑏 𝑉 𝜂〉 if 𝜄 ∈ Q

Stack: [_ (con bool True), _ (con integer 2), _ (con integer 3)]
Control: 〈builtin ifThenElse [] ([𝚋𝚘𝚘𝚕, ∗, ∗] → ∗)〉

--> Type application rule: 
    [_ 𝑉]⋅𝑠 ⊲ 〈builtin 𝑏 𝑉' (𝜄⋅𝜂)〉 ↦ 𝑠 ⊲ 〈builtin 𝑏 (𝑉'⋅𝑉) 𝜂〉 if 𝜄 ∈ U# ∪ {∗}

Stack: [_ (con integer 2), _ (con integer 3)]
Control: 〈builtin ifThenElse [(con bool True)] (integer → integer → integer)〉

--> Term application rule: 
    [_ 𝑉]⋅𝑠 ⊲ 〈builtin 𝑏 𝑉' (𝜄⋅𝜂)〉 ↦ 𝑠 ⊲ 〈builtin 𝑏 (𝑉'⋅𝑉) 𝜂〉

Stack: [_ (con integer 3)]
Control: 〈builtin ifThenElse [(con bool True), (con integer 2)] (integer → integer)〉

--> Term application rule: 
    [_ 𝑉]⋅𝑠 ⊲ 〈builtin 𝑏 𝑉' (𝜄⋅𝜂)〉 ↦ 𝑠 ⊲ 〈builtin 𝑏 (𝑉'⋅𝑉) 𝜂〉

Stack: []
Control: 〈builtin ifThenElse [(con bool True), (con integer 2), (con integer 3)] []〉

--> Built-in function execution rule: 𝑠 ⊲ 〈builtin 𝑏 𝑉 []〉 ↦ 𝑠 ⊲ ⟦𝑏⟧(𝑉)

Stack: []
Control: (con integer 2)

--> Final value (evaluation complete)
```

Now I want to make a correspondence between this evaluation trace and the "term evaluation order" thing that Purity modules operate with. 

Lets start with removing everything but the control and the final value:
```
((force (builtin ifThenElse) (con bool True) (con integer 2)) (con integer 3))
(force (builtin ifThenElse) (con bool True) (con integer 2))
(force (builtin ifThenElse) (con bool True))
(force (builtin ifThenElse))
(builtin ifThenElse)
〈builtin ifThenElse [] ([∀∗, 𝚋𝚘𝚘𝚕, ∗, ∗] → ∗)〉
〈builtin ifThenElse [] ([𝚋𝚘𝚘𝚕, ∗, ∗] → ∗)〉
〈builtin ifThenElse [(con bool True)] (integer → integer → integer)〉
〈builtin ifThenElse [(con bool True), (con integer 2)] (integer → integer)〉
〈builtin ifThenElse [(con bool True), (con integer 2), (con integer 3)] []〉
(con integer 2)
```

Note: I assume that "term evaluation order" (as a list of terms) is a thing that is useful not only for the purity calcualtion `isPure :: Term -> Bool`, but for other purposes too (otherwise why would we unfold a list of terms and then fold it to a boolean? We could have folded a term over the `Bool` accumulator without bothering with the intermediate list).  With this perspective removing "apply" terms from this order seems ... wrong?

I have added an [`ifThenElse` test](https://github.com/IntersectMBO/plutus/pull/6974/commits/9da85df8b8752eefb4e475f4fd954e7f944ebec8) to the `PlutusIR.Purity.Tests`

and it shows that term (con integer 3) is in the "evaluation order",
while this branch of the `ifThenElse` wasn't evaluated at all.

zliu41: > If it's not a proper abstraction, we may as well give up on maintaining the precise evaluation order.

If we want `EvalOrder` to contain all CEK machine evaluations, that's fine. If we want to omit some that aren't relevant for the inliner, that's also fine (I agree it's worth documenting though). But the most important thing is that whatever is in the `EvalOrder` must be in the correct order, and it must not miss anything that is impure (e.g., actually evaluating a sasturated builtin function) otherwise it isn't acceptable.

> we should probably expect every team member to be familiar with how the CEK machine works, it's the most important part of UPLC.

Yes, and that requires reading and understanding the spec. In the next meetup we should hand out swags (or maybe just napkins) with CEK machine specs printed on them.

Unisay: Ok, so in the end I replicated the same eval order that Plutus IR has, this is witnessed by a few unit tests added.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -50,11 +55,28 @@ data EvalTerm name uni fun a
...
-  where
-  prettyBy _ Unknown                    = "<unknown>"
-  prettyBy config (EvalTerm eff work t) = pretty eff <+> pretty work <> ":" <+> prettyBy config t
+ ( Show name
+ , Everywhere uni Show
+ , Show fun
+ , Show a
+ , GShow uni
+ , Closed uni
+ ) => Show (EvalTerm name uni fun a) where
```
zliu41, line 64: Why do you need Show?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
```diff
@@ -76,10 +98,11 @@ unEvalOrder (EvalOrder ts) =
   takeWhileInclusive :: (a -> Bool) -> [a] -> [a]
   takeWhileInclusive p = foldr (\x ys -> if p x then x : ys else [x]) []
 
-evalThis :: EvalTerm name uni fun a -> EvalOrder name uni fun a
-evalThis = EvalOrder . DList.singleton
+eval :: EvalTerm name uni fun a -> EvalOrder name uni fun a
```
zliu41, line 101: If you prefer renaming `evalThis` to `eval`, rename the one in PIR as well.

plutus-core/untyped-plutus-core/test/Analysis/Spec.hs
```diff
@@ -1,66 +1,148 @@
...
+      isPure builtinSemantics termIfThenElse @?= False
+
+testForceNoTypeParam :: TestTree
+testForceNoTypeParam =
+  testCase "forcing a non-existing type param is impure" $
+    isPure builtinSemantics (Force () (Builtin () EncodeUtf8)) @?= False
+
+testApplyNoTermParam :: TestTree
+testApplyNoTermParam =
+  testCase "applying a non-existing term param is impure" $
```
zliu41, line 141: There are two subcases here:
- Applying a term param, but a type param is expected.
- Applying a term param, but the builtin has already been saturated (i.e., over-saturated builtin)

The first one is definitely an error, but the second one could be valid, e.g., `ifThenElse True (\i -> i + 1) (\i -> i + 2) 42`. They are both impure, but for different reasons.

plutus-core/untyped-plutus-core/test/Analysis/Spec/Lib.hs
```diff
@@ -0,0 +1,76 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Analysis.Spec.Lib where
```
zliu41, line 3: I don't think there's any existing module named "*.Spec.Lib". For consistency, just call it `Analysis.Lib`?

plutus-core/plutus-ir/test/PlutusIR/Purity/ifThenElse
```diff
@@ -0,0 +1,8 @@
+[
```
zliu41, line 1: Where's the golden file for this?

plutus-core/untyped-plutus-core/test/Analysis/Spec.hs
```diff
@@ -1,66 +1,148 @@
...
+  testCase "evalOrderLazy" $
+    let order = termEvaluationOrder builtinSemantics dangerTerm
+        subterms = unEvalOrder order
+     in 4 @=? length subterms
+
+testEvalOrderIfThenElse :: TestTree
+testEvalOrderIfThenElse =
+  testCase "evalOrderIfThenElse" $
+    unEvalOrder (termEvaluationOrder builtinSemantics termIfThenElse)
+      %?= [ EvalTerm Pure WorkFree $ termVar 1
```
zliu41, line 47: Why not use `goldenEvalOrder` here?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Purity.hs
Unisay: For the unit test that compares eval terms: `assertEquals` needs `Eq` and `Show`

plutus-core/plutus-ir/test/PlutusIR/Purity/ifThenElse
Unisay: [plutus-core/plutus-ir/test/PlutusIR/Purity/ifThenElse.golden](https://github.com/IntersectMBO/plutus/pull/6974/files#diff-fffcfed6d95a46514c41ab031b645d30844b4d5b71a6f62c24871e04ad342a17)

zliu41 (approved):
Just need a changelog entry

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Contexts.hs
```diff
@@ -0,0 +1,41 @@
...
+  | AppCtxType a (AppCtx name uni fun a)
+  | AppCtxEnd
+
+{-| Takes a term and views it as a head plus an 'AppCtx', e.g.
+@
+  [{ f t } u v]
+  -->
+  (f, [{ _ t } u v])
+  ==
+  f (AppCtxType t (AppCtxTerm v (AppCtxTerm u AppCtxEnd)))
```
effectfully, line 20: `v` and `u` should be swapped.

plutus-core/untyped-plutus-core/test/Analysis/Spec.hs
```diff
@@ -1,66 +1,135 @@
...
+  When it comes to purity of builtin terms,
+  we want to test the following cases:
+  +---------------+---------------+
+  | Type args     | Term args     |
+  | left to apply | left to apply |
+  +---------------+---------------+
+  |     some      |     some      |
+  |     none      |     some      |
+  |     none      |     none      |
+  +---------------+---------------+
```
effectfully, line 76: Great comment and tests.

effectfully (approved):

PR #6978, opened 2025-03-25
Run default PIR/UPLC optimizations during code lifting

zliu41:
Optimizations are currently disabled during lifting, out of concern that lifting could otherwise be too slow at runtime. I don't think this is the right default - optimizations make a big difference in script efficiency, and lifting speed should rarely be a concern. For those uncommon cases where it is, the `*Unopt` variants are provided.


plutus-tx-plugin/test/Budget/9.6/listIndexing.budget.golden
```diff
@@ -1,2 +1,2 @@
-({cpu: 6684432
-| mem: 33922})
\ No newline at end of file
+({cpu: 5196432
+| mem: 24622})
```
effectfully, line 2: damn that's good

Unisay (approved):
A pleasant improvement!

PR #6980, opened 2025-03-26
Fix ByteString Agda unparsing

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1505


ramsay-t (changes requested):
When called from the command line, a certificate that uses this has compile errors because of the way it imports `Utils`: 
```
[nix-shell:~/temp/plutus]$ agda -i/Users/ramsay/plutus/plutus-metatheory/src AA2.agda
Checking AA2 (/Users/ramsay/temp/plutus/AA2.agda).
/Users/ramsay/temp/plutus/AA2.agda:21,1693-1709
Not in scope:
  Utils.encodeUtf8 at /Users/ramsay/temp/plutus/AA2.agda:21,1693-1709
    (did you mean
       'Builtin.Builtin.decodeUtf8' or
       'Builtin.Builtin.encodeUtf8' or
       'Builtin.DECODEUTF8' or
       'Builtin.ENCODEUTF8' or
       'Builtin.decodeUtf8' or
       'Builtin.encodeUtf8' or
       'DECODEUTF8' or
       'ENCODEUTF8' or
       'decodeUtf8' or
       'encodeUtf8'?)
when scope checking Utils.encodeUtf8
```
However, inside emacs or VSCode it doesn't have that error, but takes ages to typecheck (10+ mins and then failing the refl in emacs; just crashing to a grey window in VSCode)

Changing the way it imports Util creates different problems in the command line version, so I think it is only typechecking quickly because it fails quickly.

ana-pantilie:
@ramsay-t I'm unable to reproduce.

Running `agda IfInt.agda ` terminates successfully. So does compiling it in interactive mode. Are you sure you're on the right commit? Also, what command are you running to generate the certificate?

ana-pantilie:
@ramsay-t Is `cabal run uplc -- example -s IfIntegers ` non-deterministic? It seems to generate random programs every time I run it. It would have been a good thing to mention...

PR #6982, opened 2025-03-26
Add the inline-callsite-growth flag for setting inliner aggressiveness

### LINKED ISSUE #6975: Add -inline-threshold flag
Currently the callsite inliner is too conservative. It inlines an identifier at a callsite only if it doesn't increase the program size. But even if it increases the size slightly, one may still want to inline because
- if an identifier ends up being inlined at all callsites, we can remove the binding, which could end up making the program smaller
- inlining can expose additional optimizations and make the program faster.

To make this more customizable we can add a flag called `-inline-threshold=N` that controls the aggressiveness of the callsite inliner. `N` means it inlines if the size increase is no more than `N`.

zliu41:
Fixes #6975 

`inline-callsite-growth = x` means: callsite inlining is allowed to proceed as long as the new size is no more than original size plus `x`. So the higher `x` is, the more aggressive the inliner inlines. Currently the new size must be no more than the original size (i.e., equivalent to `x = 0`), which is too conservative.

I experimented with several values between 4 and 20, and found that 5 works the best overall, so I'm setting 5 as the default.


doc/docusaurus/docs/delve-deeper/plinth-compiler-options.md
```diff
@@ -34,6 +32,7 @@ For each boolean option, you can add a `no-` prefix to switch it off, such as `n
 |`dump-uplc`|Bool|False|Dump Untyped Plutus Core|
 |`inline-constants`|Bool|True|Always inline constants. Inlining constants always reduces script costs slightly, but may increase script sizes if a large constant is used more than once. Implied by `no-conservative-optimisation`.|
 |`inline-fix`|Bool|True|Always inline fixed point combinators. This is generally preferable as it often enables further optimization, though it may increase script size.|
+|`inline-threshold`|Int|5|Set inliner aggressiveness. 0 means it avoids making the program bigger. Higher values make the inliner more willing to inline.|
```
Unisay, line 35: Initially, judging on the option name only,  I thought that "inline-threshold" is maximum value of some metric, (e.g. maximum program size, or maximum number of terms etc.). This line shows me that its a "level" of inliner aggressiveness.

I propose to name the option either  `inline-force` or `inline-level`

https://g.co/gemini/share/f8e59770d1db

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Simplify/Opts.hs
```diff
@@ -20,6 +21,7 @@ data SimplifyOpts name a = SimplifyOpts
   , _soConservativeOpts        :: Bool
   , _soInlineHints             :: InlineHints name a
   , _soInlineConstants         :: Bool
+  , _soInlineThreshold         :: Int
```
Unisay, line 24: I'd use `Size` newtype (from the `PlutusCore.Size`) because this is what threshold is converted to in the end.

plutus-core/plutus-ir/src/PlutusIR/Transform/Inline/CallSiteInline.hs
```diff
@@ -144,13 +146,14 @@ callSiteInline processedTSize = go
...
+        thresh <- view iiInlineThreshold
         if costIsOk && rhsPure then do
           -- rename the rhs of the variable before any substitution
           renamedRhs <- rename headRhs
           applyAndBetaReduce renamedRhs args >>= \case
             Just inlined -> do
-              let -- Inline only if the size is no bigger than not inlining.
-                  sizeIsOk = termSize inlined <= processedTSize
+              let -- Inline only if the size is no bigger than not inlining plus threshold.
+                  sizeIsOk = termSize inlined <= processedTSize + fromIntegral (max 0 thresh)
```
Unisay, line 156: No `fromIntegral` would be needed if thresh has type `Size` (from the `PlutusCore.Size` module.

doc/docusaurus/docs/delve-deeper/plinth-compiler-options.md
Unisay: Update: after reading the rest of the code I realize that my initial intuition was right, eventually this value is used as a maximum program increase in `Size` that inliner accepts. Its the description that misled me. I also understand that a "maximum program increase in `Size`" can be seen as inliner's aggressiveness... indirectly. I mean the description is not "what the option value is" rather "what is an effect of this value".

plutus-core/plutus-ir/src/PlutusIR/Compiler/Types.hs
```diff
@@ -95,6 +95,7 @@ data CompilationOpts a = CompilationOpts {
     , _coInlineHints                    :: InlineHints PLC.Name (Provenance a)
     , _coInlineConstants                :: Bool
     , _coInlineFix                      :: Bool
+    , _coInlineThreshold                :: Int
```
Unisay, line 98: I'd really ❤️  to use the `Size` newtype here.

plutus-core/plutus-ir/test/PlutusIR/Transform/Inline/Tests.hs
```diff
@@ -76,4 +76,4 @@ prop_inline biVariant =
   withMaxSuccess numTestsForPassProp $
     testPassProp
       runQuote
-      $ \tc -> inlinePassSC True tc def (def {_biSemanticsVariant = biVariant})
+      $ \tc -> inlinePassSC 0 True tc def (def {_biSemanticsVariant = biVariant})
```
Unisay, line 79: Another benefit that using the `Size` newtype gives is that this magic 0 could be less magical:
```suggestion
      $ \tc -> inlinePassSC (Size 0) True tc def (def {_biSemanticsVariant = biVariant})
```

cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests/sorted.cbor.size.golden
```diff
@@ -1 +1 @@
-2167
\ No newline at end of file
+2164
```
ana-pantilie, line 1: How come the size was reduced? I would expect the benchmarks in this PR to have their size increased by a maximum of 5.

doc/docusaurus/docs/delve-deeper/plinth-compiler-options.md
ana-pantilie: I agree this should also say that it may increase the program's size by, at maximum, the value provided as the `inline-threshold`.

plutus-benchmark/nofib/test/9.6/clausify-F5.size.golden
```diff
@@ -1 +1 @@
-1598
\ No newline at end of file
+1615
```
ana-pantilie, line 1: Does this test run with `inline-threshold` set to something else other than the default?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -501,9 +505,10 @@ inlineSaturatedApp t
           fullyApplyAndBetaReduce varInfo args >>= \case
             Nothing -> pure t
             Just fullyApplied -> do
+              thresh <- view iiInlineThreshold
               let
-                -- Inline only if the size is no bigger than not inlining.
-                sizeIsOk = termSize fullyApplied <= termSize t
+                -- Inline only if the size is no bigger than not inlining plus threshold.
+                sizeIsOk = termSize fullyApplied <= termSize t + max 0 (fromIntegral thresh)
```
ana-pantilie, line 511: Hmm, so the `size` is local in the sense that it checks that the size didn't increase for each `inline` iteration by more than the amount.
I guess that it could lead to other optimisations being applied, therefore reducing the final size, or even potentially increasing the final size by more than the threshold! I would imagine that the worst case scenario is that the size increases by the threshold amount each iteration, so the final size increases by `numIterations * threshold`, this of course assumes that the other optimisations don't increase program size, or do they?

In any case, from the documentation my understanding was that the size threshold is global: after all inlining has been performed the final size of the program is no larger than `originalSize + 5`. At least, I think this needs to be clarified.

ana-pantilie (commented):
It looks like it gives good results in practice, or at least on the test cases we have in our repo, but I think that the description makes it sound deceptively safe, see my comments.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
zliu41: > In any case, from the documentation my understanding was that the size threshold is global

I am intentionally _not_ specifying in the documentation how the value of this flag translates to program size! Instead I'm simply saying "Higher values make the inliner more willing to inline". How do you propose modifying the documentation (without making it too verbose)?

> I guess that it could lead to other optimisations being applied, therefore reducing the final size

It could also allows to remove the binding. See https://github.com/IntersectMBO/plutus/issues/6975.

ana-pantilie: Yes I agree it's tricky to explain without getting too technical. Either we could say what you just said, that it's not a direct equivalent of program size (but then, what is it? :D ), or we could give a bit more information. Could we just mention that it specifies the maximum amount the program may increase in size per inline iteration? That, together with the idea of calling it as you suggested `inline-callsite-growth` should clear up the confusion. If there is slightly more space in the docs, maybe we could also say that it may eventually even decrease program size.

zliu41: rephrased

ana-pantilie (approved):
Looks great!

PR #6985, opened 2025-03-27
Agda metatheory as an Agda library. 

ramsay-t:
To use the plutus-metatheory package as an Agda lib it just needs an `agda-lib` file in the root (and to be in your `~/.agda/libraries` path...)

We can then create a tarball from this for versioned releases.


ana-pantilie (commented):
Why does this PR have an impact on the performance of our benchmarks?

ramsay-t:
> Why does this PR have an impact on the performance of our benchmarks?

It doesn't - it is showing two commits Ziyang made to master while it was open, for some reason.

PR #6987, opened 2025-03-28
Use right fold for PlutusTx.Data.List.length

zliu41:
It's faster than left fold. In general, tail recursion isn't useful in Plinth.


plutus-tx/src/PlutusTx/Data/List.hs
```diff
@@ -279,12 +279,9 @@ map f = coerce go
...
 -- | Get the length of a list.
 length :: List a -> Integer
-length (List l) = go l 0
+length (List l) = go l
   where
-    go =
-        B.caseList'
-            id
-            (\_ t -> B.addInteger 1 . go t)
+    go = BI.caseList' 0 (\_ -> B.addInteger 1 . go)
```
ana-pantilie, line 284: Oops, thanks for fixing this!

ana-pantilie (approved):
Great catch!

PR #6989, opened 2025-03-28
Agda certificate Serialisation tests and some fixes

ramsay-t:
Add some automated testing of certificate generation, and then running the certificates through Agda.


plutus-executables/TestCert.agda
ramsay-t, line 1: This is generated, so shouldn't be in git - not least because it then clashes with whatever I ran last :)

PR #6992, opened 2025-03-28
Add nil and singleton to PlutusTx.Data.List

zliu41:


plutus-tx/src/PlutusTx/Data/List.hs
```diff
@@ -258,7 +282,7 @@ foldMap f = go . coerce
                 let h' = unsafeFromBuiltinData h
                 in f h' <> go t
             )
-{-# INLINEABLE map #-}
+{-# INLINEABLE foldMap #-}
```
Unisay, line 285: 👍🏼

Unisay (approved):

ana-pantilie (approved):

PR #6994, opened 2025-03-29
Confine recursion to inner functions in PlutusTx.List

zliu41:
Confine recursion to inner functions and avoid making the top-level function recursive, whenever possible.


effectfully (approved):
Good job, I keep forgetting about this.

But I think we should probably have a PIR pass doing that instead of doing it manually. Also note that GHC with optimizations turned on usually lifts the local definition to the top level thus rendering the manual transformation useless, so we need to be careful to tell GHC not to optimize anything (but already do anyway, so it's OK).

zliu41:
> This sounds like a Plinth-idiom, worth mentioning in the user documentation.
Do you want to add it to the [Other optimization techniques](https://plutus.cardano.intersectmbo.org/docs/working-with-scripts/other-optimization-techniques#) maybe?

Haskell has the same idiom. But sure, it doesn't hurt to mention it in the user guide.

PR #6997, opened 2025-03-31
Add PlutusTx.Data.List.caseList and caseList'

zliu41:
for matching on Lists. Currently the only way to match on a List is to either convert it to BuiltinList BuiltinData, or use the inefficient `uncons`.


ana-pantilie (approved):
Wow, interesting! Thank you for also adding the note which documents why it's more efficient.

plutus-tx/src/PlutusTx/Data/List.hs
```diff
@@ -98,13 +100,64 @@ instance (UnsafeFromData a, Pretty a) => Pretty (List a) where
...
+    {-# INLINEABLE mempty #-}
+
+{- Note [Making the cons case non-strict in caseList and caseList']
+
+A function parameter's strictness is in theory irrelevant to the Plinth compiler.
+However, it does affect how GHC compiles the function's callsites to GHC Core.
+
+In this instance of `caseList`/`caseList'`, if `c` is strict, is passed a function
+that calls `caseList`/`caseList'`, GHC will end up creating two mutually recursive
+bindings for the application of `caseList`/`caseList'`. Since our inliner currently
```
effectfully, line 126: I didn't know it worked this way. Thank you for the Note.

plutus-tx/src/PlutusTx/Data/List.hs
```diff
@@ -98,13 +100,64 @@ instance (UnsafeFromData a, Pretty a) => Pretty (List a) where
...
+  :: forall a r
+   . (UnsafeFromData a)
+  => (() -> r)
+  -- ^ Nil case
+  -> (a -> List a -> r)
+  -- ^ Cons case
+  -> List a
+  -> r
+-- See Note [Making the cons case non-strict in caseList and caseList']
+caseList n ~c (List l) = B.caseList @BuiltinData n (\x -> c (unsafeFromBuiltinData x) . List) l
```
effectfully, line 142: Use `coerce` over the `B.caseList` application to avoid binding `l` to save on an apply-lambda pair?

plutus-tx/src/PlutusTx/Data/List.hs
```diff
@@ -245,58 +261,51 @@ mapMaybe f = go . coerce
...
+            (\h t -> f h <> go t)
 {-# INLINEABLE foldMap #-}
 
 -- | Map a function over a list.
 -- Warning: this function can be very inefficient if the list contains elements
 -- that are expensive to decode from 'BuiltinData', or if the result of applying
 -- 'f' is expensive to encode to 'BuiltinData'.
 map :: (UnsafeFromData a, ToData b) => (a -> b) -> List a -> List b
-map f = coerce go
+map f = List . go
```
effectfully, line 301: `coerce go` should be faster than `List . go`. At least in the general case.

plutus-tx/src/PlutusTx/Data/List.hs
```diff
@@ -701,4 +704,28 @@ partition pred1 l =
...
+
+fromBuiltinList :: BuiltinList BuiltinData -> List a
+fromBuiltinList = coerce
+{-# INLINEABLE fromBuiltinList #-}
+
+-- | Convert a data-backed list to a sums of products list.
+-- Warning: this function can be very inefficient if the list contains elements
+-- that are expensive to decode from 'BuiltinData'.
+toSOP :: (UnsafeFromData a) => List a -> [a]
+toSOP = go . coerce
```
effectfully, line 719: Same logic: `coerce go` is more efficient than `go . coerce`. Quoting `base`:

> -- See Note [Function coercion]
> (#.) :: Coercible b c => (b -> c) -> (a -> b) -> (a -> c)
> (#.) _f = coerce
> {-# INLINE (#.) #-}
>  
> {-
> Note [Function coercion]
>  
> Several functions here use (#.) instead of (.) to avoid potential efficiency
> problems relating to #7542. The problem, in a nutshell:
>  
> If N is a newtype constructor, then N x will always have the same
> representation as x (something similar applies for a newtype deconstructor).
> However, if f is a function,
>  
> N . f = \x -> N (f x)
>  
> This looks almost the same as f, but the eta expansion lifts it--the lhs could
> be _|_, but the rhs never is. This can lead to very inefficient code.  Thus we
> steal a technique from Shachaf and Edward Kmett and adapt it to the current
> (rather clean) setting. Instead of using  N . f,  we use  N #. f, which is
> just
>  
> coerce f `asTypeOf` (N . f)
>  
> That is, we just *pretend* that f has the right type, and thanks to the safety
> of coerce, the type checker guarantees that nothing really goes wrong. We still
> have to be a bit careful, though: remember that #. completely ignores the
> *value* of its left operand.
> -}

Applies to our stuff as well.

effectfully (approved):

plutus-tx/src/PlutusTx/Data/List.hs
zliu41: For some reason, doing what you suggested make the mutual recursion appear again. I'll make a branch to demonstrate.

This is quite annoying - we should find a way to inline recursive bindings.

plutus-tx/src/PlutusTx/Data/List.hs
zliu41: I updated this one as you suggested, but I doubt it makes a difference - the plinth compiler just ignores coercions, whether they are on types or functions.

PR #6998, opened 2025-03-31
Don't run cabal-build workflow on PRs

zeme-wana:


Unisay:
I'd appreciate you adding descriptions to tasks: I'd like to know what problem this solution is supposed to fix.

Issue #6602, opened 2024-10-28
`Case`ing on values of built-in types


effectfully:
We really want to have faster `Data` processing, but as per #6225 this is quite hard. Matching on integers is very inefficient too as one needs to perform repeated calls to `IfThenElse` and `EqualsInteger` (the possibility of using SOPs for matching on integers to solve the issue was brought up [here](https://github.com/IntersectMBO/plutus/issues/5711#issuecomment-1908255143)). Matching on built-in booleans is also [inefficient](https://github.com/IntersectMBO/plutus/issues/6578), particularly from the size point of view.

I feel like we can kill all these birds with a single stone: allow `case` to work on values of built-in types.

## `Bool`

Take booleans, we could allow

```
case b
  branchFalse
  branchTrue
```

where

- `b` is a `Bool`
- `branchFalse` is executed if `b` is `False`
- `branchTrue` is executed if `b` is `True`

(I'm making `branchFalse` come before `branchTrue` not only because `false` corresponds to `0` and `true` corresponds to `1`, but also because the `false` case is more likely to terminate execution, hence being pretty-printed an average program would be more readable with the throwing clause appearing first. But we can make the ordering the same as in `IfThenElse` instead if we feel strongly about it, I personally don't care much)

This is as close as it gets to extending the UPLC AST with a constructor for `IfThenElse` without actually doing that. It should also be easy to implement, since `Case` is currently defined as

```haskell
    | Case !ann !(Term name uni fun ann) !(Vector (Term name uni fun ann))
```

and all we need is to teach the CEK machine how to extract `Term`s from a `Vector` (plus check its length) and how to handle them depending on the type tag of the given constant.

Note how we would be able to address everything from #6598 without introducing any new builtins. For example if we had

```haskell
assertAndContinueOrTrace :: string -> bool -> a -> a
```

and wanted to write `force (assertAndContinueOrTrace msg b (delay x))`, we could instead simply write

```
case b
  (trace msg error)
  x
```

This AST is of roughly the same size, but no `force`-`delay` bookkeeping is required, just like the additional builtin. The latter code is concise and on-point.

## `Integer`

For integers we'll allow

```
case i
  branch0
  branch1
  branch2
  ...
  branchN
```

This alone will optimize `Data` processing, because one will be able to write (pseudocode)

```
let (i, args) = unConstrData d
in case i
  (applyToList branch0 args)
  (applyToList branch1 args)
  ...
  (applyToList branchN args)
```

[...]

PR #7010, opened 2025-04-03
[AST] [Performance] Use 'SmallArray' instead of 'Vector'

effectfully:
#6602 will require optimizing `case`, so let's try doing that separately.


effectfully:
(the results should be disregarded, I screwed up the implementation and made it slower than it can be)

effectfully:
#7012 looks better for `nofib` and is more "conservative" of a change, since `Vector` is backed by `Array`, so let's prefer that one in favor of this PR.

PR #7012, opened 2025-04-03
[AST] [Performance] Use 'Array' instead of 'Vector'

effectfully:
Same as #7010 except using `Array` rather than `SmallArray`.

### DIFF
--- a/plutus-core/plutus-ir/src/PlutusIR/Purity.hs
+++ b/plutus-core/plutus-ir/src/PlutusIR/Purity.hs
@@ -194,7 +194,7 @@ termEvaluationOrder binfo vinfo = goTerm
       -- first the body
       goTerm b
         -- then the whole term, but this is erased so it is work-free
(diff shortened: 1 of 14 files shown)


plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -194,7 +194,7 @@ termEvaluationOrder binfo vinfo = goTerm
       -- first the body
       goTerm b
         -- then the whole term, but this is erased so it is work-free
-        <> evalThis (EvalTerm Pure WorkFree t)
+        <> evalThis (EvalTerm MaybeImpure WorkFree t)
```
Unisay, line 197: an "accidental bugfix" :) ?

effectfully:
-3% on average for `nofib` is nice and everything, but I'm not investigating the weird benchmarks behavior and not fixing all these merge conflicts in 9 files.

PR #7014, opened 2025-04-04
`PlutusTx.Eq` instance for `ScriptPurpose`

### LINKED ISSUE #7002: Adding `PlutusTx.Eq` instance for `ScriptPurpose`

To be able to lookup for redeemers within the field `txInfoRedeemers` of script contexts, one has to rely on the (non-existing) Plutus equality instance between script purposes. This instance used to be present in V1 and V2, which have the same `ScriptPurpose` type, but seems absent from V3, where the type has been enlarged. This instance is more involved, as it goes quite deep and relies on equality between maps, but should likely be present nonetheless. 

The only alternative is to implement ad-hoc lookup and/or equality instances whenever relying on this script context field in Plutus scripts.

mmontin:
This is an attempt to fix #7002.

The content of this PR is straightforward: it adds `PlutusTx.Eq` instances to the relevant components of the PlutusV3 `ScriptContext` to eventually reach the required instance for `ScriptPurpose`.

This instance is necessary to be able to directly lookup into the redeemer map of the `ScriptContext`.

The only arguable part of this contribution might be my take on comparing map: I decided to deem them equal when their sorted lists of entries (from which duplicates are removed) are equal. If this is not satisfactory, another idea is to ensure they return the same value when looking up any key from either maps. This could yield better performance, although this is arguable if that's necessary considering the usual size of the redeemer map. I can switch to this implementation if necessary.

Feel free to let me know if there are better ways of achieving the same result, or if this instance is not present for a specific reason.


plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs
```diff
@@ -407,6 +413,26 @@ data GovernanceAction
...
+    govId PlutusTx.== govId' PlutusTx.&& cp PlutusTx.== cp' PlutusTx.&& sh PlutusTx.== sh'
+  HardForkInitiation govId vers == HardForkInitiation govId' vers' =
+    govId PlutusTx.== govId' PlutusTx.&& vers PlutusTx.== vers'
+  TreasuryWithdrawals wthdrl sh == TreasuryWithdrawals wthdrl' sh' =
+    wthdrl PlutusTx.== wthdrl' PlutusTx.&& sh PlutusTx.== sh'
+  NoConfidence govId == NoConfidence govId' =
+    govId PlutusTx.== govId'
+  UpdateCommittee govId toDel toAdd quorum == UpdateCommittee govId' toDel' toAdd' quorum' =
+    govId PlutusTx.== govId'
+      PlutusTx.&& List.sort (List.nub toDel) PlutusTx.== List.sort (List.nub toDel')
```
zliu41, line 428: This is also problematic - I imagine there are cases where you just want the plain `==`.

zliu41 (commented):
> I decided to deem them equal when their sorted lists of entries (from which duplicates are removed) are equal. If this is not satisfactory, another idea is to ensure they return the same value when looking up any key from either maps.

There are at least three reasonable ways to define equalty for Maps. Two are as you described, and the third is the plain vanilla equality check, useful if you want a fast comparison (it is O(n), while the other two are at least O(n^2)).

If there's more than one reasonable way to do something, then typeclass is not a good fit. It's better to define them as regular functions. You can then convert the Map to a list, and use `List.find`.

mmontin:
Thanks @zliu41 for your comment. I agree that there are several sensible ways to define equality between maps and thus the type class approach is not optimal. That might be the reason why this `Eq` instance was not present in the first place. 

However, the instance is required to lookup in the redeemer maps, so even if I take the function approach to implement different comparisons, one of them will be needed by default to implement this instance.

Maybe the easiest approach would be to rely on vanilla equality, however, this means that developers will need to be aware of how exactly the ledger creates the script context to rely on proper lookup. I can see many reason why that's not really a good idea. 

Another approach, which you hinted in your message, is to provide lookup helpers in `PlutusLedgerApi.V3.Contexts` which would rely on different kinds of equality. Or simply document properly why this instance is not present and how users are expected to search for redeemers in this map. 

All of this maybe indicates that having a map where equality between keys is not easily (and uniquely defined) is maybe not such a good idea. With former V1 and V2 purposes (which were much simpler), this made sense but it's arguable if it still does.

Let me know your thoughts and how you think I should proceed with this PR.

zliu41:
There are two different things here

1. Why `Map`. The `Map` is supposed to mean: there's no duplicate keys. It is guaranteed that for each unique ScriptContext, there is only one Redeemer. The question, of course, is what exactly are "duplicate keys". This is not well documented (at least not in this module - maybe it is mentioned in the ledger spec) at the moment.
2. How to look up a redeemer. This is a different question: even if we have a precise definition of ScriptPurpose equality, you still don't want to use `Map.lookup`, and rely on the `Eq` instance, because it is usually too inefficient. For example, if I just want to find the redeemer of a specific minting script, my lookup predicate should simply be
  ```haskell
case purpose of 
    Minting s -> <check further>
    _ -> False
  ```
I wouldn't want to include any of the code comparing `Spending`, `Voting`, `Proposing` and so on in my script, which would unnecessarily increase my script size, even if it isn't executed at runtime. So for lookup, you should generally use `List.find` with your own predicate.

We could certainly use some documentation in both respects. If you would like to contribute to it that would be awesome.

mmontin:
Thanks for your feedback @zliu41. I have to agree with your that this generic Eq instance would be both:
* difficult to get right due to various possible implementations
* inefficient to use for Plutus developers
As a result I am closing this PR and the related issue. I will see if I find the time to contribute in terms of documentation, as I believe this is worth mentioning for people like me expecting this instance to exist.

PR #7016, opened 2025-04-04
Doc-fix, Add required dependency for the off-chain example code 

schwarzer-swan:
off-chain project fails to build as it is missing a dependency.
## To verify the need for fix:
Follow the instruction in **[Generating Keys and Addresses](https://plutus.cardano.intersectmbo.org/docs/auction-smart-contract/end-to-end/generating-keys)**
This will give an error that you're missing the `@harmoniclabs/pair` dependency

## To verify the fix:
Follow the same instruction from the patch branch

### DIFF
--- a/doc/docusaurus/docs/auction-smart-contract/end-to-end/generating-keys.md
+++ b/doc/docusaurus/docs/auction-smart-contract/end-to-end/generating-keys.md
@@ -14,6 +14,7 @@ git clone git@github.com:IntersectMBO/plinth-template.git on-chain
 mkdir off-chain && cd $_
 yarn init -y
 yarn add @meshsdk/core
+yarn add @harmoniclabs/pair
 yarn add cbor
 ```
 


Unisay (approved):
LGTM, thanks for the fix!

schwarzer-swan:
Thank you @Unisay. On a somewhat related note, I find this example disjointed.  On the one hand, the `on-chain` code requires me to work with `nix`, which I think it is the right thing to do. However, why not go all the way. Why not make `cardano-node` node available form the `nix shell` and reduce the time it takes to create a development environment?   Instead, I have to jump through another hoop and deal with block-frost and get an API key from them, then go to the correct `faucet` to get the required funds.

schwarzer-swan:
> > Thank you @Unisay. On a somewhat related note, I find this example disjointed. On the one hand, the `on-chain` code requires me to work with `nix`, which I think it is the right thing to do. However, why not go all the way. Why not make `cardano-node` available form the `nix shell` and reduce the time it takes to create a development environment? Instead, I have to jump through another hoop and deal with block-frost and get an API key from them, then go to the correct `faucet` to get the required funds, before I could deploy a simple contract
> 
> I see your point, however, please also consider the following points:
> 
> * We want users running this example to see their transactions with blockchain explorers, so this means it has to use one of the public testnets (preprod).
> * With the blockfrost backend one doesn't have to wait until the local `cardano-node` has synced itself, can get blocks and send transactions immediately.
> * Running E2E example requires test funds, when using preprod a faucet is the only reliable way to get them.
> 
> @zliu41 maybe you can come up with more reasons?

Those are very good reasons, thank you

schwarzer-swan:
@Unisay could you please add the label `No Changelog Required` ? I don't have the access to do that

PR #7017, opened 2025-04-08
Metatheory for dropList

kwxm:
This adds the  `dropList` builtin to the Agda metatheory.

### DIFF
--- a/plutus-metatheory/src/Builtin.lagda.md
+++ b/plutus-metatheory/src/Builtin.lagda.md
@@ -156,6 +156,8 @@ data Builtin : Set where
   ripemd-160                      : Builtin
   -- Modular Exponentiation
   expModInteger                   : Builtin
+  -- DropList
+  dropList                        : Builtin
 ```
 
 ## Signatures
@@ -337,7 +339,7 @@ sig n⋆ n♯ (t₃ ∷ t₂ ∷ t₁) tᵣ
     signature countSetBits                    = ∙ [ bytestring ↑ ]⟶  integer ↑
     signature findFirstSetBit                 = ∙ [ bytestring ↑ ]⟶  integer ↑
     signature expModInteger                   = ∙ [ integer ↑ , integer ↑ , integer ↑ ]⟶  integer ↑
-
+    signature dropList                        = ∀a [ integer ↑ , list a ]⟶ list a
 open SugaredSignature using (signature) public
 
 -- The arity of a builtin, according to its signature.
@@ -440,6 +442,7 @@ Each Agda built-in name must be mapped to a Haskell name.
                                           | FindFirstSetBit
                                           | Ripemd_160
                                           | ExpModInteger
+                                          | DropList
                                           ) #-}
 ```
 
@@ -641,6 +644,7 @@ postulate
 
 -- no binding needed for appendStr
 -- no binding needed for traceStr
+-- See Utils.List for the implementation of dropList
 ```
 
 Equality of Builtins is decidable.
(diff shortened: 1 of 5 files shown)


plutus-metatheory/src/Builtin.lagda.md
```diff
@@ -641,6 +645,7 @@ postulate
 
 -- no binding needed for appendStr
 -- no binding needed for traceStr
+-- See Utils.List for the implementation of dropList
```
ana-pantilie, line 648: Why implement it there instead of here directly in Haskell using the `{-# COMPILE GHC ... #-}` pragma?

kwxm: Because the UPLC list type was already represented by an Agda list type and the [existing list builtins](https://github.com/IntersectMBO/plutus/blob/f0da50da0a3e28a16cc331c73a4786463489b4a1/plutus-metatheory/src/Untyped/CEK.lagda.md?plain=1#L348) use that representation.  When I started trying to do this I tried to use the Haskell version directly but everything immediately went wrong because the arguments had the wrong type.  

Some of the UPLC types like [bytestring](https://github.com/IntersectMBO/plutus/blob/d8904bcb51677e0c7f2cd6a044f2aa8bed18085c/plutus-metatheory/src/Utils.lagda.md?plain=1#L174) are just (I think) represented as sort of opaque things that correspond to Haskell types but others like integer and list and pair are represented by Agda types.  I'm not sure why that is: maybe they correspond neatly to stuff that already exists in Agda but there's no sensible existing Agda type for `bytestring` etc.

ana-pantilie: Yep, that's actually a problem I'm working on fixing!

I'm just trying to make sure that the Agda builtins get compiled to whatever way is expected. I see that:
```
-- no binding needed for addition
-- no binding needed for subtract
-- no binding needed for multiply

{-# COMPILE GHC div  = div  #-}
{-# COMPILE GHC quot = quot #-}
{-# COMPILE GHC rem  = rem  #-}
{-# COMPILE GHC mod  = mod  #-}
```
So for `div` and so the code gets a direct Haskell translation. I'm confused about the `no binding needed for addition`, how does it know to associate the right signature/builtin to the right implementation?

kwxm: I was trying to work out exactly what Agda's `/` and `%` do but it's a bit tricky. The problem is what happens when there are negative arguments.  What the Plutus operations do is described [here](https://github.com/IntersectMBO/plutus/blob/f810009b2320e419d15e1f622d008dca79839b3e/plutus-tx/src/PlutusTx/Ratio.hs#L371) but I'm having trouble working out what the Agda functions do because they convert negative arguments into positive ones, do everything in Nat, and then adjust the answer in some way to deal with the negativity.  

 because everything has to be converted into naturals and then converted back and maybe adjusted

kwxm:
This will close https://github.com/IntersectMBO/plutus-private/issues/1428

plutus-metatheory/src/Builtin.lagda.md
ana-pantilie: Oh, weird, so there might be a discrepancy between the Agda implementation and the Haskell one, right? If yes, then could you open an issue about this?

ana-pantilie (approved):
Thank you for this!

ana-pantilie:
> This will close https://github.com/IntersectMBO/plutus-private/issues/1428

I added "Fixes https://github.com/IntersectMBO/plutus-private/issues/1428" to the PR description. This tells Github to automatically close the issue when the PR is merged.

plutus-metatheory/src/Builtin.lagda.md
kwxm: > Oh, weird, so there might be a discrepancy between the Agda implementation and the Haskell one, right? If yes, then could you open an issue about this?

See https://github.com/IntersectMBO/plutus-private/issues/1522.

PR #7022, opened 2025-04-08
UPLC Inliner: effect safety

Unisay:
Added test coverage before changing this functionality for [#1492](https://github.com/IntersectMBO/plutus-private/issues/1492)


plutus-core/untyped-plutus-core/test/Transform/Inline/Spec.hs
```diff
@@ -0,0 +1,100 @@
...
+      , _iiHints = InlineHints \_ann _name -> MayInline
+      , _iiBuiltinSemanticsVariant = def
+      , _iiInlineConstants = True
+      , _iiInlineCallsiteGrowth = Size 1_000_000
+      }
+
+  initialState :: S Name DefaultUni DefaultFun ()
+  initialState = S{_subst = Subst (TermEnv mempty), _vars = mempty}
+
+-- Term construction: ----------------------------------------------------------
```
zliu41, line 88: Instead of placing the term construction functions here, I suggest taking this opportunity to organize the test code better. There are many test cases in UPLC tests that involve constructing terms - see `Transform.Simplify` and `Transform.CaseOfCase.Test`. If you are going to add these helper functions, put them in a test library so that it can be used in all tests.

Also, make the module names consistent - e.g. `Transform.CaseOfCase.Test` vs. `Transform.Inline.Spec`.

plutus-core/untyped-plutus-core/test/Transform/Inline/Spec.hs
```diff
@@ -0,0 +1,100 @@
...
+
+--------------------------------------------------------------------------------
+-- Tests -----------------------------------------------------------------------
+
+test_inline :: TestTree
+test_inline =
+  testGroup
+    "isFirstVarBeforeEffects"
+    [ testCase "before effects" testVarBeforeEffects
+    , testCase "after effects" testVarAfterEffects
```
zliu41, line 33: While these are useful tests on their own, they don't seem to provide any additional value compared to a simple `goldenEvalOrder` test (which is simpler, and which we already have), do they?

They also don't seem relevant to the issue you are working on: determining whether something will be eventually evaluated.

Unisay: > they don't seem to provide any additional value compared to a simple `goldenEvalOrder` test (which is simpler, and which we already have), do they?

IMO they do: they specify existing behavior of `isVarBeforeEffects` in a reasonably granular (meaning individual behaviors fail separately) way. 

I wrote a test to "fix" the main property of the `isVarBeforeEffects` function, such that if it violates its own contract with clients then a corresponding test fails. Such test doesn't make an assumption that `isVarBeforeEffects` uses `termEvalOrder` internally, by design.

The `goldenEvalOrder` specifies behavior of  a different function, which should be (and is) tested elsewhere. 

I'd also argue that golden tests are less useful as they are mini-integration tests, because they fail for a multitude of reasons. What I am optimizing for is test precision (granularity): I want a unit test to fail for one reason only (the one reason is: property this test is verifying is not held up), and keep succeeding when non-essential concerns change, like, for example, pretty-printing. Otherwise such test is called [fragile](http://xunitpatterns.com/Fragile%20Test.html).

Unisay: > They also don't seem relevant to the issue you are working on: determining whether something will be eventually evaluated.

You are probably right, but I am simply not sure if it will turn out to be relevant.  
`effectSafe` function seem to be relevant, and I first wrote these tests for it. Later I saw that `effectSafe`'s behavior fully depends on the `isVarBeforeEffects` behaviour is `purity :: Bool` parameter is fixed to `False`. 
I presume I'll be changing the `effectSafe` for my task. If I end-up extending the `effectSafe` I might re-point these tests to it, will see.

Anyway, covering existing functions with tests is also my way of understanding them.

plutus-core/untyped-plutus-core/test/Transform/Inline/Spec.hs
Unisay: I thought about making a test-reusable term building facility and using it in all tests. I decided to postpone it to other times for a few reasons: 1. not to broaden the scope of change for this particular task. 2. to accumulate some term-building code here and there before coming up with an abstraction. This way abstractions usually come out more adequate, as they are well informed by the accumulated usages.

Unisay: I made another PR with a general UPLC term construction functions that are general enough to be useful in other modulel (I demonstrated this usefulness by applying those helpers in the `Transform.Simplify` spec). This way term construction helpers from this module are replaced with the helpers from the `UntypedPlutusCore.Test.Term.Construction`.

Unisay: In which sense to you think that `mkIterAppNoAnn` is "better"?
```haskell
term = mkIterAppNoAnn (var f) [var a, var b, var c]
```
vs. 
```haskell
term = var f `app` var a `app` var b `app` var c
```
I also don't see how is it more general: I can't imagine anything that `mkIterAppNoAnn` can express that `ann` can't.

zliu41: Regardless of which is better - a typical way of introducing tech debt is to duplicate some functionality that's already there, such that two slightly different APIs co-exist, and different code use different ones.

Maybe yours is slightly better (which is questionable - for example, one way `mkIterAppNoAnn` is better is that it is more similar to the GHC API for constructing GHC Core terms - consistency is good), but I think the right approach is (1) use the exiting api for this PR (2) open an issue proposing a new API (3) let's review and agree on it, before implementing the new API and retiring the old one.

Unisay:
I have disabled logs preservation in Marlowe validators and witnessed Ex Budget improvement:
```
    +--------------------++-------------+--------------+--------+
    |                    ||   CPU units | Memory units |   Size |
    +====================++=============+==============+========+
    | Baseline (average) || 540 460 151 |    1 784 424 |  7 945 |
    |   Actual (average) || 533 611 472 |    1 742 839 |  7 831 |
    |              Delta ||      -1.27% |       -2.33% | -1.43% |
    +--------------------++-------------+--------------+--------+
```

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -419,9 +482,15 @@ effectSafe
...
   -- ^ is it pure?
   -> InlineM name uni fun a Bool
-effectSafe body n purity = do
-  immediatelyEvaluated <- isFirstVarBeforeEffects n body
-  pure $ purity || immediatelyEvaluated
+effectSafe body n termIsPure = do
+  preserveLogging <- view iiPreserveLogging
+  builtinSemantics <- view iiBuiltinSemanticsVariant
+  return $
+    termIsPure
```
zliu41, line 489: Update `Note [Inlining and purity]` and point to it here.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -419,9 +482,15 @@ effectSafe
...
+      || isFirstVarBeforeEffects builtinSemantics n body
+      || (not preserveLogging && isVarImmediatelyEvaluated n body)
+ where
+  isVarImmediatelyEvaluated name = maybe False not . isVarDelayed name
```
zliu41, line 493: `isVarImmediatelyEvaluated` -> `isVarEventuallyEvaluated` (or just `eventuallyEvaluated` since it is too long) since the var does not have to be the first thing that gets evaluated

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -411,6 +432,48 @@ isFirstVarBeforeEffects n t = do
...
+  * inside a 'delay' term
+  * inside a lambda body
+  * inside a case branch
+-}
+isVarDelayed
+  :: forall name uni fun a
+   . (InliningConstraints name uni fun)
+  => name
+  -> Term name uni fun a
+  -> Maybe Bool
```
zliu41, line 446: If this function checks `if all the variable occurrences are "delayed"`, then it's better to return `Bool` rather than `Maybe Bool`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -411,6 +432,48 @@ isFirstVarBeforeEffects n t = do
...
+        then delayed : usages
+        else usages
+    LamAbs _ann _paramName body ->
+      go True usages body
+    Apply _ann t1 t2 ->
+      go delayed usages t1 ++ go delayed usages t2
+    Force _ann t ->
+      go delayed usages t
+    Delay _ann t ->
+      go True usages t
```
zliu41, line 465: I think this function is more complex than necessary. You don't need to recurse here if you encouter a `Delay`. We don't care about what's inside `Delay`.

You can make this function return `True` if at least one occurrence is _not_ delayed. Then, avoid descending into subterms of Delay, Lambda and Case branches. Then in the Var case, if it is match, just return True.

Unisay: > You don't need to recurse here if you encouter a Delay.
Good point, will do.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
Unisay: I can do it, but this makes the function less general and therefore less potentially applicable outside of the inliner, since the choice of default value is determined by its usage in the inliner.

zliu41: I think that's very premature generalization - there's a good chance it will never be needed outside of the inliner, and even if it is needed by something else, the simplified version may still be sufficient.

Unisay: 👌🏼

plutus-core/testlib/UntypedPlutusCore/Test/Term/Construction.hs
```diff
@@ -0,0 +1,515 @@
+{-# OPTIONS_GHC -Wno-simplifiable-class-constraints #-}
+
+module UntypedPlutusCore.Test.Term.Construction where
```
zliu41, line 3: Put this in a separate PR. The rest of the PR looks good, but this API has a number of issues, and needs to be reviewed separately.

First, this is only acceptable to me if it replaces `MkPlc`. Otherwise we shouldn't have two APIs that do the same thing. But it doesn't: `MkPlc` is polymorphic and can be used to construct PIR, TPLC and UPLC terms. This can't.

plutus-core/testlib/UntypedPlutusCore/Test/Term/Construction.hs
```diff
@@ -0,0 +1,515 @@
...
+--------------------------------------------------------------------------------
+-- Unique names
+--------------------------------------------------------------------------------
+
+uniqueNames2
+  :: String
+  -> String
+  -> (Name, Name)
+uniqueNames2 n1 n2 =
+  (name n1 0, name n2 1)
```
zliu41, line 444: Second, this is a strange way of getting unique names. In most if not every other API (e.g., template haskell API, GHC API), getting uinque names is a monadic action. This way the names are truly unique. In your method, calling `uniqueNames2 "a" "b"` the second time returns the same names, which are no longer unique.

plutus-core/testlib/UntypedPlutusCore/Test/Term/Construction.hs
```diff
@@ -0,0 +1,515 @@
...
+
+chooseData :: T -> T -> T -> T -> T -> T -> T
+chooseData d constr m xs i b =
+  force (builtin ChooseData)
+    `app` d
+    `app` constr
+    `app` m
+    `app` xs
+    `app` i
+    `app` b
```
zliu41, line 235: Also: the heavy use of infix functions like `` `app` `` looks visually unappealing.

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
```diff
@@ -1,206 +1,179 @@
...
 -}
 inlineImpure4 :: Term Name PLC.DefaultUni PLC.DefaultFun ()
 inlineImpure4 =
-  mkInlinePurityTest $
-    Force () . Force () . Force () . Delay () . Delay () . Var () <$> freshName "a"
+  mkInlinePurityTest $ force . force . force . delay . delay . var $ name "a" 0
 
-{- | @(\a -> f (a 0 1) (a 2)) (\x y -> g x y)@
+inlineImpure5 :: Term Name PLC.DefaultUni PLC.DefaultFun ()
+inlineImpure5 =
```
zliu41, line 168: Have you verified that before your change, the inliner won't inline this? I think that should be the case but it's worth verifying to be certain.

plutus-core/untyped-plutus-core/test/Transform/Inline/Spec.hs
```diff
@@ -0,0 +1,161 @@
+{-# LANGUAGE BlockArguments      #-}
+{-# LANGUAGE NumericUnderscores  #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Transform.Inline.Spec where
```
zliu41, line 6: The test module layout is becoming confusing, because we have a `Transform.Inline.Spec`, but then there are also a bunch of inliner tests in `Transform.Simplify` (and the module name doesn't have `Spec`).

Is there a quick way to make things less confusing? If not, open a new issue and add a TODO.

zliu41 (commented):
The title of the PR should mention UPLC, as you are only updating the UPLC inliner, not the PIR inliner.

zliu41:
Also this should have a changelog entry. Please do not default to using the "no changelog required" tag. If you are in doubt about whether a changelog entry is needed, err on the side of adding one.

plutus-core/testlib/UntypedPlutusCore/Test/Term/Construction.hs
Unisay: > Put this in a separate PR. 

It already was in a [separate PR](https://github.com/IntersectMBO/plutus/pull/7028). 

> this API has a number of issues, and needs to be reviewed separately.

It got reviewed by Ana and merged into this PR so that I can use these functions.

Unisay: > First, this is only acceptable to me if it replaces MkPlc. ...  Otherwise we shouldn't have two APIs that do the same thing. 

My problem with `MkPlc` is that it doesn't allow me to construct UPLC terms in a way that's clean and readable, e.g. `MkPlc` example is very awkward, noisy, hard to reason:
```haskell
caseOfCase3 :: Term Name PLC.DefaultUni PLC.DefaultFun ()
caseOfCase3 = runQuote $ do
  b <- freshName "b"
  x <- freshName "x"
  xs <- freshName "xs"
  f <- freshName "f"
  let ite = Force () (Builtin () PLC.IfThenElse)
      true = Constr () 0 [Var () x, Var () xs]
      false = Constr () 1 []
      altTrue = Var () f
      altFalse = mkConstant @Integer () 2
      alts = V.fromList [altTrue, altFalse]
  pure $ Case () (mkIterApp ite [((), Var () b), ((), true), ((), false)]) alts
```

`UntypedPlutusCore.Test.Term.Construction` is way cleaner, easier to make sense of:
```haskell
caseOfCase3 :: Term Name PLC.DefaultUni PLC.DefaultFun ()
caseOfCase3 =
  case_ (ifThenElse (var b) (ctor 0 [var x, var xs]) ctorFalse)
    [ var f
    , int 2 
    ]
 where
  (b, x, xs, f) = uniqueNames4 "b" "x" "xs" "f"

```

Conceptually both of these APIs do "term construction" but the difference in the term-construction expressions  is a reason why I created the latter. 
IMO
```haskell
  ...
 where
  (b, x, xs, f) = uniqueNames4 "b" "x" "xs" "f"
```
is much better than:

```haskell
runQuote $ do
  b <- freshName "b"
  x <- freshName "x"
  xs <- freshName "xs"
  f <- freshName "f"
  ...
  pure $ ...
```

because:
* its short, consise, one-liner.
* it de-emphasises this secondary concern of unique names creation moving it to the bottom (`where` clause after the constructed term) whereas the latter sticks it into your face before you even get to read the term (which is the main purpose)

> MkPlc is polymorphic and can be used to construct PIR, TPLC and UPLC terms. 

This is, actually, more of a problem here. I don't need to construct PIR and TPLC terms in this tests. Polymorphism is not needed for the UPLC term construction, its actually a drag.

TLDR:
My goal is to construct clean, concise and readable UPLC terms for tests:
- Without having to put term inside a monadic context;
- Without having to deal with types and annotations;
- With a palette of pre-constructed most popular terms like builtins, true/false constructors; etc.

I don't know how to achieve this goal with what is currently in the `MkPlc`. It also doesn't make sense to add the functions I want to have to MkPlc, as they are of a different nature: created for UPLC terms for tests only: intentionally monomorphic and hardcode `ann = ()`, `DefaultFun`, `DefaultUni`.

plutus-core/testlib/UntypedPlutusCore/Test/Term/Construction.hs
Unisay: I guess its a matter of taste 🤷🏼‍♂️ 
I was thinking of having a variation of `$` operator for infix applications, something like:
```haskell
  term
    $. m
    $. xs
    $. i
    $. b
```

plutus-core/untyped-plutus-core/test/Transform/Simplify.hs
Unisay: Yes, I did.
```
inlineImpure5:           FAIL
  Test output was different from 'untyped-plutus-core/test/Transform/inlineImpure5.uplc.golden'. 
  It was:
  \a -> (\b -> addInteger (addInteger 1 a) b) (addInteger 3 a)
  Use -p '/simplify/&&/inlineImpure5/' to rerun this test only.
```

plutus-core/untyped-plutus-core/test/Transform/Inline/Spec.hs
Unisay: https://github.com/IntersectMBO/plutus/issues/7041

PR #7023, opened 2025-04-08
Explain lifting in the user guide

zliu41:
This doesn't currently have sufficient explanation.


ana-pantilie (approved):
Thank you for adding this! Helped my understanding of lifting as well.

Unisay (approved):
🥇

PR #7025, opened 2025-04-09
Conformance tests for `dropList`

kwxm:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1431.

This adds some conformance tests for `dropList`.  It can't fail (execpt for running out of budget) so the tests are quite simple.


zliu41 (approved):
Looks reasonable!

PR #7026, opened 2025-04-09
Update ghc966 to ghc967

zeme-wana:


.editorconfig-checker.json
```diff
@@ -0,0 +1,19 @@
+{
```
zeme-wana, line 1: Newer version of editorconfig-checker uses json now

zliu41: This doesn't seem equivalent to the old config - for instance it doesn't specify max line length?

zeme-wana:
Closing as obsolete

PR #7028, opened 2025-04-10
UPLC Term construction functions for testing

Unisay:
Closes https://github.com/IntersectMBO/plutus/issues/7024


plutus-core/untyped-plutus-core/test/Transform/cse2.uplc.golden
```diff
@@ -1,6 +1,5 @@
 force
-  (force
-     (ifThenElse
```
Unisay, line 3: The `force` should be applied to the builtin before its arguments are applied.

ana-pantilie (approved):
Nice!

PR #7029, opened 2025-04-10
[Builtins] Allow casing on booleans and integers

effectfully:
This adds support for casing on booleans and integers using `Case`, which is the first part of #6602.


plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -538,20 +538,32 @@ instance KnownBuiltinTypeIn DefaultUni term Integer => ReadKnownIn DefaultUni te
 
 instance UniOf term ~ DefaultUni => CaseBuiltin term DefaultUni where
     caseBuiltin (Some (ValueOf uni x)) branches = case uni of
+        -- TODO: It feels like we should support having only one branch to verify that a condition
```
Unisay, line 525: A gentle reminder that we've agreed that TODO/FIXME tags should be accompanied by a link to the corresponding github issue.

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
```diff
@@ -364,18 +367,23 @@ defineBuiltinTerms = do
...
+                        a <- freshTyName "a"
+                        b <- freshName "b"
+                        x <- freshName "x"
+                        y <- freshName "y"
+                        return
+                            . PIR.tyAbs () a (PLC.Type ())
+                            . PIR.lamAbs () b (PLC.mkTyBuiltin @_ @Bool ())
+                            . PIR.lamAbs () x (PLC.TyVar () a)
+                            . PIR.lamAbs () y (PLC.TyVar () a)
+                            $ PIR.kase ()
```
Unisay, line 383: [🧀 ](https://de.wikipedia.org/wiki/K%C3%A4se)

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -408,7 +408,8 @@ type GivenCekCosts = (?cekCosts :: CekMachineCosts)
 type GivenCekReqs uni fun ann s = (GivenCekRuntime uni fun ann, GivenCekEmitter uni fun s, GivenCekSpender uni fun s, GivenCekSlippage, GivenCekStepCounter s, GivenCekCosts)
 
 data CekUserError
-    = CekOutOfExError !ExRestrictingBudget -- ^ The final overspent (i.e. negative) budget.
+    = CaseBuiltinError Text -- ^ 'Case' over a value of a built-in type failed.
```
effectfully, line 411: I'm not sure if I should `newtype`-wrap this text into a proper error-type. I feel like this is OK?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -785,6 +790,9 @@ enterComputeCek = computeCek
         (VConstr i args) -> case (V.!?) cs (fromIntegral i) of
             Just t  -> computeCek (transferArgStack args ctx) env t
             Nothing -> throwingDischarged _StructuralError (MissingCaseBranchMachineError i) e
+        VCon val -> case caseBuiltin val cs of
+            Left err  -> throwingDischarged _OperationalError (CaseBuiltinError err) e
+            Right res -> computeCek ctx env res
```
effectfully, line 795: Main change part 1.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
```diff
@@ -47,7 +72,18 @@ processTerm = \case
           mkIterApp
             ite
             [ cond
-            , (trueAnn, Delay trueAnn (Case ann true alts))
-            , (falseAnn, Delay falseAnn (Case ann false alts))
+            , (trueAnn, Delay trueAnn . CaseReduce.processTerm $ Case ann true alts)
+            , (falseAnn, Delay falseAnn . CaseReduce.processTerm $ Case ann false alts)
```
effectfully, line 75: TODO: document.

plutus-core/plutus-core/src/PlutusCore/Builtin/Case.hs
```diff
@@ -0,0 +1,26 @@
...
+        :: UniOf term ~ uni
+        => SomeTypeIn uni
+        -> [term]
+        -> Either Text [(term, [Type TyName uni ann])]
+
+class UniOf term ~ uni => CaseBuiltin term uni where
+    -- | Given a constant with its type tag and a vector of branches, choose the appropriate branch
+    -- or fail if the constant doesn't correspond to any of the branches (or casing on constants of
+    -- this type isn't supported at all).
+    caseBuiltin :: Some (ValueOf uni) -> Vector term -> Either Text term
```
effectfully, line 26: Main change part 2.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -517,6 +521,36 @@ instance KnownBuiltinTypeIn DefaultUni term Integer => ReadKnownIn DefaultUni te
...
+            -- 'Error' anyway.
+            False | len == 1 || len == 2 -> Right $ branches Vector.! 0
+            True  |             len == 2 -> Right $ branches Vector.! 1
+            _                            -> Left  $ outOfBoundsErr x branches
+        DefaultUniInteger
+            | 0 <= x && x < toInteger len -> Right $ branches Vector.! fromInteger x
+            | otherwise                   -> Left  $ outOfBoundsErr x branches
+        _ -> Left $ display uni <> " isn't supported in 'case'"
+      where
+        !len = Vector.length branches
```
effectfully, line 552: Main change part 3.

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
```diff
@@ -364,7 +367,23 @@ defineBuiltinTerms = do
...
+                                (PLC.TyVar () a)
+                                (PIR.Var () b)
+                                [PIR.Var () y, PIR.Var () x]
```
effectfully, line 386: Main change part 4. I'm using `datatypeStyle` here, because that's what determines whether we have `Case` available or not. I'll document it if we're fine with this logic.

plutus-core/plutus-ir/src/PlutusIR/Transform/CaseReduce.hs
```diff
@@ -6,25 +6,36 @@ module PlutusIR.Transform.CaseReduce
...
 import Control.Lens (transformOf, (^?))
 import Data.List.Extras
+import GHC.IsList (fromList)
 import PlutusCore qualified as PLC
 import PlutusIR.Pass
 import PlutusIR.TypeCheck qualified as TC
 
 caseReducePass
-  :: (PLC.Typecheckable uni fun, PLC.GEq uni, Applicative m)
+  :: ( PLC.Typecheckable uni fun, CaseBuiltin (Term TyName Name uni fun a) uni
```
Unisay, line 21:
```suggestion
  :: ( PLC.Typecheckable uni fun
     , CaseBuiltin (Term TyName Name uni fun a) uni
```

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
Unisay: I couldn't find test coverage for this logic. Did you also update the tests?

Unisay (approved):
LGTM!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek.hs
```diff
@@ -107,7 +107,9 @@ evaluateCekNoEmit = Common.evaluateCekNoEmit runCekDeBruijn
 -- | Unlift a value using the CEK machine.
 -- *THIS FUNCTION IS PARTIAL if the input term contains free variables*
 readKnownCek
-    :: (ThrowableBuiltins uni fun, ReadKnown (Term Name uni fun ()) a)
+    :: ( ThrowableBuiltins uni fun, CaseBuiltin (NTerm uni fun ann) uni
```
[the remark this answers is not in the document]
effectfully: This is intentional, I often put together constraints that are related and these two are related in that they're required by the CEK machine unlike the one below.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -517,6 +521,36 @@ instance KnownBuiltinTypeIn DefaultUni term Integer => ReadKnownIn DefaultUni te
...
+    ]
+
+instance AnnotateCaseBuiltin DefaultUni where
+    annotateCaseBuiltin (SomeTypeIn uni) branches = case uni of
+        DefaultUniBool    -> Right $ map (, []) branches
+        DefaultUniInteger -> Right $ map (, []) branches
+        _                 -> Left $ display uni <> " isn't supported in 'case'"
+
+instance UniOf term ~ DefaultUni => CaseBuiltin term DefaultUni where
+    caseBuiltin (Some (ValueOf uni x)) branches = case uni of
```
zliu41, line 539: Don't you want to `INLINE` this?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -517,6 +521,36 @@ instance KnownBuiltinTypeIn DefaultUni term Integer => ReadKnownIn DefaultUni te
...
+        DefaultUniBool -> case x of
+            -- We allow there to be only one branch as long as the scrutinee is 'False'.
+            -- This is strictly to save size by not having the 'True' branch if it was gonna be
+            -- 'Error' anyway.
+            False | len == 1 || len == 2 -> Right $ branches Vector.! 0
+            True  |             len == 2 -> Right $ branches Vector.! 1
+            _                            -> Left  $ outOfBoundsErr x branches
```
zliu41, line 546: Looks like casing on SOP allows extra branches. Should it also be allowed when casing on builtins?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseReduce.hs
```diff
@@ -13,16 +15,19 @@ import Control.Lens (transformOf)
...
     return result
 
-processTerm :: Term name uni fun a -> Term name uni fun a
+processTerm :: CaseBuiltin (Term name uni fun a) uni => Term name uni fun a -> Term name uni fun a
 processTerm = \case
     Case ann (Constr _ i args) cs | Just c <- (V.!?) cs (fromIntegral i) ->
                                     mkIterApp c ((ann,) <$> args)
-    t                                                     -> t
+    Case ann (Constant _ con) cs -> case caseBuiltin con cs of
+        Left _    -> Error ann
```
zliu41, line 31: Worth a comment: why we rewrite it to `Error` here, but not in the `Constr` case.

zliu41 (changes requested):
Needs a mechanism to disable casing on booleans/integers until HF, and I don't think there currently is one

zliu41:
The simplest solution is perhaps adding a `BCaseBuiltin` step with infinite budget. Or alternatively, introduce a new plutus core versoin (1.2.0) and make the CEK machine aware of the version, but that's a bigger change and may have performance overhead.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
```diff
@@ -47,7 +71,18 @@ processTerm = \case
...
             ]
+  original@(Case annOuter (Case annInner scrut altsInner) altsOuter) ->
```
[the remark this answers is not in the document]
effectfully: It's not that orthogonal, we already had case-of-ifThenElse, we need case-of-case to have an accurate comparison here, otherwise we aren't optimizing with the new machinery the things that we were optimizing with the old one. I can still take it out if you want though.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: > Looks like casing on SOP allows extra branches.

I'm not sure if the concept of extra branches is even defined in UPLC for SOPs? There's no way to tell if `x` in `case x` is always `Constr i` with `i` being less than the given number of branches.

So it's like the `Integer` case here:

```haskell
        DefaultUniInteger
            | 0 <= x && x < toInteger len -> Right $ branches Vector.! fromInteger x
            | otherwise                   -> Left  $ outOfBoundsErr x branches
```

Do we allow extra branches? Well, who knows, we don't know if `x` can ever reach `toInteger len - 1` or not.

So I don't think there's a precedent here and we're free to make one for `Bool`.

I don't why allow extra branches. Is it useful in any way? If not, can it happen other than by mistake? If no, then I don't see a point.

plutus-core/plutus-core/src/PlutusCore/Builtin/Case.hs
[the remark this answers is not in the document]
effectfully: Hm, I feel like I had some kind of idea, but I can't possibly think what it could've been. Probably just a thinko.

Thanks for spotting! Fixed.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseReduce.hs
effectfully: I... don't know. I just thought that if it was an error, I can make it `Error` to save some size.

I can make the `Constr` case rewrite it to `Error` as well if you're fine with that.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
zliu41: Yeah, that's very valid point.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseReduce.hs
zliu41: I think there's a good reason: in the `Constr` case, index out of bounds is a structural error, and rewriting it to `Error` would change it to an operational error, so we shouldn't do that.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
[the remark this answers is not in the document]
zliu41: To make the comparison accurate, I think we should merge this change first, then rebase this PR on top of it.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseReduce.hs
effectfully: Hm, I'm not sure how much it matters, but I don't mind punishing someone who created an invalid script by not optimizing it fully. OK, I'll remove the `Error` part. Thanks!

plutus-core/plutus-core/src/PlutusCore/Builtin/Case.hs
```diff
@@ -0,0 +1,26 @@
...
+import PlutusCore.Core.Type (Type, UniOf)
+import PlutusCore.Name.Unique
+
+import Data.Text (Text)
+import Data.Vector (Vector)
+import Universe
+
+class AnnotateCaseBuiltin uni where
+    -- | Given a tag for a built-in type and a list of branches, annotate each of the branches with
+    -- its expected type or fail if casing on values of the built-in type isn't supported.
```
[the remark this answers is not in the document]
effectfully: Yes, thank you for spotting, it was the whole type initially but then I changed it without changing the docs.

plutus-core/plutus-core/src/PlutusCore/Builtin/Case.hs
```diff
@@ -0,0 +1,26 @@
...
+    annotateCaseBuiltin
+        :: UniOf term ~ uni
+        => SomeTypeIn uni
+        -> [term]
```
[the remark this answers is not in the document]
effectfully: Yes, the list part, not the `term` part. For example matching on built-in lists will require to annotate the `nil` and the `cons` branches differently.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: Doesn't really matter, because there's no place where it'd be useful to inline it, but I can annotate it with `INLINE` just in case it becomes useful in future. Thanks!

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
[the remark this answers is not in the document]
effectfully: Fair, I'll create that PR.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
```diff
@@ -47,7 +71,18 @@ processTerm = \case
...
+    maybe
+      original
+      (Case annInner scrut)
+      (do
+        constrs <- for altsInner $ \case
+          c@(Constr _ i _) -> Just (i, c)
+          _                -> Nothing
```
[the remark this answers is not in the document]
effectfully: That's a very good point, thank you.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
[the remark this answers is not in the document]
effectfully: Apparently, I've already done that investigation and forgot about it: #7046.

Should I reopen the PR? I kinda don't wanna get nerd-sniped writing tests for it given that we have those tests here anyway.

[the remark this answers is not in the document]
effectfully: Plus I'll need to change it here anyway to all case-of-case-returning-constants.

michaelpj:
This is a non-backwards-compatible change to the behaviour of a core expression type, so it's pretty tricky indeed. Possibly we need separate `Case` and `Case2` nodes and we deserialise new case expressions into the latter. From the point of view of the evaluator, it needs to support both behaviours and it needs to distinguish them. We don't have a way to plug in evaluators like we do for builtins, so otherwise I guess we'd need to add something like that (and be very careful about performance).

effectfully:
@michaelpj @zliu41 

> otherwise I guess we'd need to add something like that (and be very careful about performance).

Added in 5f3483866477eee84e04ccc81147a39eacdddc47, please review.

(not complete, some builds and tests fail, but it's enough to run the benchmarks)

effectfully:
Performance looks pretty much unaffected.

plutus-core/plutus-core/src/PlutusCore/Builtin/Case.hs
```diff
@@ -0,0 +1,51 @@
...
+        -> Either Text [(term, [Type TyName uni ann])]
+
+class CaseBuiltin uni where
+    -- | Given a constant with its type tag and a vector of branches, choose the appropriate branch
+    -- or fail if the constant doesn't correspond to any of the branches (or casing on constants of
+    -- this type isn't supported at all).
+    caseBuiltin :: UniOf term ~ uni => Some (ValueOf uni) -> Vector term -> Either Text term
+
+data CaserBuiltin uni = CaserBuiltin
```
zliu41, line 38: It seems `CaserBuiltin` should render the `CaseBuiltin` class superfluous, since an instance of `CaseBuiltin` can just be represented with a `CaserBuiltin` value?

Also, why not newtype?

zliu41 (commented):
One comment, otherwise looks good.

plutus-core/plutus-core/src/PlutusCore/Builtin/Case.hs
effectfully: We do use `CaseBuiltin` as an actual constraint

image

> Also, why not newtype?

See [this](https://github.com/IntersectMBO/plutus/blob/5f3483866477eee84e04ccc81147a39eacdddc47/plutus-core/plutus-core/src/PlutusCore/Builtin/Runtime.hs#L64-L71) for a similar case.

zliu41: > We do use CaseBuiltin as an actual constraint

Ok, so `Case` is used to define how a universe normally handles casing on builtins, and `Caser` exists so that it can vary based on protocol version. I guess that's fine, but not easy to tell, and their similar names don't help.

> See [this](https://github.com/IntersectMBO/plutus/blob/5f3483866477eee84e04ccc81147a39eacdddc47/plutus-core/plutus-core/src/PlutusCore/Builtin/Runtime.hs#L64-L71) for a similar case.

Create a Note and link to it?

zliu41 (approved):
This overall looks great. Very nice improvement to the language.

kwxm:
I'll review this soon so that I actually have some idea of what's going on.

If we're committed to releasing this then there are some other things that'll need to be done:

 * Do we have to worry about costing?  I guess it'll just be the standard machine step cost but we should think about that.
 * Specification
 * Metatheory
 * More conformance tests?  We'll need to make sure that `case` works on all the things that it's supposed to.
 * End-to-end tests.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
effectfully: We do not call an entire optimization pass here, just its single step. Added a comment:

```
            -- Here we call a single step of case-reduce in order to immediately clean up the
            -- duplication of @alts@. Otherwise optimizing case-of-case-of-case-of... would create
            -- exponential blowup of the case branches, which would eventually get deduplicated
            -- with case-reduce, but only after that exponential blowup has already slowed the
            -- optimizer down unnecessarily.
```

plutus-core/plutus-core/test/TypeSynthesis/Golden/DefaultFun/ReplicateByteString.plc.golden
effectfully, line 1: A bunch of obsolete golden files that I removed because I had lots of git conflicts creating obsolete golden files, these ones were caught in the crossfire.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
```diff
@@ -11,32 +29,42 @@
 into
 
 @
-    force ifThenElse b (delay (case (constr t) alts)) (delay (case (constr f) alts))
+    force (force ifThenElse b (delay (case (constr t) alts)) (delay (case (constr f) alts)))
```
effectfully, line 32: Fixed a documentation bug while I was here.

plutus-core/plutus-ir/test/PlutusIR/Transform/StrictLetRec/Tests/Lib.hs
```diff
@@ -87,11 +88,14 @@ evaluateUplcProgramWithTraces uplcProg =
...
-    :: MachineParameters CekMachineCosts DefaultFun (CekValue DefaultUni DefaultFun ()) =
-      mkMachineParameters def costModel
+  costModel :: CostModel CekMachineCosts BuiltinCostModel
+  costModel =
+      CostModel defaultCekMachineCostsForTesting defaultBuiltinCostModelForTesting
+
+  machineParameters :: DefaultMachineParameters
+  machineParameters =
+      -- TODO: proper semantic variant. What should def be?
+      MachineParameters def $ mkMachineVariantParameters def costModel
```
effectfully, line 97: We have a lot of this duplication where `MachineParameters` is used together with `mkMachineVariantParameters`, but I'm gonna argue that it's a feature. The right way to create a `MachineParameters` is via 

```haskell
toMachineParameters :: MajorProtocolVersion -> EvaluationContext -> DefaultMachineParameters
toMachineParameters pv (EvaluationContext ll toCaser toSemVar machParsList) =
    case lookup (toSemVar pv) machParsList of
        Nothing -> error $ Prelude.concat
            ["Internal error: ", show ll, " does not support protocol version ", show pv]
        Just machVarPars -> MachineParameters (toCaser pv) machVarPars
```

or the like and there's no duplication there. We only get duplication with "improper" `MachineParameters` creation and "worst practices should be hard".

plutus-ledger-api/src/PlutusLedgerApi/MachineParameters.hs
effectfully, line 1: This whole file needs to be deleted.

plutus-ledger-api/src/PlutusLedgerApi/V1/EvaluationContext.hs
```diff
@@ -41,10 +43,13 @@ mkEvaluationContext =
     >=> pure . toCostModelParams
     >=> mkDynEvaluationContext
       PlutusV1
+      (\pv ->
+          if pv < futurePV
```
effectfully, line 46: @zliu41 this will need to get updated once we know what `futurePV` actually is.

And in other places.

plutus-tx/src/PlutusTx/Lift.hs
effectfully, line 1: These changes are ridiculous, we need to introduce an alias for these constraint set some time.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
```diff
@@ -47,7 +75,24 @@ processTerm = \case
...
+      original
+      (Case annInner scrut)
+      (do
+        constrs <- for altsInner $ \case
+          c@(Constr _ i _)   -> Just (Left i, c)
+          c@(Constant _ val) -> Just (Right val, c)
+          _                  -> Nothing
+        -- See Note [Case-of-case and duplicating code].
+        guard $ length (nub . toList $ fmap fst constrs) == length constrs
+        pure $ constrs <&> \(_, c) -> CaseReduce.processTerm $ Case annOuter c altsOuter)
```
effectfully, line 97: I think this is overly restrictive. We should allow duplicates as long as they're of a small size. I'll implement that as a follow-up.

plutus-conformance/test-cases/uplc/evaluation/term/case/case-05/case-05.uplc
kwxm, line 1: @effectfully We probably need more conformance tests for `case` here, like testing that it does work on booleans and integers (depending on the Plutus version) and that it doesn't work on other types (but that's OK as long as it's not fully applied).  There's also the question of what happens about missing branches. 

Any new tests will probably have to be [excluded](https://github.com/IntersectMBO/plutus/blob/71afec07442fd800ff79f03a60548f5a151f7b6b/plutus-conformance/agda/Spec.hs#L119) from `agda-conformance` because I don't think the metatheory knows about `case` and  non-`constr` yet.  

In some cases in the past we've obtained end-to-end tests by promoting the conformance tests, so if these are as comprehesive as possible it might make the end-to-end tests a bit easier.

PR #7030, opened 2025-04-10
[UPLC] [Optimization] Push 'force' into 'case' of 'delay's

effectfully:
This pushes `force` into a `case` whenever all of the branches start with a `delay`.


effectfully:
@zliu41 pointed out that this is no safer than the general optimization, because here we can also turn failure into success, for example by transforming

```haskell
force (case (constr 0 [3, 4]) [delay (\x y -> x + y)])
```

so we should either give up or implement the general approach (I prefer the latter).

PR #7038, opened 2025-04-14
Add goldenBudgetAndSize test function

zeme-wana:
We previously used the `goldenBudget` and `goldenSize` functions to check a script's expected budget and size, respectively. This PR introduces a new function, `goldenBudgetAndSize`, which performs both checks.

All occurrences of `goldenBudget` have been replaced with `goldenBudgetAndSize`.
In tests where both `goldenSize` and `goldenBudget` were used, they have been consolidated into a single call to `goldenBudgetAndSize`, and the golden files have been renamed from `.budget.golden` to `.budget-and-size.golden`.

As a result, the `goldenBudget` function is no longer used anywhere, and has been removed.

Also, `String` has been replaced by `TestName` where appropriate in `PlutusCore.Test`.


zliu41 (changes requested):
Instead of making a new function `goldenBudgetAndSize`, we can just modify `goldenBudget`, so that it additionally prints the size in the same golden file. This is better than producing two separate golden files.

effectfully:
I totally agree with Ziyang.

plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -111,10 +113,13 @@ renderExcess tData mData diff =
     <> "Remaining headroom: "
     <> show diff
 
--- Budget testing
-
-goldenBudget :: TestName -> CompiledCode a -> TestNested
-goldenBudget name compiledCode = goldenUEvalBudget name [compiledCode]
+goldenBudgetAndSize :: TestName -> CompiledCode a -> TestNested
```
zeme-wana, line 116: `goldenBudget` used to check just the `ExBudget` (cpu + memory)
If we include the size, the name `goldenBudget` is now a bit misleading.
Are we sure we don't want to name this function `goldenBudgetAndSize` and the golden files `budget-and-size`?

plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -111,10 +113,13 @@ renderExcess tData mData diff =
...
+goldenBudgetAndSize name compiledCode = do
+  nestedGoldenVsDocM name ".budget-and-size" $ ppCatch $ do
+    ExBudget cpu mem <- runUPlcBudget [compiledCode]
+    size <- UPLC.programSize <$> toUPlc compiledCode
+    let contents = "cpu: " <> pretty cpu <> "\nmem: " <> pretty mem <> "\nsize: " <> pretty size
+    pure (render @Text contents)
```
zeme-wana, line 122: The above prints:
```
cpu: 203202927
mem: 641180
size: 863
```
We lose the `instance Pretty ExBudget` but if we didn't use `pretty` explicitly here, and did something like:
`pure (budget, size)` instead, we would get this: which is not very readable:
```
(({cpu: 203202927 | mem: 641180}), 863)
```

plutus-tx/testlib/PlutusTx/Test.hs
zliu41: Technically size is a form of budget since it is limited. It's slightly inconsistent, but `goldenBudgetAndSize` is just too long for my taste.

plutus-tx/testlib/PlutusTx/Test.hs
zliu41: Yeah, the former is better.

plutus-tx/testlib/PlutusTx/Test.hs
zeme-wana: Alright then, `goldenBudget` it is.

PR #7039, opened 2025-04-14
[UPLC] [Optimization] Cancel delay-force pairs

effectfully:
We can turn `delay (force t)` into just `t`, if the whole expression occurs under something blocking evaluation (`Delay`, `LamAbs`, `Case`).


plutus-benchmark/script-contexts/test/V2/9.6/dataFwdStakeTrickManual.uplc.golden
```diff
@@ -15,19 +15,16 @@
...
-                                                              rest)))))))))))
+                                         (force
+                                            (force ifThenElse
+                                               (equalsData
+                                                  r_stake_cred
+                                                  wdrlAtOne)
+                                               (delay (delay (constr 0 [])))
+                                               (delay
+                                                  (delay
+                                                     (force lookForCred
```
effectfully, line 26: How is this `delay-delay-force` not turned into just `delay`?

effectfully:
OK, this isn't semantics-preserving.

This doesn't throw an error:

```haskell
let ddf = delay (delay (force error))
    df  = force ddf
in ()
```

but after being optimized to

```haskell
let ddf = delay error
    df  = force ddf
in ()
```

it does.

So given that pushing `force` into `case` is simpler, more helpful, less likely to backfire and largely covers the same use cases and more, I'm closing this PR.

PR #7040, opened 2025-04-15
[UPLC] [Optimization] Make case-of-case clutter the output less

effectfully:
I thought this would be a good idea, but it isn't, see below.


cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests/unsorted.uplc.golden
```diff
@@ -1376,18 +1341,15 @@ program
...
-                                  (delay (delay (constr 1 []))))))))))))
+             (force ifThenElse
+                (equalsInteger eta eta)
+                (delay (constr 0 []))
+                (delay
+                   (force
+                      (force ifThenElse
+                         (lessThanEqualsInteger eta eta)
+                         (delay (constr 2 []))
+                         (delay (constr 1 []))))))))
```
effectfully, line 1352: Instead of trying to make case-of-case clever, we should just propagate `force` into the branches whenever appropriate, because that covers more cases, like the one we see here for example.

PR #7044, opened 2025-04-15
refactor: apply fourmolu to plutus-conformance

Unisay:
- [x] Wrap long lines
- [x] `fourmolu`-ize

### DIFF
--- a/plutus-conformance/src/PlutusConformance/Common.hs
+++ b/plutus-conformance/src/PlutusConformance/Common.hs
@@ -1,11 +1,13 @@
-{-# LANGUAGE BangPatterns         #-}
 {-# LANGUAGE FlexibleInstances    #-}
 {-# LANGUAGE OverloadedStrings    #-}
 {-# LANGUAGE UndecidableInstances #-}
 
-{- | Plutus conformance test suite library. -}
+-- | Plutus conformance test suite library.
 module PlutusConformance.Common where
 
+import Data.Maybe (fromJust)
+import Data.Text qualified as T
+import Data.Text.IO qualified as T
 import PlutusCore.Annotation
 import PlutusCore.Default (DefaultFun, DefaultUni)
 import PlutusCore.Error (ParserErrorBundle)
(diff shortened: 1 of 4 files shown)


plutus-conformance/src/PlutusConformance/Common.hs
```diff
@@ -58,20 +56,23 @@ type UplcProg = UPLC.Program Name DefaultUni DefaultFun ()
...
 type UplcEvaluatorFun res = UplcProg -> Maybe res
 
--- TODO: consider splitting up the evaluator with costing into a part that parses the model
--- and a part that consumes it. Currently the tests are fast enough regardless so it doesn't
--- matter.
+-- TODO: consider splitting up the evaluator with costing into a part that
+-- parses the model and a part that consumes it. Currently the tests are fast
+-- enough regardless so it doesn't matter.
+
 -- | The evaluator to be tested.
```
effectfully, line 63: Added the retarded newline.

plutus-conformance/src/PlutusConformance/Common.hs
```diff
@@ -191,75 +212,95 @@ getTestedValue eval file = do
...
-    -- whether it should labelled as `ExpectedBudgetFailure`.
-    -> IO ()
+{-| Run the UPLC evaluation tests given an `evaluator` that evaluates UPLC
+programs.
+-}
+runUplcEvalTests
+  :: UplcEvaluator
+  -- ^ The action to run the input through for the tests.
+  -> (FilePath -> Bool)
+  {- ^ A function that takes a test name and returns
```
effectfully, line 289: Why `{- ^` with the space, but `{-|` without the space? I'd prefer the latter to have a space too.

effectfully (approved):
Other than the despicable newline, this is an improvement.

plutus-conformance/src/PlutusConformance/Common.hs
Unisay: This is because in our fourmolu config we have `haddock-style: multiline-compact`, could be fixed by `haddock-style: multiline`.

kwxm (commented):
I don't see anything that looks too terrible.  It looks as if it's reformatting comments to be a maximum of 80 characters wide; I could probably cope with 100 or 120, but it's not a big deal.  We've probably argued about this at length before anyway.

PR #7045, opened 2025-04-15
builtin array conformance tests

Unisay:
Same tests as for list conformance but for arrays.

The `agda-conformance` tests for arrays are ignored until `plutus-metatheory` is [updated to include arrays](https://github.com/IntersectMBO/plutus-private/issues/1465).

Closes https://github.com/IntersectMBO/plutus-private/issues/1464


kwxm:
Maybe you're planning to do this in a separate PR, but we need tests for the semantics of each new function as well: see [this directory](https://github.com/IntersectMBO/plutus/tree/master/plutus-conformance/test-cases/uplc/evaluation/builtin/semantics).  Generally we have tests for a few typical succeeding cases and all of the error cases (if there are any: I don't think that `lengthOfArray` can fail, but for `indexArray` you can fail by trying to look up something out of bounds, for example). Usually adding these is straightforward but very tedious,

kwxm:
> we need tests for the semantics of each new function as well
On the other hand the conformance tests include the budgets as well, so you might want to wait until the costing's fixed first, at least for `listToArray`.

Unisay:
> we need tests for the semantics of each new function as well

I've missed those initially, 
fixed now.

kwxm (approved):
This all looks fine, but it might be worth adding a few examples containing big bytestrings just so we have a check that the behaviour and budget don't behave differently depending on the type of the array elements. You'd actually have to look at the results to check that the budgets are the same, but if one budget was to change in future we'd get a failure to alert us.

PR #7046, opened 2025-04-16
[UPLC] [Optimization] Add proper case-of-case

effectfully:
Just to show that proper case-of-case doesn't affect anything in UPLC unless we also have casing on booleans.


effectfully:
So proper case-of-case in UPLC doesn't affect anything.

Makes me wonder why it helps when we have casing on booleans.

@zliu41 I don't really know how PIR case-of-case works and I don't feel like writing tests separately for what we have here, since in the casing on booleans PR we have plenty of instances where the new case-of-case affects golden files (programs, budgets, sizes). So I'm inclined to close this PR and just add the new case-of-case in the casing on booleans one without any extra investigations and testing. This all is a derailment anyway.

So I'm closing this PR, but do feel free to tell me to do things differently.

PR #7049, opened 2025-04-17
Fix memory usage for polymorphic types.

kwxm:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1498

While [costing the array builtins](https://github.com/IntersectMBO/plutus/pull/6950) it appeared that the time taken by `listToArray` differed for different types of list elements.  What was actually happening was that the list was wrapped in a `ListCostedByLength` wrapper but this was ignored when the denotation was executed, so the default size measure was used instead, and this measured the total memory footprint of the list.  So two lists of the same length could have different "sizes" and when working out the dependency of time on size  you would get different slopes for lists with elements of different types.

This fixes that problem (which should only happen when `SomeConstant` is involved, which at present only happens for polymorphic types) by changing the default memory usage for lists and arrays to be the length, and the memory usage for pairs to be equal to 1.  This is arguably the correct thing to do because polymorphic builtins should be parametrically polymorphic, and the time taken by such functions should not depend on the contents of polymorhpic arguments (although if you really want to you can do weird things with the builtin machinery, so it may be possible to produce counterexamples to this) .  The polymorphic builtins which have previously been deployed on the chain all have constant time and memory usage, so this change will not affect any existing code.  The only function which this will affect is `listToArray`, which has not yet been deployed.

This also has the benefit of making a number of Agda conformance tests pass which did not pass previously, because the Agda code has the default cost measures hard-coded and doesn't know about newtype wrappers for alternative size measures.  Since we now only have one size measure for the polymorphic types this problem has now gone away for polymorphic functions, although it still persists for a few non-polymorphic functions.

This may still be a bit fragile: see https://github.com/IntersectMBO/plutus-private/issues/1046 for a possible longer-term solution.


plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -158,8 +156,50 @@ class ExMemoryUsage a where
...
+ actual value of `n`, which will be exponentially greater than the memoory
+ occupied by `n`, so this case the memory usage is not a sensible size measure.
+ In most cases the default `memoryUsage` function returns the actual memory
+ usage, butto deal with cases like `replicateByte` we occasionally use newtype
+ wrappers which define a different size measure (see `IntegerCostedLiterally`
+ below).  Polymorhpic types require some care though: see Note [ExMemoryUsage
+ for polymorphic types].
+-}
+
+{- Note [ExMemoryUsage for polymorphic types].  For polymorphic types such as
```
kwxm, line 180: @effectfully I'm not confident that this comments is completely accurate, so feel free to make corrections.

kwxm: Maybewe should say something about this in `Builtins.hs` as well, although there's already hundreds of liness of comments in there.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -158,8 +156,50 @@ class ExMemoryUsage a where
...
+ will depend on the size of the inputs.  The name `memoryUsage` is perhaps a
+ misnomer: it was originally supposed to measure (in 64-bit words) the heap
+ space required to store an object, but this is not always the correct measure
+ to use.  For example, the time taken by `AddInteger` or `MultiplyInteger` will
+ depend on the logarithms of the inputs (and the logarithm is proportional to
+ the memory occupied by the inputs), and the memory occupied by the result will
+ be some function of the memory occupied by the inputs, so for these functions
+ the actual memory usage is a sensible size measure.  However, calling
+ `replicateByte n b` function allocates a number of bytes which is equal to the
+ actual value of `n`, which will be exponentially greater than the memoory
```
Unisay, line 171:
```suggestion
 actual value of `n`, which will be exponentially greater than the memory
```

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -158,8 +156,50 @@ class ExMemoryUsage a where
...
+ usage, butto deal with cases like `replicateByte` we occasionally use newtype
+ wrappers which define a different size measure (see `IntegerCostedLiterally`
+ below).  Polymorhpic types require some care though: see Note [ExMemoryUsage
+ for polymorphic types].
+-}
+
+{- Note [ExMemoryUsage for polymorphic types].  For polymorphic types such as
+ `pair, `list`, and `array` DO NOT use newtype wrappers to define alternative
+ size measures.  The denotations of functions which take polymorphic arguments
+ use `SomeConstant` and this willl ignore newtype wrappers and will only use the
```
Unisay, line 183:
```suggestion
 use `SomeConstant` and this will ignore newtype wrappers and will only use the
```

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -158,8 +156,50 @@ class ExMemoryUsage a where
...
+ polymorphic arguments: the time taken to process a list, for example, will
+ typically depend only on the length of the list and not the size of the
+ contents.  Currently all such functions are parametrically polymorphic and only
+ manipulate pointers without inspecting the contents of their polymorphic
+ arguments, so it is reasonable to use size measures which depend only on the
+ surface structure of polymorphic objects. -}
+
+instance ExMemoryUsage (a, b) where
+    memoryUsage _ = singletonRose 1
+    {-# INLINE memoryUsage #-}
```
Unisay, line 195: The indentation is inconsistent across these instances: some use 4 spaces, some 3.
I suggest raking the whole file with `fourmolu`

```suggestion
  memoryUsage _ = singletonRose 1
  {-# INLINE memoryUsage #-}
```

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -158,8 +156,50 @@ class ExMemoryUsage a where
...
+
+instance ExMemoryUsage [a] where
+  memoryUsage l = singletonRose . fromIntegral $ length l
+  {-# INLINE memoryUsage #-}
+
+instance ExMemoryUsage (Vector a) where
+    memoryUsage l = singletonRose . fromIntegral $ Vector.length l
```
Unisay, line 202:
```suggestion
    memoryUsage = singletonRose . fromIntegral . Vector.length
```

Unisay (approved):
Looks good! (modulo formatting/typos).

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
kwxm: Hmm. Bouncy keyboard.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
kwxm: >  I suggest raking the whole file with fourmolu
I'll do that before merging.  It changes almost every line though, so I won't do it right now.

Unisay: Thanks!

plutus-metatheory/src/Cost/Size.lagda.md
```diff
@@ -75,12 +75,8 @@ defaultConstantMeasure (tmCon (atomic aBls12-381-g1-element) x) = g1ElementSize
...
 defaultConstantMeasure (tmCon (atomic aBls12-381-mlresult) x) = mlResultElementSize x
 defaultConstantMeasure (tmCon (list t) []) = 1
-defaultConstantMeasure (tmCon (list t) (x ∷ xs)) =
-      3 + defaultConstantMeasure (tmCon t x)
-      + defaultConstantMeasure (tmCon (list t) xs)
-defaultConstantMeasure (tmCon (pair t u) (x , y)) =
-      1 + defaultConstantMeasure (tmCon t x)
-      + defaultConstantMeasure (tmCon u y)
+defaultConstantMeasure (tmCon (list t) l) = Utils.length l
+defaultConstantMeasure (tmCon (pair t u) (x , y)) = 1
```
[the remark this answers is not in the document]
kwxm: > What logic is applied here?

No logic.  Nothing we ever do with pairs should depend on the size (famous last words), so it shouldn't matter what it is.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -158,8 +156,50 @@ class ExMemoryUsage a where
...
+ manipulate pointers without inspecting the contents of their polymorphic
+ arguments, so it is reasonable to use size measures which depend only on the
+ surface structure of polymorphic objects. -}
+
+instance ExMemoryUsage (a, b) where
+    memoryUsage _ = singletonRose 1
+    {-# INLINE memoryUsage #-}
+
+instance ExMemoryUsage [a] where
+  memoryUsage l = singletonRose . fromIntegral $ length l
```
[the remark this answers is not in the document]
kwxm: For some reason I thought that we weren't doing that in this file, but maybe I'm just imagining that.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -304,40 +322,20 @@ addConstantRose (CostRose cost1 forest1) (CostRose cost2 forest2) =
...
-   won't be called.  Thus if we just counted the number of nodes the sizes of
-   'I 2' and 'B <huge bytestring>' would be the same but they'd take different
-   amounts of time to compare.  It's not clear how to trade off the costs of
-   processing a node and processing the contents of nodes: the implementation
-   below compromises by charging four units per node, but we may wish to revise
-   this after experimentation.
+{- A naive traversal for size.  This accounts for the number of nodes in a Data
+   object and also the sizes of the contents of the nodes.  This is not ideal,
+   but it seems to be the best we can do.  At present this only comes into play
+   for 'equalsData', which is implemented using the derived implementation of
```
effectfully, line 328: Not just `equalsData`, but also `serialiseData`?

plutus-metatheory/src/Cost/Size.lagda.md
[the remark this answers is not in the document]
effectfully: It may, if add a builtin serializing a list of pairs or whatever.

But for the time being let's just make it `maxBound` or whatever instead of `1`? So that if our assumption that we'll never need a builtin where that matters gets violated, we'll notice that.

effectfully (approved):
Great stuff.

zliu41:
> What was actually happening was that the list was wrapped in a ListCostedByLength wrapper but this was ignored when the denotation was executed

Why was it ignored- is it a bug and is it fixed in this PR? I'm asking because there are other newtype wrappers like `NumBytesCostedAsNumWords` and we don't want those to be ignored.

kwxm:
> Why was it ignored- is it a bug and is it fixed in this PR? I'm asking because there are other newtype wrappers like `NumBytesCostedAsNumWords` and we don't want those to be ignored.

@effectfully will be able to fill in the details, but my understanding is that this only happens for wrappers around polymorphic types: the code that converts those back into Haskell types is unaware of the wrappers and so it always uses the default size measure.  For monomorphic types like integers, bytestrings, etc, the wrappers are recognised and the correct size measure is used.  This hasn't shown up before because all of the functions that we have that operate on polymorphic types were constant time, unlike `listToArray`.

plutus-metatheory/src/Cost/Size.lagda.md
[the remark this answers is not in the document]
kwxm: > It may, if add a builtin serializing a list of pairs or whatever.

I don't think that we can do that polymorphically unless we assume that everything in the universe of types is serialisable.  We could write a function that would serialise `(integer,bytestring)` pairs or something but that would be monomorphic and I think that in that case it'd be ok to use a wrapper.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -158,8 +156,50 @@ class ExMemoryUsage a where
...
+instance ExMemoryUsage (a, b) where
+    memoryUsage _ = singletonRose 1
+    {-# INLINE memoryUsage #-}
+
+instance ExMemoryUsage [a] where
+  memoryUsage l = singletonRose . fromIntegral $ length l
+  {-# INLINE memoryUsage #-}
+
+instance ExMemoryUsage (Vector a) where
+    memoryUsage l = singletonRose . fromIntegral $ Vector.length l
```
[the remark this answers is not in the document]
kwxm: I've added a comment based on this one.

PR #7054, opened 2025-04-22
Add array built-in functions to the spec

Unisay:


doc/plutus-core-spec/cardano/builtins6.tex
```diff
@@ -63,6 +63,36 @@ \subsubsection{Built-in functions}
...
+        $[x_1,\ldots,x_n] \mapsto n$ 
+      &  
+      & \\
+    \TT{listToArray} 
+      & $[\forall a_\#, \listOf{a_\#}] \to \arrayOf{a_\#}$ 
+      & $[] \mapsto []$
+        \newline 
+        $[x_1] \mapsto [x_1]$
+        \newline 
+        $[x_1,\ldots,x_n] \mapsto [x_1,\ldots,x_n]$
```
Unisay, line 81: I don't know a better denotation. Maybe we should use a different syntax to denotate arrays? @effectfully please suggest, what would you do?

Unisay: This is how the whole table looks when rendered:
image

kwxm: This looks good, although I think you could miss out the two `[x₁] ↦ ...` cases since they're covered by `[x₁, ..., xₙ]` with n=1 (and arguably the `[]` case is covered by n=0, but that might be a bit confusing).

Also,
```
[x₁, ..., xₙ] ↦ [x₁, ..., xₙ]
```
might look a bit weird,but on page 33 we've already got `(x,y) ↦ (x,y)` for the denotation of `mkPairData`, so its maybe not such a big deal.

doc/plutus-core-spec/flat-serialisation.tex
```diff
@@ -528,7 +531,8 @@ \subsection{Constants}
   & \Econstant{\ty{bool}}(s, \texttt{False})       &&= s \cdot \bits{0}\\
   & \Econstant{\ty{bool}}(s, \texttt{True})        &&= s \cdot \bits{1}\\
   & \Econstant{\listOf{\tn}}(s,l)                  &&= \Elist^{\tn}_{\mathsf{constant}}(s, l) \\
-  & \Econstant{\pairOf{\tn_1}{\tn_2}}(s,(c_1,c_2))  &&= \Econstant{\tn_2}(\Econstant{\tn_1}(s, c_1), c_2)\\
+  & \Econstant{\arrayOf{\tn}}(s,l)                 &&= \Earray^{\tn}_{\mathsf{constant}}(s, l) \\
```
kwxm, line 534: I think that `flat` encodes arrays in a different way from lists.  I have a program [here](https://github.com/kwxm/debug-flat) that'll take a `flat` file and decode it, describing what the bits of the encoding mean.  I wrote that to help me understand the flat encoding so that I could write a proper specifcation of the parts that we use.  I don't quite remember why, but I added the array operations a few weeks ago and had to write some [code to decode arrays](https://github.com/kwxm/debug-flat/blob/d0c8fc0469ba04c6a117fed6b7928c8bc2be6c4e/Main.hs#L378).  After a bit of trial and error, it seemed that arrays are encoded by emitting a natural number that says how many elements it has, then emitting the encodings of the individual elements, and then for some reason emitting 8 zero bits afterwards.  I didn't actually check the implementation though, so maybe you should have a look and see if that's  correct.

kwxm (changes requested):
Well done for getting to grips with the TeX!  This looks pretty good, although I think the flat encoding may be wrong.

doc/plutus-core-spec/flat-serialisation.tex
kwxm: I had a look but I couldn''t work out how the flat encoding for arrays is determined.  I checked again and arrays are definitely encoded differently from lists.

doc/plutus-core-spec/cardano/builtins6.tex
Unisay: > I think you could miss out the two [x₁] ↦ ... cases since they're covered by [x₁, ..., xₙ] with n=1 (and arguably the [] case is covered by n=0, but that might be a bit confusing).

Yeah, I wasn't sure if `[x₁, ..., xₙ] ↦ [x₁, ..., xₙ]` subsumes `[x₁] ↦ [x₁]` or even `[] ↦ []` so I went for more explicit notation. Happy to simplify it if you think that there is an ambiguity.

doc/plutus-core-spec/cardano/builtins6.tex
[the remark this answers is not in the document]
Unisay: Added a Batch 6 sub-sub-section for the array type operator and its concrete syntax.

doc/plutus-core-spec/flat-serialisation.tex
kwxm: The haddock for the `Vector` instance says

> Vectors are encoded as arrays.

but that seems to be very misleading.  The Haskell [Array](https://hackage.haskell.org/package/array-0.5.4.0/docs/Data-Array.html#t:Array) type is more like some kind of association list and you can see that in the examples for the `Array` instance: it's encoded as a sequence of (index,element) pairs.  That's definitely not what it's doing with UPLC arrays.  It seems to be doing something much more like what you'd do with standard arrays.  In the source for the `flat` library it has
```
instance Flat a => Flat (V.Vector a) where
  size   = sizeSequence
  encode = encodeSequence
  decode = decodeSequence
```
and in the [the file that defines `encodeSequence](https://github.com/Quid2/flat/blob/3014e2dde892635c4f46037ba929ec514e0f130f/src/Flat/Instances/Mono.hs#L40) it says
> In practice, this means that the sequence is encoded as a sequence of blocks of up to 255 elements, with every block preceded by the count of the elements in the block and a final 0-length block.

This looks much more like what I observed, and my flat debugger fails when I give it an array with more than 255 elements because I assumed that there was only one block.  I think (but haven't checked in detail) that it's encoding an array as a sequence of blocks, each one consisting of a size (eight bits long?) followed by that many elements, and then the 00000000 at the end is an empty block that tells you thhat you've finished.  Presumably that means that you can serialise an array in many different ways depending on how  you break it into blocks.  I guess we''ll have work out the details and put it all into the specification.   I think something similar happens in CBOR and that's described in our specification, so maybe we could steal that, or at least a modified version.

Unisay: You're right, my experiments show that `Strict.Vector`s are encoded as sequences:
```haskell
#!/usr/bin/env runhaskell

import Data.Bits (testBit)
import Data.ByteString (ByteString, unpack)
import Data.Vector.Strict (Vector)
import Data.Vector.Strict qualified as StrictVector
import Data.Word (Word8)
import Flat (DecodeException, Flat (..), flat, unflat)
import Flat.Instances.Test (tstBits)
import Flat.Instances.Vector ()

main :: IO ()
main = do
  let
    words :: Vector Word8
    words = StrictVector.fromList [1, 2, 3, 4, 5]

    bytes :: ByteString
    bytes = flat words

  print $ tstBits words
  -- (True,56,"00000101 00000001 00000010 00000011 00000100 00000101 00000000")

  putStrLn $ renderBits bytes
  -- 00000101 00000001 00000010 00000011 00000100 00000101 00000000 00000001

  let
    words' :: Either DecodeException (Vector Word8)
    words' = unflat bytes

  print words' -- Right [1,2,3,4,5]

renderBits :: ByteString -> String
renderBits bs = unwords [byteToBits w | w <- unpack bs]
 where
  byteToBits :: Word8 -> String
  byteToBits byte = [if testBit byte i then '1' else '0' | i <- [7, 6 .. 0]]

-- from module Data.Vector.Orphans () in the Plutus repo
instance (Flat a) => Flat (Vector a) where
  size = size . StrictVector.toLazy -- Strict to Lazy is O(1)
  encode = encode . StrictVector.toLazy
  decode = StrictVector.fromLazy <$> decode -- Strict from Lazy is O(1)
```

I interpret the encoded bits as:
```
00000101: size = 5 
00000001: 1
00000010: 2
00000011: 3
00000100: 4
00000101: 5
00000000: end of the vector
00000001: padding
```

Unisay: image

ana-pantilie: @Unisay Maybe the binary serialisation/deserialisation section of https://cips.cardano.org/cip/CIP-0138 helps. It describes how the `flat` encoding works.

ana-pantilie: Also, if it turns out that what I wrote in the CIP is wrong, we should amend it.

kwxm: > Also, if it turns out that what I wrote in the CIP is wrong, we should amend it.

It says "The decoding of arrays is, of course, the inverse of the encoding function".  I think that may not be exactly correct.  For encoding it specifies that you get a bunch of initial 255-byte blocks possibly followed by a smaller final one.  I suspect (but haven't checked) that the decoder will accept a sequence of blocks of any nonzero length.  There's a similar thing for bytestrings in flat (see p62 of the specification) and certain items in `data`: see the first bullet point on p52 and also Section B.5 on p55. See also https://github.com/IntersectMBO/plutus/issues/5852.

ana-pantilie: @Unisay looks like @kwxm is who uncovered the possible bug. It might make sense to do what we discussed in the technical discussion: modify https://cips.cardano.org/cip/CIP-0035 to not require Plutus CIP authors to provide binary encodings themselves, and then modify the Array CIP saying that the encoding is based on the flat encoding for arrays and provide no other details. What do you think?

kwxm: > It might make sense to do what we discussed in the technical discussion: modify https://cips.cardano.org/cip/CIP-0035 to not require Plutus CIP authors to provide binary encodings themselves.

@ana I agree.  In many cases `flat` will already provide a serialisation instance and it would be natural to use that, so we should only require CIP authors to specify the flat encoding if it's something that flat doesn't already support.  Also, we don't even support flat encodings for the BLS12-381 types because the decoder could take up a lot of time during deserialisation, and that isn't covered by fees. So only some types are flat-encodable, and most of the time there'll be a default that handles the encoding already.

kwxm (changes requested):
This looks good modulo a couple of small remarks (getting rid of the separate cases for one-element arrays).  Could you update [the date on the front page](https://github.com/IntersectMBO/plutus/blob/07963445c71e1cc27e9f6412c389f53376651083/doc/plutus-core-spec/plutus-core-specification.tex#L8) as well please? That makes it easier to check that someone's using the correct version if thet have complaints or questions.

doc/plutus-core-spec/cardano/builtins6.tex
Unisay: image

doc/plutus-core-spec/cardano/builtins6.tex
```diff
@@ -11,9 +11,39 @@
 \subsection{Batch 6}
 \label{sec:default-builtins-6}
 
+\subsubsection{Built-in type operators}
+\label{sec:built-in-type-operatos-6}
```
kwxm, line 15:
```suggestion
\label{sec:built-in-type-operators-6}
```

doc/plutus-core-spec/cardano/builtins6.tex
```diff
@@ -11,9 +11,39 @@
...
+Some valid constant expressions are thus:
+\begin{verbatim}
+   (con (array bool) [True, False, True])
+   (con (array integer) [])
+\end{verbatim}
+
 \subsubsection{Built-in functions}
 \label{sec:built-in-functions-6}
-The sixth batch of builtin operations is defined in Table~\ref{table:built-in-functions-6}.
+The sixth batch of built-in types is defined in Table~\ref{table:built-in-types-6}.
```
kwxm, line 45:
```suggestion
The sixth batch of built-in types is defined in Table~\ref{table:built-in-type-operators-6}.
```

kwxm (changes requested):
I found a couple of typos in cross-references and added suggestions to fix them.  If it'll let me I'll commit those myself.  Apart from that I think it can be merged.  I might come back later and try to refactor the document a bit because the flat encodings for bytestring and arrays are very similar:  see https://github.com/IntersectMBO/plutus-private/issues/1581.

kwxm (approved):
OK, so it let me commit my changes to fix the typos.  I'll set it to auto-merge.

doc/plutus-core-spec/cardano/builtins6.tex
```diff
@@ -63,6 +93,31 @@ \subsubsection{Built-in functions}
...
+        \newline 
+        $[x_1,\ldots,x_n] \mapsto n\ (n \geq 1)$ 
+      &  
+      & \\
+    \TT{listToArray} 
+      & $[\forall a_\#, \listOf{a_\#}] \to \arrayOf{a_\#}$ 
+      & $[] \mapsto []$
+        \newline 
+        $[x_1,\ldots,x_n] \mapsto [x_1,\ldots,x_n]\ (n \geq 1)$
+      & \\
```
kwxm, line 108:
```suggestion
      & & \\
```

kwxm (approved):
There was an `&` missing in the table too.

PR #7059, opened 2025-04-25
Fix Haskell version of certifier builtinEq

ana-pantilie:
~Needs https://github.com/IntersectMBO/plutus/pull/7051 before it can be reviewed properly~ Done.

Fixes https://github.com/IntersectMBO/plutus-private/issues/1534
Fixes https://github.com/IntersectMBO/plutus-private/issues/1528


ramsay-t (approved):
This looks great (I mean, it looks like an awful kludge but that's Agda's fault, not yours :) ). Getting it to handle two different versions in two different settings is really useful.

PR #7060, opened 2025-04-27
doc: fix missing await on wallet.signTx

dev6699:
Fix missing await in <b>Minting the Token to be Auctioned</b> example doc.
`wallet.signTx` return a Promise, getting `{"code":"ERR_INVALID_ARG_TYPE"}` when calling `wallet.submitTx`


zliu41 (approved):
Thanks

PR #7037, opened 2025-04-14
Drop Support for GHC 8.10

zeme-wana:
This PR removes support for GHC 8.10.

PR #7063, opened 2025-04-29
Forgot changelog entries

kwxm:
We forgot to add changelog entries in #7037.  This adds them


kwxm:
I've merged it without review since it doesn't affect any code.

PR #7064, opened 2025-04-29
Fix `expModInteger` bug

### LINKED ISSUE #7061: Fix `expModInteger` bug
The builtin call 
``` 
   expModInteger 0 e m
```
returns 0 for e<0.  This is wrong for all m except m=1: for m>1 it should actually fail because 0 is not invertible (and `expModInteger b e m` does fail when b is a nonzero muiltiple of m).  This is because of a bug in [integerPowMod#](https://hackage.haskell.org/package/ghc-bignum-1.3/docs/src/GHC.Num.Integer.html#integerPowMod%23) in `GHC.Num.Integer`, which begins

```
integerPowMod# !b !e !m
   | naturalIsZero m   = (# | () #)
   | naturalIsOne  m   = (# naturalZero | #)
   | integerIsZero e     = (# naturalOne  | #)
   | integerIsZero b     = (# naturalZero | #)  <--------
   | integerIsOne  b     = (# naturalOne  | #)
     -- when the exponent is negative, try to find the modular multiplicative
     -- inverse and use it instead
   | integerIsNegative e = case integerRecipMod# b m of ...
```

The case for `integerIsZero b` should only return if `e>=0`; if `e<0` then it should carry on to the `integerIsNegative` case.   We should add a fix for this in our [expMod](https://github.com/IntersectMBO/plutus/blob/840af899086d1870de61a820795ae0ceb8374b85/plutus-core/plutus-core/src/PlutusCore/Crypto/ExpMod.hs#L19) function and report the bug to the GHC maintainers. 

For m=1 the behaviour is correct: in that case we're effectively working in the zero ring, where 0 is invertible with inverse 0 since 0 is also the identity element.  The `naturalIsOne` case takes care of this.  However, `GHC.Num.Integer`'s [integerRecipMod#](https://hackage.haskell.org/package/ghc-bignum-1.3/docs/src/GHC.Num.Integer.html#integerRecipMod%23) function doesn't handle the m=1 case correctly: it's implemented as
```
integerRecipMod# x m
   | integerIsZero x = (# | () #)
   | naturalIsZero m = (# | () #)
   | naturalIsOne  m = (# | () #)
   | True            = Backend.integer_recip_mod x m
```
and so it fails when m=1 and n=0.  It should return 0 in this case.  This doesn't affect us directly, but we should report it anyway.

kwxm:
This fixes #7061 (see that for details).  I'll add the property tests that discovered the problem in a later PR.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Crypto/ExpMod.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Crypto/ExpMod.hs
@@ -21,7 +21,7 @@ expMod b e m
   -- ^ Just in case: GHC.Num.Integer.integerRecip# gets this wrong.  Note that 0
   -- is invertible modulo 1, with inverse 0.
   | b == 0 && e < 0 = failNonInvertible 0 m
-  -- ^ GHC.Num.Integer.integerPowMod# incorrectly returns 0 in this case.
+  -- ^ integerPowMod# incorrectly returns 0 in this case.
   | otherwise =
       case integerPowMod# b e m of
         (# n | #)  -> pure n
--- a/plutus-core/plutus-core/src/PlutusCore/Crypto/ExpMod.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Crypto/ExpMod.hs
@@ -10,16 +10,23 @@ import PlutusCore.Builtin
 import GHC.Natural
 import GHC.Num.Integer
 
--- similar to `Default.Builtins.nonZeroSecondArg`
--- We don't really need it because integerPowMod# returns `()` on zero mod, but we put
--- in case of future implementation changes.
--- TODO https://github.com/IntersectMBO/plutus-private/issues/1474
-expMod _ _ 0 = fail "Cannot divide by zero"
-expMod b e m =
-    case integerPowMod# b e m of
-        (# n | #)  -> pure n
-        (# | () #) -> fail "expMod: failure"
-
-
+-- | Modular exponentiation.  This uses GHC.Num.integerPowMod#, which gives the
+-- wrong answer in some cases.  TODO: we'll be able to remove some of the guards
+-- when/if integerPowMod# gets fixed.
 expMod :: Integer -> Integer -> Natural -> BuiltinResult Natural
+expMod b e m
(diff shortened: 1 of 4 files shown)


plutus-core/plutus-core/src/PlutusCore/Crypto/ExpMod.hs
```diff
@@ -10,16 +10,23 @@ import PlutusCore.Builtin
...
-        (# n | #)  -> pure n
-        (# | () #) -> fail "expMod: failure"
-
-
+-- | Modular exponentiation.  This uses GHC.Num.integerPowMod#, which gives the
+-- wrong answer in some cases.  TODO: we'll be able to remove some of the guards
+-- when/if integerPowMod# gets fixed.
 expMod :: Integer -> Integer -> Natural -> BuiltinResult Natural
+expMod b e m
+  | m <= 0 = fail "expMod: invalid modulus"
```
zliu41, line 18: Is this the right thing to do - not allowing negative modulus? This would be inconsistent with `modInteger` which does allow negative modulus.

kwxm: The modulus in `integerPowMod#` is a Natural, so if we wanted to allow negative moduli we'd have make a decision about how to deal with that and add some code to do so.  Mathematically the integers modulo -m are the same as the integers modulo m, so the easiest thing to do would be to replace m by its absolute value.  However, we have eg `mod 10 (-7) = -4` so if we wanted to the case e=1 to reproduce the behaviour of `mod` we couldn't do that and we'd also have to abandon desirable properties of `expMood` like the result always being positive and being periodic modulo m.  Note that `rem` also works with negative moduli but it does something different: `rem 10 (-7) = 3`, but `rem (-10) (-7) = -3`.    

I've never seen anyone using a negative modulus for any purpose and trying to support it would probably just be confusing (and it's already confusing enough with div/mod and quot/rem).

zliu41: Ok, the CIP says the modulus is positive, so we can tentatively make it positive only. If one needs a negative modulus they can still get the answer easily with an extra step.

zliu41 (approved):

Issue #7015, opened 2025-04-04
Extra conformance tests for `expModInteger`


kwxm:
We could do with some more conformance tests, in particular ones with large inputs, eg calculating a^b mod m with |a| and |b| >> m.  Also there's an edge case for m=1 (a^b is zero for all a and b) that could be checked  more thoroughly.

kwxm:
We really need accurate costing to do this properly, although the tests could be added with preliminary costing and the results could be updated later.

PR #7066, opened 2025-05-01
Property tests for `expModInteger`

kwxm:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1474.

Testing for `expModInteger` was limited because we were still supporting GHC 8.10 which didn't support `expModIntegers`.  GHC 8.10 is now gone, so this adds some property tests.  These uncovered a bug: see issue #7061.

I intend to extend the conformance tests in a later PR once costing is complete: see #7015.

I've re-used some of the testing machinery from the BLS12-381 property tests, which led to some refactoring to put the common functionality in one place.  Similar code probably occurs in multiple other places in the codebase: I think we're doomed to re-invent this stuff every time we update the tests.  Possibly we could try to collect all testing functionality in one place, but that would be quite tricky and probably still wouldn't satisfy everyone.


plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Definition.hs
```diff
@@ -652,6 +654,8 @@ test_ExpModInteger = testNestedM "ExpMod" $ do
     evals @Integer 0 b [] [int 500, int (-1777), one] -- base:*, exp: neg, mod:1
     fails "mod-zero" b [] [one, one, zero] -- base:*, exp:*, mod: 0
     fails "mod-neg" b [] [one, one, int (-3)] -- base:*, exp:*, mod: neg
+    -- base is zero, negative exponent
+    fails "exp-neg-non-inverse0" b [] [int 0, int (-1), int 7]
```
kwxm, line 658: This is the case where the bug gave the worng result: none of the previous tests detected that.

zliu41 (approved):
Looks great

PR #7069, opened 2025-05-02
Add the certifier to plutus-tx-plugin

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1348


ramsay-t (approved):
Looks fine - does this add any extra constraints to what we can change in the certifier? 

Also: is there a nice way to build some tests of this interface? Can we add some to the general plugin tests?

ana-pantilie:
> Looks fine - does this add any extra constraints to what we can change in the certifier?

Not sure what you mean.

> Also: is there a nice way to build some tests of this interface? Can we add some to the general plugin tests?

I have a separate issue for adding tests.

PR #7071, opened 2025-05-06
Costing for `expModInteger`

kwxm:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1448
Fixes https://github.com/IntersectMBO/plutus-private/issues/1554

This adds costing for `expModInteger`.  Here are the costs of the builtin with inputs of various sizes, as a percentage of the maximum execution budget.  For cryptographic purposes you might go up to about 512 bits, so it doesn't look as if it's prohibitively expensive.

| Size in bytes | Size in bits | Percentage of maximum budget |
| :----------- | :-------- | :----------- |
| 16 | 128 | 0.5% |
| 32  | 256 | 0.11% |
| 64 | 512  | 0.49% |
| 128 | 1024 | 3.07% |
| 256 | 2048 | 21% |


plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -169,7 +169,7 @@ builtinMemoryModels = BuiltinCostModelBase
   , paramCountSetBits                    = Id $ ModelOneArgumentConstantCost 1
   , paramFindFirstSetBit                 = Id $ ModelOneArgumentConstantCost 1
   , paramRipemd_160                      = Id $ hashMemModel Hash.ripemd_160
-  , paramExpModInteger                   = Id $ ModelThreeArgumentsConstantCost 100000000000 -- FIXME: stub
+  , paramExpModInteger                   = Id $ ModelThreeArgumentsLinearInZ identityFunction
```
kwxm, line 172: Oh wait!  This is the size in bytes, not words, so it'll be 8 times bigger than it should be.  The size of the result will be the same as the size of the modulus, but we're using a non-standard measure here.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -383,22 +392,53 @@ evaluateTwoVariableQuadraticFunction
...
--- ModelTwoArgumentsSubtractedSizes instead, but that would change the order of
--- the cost model parameters since the minimum value would come first instead of
--- last.
--- Tracked by https://github.com/IntersectMBO/plutus-private/issues/1554.
+-- | c00 + c01x*y + c12x*y^2
+-- This is used only for `expModInteger`, whose costing is quite complex.
+data ExpModCostingFunction = ExpModCostingFunction
+  { coefficient00 :: Coefficient00
+  , coefficient11 :: Coefficient11
+  , coefficient12 :: Coefficient12
```
kwxm, line 400: This seemed so specific that it'd be pointless to give it a more general name.  We could have made it a general cubic function in two variables, but that would have ten coefficients and we'd only be using three.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -383,22 +392,53 @@ evaluateTwoVariableQuadraticFunction
...
+   (ExpModCostingFunction
+    (Coefficient00 c00) (Coefficient11 c11) (Coefficient12 c12))
+  aa ee mm = if aa <= 5*mm
+          then cost0
+          else cost0 + (cost0 `dividedBy` 2)
+  where cost0 = c00 + c11*ee*mm + c12*ee*mm*mm
+{-# INLINE evaluateExpModCostingFunction #-}
+
 -- | s * (x - y) + I
+{- In principle we could use ModelConstantOrOneArgument here, but that would
```
kwxm, line 426: Fixing this because of https://github.com/IntersectMBO/plutus-private/issues/1554

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -226,6 +227,53 @@ instance ExMemoryUsage () where
...
+    memoryUsage n = memoryUsage $ toInteger n
+    {-# INLINE memoryUsage #-}
+
+instance ExMemoryUsage Word8 where
+    memoryUsage _ = singletonRose 1
+    {-# INLINE memoryUsage #-}
+
+{- | Calculate a 'CostingInteger' for the size of the given 'Integer', measured in
+   8-bit bytes. -}
+memoryUsageBytes :: Integer -> CostingInteger
```
kwxm, line 259: I added this because using the standard measure would give very coarse-grained results and this is much more accurate.  Now I'm worried that it's jsut adding more complexity.  Maybe we should go for inaccurate but simple.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -383,22 +392,53 @@ evaluateTwoVariableQuadraticFunction
...
+  -> CostingInteger
+  -> CostingInteger
+  -> CostingInteger
+  -> CostingInteger
+evaluateExpModCostingFunction
+   (ExpModCostingFunction
+    (Coefficient00 c00) (Coefficient11 c11) (Coefficient12 c12))
+  aa ee mm = if aa <= 5*mm
+          then cost0
+          else cost0 + (cost0 `dividedBy` 2)
```
kwxm: ie, 1.5 times the cost.  It seems a bit risky to multiply saturated integers by fractions though.  

I'm a bit worried about this being hard-coded since it'll have to be reporduced in eg the metatheory.  Maybe this should go in the cost model parameters along with the cofficients.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -383,22 +392,53 @@ evaluateTwoVariableQuadraticFunction
...
+evaluateExpModCostingFunction
+  :: ExpModCostingFunction
+  -> CostingInteger
+  -> CostingInteger
+  -> CostingInteger
+  -> CostingInteger
+evaluateExpModCostingFunction
+   (ExpModCostingFunction
+    (Coefficient00 c00) (Coefficient11 c11) (Coefficient12 c12))
+  aa ee mm = if aa <= 5*mm
```
zliu41, line 419: I thought it was `aa <= mm` - how did you arrive at `5*mm`?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -2020,8 +2022,18 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
...
-        let expModIntegerDenotation :: Integer -> Integer -> Natural -> BuiltinResult Natural
-            expModIntegerDenotation = ExpMod.expMod
+        let expModIntegerDenotation
+              :: IntegerCostedByNumBytes
+              -> IntegerCostedByNumBytes
+              -> IntegerCostedByNumBytes
+              -> BuiltinResult Natural
+            expModIntegerDenotation
+              (IntegerCostedByNumBytes a)
+              (IntegerCostedByNumBytes b)
```
zliu41, line 2032: `b` -> `e` for consistency?

zliu41 (approved):
Looks sensible!

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
kwxm: > I thought it was `aa <= mm` - how did you arrive at `5*mm`?

I thought I'd be generous and allow callers to go a little bit over the limit: if you're calculating `(ab)^e mod m` for example, `ab` might be close to twice the size of `m`.  I'm thinking that I should probably just make it `aa < mm` though.

kwxm:
I'm closing this in favour of #7080.

PR #7072, opened 2025-05-06
Evaluation of CompiledCode for testing.

Unisay:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-tx/testlib/PlutusTx/Test/Golden.hs
```diff
@@ -0,0 +1,135 @@
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE KindSignatures        #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE TypeApplications      #-}
+{-# LANGUAGE TypeOperators         #-}
+
+module PlutusTx.Test.Golden (
```
Unisay, line 8: The code has been moved here from `PlutusTx.Test` without changes.

plutus-tx/testlib/PlutusTx/Test/Orphans.hs
```diff
@@ -0,0 +1,51 @@
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE LambdaCase            #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE UndecidableInstances  #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module PlutusTx.Test.Orphans where
```
Unisay, line 9: Code migrated without changes.

zliu41 (commented):
I'll take a closer look soon but I don't like putting this in `plutus-tx-testlib`. First, it's a library with utilities for our internal testing, not for the users to test their own code. There may be some overlap but they are not quite the same. Second, it may not be used exclusively for testing - users may depend on it in their own libraries or executables that serve other purposes. I'd just put it in the `plutus-tx` package, and call it "PlutusTx.Run" or something like that.

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
```diff
@@ -0,0 +1,201 @@
...
+```haskell
+import PlutusTx.Test (
+  CodeResult,
+  applyLifted,
+  evaluateCompiledCode,
+  prettyCodeResult
+ )
+```
+
+Our example code represents a function that expects an `Integer` argument
```
zliu41, line 41: Perhaps show the actual example code?

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
```diff
@@ -0,0 +1,201 @@
...
+and returns the `Integer` value. 
+
+It is possible to evaluate this compiled code without applying it to any arguments, 
+and evaluation will succeed, returning the value of type `Integer -> Integer`:
+
+```haskell
+result :: CodeResult
+result = evaluateCompiledCode compiledCode
```
zliu41, line 49: The cost model and parameters depend on `MajorProtocolVersion` and ledger language version (V1, V2, V3), so I'd expect `evaluateCompiledCode`  to take them as parameters. It could also take arguments to `compiledCode` as a parameter.

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
```diff
@@ -0,0 +1,201 @@
+---
+sidebar_position: 12
+---
+
+# Evaluating CompiledCode for testing
```
zliu41, line 5: I'd just call it `Evaluating CompiledCode` or `Running CompiledCode`

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+Our example code represents a function that expects an `Integer` argument
+and returns the `Integer` value. 
+
+It is possible to evaluate this compiled code without applying it to any arguments, 
+and evaluation will succeed, returning the value of type `Integer -> Integer`:
+
+```haskell
+result :: CodeResult
```
zliu41, line 48: `CodeResult` sounds weird - perhaps `EvalResult`?

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
```diff
@@ -0,0 +1,201 @@
...
+result = evaluateCompiledCode compiledCode
+```
+
+The `CodeResult` type is declared as follows:
+
+```haskell
+data CodeResult = CodeResult
```
zliu41, line 55: Can you `LiteralInclude` this?

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
```diff
@@ -0,0 +1,201 @@
...
+No traces were emitted
+```
+
+This output reveals that the evaluation was successful, and the resultng UPLC 
+value is an (un-applied) lambda abstraction. 
+
+The [Lifting Values into CompiledCode](./lifting.md) section contains and example 
+of applying the `CompiledCode` to an argument.
+
+Alternatively, we can use the imported `applyLifted` function like this:
```
zliu41, line 146: This part makes it sound like arguments must be a lifted value, but that's not true. You can compile the function using `$$(compile..)`, and compile the argument also using `$$(compile..)`, and then apply the function to the argument using `unsafeApplyCode`. Lifting isn't necessary.

plutus-tx-plugin/test/size/Main.hs
```diff
@@ -4,167 +4,245 @@
...
-              ]
-          , testNested "Other"
-              [ goldenSize "numerator" ratNumerator
-              , goldenSize "denominator" ratDenominator
-              , goldenSize "round" ratRound
-              , goldenSize "truncate" ratTruncate
-              , goldenSize "properFraction" ratProperFraction
-              , goldenSize "recip" ratRecip
-              , goldenSize "abs-specialized" ratAbs
+main =
```
zliu41, line 22: I went through this whole function trying to figure out what changed, only to realize it is just formatting. Please avoid mixing formatting with other changes.

zliu41 (commented):
This PR does multiple things - not just "Evaluation of CompiledCode for testing" but also lots of code reorganization. I find it difficult to review because I wanted to focus on the former, but the latter makes it much harder to do so - for example, "Updated existing tests to evaluate CompiledCode directly" - where is that, exactly?

In the future I'd highly recommend not doing both together.

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
Unisay: I agree that suffix "for testing" makes title less concise, but I felt like it serves an important purpose:
under the hood evaluator uses the `defaultCekParametersForTesting` and I wanted to highlight the fact that compiled code evaluation "for real" is likely going to produce different `ExBudget` values than "for testing".

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
Unisay: Done.

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
Unisay: It would have to reference the file `plutus-tx/testlib/PlutusTx/Test/Run/Code.hs` outside of the `doc` folder; This would make the `doc/docusaurus` not self-contained. 

(I gave it a try anyway, and it didn't work: `Error loading code block`)

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
Unisay: You're right that its not always necessary: this is an optional convenience function if one wants to apply and lift in one hop. 
I don't have strong feelings about it, just wanted to come up with the most concise way to apply lifted value: we have 
```haskell
`unsafeApplyCode` liftCodeDef ...
```
in many places...

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
Unisay: I was referring to the example above
grafik

Perhaps I should place this sentence right after the code block.

Unisay: I have reworked this part to be clearer.

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
Unisay: I see a trade-off here: on one hand I want to expose the most simple API to the users, and the less decisions they have to make - the better it is. On the other - we want it to be flexible: e.g. users can decide which CEK Parameters to use, but it means that CEK Parameters parameter needs to be exposed. 
I decided to hardcode the `defaultCekParametersForTesting` thus limiting the function applicability to testing. 
From your comment I read that you'd go different route: ask users to chose the CEK parameters, directly or indirectly. I am thinking maybe we can have two functions for both sides of the trade-off...

doc/docusaurus/docs/using-plinth/evaluating-plinth.md
zliu41: See above comment

zliu41 (commented):
> What is an alternative approach to making a series of dependent changes that you recommend?

This is discussed in the contributing guide - you can either open the second PR after the first is merged, or open the second PR against the first PR.

This is much better than trying to carefully organize the commits in a single PR, and asking reviewers to review commit-by-commit. The latter has too many problems:
- Most people don't do this. Often times they prefer to commit and push whenever convenient - e.g., before the lunch break, when they finish their day, etc.
- Going back and altering and earlier commit is non-trivial - it's much easier to add new commits.
- When addressing comments, you could make a new commit that significantly alter the changes introduced in a previous commit. Then if a new reviewer comes in and starts reviewing it commit-by-commit, they'll have to first review the earlier commit, only to realize that the later commit rendered it largely obsolete.
- When merging into master the commits are squashed so you'd get a big master commit.

Unisay:
@zliu41 I've moved functions to evaluate code out of the testlib. 
Added a way to provide a custom cost model and convenient way to get one from a `(PlutusLanguageVersion, MajorProtocolVersion)` value combo.
The user guide is updated accordingly.

Please tell me if there remain any other concerns you want to get addressed before merging this PR.

plutus-ledger-api/src/PlutusLedgerApi/MachineParameters.hs
```diff
@@ -0,0 +1,24 @@
+module PlutusLedgerApi.MachineParameters where
+
+import PlutusCore.Default (BuiltinSemanticsVariant (..))
+import PlutusCore.Evaluation.Machine.ExBudgetingDefaults (cekCostModelForVariant)
+import PlutusCore.Evaluation.Machine.MachineParameters (mkMachineParameters)
+import PlutusCore.Evaluation.Machine.MachineParameters.Default (DefaultMachineParameters)
+import PlutusLedgerApi.Common
+
+machineParametersFor
```
zliu41, line 9: Reuse this function in `PlutusLedgerApi.V1/V2/V3.mkEvaluationContext`?

Unisay:
> > I've moved functions to evaluate code out of the testlib.
> 
> It seems they are still in testlib - `PlutusTx.Test.Run.Uplc`. Am I missing something?

No, I totally missed this one myself :) Fixed.

zliu41 (commented):
Looks good. I think the doc needs to be updated - it still says `plutus-tx-testlib`.

zliu41:
I'm still seeing `plutus-tx-testlib` in the doc - looks like it hasn't been updated?

Unisay:
> I'm still seeing `plutus-tx-testlib` in the doc - looks like it hasn't been updated?

Good catch! I have:

1. Updated the package deps in the doc.
2. Fixed grammar, spelling, style of the doc.
3. Made a reference to the full source code clickable.
4. Rebased PR on master.

zliu41 (approved):
LGTM

PR #7078, opened 2025-05-07
Enable & Publish Project Coverage Reports For plutus-core

zeme-wana:
This PR enables the build of the project coverage report.
A report is built on each PR, but only published on push to master.
The report will be available [here](https://plutus.cardano.intersectmbo.org/dev/coverage) 
The report currently is only available for the `plutus-core` package, this is because enabling HPC (haskell program coverage) on other packages breaks several tests in their test-suites, see [this PR](https://github.com/IntersectMBO/plutus/pull/6979) for an effort to address this.
There is one test that also fails in plutus-core, when HPC is enabled:
```
      TrackCosts: retaining:                                                             FAIL (0.43s)
        untyped-plutus-core/testlib/Evaluation/Builtins/Definition.hs:526:
        Too many elements picked up by GC
        Expected at most: 5
        But got: 7180
        The result was: [106,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0>
        Use -p '/TrackCosts: retaining/' to rerun this test only.
```
For this reason we use a new custom cabal flag named `__HPC_ENABLED__` and a new function `Test.Tasty.Extras.ignoreTestIfHpcEnabled`. This could be avoided if we somehow got the `TrackCosts: retaining:` not to fail when HPC is enabled

### DIFF
--- a/.github/workflows/code-coverage-report.yml
+++ b/.github/workflows/code-coverage-report.yml
@@ -10,7 +10,7 @@ on:
   workflow_dispatch:
   pull_request:
     branches: 
-      - test-coverage # TODO remove me before merging
+      - coverage-plutus-core # TODO remove me before merging
 
 jobs:
   deploy:
--- a/.github/workflows/code-coverage-report.yml
+++ b/.github/workflows/code-coverage-report.yml
@@ -8,9 +8,6 @@ on:
     branches: 
       - master 
   workflow_dispatch:
-  pull_request:
-    branches: 
-      - coverage-plutus-core # TODO remove me before merging
 
 jobs:
   deploy:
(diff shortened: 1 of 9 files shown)


ramsay-t (approved):
This is really cool! I'm not at all surprised that we do a terrible job of testing error branches (which would actually give us much better regression tests when we change things...) but it is nice to see the full HPC reports.

zliu41 (approved):
:+1:

PR #7080, opened 2025-05-08
Costing for `expModIntger`, take two.

kwxm:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1448
Fixes https://github.com/IntersectMBO/plutus-private/issues/1554

The earlier PR (#7071) for this measured sizes of inputs in bytes (not words) in order to get a high-resolution view of the data.  I redid it with the standard size measure (8-byte words) and the results were just as accurate, so I decided to use that instead and open a new PR since it  simplifies things a lot.

Some costs calculated according to the new costing function:

| Size in words | Size in bits | Percentage of maximum budget |
| :----------- | :-------- | :----------- |
| 2 | 128 | 0.02% |
| 4  | 256 | 0.07% |
| 8 | 512  | 0.42% |
| 16 | 1024 | 2.77% |
| 32 | 2048 | 19.8% |
| 56 | 3584 | 100.6% |

This seems reasonable for realistic input sizes, and it's actaully a bit cheaper than the previous version.


plutus-core/plutus-core/test/TypeSynthesis/Golden/Signatures/DefaultFun/ExpModInteger.sig.golden
```diff
@@ -1 +1 @@
-Integer -> Integer -> Natural -> BuiltinResult Natural
\ No newline at end of file
+IntegerCostedByNumBytes -> IntegerCostedByNumBytes -> IntegerCostedByNumBytes -> BuiltinResult Natural
```
zliu41, line 1: How are these golden files generated? I don't see a type called `IntegerCostedByNumBytes` in the codebase?

zliu41 (approved):

plutus-core/plutus-core/test/TypeSynthesis/Golden/Signatures/DefaultFun/ExpModInteger.sig.golden
kwxm: Oh, I removed that.  I updated most of the golden files but must have forgotten this one.  I'll fix that.

PR #7082, opened 2025-05-09
[Plinth] [Optimization] Tweak eta and strictness codegen

effectfully:
I wanted to make `Data.List.drop` based on the respectieve, but saw some opportunities and decided to lump them in here as well.


plutus-core/plutus-ir/src/PlutusIR/Purity.hs
```diff
@@ -140,17 +141,20 @@ termEvaluationOrder binfo vinfo = goTerm
...
-        -- then the body
-        <> goTerm b
-        -- then the whole term, which will lead to applications (so work)
-        <> evalThis (EvalTerm Pure MaybeWork t)
-    Let _ Rec _ _ ->
-      -- Hard to know what gets evaluated first in a recursive let-binding,
-      -- just give up
-      evalThis Unknown
+    t@(Let _ rec bs b) ->
+      if rec == NonRec || length (filter isTermBind $ NE.toList bs) <= 1
```
Unisay, line 145:
```suggestion
      if rec == NonRec || length (NE.filter isTermBind bs) <= 1
```

effectfully:
Not gonna pursue it.

PR #7084, opened 2025-05-09
Manually Generate and Version Control plutus-metatheory/src/MAlonzo Modules

zeme-wana:
plutus-metatheory has a custom Setup.hs that invokes agda to generate haskell modules.
This means that we can't build the package on CHaP, for example, as it requires agda in PATH.
This means that externals users can't cabal install plutus-metatheory unless they have a proper agda setup.
We want to lift this restriction.
This PR removes the Setup.hs and introduces instructions on how to generate the MAlonzo.Code modules manually.
Maintainers will have to remember to run this step manually during development.
The generated MAlonzo.Code file will now be in version control.
A new CI check is put in place to detect when devs forget to do this.


ana-pantilie (approved):
Thank you ❤️

PR #7085, opened 2025-05-12
[Docs] Add 'BuiltinResult' to 'BuiltinsOverview.md'

effectfully:
I was looking into this file and noticed that some of it bitrotted, so here are fixes.


Unisay (approved):
👍🏼

PR #7089, opened 2025-05-12
Add missing example to docusaurus cabal file

ana-pantilie:
AuctionMintingPolicy was not included in the cabal file therefore wasn't being built when running `cabal build docusaurus-examples`


Unisay (approved):
Well spotted!

PR #7090, opened 2025-05-13
Add extra conformance tests for `expModInteger`

kwxm:
Fixes https://github.com/IntersectMBO/plutus/issues/7015.

There are a lot of edge cases in `expModInteger` (the existing conformance tests didn't find [this bug](https://gitlab.haskell.org/ghc/ghc/-/issues/26017)) and the costing is quite complex, especially for large inputs.  This PR adds a large number of new tests that should exercise the implementation a bit more comprehensively.  Property tests (like [these](https://github.com/IntersectMBO/plutus/pull/7066)) would be better, but the conformance testing machinery doesn't support property testing yet.  I've kept the original tests because they're not doing any harm.


zliu41 (approved):
:+1:

PR #7091, opened 2025-05-13
Fix force delay test failures

ramsay-t:
The force delay relation had some subtle invariants that aren't actually needed. This PR converts it to the Zipper form and makes it much clearer. 

Fixes https://github.com/IntersectMBO/plutus-private/issues/1541
Fixes https://github.com/IntersectMBO/plutus-private/issues/1555


ana-pantilie (changes requested):
As the issue mentions, there are two force-delay tests which are removed here: https://github.com/IntersectMBO/plutus/blob/1eb5c18f06d597391d78e6d4b0513f5710006b15/plutus-metatheory/test/certifier/Test/Certifier/Optimizer.hs#L59

Do these changes make them pass? If so, please add them back to the test suite.

Also, such a large change requires a lot more testing.

ramsay-t:
> As the issue mentions, there are two force-delay tests which are removed here:
> 
> https://github.com/IntersectMBO/plutus/blob/1eb5c18f06d597391d78e6d4b0513f5710006b15/plutus-metatheory/test/certifier/Test/Certifier/Optimizer.hs#L59
> 
> Do these changes make them pass? If so, please add them back to the test suite.
> 

I told you I wasn't concentrating :) Now with that spec file committed!

> Also, such a large change requires a lot more testing.

There are lots of tests in the Agda file that should move to the actual test spec.

ana-pantilie:
Also don't forget to `generate-malonzo-code`!

ramsay-t:
> This fix doesn't work for one example I'm working with, here's the certificate which says no: [FWSTCert-559113626.zip](https://github.com/user-attachments/files/20188059/FWSTCert-559113626.zip)

Could we make this an actual test? There was a folder full of UPLC files that "just ran".

ramsay-t:
> Also don't forget to `generate-malonzo-code`!

Do we have a test that tests this? (And generates an actual response that says "you didn't do `generate-malonzo-code`"? :) )

ramsay-t:
> This fix doesn't work for one example I'm working with, here's the certificate which says no: [FWSTCert-559113626.zip](https://github.com/user-attachments/files/20188059/FWSTCert-559113626.zip)

This seems to have IfThenElse stuff in it - I haven't supported that change yet; I was going to do it as a different PR but I can solve it here too if you can wait a bit.

ana-pantilie (changes requested):
Please add the MAlonzo modules and please add Haskell tests, because we also need to be testing the runtime version of this!

Regarding your questions, Lorenzo is working on adding a CI check which fails if you haven't added the MAlonzo files. We also don't need to support the new `if then else` stuff in this PR, please open a separate PR for that. Regarding adding new tests like the one I attached here, I have an issue on that and will be working on it.

plutus-metatheory/test/certifier/Test/Certifier/AST/ForceDelay.hs
```diff
@@ -0,0 +1,295 @@
...
+            , Term Name PLC.DefaultUni PLC.DefaultFun ()
+            )
+            ]
+successItems =
+    [
+        ("Simple one lambda", ForceDelay
+            , simpleSuccessBefore, simpleSuccessAfter)
+        ,("Nested", ForceDelay
+            , nestedBefore, nestedAfter)
+        -- FIXME: This isn't yet implemented in the certifier
```
ana-pantilie, line 261: Please link to the issue tracking this

PR #7092, opened 2025-05-13
Builtin Array example

Unisay:
Closes https://github.com/IntersectMBO/plutus-private/issues/1590

Preview: https://plutus.cardano.intersectmbo.org/pr-preview/pr-7092/working-with-scripts/other-optimization-techniques#using-builtinarray-for-index-based-lookups


doc/docusaurus/docs/working-with-scripts/other-optimization-techniques.md
```diff
@@ -128,6 +128,37 @@ Traces can be expensive especially in terms of script sizes.
...
 
+## Using `BuiltinArray` for index-based lookups
+
+For multiple lookups by index, converting a list to a `BuiltinArray` can be more cost-effective than performing lookups on the list directly. The initial cost of converting the list to an array can be offset by the significantly faster lookups.
+
+As a rule of thumb, if you only need to perform a single lookup, it is likely not worth the conversion cost. The benefits of using a `BuiltinArray` become apparent when you need to perform several lookups on the same data structure.
+
+For instance, a single lookup at index 99 of a 100-element list demonstrates this trade-off:
+*   **Lookup Cost:** The CPU cost for the lookup on a standard Plinth list (a sum-of-products type) is 206 times higher than on a `BuiltinArray`.
+*   **Creation Cost:** However, creating the `BuiltinArray` (by first creating a list and then converting it) is 11 times more expensive than creating the list alone.
```
zliu41: Really? Creating a list is linear, creating an array is also linear, how can it be 11x worse?

Are you possibly trying to say: if you _already_ have a list, then performing a single lookup is 11x faster than converting the entire list to an array, and then performing a single lookup?

doc/docusaurus/docs/working-with-scripts/other-optimization-techniques.md
```diff
@@ -128,6 +128,37 @@ Traces can be expensive especially in terms of script sizes.
 It is advisable to use traces during development, but to remove them when deploying your scripts on mainnet.
 Traces can be removed via the `remove-trace` plugin flag.
 
+## Using `BuiltinArray` for index-based lookups
+
+For multiple lookups by index, converting a list to a `BuiltinArray` can be more cost-effective than performing lookups on the list directly. The initial cost of converting the list to an array can be offset by the significantly faster lookups.
+
+As a rule of thumb, if you only need to perform a single lookup, it is likely not worth the conversion cost. The benefits of using a `BuiltinArray` become apparent when you need to perform several lookups on the same data structure.
```
zliu41: > it is likely not worth the conversion cost.

What about picking list vs. array to use in the first place? If you use array in your data type in the first place, then you won't need the conversion. So should one generally prefer using array or list in the first place, and why?

zliu41 (commented):
It should be clarified that this is an upcoming feature - not usable yet. And I think it deserves more publicity than being a part of "Other Optimization Techniques".

I suggest creating a category of pages, called "Upcoming Features". Under it, add a page called "Builtin Arrays", and put the content there. You can reference it in "Other Optimization Techniques".

doc/docusaurus/docs/working-with-scripts/other-optimization-techniques.md
Unisay: > Really? Creating a list is linear, creating an array is also linear, how can it be 11x worse?

You're right - that comparison doesn't make sense. The benchmark was comparing referencing a compile-time list constant against doing a runtime conversion, which isn't really apples-to-apples.

I've updated the docs to remove the misleading 11x claim and focus on what actually matters: the **206x lookup performance difference**. That's the real story here.

The key points now are:
  - Arrays are much faster for lookups (206x)
  - Array creation currently requires list conversion (with overhead)
  - Future versions will add arrays to `ScriptContext` to avoid conversion costs

Thanks for catching this - the original wording was confusing.

doc/docusaurus/docs/working-with-scripts/other-optimization-techniques.md
Unisay: Good point - avoiding conversion is definitely better than doing it.

- [x]  I added a section on when to choose arrays vs lists from the start. 

(The tricky part is you can't always choose arrays upfront since data often comes as lists from external sources like `ScriptContext`). That's why conversion is common now, and why future versions will provide arrays directly in those cases.)

doc/docusaurus/docs/upcoming-features/builtin-arrays.md
```diff
@@ -0,0 +1,60 @@
...
+**This is an upcoming feature that is not yet available for use.** The `BuiltinArray` type and related functions are currently under development and will be included in a future version of Plutus. This documentation is provided for preview purposes only.
+:::
+
+For multiple lookups by index, `BuiltinArray` provides significantly better performance than lists. The key advantage is in the lookup operations themselves.
+
+**Lookup Performance Comparison:**
+A single lookup at index 99 of a 100-element data structure shows that the CPU cost for lookup on a standard Plinth list (`[Integer]`, a sum-of-products type) is **206 times higher** than on a `BuiltinArray`.
+
+**Important Considerations:**
+Currently, `BuiltinArray` creation is implemented as conversion from lists, which involves traversing the entire list. This conversion cost should be factored into your performance calculations - the dramatic lookup performance improvement needs to be amortized over multiple lookups to justify the conversion overhead.
```
zliu41, line 17: What about the other direction - turning Array into List or Data? If this is not supported, there should be an explanation for why it is not supported.

doc/docusaurus/static/code/Example/Builtin/Array/Main.hs
```diff
@@ -0,0 +1,191 @@
...
+  $$(Plinth.compile [||lookupByIndex (sopListToArray sopListOfInts)||])
+ where
+  lookupByIndex :: BuiltinArray Integer -> Integer
+  lookupByIndex xs = indexArray xs 99
+
+sopListConstruction :: Plinth.CompiledCode [Integer]
+sopListConstruction = $$(Plinth.compile [||sopListOfInts||])
+
+builtinListConstruction :: Plinth.CompiledCode (BuiltinList Integer)
+builtinListConstruction = $$(Plinth.compile [||toOpaque sopListOfInts||])
```
SeungheonOh, line 55: This is unfair against builtinList. It really should build builtin list using `MkCons` instead of first constructing SOP list and converting to builtin list, which will be inevitably more expensive than SOP list. 

Same thing goes for builtinArray construction case; it should first construct builtin list and convert that into array.

doc/docusaurus/docs/upcoming-features/builtin-arrays.md
SeungheonOh: Array to List is possible. You just need to iterate "length of array" times and access each elements. Maybe this could be noted.

SeungheonOh (changes requested):
> I have a comment on builtin list construction not being fair. 

I realized that the way results were printed is misleading in a sense that "construction" budget measurements weren't meant to represent the real cost of construction, but rather to subtract this budget from total `(lookup + construction) - construction` in order to isolate the lookup budget.

I have removed printing construction budgets:
image

doc/docusaurus/docs/upcoming-features/builtin-arrays.md
zliu41: Right. Then the consideration is not just "`BuiltinArray` creation", but also, e.g., storing `BuiltinArray` in `Data`.

doc/docusaurus/static/code/Example/Builtin/Array/Main.hs
Unisay: see my comment above.

doc/docusaurus/docs/upcoming-features/builtin-arrays.md
Unisay: This consideration is written because using arrays for lookups is not universally better than using lists - hence the warning.

Otherwise, this isn't and attempt to document all possible array functionlity, only to answer question "Why arrays?"

SeungheonOh (approved):
Thanks, benchmark cools good now.

PR #7093, opened 2025-05-13
Add `goldenEvalCekCatchBudget` and change existing cases to use it

### LINKED ISSUE #6846: `goldenBudget` should show evaluation result
Currently it only shows the budget. Showing the result is almost always desirable, since we should also verify that the result is correct.

SeungheonOh:
`goldenEvalCekCatchBudget` combines `goldenBudget` and `goldenEvalCekCatch`. Unlike calling two separately, this will only run CEK once.

Previous cases that used `goldenBudget` will use `goldenEvalCekCatchBudget` since it's more sensible to test their evaluation result as well.

Also, changed some cases to use `goldenBundle'`. Looks like these were missed out from #7057.

Closes #6846


plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -259,9 +282,29 @@ runPlcCek values = do
...
-    UPLC.evaluateCekNoEmit
-      PLC.defaultCekParametersForTesting
-      (p ^. UPLC.progTerm)
+        UPLC.evaluateCekNoEmit
+        PLC.defaultCekParametersForTesting
+        (p ^. UPLC.progTerm)
+
+runPlcCekBudget
+  :: (ToUPlc a PLC.DefaultUni PLC.DefaultFun)
+  => [a]
```
effectfully, line 291: We don't seem to use this list functionality anywhere? So can make it just `a`?

plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -139,8 +144,7 @@ goldenBundle
 goldenBundle name x y = do
   goldenPirReadable name x
   goldenUPlcReadable name x
-  goldenEvalCekCatch name [y]
-  goldenBudget name y
```
effectfully, line 143: Do we even `goldenEvalCekCatch` and `goldenBudget` anymore? If not, let's remove them (or one of them, if we need the other) to ensure only `goldenEvalCekCatchBudget` gets used from now on.

effectfully (approved):
Less noise is always a good thing.

plutus-tx/testlib/PlutusTx/Test.hs
SeungheonOh: I agree. It would be simpler to make this just `a` and have some other helper that does `ToUplc a => [a] -> a`.

plutus-tx/testlib/PlutusTx/Test.hs
SeungheonOh: Okay, I'll count this as a greenlight to kill them all. I was little reluctant to just take out functions

plutus-tx/testlib/PlutusTx/Test.hs
Unisay: Yep, when I was implementing `evaluateCompiledCode` I simplified interface to accept a single argument. Users can chain infix applications externally.

ana-pantilie:
@SeungheonOh I added the "no changelog required" label. It's fine to add it to make CI happy if your changes aren't user facing (like in this case).

SeungheonOh:
I don't have permission to add labels nor merge anything(even with CI passing and approved review!) yet. For now, please feel free to merge this whenever CI's happy.

plutus-tx/testlib/PlutusTx/Test.hs
```diff
@@ -201,21 +184,34 @@ goldenPirBy config name value =
...
-  nestedGoldenVsDocM name ".eval" $
-    prettyPlcClassicSimple . view _1 <$> (rethrow $ runPlcCekTrace values)
+goldenEvalCekCatchBudget :: TestName -> CompiledCode a -> TestNested
+goldenEvalCekCatchBudget name compiledCode = do
+  let
+    evalRes = runPlcCekBudget compiledCode
+  nestedGoldenVsDocM name ".eval" $ do
+    either (pretty . show) prettyPlcClassicSimple
+      <$> runExceptT (fst <$> evalRes)
+  nestedGoldenVsDocM name ".budget" $ ppCatch $ do
```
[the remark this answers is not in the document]
effectfully: I think so too. Budget should go first in case the result is large (can happen for early Plutus version of if error message is long).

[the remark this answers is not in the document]
SeungheonOh: My reasoning here is based on this: https://github.com/IntersectMBO/plutus/issues/6846#issuecomment-2662513745

Having two files will give better feedback: one can easily check if budgeting is broken or evaluator is broken or both are broken from file changes without going over each files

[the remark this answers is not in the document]
zliu41: I disagree - you don't need to go over the entire file if the budget changed and the result didn't - the diff would clearly show that, especially if the budget appears before the result like @effectfully said.

Having both in a single file makes it easier to review the change (which benefits both the author and the reviewers): when the budget changes, I always want to verify whether the result changed too. Putting them in separate files means I'd need to look for the corresponding `.eval` file,  which may not be adjacent to the `.budget` file, or included in the PR. With a single file, I can scan through changes much more efficiently.

zliu41 (changes requested):
There are some golden files that need to be removed, e.g. `checkScriptContext2-20.budget.golden`. You can delete all golden files and then re-run the tests.

zliu41 (commented):
Try
```
cabal test plutus-tx-plugin --test-options=--accept
cabal test plutus-benchmark --test-options=--accept
cabal test plutus-ledger-api:plutus-ledger-api-plugin-test --test-options=--accept
cabal test cardano-constitution --test-options=--accept
```

This should fix all the golden files.

zliu41 (approved):

PR #7097, opened 2025-05-14
Use extra-source-files instead of data-files

kwxm:
If the JSON cost model files were altered then the Plutus code wasn't rebuilt, so the changed costs weren't reflected in cost calculations.  It turns out that putting the JSON files in `extra-source-files` instead of `data-files` in `plutus-core.cabal` solves this problem.  We used to do this, but I cleverly changed it in #6149 (despite @zliu41 querying the change) and then spent a long time wondering why rebuilds weren't happening.  I blame Cabal's [uniformative documentation](https://cabal.readthedocs.io/en/latest/cabal-package-description-file.html#pkg-field-extra-source-files).

### DIFF
--- a/plutus-core/plutus-core.cabal
+++ b/plutus-core/plutus-core.cabal
@@ -1,22 +1,27 @@
-cabal-version:   3.0
-name:            plutus-core
-version:         1.45.0.0
-license:         Apache-2.0
+cabal-version:      3.0
+name:               plutus-core
+version:            1.45.0.0
+license:            Apache-2.0
 license-files:
   LICENSE
   NOTICE
 
-maintainer:      michael.peyton-jones@iohk.io
-author:          Plutus Core Team
-synopsis:        Language library for Plutus Core
-description:     Pretty-printer, parser, and typechecker for Plutus Core.
-category:        Language, Plutus
-build-type:      Simple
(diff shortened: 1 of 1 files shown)


plutus-core/plutus-core.cabal
```diff
@@ -1,22 +1,27 @@
-cabal-version:   3.0
-name:            plutus-core
-version:         1.45.0.0
-license:         Apache-2.0
+cabal-version:      3.0
```
kwxm, line 1: Not sure why this extra reformatting happened.

plutus-core/plutus-core.cabal
```diff
@@ -1,22 +1,27 @@
...
   LICENSE
   NOTICE
 
-maintainer:      michael.peyton-jones@iohk.io
-author:          Plutus Core Team
-synopsis:        Language library for Plutus Core
-description:     Pretty-printer, parser, and typechecker for Plutus Core.
-category:        Language, Plutus
-build-type:      Simple
+maintainer:         michael.peyton-jones@iohk.io
```
kwxm, line 9: Maybe we should change this...

zeme-wana (approved):

Unisay (approved):

PR #7099, opened 2025-05-15
Improve MAlzono Code Generation Workflow

zeme-wana:
With this PR:
- stylish-haskell no longer formats `plutus-metatheory/src/MAlonzo` files
- `generate-malonzo-code` is run automatically as a pre-push git hook, and only when there are changes to `plutus-metatheory/src`


ana-pantilie (approved):
Thank you!!

ramsay-t (approved):
LGTM - thanks!

PR #7106, opened 2025-05-19
[Errors] Make '_OperationalEvaluationError' work in the machines

effectfully:
This replaces

```haskell
instance AsEvaluationFailure err => AsEvaluationFailure (ErrorWithCause err cause) 
```

with

```haskell
instance AsEvaluationError err structural operational =>
        AsEvaluationError (ErrorWithCause err cause) structural operational where
```

making error throwing more uniform in the machines (plus it adds functionality that I need in #7029).

Also makes `throwingWithCause` accept a `term` rather than a `Maybe term`, as the name indicates there really should be a cause.

Also replaces 

```haskell
throwing_ _EvaluationFailure
```

with

```haskell
throwingWithCause _OperationalEvaluationError CekEvaluationFailure $ Error ()
```

in the machines, so that the logs clearly indicate that it was `error` that caused an evaluation failure -- not a builtin.

### DIFF
--- a/plutus-core/changelog.d/20250520_160112_effectfully_make__OperationalEvaluationError_work_in_the_machines.md
+++ b/plutus-core/changelog.d/20250520_160112_effectfully_make__OperationalEvaluationError_work_in_the_machines.md
@@ -1,0 +1,3 @@
+### Changed
+
+- In #7106 improved error reporting in the evaluators.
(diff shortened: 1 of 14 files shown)


zliu41:
It seems the key is to replace `throwing_` with `throwingWithCause` - adding the cause. What's the purpose of using `AsEvaluationError` instead of `AsEvaluationFailure`? What's the relationship between these two?

effectfully:
`AsEvaluationError` is more general, it allows one to throw both structural and operational errors. `AsEvaluationFailure` only allows one to throw the single operational error corresponding to an `Error` call. This is way too restrictive and I need other operational errors in the casing-on-booleans branch, so I figured I'd just forbid throwing `Error` this way, so that we uniformly throw all operational errors through `AsEvaluationError`.

Maybe we should replace `_MachineError` with `_StructuralEvaluationError` for consistency too, since the former is a synonym for the latter. Not sure. The former is kinda more readable and is shorter too.

zliu41 (approved):
LGTM. The error handling could indeed use some more restructuring or refactoring. The naming - `AsEvaluationError` vs `AsEvaluationFailure` - doesn't help either.

Issue #7065, opened 2025-04-30
Support base's `==` on Integer


zliu41:
This would allow matching on integer literals. The plugin could match on `integerEq` or `integerEq#`.

PR #7108, opened 2025-05-20
Add support for base integer `==`  (#7065)

SeungheonOh:
Add support for base integer equality, thus integer pattern matching. This is accomplished by injecting `(PlutusTx.==) @Integer` into the code given to `compile` and converting all `GHC.Num.Integer.integerEq` into `PlutusTx.Builtins.equalsInteger` on the plugin side. This does bloat the PIR a little bit because it will include binding for `equalsInteger` even when it's not used, but this won't effect uplc at the end as we expect unused bindings from PIR to be correctly erased trivially. 

This feels like a rather sloppy method. In fact, I had other approach in mind, which is to precompile `(PlutusTx.==) @Integer` and load precompiled PIR from the plugin side, but I wasn't able to figure out a way to get this working with `Bool` being defined twice both from imported snippet and the code that snippet will get embedded to. I'll try to pursue that route again if that one is more preferable.

### DIFF
--- a/plutus-tx-plugin/src/PlutusTx/Plugin.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Plugin.hs
@@ -15,7 +15,7 @@ import Data.Bifunctor
 import PlutusPrelude
 import PlutusTx.AsData.Internal qualified
 import PlutusTx.Bool ((&&), (||))
-import PlutusTx.Builtins (mkNilOpaque, useFromOpaque, useToOpaque)
+import PlutusTx.Builtins (equalsInteger, mkNilOpaque, useFromOpaque, useToOpaque)
 import PlutusTx.Code
 import PlutusTx.Compiler.Builtins
 import PlutusTx.Compiler.Error
@@ -412,6 +412,7 @@ compileMarkedExpr locStr codeTy origE = do
            , 'useToOpaque
            , 'useFromOpaque
            , 'mkNilOpaque
+           , 'PlutusTx.Builtins.equalsInteger           
            ]
   modBreaks <- asks pcModuleModBreaks
   let coverage =
--- a/plutus-tx-plugin/changelog.d/20250523_172915_seungheon.ooh_7065_2.md
+++ b/plutus-tx-plugin/changelog.d/20250523_172915_seungheon.ooh_7065_2.md
@@ -1,0 +1,3 @@
+### Added
+
+- Added support for `integerEq` into the plugin, enabling use of `(==) @Integer` and pattern matching on `Integer` type values. 
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
@@ -1028,9 +1028,8 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
       | GHC.getName n == GHC.eqName ->
           throwPlain $ UnsupportedError "Use of == from the Haskell Eq typeclass"
     GHC.Var n
-      | isProbablyIntegerEq n ->
-          throwPlain $
-            UnsupportedError "Use of Haskell Integer equality, possibly via the Haskell Eq typeclass"
+      | isProbablyIntegerEq n -> do
+          lookupGhcId 'Builtins.equalsInteger >>= compileExpr . GHC.Var
     GHC.Var n
       | isProbablyBytestringEq n ->
           throwPlain $
--- a/plutus-tx-plugin/test/Plugin/Basic/9.6/integerCase.pir.golden
+++ b/plutus-tx-plugin/test/Plugin/Basic/9.6/integerCase.pir.golden
@@ -1,0 +1,32 @@
+let
+  ~ds : integer = 2
+  data Bool | Bool_match where
+    True : Bool
+    False : Bool
+  !equalsInteger : integer -> integer -> bool = equalsInteger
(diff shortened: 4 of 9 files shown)


SeungheonOh:
That was one of my initial attempt, but replacing only works when the program uses `PlutusTx.==` before its encounter with `Base.==` because without having `PlutusTx.==`, program will not include `PlutusTx.Builtins.equalsInteger`.

Also, we can't use `PlutusTx.Builtins.Internals.equalsInteger` because this works with BuiltinBool.

So we need to inject `PlutusTx.==` the way I did it on the `PlutusTx.TH`

zliu41:
I'm not sure I follow. You are effectively rewriting all `integerEq` in the program being compiled with `equalsInteger` - turning
```
...integerEq x y...
```

into
```
...equalsInteger x y...
```

How does it not work? It surely does work if you write the second program by hand?

SeungheonOh:
This doesn't work unless equalsInteger appears before in the program; it will give "equalsInteger not found" error. 

Note this is not the raw builtin. Raw builtin equalsInteger won't work because it returns builtin bool.

zliu41:
> This doesn't work unless equalsInteger appears before in the program; it will give "equalsInteger not found" error.

But how is it different than manually changing all `integerEq` with `equalsInteger` in the source program?

The first time the plugin encounters `equalsInteger` (whether it's originally in the program, or introduced by the compiler to replace `integerEq`) - or any other similar identifier, it should enter [this branch](https://github.com/IntersectMBO/plutus/blob/541aff5c501179cce9c419bc27bc49c0a8ef9a81/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs#L1054), and call `hoistExpr` with the unfolding of `equalsInteger`, to create a binding. If it doesn't, I'd be curious to know why.

SeungheonOh:
Oh I see. Yeah manually replacing all instances of integerEq to equalsInteger will do the same thing. I'll try that and see if that can will improve. 

We would still need the "injection" part in the PlutusTx.

plutus-tx/src/PlutusTx/TH.hs
```diff
@@ -42,11 +46,30 @@ This isn't so bad, since our plc function accepts an argument of any type, so th
...
+PIR and the actual compilation output is not allowed. Instead, we inject the desired functions into
+the plugin's scope at compile time. Once a function is in scope, the plugin can reference it freely,
+removing the need to maintain separate PIR definitions.
+-}
+
+-- This function injects the specified functions into the user’s code by creating let-bindings
+-- that assign each target expression to `_`. These bindings guarantee that the plugin can locate
+-- and reference the functions during lookup.
+-- See Note [Dependency Injection for Plugin]
+injectDeps :: TH.Q TH.Exp -> TH.Q TH.Exp
```
zliu41, line 61: I don't think we need the injection, do we?

SeungheonOh: Without this plugin cannot find definition of the equalsInteger.

zliu41: That's not right. Whatever mechanism that makes it work when one hand-writes `equalsInteger` in the source code, it should also work if the plugin substitutes it in. The plugin is simply replacing `integerEq` with `equalsInteger`, which is no different than if the programmer wrote `equalsInteger` themselves.

SeungheonOh: When programmer writes equalsInteger, GHC will bring the definition of equalsInteger into the CoreExpr. If we just replace the function name, it is possible we are essentially introducing an unbound variable. 

Because plugin is working with CoreExpr, we cannot add new function from a module and expect CoreExpr to bring in those definitons. 

More concretely, without injection,

```hs
(1 PlutusTx.== 2) && (1 Base.== 2)
```
compiles because `PlutusTx.==` brings `PlutusTx.Builtins.equalsInteger` in.

But 
```hs
1 Base.== 2
```
Will not compile because `integerEq` will be converted into `PlutusTx.Builtins.equalsInteger` which is not bound.



The one plugin have available without injecting definition would be the raw builtin, which very confusingly have the same name, `PlutusTx.Builtins.Internal.equalsInteger`. This one doesn't need any injection, but it works with BuiltinBool so this cannot replace `Base.== :: Integer -> Integer -> Bool`

zliu41: > If we just replace the function name, it is possible we are essentially introducing an unbound variable.

If this is true it would be surprising, and probably a GHC bug. Because it implies that the `Id` returned from `GHC.lookupId` contains no unfolding, even when the unfolding should be available.

I'll look into whether this is the case tomorrow.

SeungheonOh: I tried this again, and with the changes I made on plugin, now it works without injection. I apologize for my ignorance, wasted way too much time on this :upside_down_face:

plutus-tx-plugin/src/PlutusTx/Plugin.hs
```diff
@@ -390,6 +390,7 @@ compileMarkedExpr locStr codeTy origE = do
           , 'useToOpaque
           , 'useFromOpaque
           , 'mkNilOpaque
+          , 'PlutusTx.Builtins.equalsInteger
```
SeungheonOh, line 393: Will GHC get pick over `PlutusTx.Builtins.equalsInteger` vs `Builtins.equalsInteger`? Initially, I had `Builtins.equalsInteger` in `builtinNames` but that didn't provided `equalsInteger` in the context and I got `PlutusTx.Builtins.equalsInteger` not found error from plugin; hence I went through the rabbit hole.

PR #7109, opened 2025-05-20
Amend documentation in Builtins.Internal about budgeting and invariants on Bytestring builtins

SeungheonOh:
Adds more notes in Builtins.Internal:
1. notes about execution budget and script failure due to that
2. Invariants on `replicateByte` and `integerToByteString`

~~I will add notes on which function have dynamic costing and which doesn't later today~~

### DIFF
--- a/plutus-tx/src/PlutusTx/Builtins/Internal.hs
+++ b/plutus-tx/src/PlutusTx/Builtins/Internal.hs
@@ -18,8 +18,15 @@
   Most users should not use this module directly, but rather use 'PlutusTx.Builtins'.
 
   Please note that the documentation for each function will only include operational invariants
-  if there are any. This documentation assumes type system correctly enforces and prevents any structural
-  errors on the generated UPLC. See Note [Structural vs operational errors within builtins].
+  if there are any. This documentation assumes that the type system correctly enforces and
+  prevents any structural errors on the generated UPLC. See Note [Structural vs operational errors
+  within builtins].
+
+  Also note that all builtin functions will fail if the CEK machine exceeds its evaluation budget.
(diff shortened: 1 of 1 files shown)


plutus-tx/src/PlutusTx/Builtins/Internal.hs
```diff
@@ -875,7 +881,9 @@ CONVERSION
 
 -- | Converts the given integer to a bytestring. The first argument specifies
 -- endianness (True for big-endian), followed by the target length of the resulting bytestring
--- and the integer itself. See 'PlutusCore.Bitwise.integerToByteString' for its invariances.
+-- and the integer itself. Fails if the target length is greater than 8192 or if the length
+-- argument is 0 and the result won't fit into 8192 bytes.
+-- See 'PlutusCore.Bitwise.integerToByteString' for its invariances in detail.
```
kwxm:
```suggestion
-- See 'PlutusCore.Bitwise.integerToByteString' for its invariants in detail.
```

plutus-tx/src/PlutusTx/Builtins/Internal.hs
```diff
@@ -21,6 +21,12 @@
   Please note that the documentation for each function will only include operational invariants
   if there are any. This documentation assumes type system correctly enforces and prevents any structural
   errors on the generated UPLC. See Note [Structural vs operational errors within builtins].
+
```
kwxm, line 24: A couple of lines above it says "assumes type system". That should probably be "assumes that the type system".

kwxm (approved):
Sorry, just noticed this.  Looks good!

PR #7111, opened 2025-05-21
[Refectoring] Remove Prism from type checking/compiler error handling 

### LINKED ISSUE #6160: Get rid of prismatic error handling?
We use prisms for error handling a lot, an example:

```haskell
instance AsEvaluationError UnliftingEvaluationError UnliftingError UnliftingError where
    _EvaluationError = coerced
    {-# INLINE _EvaluationError #-}

-- | An 'UnliftingEvaluationError' /is/ an 'EvaluationError', hence for this instance we only
-- require both @operational@ and @structural@ to have '_UnliftingError' prisms, so that we can
-- handle both the cases pointwisely.
        AsUnliftingEvaluationError (EvaluationError operational structural) where
        go =
            prism'
                (bimap
                    (review _UnliftingError)
                    (review _UnliftingError))
                (bitraverse
    {-# INLINE _UnliftingEvaluationError #-}

instance AsUnliftingEvaluationError BuiltinError where
    _UnliftingEvaluationError = _BuiltinUnliftingEvaluationError . _UnliftingEvaluationError
    {-# INLINE _UnliftingEvaluationError #-}
```

This all is pretty complex and I've got parts of it wrong in the past. Do we really need it? I don't think we use the `matching` part of any of these prisms anywhere, only the `review` one? So can we just make those into regular functions?

We'll lose some Template Haskell convenience that comes with prismatic error handling, but given the amount of hand-written code, maybe even that will be net positive.

Or do we use the `matching` part somewhere?

SeungheonOh:
closes #6160 

Removed all prisms from type checking and compiler error handling. All functions previously threw unrealized error types were given minimal error type that can cover all kinds of error the function throws. Function calls within a function that throws different error type now requires changing error type using `modifyError` manually. 

[hoist-error](https://hackage.haskell.org/package/hoist-error-0.3.0.0/docs/Control-Monad-Error-Hoist.html) seems like a viable option at first, but all we need is `Control.Monad.Except.modifyError` since error-throwing function calls are always made within another MonadError. `hoistError` is just a more general version of `modifyError`, and in this case we don't need extra generality!

Changes made in this PR arguably increases verbosity as this requires manually lifting error types of each function calls.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/TypeCheck.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/TypeCheck.hs
@@ -7,6 +7,7 @@ module PlutusCore.TypeCheck
     ( ToKind
     , MonadKindCheck
     , MonadTypeCheck
+    , TypeErrorPlc
     , Typecheckable
     -- * Configuration.
     , BuiltinTypes (..)
@@ -30,6 +31,7 @@ import PlutusPrelude
 import PlutusCore.Builtin
 import PlutusCore.Core
 import PlutusCore.Default
+import PlutusCore.Error
 import PlutusCore.Name.Unique
 import PlutusCore.Normalize
 import PlutusCore.Quote
@@ -52,7 +54,7 @@ defKindCheckConfig = KindCheckConfig DetectNameMismatches
 -- | Extract the 'TypeScheme' from a 'BuiltinMeaning' and convert it to the
 -- corresponding 'Type' for each built-in function.
 builtinMeaningsToTypes
-    :: (MonadKindCheck err term uni fun ann m, Typecheckable uni fun)
+    :: (MonadKindCheck (TypeError term uni fun ann) term uni fun ann m, Typecheckable uni fun)
     => BuiltinSemanticsVariant fun
     -> ann
     -> m (BuiltinTypes uni fun)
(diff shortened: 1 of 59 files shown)


plutus-core/plutus-core/src/PlutusCore/TypeCheck/Internal.hs
```diff
@@ -192,9 +191,8 @@ makeLenses ''TypeCheckEnv
 type TypeCheckT uni fun cfg m = ReaderT (TypeCheckEnv uni fun cfg) m
 
 -- | The constraints that are required for kind checking.
-type MonadKindCheck err term uni fun ann m =
-    ( MonadError err m                  -- Kind/type checking can fail
-    , AsTypeError err term uni fun ann  -- with a 'TypeError'.
+type MonadKindCheck error term uni fun ann m =
+    ( MonadError error m                -- Kind/type checking can fail
```
effectfully, line 195: At this point is there even benefit to keeping `MonadError` here? Can it be just `Either`? Can we simply move this functionality into `TypeCheckT`?

Anyway, that's probably is best to investigate separately, so feel free to ignore it here.

SeungheonOh: I'd be happy to kill MonadError as well. It's probably best in a separate PR however. I'll make an issue.

plutus-core/plutus-core/src/PlutusCore/TypeCheck.hs
```diff
@@ -64,28 +66,28 @@ builtinMeaningsToTypes semvar ann =
...
+    :: MonadKindCheck (TypeError term uni fun ann) term uni fun ann m
     => KindCheckConfig -> Type TyName uni ann -> m (Kind ())
 inferKind config = runTypeCheckM config . inferKindM
 
 -- | Check a type against a kind.
 -- Infers the kind of the type and checks that it's equal to the given kind
 -- throwing a 'TypeError' (annotated with the value of the @ann@ argument) otherwise.
 checkKind
-    :: MonadKindCheck err term uni fun ann m
+    :: MonadKindCheck (TypeError term uni fun ann) term uni fun ann m
```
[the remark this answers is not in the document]
SeungheonOh: We need this parameterized since `MonadKindCheck` is also used in PIR type checker.

PR #7113, opened 2025-05-26
Refactor BuiltinList tests to reduce duplication and add index-out-of-bounds case

Unisay:
Chore:
- [x] Replace wildcard imports with explicit ones
- [x] Use `PlutusTx.Prelude` by default to reduce noise.
- [x] Reduce code duplication by extracting a helper function.

Meaning:
- [x] Add a "lookup with an out-ouf-bounds index"  test case.

### DIFF
--- a/plutus-tx-plugin/test/BuiltinList/Budget/9.6/index-out-of-bounds.pir.golden
+++ b/plutus-tx-plugin/test/BuiltinList/Budget/9.6/index-out-of-bounds.pir.golden
@@ -5,34 +5,27 @@ let
   data Unit | Unit_match where
     Unit : Unit
 in
-letrec
-  !go : list integer -> integer -> integer
-    = \(xs : list integer) (i : integer) ->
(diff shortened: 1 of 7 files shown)


plutus-tx-plugin/test/BuiltinList/Budget/Spec.hs
```diff
@@ -1,49 +1,61 @@
-{-# LANGUAGE DataKinds       #-}
-{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE TupleSections     #-}
+{-# LANGUAGE TypeApplications  #-}
```
SeungheonOh, line 5: TypeApplication is not used it seems, and so is TupleSections.

PR #7114, opened 2025-05-26
BuiltinList lookup uses `dropList` builtin to improve performance.

Unisay:
`BuiltinList` lookup made many times cheaper by using the new `dropList` builtin instead of recursion.

Note: this PR is built on top of another PR as suggested by @zliu41 in order to separate concerns and make code review easier.

Closes: https://github.com/IntersectMBO/plutus-private/issues/1589

### DIFF
--- a/plutus-tx/src/PlutusTx/BuiltinList.hs
+++ b/plutus-tx/src/PlutusTx/BuiltinList.hs
@@ -1,4 +1,6 @@
-{-# LANGUAGE LambdaCase #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+
+{-# HLINT ignore "Redundant if" #-}
 
 -- | Functions operating on `BuiltinList`.
 module PlutusTx.BuiltinList (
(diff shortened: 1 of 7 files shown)


effectfully (approved):

plutus-tx/src/PlutusTx/BuiltinList.hs
```diff
@@ -69,21 +71,15 @@ all p = go
...
-      xs
-      ()
+(!!) xs i
+  | i `B.lessThanInteger` 0 = traceError builtinListNegativeIndexError
+  | otherwise =
+      B.caseList
+        (\_ann -> traceError builtinListIndexTooLargeError)
+        (\y _rest _ann -> y)
+        (BI.drop i xs)
+        ()
```
SeungheonOh, line 85: This is outside of the scope of this PR, but these conditions to throw specific trace messages make function less efficient. And when script is compiled with traces turned off, these are checks are with no purpose really. 

For example, this function will be much more efficient as `BI.head $ BI.drop i xs` when script is compiled with traces turned off.

Unisay: In fact, my first implementation was exactly what you wrote `BI.head (BI.drop i xs)` but then I saw it as a breaking backward compatibility in the negative case and decided to preserve the behavior as is (at least for this PR).

Your observation is legit! I propose you bring it up in a more general form "How much of runtime checks we want to use in Plinth std lib" for us to come up with a guiding principle (rule of 👍🏼 ), if possible.

SeungheonOh: Ideally, we want to have two implementation that we can choose depending on rather we are compiling to tracing script or not as we have no use whatsoever for these runtime checks for non-tracing scripts.

This is kind of tricky since this means we either 
- have two separate implementations of same functions 
- or, have some way to signal(Perhaps implicit parameter or nullary typeclass--I don't like either :/ ) implementation to automatically pick what implementation to use.

Both option does feel rather janky at the moment.

PR #7116, opened 2025-05-27
[Refactoring] Remove prismatic error handling from Ck/Cek machines

### LINKED ISSUE #6160: Get rid of prismatic error handling?

SeungheonOh:
Closes #6160 

Removes prismatic error handling from Ck/Cek machine implementations.

There is a very minor performance degradation, and I don't know where it's coming from. I'll see if I can dig into ghc core tomorrow

### DIFF
--- a/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
+++ b/plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
@@ -24,7 +24,6 @@ import PlutusCore.Data
 import PlutusCore.Evaluation.Machine.BuiltinCostModel
 import PlutusCore.Evaluation.Machine.ExBudget
 import PlutusCore.Evaluation.Machine.ExBudgetStream
-import PlutusCore.Evaluation.Result (evaluationFailure)
 import PlutusCore.Pretty
 
 import PlutusCore.StdLib.Data.ScottList qualified as Plc
@@ -32,6 +31,7 @@ import PlutusCore.StdLib.Data.ScottList qualified as Plc
 import Control.Concurrent.MVar
 import Control.Exception
 import Control.Monad
+import Control.Monad.Except
 import Data.Default.Class
 import Data.Either
 import Data.Kind qualified as GHC (Type)
@@ -199,7 +199,7 @@ instance tyname ~ TyName => KnownTypeAst tyname DefaultUni Void where
 instance UniOf term ~ DefaultUni => MakeKnownIn DefaultUni term Void where
     makeKnown = absurd
 instance UniOf term ~ DefaultUni => ReadKnownIn DefaultUni term Void where
-    readKnown _ = throwing _StructuralUnliftingError "Can't unlift to 'Void'"
+    readKnown _ = throwError $ structuralUnliftingError "Can't unlift to 'Void'"
 
 data BuiltinErrorCall = BuiltinErrorCall
     deriving stock (Show, Eq)
@@ -286,7 +286,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
         idAssumeCheckBoolPlc val =
             case asConstant val of
                 Right (Some (ValueOf DefaultUniBool b)) -> pure b
-                _                                       -> evaluationFailure
+                _                                       -> builtinResultFailure
 
     toBuiltinMeaning _semvar IdSomeConstantBool =
         makeBuiltinMeaning
@@ -296,7 +296,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
         idSomeConstantBoolPlc :: SomeConstant uni Bool -> BuiltinResult Bool
         idSomeConstantBoolPlc = \case
             SomeConstant (Some (ValueOf DefaultUniBool b)) -> pure b
-            _                                              -> evaluationFailure
+            _                                              -> builtinResultFailure
 
     toBuiltinMeaning _semvar IdIntegerAsBool =
         makeBuiltinMeaning
@@ -306,7 +306,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
         idIntegerAsBool :: SomeConstant uni Integer -> BuiltinResult (SomeConstant uni Integer)
         idIntegerAsBool = \case
             con@(SomeConstant (Some (ValueOf DefaultUniBool _))) -> pure con
-            _                                                    -> evaluationFailure
+            _                                                    -> builtinResultFailure
 
     toBuiltinMeaning _semvar IdFInteger =
         makeBuiltinMeaning
@@ -397,7 +397,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni ExtensionFun where
 
     toBuiltinMeaning _semvar ErrorPrime =
         makeBuiltinMeaning
-            (evaluationFailure :: forall a. BuiltinResult a)
+            (builtinResultFailure :: forall a. BuiltinResult a)
             whatever
 
     toBuiltinMeaning _semvar Comma =
--- a/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
@@ -17,7 +17,6 @@
 
 module PlutusCore.Builtin.KnownType
     ( BuiltinError
-    , throwBuiltinErrorWithCause
     , KnownBuiltinTypeIn
     , KnownBuiltinType
     , BuiltinResult (..)
@@ -39,10 +38,11 @@ import PlutusCore.Builtin.HasConstant
 import PlutusCore.Builtin.Polymorphism
 import PlutusCore.Builtin.Result
 import PlutusCore.Core
-import PlutusCore.Evaluation.ErrorWithCause
+import PlutusCore.Evaluation.Machine.Exception
 import PlutusCore.Evaluation.Result
 import PlutusCore.Pretty
 
+import Control.Monad.Except
 import Data.Either.Extras
 import Data.Functor.Identity
 import Data.String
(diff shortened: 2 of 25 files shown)


SeungheonOh:
I think this is in reasonable margins of error.

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
```diff
@@ -361,13 +358,6 @@ makeKnownOrFail x = case makeKnown x of
     BuiltinFailure _ _           -> EvaluationFailure
 {-# INLINE makeKnownOrFail #-}
 
--- | Same as 'readKnown', but the cause of a potential failure is the provided term itself.
-readKnownSelf
-    :: (ReadKnown val a, AsUnliftingEvaluationError err, AsEvaluationFailure err)
-    => val -> Either (ErrorWithCause err val) a
-readKnownSelf val = fromRightM (throwBuiltinErrorWithCause val) $ readKnown val
-{-# INLINE readKnownSelf #-}
-
```
Unisay, line 370: Is this change related to the task or is it an opportunistic improvement?

plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
```diff
@@ -184,28 +126,25 @@ variable).
...
-{-# INLINE _StructuralUnliftingError #-}
-
--- | See Note [Structural vs operational errors within builtins]
-_OperationalUnliftingError :: AsBuiltinError err => Prism' err UnliftingError
-_OperationalUnliftingError = _BuiltinUnliftingEvaluationError . _OperationalError
-{-# INLINE _OperationalUnliftingError #-}
 
 throwNotAConstant :: MonadError BuiltinError m => m void
-throwNotAConstant = throwing _StructuralUnliftingError "Not a constant"
+throwNotAConstant = throwError (BuiltinUnliftingEvaluationError $ MkUnliftingEvaluationError $ StructuralError $ MkUnliftingError "Not a constant")
```
Unisay, line 131: Same sentiment here: don't see much value in factoring `throwError` in.

It feels like syntactic win (`throw` prefix is shorter than `throwError .`) IMO doesn't justify the loss of utility (`notAConstantError :: BuiltinError` is potentially usable without throwing)

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -401,7 +401,7 @@ instance (KnownBuiltinTypeIn DefaultUni term Integer, Integral a, Bounded a, Typ
             -- TODO: benchmark an alternative 'integerToIntMaybe', modified from @ghc-bignum@
             if fromIntegral (minBound :: a) <= i && i <= fromIntegral (maxBound :: a)
                 then pure . AsInteger $ fromIntegral i
-                else throwing _OperationalUnliftingError . MkUnliftingError $ fold
+                else throwOperationalUnliftingError . MkUnliftingError $ fold
```
Unisay, line 404:
```suggestion
                else throwError . mkUnliftingError $ fold
```
(Don't apply, this is an example of idea)

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Ck.hs
```diff
@@ -98,15 +100,30 @@ type CkM uni fun s =
...
-    _UnliftingError = _UnliftingErrorVia CkEvaluationFailure
-
 instance Pretty CkUserError where
     pretty CkEvaluationFailure = "The provided Plutus code called 'error'."
 
+builtinErrorToCkEvaluationException
+  :: Maybe (Term TyName Name uni fun ())
+  -> BuiltinError
+  -> CkEvaluationException uni fun
+builtinErrorToCkEvaluationException cause =
```
Unisay, line 110: Observation: you're not factoring in throwing an exception here (which arguably makes more sense for exception than for errors!)

plutus-core/plutus-core/examples/PlutusCore/Examples/Builtins.hs
```diff
@@ -199,7 +198,7 @@ instance tyname ~ TyName => KnownTypeAst tyname DefaultUni Void where
 instance UniOf term ~ DefaultUni => MakeKnownIn DefaultUni term Void where
     makeKnown = absurd
 instance UniOf term ~ DefaultUni => ReadKnownIn DefaultUni term Void where
-    readKnown _ = throwing _StructuralUnliftingError "Can't unlift to 'Void'"
+    readKnown _ = throwStructuralUnliftingError "Can't unlift to 'Void'"
```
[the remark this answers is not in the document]
SeungheonOh: good idea

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
SeungheonOh: It's not possible to define this function without prismatic error type constraint unless I have a copy of this function per error type. 

Instead, I removed this function entirely and handled error when `readKnown` is used where error type is actually realized

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Ck.hs
SeungheonOh: I intentionally prevented myself from making any semantic changes on error types.

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
Unisay: Gotcha! Thanks for the clarification.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/Ck.hs
Unisay: I didn't mean to change the `CkEvaluationException` type (I could have, but that wasn't my point). 
I meant this direction of change `builtinErrorToCkEvaluationException` -> `throwBuiltinErrorToCkEvaluationException`.
The PR doesn't follow this path consistently.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
SeungheonOh: Ah, yes `UnliftingError ~ Text` anyways, and on other cases we were just relying on `IsString UnliftingError` anyways. Making `operationalUnliftingError :: Text -> BuiltinError` would be most reasonable; I agree.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -451,14 +453,34 @@ But in our case this is okay, because:
...
+builtinErrorToCekEvaluationException
+  :: Maybe (Term name uni fun ())
+  -> BuiltinError
+  -> CekEvaluationException name uni fun
+builtinErrorToCekEvaluationException cause = \case
+    (BuiltinUnliftingEvaluationError err) ->
+      ErrorWithCause
+        (bimap UnliftingMachineError (const CekEvaluationFailure) $ unUnliftingEvaluationError err)
+        cause
+    BuiltinEvaluationFailure -> ErrorWithCause (OperationalError CekEvaluationFailure) cause
```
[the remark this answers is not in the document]
SeungheonOh: I'm personally not a huge fan of adding a whole type class for this, but unless we want to change(or merge) `CekUserError` and `CkUserError`, I can't think of a better way without code duplication I previously had.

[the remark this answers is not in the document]
effectfully: It's not the code duplication that I particularly hate here, it's that it propagates into user space (where by "user" I mostly mean "someone who writes tests"). `readKnownSelf` is a handy thing, I don't want to lose it. If it was only about the machines, I'd swallow the pill.

We can't unify `CkEvaluationError` and `CekEvaluationError` into a single type. It's pretty much a coincidence that both are based on `MachineError`, we could've had (and did have) an evaluator that isn't an abstract machine. Plus, `CekUserError` and `CkUserError` do differ, since the CK machine can't with an out-of-budget error.

SeungheonOh:
@effectfully We can merge this if everything looks good

plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
```diff
@@ -184,27 +126,24 @@ variable).
...
+operationalUnliftingError :: Text -> BuiltinError
+operationalUnliftingError =
+  BuiltinUnliftingEvaluationError . MkUnliftingEvaluationError . OperationalError . MkUnliftingError
+{-# INLINE operationalUnliftingError #-}
 
--- | See Note [Structural vs operational errors within builtins]
-_OperationalUnliftingError :: AsBuiltinError err => Prism' err UnliftingError
-_OperationalUnliftingError = _BuiltinUnliftingEvaluationError . _OperationalError
-{-# INLINE _OperationalUnliftingError #-}
+structuralUnliftingError :: Text -> BuiltinError
```
effectfully, line 135: Would be good to keep the "See Note" part, but I'm not gonna block the PR on that.

PR #7121, opened 2025-05-28
Add `makeIsDataAsList` for generating `IsData` instances that uses `List` internally

### LINKED ISSUE #7115: IsData helper for list-encoded product types

Product types in on-chain as Data can be more efficiently stored as `List` instead of `Constr` since former requires one less step for deconstruction. Currently, Plinth users are only given option to use `Constr` for any datatype: `unstableMakeIsData`. Providing template-haskell helper for building list-encoded product types would allow better optimization and control for the users. 

Example provided by @colll78 
```hs
data Neighbor = Neighbor
  { nibble :: Integer
  , prefix :: BuiltinByteString
  , root   :: BuiltinByteString
  }
  deriving stock (Show, Eq, Generic)

instance PlutusTx.ToData Neighbor where
  {-# INLINABLE toBuiltinData #-}
  toBuiltinData :: Neighbor -> PlutusTx.BuiltinData
  toBuiltinData (Neighbor{nibble, prefix, root}) = PlutusTx.toBuiltinData [PlutusTx.toBuiltinData nibble, PlutusTx.toBuiltinData prefix, PlutusTx.toBuiltinData root]

instance PlutusTx.FromData Neighbor where
  {-# INLINABLE fromBuiltinData #-}
  fromBuiltinData :: PlutusTx.BuiltinData -> Maybe Neighbor
  fromBuiltinData neighbor =
    let toNeighbor :: BI.BuiltinData -> Maybe Neighbor
        toNeighbor neighborList' = do
          let neighborList = BI.unsafeDataAsList neighborList'
          nibble <- PlutusTx.fromBuiltinData $ BI.head neighborList
          prefix <- PlutusTx.fromBuiltinData $ BI.head (BI.tail neighborList)
          root <- PlutusTx.fromBuiltinData $ BI.head (BI.tail $ BI.tail neighborList)
          return Neighbor{nibble, prefix, root}
    in
      BI.chooseData neighbor
          Nothing
          Nothing
          (toNeighbor neighbor)
          Nothing
          Nothing

instance PlutusTx.UnsafeFromData Neighbor where
  {-# INLINABLE unsafeFromBuiltinData #-}
  unsafeFromBuiltinData :: PlutusTx.BuiltinData -> Neighbor
  unsafeFromBuiltinData neighbor =
    let bd = BI.unsafeDataAsList neighbor
        nibble = PlutusTx.unsafeFromBuiltinData $ BI.head bd
        prefix = PlutusTx.unsafeFromBuiltinData $ BI.head $ BI.tail bd
        root = PlutusTx.unsafeFromBuiltinData $ BI.head $ BI.tail $ BI.tail bd
    in Neighbor{nibble, prefix, root}
```

_No response_

SeungheonOh:
closes #7115 

`PlutusTx.IsData.TH` in general is rather a mess; there's a ton of unnecessary "helpers" and obfuscation. It would be nice to redo parts of it from scratch since now I introduced a bit of code duplication on top. But perhaps in a separate refactoring PR. 

I'll add some more test cases tomorrow

### DIFF
--- a/plutus-tx-plugin/test/IsData/Spec.hs
+++ b/plutus-tx-plugin/test/IsData/Spec.hs
@@ -128,9 +128,9 @@ tests :: TestNested
 tests =
   testNested "IsData" . pure $
     testNestedGhc
-      [ assertResultTrue "int" (plc (Proxy @"int") (isDataRoundtrip (1 :: Integer)))
-      , assertResultTrue "tuple" (plc (Proxy @"tuple") (isDataRoundtrip (1 :: Integer, 2 :: Integer)))
-      , assertResultTrue
+      [ assertResult "int" (plc (Proxy @"int") (isDataRoundtrip (1 :: Integer)))
+      , assertResult "tuple" (plc (Proxy @"tuple") (isDataRoundtrip (1 :: Integer, 2 :: Integer)))
+      , assertResult
           "tupleInterop"
           ( compiledCodeToHaskUnsafe
               ( plc (Proxy @"tupleInterop")
@@ -142,7 +142,7 @@ tests =
               )
               (plc (Proxy @"tupleInteropArg") (P.toBuiltinData (1 :: Integer, 2 :: Integer)))
           )
-      , assertResultTrue
+      , assertResult
           "unsafeTupleInterop"
           ( compiledCodeToHaskUnsafe
               ( plc (Proxy @"unsafeTupleInterop")
@@ -152,8 +152,8 @@ tests =
               )
               (plc (Proxy @"unsafeTupleInteropArg") (P.toBuiltinData (1 :: Integer, 2 :: Integer)))
           )
-      , assertResultTrue "unit" (plc (Proxy @"unit") (isDataRoundtrip ()))
-      , assertResultTrue
+      , assertResult "unit" (plc (Proxy @"unit") (isDataRoundtrip ()))
+      , assertResult
           "unitInterop"
           ( compiledCodeToHaskUnsafe
               ( plc (Proxy @"unitInterop")
@@ -165,20 +165,20 @@ tests =
               )
               (plc (Proxy @"unitInteropArg") (P.toBuiltinData ()))
           )
-      , assertResultTrue "mono" (plc (Proxy @"mono") (isDataRoundtrip (Mono2 2)))
-      , assertResultTrue "poly" (plc (Proxy @"poly") (isDataRoundtrip (Poly1 (1 :: Integer) (2 :: Integer))))
-      , assertResultTrue "record" (plc (Proxy @"record") (isDataRoundtrip (MyMonoRecord 1 2)))
-      , assertResultTrue "recordAsList" (plc (Proxy @"record") (isDataRoundtrip (MyMonoRecordAsList 1 2)))
-      , assertResultTrue "recordAsList is List"
+      , assertResult "mono" (plc (Proxy @"mono") (isDataRoundtrip (Mono2 2)))
+      , assertResult "poly" (plc (Proxy @"poly") (isDataRoundtrip (Poly1 (1 :: Integer) (2 :: Integer))))
+      , assertResult "record" (plc (Proxy @"record") (isDataRoundtrip (MyMonoRecord 1 2)))
+      , assertResult "recordAsList" (plc (Proxy @"record") (isDataRoundtrip (MyMonoRecordAsList 1 2)))
+      , assertResult "recordAsList is List"
           (plc (Proxy @"record") (
               P.toBuiltinData (MyMonoRecordAsList 1 2)
               P.== (BI.mkList $
                       BI.mkCons (P.toBuiltinData @Integer 1) $
                       BI.mkCons (P.toBuiltinData @Integer 2) $
                       BI.mkNilData BI.unitval)))
-      , assertResultTrue "list" (plc (Proxy @"list") (isDataRoundtrip ([1] :: [Integer])))
-      , assertResultTrue "nested" (plc (Proxy @"nested") (isDataRoundtrip (NestedRecord (Just (1, 2)))))
-      , assertResultTrue
+      , assertResult "list" (plc (Proxy @"list") (isDataRoundtrip ([1] :: [Integer])))
+      , assertResult "nested" (plc (Proxy @"nested") (isDataRoundtrip (NestedRecord (Just (1, 2)))))
+      , assertResult
           "bytestring"
           (plc (Proxy @"bytestring") (isDataRoundtrip (WrappedBS Builtins.emptyByteString)))
       , goldenPirReadable "deconstructData" deconstructData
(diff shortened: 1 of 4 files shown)


plutus-tx-plugin/test/IsData/Spec.hs
```diff
@@ -158,6 +164,7 @@ tests =
       , goldenUEval "mono" [plc (Proxy @"mono") (isDataRoundtrip (Mono2 2))]
       , goldenUEval "poly" [plc (Proxy @"poly") (isDataRoundtrip (Poly1 (1 :: Integer) (2 :: Integer)))]
       , goldenUEval "record" [plc (Proxy @"record") (isDataRoundtrip (MyMonoRecord 1 2))]
+      , goldenUEval "recordAsList" [plc (Proxy @"record") (isDataRoundtrip (MyMonoRecordAsList 1 2))]
```
SeungheonOh, line 167: Need something better than this since `goldenUEval` doesn't check if given value actually is `True` value.

plutus-tx/src/PlutusTx/IsData/TH.hs
```diff
@@ -31,22 +33,23 @@ import PlutusTx.Trace (traceError)
...
-   in
-    createExpr
+  [|BI.mkConstr (conIx :: Integer) $(mkListCreateExpr createFieldNames)|]
 
 mkConstrPartsMatchPattern :: Integer -> [TH.Name] -> TH.PatQ
 mkConstrPartsMatchPattern conIx extractFieldNames =
   let
+    -- Do not change this to raw integer pattern matching.
+    -- GHC will inline `equalsEq` and do integer equality by matching
+    -- them structurally, which the plugin does not know how to handle.
```
Unisay, line 52: Valuable insight 👍🏼

plutus-tx-plugin/test/IsData/Spec.hs
```diff
@@ -40,6 +40,12 @@ import UntypedPlutusCore qualified as UPLC
 
 import Data.Proxy
 
+data MyMonoRecordAsList = MyMonoRecordAsList {mrlA :: Integer, mrlB :: Integer}
+instance P.Eq MyMonoRecordAsList where
+  {-# INLINEABLE (==) #-}
+  (MyMonoRecordAsList x1 y1) == (MyMonoRecordAsList x2 y2) = x1 P.== x2 && y1 P.== y2
+IsData.makeIsDataAsList ''MyMonoRecordAsList
```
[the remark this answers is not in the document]
SeungheonOh: Yes, do we want that?

plutus-tx/src/PlutusTx/IsData/TH.hs
```diff
@@ -31,22 +33,23 @@ import PlutusTx.Trace (traceError)
...
-    createExpr = [|BI.mkConstr (conIx :: Integer) $createArgsExpr|]
-   in
-    createExpr
+  [|BI.mkConstr (conIx :: Integer) $(mkListCreateExpr createFieldNames)|]
 
 mkConstrPartsMatchPattern :: Integer -> [TH.Name] -> TH.PatQ
 mkConstrPartsMatchPattern conIx extractFieldNames =
   let
+    -- Do not change this to raw integer pattern matching.
+    -- GHC will inline `equalsEq` and do integer equality by matching
```
zliu41:
```suggestion
    -- GHC may inline `integerEq` and do integer equality by matching
```

plutus-tx/src/PlutusTx/IsData/TH.hs
```diff
@@ -277,6 +359,69 @@ makeIsDataIndexed dataTypeName indices = do
...
+-}
+makeIsDataAsList :: TH.Name -> TH.Q [TH.Dec]
+makeIsDataAsList dataTypeName = do
+  dataTypeInfo <- TH.reifyDatatype dataTypeName
+  let appliedType = TH.datatypeType dataTypeInfo
+  let nonOverlapInstance = TH.InstanceD Nothing
+
+  cons <-
+    case TH.datatypeCons dataTypeInfo of
+      [cons'] -> pure cons'
```
zliu41:
```suggestion
      [cons] -> pure cons
```

plutus-tx-plugin/test/IsData/Spec.hs
zliu41: Is your concern that `constr 0 []` could represent something else other than True? True, but highly unlikely.

plutus-tx-plugin/test/IsData/Spec.hs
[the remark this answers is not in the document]
zliu41: Yes - because otherwise there's no evidence that `makeIsDataAsList` actually uses the `List` constructor. The existing tests would succeed if it was using `Constr`.

zliu41 (approved):
This needs a changelog - it's a user-facing change. Also needs a golden test showing that it uses the List constructor. Otherwise looks good.

plutus-tx-plugin/test/IsData/Spec.hs
SeungheonOh: It should be okay for now, but we've seen representation of Bool changing(with introduction of SOP) and also more importantly, there's no reason to have this as golden at all! Golden from these cases are just purely unnecessary noises 

I'll make some changes

plutus-tx/testlib/PlutusTx/Test/Util/Apply.hs
```diff
@@ -0,0 +1,91 @@
...
+arguments to already built script in a type safe manner. Example:
+```hs
+foo :: CompiledCode (Integer -> () -> Bool)
+bar :: CompiledCode Integer
+baz :: CompiledCode ()
+
+compiledCodeToHask foo bar baz :: Either String (CompiledCode ())
+```
+-}
+compiledCodeToHask
```
SeungheonOh, line 68: @Unisay I made this cute little helper. You might be interested; no need for spamming `unsafeApplyCode`.

plutus-tx/testlib/PlutusTx/Test/Run/Code.hs
```diff
@@ -64,3 +67,10 @@ assertEvaluatesWithError code = do
             , "Evaluation traces:"
             , Text.unlines evalResultTraces
             ]
+
+assertResultTrue :: TestName -> CompiledCode Bool -> TestNested
+assertResultTrue name code =
```
Unisay: Looks useful! I suggest to drop the `True` suffix like all `asserts` [out there](https://hoogle.haskell.org/?hoogle=assert) do.
```suggestion
assertResult :: TestName -> CompiledCode Bool -> TestNested
assertResult name code =
```

plutus-tx/testlib/PlutusTx/Test/Util/Apply.hs
Unisay: I like ergonomics of it but not sure about semantics: does it translate to the `Apply` constructor in UPLC?

SeungheonOh:
@zliu41 Okay, I'll add pir goldens too. But, I think TH splice goldens are still be very helpful as a quick reference point to see how codegen currently behaves. It would be especially  useful when we need to make changes later on and my memories on how generated code looks like got faded away.

plutus-tx/testlib/PlutusTx/Test/Util/Apply.hs
SeungheonOh: this is just fancy `applyCode` and `unsafeApplyCode`, so it will behave identically to these, which iirc will create `Apply` node on UPLC. 

Perhaps name can be made less confusing. My reasoning for `compiledCodeToHask` was that it lifts functions in UPLC(`CompiledCode (a->b->c)`) into functions in Hask(`CompiledCode a -> CompiledCode b -> CompiledCode c`).

zliu41 (approved):

plutus-tx/testlib/PlutusTx/Test/Util/Apply.hs
Unisay: If it has the same semantics (we're not substituting UPLC application with Haskell application) then I support it!

PR #7124, opened 2025-05-30
Fake PR for benchmarks

kwxm:
Checking to see  if benchmark results have become unstable again.


kwxm:
The results look pretty consistent apart from [this lot](https://github.com/IntersectMBO/plutus/pull/7124#issuecomment-2922722079).  We were getting occasional results like that before though.

PR #7125, opened 2025-05-30
Refactor budget output of CPU, Memory, and Size in golden tests to use numeric underscores for better readability.

Unisay:
Closes https://github.com/IntersectMBO/plutus/issues/7118


plutus-tx/testlib/PlutusTx/Test/Golden.hs
```diff
@@ -169,3 +168,25 @@ goldenEvalCekLog
...
+      . printf "%19s"
+      . reverse
+      . List.intercalate "_"
+      . chunksOf 3
+      . reverse
+      . show @Integer
+      . fromIntegral
+   where
+    chunksOf _ [] = []
+    chunksOf n xs = take n xs : chunksOf n (drop n xs)
```
SeungheonOh, line 192: There's `Data.Text.chunksOf` but `printf` needs `String` :/

plutus-tx/testlib/PlutusTx/Test/Golden.hs
```diff
@@ -169,3 +168,25 @@ goldenEvalCekLog
 goldenEvalCekLog name value =
   nestedGoldenVsDocM name ".eval" $
     prettyPlcClassicSimple . view _1 <$> rethrow (runPlcCekTrace value)
+
+prettyBudget :: PLC.ExBudget -> Size -> Doc ann
+prettyBudget (PLC.ExBudget (ExCPU cpu) (ExMemory mem)) (Size size) =
+  vsep
+    [ fill 8 "CPU:" <+> prettyIntRightAligned (fromSatInt @Int cpu)
+    , fill 8 "Memory:" <+> prettyIntRightAligned (fromSatInt @Int mem)
```
SeungheonOh, line 176: What was the reason for the change to "CPU" and "Memory"? For consistency with `Pretty ExBudget` instance, I think we should stay with "cpu" and "mem".

https://github.com/IntersectMBO/plutus/blob/de1b3ef1a5677d8b636054cf4005a91dcc6cb733/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExBudget.hs#L198-L202

SeungheonOh (changes requested):
Left some comments, looks good in general. This is great!

plutus-tx/testlib/PlutusTx/Test/Golden.hs
```diff
@@ -169,3 +168,25 @@ goldenEvalCekLog
...
+    , fill 8 "Size:" <+> prettyIntRightAligned size
+    ]
+ where
+  prettyIntRightAligned :: (Integral i) => i -> Doc ann
```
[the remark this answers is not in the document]
Unisay: I made it "private" intentionally because its specific to the `prettyBudget`: it hardcodes the maximum width of a label (8), in this case the longest label is "Memory:". If someone calls it with a longer label then it will misbehave.

plutus-tx/testlib/PlutusTx/Test/Golden.hs
effectfully: We should make the official names "CPU", "MEM" and "Size". The first two are metrics of script execution and they should look similar. The last one isn't and it's fine that it doesn't look alike.

effectfully (approved):
Noice.

plutus-tx/testlib/PlutusTx/Test/Golden.hs
SeungheonOh: I won't block the PR over this, after all it's just aesthetic nitpicking.

This is my further reasoning for documentation purposes:
For me, CPU and Memory having different lengths made it feel more disconnected although they are both ExBudget and would always go together--this is unlike size which is outside of ExBudget.

effectfully: @SeungheonOh wanna shepherd that discussion? Move it to Slack, raise it during the technical meeting etc.

PR #7128, opened 2025-05-30
Move `caseList` and `caseData` to the end of the flat encoding

kwxm:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1613.

It seems that we may decide not to release the `caseList` and `caseData` builtins.  This PR moves them to the very end of the flat encoding (tags 126 and 127) and adjusts the tags for `dropList` and the array builtins (which we're definitely going to release) to fill the gap.  The point of this is to make the changes as early as possible to stabilise the encoding before anything gets released and to make it easier to remove `caseList` and `caseData` if we decide to do  so (which will involve further changes to the implementation and specification).

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
@@ -178,13 +178,14 @@ data DefaultFun
     | Ripemd_160
     -- Batch 6
     | ExpModInteger
-    | CaseList
-    | CaseData
     | DropList
     -- Arrays
     | LengthOfArray
     | ListToArray
     | IndexArray
+    -- Case
+    | CaseList
+    | CaseData
     deriving stock (Show, Eq, Ord, Enum, Bounded, Generic, Ix)
     deriving anyclass (NFData, Hashable, PrettyBy PrettyConfigPlc)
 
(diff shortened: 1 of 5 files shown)


plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -178,13 +178,14 @@ data DefaultFun
...
     | ExpModInteger
-    | CaseList
-    | CaseData
     | DropList
     -- Arrays
     | LengthOfArray
     | ListToArray
     | IndexArray
+    -- Case
+    | CaseList
```
zliu41, line 187: Does the order of constructors here matter?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -2281,111 +2282,111 @@ instance Flat DefaultFun where
...
 
-              DropList                        -> 90
+              LengthOfArray                   -> 89
+              ListToArray                     -> 90
+              IndexArray                      -> 91
 
-              LengthOfArray                   -> 91
-              ListToArray                     -> 92
-              IndexArray                      -> 93
+              CaseList                        -> 126
```
zliu41, line 2291: Is there a max bound for the tag number?

zliu41 (approved):

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
kwxm: > Does the order of constructors here matter?

No, I don't think so.  The main thing that should never change after deployment is the flat tag.  I think there are some things (tests etc) that iterate over the enumeration of all the builtins, so rearranging the constructors may alter the order that things happen in.  I just moved these to the end to make them easy to find.

PR #7138, opened 2025-06-11
[Builtins] Make costing of 'equalsData' more accurate

effectfully:
This makes the user not pay for so the nodes that `equalsData` ends up not processing (when `False` is returned early, because two nodes don't match and there's no point in traversing the `Data` objects further).


SeungheonOh:
As demonstrated [here](https://github.com/IntersectMBO/plutus/issues/7142#issuecomment-3034295837), costing update here doesn't really benefit script evaluations in the node. 

In short, this is because most validators when they use `equalsData` either use it on two data of same structure, like checking equality of `TxOutRef`. When they use `equalsData` on data of unknown structure, it's frequently comparing `Redeemer` or `Datum` which validators often expect `equalsData` to be true. So cases where new costing of `equalsData` actually being beneficial is very rare.

PR #7140, opened 2025-06-11
List handling in uglyTmCon

ramsay-t:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1603

### DIFF
--- a/plutus-metatheory/src/Untyped.lagda.md
+++ b/plutus-metatheory/src/Untyped.lagda.md
@@ -76,8 +76,8 @@ uglyTmCon : TmCon → String
 {-# TERMINATING #-}
 uglyTmConList : (t : TyTag) → ⟦ list t ⟧tag → String
 uglyTmConList t [] = ""
-uglyTmConList t (x ∷ []) = uglyTmCon ((tmCon t x))
-uglyTmConList t (x ∷ l ∷ ls) = uglyTmCon ((tmCon t x)) ++ " , " ++ (uglyTmConList t (l ∷ ls))
+uglyTmConList t (x ∷ []) = uglyTmCon (tmCon t x)
+uglyTmConList t (x ∷ l ∷ ls) = uglyTmCon (tmCon t x) ++ " , " ++ (uglyTmConList t (l ∷ ls))
 
 uglyTmCon (tmCon integer x)              = "(integer " ++ show x ++ ")"
 uglyTmCon (tmCon bytestring x)           = "bytestring"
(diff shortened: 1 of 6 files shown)


plutus-metatheory/src/Untyped.lagda.md
```diff
@@ -72,6 +72,13 @@ uglyDATA : DATA → String
 uglyDATA d = "(DATA)"
 
 uglyTmCon : TmCon → String
+
+{-# TERMINATING #-}
+uglyTmConList : (t : TyTag) → ⟦ list t ⟧tag → String
+uglyTmConList t [] = ""
+uglyTmConList t (x ∷ []) = uglyTmCon ((tmCon t x))
```
kwxm: Are the double parentheses significant, or just accidental?  (Similarly in the next line).

ramsay-t: Huh.... No, double curly brackets are Agda significant, but not these :)

plutus-metatheory/src/Untyped.lagda.md
```diff
@@ -72,6 +72,13 @@ uglyDATA : DATA → String
 uglyDATA d = "(DATA)"
 
 uglyTmCon : TmCon → String
+
+{-# TERMINATING #-}
+uglyTmConList : (t : TyTag) → ⟦ list t ⟧tag → String
+uglyTmConList t [] = ""
```
kwxm, line 78: I guess there's are no enclosing `[]` and maybe a trailing comma?  That really is ugly!

plutus-metatheory/src/Untyped.lagda.md
```diff
@@ -72,6 +72,13 @@ uglyDATA : DATA → String
 uglyDATA d = "(DATA)"
 
 uglyTmCon : TmCon → String
+
+{-# TERMINATING #-}
```
kwxm, line 76: Can Agda not work out that it's terminating itself?

kwxm (approved):
Looks good.

plutus-metatheory/src/Untyped.lagda.md
ramsay-t: The brackets are added in the parent function. There are three patterns, not two, precisely to avoid the trailing comma...

plutus-metatheory/src/Untyped.lagda.md
ramsay-t: No, not when it is mutually recursive. It doesn't know that `(list (list (list t)))` is finite.

plutus-metatheory/src/Untyped.lagda.md
kwxm: Oh right: I was looking for [x] instead of x::[]. In that case it's not ugly at all!

PR #7141, opened 2025-06-11
Kwxm/spec/indices start at one

kwxm:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1566

In UPLC you can have case expressions of the form `(case M N_1 ... N_n)` and the i'th branch is selected if M evaluates to an term of the form `(constr i M_1 ... M_m)`.  However, the list of alternatives in `case` is allowed to be empty: `(lam x (case x))` is a legal term (which can be returned by a lambda for example), but an error will occur if it's ever applied to an argument because there are no branches to choose from.  The specification didn't cover this in the reduction semantics and the CEK machine, where the argument list was of the form `M_0 ... M_n`, so I've changed those so that indexing starts at 1, with the empty list covered by the case n=0: this notation is already used elsewhere in the specification.  

It's arguable that this choice is bad, since for example, if `case` is given a constructor with tag `i`, branch `i+1` is selected. We could instead have continued to use 0-based indexing but then we'd either have needed special cases for empty lists or allowed n=-1.  Neither of these seemed very appealing.


doc/plutus-core-spec/notation.tex
```diff
@@ -114,13 +114,14 @@ \subsection{Bytestrings and bitstrings}
...
   \textit{left} end but the bits in bitstrings from the \textit{right} end.
 
-\item The bytestring $[c_0, \ldots, c_n]$ may be denoted by $c_0{\cdots}c_n$ ($n
-  \geq 1$, $c_i \in \B$);  the empty bytestring may be denoted by $\epsilon$.
-
-\item The bitstring $[b_n, \ldots, b_0]$ may be denoted by $b_n{\cdots}b_0$ ($n
-  \geq 1$, $b_i \in \b$); the empty bitstring may be denoted by $\epsilon$: we
-  also use this symbol for the empty bytestring, but this should not cause any
-  confusion.
+\item The bytestring $[c_0, \ldots, c_{n-1}]$ may be denoted by
```
kwxm, line 117: It's maybe a bit inconsistent to use zero-based indexing here, but it makes things a lot easier when dealing with the bitwise operations.  Note also that it was already like this: the only thing that's changed is the formatting.

doc/plutus-core-spec/cardano/builtins1.tex
```diff
@@ -236,13 +236,13 @@ \subsubsection{Built-in functions}
                                                          \errorX & \text{otherwise}
                                                        \end{cases}$} & Yes & \ref{note:consbytestring}\\[2mm]
     \TT{sliceByteString}        & $[\ty{integer}, \ty{integer}, \ty{bytestring]} $  \text {$\;\; \to  \ty{bytestring}$}
-                                                   &   $(s,k,[c_0,\ldots,c_n])$ \text{$\;\;\mapsto [c_{\max(s,0)},\ldots,c_{\min(s+k-1,n-1)}]$}
+                                                   &   $(s,k,[c_1,\ldots,c_n])$ \text{$\;\;\mapsto [c_{\max(s+1,1)},\ldots,c_{\min(s+k,n)}]$}
```
kwxm, line 239: I changed these to try to be a bit more consistent.  That might be confusing because as far as the builtins are concerned the indices start at zero (but again we'd need special cases for the emty list handle it by allowing or allowing n=-1).  The specifications of the array operations already use 1-based indices though.

zliu41 (approved):
:+1:

kwxm:
Hydra's stuck so I'll just merge this without waiting.

PR #7144, opened 2025-06-12
Add plutus-metatheory to CHaP release process

kwxm:
The CHaP PR in the release process was failing, apparently because it didn't know about `plutus-metatheory`.  This adds `plutus-metatheory` to the list of stuff to be released to CHaP.

Also there were a lot of trailing spaces which I've deleted, which may confuse matters.

### DIFF
--- a/scripts/interactive-release.sh
+++ b/scripts/interactive-release.sh
@@ -4,11 +4,11 @@
 #
 # 1. Open a new Release PR in plutus
 # 2. Review and approve the Release PR in plutus, do not merge it yet
-# 3. Open and merge a new Plutus Release PR in CHaP 
+# 3. Open and merge a new Plutus Release PR in CHaP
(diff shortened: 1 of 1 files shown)


scripts/interactive-release.sh
```diff
@@ -185,7 +185,7 @@ open-chap-pr() {
   cd cardano-haskell-packages
   git checkout -b $PR_BRANCH
   tell "Running ./scripts/add-from-github.sh with COMMIT_SHA=$COMMIT_SHA"
-  local RELEASE_PACKAGES=(plutus-core plutus-ledger-api plutus-tx plutus-tx-plugin)
+  local RELEASE_PACKAGES=(plutus-core plutus-ledger-api plutus-metatheory plutus-tx plutus-tx-plugin)
```
kwxm, line 188: This is the only change that matters!

ana-pantilie (approved):

PR #7148, opened 2025-06-16
Add a way to generate callstack trace message

SeungheonOh:
Add plugin option to generate callstack trace message. Currently this won't handle cases when a function has `PlutusTx.Trace.callStack` and that function is called in multiple places (working on it. more detail on first commit message).

<details>
<summary>Example</summary>
```hs
-- SillyCalss.hs
class SillyClass a where
  sillyThing :: a -> Integer

instance SillyClass Integer where
  sillyThing = traceError callStack

-- Demo.hs
rob :: Integer -> Integer -> Integer
rob x = sillyThing @Integer

-- entrypoint
tom :: Integer -> Integer
tom = rob (x PTx.+ 20)

{-
tracemsg :
tom:demo/Demo.hs:44:1-44:3
\-rob:demo/Demo.hs:32:1-32:3
\-$fSillyClassInteger
\-$fSillyClassInteger_$csillyThing
-}
```
<\details>

### DIFF
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE MultiWayIf            #-}
 {-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE RecordWildCards       #-}
 {-# LANGUAGE TemplateHaskellQuotes #-}
 {-# LANGUAGE TupleSections         #-}
 {-# LANGUAGE TypeApplications      #-}
@@ -72,12 +73,13 @@ import Control.Exception (displayException)
 import Control.Lens hiding (index, strict, transform)
 import Control.Monad
 import Control.Monad.Reader (ask, asks, local)
+import Control.Monad.State (get)
 import Data.Array qualified as Array
 import Data.ByteString qualified as BS
 import Data.ByteString.Base16 qualified as Base16
 import Data.ByteString.Char8 qualified as BSC
 import Data.Generics.Uniplate.Data (transform, universeBi)
-import Data.List (elemIndex, isPrefixOf, isSuffixOf)
+import Data.List (elemIndex, intercalate, isPrefixOf, isSuffixOf)
 import Data.Map qualified as Map
 import Data.Maybe (mapMaybe)
 import Data.Set qualified as Set
@@ -838,6 +840,8 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
     } <-
     ask
 
+  CompileOptions {..} <- asks ccOpts
+
   -- TODO: Maybe share this to avoid repeated lookups. Probably cheap, though.
   builtinIntegerTyCon <- lookupGhcTyCon ''BI.BuiltinInteger
   builtinBoolTyCon <- lookupGhcTyCon ''BI.BuiltinBool
@@ -857,6 +861,7 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
   boolOperatorOr <- lookupGhcName '(PlutusTx.Bool.||)
   boolOperatorAnd <- lookupGhcName '(PlutusTx.Bool.&&)
   inlineName <- lookupGhcName 'PlutusTx.Optimize.Inline.inline
+  callStackName <- lookupGhcName 'PlutusTx.Trace.callStack
 
   case e of
     {- Note [Lazy boolean operators]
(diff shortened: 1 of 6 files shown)


SeungheonOh:
Figured out a way to do "real" callstack, with some runtime overhead. This will handle something like this correctly:

```hs
silly, silly2, silly3, silly4 :: Integer -> Integer
silly x = silly2 x
silly2 x = silly3 x
silly3 x = silly4 x
silly4 x = bob

bob, rob, cob :: Integer
bob = traceError callStack
rob = traceError callStack
cob = traceError callStack

tom :: Integer -> Integer
tom 1 = silly 1
tom 2 = rob
tom 3 = cob
tom _ = 42
```
```
[||tom 1||]
tom:demo/Demo.hs:62:1-62:3
\-silly:demo/Demo.hs:51:1-51:5
\-silly2:demo/Demo.hs:52:1-52:6
\-silly3:demo/Demo.hs:53:1-53:6
\-silly4:demo/Demo.hs:54:1-54:6
\-bob:demo/Demo.hs:57:1-57:3

[||tom 2||]
tom:demo/Demo.hs:62:1-62:3
\-rob:demo/Demo.hs:58:1-58:3

[||tom 3||]
tom:demo/Demo.hs:62:1-62:3
\-cob:demo/Demo.hs:59:1-59:3
```

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -610,8 +612,49 @@ for any variables that were freshly created by the simplifier. That's easy to fi
...
+  csFuns <- gets csNeedsCallStack
+  if Set.member (LexName $ GHC.varName currVar) csFuns
+    then do
+      lastCs <- lastCallStackName
+      let
+        nameStr v = GHC.occNameString $ GHC.occName $ GHC.varName $ v
+        printCallStackEntry v =
+          case getVarSourceSpan v of
+            -- FIXME: Variables will miss span information when the module they are defined
+            -- in is loaded from cache instead of getting compiled.
```
SeungheonOh, line 631: Is there a workaround for this?

plutus-tx-plugin/test/Plugin/CallStack/9.6/func01.eval.golden
```diff
@@ -0,0 +1,10 @@
+An error has occurred:
+The machine terminated because of an error, either from a built-in function or from an explicit use of 'error'.
+Caused by: error
+
+Trace:
+func:test/Plugin/CallStack/Spec.hs:70:1-70:4
+\-nestedLinear:test/Plugin/CallStack/Spec.hs:63:1-63:12
+\-nestedLinear2:test/Plugin/CallStack/Spec.hs:64:1-64:13
+\-nestedLinear3:test/Plugin/CallStack/Spec.hs:65:1-65:13
+\-nestedLinear4:test/Plugin/CallStack/Spec.hs:66:1-66:13
```
SeungheonOh, line 10: The span provided points to the definition of executed function instead of location of invocation. Is this fine? 

... I think it's fine

plutus-tx-plugin/changelog.d/20250618_025657_seungheon.ooh_7146.md
```diff
@@ -0,0 +1,4 @@
+### Added
+- `generate-callstack` option: When enabled, the plugin replaces every occurrence of `PlutusTx.Trace.callStack` with a `BuiltinString` containing the callstack at that relative point. Enabling this option incurs execution-cost overhead: it adds an extra callstack argument to all user-defined functions, and every time they are invoked it requires having extra application node for providing the current callstack. When `gennerate-callstack` is not enabled, `PlutusTx.Trace.callStack` is string "<CallStack>".
```
Unisay, line 2:
```suggestion
- `generate-callstack` option: When enabled, the plugin replaces every occurrence of `PlutusTx.Trace.callStack` with a `BuiltinString` containing the callstack at that relative point. Enabling this option incurs execution-cost overhead: it adds an extra callstack argument to all user-defined functions, and every time they are invoked it requires having extra application node for providing the current callstack. When `generate-callstack` is not enabled, `PlutusTx.Trace.callStack` is string "<CallStack>".
```

plutus-tx-plugin/src/PlutusTx/Compiler/Types.hs
```diff
@@ -69,16 +71,37 @@ data CompileContext uni fun = CompileContext
...
+  most recent callstack.
+  -}
+  , csCallStackDeps  :: Set LexName
+  {- ^ Set of names that requires callstack. A name being in this set indicates that their
+  corresponding term is modified to have extra string argument and need to apply callstack
+  before using the function.
+  -}
   }
 
+lastCallStackName :: MonadState CompileState m => m (Maybe (PIR.Name))
```
Unisay, line 93:
```suggestion
lastCallStackName :: MonadState CompileState m => m (Maybe PIR.Name)
```

plutus-tx-plugin/src/PlutusTx/Options.hs
```diff
@@ -309,6 +310,9 @@ pluginOptions =
             rest <- many (alphaNumChar <|> char '_' <|> char '\\')
             pure (firstC : rest)
        in (k, PluginOption typeRep (plcParserOption p k) posCertify desc [])
+    , let k = "generate-callstack"
+          desc = "Generate callstack string and replace 'callStack' with callstack at the location."
```
Unisay, line 314: What does the "replace 'callStack' with callstack at the location." part mean for a compiler user?
I don't understand it myself, don't think that end-users will get it either.

plutus-tx-plugin/test/Plugin/CallStack/9.6/func01.eval.golden
Unisay: This PR is  meant to achieve the goal stated in the issue it is associated with. From the issue description "is useful for determining which part of validator in particular cause the failure." I deduce that goal is for on-chain program authors to be able to determine "which part of a validator causes the failure". 

Given two possible implementations:
- One that prints locations of invocations.
- Another that prints locations of definitions.

I'd say that its straightforward how the former one could help authors to determine the failing validator part, and is *not straightforward* if the latter always helps.

Is there a fundamental reason why we can't have invocation locations?
Please document it in the note, if that's the case.

plutus-tx-plugin/test/Plugin/CallStack/9.6/func02.eval.golden
```diff
@@ -0,0 +1,9 @@
+An error has occurred:
+The machine terminated because of an error, either from a built-in function or from an explicit use of 'error'.
+Caused by: error
+
+Trace:
+func:test/Plugin/CallStack/Spec.hs:70:1-70:4
+\-nestedLinear2:test/Plugin/CallStack/Spec.hs:64:1-64:13
+\-nestedLinear3:test/Plugin/CallStack/Spec.hs:65:1-65:13
+\-nestedLinear4:test/Plugin/CallStack/Spec.hs:66:1-66:13
```
Unisay, line 9: Why did you use `\` backslash?  I'd find it more useful to have some notion of a direction here: should I read it top to bottom or vice-versa? Maybe 1. 2. 3. is better then?

SeungheonOh:
superseded by #7178 unless someone actually prefers this method

PR #7149, opened 2025-06-16
Range check for constr tags in parsers

kwxm:
The tag in a `constr` term must be a Word64, but the parsers for textual UPLC/PLC/PUR would accept larger values and wrap them round into the Word64 range without complaining.  This adds a check in the parsers and also some conformance tests involving illegally large tags.

This problem doesn't arise in `flat`: an error will occur during deserialisation if the tag is too large.


plutus-conformance/test-cases/uplc/evaluation/term/constr/constr-01/constr-01.uplc
kwxm, line 1: I renamed all of these because we now have ten tests and I wanted them to remain in numerical order (so `constr-10` comes after `constr-09`, not between `constr-1` and `constr-2`)

plutus-core/plutus-ir/src/PlutusIR/Parser.hs
```diff
@@ -104,10 +105,15 @@ errorTerm _tm = withSpan $ \sp ->
...
-    pure res
+constrTerm tm = withSpan $ \sp ->
+    inParens $ do
+      let maxTag = fromIntegral (maxBound :: Word64)
+      ty <- symbol "constr" *> pType
+      tag :: Integer <- lexeme Lex.decimal
+      args <- many tm
+      whenVersion (\v -> v < plcVersion110) $ fail "'constr' is not allowed before version 1.1.0"
+      when (tag > maxTag) $ fail "constr tag too large: must be a legal Word64 value"
+      pure $ Constr sp ty (fromIntegral tag) args
```
SeungheonOh: `PIR.constr` as it was originally

kwxm: > `PIR.constr` as it was originally

Thanks!  I pretty much just pasted in what I'd done for PLC without looking to closely at the old version.  I've fixed it now.

SeungheonOh (approved):

kwxm:
I'm going to bypass CI because it's all broken.

PR #7150, opened 2025-06-16
Extra conformance test for empty case

kwxm:
This just adds an extra conformance test to check that `(lam x (case x))` succeeds.  This is an edge case that was unclear in the specification until recently (see #7141) and which we weren't previously checking.

### DIFF
--- a/plutus-conformance/test-cases/uplc/evaluation/term/case/case-09/case-09.uplc
+++ b/plutus-conformance/test-cases/uplc/evaluation/term/case/case-09/case-09.uplc
@@ -1,5 +1,5 @@
--- It's legal to have a case term with no branches,
--- but applying it causes an error.
+-- It's legal to have a case term with no branches, but using it to branch on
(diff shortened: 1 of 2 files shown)


plutus-conformance/test-cases/uplc/evaluation/term/case/case-09/case-09.uplc
```diff
@@ -0,0 +1,5 @@
+-- It's legal to have a case term with no branches,
+-- but applying it causes an error.
```
SeungheonOh: Probably I'm dumb but this took me a while to process. Maybe we can use a clearer wording: "but casing anything on no-branch case term causes an error".

kwxm: OK, I've tried to clarify that.

SeungheonOh (approved):

kwxm:
I'm going to bypass CI here too (too many failures).

PR #7154, opened 2025-06-18
Correct the specification of the flat encoding of `constr` tags

kwxm:
Fixes https://github.com/cardano-scaling/cardano-blueprint/issues/32
Fixes https://github.com/IntersectMBO/plutus-private/issues/1620

The specification said that tags in `constr` terms should be encoded using a bitstring of length 64.  This is 
incorrect: they should be encoded using the standard variable-length encoder for natural numbers. 

See also #7149, which made the textual parser check tag values more carefully.

### DIFF
--- a/doc/plutus-core-spec/plutus-core-specification.tex
+++ b/doc/plutus-core-spec/plutus-core-specification.tex
@@ -5,7 +5,7 @@
   \LARGE{\red{\textsf{DRAFT}}}
 }
 
-\date{11th June 2025}
+\date{18th June 2025}
 \author{Plutus Core Team}
 
 \input{header.tex}


doc/plutus-core-spec/flat-serialisation.tex
```diff
@@ -133,11 +133,10 @@ \subsection{Fixed-width natural numbers}
 by
 $$
 \D_n(b_{n-1}\cdots{b_0} \cdot s)= (s,\sum^{n-1}_{i=0}b_i2^i).
-$$ As in Appendix~\ref{appendix:data-cbor-encoding}, $\Nab{a}{b}$
-denotes the closed interval of integers $\{n \in \Z : a \leq n \leq b\}$.  Note
-that $n$ here is a variable (not a fixed label) so we are defining whole
-families of encoders $\E_1, \E_2, \E_3, \ldots$ and and decoders $\D_1, \D_2,
```
kwxm, line 139: "and and"

kwxm:
Bypassing checks because of Hydra problems again.

doc/plutus-core-spec/untyped-grammar.tex
kwxm, line 1: Just fixing a few small formatting discrepancies while I'm here.

PR #7156, opened 2025-06-18
Add cardano open oracle protocol scripts for `plutus-benchmark`

SeungheonOh:
Implements [Cardano Open Oracle Protocol](https://github.com/mlabs-haskell/cardano-open-oracle-protocol/) scripts.

### DIFF
--- a/plutus-benchmark/coop/src/PlutusBenchmark/Coop/Utils.hs
+++ b/plutus-benchmark/coop/src/PlutusBenchmark/Coop/Utils.hs
@@ -1,0 +1,61 @@
+{-# LANGUAGE NamedFieldPuns    #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications  #-}
+
+module PlutusBenchmark.Coop.Utils where
+
+import PlutusTx.Prelude
+import Prelude ()
+
+import PlutusLedgerApi.V1.Value
+import PlutusLedgerApi.V3
+import PlutusLedgerApi.V3.MintValue
+import PlutusTx.AssocMap
+import PlutusTx.AssocMap qualified as AssocMap
+import PlutusTx.Builtins.Internal qualified as BI
+import PlutusTx.List
+
+findOwnInput :: [TxInInfo] -> TxOutRef -> TxInInfo
+findOwnInput inputs oref =
+  case find (\i -> txInInfoOutRef i == oref) inputs of
+    Nothing -> traceError "findOwnInput: not found"
+    Just x  -> x
+
+mustBurnOwnSingletonValue :: ScriptContext -> BuiltinUnit
+mustBurnOwnSingletonValue (ScriptContext (TxInfo {..}) _ (SpendingScript oref _)) =
+  let (TxInInfo _ (TxOut {txOutValue = ownInputValue})) = findOwnInput txInfoInputs oref
+  in case flattenValue ownInputValue of
+    [_ada, (cs, tk, q)] ->
+      if valueOf (mintValueToValue txInfoMint) cs tk == negate q
+      then BI.unitval
+      else traceError "Must burn the all of the single asset this utxo was holding"
+    _ -> traceError "The UTXO should exactly have one assets besides Lovelace"
+mustBurnOwnSingletonValue _ = traceError "Only spending purpose is supported"
+
+resolveDatum :: forall a. UnsafeFromData a => Map DatumHash Datum -> OutputDatum -> a
+resolveDatum datums outputDatum =
+  case outputDatum of
+    NoOutputDatum -> traceError "expected datum but got no datum"
+    OutputDatumHash hash ->
+      case lookup hash datums of
+        Nothing        -> traceError "expected datum but given datum hash have no associated datum"
+        Just (Datum d) -> unsafeFromBuiltinData @a d
+    OutputDatum (Datum d) -> unsafeFromBuiltinData @a d
+
+currencyValue :: CurrencySymbol -> Value -> Value
+currencyValue cs val = withCurrencySymbol cs val mempty (\v -> Value $ AssocMap.singleton cs v)
+
+unsafeMergeMap :: AssocMap.Map k v -> AssocMap.Map k v -> AssocMap.Map k v
+unsafeMergeMap x y = AssocMap.unsafeFromList (AssocMap.toList x <> AssocMap.toList y)
+
+hashInput :: TxInInfo -> BuiltinByteString
+hashInput (TxInInfo (TxOutRef (TxId hash) idx) _)
+  | idx < 256 = blake2b_256 (consByteString idx hash)
+  | otherwise = traceError "hashInput: Transaction output index must fit in an octet"
+
+errorIfFalse :: BuiltinString -> Bool -> BuiltinUnit
+errorIfFalse msg False = traceError msg
+errorIfFalse _ True    = BI.unitval
+
+hasCurrency :: CurrencySymbol -> Value -> Bool
+hasCurrency cs (Value val) = AssocMap.member cs val
--- a/plutus-ledger-api/src/PlutusLedgerApi/V3/MintValue.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V3/MintValue.hs
@@ -20,6 +20,7 @@ module PlutusLedgerApi.V3.MintValue
   ( MintValue (..) -- Constructor is exported for testing
   , emptyMintValue
   , mintValueToMap
+  , mintValueToValue
   , mintValueMinted
   , mintValueBurned
   )
@@ -58,7 +59,7 @@ Users should project 'MintValue' into 'Value' using 'mintValueMinted' or 'mintVa
 -}
 
 -- | A 'MintValue' represents assets that are minted and burned in a transaction.
-newtype MintValue = UnsafeMintValue (Map CurrencySymbol (Map TokenName Integer))
+newtype MintValue = UnsafeMintValue {unMintValue ::(Map CurrencySymbol (Map TokenName Integer))}
   deriving stock (Generic, Data, Haskell.Show)
   deriving anyclass (NFData)
   deriving newtype (ToData, FromData, UnsafeFromData)
@@ -98,6 +99,10 @@ emptyMintValue = UnsafeMintValue Map.empty
 mintValueToMap :: MintValue -> Map CurrencySymbol (Map TokenName Integer)
 mintValueToMap (UnsafeMintValue m) = m
 {-# INLINEABLE mintValueToMap #-}
+
+mintValueToValue :: MintValue -> Value
+mintValueToValue (UnsafeMintValue m) = Value m
+{-# INLINEABLE mintValueToValue #-}
 
 -- | Get the 'Value' minted by the 'MintValue'.
 mintValueMinted :: MintValue -> Value
--- a/plutus-benchmark/coop/README.md
+++ b/plutus-benchmark/coop/README.md
@@ -1,0 +1,10 @@
+# Cardano Open Oracle Protocol
+Cardano Open Oracle Protocol(COOP for short) is an opensource oracle implementation. More information can be found at [mlabs-haskell/cardano-open-oracle-protocol](https://github.com/mlabs-haskell/cardano-open-oracle-protocol/).
+
+## Scripts
+There are in total five scripts; validation requirements can be found on the protocol specification which is linked.
+- [Fs Validator](https://github.com/mlabs-haskell/cardano-open-oracle-protocol/blob/main/coop-docs/02-plutus-protocol.md#fs-validator)
+- [Cert Validator](https://github.com/mlabs-haskell/cardano-open-oracle-protocol/blob/main/coop-docs/02-plutus-protocol.md#cert-validator)
+- [Fs Policy](https://github.com/mlabs-haskell/cardano-open-oracle-protocol/blob/main/coop-docs/02-plutus-protocol.md#fs-policy)
+- [Cert Policy](https://github.com/mlabs-haskell/cardano-open-oracle-protocol/blob/main/coop-docs/02-plutus-protocol.md#cert-policy)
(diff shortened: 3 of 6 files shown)


plutus-core/executables/src/PlutusCore/Executable/AstIO.hs
```diff
@@ -160,9 +160,9 @@ serialiseUplcProgramFlat
     -> BSL.ByteString
 serialiseUplcProgramFlat =
     \case
-     Named         -> BSL.fromStrict . flat. UPLC.UnrestrictedProgram
-     DeBruijn      -> BSL.fromStrict . flat. UPLC.UnrestrictedProgram . toDeBruijnUPLC
-     NamedDeBruijn -> BSL.fromStrict . flat .UPLC.UnrestrictedProgram . toNamedDeBruijnUPLC
+     Named         -> BSL.fromStrict . flat . UPLC.UnrestrictedProgram
+     DeBruijn      -> BSL.fromStrict . flat . UPLC.UnrestrictedProgram . toDeBruijnUPLC
+     NamedDeBruijn -> BSL.fromStrict . flat . UPLC.UnrestrictedProgram . toNamedDeBruijnUPLC
```
SeungheonOh, line 165: I wasn't able to resist here

plutus-benchmark/coop/src/PlutusBenchmark/Coop/Scripts.hs
```diff
@@ -0,0 +1,296 @@
...
+fsMp :: CompiledCode (FsMpParams -> BuiltinData -> BuiltinData -> BuiltinUnit)
+fsMp = $$(compile [|| \p r sc -> fsMp' p (unsafeFromBuiltinData r) (unsafeFromBuiltinData sc) ||])
+
+authMp :: CompiledCode (AuthMpParams -> BuiltinData -> BuiltinData -> BuiltinUnit)
+authMp = $$(compile [|| \p r sc -> authMp' p (unsafeFromBuiltinData r) (unsafeFromBuiltinData sc) ||])
+
+certMp :: CompiledCode (CertMpParams -> BuiltinData -> BuiltinData -> BuiltinUnit)
+certMp = $$(compile [|| \p r sc -> certMp' p (unsafeFromBuiltinData r) (unsafeFromBuiltinData sc) ||])
+
+fsMp' :: FsMpParams -> FsMpRedeemer -> ScriptContext -> BuiltinUnit
```
SeungheonOh, line 50: I'm aware there's Data backed versions, which I found out later. Sadly, pattern synonym doesn't work with wildcards. 

We need to have discussion around this. I think we should just have arbitrary datatype to have configurable underlying representation, which allows other optimizations as well.

plutus-benchmark/coop/exe/Main.hs
```diff
@@ -0,0 +1,156 @@
+{-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell   #-}
```
Unisay: Unused pragma:
```suggestion
```

plutus-benchmark/coop/exe/Main.hs
```diff
@@ -0,0 +1,155 @@
...
+      , compiledCodeToHaskUnsafe
+          Scripts.authMp
+          (liftCodeDef authMpParams)
+          (liftCodeDefAsData (Redeemer $ toBuiltinData AuthMpBurn))
+          (liftCodeDefAsData $
+             runGen (genCorrectAuthMpBurningCtx authCs))
+      ]
+
+  _ <- traverse (uncurry createIfNotExists) (zip ((\i -> "coop-" <> show @Integer i) <$> [1..]) scripts)
+  pure ()
```
Unisay: (`traverse_` needs to be imported)

```suggestion
  traverse_ (uncurry createIfNotExists) (zip ((\i -> "coop-" <> show @Integer i) <$> [1..]) scripts)
```

plutus-benchmark/coop/exe/Main.hs
```diff
@@ -0,0 +1,155 @@
+{-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications  #-}
+
+{-
+This module itself won't run any benchmark on it's own. It will only generate `.flat` file, if missing,
+in `validation/data` directory for `validation` benchmark runner to run COOP scripts.
```
Unisay:
```suggestion
This module itself won't run any benchmark on it's own. It will only generate
`.flat` file, if missing, in `validation/data` directory for `validation`
 benchmark runner to run COOP scripts.
```

plutus-benchmark/coop/exe/Main.hs
```diff
@@ -0,0 +1,155 @@
...
+    fullName = p </> name <.> "flat"
+    eraseName (UPLC.Program ann ver t) =
+      () <$ UPLC.Program ann ver (UPLC.termMapNames UPLC.unNameDeBruijn t)
+  termAsBS <-
+    case term of
+      SerializedCode bs _ _     ->
+        let
+          parsed =
+            UPLC.unUnrestrictedProgram
+            <$> (unflat @(UPLC.UnrestrictedProgram UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun SrcSpans) $ bs)
```
Unisay:
```suggestion
          parsed =
            UPLC.unUnrestrictedProgram
              <$> unflat
                @( UPLC.UnrestrictedProgram
                     UPLC.NamedDeBruijn
                     UPLC.DefaultUni
                     UPLC.DefaultFun
                     SrcSpans
                 )
                bs
```

plutus-benchmark/coop/exe/Main.hs
```diff
@@ -0,0 +1,155 @@
...
+
+liftCodeDefAsData :: ToData a => a -> CompiledCode BuiltinData
+liftCodeDefAsData = liftCodeDef . toBuiltinData
+
+main :: IO ()
+main = do
+  let
+    seed = 1
+    runGen :: Gen a -> a
+    runGen g = unGen g (mkQCGen seed) 1
```
Unisay:
```suggestion
    runGen :: Gen a -> a
    runGen g =
      let seed = 1
          size = 1
       in unGen g (mkQCGen seed) size
```

plutus-benchmark/coop/src/PlutusBenchmark/Coop/Utils.hs
```diff
@@ -7,12 +7,17 @@ module PlutusBenchmark.Coop.Utils where
 import PlutusTx.Prelude
 import Prelude ()
 
-import PlutusLedgerApi.V1.Value
-import PlutusLedgerApi.V2
-import PlutusTx.AssocMap
+import PlutusLedgerApi.V1.Value (Value (Value), flattenValue, valueOf, withCurrencySymbol)
+import PlutusLedgerApi.V2 (CurrencySymbol, Datum (Datum), DatumHash,
```
Unisay, line 11: Why V2 not V3 ?

Unisay (commented):
Thanks for the explicit imports!

I was expecting all types from the Plutus Ledger API to be imported from the `PlutusLedgerApi.V3` namespace (or the `PlutusLedgerApi.Data.V3`). Could you verify this is the case, please?

plutus-benchmark/coop/src/PlutusBenchmark/Coop/Scripts.hs
Unisay: Please remember to bring up this topic at the next technical discussion meeting!

SeungheonOh:
The original scripts were V2 scripts; I replicated original script exactly for both ease of testing and following code

Unisay:
> The original scripts were V2 scripts; 

It could be that we've improved something in V3 so I see no point in bringing something potentially sub-optimal.

> I replicated original script exactly for both ease of testing and following code

Sorry, I don't understand how using V2 makes it easier to test or follow 🤷🏼‍♂️

plutus-ledger-api/src/PlutusLedgerApi/V3/MintValue.hs
```diff
@@ -99,6 +100,10 @@ mintValueToMap :: MintValue -> Map CurrencySymbol (Map TokenName Integer)
 mintValueToMap (UnsafeMintValue m) = m
 {-# INLINEABLE mintValueToMap #-}
 
+mintValueToValue :: MintValue -> Value
+mintValueToValue (UnsafeMintValue m) = Value m
+{-# INLINEABLE mintValueToValue #-}
```
Unisay: Oh noes!

This defeats the whole purpose: we want users to explicitly decide if the `MintValue` represent a burned `Value` or minted `Value`. They have different invariants internally wrt negative values - this function is unsafe!

plutus-ledger-api/src/PlutusLedgerApi/V3/MintValue.hs
```diff
@@ -20,6 +20,7 @@ module PlutusLedgerApi.V3.MintValue
   ( MintValue (..) -- Constructor is exported for testing
   , emptyMintValue
   , mintValueToMap
+  , mintValueToValue
```
Unisay:
```suggestion
```

plutus-ledger-api/src/PlutusLedgerApi/V3/MintValue.hs
SeungheonOh: How about having it renamed to `unsafeMintValueToValue`? I think this function is really helpful because we already have a bunch of utilities for `Value` and sometimes we want to use those on `MintValue` as well

SeungheonOh:
V2 makes comparison more effective in this case because of two reasons mainly:
- Script context deconstruction cost can be differ between different plutus ledger api versions. For example, deconstructing script context in V1 context is often always more expensive than doing it for V2 even with data because they have a bunch of types like `CurrencySymbol` or `PublicKeyHash` wrapped in `Constr 0`. So when comparing costs between two implementations, it's better to keep ledger api version same for a pair comparison. 
- Secondly, It's much easier more so in this case because we have vendor provided generators for script contexts that will generate V2 script context. Having two script that shares same input ledger api format would be simple to do.

Unisay:
> * Script context deconstruction cost can be differ between different plutus ledger api versions. For example, deconstructing script context in V1 context is often always more expensive than doing it for V2 even with data because they have a bunch of types like `CurrencySymbol` or `PublicKeyHash` wrapped in `Constr 0`. So when comparing costs between two implementations, it's better to keep ledger api version same for a pair comparison.

What is it you're comparing it against?

SeungheonOh:
The plutarch implementation: https://github.com/mlabs-haskell/cardano-open-oracle-protocol

SeungheonOh:
@zliu41 Do we want this comparison against plutarch to be inside of `plutus-benchmark`. If not, I'm just going to run it separately and report numbers here. Since Plutarch implementation won't change, we just need to run the numbers for plutarch scripts once

plutus-ledger-api/src/PlutusLedgerApi/V3/MintValue.hs
SeungheonOh: Oh wait, I haven't even used this. silly me. I'll remove

Unisay:
> @zliu41 Do we want this comparison against plutarch to be inside of `plutus-benchmark`. If not, I'm just going to run it separately and report numbers here. Since Plutarch implementation won't change, we just need to run the numbers for plutarch scripts once

I have a task to collect and write down such requirements, its in progress. For now I'd say it is enough to do a minimal thing, as we haven't agreed yet on what the maximal thing is going to be.

plutus-ledger-api/src/PlutusLedgerApi/V3/MintValue.hs
Unisay: Ok, could be. Please don't export it in the `PlutusLedgerApi.V3` and add a haddock explaining why this function is unsafe, what should be used instead. Here is the context: 
https://github.com/IntersectMBO/plutus/issues/6445

zliu41 (commented):
A high level comment: we need golden tests that measure the execution units of these scripts in CI, similar to [what we do for Marlowe scripts](https://github.com/IntersectMBO/plutus/blob/5bdbb8df36e2cf29b5e7d8a8526b1ae9c9f658b5/plutus-benchmark/marlowe/test/Spec.hs). This is the main purpose of porting these scripts, which is to allow us to observe the effectiveness of compiler changes. I think you've done most of the work, but the tests aren't there yet.

It would also be very helpful to compare the sizes and costs of the Plutarch and Plinth versions. You don't need to open a PR or commit any code for this - just do some offline analysis and document any observations about optimization opportunities for the Plinth compiler. The cross-language benchmark project will give us more scripts to compare and analyze, but we still need to analyze them individually, so it's good to start with this one.

And to do that, you'll likely need to adapt the script to use Data encoding more (at least for the ledger types) - otherwise the performance will likely be very bad and the comparison won't be meaningful.

SeungheonOh:
@zliu41 

Here are results from plutarch implementation: 

```
authMpBurnApplied.flat
(con unit ())
CPU budget:    70_105_072
Memory budget: 196_783

authMpMintApplied.flat
(con unit ())
CPU budget:    430_521_147
Memory budget: 1_184_919

certMpBurnApplied.flat
An error has occurred:
Caused by: error
CPU budget:    1_229_697_221
Memory budget: 3_595_798

certMpMintApplied.flat
(con unit ())
CPU budget:    510_411_469
Memory budget: 1_457_063

fsMpBurnApplied.flat
An error has occurred:
Caused by: error
CPU budget:    474_651_914
Memory budget: 1_516_415

fsMpMintApplied.flat
An error has occurred:
Caused by: error
CPU budget:    768_545_412
Memory budget: 2_404_394
```

Some scripts fails, even though script context used here is provided by the plutarch implementation. I can research more if necessary, but I prefer not to since code there is **_archaic_**. In terms of result, Plinth implementation is 30~50% slower, but this is with current plinth implementation using slower plutus ledger api types that will deconstruct all fields at once.

Issue #7157, opened 2025-06-18
Rendering of Unicode in textual output


kwxm:
It appears that non-ascii Unicode characters in textual renderings of UPLC programs (for example, as produced by `uplc` and in the expected results of conformance tests) are now represented by escape sequences.  For example, the conformance tests [contain the following input](https://github.com/IntersectMBO/plutus/blob/master/plutus-conformance/test-cases/uplc/evaluation/builtin/constant/string/string-05/string-05.uplc):
```
(program 1.0.0
    (con string "x ∈ ℝ ⇒ x² ≥ 0; z ∈ ℂ\\ℝ ⇒ z² ∉ {x ∈ ℝ: x ≥ 0}.")
)
```
The output of that program used to be readable but is now
```
(program
  1.0.0
  (con
    string
    "x \8712 \8477 \8658 x\178 \8805 0; z \8712 \8450\\\8477 \8658 z\178 \8713 {x \8712 \8477: x \8805 0}."
  )
)
```
Do we want to change it back?

PR #7158, opened 2025-06-18
Kwxm/conformance/update test results

kwxm:
This updates the conformance README to clarify the fact that the results of conformance tests only have to agree with our expected results up to α-equivalence.  Thanks to @angerman for pointing out the potential for confusion.

The outputs in the `.uplc.expected` files were also a bit confusing.  Some of them had plain names, some had names involving underscores (perhaps indicating de Bruijn indices?) and some had uniques.  I've updated them so that all names include uniques, but I'm not sure if that's better than having plain names.

I got the updated `.uplc.expected` files by deleting them all and re-running the tests.  This revealed that the way we print Unicode characters has changed.  I may come back and change that in a later PR (see #7157).


plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/bls12_381_G1_uncompress/on-curve-bit3-set/on-curve-bit3-set.uplc.expected
```diff
@@ -1,5 +1,3 @@
--- This value was obtained by hashing 0x0102030405 to G2 and has the compression
--- bit set, so uncompression should succeed.
```
effectfully, line 2: Keep the comment?

effectfully (approved):

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/bls12_381_G1_uncompress/on-curve-bit3-set/on-curve-bit3-set.uplc.expected
kwxm: Well-spotted, but I think we can lose the comment.  I think that's a stray comment from somewhere else (it mentions G2 but these tests are about G1.  I'm not sure how it got in there in the first place.

kwxm:
Bypassing CI checks because of Hydra problems.

kwxm:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1636

PR #7159, opened 2025-06-18
[Refac] Make UPLC and TPLC tests more uniform

effectfully:
Polished things a little, added some functionality to the `testlib` of UPLC that we already had for TPLC, added a UPLC test.


plutus-core/plutus-core/test/Spec.hs
```diff
@@ -65,83 +65,10 @@ main = do
...
-compareName = (==) `on` _nameText
-
-compareTyName :: TyName -> TyName -> Bool
-compareTyName (TyName n) (TyName n') = compareName n n'
-
-compareTerm ::
-  (GEq uni, Closed uni, uni `Everywhere` Eq, Eq fun, Eq a) =>
-  Term TyName Name uni fun a ->
-  Term TyName Name uni fun a ->
-  Bool
```
effectfully, line 78: Now that we print uniques, we don't need any of that garbage.

plutus-core/plutus-core/test/Parser/Spec.hs
```diff
@@ -20,7 +20,7 @@ import Test.Tasty.Hedgehog
 propTermSrcSpan :: Property
 propTermSrcSpan = property $ do
     term <- _progTerm <$>
-        forAllWith display (runAstGen $ discardIfAnyConstant (not . isSerialisable) genProgram)
+        forAllWith display (runAstGen $ regenConstantsUntil isSerialisable =<< genProgram)
```
effectfully, line 23: It's much faster to regenerate a specific unsupported constant than to discard the whole term and regenerate a new one from scratch.

plutus-core/plutus-ir/src/PlutusIR/Transform/Substitute.hs
effectfully, line 1: Moved this stuff to a `Subst` file just like in UPLC and TPLC.

plutus-core/testlib/PlutusCore/Generators/Hedgehog/AST.hs
```diff
@@ -17,18 +18,18 @@ module PlutusCore.Generators.Hedgehog.AST
...
     , genTerm
     , genProgram
+    , genNameMangler
     , mangleNames
     ) where
 
 import PlutusPrelude
 
 import PlutusCore
-import PlutusCore.Core.Plated (termConstantsDeep)
```
Unisay, line 26: This `termConstantsDeep` doesn't seem to be used anywhere anymore.

plutus-core/testlib/PlutusCore/Generators/Hedgehog/AST.hs
```diff
@@ -60,12 +61,23 @@ runAstGen a = do
     names <- genNames
     Gen.fromGenT $ hoist (return . flip runReader names) a
 
-discardIfAnyConstant
+regenConstantUntil
```
Unisay, line 64: I'd appreciate a haddock explaining how this function is useful and what is the meaning of its parameters.

plutus-core/testlib/PlutusCore/Generators/Hedgehog/AST.hs
effectfully: It's OK, it's a library function, maybe someone will need it for something eventually. It's not expensive to maintain.

plutus-core/testlib/PlutusCore/Generators/Hedgehog/AST.hs
effectfully: Added

```
-- | Replace all constants in a program failing the given predicate with newly generated ones that
-- are guaranteed to satisfy the predicate (if such constants cannot be generated, this generator
-- will fail).
```

kwxm:
Hydra timing out - merging manually.

PR #7161, opened 2025-06-19
Implement force-case-delay optimisation

ana-pantilie:
Needs https://github.com/IntersectMBO/plutus/pull/7163

Fixes https://github.com/IntersectMBO/plutus-private/issues/1557


plutus-benchmark/bitwise/test/9.6/8 queens.uplc.golden
```diff
@@ -5,142 +5,132 @@
...
+                                                        , (constr 0 []) ]))
+                                                  (delay
+                                                     ((\newRow ->
+                                                         (\cse ->
+                                                            (\newRight ->
+                                                               (\newLeft ->
+                                                                  (\newDown ->
+                                                                     (\newControl ->
+                                                                        case
+                                                                          (s
```
ana-pantilie: It seems like `cse` didn't fire here anymore, because the `(s s 0 newRow ...)` term is inlined. I don't think this can be controlled, since we know `cse` to be finicky. But I would like someone to confirm.

plutus-benchmark/bitwise/test/9.6/8 queens.eval.golden
```diff
@@ -1,6 +1,6 @@
-CPU:         381_725_318_016
-Memory:        1_403_857_547
-Size:                    380
+CPU:         795_331_345_376
+Memory:        2_751_105_493
+Size:                    369
```
effectfully: This is very bad.

plutus-benchmark/bitwise/test/9.6/8 queens.uplc.golden
effectfully: Sounds like a plausible explanation to me.

PR #7162, opened 2025-06-19
Bug: `Cannot construct a value of type: PlutusTx.Builtins.Internal.BuiltinUnit`

Unisay:
Closes https://github.com/IntersectMBO/plutus-private/issues/1626

### DIFF
--- a/plutus-tx-plugin/changelog.d/20260205_095326_yuriy.lazaryev_Cannot_construct_a_value_of_type.md
+++ b/plutus-tx-plugin/changelog.d/20260205_095326_yuriy.lazaryev_Cannot_construct_a_value_of_type.md
@@ -1,0 +1,4 @@
+
+### Changed
+
+- Improved "Cannot construct a value of type" error messages for opaque builtin types to explain stage violations and suggest moving where-bindings to top-level


doc/notes/plutus-tx-plugin/NO_SIMPLE_PASS.md
```diff
@@ -0,0 +1,135 @@
...
+```
+Error: Reference to a name which is not a local, a builtin, or an external INLINABLE function: Variable Plugin.Coverage.Spec.fun
+OtherCon []
+```
+- **Root cause**: GHC #16615 - local bindings lack unfoldings without simplifier
+- **Pattern**: Functions marked `{-# INLINEABLE #-}` but defined after their use site
+- **Solutions**:
+  - Move functions to top-level
+  - Generate minimal unfoldings during desugaring (GHC enhancement needed)
+  - Create lightweight unfolding pass (lighter than full simplifier)
```
SeungheonOh: is this possible without making changes to GHC itself? From what I can see(which can be very wrong), we can only really change the plugin pass and the plugin pass doesn't have access to other stages of compilation.

This would be very easy if we have whole compilation pipeline under control.

doc/notes/plutus-tx-plugin/NO_SIMPLE_PASS.md
```diff
@@ -0,0 +1,135 @@
...
+- **BuiltinUnit test cases**: Now produce clear "No unfolding" errors instead of confusing builtin constructor errors
+- **This is the desired behavior** - code that violates PlutusTx compilation rules should fail with meaningful messages
+
+#### 2. Missing Unfoldings for INLINEABLE Functions (🔧 Potentially Fixable)
+```
+Error: Reference to a name which is not a local, a builtin, or an external INLINABLE function: Variable Plugin.Coverage.Spec.fun
+OtherCon []
+```
+- **Root cause**: GHC #16615 - local bindings lack unfoldings without simplifier
+- **Pattern**: Functions marked `{-# INLINEABLE #-}` but defined after their use site
```
SeungheonOh: It's probably also the case for function marked inlineable but defined on the other module.

SeungheonOh (commented):
I prefer adding special handling for providing better error message instead of disabling simplpass entirely. It seems writing a new simplifier(or rather, unfolder) would require some changes to GHC itself which would be undesirable.

doc/docusaurus/docs/troubleshooting.md
```diff
@@ -49,7 +49,57 @@ case (x :: Maybe Integer) of Just y | y PlutusTx.== 42 -> ...
...
+     where
+       ~validator = unitval
+   ```
+
+3. **Define the binding inside the quotation:**
+   ```haskell
+   code = $$(compile [|| let validator = unitval in validator ||])
+   ```
+
+Variables inside `compile` quotations must be either top-level or bound inside the quotation itself to avoid stage violations.
```
zliu41: But this contradicts with workaround 2, where the variable is still outside the quotation?

zliu41 (approved):

PR #7166, opened 2025-06-25
Benchmark: Linear Vesting

Unisay:
Closes https://github.com/IntersectMBO/plutus-private/issues/1617
```
Evaluation was SUCCESSFUL, result is:
  ()

Execution budget spent:
  CPU 101,442,633
  MEM 474,786

Evaluation traces:
  1. Parsing ScriptContext...
  2. Parsed ScriptContext
  3. Parsed Redeemer
  4. Full unlock requested
  5. Validation completed
```

After switching to the `Data`-backed `ScriptContext`:

```
Execution budget spent:
  CPU 39,768,848
  MEM 155,049  
```

### DIFF
--- a/plutus-benchmark/linear-vesting/src/LinearVesting.hs
+++ b/plutus-benchmark/linear-vesting/src/LinearVesting.hs
@@ -1,0 +1,246 @@
+{-# LANGUAGE BangPatterns          #-}
+{-# LANGUAGE BlockArguments        #-}
+{-# LANGUAGE DataKinds             #-}
+{-# LANGUAGE LambdaCase            #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE MultiWayIf            #-}
+{-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE NoImplicitPrelude     #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE PatternSynonyms       #-}
+{-# LANGUAGE Strict                #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE ViewPatterns          #-}
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+{-# OPTIONS_GHC -fno-full-laziness #-}
+{-# OPTIONS_GHC -fno-ignore-interface-pragmas #-}
+{-# OPTIONS_GHC -fno-omit-interface-pragmas #-}
+{-# OPTIONS_GHC -fno-spec-constr #-}
+{-# OPTIONS_GHC -fno-specialise #-}
+{-# OPTIONS_GHC -fno-strictness #-}
+{-# OPTIONS_GHC -fno-unbox-small-strict-fields #-}
+{-# OPTIONS_GHC -fno-unbox-strict-fields #-}
+{-# OPTIONS_GHC -fplugin PlutusTx.Plugin #-}
+{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:conservative-optimisation #-}
+{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:no-remove-trace #-}
+{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:preserve-logging #-}
+
+module LinearVesting where
+
+import PlutusTx
+import PlutusTx.Prelude
+import Prelude qualified as Haskell
+
+import PlutusLedgerApi.Data.V3
+import PlutusLedgerApi.V1.Data.Value (AssetClass, assetClass, assetClassValueOf)
+import PlutusLedgerApi.V3.Data.Contexts (txSignedBy)
+import PlutusTx.Data.AssocMap qualified as Map
+import PlutusTx.Data.List (List)
+import PlutusTx.Data.List qualified as List
+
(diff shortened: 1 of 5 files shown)


SeungheonOh:
For the reference, this script is from

https://github.com/Anastasia-Labs/linear-vesting/blob/main/src/LinearVesting.hs

plutus-benchmark/linear-vesting/src/LinearVesting.hs
```diff
@@ -0,0 +1,243 @@
...
+    Just ctx -> check $ typedValidator $ trace "Parsed ScriptContext" ctx
+
+validatorCode :: CompiledCode (BuiltinData -> BuiltinUnit)
+validatorCode = $$(compile [||untypedValidator||])
+
+validatorCodeFullyApplied :: CompiledCode BuiltinUnit
+validatorCodeFullyApplied =
+  validatorCode `unsafeApplyCode` liftCodeDef (toBuiltinData testScriptContext)
+
+testScriptContext :: ScriptContext
```
SeungheonOh: Let's have this as an another file.

plutus-benchmark/linear-vesting/src/LinearVesting.hs
```diff
@@ -0,0 +1,243 @@
...
+      addressCredential (beneficiary vestingDatum)
+   in
+    if
+      | not (txSignedBy txInfo beneficiaryHash) ->
+          traceError "Missing beneficiary signature"
+      | firstUnlockPossibleAfter vestingDatum >= currentTimeApproximation ->
+          traceError "Unlock not permitted until firstUnlockPossibleAfter time"
+      | newRemainingQty <= 0 ->
+          traceError "Zero remaining assets not allowed"
+      | newRemainingQty > oldRemainingQty ->
```
SeungheonOh: `newRemainingQty >= oldRemainingQty`?

https://github.com/Anastasia-Labs/linear-vesting/blob/383b388c77bdf6d80242fdbe2a6b0167b42f4b2c/src/LinearVesting.hs#L166

It doesn't really matter though

plutus-benchmark/linear-vesting/src/LinearVesting.hs
```diff
@@ -0,0 +1,243 @@
...
+    futureInstallments :: Integer =
+      vestingTimeRemaining `divCeil` timeBetweenTwoInstallments
+    expectedRemainingQty :: Integer =
+      (futureInstallments * totalVestingQty vestingDatum)
+        `divCeil` totalInstallments vestingDatum
+    PubKeyCredential beneficiaryHash =
+      addressCredential (beneficiary vestingDatum)
+   in
+    if
+      | not (txSignedBy txInfo beneficiaryHash) ->
```
SeungheonOh: Generally, using `not` is not preferable. I'm not sure our optimizer or GHC deal with this like

```hs
if not predicate then A else B
```
to
```hs
if predicate then B then A
```

SeungheonOh (commented):
Script looks good. 

About running tests, do we want to pursue some unified format? For me, I made it so that script will be exported and save to `validation/data` so that it can be run by the existing "test runner".

plutus-benchmark/linear-vesting/src/LinearVesting.hs
Unisay: Moving it to another module requires domain types (`VestingDatum`, `VestingRedeemer`) to be moved out too. 
I was thinking that such break-down to smaller modules is an overkill for the linear vesting.
What's your reasoning for moving it out? What are you trying to optimize for? Will a comment section that separates testing code from the contract code be enough?

plutus-benchmark/linear-vesting/src/LinearVesting.hs
Unisay: Good point, lets see:

```haskell
  $$(compile [||\x -> if PlutusTx.not x then 1 else 2||])
    `unsafeApplyCode` liftCodeDef True
```
compiles to this UPLC:

```uplc
(program 1.1.0 ((\x -> case x [2, 1]) (constr 0 [])))
```
which evaluates to

```uplc
(con integer 2)
```

So `not` is optimized away, no problem.

Unisay:
> About running tests, do we want to pursue some unified format? For me, I made it so that script will be exported and save to `validation/data` so that it can be run by the existing "test runner".

We sure do want to unify the way we exercise the validators! I am planning to have a meeting round to collect such ideas/requirements before we implement something. The testing functionality in this PR is provisional.

plutus-benchmark/linear-vesting/src/LinearVesting.hs
SeungheonOh: Domain types can stay in `LinearVesting.hs` and the test cases file can import the domain types and the script itself. 

My argument is if we have it on same file, it will be more difficult to differentiate changes made in test cases and/or that in the script itself in the future. Basically for future tracking and review.

If you see this as something we add once here and rarely make changes or additions, I'm fine with having everything in one file.

Unisay: Actually, you convinced me that such validator/test code split is possible the other way around without much hassle;

I agree that it makes it easier to work with such code and usually I am pro-split. It just didn't feel like the code is big enough to exhibit the problem yet, but why not. I will separate it.

Unisay: Done.

PR #7167, opened 2025-06-25
Release 1.48.0.0

zeme-wana:
Release 1.48.0.0

### DIFF
--- a/plutus-executables/plutus-executables.cabal
+++ b/plutus-executables/plutus-executables.cabal
@@ -1,6 +1,6 @@
 cabal-version:   3.0
 name:            plutus-executables
-version:         1.47.0.0
+version:         1.48.0.0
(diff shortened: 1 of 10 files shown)


plutus-executables/plutus-executables.cabal
```diff
@@ -109,7 +109,7 @@ executable uplc
     , haskeline
     , mtl
     , optparse-applicative
-    , plutus-core                      ^>=1.47
+    , plutus-core                      ^>=1.48
```
ana-pantilie, line 112: I can't comment on the right line, so this is for line 115: should we start enforcing bounds on `plutus-metatheory` as well now that we release it to chap?

zeme-wana: Yes, good catch.

PR #7169, opened 2025-06-25
[Names] Remove 'Monoid' instances

effectfully:
This shows just how annoying it is to address https://github.com/IntersectMBO/plutus/issues/6138. Dropping `mempty` is fine, but now half the optimization are broken in non-trivial ways. I'm not sure what to do about it, maybe it's worth fixing all of that.


effectfully:
This is pretty useless for a self-review, so no point in pursuing it.

PR #7171, opened 2025-06-30
Tool to dump cost model parameters in order expected by ledger

kwxm:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1650.

The ledger and the procol parameters view the cost model parameters as a list of integers with no other information.  This makes them quite difficult to deal with (in particular to check that they're in the right order).  This PR adds a simple tool to dump the cost model parameters in the expected order, in various formats including the JSON format used by `cardano-cli`.  See the attached files for examples of the output.

[cost-model-parameters-untagged.txt](https://github.com/user-attachments/files/20984174/cost-model-parameters-untagged.txt)
[cost-model-parameters-tagged.txt](https://github.com/user-attachments/files/20984176/cost-model-parameters-tagged.txt)
[cost-model-parameters-json.txt](https://github.com/user-attachments/files/20984178/cost-model-parameters-json.txt)

The `uplc` executable has a `dump-cost-model` command, but that doesn't know anything about `plutus-ledger-api` and just produces a list ordered by parameter name,

### DIFF
--- a/plutus-ledger-api/exe/dump-cost-model-parameters/Parsers.hs
+++ b/plutus-ledger-api/exe/dump-cost-model-parameters/Parsers.hs
@@ -53,5 +53,5 @@ parseDumpOptions =
            progDesc ("Print the current (and possibly undeployed) cost model parameters "
                      ++ " in the plutus repository in the order used in the protocol parameters.\n"
                      ++ "The purpose of this tool is to help with the deployment and verification "
-                     ++ "of updated cost model parameters. It MUST be kept up to date with the "
-                     ++ "EvaluationContext files in plututus-ledger-api."))
+                     ++ "of updated cost model parameters: it MUST be kept up to date with the "
+                     ++ "`mkEvaluationContext` functions in plututus-ledger-api."))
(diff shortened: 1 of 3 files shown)


plutus-ledger-api/exe/dump-cost-model-parameters/Parsers.hs
```diff
@@ -0,0 +1,57 @@
...
+
+parseVersion :: ReadM WhichLL
+parseVersion = eitherReader $ \case
+  "1" -> Right $ One PlutusV1
+  "2" -> Right $ One PlutusV2
+  "3" -> Right $ One PlutusV3
+  s -> Left $ "Unknown ledger language version: " ++ s
+
+whichll :: Parser WhichLL
+whichll =
```
Unisay, line 22: The indentation is inconsistent across newly added files,
do you mind applying Fourmolu to all the Haskell code in this PR?

kwxm: > do you mind applying Fourmolu to all the Haskell code in this PR?

I've done that, but then stylish-haskell changed some of it back when I committed the changes.  Specifically, it lined up the `->`s in case expressions and also the `-#}`s at the end of the pragmas at the start.  I like both of those better than what Fourmolu does, but you might object!

kwxm: I'm not sure but I think that also Fourmolu aligned `let ... in ...` like this
```
  let xxx
   in yyy
```
and stylish-haskell changed it back to
```
  let xxx
  in yyy
```
The first one looks a bit strange to me but I see why it's done it.  It's a pity that `let` and `in` don't have the same number of letters...

PR #7174, opened 2025-07-01
Re-run the Agda benchmarks.

kwxm:
We have some benchmarks for the Agda evaluator that we haven't run for a long time.  This is just to re-run them to compare the results with the Haskell CEK machine.


kwxm:
OK, I found out what I wanted to know: the Agda evaluator takes about 3-10 times as long as the Haskell one.  

Also, `nofib-agda-cek` is failing.  It seems that `plc-agda` can't handle the compiled files for some reason (and experiments with `nofib-exe` confirm this).  I'll look into that.

PR #7177, opened 2025-07-02
[Benchmark] Pull 'force' out of 'bench'

effectfully:
I was investigating an issue with #7012 not being faster than the baseline despite clearly doing strictly less work and came to the conclusion that the only reasonable explanation is that something is wrong with the benchmarks. So here's an investigation.

### DIFF
--- a/plutus-benchmark/agda-common/PlutusBenchmark/Agda/Common.hs
+++ b/plutus-benchmark/agda-common/PlutusBenchmark/Agda/Common.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE BangPatterns #-}
+
 module PlutusBenchmark.Agda.Common
   ( benchTermAgdaCek
   , benchProgramAgdaCek
@@ -10,7 +12,8 @@ import UntypedPlutusCore qualified as UPLC
 
 import MAlonzo.Code.Evaluator.Term (runUAgda)
 
-import Criterion.Main (Benchmarkable, nf)
+import Control.DeepSeq (force)
+import Criterion.Main (Benchmark, bench, whnf)
 
 -- This code is in its own file so that we only build the metatheory when we really need it.
 
@@ -19,13 +22,15 @@ type Program = UPLC.Program PLC.NamedDeBruijn DefaultUni DefaultFun ()
 
 ---------------- Run a term or program using the plutus-metatheory CEK evaluator ----------------
 
-benchTermAgdaCek :: Term -> Benchmarkable
-benchTermAgdaCek term =
-    nf unsafeRunAgdaCek $! term
+benchTermAgdaCek :: String -> Term -> Benchmark
+benchTermAgdaCek name term =
+    let !term' = force term
+    in bench name $ whnf unsafeRunAgdaCek term'
 
-benchProgramAgdaCek :: Program -> Benchmarkable
-benchProgramAgdaCek (UPLC.Program _ _ term) =
-    nf unsafeRunAgdaCek $! term
+benchProgramAgdaCek :: String -> Program -> Benchmark
+benchProgramAgdaCek name (UPLC.Program _ _ term) =
+    let !term' = force term
+    in bench name $ whnf unsafeRunAgdaCek term'
 
 unsafeRunAgdaCek :: Term -> PLC.EvaluationResult Term
 unsafeRunAgdaCek =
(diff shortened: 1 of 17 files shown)


plutus-benchmark/agda-common/PlutusBenchmark/Agda/Common.hs
```diff
@@ -19,13 +22,15 @@ type Program = UPLC.Program PLC.NamedDeBruijn DefaultUni DefaultFun ()
 
 ---------------- Run a term or program using the plutus-metatheory CEK evaluator ----------------
 
-benchTermAgdaCek :: Term -> Benchmarkable
-benchTermAgdaCek term =
-    nf unsafeRunAgdaCek $! term
+benchTermAgdaCek :: String -> Term -> Benchmark
+benchTermAgdaCek name term =
+    let !term' = force term
+    in bench name $ whnf unsafeRunAgdaCek term'
```
effectfully, line 28: Added `force` and changed `nf` to `whnf` for consistency.

plutus-benchmark/common/PlutusBenchmark/Common.hs
```diff
@@ -134,14 +134,14 @@ evaluateCekForBench
     -> ()
 evaluateCekForBench evalCtx = either (error . show) (\_ -> ()) . evaluateCekLikeInProd evalCtx
 
-benchTermCek :: LedgerApi.EvaluationContext -> Term -> Benchmarkable
-benchTermCek evalCtx term =
+benchTermCek :: String -> LedgerApi.EvaluationContext -> Term -> Benchmark
+benchTermCek name evalCtx term =
     let !term' = force term
-    in whnf (evaluateCekForBench evalCtx) term'
+    in bench name $ whnf (evaluateCekForBench evalCtx) term'
```
effectfully, line 140: The gist of the PR.

plutus-benchmark/agda-common/PlutusBenchmark/Agda/Common.hs
```diff
@@ -19,13 +22,15 @@ type Program = UPLC.Program PLC.NamedDeBruijn DefaultUni DefaultFun ()
...
 
-benchProgramAgdaCek :: Program -> Benchmarkable
-benchProgramAgdaCek (UPLC.Program _ _ term) =
-    nf unsafeRunAgdaCek $! term
+benchProgramAgdaCek :: String -> Program -> Benchmark
+benchProgramAgdaCek name (UPLC.Program _ _ term) =
+    let !term' = force term
+    in bench name $ whnf unsafeRunAgdaCek term'
```
effectfully, line 33: Should just call `benchTermAgdaCek` instead. Will fix.

plutus-benchmark/nofib/bench/BenchHaskell.hs
```diff
@@ -12,17 +12,17 @@ import PlutusBenchmark.NoFib.Knights qualified as Knights
 import PlutusBenchmark.NoFib.Prime qualified as Prime
 import PlutusBenchmark.NoFib.Queens qualified as Queens
 
-benchClausify :: Clausify.StaticFormula -> Benchmarkable
-benchClausify f = nf Clausify.runClausify f
+benchClausify :: String -> Clausify.StaticFormula -> Benchmark
+benchClausify name f = bench name $ whnf Clausify.runClausify f
```
effectfully, line 16: Changed `nf` to `whnf` for consistency. Note that this is just the Haskell version of the benchmarks, the CEK one is unaffected by this change.

effectfully:
That somehow broke the `validation` benchmarks... I'm seeing this:

```
Benchmark validation: RUNNING...
validation: Criterion atttempted to retrieve a non-existent environment!
	Perhaps you forgot to use lazy pattern matching in a function which
	constructs benchmarks from an environment?
	(see the documentation for `env` for details)
```

and I keep seeing it even if I change

```haskell
benchTermCek :: String -> LedgerApi.EvaluationContext -> Term -> Benchmark
benchTermCek name evalCtx term =
    let !term' = force term
    in bench name $ whnf (evaluateCekForBench evalCtx) term'
```

to

```haskell

benchTermCek :: String -> LedgerApi.EvaluationContext -> Term -> Benchmark
benchTermCek name evalCtx term =
    bench name $ let !term' = force term in whnf (evaluateCekForBench evalCtx) term'
```

which really should restore the old behavior?

For the love of God, what can possibly be wrong here?

effectfully:
wtf...

image

effectfully:
This turned out to be way harder than I expected. Unfortunately, it seems to be above my performance grade ("meeting expectations") and I do not feel like I'm capable of taking on such a complex task. I believe a person with a better performance grade is warranted here.

Closing this PR, since I'm sure a person with a better performance grade will want to run their own investigation and write superior code, matching their elevated effort level.

PR #7178, opened 2025-07-02
Call trace for failing evaluation via new emitter mode

### LINKED ISSUE #7146: Add bind tracing mode

Plutarch and Aiken have `TracingAndBinds` compilation mode which will print all the binds(function calls) when validator fails. This is useful for determining which part of validator in particular cause the failure. 

For Plutarch, this feature is implemented using Haskell's `HasCallStack` to track all function calls. For Plinth, we can likely generate bindings by having an extra plugin pass. Note, this pass will need to be before `simplPass` since `simplPass` will inline many function.

_No response_

SeungheonOh:
closes #7146.

This replaces https://github.com/IntersectMBO/plutus/pull/7148. Unlike my original PR which made the plugin generate callstack message, this PR uses already existing `profile-all` option and make no addition to plugin. Since `profile-all` option gives function entrance and exit, it is possible to just have another EmitterMode that can make call trace from the trace messages. 

I decided added new machinery just for callstack into plugin while we have `profile-all` which does something similar is unfavorable. This PR is much simpler and doesn't add maintenance burden to the plugin. 

We can do something more on the emitter: like putting all call traces together. I kept it simple but let me know if there something that can be improved on emitter

### DIFF
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
@@ -669,7 +669,7 @@ maybeProfileRhs ghcVar var t = do
       case getVarSourceSpan ghcVar of
         -- When module is not compiled and GHC is using cached build from previous build, it will
         -- lack source span. There's nothing much we can do about this here since this is GHC
-        -- behavior
+        -- behavior. Issue #7203
         Nothing  -> nameStr
         Just src -> nameStr <> " (" <> show (src ^. srcSpanIso) <> ")"
 


plutus-core/executables/src/PlutusCore/Executable/Types.hs
```diff
@@ -55,7 +55,7 @@ data TimingMode = NoTiming | Timing Integer deriving stock (Eq) -- Report progra
 data CekModel = Default | Unit -- Which cost model should we use for CEK machine steps?
 data PrintMode = Classic | Simple | Readable | ReadableSimple deriving stock (Show, Read)
 data NameFormat = IdNames | DeBruijnNames -- Format for textual output of names
-data TraceMode = None | Logs | LogsWithTimestamps | LogsWithBudgets deriving stock (Show, Read)
+data TraceMode = None | Logs | LogsWithTimestamps | LogsWithBudgets | LogsWithCallTrace deriving stock (Show, Read)
```
Unisay:
```suggestion
data TraceMode 
  = None 
  | Logs 
  | LogsWithTimestamps 
  | LogsWithBudgets 
  | LogsWithCallTrace 
  deriving stock (Show, Read)
```
(Since the line is longer than 100)

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/EmitterMode.hs
```diff
@@ -68,3 +70,36 @@ logWithBudgetEmitter = EmitterMode $ \getBudget -> do
...
+  logsRef <- newSTRef DList.empty
+  let
+    addTrace DList.Nil logs = logs
+    addTrace newLogs DList.Nil = newLogs
+    addTrace newLogs logs = DList.fromList $ go (DList.toList newLogs) (DList.toList logs)
+      where
+        go l [] = l
+        go [] l = l
+        go (x:xs) l =
+          case ((T.words $ last l), (T.words x)) of
```
Unisay:
```suggestion
          case (T.words (last l), T.words x) of
```

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/EmitterMode.hs
```diff
@@ -68,3 +70,36 @@ logWithBudgetEmitter = EmitterMode $ \getBudget -> do
...
+on compiler-generated trace calls that notifies entrance and exit of a function call. These traces that mark
+entrance and exit are ordinary traces like "entering rob:Example.hs:3:1-3:15" and
+"exiting bob:Example.hs:1:1-1:13" with "entering" and "exiting" prefixies, where "bob" and "rob" is the name of the
+function with source span. If regular script with no entrance/exit marker is given, this emitter will behave identically to
+'logEmitter'.
+
+When script evaluation fails, this emitter will give call trace of the functions that led to the evaluation failure.
+This is useful for pin-pointing specific area of the codebase that caused failure when debugging a script. When
+script evaluation passes, every trace message generated by `profile-all` flag will be removed, and this emitter
+will behave identically to 'logEmitter'.
```
Unisay: These lines exceed [our conventional line width limit of 100 characters](https://github.com/IntersectMBO/plutus/blame/1df397c6fc5f7957982931008f7deebacdc59e12/STYLEGUIDE.adoc#L41).

Please re-format them to fit 100.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/EmitterMode.hs
```diff
@@ -68,3 +70,36 @@ logWithBudgetEmitter = EmitterMode $ \getBudget -> do
...
+-}
+logWithCallTraceEmitter :: EmitterMode uni fun
+logWithCallTraceEmitter = EmitterMode $ \_ -> do
+  logsRef <- newSTRef DList.empty
+  let
+    addTrace DList.Nil logs = logs
+    addTrace newLogs DList.Nil = newLogs
+    addTrace newLogs logs = DList.fromList $ go (DList.toList newLogs) (DList.toList logs)
```
Unisay, line 94: I wonder if its possible to stay in the `DList` avoiding the `DList -> [] -> DList` conversion?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/EmitterMode.hs
```diff
@@ -68,3 +70,36 @@ logWithBudgetEmitter = EmitterMode $ \getBudget -> do
...
+      where
+        go l [] = l
+        go [] l = l
+        go (x:xs) l =
+          case ((T.words $ last l), (T.words x)) of
+            ("entering":enterRest, "exiting":exitRest) | enterRest == exitRest -> go xs (init l)
+            _                                                                  -> go xs (l <> [x])
+
+    emitter logs =
+      CekM $ modifySTRef logsRef (addTrace logs)
```
Unisay:
```suggestion
    emitter = CekM . modifySTRef logsRef . addTrace 
```

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/EmitterMode.hs
SeungheonOh: `DList` lacks `init` and `last`. I wasn't sure if `DList -> [] -> DList` is better or write `init` and `last` for `DList` is better

plutus-tx-plugin/test/CallTrace/9.6/func01.eval.golden
```diff
@@ -0,0 +1,12 @@
...
+Caused by: error
+
+Trace:
+entering func:test/CallTrace/Spec.hs:69:1-69:4
+func 1
+entering nestedLinear:test/CallTrace/Spec.hs:62:1-62:12
+entering nestedLinear2:test/CallTrace/Spec.hs:63:1-63:13
+entering nestedLinear3:test/CallTrace/Spec.hs:64:1-64:13
+entering nestedLinear4:test/CallTrace/Spec.hs:65:1-65:13
+entering error
```
Unisay: I've noticed that stack traces are only "entering" functions, there is never an "exiting" label.
Let's either demonstrate "exiting" in a test or remove it all together?

plutus-tx-plugin/test/CallTrace/Utils.hs
```diff
@@ -0,0 +1,59 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications  #-}
+
+module CallTrace.Utils where
```
Unisay: FYI: we have a convention to name supplementary modules in tests `Lib`
```suggestion
module CallTrace.Lib where
```

plutus-tx-plugin/test/CallTrace/Utils.hs
```diff
@@ -0,0 +1,59 @@
...
+          UPLC.counting
+          emitter
+          (uplc ^. UPLC.progTerm)
+      termSize = UPLC.programSize uplc
+      flatSize = UPLC.serialisedSize $ UPLC.UnrestrictedProgram uplc
+
+      traceMsg =
+        case logOut of
+          [] -> ["Trace: <empty>"]
+          x  -> [ "Trace:" , vsep $ pretty <$> x]
```
Unisay: Fourmolu?

plutus-tx-plugin/test/CallTrace/Utils.hs
```diff
@@ -0,0 +1,59 @@
...
+import PlutusCore.Evaluation.Machine.ExBudgetingDefaults qualified as PLC
+import PlutusCore.Pretty (Pretty (pretty), Render (render), prettyPlcClassicSimple, prettyReadable)
+import PlutusCore.Test (TestNested, ToUPlc (..), nestedGoldenVsDocM, ppCatch)
+import UntypedPlutusCore qualified as UPLC
+import UntypedPlutusCore.Evaluation.Machine.Cek qualified as UPLC
+import UntypedPlutusCore.Evaluation.Machine.Cek.Internal qualified as UPLC
+
+import PlutusTx.Test (prettyBudget)
+
+goldenEvalCekTraceWithEmitter :: UPLC.EmitterMode UPLC.DefaultUni UPLC.DefaultFun -> TestName -> CompiledCode a -> TestNested
```
Unisay:
```suggestion
goldenEvalCekTraceWithEmitter 
  :: UPLC.EmitterMode UPLC.DefaultUni UPLC.DefaultFun 
  -> TestName 
  -> CompiledCode a 
  -> TestNested
```

plutus-tx-plugin/test/CallTrace/Utils.hs
```diff
@@ -0,0 +1,59 @@
...
+    uplc <- toUPlc compiledCode
+    let
+      (evalRes, UPLC.CountingSt budget, logOut) =
+        UPLC.runCek
+          PLC.defaultCekParametersForTesting
+          UPLC.counting
+          emitter
+          (uplc ^. UPLC.progTerm)
+      termSize = UPLC.programSize uplc
+      flatSize = UPLC.serialisedSize $ UPLC.UnrestrictedProgram uplc
```
Unisay: You should be able to use `countAstNodes` and `countFlatBytes` functions once [this PR](https://github.com/IntersectMBO/plutus/pull/7172) is merged.

plutus-tx-plugin/test/CallTrace/Utils.hs
```diff
@@ -0,0 +1,59 @@
...
+      Right termRes ->
+        pure $
+          render @Text $
+            vsep
+              [ prettyBudget budget termSize flatSize
+              , mempty
+              , vsep traceMsg
+              , mempty
+              , prettyPlcClassicSimple termRes
+              ]
```
Unisay:
```suggestion
    pure . render @Text . vsep $
      case evalRes of
        Left evalErr ->
          [ prettyReadable evalErr
          , mempty
          , vsep traceMsg
          ]
        Right termRes ->
          [ prettyBudget budget termSize flatSize
          , mempty
          , vsep traceMsg
          , mempty
          , prettyPlcClassicSimple termRes
          ]
```

plutus-tx-plugin/test/CallTrace/Spec.hs
```diff
@@ -0,0 +1,94 @@
...
+func 7 = trace "func 7" $ myClassFuncInOtherModule @Integer 7
+func 8 = myClassFuncInOtherModule @Integer 8
+func 9 = trace "func 9" $ myClassFuncInOtherModule ()
+func n =
+  if bob 42 n == ()
+  then error ()
+  else ""
+
+functionFromOtherModule :: Bool -> BuiltinString
+functionFromOtherModule x = wraps (not $ not x)
```
Unisay:
```suggestion
functionFromOtherModule x = wraps $ not $ not x
```
(HLint suggestion)

plutus-tx-plugin/test/CallTrace/Spec.hs
```diff
@@ -0,0 +1,94 @@
...
+
+nestedLinear, nestedLinear2, nestedLinear3, nestedLinear4 :: Bool -> BuiltinString
+nestedLinear x = nestedLinear2 x
+nestedLinear2 x = nestedLinear3 x
+nestedLinear3 x = nestedLinear4 x
+nestedLinear4 True  = error ()
+nestedLinear4 False = error ()
+
+func :: Integer -> BuiltinString
+func 1 = (trace "func 1" (\_thunk -> nestedLinear True)) ()
```
Unisay:
```suggestion
func 1 = trace "func 1" (\_thunk -> nestedLinear True) ()
```

Unisay (commented):
This approach looks less invasive, indeed!

SeungheonOh:
Sorry for all the formatting error!

plutus-tx-plugin/test/CallTrace/9.6/func01.eval.golden
SeungheonOh: The emitter is intentionally removing all `exiting` labels because if we have both enter and exit trace that means that function ran without causing error. X

If we would've used `logEmitter` it will report all entering and exiting. The idea of `logWithCallTraceEmitter` is to remove these enter/exit for functions that we don't care(because they didn't cause failure).

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/EmitterMode.hs
```diff
@@ -68,3 +70,35 @@ logWithBudgetEmitter = EmitterMode $ \getBudget -> do
...
+    addTrace DList.Nil logs = logs
+    addTrace newLogs DList.Nil = newLogs
+    addTrace newLogs logs = DList.fromList $ go (DList.toList newLogs) (DList.toList logs)
+      where
+        go l [] = l
+        go [] l = l
+        go (x:xs) l =
+          case (T.words (last l), T.words x) of
+            ("->":enterRest, "<-":exitRest) | enterRest == exitRest -> go xs (init l)
+            _                                                       -> go xs (l <> [x])
```
Unisay: Fourmolu changes this code to:
image

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -780,9 +791,9 @@ entryExitTracing lamName displayName e ty =
         annMayInline
         ( mkTrace
             (PLC.TyFun annMayInline defaultUnitTy ty) -- ()-> ty
-            ("entering " <> displayName)
+            ("-> " <> displayName)
```
Unisay, line 808: I see that these special strings (`->`, `<-`) are used in several places:
`plutus-core/executables/traceToStacks/Common.hs`
and here.

I suggest to define them once (as a binding) and then refer to their defined names from all places using them. This helps static code navigation, makes the link more explicit.

plutus-tx-plugin/test/CallTrace/9.6/func01.eval.golden
```diff
@@ -0,0 +1,12 @@
...
+Caused by: error
+
+Trace:
+-> func:test/CallTrace/Spec.hs:78:1-78:4
+func 1
+-> nestedLinear:test/CallTrace/Spec.hs:71:1-71:12
+-> nestedLinear2:test/CallTrace/Spec.hs:72:1-72:13
+-> nestedLinear3:test/CallTrace/Spec.hs:73:1-73:13
+-> nestedLinear4:test/CallTrace/Spec.hs:74:1-74:13
+-> error
```
Unisay: Looks nice, IMO: the direction of calls is clear.
Having a space between the function name and its source location makes it even easier to 👀  scan:
```
Trace:
-> func (test/CallTrace/Spec.hs @ 78:1-78:4)
func 1
-> nestedLinear (test/CallTrace/Spec.hs @ 71:1-71:12)
-> nestedLinear2 (test/CallTrace/Spec.hs @ 72:1-72:13)
-> nestedLinear3 (test/CallTrace/Spec.hs @ 73:1-73:13)
-> nestedLinear4 (test/CallTrace/Spec.hs @ 74:1-74:13)
-> error
```

plutus-tx-plugin/test/CallTrace/9.6/funcionFromOtherModule.eval.golden
Unisay, line 1: I don't see a value in having this golden file: what exactly is being tested here?

plutus-tx-plugin/test/CallTrace/Lib.hs
```diff
@@ -0,0 +1,58 @@
...
+      (evalRes, UPLC.CountingSt budget, logOut) =
+        UPLC.runCek
+          PLC.defaultCekParametersForTesting
+          UPLC.counting
+          emitter
+          (uplc ^. UPLC.progTerm)
+
+      traceMsg =
+        case logOut of
+          [] -> ["Trace: <empty>"]
```
Unisay: What if I `trace "<empty>"` ?
Will it be distinguishable from not tracing anything?

plutus-tx-plugin/test/CallTrace/Spec.hs
```diff
@@ -0,0 +1,103 @@
...
+{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:profile-all #-}
+
+{-# HLINT ignore "Eta reduce" #-}
+{-# HLINT ignore "Redundant if" #-}
+
+module CallTrace.Spec where
+
+import PlutusTx
+import PlutusTx.Prelude
+import Prelude ()
```
Unisay:
```suggestion
```

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
SeungheonOh: I wasn't sure where the best place for this would be:

This enter/exit marker is used in `plutus-executable:traceToStack`, `plutus-tx-plugin`, and `plutus-core:untyped-plutus-core`. For dependency purpose, this would need to go into `plutus-core` but logically, this does not belong in `plutus-core` since this is profiling marker and `plutus-tx-plugin` creates scripts with these profile markers.

plutus-tx-plugin/test/CallTrace/9.6/funcionFromOtherModule.eval.golden
SeungheonOh: This tests that when program got evaluated without failure, there's no enter/exit marker left on the trace log.

plutus-tx-plugin/test/CallTrace/Lib.hs
SeungheonOh: Good point!

plutus-tx-plugin/test/CallTrace/Spec.hs
SeungheonOh: We need this because we don't have `NoImplicitPrelude` on

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -650,19 +649,31 @@ hoistExpr var t = do
...
-  let ty = PLC._varDeclType var
-      varName = PLC._varDeclName var
-      displayName = T.pack $ PP.displayPlcSimple varName
-      isFunctionOrAbstraction = case ty of PLC.TyFun{} -> True; PLC.TyForall{} -> True; _ -> False
+  let
+    nameStr = GHC.occNameString $ GHC.occName $ GHC.varName $ ghcVar
+    displayName = T.pack $
+      case getVarSourceSpan ghcVar of
+        -- FIXME: Variables will miss span information when the module they are defined
+        -- in is loaded from cache instead of getting compiled.
```
Unisay: We have recently agreed to avoid leaving orphaned `TODO`, `FIXME` comments in the code. The best course of action would be:

- Create a github issue, such that a FIXME or TODO work can be a part of our dev process: triaged, assigned to a sprint, assigned to a person ("owned") etc.
- Amend your `FIXME` comment with a link to the issue.

(The "triage" role that you assume this sprint includes work on removing such orphaned comments ;-) )

plutus-tx-plugin/test/CallTrace/9.6/func02.eval.golden
```diff
@@ -0,0 +1,10 @@
+An error has occurred:
+The machine terminated because of an error, either from a built-in function or from an explicit use of 'error'.
+Caused by: error
+
+Trace:
+-> func (test/CallTrace/Spec.hs:78:1-78:4)
```
Unisay, line 6: This encoding `:78:1-78:4` confuses me: what does each section mean ?

plutus-tx-plugin/test/CallTrace/9.6/funcionFromOtherModule.eval.golden
Unisay: This explanation makes sense to me. Do you mind embedding it into the name of the golden file, e.g.:
`successfullEvaluationYieldsNoTraceLog` or something along these lines?

plutus-tx-plugin/test/CallTrace/OtherModule.hs
```diff
@@ -0,0 +1,23 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module CallTrace.OtherModule where
+
+import PlutusTx.Prelude
+import Prelude ()
```
Unisay:
```suggestion
```

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
Unisay: My concern is that its a hidden dependency:
`plutus-tx-plugin` depends on the fact that `plutus-core` uses `->`, `<-` markers but this dependency is not typechecked, so there is a higher chance for it to get broken, as its implicit.

I'd rather have a honest (but not so beautiful) approach: make plutus core aware of the fact that it has to emit certain makers used by other tools for profiling.

Unisay: @effectfully what is your opinion on this matter?

Unisay (changes requested):
Left a few comments.

plutus-tx-plugin/test/CallTrace/9.6/func02.eval.golden
SeungheonOh: from line 78 column 1 to line 78 column 4. this is what SrcSpan pretty printer does. I don't really want to change the pretty instance.

plutus-tx-plugin/test/CallTrace/OtherModule.hs
SeungheonOh: Again, we need this because we don't have `NoImplicitPrelude`

Unisay: Add `NoImplicitPrelude` ? It seems to be more explicit in its purpose than `import Prelude ()`

plutus-tx-plugin/test/CallTrace/9.6/func02.eval.golden
Unisay: I see!
My first attempt to parse it in the head produced this result:

1. `test/CallTrace/Spec.hs`
1. `78`
1. `1-78`
1. `4`

🤣 

I think that my mental parser trips over the `:` as a separator between the file path (LHS) and `SrcSpan` (RHS).
I agree that we don't want to change the way `SrcSpan` is printed here,
but we can avoid mixing the same char `:` to separate 
1. File path / Src Span
2. Line / Column

I suggested to use ` @ ` as a 1. separator before.

plutus-tx-plugin/test/CallTrace/Spec.hs
Unisay: I'd much prefer `NoImplicitPrelude` for this purpose.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
SeungheonOh: I added a note:
```
{- Note [Profiling Markers]
   The @profile-all@ will insert trarces when entering and exciting functions. These
   traces have a string marker to indicate that a given traces message is for enter/exit
   marking. Markers are just simple strings: "->" and "<-". So for any reason in the
   future this marker needs to be changed, all of utilities that uses this marker will
   need to be updated.

   This list will track of all of the utilities that uses this marker:
   - plutus-core:traceToStacks
   - @UntypedPlutusCore.Evaluation.Machine.Cek.EmitterMode.logWithCallTraceEmitter@
-}
```

I think this is sufficient.

plutus-tx-plugin/test/CallTrace/OtherModule.hs
SeungheonOh: Ah, necessity to add **explicitly** add `NoImplicitPrelude` to make prelude not implicit!

I generally not prefer adding extra language extension, but I don't really mind here.

plutus-tx-plugin/test/CallTrace/9.6/func02.eval.golden
SeungheonOh: We can't change this without changing `Pretty` instances

```hs
Just src -> nameStr <> " (" <> show (src ^. srcSpanIso) <> ")"
```

I think this is okay honestly. This is standard format used by GHC(and other languages as well!), e.g. `test/CallTrace/Spec.hs:88:10`

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
SeungheonOh: done #7203

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
effectfully: > I'd rather have a honest (but not so beautiful) approach: make plutus core aware of the fact that it has to emit certain makers used by other tools for profiling.

I agree, I'd prefer those special symbols to be given names in Haskell. But don't care much, a Note is good enough.

PR #7182, opened 2025-07-03
QuickCheck-2.16 support

coot:
We also need a new release of `1.48` for `cardano-node-10.5`.  To avoid integration changes ideally it would be just this patch over `1.48.0.0` tag (`master` is already incompatible with `ouroboros-consensus` which is integrated with `plutus-core-1.48.0.0`).


nix/project.nix
```diff
@@ -12,6 +12,9 @@ let
 
       src = ../.;
 
+      index-state = "2025-06-17T07:53:04Z";
+      index-sha256 = "sha256-/9/Z6Fpdzil6tDmMrKMKKVcnXksfIpLOQ/2drHo1Rts=";
+
```
Unisay: Why do we need to pin cabal index two times, once in the `cabal.project` and 2nd time here?

nix/project.nix
```diff
@@ -91,6 +94,9 @@ let
             plutus-tx-plugin.ghcOptions = [ "-Werror" ];
           };
         }
+        ({ lib, pkgs, ... }: lib.mkIf (pkgs.stdenv.hostPlatform.isWindows) {
+          packages.basement.configureFlags = [ "--hsc2hs-option=--cflag=-Wno-int-conversion" ];
```
Unisay, line 101: Is there a problem with int-conversion on Windows? Would be nice to have a comment explaining this special case with a reference to an issue if applicable.

cabal.project
```diff
@@ -87,3 +87,6 @@ allow-newer:
 --   flags: +force-build
 -- allow-newer: *:plutus-ledger-api
 -- allow-older: *:nothunks
+
+allow-newer:
+  *:QuickCheck
```
Unisay: Why is it needed?

We'd like to minimise amount of exceptions like this relying on cabal being able to create a build plan. 

That being said, I understand that sometimes such workarounds are adequate. If this is the case then we'd want to eventually remove it, but its important to track the original problem to know when it could be removed. 

Please add a comment explaining the conflict this exception is working around.

nix/project.nix
coot: Afaik, this is a requirement of a new version of `haskall.nix`.  But there's also an upside, one doesn't need to run `nix flake update --update-input haddock` anymore, nor one needs to wait a day until the `hackage.nix` is rebuild after publishing a package on hackage.

cabal.project
coot: As of today there are two packages which require it (it was more than that :grin:):
* https://github.com/hedgehogqa/haskell-hedgehog/issues/548
* https://github.com/phadej/vec/issues/121

coot: I updated this `allow-newer` to not use a `*`.

nix/pkgs.nix
```diff
@@ -9,5 +9,17 @@ import inputs.nixpkgs {
     inputs.haskell-nix.overlay
     inputs.iohk-nix.overlays.haskell-nix-crypto
     inputs.iohk-nix.overlays.haskell-nix-extra
+    # Pin R to 4.4.3 until inline-r is fixed.
```
hamishmack: We should be able to remove this now and `nix flake update hackage` instead.  See https://github.com/tweag/HaskellR/pull/433, which was released to https://hackage.haskell.org/package/inline-r as 1.0.2.

coot: I removed that commit, and force-pushed the branch.

nix/project.nix
coot: This fixes [this error](https://ci.iog.io/build/8529222/nixlog/1):
```
Preprocessing library for basement-0.0.16...
Size.hsc:126:30: error: initialization of ‘long long int’ from ‘void *’ makes integer from pointer without a cast []
compilation failed
```

`basement` repo was archived, so there's no way to open a ticket.  Maybe you'll need to consider moving away of it.

Unisay: This is how we inherit `basement`:
```
 plutus-core-1.49.0.0
   └─ cardano-crypto-class-2.2.0.0
       └─ crypton-1.0.1
           └─ memory-0.18.0
               └─ basement-0.0.16
  ```
The newest `cardano-crypto-class` depends on `crypton` https://chap.intersectmbo.org/package/cardano-crypto-class-2.2.2.0/

The newest `crypton` depends on `basement`:
https://flora.pm/packages/@hackage/crypton/1.0.4/dependencies

@coot please add the explanation you gave in the github comment above as a code comment in the `nix` file, where the flags are set.

File: nix/project.nix (hunk 1 above)
Comment (Unisay) [reply nix/project.nix]: @zeme-wana wdyt?

File: nix/project.nix (hunk 2 above)
Comment (coot) [reply nix/project.nix:101]: I opened an issue for it: kazu-yamamoto/crypton#57

File: nix/project.nix (hunk 1 above)
Comment (zeme-wana) [reply nix/project.nix]: We've never needed to pin the index twice. I'd rather not do it unless necessary. We're happy with and used to a workflow that includes `nix flake update --update-input hackage`. I'll try and remove it and see if haskell.nix is still happy

Comment (coot) [reply nix/project.nix]: Unfortunately, it is necessary.

Comment (coot) [reply nix/project.nix]: @hamishmack, you know better if pinning twice (e.g. in `cabal.project` and in `callProject') is really required here (_I didn't get things to work without it_).

Comment (zeme-wana) [reply nix/project.nix]: Is it really necessary? I just removed it and CI is green.

Comment (coot) [reply nix/project.nix]: When I do that localy and I run:
```bash
nix build -j12 .\#hydraJobs.x86_64-linux.required
```
then I get this error:
```
Ignoring ./doc/notes/fomega/package.yaml as ./doc/notes/fomega/fomega.cabal exists
Cloning into '/build/tmp.YLIDDwqEuc/dist-newstyle/src/wssdcgk0i_-6c4b706a3e6469a07bcb1f0f008f40c18fb145572f356ba8d1555be83d87e6b2'.
..
remote: Enumerating objects: 454, done.
remote: Counting objects: 100% (454/454), done.
remote: Compressing objects: 100% (416/416), done.
remote: Total 454 (delta 22), reused 0 (delta 0), pack-reused 0 (from 0)
Receiving objects: 100% (454/454), 1.27 MiB | 9.07 MiB/s, done.
Resolving deltas: 100% (22/22), done.
HEAD is now at 516b4cb Minimal Repo For Haskell.Nix
Configuration is affected by the following files:
- cabal.project
Warning: The package list for 'hackage.haskell.org' is 20285 days old.
Run 'cabal update' to get the latest list of available packages.
Error: [Cabal-7159]
Latest known index-state for 'hackage.haskell.org' (2025-07-03T02:06:05Z) is older than the requested index-state (2025-07-11T12:32
:55Z).
Run 'cabal update' or set the index-state to a value at or before 2025-07-03T02:06:05Z.
```

coot: oh, I forgot to update `flake.nix`; I'll try your changes.

coot: It seems you're right (it's building but it didn't fail early as it used to).  I think we can resolve this thread.

jasagredo:
I just rebased the PR to get the filename fix in #7216 on the branch.

Unisay (approved):
Thank you, @coot for making the update and addressing comments!
Lets get it merged (if it compiles then it works, right? :) )

PR #7188, opened 2025-07-07
[Builtins] Allow casing on lists

SeungheonOh:
Part of #6602

We haven't fully decided how casing on list should be. I tried simply casing into nil case and cons case.


SeungheonOh:
This casing on list is not actually used anywhere, this is just to see if adding `ann` field in `FrameCases` changes anything significantly. 

/benchmark lists

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -614,7 +614,7 @@ data Context uni fun ann
     -- See Note [Accumulators for terms]
     | FrameConstr !(CekValEnv uni fun ann) {-# UNPACK #-} !Word64 ![NTerm uni fun ann] !(ArgStack uni fun ann) !(Context uni fun ann)
     -- ^ @(constr i V0 ... Vj-1 _ Nj ... Nn)@
-    | FrameCases !(CekValEnv uni fun ann) !(V.Vector (NTerm uni fun ann)) !(Context uni fun ann)
+    | FrameCases ann !(CekValEnv uni fun ann) !(V.Vector (NTerm uni fun ann)) !(Context uni fun ann) -- TODO: This is bad
```
SeungheonOh: Interestingly, this doesn't hurt the performance. I still think this is bad tough

SeungheonOh: Okay, #7189 will allow us to fix this by constructing term with arbitrary annotation.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -530,9 +532,19 @@ outOfBoundsErr x branches = fold
...
+        DefaultUniList ty ->
+          case branches of
+            [nil, cons] ->
+              Right
+                [ (nil, [])
+                , (cons, [ TyBuiltin @TyName ann $ SomeTypeIn ty
+                         , TyBuiltin @TyName ann $ SomeTypeIn $ DefaultUniList ty
+                         ])
+                ]
+            _ -> Left $ "Casing on list requires two branches"
```
effectfully: I discussed it with @zliu41 and IIRC we agreed that we should allow only one branch too for the cons case, since a lot of time it's expected that the list is gonna be non-empty and it would be wasteful to have a branch for the nil case when it's supposed to fail evaluation anyway.

@zliu41 we still need to reconcile it with `False` being the default for `Bool`. It's not too late to change things I think. Maybe we should make `True` and `(:)` the default branches for consistency. We can do it with or without reordering them. So let's discuss it some time.

plutus-core/plutus-core/src/PlutusCore/Builtin/Case.hs
```diff
@@ -19,25 +19,32 @@ import Universe
...
 class CaseBuiltin uni where
     -- | Given a constant with its type tag and a vector of branches, choose the appropriate branch
     -- or fail if the constant doesn't correspond to any of the branches (or casing on constants of
     -- this type isn't supported at all).
-    caseBuiltin :: UniOf term ~ uni => Some (ValueOf uni) -> Vector term -> Either Text term
+    caseBuiltin
+        :: UniOf term ~ uni
+        => Some (ValueOf uni)
+        -> Vector term
+        -> Either Text ([Some (ValueOf uni)], term)
```
effectfully: We should use `PlutusCore.Builtin.KnownType.HeadSpine` here except it needs to be changed to allow for the spine to contain elements of a different type than the head. What you have here isn't efficient.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -793,20 +793,21 @@ enterComputeCek = computeCek
...
+        -- Otherwise, we can safely convert the index to an Int and use it.
         (VConstr i args) -> case (V.!?) cs (fromIntegral i) of
             Just t  -> computeCek (transferArgStack args ctx) env t
             Nothing -> throwErrorDischarged (StructuralError $ MissingCaseBranchMachineError i) e
+        -- Proceed with caser when expression given is not Constr.
         VCon val -> case unCaserBuiltin ?cekCaserBuiltin val cs of
-            Left err  -> throwErrorDischarged (OperationalError $ CekCaseBuiltinError err) e
-            Right res -> computeCek ctx env res
+            Left err          -> throwErrorDischarged (OperationalError $ CekCaseBuiltinError err) e
+            Right (args, res) -> computeCek ctx env (foldl (Apply ann) res (Constant ann <$> args))
```
effectfully: This is the source of the annotation issues. You aren't supposed to create new terms in the CEK machine. Use

```haskell
transferConstantSpine
    :: Spine (Some (ValueOf uni))
    -> Context uni fun ann
    -> Context uni fun ann
transferConstantSpine args ctx = foldr (FrameAwaitFunValue . VCon) ctx args
{-# INLINE transferConstantSpine #-}
```

to make it something like

```haskell
computeCek (transferConstantSpine args ctx) env res
```

Rename `transferSpine` to `transferValueSpine` while you're there and don't define `transferConstantSpine` in terms of it to avoid reallocating the spine (it's not a list and so you can't rely on fusion to deforest it for you).

What you're doing know is creating a term just to chop it up back into new context entries later, much better to do the latter directly.

plutus-core/plutus-core/src/PlutusCore/Builtin/Case.hs
SeungheonOh: `-- Used in built-in functions returning function applications such as 'CaseList'.` alright, should've read more 😔 Thank you.

effectfully: It wasn't obvious at all.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
SeungheonOh: I added single case as well

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
```diff
@@ -287,16 +290,32 @@ data Spine a
...
+  bimap headF spineF (HeadSpine a b) = HeadSpine (headF a) (spineF <$> b)
+
+-- | Construct @HeadSpine@ from head and list.
+headSpine :: a -> [b] -> HeadSpine a b
+headSpine h [] = HeadOnly h
+headSpine h (x:xs) =
+  -- It's critical to use 'foldr' here, so that deforestation kicks in.
+  -- See Note [Definition of foldl'] in "GHC.List" and related Notes around for an explanation
+  -- of the trick.
+  HeadSpine h $ foldr (\x2 r x1 -> SpineCons x1 $ r x2) SpineLast xs x
```
effectfully, line 318: Make sure to add the `INLINE` pragma.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: Change the error message too.

I suppose do it for booleans too while you're here? To make it consistent.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -541,12 +559,22 @@ instance CaseBuiltin DefaultUni where
...
-            True  |             len == 2 -> Right $ branches Vector.! 1
+            False | len == 1 || len == 2 -> Right $ HeadOnly $ branches Vector.! 0
+            True  |             len == 2 -> Right $ HeadOnly $ branches Vector.! 1
             _                            -> Left  $ outOfBoundsErr x branches
         DefaultUniInteger
-            | 0 <= x && x < toInteger len -> Right $ branches Vector.! fromInteger x
+            | 0 <= x && x < toInteger len -> Right $ HeadOnly $ branches Vector.! fromInteger x
             | otherwise                   -> Left  $ outOfBoundsErr x branches
+        DefaultUniList ty
+            | length branches == 1 ->
```
effectfully: Don't recompute the length, especially using the slow definition from `Data.Foldable`. Use `len` bound in the `where` branch below.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -541,12 +559,22 @@ instance CaseBuiltin DefaultUni where
...
+              case x of
+                [] -> Left "Expected non-empty list, got empty list"
+                (y : ys) -> Right $ headSpine (branches Vector.! 0) [Some (ValueOf ty y), Some (ValueOf uni ys)]
+            | length branches == 2 ->
+              case x of
+                []       -> Right $ HeadOnly $ branches Vector.! 0
+                (y : ys) -> Right $ headSpine (branches Vector.! 1) [Some (ValueOf ty y), Some (ValueOf uni ys)]
+            | otherwise            -> Left "Casing on builtin list requires exactly two branches"
```
effectfully: Please use `outOfBoundsErr` for all the error messages here. If it doesn't provide some information you want it to provide, add it in there.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1107,15 +1107,15 @@ functions.
 -}
 
 -- | Take a function and a list of arguments and apply the former to the latter.
-headSpine :: Opaque val asToB -> [val] -> Opaque (HeadSpine val) b
-headSpine (Opaque f) = Opaque . \case
+headSpineOpaque :: Opaque val asToB -> [val] -> Opaque (MonoHeadSpine val) b
+headSpineOpaque (Opaque f) = Opaque . \case
```
effectfully, line 1108: Reuse `headSpine`? Doesn't matter much, I'm gonna remove this anyway.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -541,12 +559,22 @@ instance CaseBuiltin DefaultUni where
...
             _                            -> Left  $ outOfBoundsErr x branches
         DefaultUniInteger
-            | 0 <= x && x < toInteger len -> Right $ branches Vector.! fromInteger x
+            | 0 <= x && x < toInteger len -> Right $ HeadOnly $ branches Vector.! fromInteger x
             | otherwise                   -> Left  $ outOfBoundsErr x branches
+        DefaultUniList ty
+            | length branches == 1 ->
+              case x of
+                [] -> Left "Expected non-empty list, got empty list"
+                (y : ys) -> Right $ headSpine (branches Vector.! 0) [Some (ValueOf ty y), Some (ValueOf uni ys)]
```
effectfully: You can save some typing by using `someValueOf`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -614,7 +614,7 @@ data Context uni fun ann
     -- See Note [Accumulators for terms]
     | FrameConstr !(CekValEnv uni fun ann) {-# UNPACK #-} !Word64 ![NTerm uni fun ann] !(ArgStack uni fun ann) !(Context uni fun ann)
     -- ^ @(constr i V0 ... Vj-1 _ Nj ... Nn)@
-    | FrameCases !(CekValEnv uni fun ann) !(V.Vector (NTerm uni fun ann)) !(Context uni fun ann)
+    | FrameCases !(CekValEnv uni fun ann) !(V.Vector (NTerm uni fun ann)) !(Context uni fun ann) -- TODO: This is bad
```
effectfully: No longer bad :slightly_smiling_face:

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -800,23 +808,25 @@ enterComputeCek = computeCek
...
         (VConstr i args) -> case (V.!?) cs (fromIntegral i) of
             Just t  -> computeCek (transferArgStack args ctx) env t
             Nothing -> throwErrorDischarged (StructuralError $ MissingCaseBranchMachineError i) e
+        -- Proceed with caser when expression given is not Constr.
         VCon val -> case unCaserBuiltin ?cekCaserBuiltin val cs of
-            Left err  -> throwErrorDischarged (OperationalError $ CekCaseBuiltinError err) e
-            Right res -> computeCek ctx env res
+            Left err          -> throwErrorDischarged (OperationalError $ CekCaseBuiltinError err) e
+            Right (HeadOnly fX) -> computeCek ctx env fX
+            Right (HeadSpine f xs) -> computeCek (transferConstantSpine xs ctx) env f
```
effectfully, line 819: Using `Either` here was stupid on my part, should've introduced a dedicated strict data type.

But now that I look at it, it's probably best to inline `HeadSpine` business in such a data type, so that it has three constructors (error / success with only head / success with head and spine).

Not in this PR though, let's do it separately and measure whether it helps.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -530,9 +532,25 @@ outOfBoundsErr x branches = fold
...
-    annotateCaseBuiltin (SomeTypeIn uni) branches = case uni of
+    annotateCaseBuiltin ann (SomeTypeIn uni) branches = case uni of
         DefaultUniBool    -> Right $ map (, []) branches
         DefaultUniInteger -> Right $ map (, []) branches
+        DefaultUniList ty ->
+          case branches of
+            [cons] ->
+              Right
+                [ (cons, [ TyBuiltin @TyName ann $ SomeTypeIn ty
+                         , TyBuiltin @TyName ann $ SomeTypeIn $ DefaultUniList ty
```
effectfully: You can save some typing with `PlutusCore.MkPlc.mkTyBuiltinOf`.

You shouldn't need the `TyName` type application I think?

plutus-core/plutus-core/src/PlutusCore/MkPlc.hs
```diff
@@ -339,5 +340,11 @@ mkFreshTermLet aT a = do
     pure (var mempty genName, termLet mempty (Def (VarDecl mempty genName aT) a))
 
 -- | 'apply' the head of the application to the arguments iteratively.
-headSpineToTerm :: TermLike term tyname name uni fun => HeadSpine (term ()) -> term ()
-headSpineToTerm = foldl1 (apply ())
+headSpineToTerm :: TermLike term tyname name uni fun => ann -> MonoHeadSpine (term ann) -> term ann
+headSpineToTerm _ (HeadOnly t)       = t
+headSpineToTerm ann (HeadSpine t ts) = apply ann t $ foldl1 (apply ann) ts
```
effectfully: This doesn't look right at all. Does it not fail any tests?

Just use `foldl`.

plutus-core/plutus-core/src/PlutusCore/MkPlc.hs
```diff
@@ -339,5 +340,11 @@ mkFreshTermLet aT a = do
...
+
+-- | @headSpineToTerm@ but without annotation.
+headSpineToTermNoAnn :: TermLike term tyname name uni fun => MonoHeadSpine (term ()) -> term ()
+headSpineToTermNoAnn (HeadOnly t)     = t
+headSpineToTermNoAnn (HeadSpine t ts) = apply () t $ foldl1 (apply ()) ts
```
effectfully: Same.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
SeungheonOh: I'll make a PR for this after this one

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
SeungheonOh: Yes, this is vestige of hostile conversation I had with GHC

plutus-core/plutus-core/src/PlutusCore/MkPlc.hs
SeungheonOh: brainfart!

plutus-core/plutus-core/test/TypeSynthesis/Golden/Signatures/DefaultFun/CaseData.sig.golden
```diff
@@ -1 +1 @@
-forall a. Opaque Val (Integer -> [Data] -> TyVarRep * ('TyNameRep * "a" 0)) -> Opaque Val ([(Data,Data)] -> TyVarRep * ('TyNameRep * "a" 0)) -> Opaque Val ([Data] -> TyVarRep * ('TyNameRep * "a" 0)) -> Opaque Val (Integer -> TyVarRep * ('TyNameRep * "a" 0)) -> Opaque Val (ByteString -> TyVarRep * ('TyNameRep * "a" 0)) -> Data -> Opaque (HeadSpine Val) (TyVarRep * ('TyNameRep * "a" 0))
```
SeungheonOh: This will go anyways, but need to appease CI

plutus-core/plutus-core/src/PlutusCore/MkPlc.hs
effectfully: I need to know if it failed any tests! 'Cause if it didn't, that's a problem on its own.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
SeungheonOh: For Bool, do we want to enforce that number of branches can only be exactly one or two? 

Current the type annotation allows some other "trailing" branches to be present--like `case (con bool False) foo bar baz bob rob cob)` where everything besides `foo` and `bar` will be ignored. However, current builtin casing itself doesn't allow it, so it would need to be fixed to match one behavior or another. 

I don't see any use of this; bool casing with more than two branches would need to be statically removed anyways for script size optimization. We should only allow one or two branches on bool casing

SeungheonOh: I'll go forward with only one or two branches allowed, but please let me know if there is a good argument against this.

effectfully: Yeah, that's what I attempted to suggest. Thanks.

plutus-core/plutus-core/src/PlutusCore/Builtin/Case.hs
```diff
@@ -19,25 +20,31 @@ import Universe
 class AnnotateCaseBuiltin uni where
     -- | Given a tag for a built-in type and a list of branches, annotate each of the branches with
     -- its expected argument types or fail if casing on values of the built-in type isn't supported.
+    -- Note: you don't need to include the resulting type of the whole case matching in the
+    -- returning list here.
     annotateCaseBuiltin
         :: UniOf term ~ uni
-        => SomeTypeIn uni
+        => Type TyName uni ann
```
SeungheonOh, line 27: This is necessary to make typechecker work on higher typed builtin types with type variable argument like `[(con list) a]`.

effectfully:
Not bad!

SeungheonOh:
`sum/compiled-from-Haskell/sum-right-builtin` improvements are nice! Almost 35% avg on that category

effectfully:
Make sure to check that the budget reductions are aligned with the performance improvements. We don't want to see, say, a 70% budget reduction when the code is only 40% faster.

cardano-constitution/test/Cardano/Constitution/Validator/GoldenTests/unsorted.pir.golden
```diff
@@ -416,14 +416,7 @@ program
...
           /\r ->
-            \(z : r) (f : a -> list a -> r) (xs : list a) ->
-              chooseList
-                {a}
-                {all dead. r}
-                xs
-                (/\dead -> z)
-                (/\dead -> f (headList {a} xs) (tailList {a} xs))
-                {r})
+            \(z : r) (f : a -> list a -> r) (xs : list a) -> case r xs [z, f])
```
effectfully, line 419: Hm, check how `annAlwaysInline` instead of `annMayInline` in `PlutusTx.Compiler.Builtins` affects things.

SeungheonOh:
Do we currently have configurable budgeting mechanism for casing? I haven't seen any budget related code on CEK for casing.

plutus-benchmark/script-contexts/test/V1/9.6/checkScriptContext1-20.eval.golden
```diff
@@ -1,6 +1,6 @@
-CPU:               239_576_808
-Memory:                906_600
-Term Size:               1_391
-Flat Size:               1_977
+CPU:               186_924_676
+Memory:                721_804
```
effectfully, line 2: 20% MEM reduction is pretty wild.

plutus-benchmark/script-contexts/test/V1/Data/9.6/checkScriptContext1-20.eval.golden
```diff
@@ -1,6 +1,6 @@
-CPU:                20_088_713
-Memory:                 72_250
-Term Size:                  89
-Flat Size:                 654
+CPU:                 8_111_579
+Memory:                 33_498
+Term Size:                  71
+Flat Size:                 641
```
effectfully, line 4: Wow.

plutus-benchmark/script-contexts/test/V2/9.6/dataFwdStakeTrickManual.eval.golden
```diff
@@ -1,6 +1,6 @@
 CPU:                 5_706_022
 Memory:                 10_678
-Term Size:                 165
-Flat Size:               1_208
+Term Size:                 147
+Flat Size:               1_194
```
effectfully, line 4: Wait, how is is possible that CPU and MEM didn't change?

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Definition.hs
```diff
@@ -1251,8 +1251,8 @@ test_Case =
                             (mkConstant () scrut)
                             (map (mkConstant ()) is)
                 in case typecheckEvaluateCekNoEmit def defaultBuiltinCostModelForTesting term of
-                    Left _                        -> False
-                    Right EvaluationFailure       -> length is /= 2 && (scrut || length is /= 1)
+                    Left _                        -> length is /= 1 && length is /= 2
+                    Right EvaluationFailure       -> scrut || length is == 2 -- scrut implies length = 2
```
effectfully: Hm. So let's say there's a bug in the evaluator causing it to fail whenever `scrut` is `True`. Type checking succeeds, evaluation fails and you say it's ok? Or am I reading it wrong?

plutus-tx-plugin/test/Budget/9.6/map2.uplc.golden
```diff
@@ -149,11 +150,16 @@
                            (case d [(\d d -> mkPairData d d)])
                            (s s ds)) ])))
          ((\s -> s s)
-            (\s acc ->
-               caseList' acc (\hd -> (\x -> s s x) (force mkCons hd acc)))))
-      (\z f xs ->
-         force
-           (force (force chooseList)
+            (\s acc xs ->
+               case xs [acc, (\hd -> (\x -> s s x) (force mkCons hd acc))])))
```
effectfully, line 154: Nevermind about inlining `caseList`, it seems to inline perfectly well in the end.

plutus-tx-plugin/test/BuiltinList/Budget/9.6/and.eval.golden
```diff
@@ -1,6 +1,6 @@
-CPU:                 1_827_714
-Memory:                  7_992
-Term Size:                  81
-Flat Size:                  76
+CPU:                   752_100
+Memory:                  4_800
+Term Size:                  66
+Flat Size:                  66
```
effectfully, line 4: Cool.

plutus-tx-plugin/test/BuiltinList/Budget/9.6/drop.uplc.golden
```diff
@@ -10,17 +10,13 @@
...
-                                       `$dMkNil`
-                                       (subtractInteger n 1)
-                                       xs)
-                                    (force headList l)
-                                    (force tailList l)))))
+                      [ (case
+                           l
+                           [ `$dMkNil`
+                           , (\ds xs ->
+                                force (force (drop (delay (\x -> x))))
```
effectfully, line 17: "force force drop delay id" is a piece of modern art.

plutus-tx-plugin/test/BuiltinList/Budget/9.6/lastOk.eval.golden
```diff
@@ -1,6 +1,6 @@
-CPU:                19_055_427
-Memory:                 84_856
-Term Size:                 119
-Flat Size:                 122
+CPU:                 4_736_100
+Memory:                 29_700
+Term Size:                  86
+Flat Size:                  96
```
effectfully, line 4: Impressive.

effectfully:
> Do we currently have configurable budgeting mechanism for casing? I haven't seen any budget related code on CEK for casing.

We don't, it's just the regular cost of a single CEK step.

cardano-constitution/test/Cardano/Constitution/Validator/GoldenTests/unsorted.pir.golden
effectfully: [Nevermind](https://github.com/IntersectMBO/plutus/pull/7188#discussion_r2198804689).

plutus-benchmark/script-contexts/test/V2/9.6/dataFwdStakeTrickManual.eval.golden
SeungheonOh: probably because the script never needed to match on list. This is before fully evaluating term so makes sense

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Definition.hs
SeungheonOh: Type checker is checking is correct number of branches and correct type for branches. If type checking passes but evaluation fails then it means that's matching True with single branch.

SeungheonOh: Oh wait, I think I messed up

SeungheonOh: oh wait, I think I messed up. It should be ` scrut && length is /= 2`

effectfully: Then it'll say that `scrut` cannot be `False`, which it obviously can.

effectfully: > If type checking passes but evaluation fails then it means that's matching True with single branch.

No it doesn't, because maybe I implemented casing on lists wrong and it just always throws on `True` or something.

cardano-constitution/test/Cardano/Constitution/Validator/GoldenTests/unsorted.pir.golden
SeungheonOh: Yeah, can confirm, I checked too, no changes

SeungheonOh:
Should be good to go now

SeungheonOh:
Performance improvement here is pretty sweat, but multi-let will improve it much more. Since multi-bind will remove the need for matching on list repeatedly which will be a massive improvement.  

Alternatively, we can have annotated case matching which would be also very useful.

nix/shell.nix
```diff
@@ -69,14 +69,15 @@ let
...
+    pkgs.ghcid
     agda-tools.agda
     agda-tools.agda-with-stdlib
     agda-tools.agda-mode
 
     metatheory.generate-malonzo-code
     metatheory.agda-with-stdlib-and-metatheory
 
-    r-with-packages
+    # r-with-packages
```
effectfully: What's going on here?

plutus-benchmark/casing/src/PlutusBenchmark/Casing.hs
```diff
@@ -0,0 +1,100 @@
...
+    fromRight (Prelude.error "debruijnTermUnsafe") . runExcept @UPLC.FreeVariableError . UPLC.deBruijnTerm
+
+nonMatchingBranch :: TermLike term tyname name UPLC.DefaultUni UPLC.DefaultFun => term ()
+nonMatchingBranch = constant () $ someValueOf DefaultUniInteger $ -1
+
+-- Note that we don't need to generate casings for the none maching branches because
+-- only matching branches get executed.
+-- | Generate a term that does a lot of casing on boolean
+casingBool :: Integer -> Term
+casingBool 0 = constant () $ someValueOf DefaultUniInteger 42
```
effectfully, line 25: Great benchmarks.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
effectfully: `headSpineOpaque` is no longer needed at all?

plutus-core/plutus-core/src/PlutusCore/TypeCheck/Internal.hs
```diff
@@ -576,14 +576,12 @@ inferTypeM (Case ann resTy scrut branches) = do
             -- scrutinee does not have a SOP type with the right number of alternatives
             -- for the number of branches
             Nothing -> throwError (TypeMismatch ann (void scrut) expectedSop vScrutTy)
-        TyBuiltin _ someUni -> case annotateCaseBuiltin someUni branches of
+        ty -> case annotateCaseBuiltin ty branches of
```
effectfully: Please rename it to `vTy` to indicate that it's normalized (you can see all these comments about normalization, it's a consequence of us originally having trouble ensuring that normalization happens in the right places).

And update the PIR type checker too.

plutus-core/untyped-plutus-core/test/Evaluation/Golden/caseNonTag.type.golden
```diff
@@ -1,7 +1 @@
-(Left Type mismatch at ()
-Expected a term of type
-  '(sop)'
-But found one of type
-  '(fun (con integer) (fun (con integer) (con integer)))'
-Namely,
-  '(builtin addInteger)')
\ No newline at end of file
+(Left Unsupported 'case' of a value of a built-in type at (): (fun (con integer) (fun (con integer) (con integer))) isn't supported in 'case')
```
effectfully: Make it readable again?

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Definition.hs
```diff
@@ -1277,6 +1278,47 @@ test_Case =
...
+                    xs <- freshName "xs"
+                    let
+                      listElem = mkTyBuiltin @_ @Integer ()
+                      list = mkTyBuiltin @_ @[Integer] ()
+                    pure $
+                      kase ()
+                        listElem
+                        (mkConstant () scrut)
+                        [ mkConstant @Integer () 42
+                        , lamAbs () x listElem $ lamAbs () xs list $ var () x
```
effectfully, line 1296: We should rename `Bool success` to `Bool success 2` and add `Bool success 1` for when there's only one branch. And do the same here. Because `Bool any` doesn't properly cover the case with a single branch and the `False` scrutinee.

But that can be a separate PR.

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Definition.hs
```diff
@@ -1277,6 +1278,47 @@ test_Case =
...
+                    x <- freshName "x"
+                    xs <- freshName "xs"
+                    let
+                      listElem = mkTyBuiltin @_ @Integer ()
+                      list = mkTyBuiltin @_ @[Integer] ()
+                    pure $
+                      kase ()
+                        listElem
+                        (mkConstant () scrut)
+                        [ lamAbs () x listElem $ lamAbs () xs list $ var () x
```
effectfully, line 1316: This doesn't cover the case when there's more branches than expected.

It's also out of sync with the `Bool` tests, which have very similar constraints. I do think it's a good idea to have the two that you have there, but then we should also

1. add a test for when there's more branches than expected
2. structure the `Bool` tests the same way

I felt like `Bool any` was an OK way to merge what you have here and the too-many-branches test into a single test, but I never actually measured coverage, so if you want to split it into two, that's even better I guess.

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Definition.hs
```diff
@@ -1251,8 +1251,9 @@ test_Case =
                             (mkConstant () scrut)
                             (map (mkConstant ()) is)
                 in case typecheckEvaluateCekNoEmit def defaultBuiltinCostModelForTesting term of
-                    Left _                        -> False
-                    Right EvaluationFailure       -> length is /= 2 && (scrut || length is /= 1)
+                    Left _                        -> length is /= 1 && length is /= 2
+                    -- Bool casing only fails if there's one branch and its trying to case True.
+                    Right EvaluationFailure       -> scrut && length is /= 2
```
effectfully, line 1256: This still looks wrong to me. OK, I'll try it out myself.

effectfully (approved):

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Definition.hs
effectfully: Nope, it's right, I was wrong.

nix/shell.nix
SeungheonOh: Oh, oops. Turns out apple decided to change linker behavior. It broke R packages. This is what I had to do to make it build dev shell.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
SeungheonOh: Yes, with the removal of `CaseData` and `CaseList`. I will remove

SeungheonOh: I think it's okay to leave this actually. This can be useful future for adding other builtins

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Definition.hs
SeungheonOh: I tested coverage, and it was pretty okay. Regardlessly, I think it's better to have separate cases.

PR #7189, opened 2025-07-07
Force ambiguity for annotations in CEK

SeungheonOh:
For Cek, we intentionally maintain ambiguity of the annotation type for performance reasons(I assume). This PR makes it so that this ambiguity is forcefully enforced by using `GHC.Exts.Any` as the annotation type. This reduces some noise on the types while also allowing construction of UPLC nodes within CEK in nicer way, which is what I needed for [this](https://github.com/IntersectMBO/plutus/pull/7188/files#r2191103734) 

SteppableCek is a bit messy because its context can hold annotations, so it would require more coercsions.

### DIFF
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
@@ -60,6 +60,8 @@ module UntypedPlutusCore.Evaluation.Machine.Cek.Internal
     , Slippage
     , defaultSlippage
     , NTerm
+    , forgetAnn
+    , rememberAnn
     , runCekM
     )
 where
@@ -101,10 +103,12 @@ import Data.Semigroup (stimes)
 import Data.Text (Text)
 import Data.Vector qualified as V
 import Data.Word
+import GHC.Exts (Any)
 import GHC.Generics
 import GHC.TypeLits
 import Prettyprinter
 import Universe
+import Unsafe.Coerce (unsafeCoerce)
 
 {- Note [Compilation peculiarities]
 READ THIS BEFORE TOUCHING ANYTHING IN THIS FILE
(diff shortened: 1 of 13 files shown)


SeungheonOh:
/benchmark lists

it shouldn't differ really. core didn't change

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -163,6 +167,12 @@ this can make a surprisingly large difference.
 -- 'Name' is not necessary but we leave it here for simplicity and debuggability.
 type NTerm uni fun = Term NamedDeBruijn uni fun
 
+forgetAnn :: forall ann uni fun. NTerm uni fun ann -> NTerm uni fun Any
+forgetAnn = unsafeCoerce
+
+rememberAnn :: forall ann uni fun. NTerm uni fun Any -> NTerm uni fun ann
```
SeungheonOh, line 173: This is unsafe

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/SteppableCek/Internal.hs
```diff
@@ -328,8 +329,8 @@ cekTrans :: forall uni fun ann s
            . (ThrowableBuiltins uni fun, GivenCekReqs uni fun ann s)
            => CekTrans uni fun ann s
 cekTrans = \case
-    Starting term          -> pure $ Computing NoFrame Env.empty term
-    Computing ctx env term -> computeCek ctx env term
+    Starting term          -> pure $ Computing NoFrame Env.empty (forgetAnn term)
+    Computing ctx env term -> computeCek ctx env (rememberAnn term)
```
SeungheonOh, line 333: This is only safe if `Computing ...` is constructed by `cekTrans` which is not ideal. Probably, it is possible to have some witness value to allow this conditionally

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -163,6 +167,12 @@ this can make a surprisingly large difference.
...
+rememberAnn = unsafeCoerce
```
effectfully, line 174: Ugh, I'm really uncomfortable about that. Let's discuss it during the tech meeting.

effectfully (changes requested):
Hate it.

SeungheonOh:
I agree. Also, less reason to pursue this as there's a proper way for #7188. Closing

PR #7196, opened 2025-07-09
[Optimization] Push 'force' into 'chooseList' and 'chooseData'

effectfully:
Currently we only push `force` into `ifThenElse`. This PR makes it so we push `force` into `chooseList` and `chooseData` as well. Somehow that doesn't seem to make any difference whatsoever. Did I implement it wrong? Does it really not matter? Unclear.

I could've added some tests, but this all is rather insignificant and I'm inclined to just remove `chooseData`-based `caseData` as per [this](https://github.com/IntersectMBO/plutus/pull/7190#discussion_r2193686295) comment and not worry about `chooseList`, because we're gonna have casing on lists soon anyway, which makes `chooseList` pretty much obsolete.


SeungheonOh:
This is no longer relevant since we removed `CaseData` and `CaseList` altogether and since we don't have to deal with this with `case`ing

effectfully:
`chooseList` and `chooseData` are still a thing, so this is still relevant, but I do believe it's insignificant and we can just forget about it.

PR #7197, opened 2025-07-09
PlutusLedgerApi.Envelope module and tests for JSON envelope generation

Unisay:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


colll78 (approved):
Perfect, thank you! This will help a lot of users, as I can refer them to this function for exporting their plutus scripts.

cabal.project
```diff
@@ -81,9 +81,3 @@ allow-newer:
   , inline-r:containers
   , inline-r:primitive
 
--- Uncomment the following lines to make cardano-constitution:create-json-envelope buildable:True
---
--- package cardano-constitution
---   flags: +force-build
--- allow-newer: *:plutus-ledger-api
--- allow-older: *:nothunks
```
SeungheonOh, line 93: Awesome!

plutus-ledger-api/src/PlutusLedgerApi/Envelope.hs
```diff
@@ -0,0 +1,71 @@
...
+  -- ^ Description of the code
+  -> CompiledCode a
+  -- ^ Compiled code to wrap in the envelope
+  -> Json.Value
+  -- ^ JSON envelope
+compiledCodeEnvelope desc code =
+  Json.object ["type" .= typ, "description" .= desc, "cborHex" .= hex]
+ where
+  typ :: Text
+  typ = "PlutusScriptV3"
```
SeungheonOh: Supporting other ledger version may be a good idea. We can use `PlutusLedgerApi.Common.Versions.PlutusLedgerLanguage` as an extra parameter.

plutus-ledger-api/src/PlutusLedgerApi/Envelope.hs
```diff
@@ -0,0 +1,71 @@
...
+  "type": "PlutusScriptV3",
+  "description": "A description of the code",
+  "cborHex": "..."
+}
+@
+-}
+compiledCodeEnvelope
+  :: Text
+  -- ^ Description of the code
+  -> CompiledCode a
```
SeungheonOh, line 37: Do you know if envelope requires the term to be in specific type or does it allow the encoded term to be of any type? 

I don't think we really need to enforce anything necessarily here, but it would be something nice to know.

Unisay: @colll78 maybe you can answer this question?
I am not aware of any envelope standard or spec.

plutus-ledger-api/src/PlutusLedgerApi/Envelope.hs
Unisay: I assessed the probability of a user having the new `plutus-ledger-api` (in which this functionality is) and outdated (V2) `plutus-tx-plugin` (as a source of `CompiledCode` value) as low enough to justify hardcoding `type = "PlutusScriptV3"`.

SeungheonOh: It can be the case someone needs V1 or V2 script specifically. For any new projects, we can safely assume they will use latest ledger api. However, when someone needs to make changes to existing protocols that was written in V1 or V2 ledger api, one would need to produce script of matching version because interoperability between different script versions are quite messy.

plutus-ledger-api/src/PlutusLedgerApi/Envelope.hs
colll78: The `cborHex` should be in the same encoding format as produced by `writeFileTextEnvelope` from `Cardano.Api` which from what I see is how it works in this PR.  That is what all the offchain frameworks expect. Other than that, I don't think there are any other restrictions.

SeungheonOh (approved):

plutus-ledger-api/src/PlutusLedgerApi/Envelope.hs
SeungheonOh: Honestly, this is something we can just patch up later someone find this troublesome

colll78 (approved):

plutus-ledger-api/src/PlutusLedgerApi/Envelope.hs
Unisay: You also conviced me that people might want to use it with V1 or V2 so I added a function that allows to speicfy version without removing the one using V3 as default.

PR #7201, opened 2025-07-10
[Benchmark] Remove preparation noise from the 'validation' benchmarks

effectfully:
This fixes the `validation` benchmarks unintentionally counting in preparation time (deserialization + conversion to `NamedDeBruijn`) and partially addresses the issue with `validation` benchmarks that I ran into in #7177.

### DIFF
--- a/plutus-benchmark/validation/bench/Common.hs
+++ b/plutus-benchmark/validation/bench/Common.hs
@@ -132,7 +132,7 @@ benchWith act = do
     mkScriptBM dir file =
         let readAndPrep = do
                 script <- BS.readFile (dir </> file)
-                pure (script, UPLC.Error ()) -- toNamedDeBruijnTerm . UPLC._progTerm $ unsafeUnflat file script)
+                pure (script, toNamedDeBruijnTerm . UPLC._progTerm $ unsafeUnflat file script)
         in env readAndPrep $ \(~scriptAndTerm) ->
             -- We use 'uncurry', because for whatever reason lazy matching on a tuple doesn't work
             -- when @Strict@ is enabled.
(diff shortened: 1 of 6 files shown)


plutus-benchmark/validation/bench/Common.hs
```diff
@@ -122,8 +124,9 @@ benchWith act = do
 
     mkScriptBM :: FilePath -> FilePath -> Benchmark
     mkScriptBM dir file =
-        env (BS.readFile $ dir </> file) $ \(~scriptBS) ->
-            bench (dropExtension file) $ act file scriptBS
+        let prep = toNamedDeBruijnTerm . UPLC._progTerm . unsafeUnflat file
+        in env (prep <$> BS.readFile (dir </> file)) $ \(~term) ->
+            bench (dropExtension file) $ act term
```
effectfully: Note that this doesn't address the issue of `term` being deep-forced by `benchTermCek`, which also is unnecessarily counted in.

Unisay: Do you intend to address the issue in this PR ?

effectfully: No, that's for #7201. I'll also write documentation there, this PR is just to fix the very specific issue of the `validation` benchmarks.

effectfully:
validation-agda-cek

effectfully:
This turned out to be way harder than I expected. Unfortunately, it seems to be above my performance grade ("meeting expectations") and I do not feel like I'm capable of taking on such a complex task. I believe a person with a better performance grade is warranted here.

Closing this PR, since I'm sure a person with a better performance grade will want to run their own investigation and write superior code, matching their elevated effort level.

PR #7204, opened 2025-07-11
Add Strict to nofib examples

kwxm:
The nofib examples were written a number of years ago.  We're now recommending that Plinth programs should use `{-# LANGUAGE Strict #-}` but they don't include it. I just want to see if that makes any difference.


kwxm:
Looks as if this doesn't make a great deal of difference.  We should probably review all of the benchmarks at some time to make sure that we're doing the right thing.

PR #7207, opened 2025-07-14
Bug in plutus-tx traces unused code

zeme-wana:
An unused binding in a validator's `where` clause is surprisingly executed at run time.
The `noCollateralTaken` binding in Validator.hs on line 153 is unused, and yet when evaluating the validator, it is executed. 
In addition, when uncommenting the real definition of `noCollateralTaken` evaluation fails with an unknown error.
To reproduce, run `cabal run plutus-benchmark:cardano-loans`


Unisay:
Not a bug: in a strict setting this unused `where` binding is evaluated as if it is a let binding:
```haskell
foo = unitval
  where
    _boom = traceError "boom"
```
is equivalent to:
```haskell
foo =
  let 
    !_boom = traceError "boom"
   in 
    unitval
```

If the binding `noCollateralTaken` is promoted to the top level then its not evaluated anymore.

PR #7209, opened 2025-07-14
[Builtin] Add "dumb" `Data` matching

SeungheonOh:
part of #6602 

Here we don't do any elaborate casing on `Constr` constructor of `Data`; instead, it will simply case on each constructors of `Data`. This might not be useful since this style of casing isn't used frequently in real-world script.

Feel free to close this PR, once we have smarter/more useful casing on Constr.


effectfully (commented):
This all is well-implemented, but I believe we shouldn't do it.

We had a [discussion](https://input-output-rnd.slack.com/archives/G01EE4NQ9U0/p1748536241753669?thread_ts=1748527592.665449&cid=G01EE4NQ9U0) with Philip DiSarro on Slack. My question to him was

> I'm asking because if `chooseData` isn't particularly useful, then neither `caseData` will be and so instead of worrying about the latter I should just implement proper casing on `Data` using the `Case` constructor of the AST

And his answer was

> Yes that would be much more impactful from my perspective.

He also said

> Either way, this is a very small edge case relative to unwrapping expected constructor. `chooseData` is rarely ever used.

Basically, if `chooseData` is rarely ever used and people just unwrap the expected constructor (`I`, `List` etc), then casing on `Data` like implemented here isn't very useful and so we should do the more useful alternative of casing on `Constr` specifically to mimic the behavior of SOPs and approximate their efficiency, which is something that is definitely going to be used a lot on the chain.

PR #7210, opened 2025-07-15
[UPLC] [Optimization] Improve case-of-case

effectfully:
Makes UPLC case-of-case a bit less restrictive by allowing it to produce duplicate branches if they're small enough. Doesn't seem to have any effect on anything (provided I implemented it right, there's no tests), so is probably not worth it.

### DIFF
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
@@ -34,7 +34,7 @@ into
 
 This is always an improvement.
 -}
-module UntypedPlutusCore.Transform.CaseOfCase (caseOfCase) where
+module UntypedPlutusCore.Transform.CaseOfCase (caseOfCase, processTerm, annotateIsDuplicatedOn) where
 
 import PlutusPrelude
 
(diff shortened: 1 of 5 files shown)


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/CaseOfCase.hs
```diff
@@ -88,11 +96,13 @@ processTerm = \case
...
+        constrsDupl <- fmap (annotateIsDuplicatedOn fst) . for altsInner $ \case
           c@(Constr _ i _)   -> Just (Left i, c)
           c@(Constant _ val) -> Just (Right val, c)
           _                  -> Nothing
         -- See Note [Case-of-case and duplicating code].
-        guard $ length (nub . toList $ fmap fst constrs) == length constrs
-        pure $ constrs <&> \(_, c) -> CaseReduce.processTerm $ Case annOuter c altsOuter)
+        for constrsDupl $ \((_, c), dupl) -> do
+            let alt = CaseReduce.processTerm $ Case annOuter c altsOuter
+            guard $ not dupl || termSize alt <= 3
```
effectfully, line 106: Tried making it `300`, no effect either.

effectfully:
Zero impact on existing tests => doesn't make much sense to investigate it further. Practically useless for a IOE Team Success Meeting review.

PR #7211, opened 2025-07-15
[Builtins] Remove 'HeadSpine' from 'BuiltinRuntime'

effectfully:
This removes `HeadSpine` from `BuiltinRuntime` now that we don't need it for pattern matching builtins (because they're removed).


SeungheonOh (approved):
Good changes. 

Now that `HeadSpine` is not used elsewhere, is it okay for me to add extra "empty" constructor to `HeadSpine` for discussed optimization?

effectfully:
Yep. Except it should carry some error, not just be empty. If I understood you right.

SeungheonOh:
Good. Thanks

PR #7214, opened 2025-07-15
Remove caseData and caseList builtins from specification

kwxm:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1670

This removes the `caseList` and `caseData` builtins from the PLC specification to reflect #7190.  These were added to the specification in #6895, which required reworking the specification significantly to allow builtins to return "deferred applications" (`HeadSpine` objects) of the form `(v|v₁,...,vₖ)`.  These are no longer required and may be removed in a separate PR to make it easier to restore them later if required.


kwxm:
Stuck in CI.

Issue #7213, opened 2025-07-15
Conformance, End-to-End testing, and specification changes for new builtin casing


SeungheonOh:
`Case`ing on builtin were added on #7029 and #7188. We need to 

- Update spec
- Add conformance testing for casing
- Add End-to-End tests
- Update Metatheory

PR #7218, opened 2025-07-16
More tests, conformance test cases for builtin integer, bool, list casing

SeungheonOh:
part of #7213 

Make property test slightly clearer. Adds conformance tests for builtin integer, bool, and list.


plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Definition.hs
Unisay, line 1: Not related to your task, just an observation:
it seems wrong to me that a package named `testlib` contains actual test cases. I'd expect a `testlib` to contain functionality intended to help test authors, not the actual tests, analogous to how i.e. `tasty` library doesn't ship tests but a functionality to help building them.

kwxm:
This fails if I try to build it locally.  Eg

```
testlib/PlutusCore/Generators/QuickCheck/Builtin.hs:435:21: error:
    Ambiguous occurrence ‘Some’
    It could refer to
       either ‘Universe.Some’,
              imported from ‘Universe’ at testlib/PlutusCore/Generators/QuickCheck/Builtin.hs:36:1-15
              (and originally defined in ‘some-1.0.6:Data.Some.Newtype’)
           or ‘Test.QuickCheck.Some’,
              imported from ‘Test.QuickCheck’ at testlib/PlutusCore/Generators/QuickCheck/Builtin.hs:33:1-22
              (and originally defined in ‘Test.QuickCheck.Monoids’)
    |
435 | instance Arbitrary (Some (ValueOf DefaultUni)) where
    |                     ^^^^
```

That's fixed by `import Test.QuickCheck hiding (Some)`.  I'm not sure why it's passing the CI tests though.

[Later] No, ignore that.  The same thing happened in some completely unrelated files and I eventually managed to fix it by restarting nix (and also `git pull` and `cabal update`).  I've no idea what was going on there.

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Definition.hs
kwxm: I was a bit confused by the difference between `test` and `testlib` recently: maybe that's why. That's unrelated to this PR, but maybe we should remember to think about it.

SeungheonOh: Yes, I agree. we shouldn't have test cases in `testlib`

SeungheonOh:
I'll add suggested cases on the next PR for constant casing conformance tests which we will need soon for casing on unit, pair, and data.

plutus-conformance/test-cases/uplc/evaluation/term/constant-case/bool/bool-01/bool-01.uplc
```diff
@@ -0,0 +1,4 @@
+-- case on false
+(program 1.1.0
+    (case (con bool False) (con integer 0) (con integer 1))
+)
```
effectfully, line 4: The choice of constants is kinda unfortunate: what if `case` simply converts the given boolean to the respective integer instead of going into the right branch? It'd be better to make it more random, like return 69 and 42 or whatever.

plutus-conformance/test-cases/uplc/evaluation/term/constant-case/bool/bool-04/bool-04.uplc.budget.expected
```diff
@@ -0,0 +1 @@
+evaluation failure
```
effectfully, line 1: That's weird. A failing program still consumes budget and we should report it. I do of course understand that it's not a problem with this PR.

effectfully (commented):
Great stuff, thank you.

PR #7221, opened 2025-07-17
Add constant casing for builtin unit and pair

SeungheonOh:
part of #6602


SeungheonOh:
I made changes to plugin so that it will use constant casing for `FstPair`, `SndPair`, and `ChooseUnit`. This have improved execution budget around 10% to 20%; however, introduced rather impressive ~30% slowdown.

plutus-tx-plugin/test/Budget/9.6/matchAsDataE.pir.golden
```diff
@@ -18,8 +18,8 @@
            Tuple2
              {integer}
              {b}
-             (fstPair {integer} {b} tup)
-             (sndPair {integer} {b} tup))
+             (case integer tup [(\(l : integer) (r : b) -> l)])
+             (case b tup [(\(l : integer) (r : b) -> r)]))
```
effectfully: There's of course the issue that we want to only `case` on `tup` once, i.e. we currently generate sub-optimal code.

Not for this PR though, should be addressed in a follow-up.

plutus-tx-plugin/test/Plugin/Strict/9.6/strictPair.pir.golden
```diff
@@ -1 +1,2 @@
-\(p : pair integer integer) -> fstPair {integer} {integer} p
\ No newline at end of file
+\(p : pair integer integer) ->
+  case integer p [(\(l : integer) (r : integer) -> l)]
```
effectfully: Well, I guess it's not entirely unreasonable that calling a builtin once can be faster than evaluating a `Case` node, two function applications and a variable lookup.

effectfully: Especially given that a builtin does something very similar to what a `case` does, except with less noise and with better optimizations.

SeungheonOh: Yeah, and cost model doesn't reflect this correctly. The cost might only even out when we compare taking both values from pair, running `FstPair` and `SndPair` both.

I don't think we can replace `FstPair` and `SndPair`

SeungheonOh:
Okay, we just got trolled by benchmark machine

plutus-benchmark/casing/bench/Bench.hs
```diff
@@ -19,7 +19,13 @@ import Data.Functor
...
 benchmarks ctx =
     [ bgroup "casing"
-      [mkBMs "bool" Casing.casingBool
+      [ mkBMs "pairFstSnd" Casing.pairFstSnd
+      , mkBMs "pairCasing" Casing.pairCasing
+      , mkBMs "chooseUnit" Casing.chooseUnit
+      , mkBMs "unitCasing" Casing.unitCasing
+      , mkBMs "headList" Casing.headList
+      , mkBMs "headListCasing" Casing.headListCasing
+      , mkBMs "bool" Casing.casingBool
```
effectfully: These are not backported to `master`, right? So the whole benchmarking result is meaningless.

SeungheonOh: No, the point is to compare builtin version of fst/snd pair to the casing version of it. Same with HeadList and ChooseUnit

SeungheonOh:
Benchmark was to show pair fst/snd via casing vs builtin, here are results

| -                 | builtin     | casing       | percent change   |
| ----------------- | ----------- | ------------ | ---------------- |
| pairFstSnd2000    | 0.5482 ms   | 0.4223 ms    | -22.97%          |
| pairFstSnd4000    | 1.204 ms    | 0.9932 ms    | -17.51%          |
| pairFstSnd6000    | 2.017 ms    | 1.774 ms     | -12.05%          |
| pairFstSnd8000    | 2.901 ms    | 2.617 ms     | -9.79%           |
| pairFstSnd10000   | 3.841 ms    | 3.513 ms     | -8.54%           |
| pairFstSnd12000   | 4.675 ms    | 4.422 ms     | -5.41%           |
| ----------------- | ----------- | ------------ | ---------------- |
| chooseUnit2000    | 0.2204 ms   | 0.04756 ms   | -78.42%          |
| chooseUnit4000    | 0.4926 ms   | 0.09568 ms   | -80.58%          |
| chooseUnit6000    | 0.8098 ms   | 0.1426 ms    | -82.39%          |
| chooseUnit8000    | 1.194 ms    | 0.1906 ms    | -84.04%          |
| chooseUnit10000   | 1.681 ms    | 0.2371 ms    | -85.90%          |
| chooseUnit12000   | 2.263 ms    | 0.2815 ms    | -87.56%          |
| ----------------- | ----------- | ------------ | ---------------- |
| headList2000      | 0.2384 ms   | 0.1947 ms    | -18.33%          |
| headList4000      | 0.4950 ms   | 0.4162 ms    | -15.92%          |
| headList6000      | 0.7753 ms   | 0.6706 ms    | -13.50%          |
| headList8000      | 1.080 ms    | 0.9773 ms    | -9.51%           |
| headList10000     | 1.428 ms    | 1.350 ms     | -5.46%           |
| headList12000     | 1.811 ms    | 1.770 ms     | -2.26%           |

So we do get a meaningful improvements in CEK performance as well. 

ps. I forgot to click post button 🙃

SeungheonOh:
```
art

**Comment by SeungheonOh** [comment]
| name             | using builtin | using casing | d%      |
|------------------|---------------|--------------|---------|
| pairFstSnd/2000  | 326.7 us      | 241.8 us     | -25.99% |
| pairFstSnd/4000  | 711.5 us      | 559.3 us     | -21.39% |
| pairFstSnd/6000  | 1.207 ms      | 982.7 us     | -18.58% |
| pairFstSnd/8000  | 1.727 ms      | 1.409 ms     | -18.41% |
| pairFstSnd/10000 | 2.278 ms      | 1.843 ms     | -19.10% |
| pairFstSnd/12000 | 2.795 ms      | 2.254 ms     | -19.36% |
| chooseUnit/2000  | 141.5 us      | 24.28 us     | -82.85% |
| chooseUnit/4000  | 311.5 us      | 48.23 us     | -84.52% |
| chooseUnit/6000  | 506.4 us      | 71.87 us     | -85.81% |
| chooseUnit/8000  | 730.5 us      | 95.73 us     | -86.89% |
| chooseUnit/10000 | 1.002 ms      | 119.9 us     | -88.04% |
| chooseUnit/12000 | 1.341 ms      | 141.8 us     | -89.43% |
| headList/2000    | 143.8 us      | 116.0 us     | -19.37% |
| headList/4000    | 302.6 us      | 237.1 us     | -21.67% |
| headList/6000    | 466.7 us      | 383.4 us     | -17.84% |
| headList/8000    | 655.3 us      | 553.8 us     | -15.49% |
| headList/10000   | 877.3 us      | 750.8 us     | -14.42% |
| headList/12000   | 1.090 ms      | 969.3 us     | -11.08% |

Ran benchmark again locally, this compares implementation of `headList`, `sndPair`, and `fstPair` based on builtin casing to that based on builtins.

File: cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests/sorted.uplc.golden [hunk 4]
```diff
@@ -37,689 +37,693 @@ program
...
+                                                                                                                                         (unConstrData
+                                                                                                                                            ((\tup ->
+                                                                                                                                                case
+                                                                                                                                                  (equalsInteger
+                                                                                                                                                     5
+                                                                                                                                                     (case
+                                                                                                                                                        tup
```
effectfully: Interesting, so case-of-case can be generalized such that it transforms, say,

```
case (nullList (case tup (\x y -> x : xs)))
  falseAlt
  trueAlt
```

into

```
case tup
  \x y ->
    case nullList (x : xs)
      falseAlt
      trueAlt
```

which then becomes

```
case tup
  \x y -> falseAlt
```

This wouldn't help here, but may or may not be a useful optimization in general.

Actually, we don't even need to care about the outer `caseList`, we just need to flip `nullList` and the inner case. Which is kinda what we do for `ifThenElse`. So yeah, it makes sense `nullList` is itself a glorified `case`.

And it works the other way too: case-of-nullList is as interesting to optimize as nullList-of-case. Obviously applies to other builtins like `headList` and `fstPair` etc. We do already handle `ifThenElse` (even though we don't actually generate it anymore), but only in one direction.

cardano-constitution/test/Cardano/Constitution/Validator/GoldenTests/unsorted.uplc.golden
```diff
@@ -912,8 +918,8 @@ program
                                     [ (\x xs ->
                                          constr 1
                                            [ (constr 0
-                                                [ (force (force fstPair) x)
-                                                , (force (force sndPair) x) ])
+                                                [ (case x [(\l r -> l)])
+                                                , (case x [(\l r -> r)]) ])
```
effectfully: That should have a single `case` with `constr` inside of it. Do you happen to have any idea where we generate this code?

plutus-benchmark/cardano-loans/test/9.6/main.uplc.golden
```diff
@@ -6150,14 +6215,16 @@
...
+                                                                            (case
+                                                                               tup
+                                                                               [ (\l
+                                                                                   r ->
+                                                                                    r) ]))
+                                                                         (case
+                                                                            tup
+                                                                            [ (\l
+                                                                                r ->
+                                                                                 l) ]))
```
effectfully: I feel like we should cache these two somehow? They appear quite a few times in the file.

plutus-benchmark/casing/src/PlutusBenchmark/Casing.hs
```diff
@@ -98,3 +104,114 @@ casingListOneBranch i = debruijnTermUnsafe $ go i arg
...
+-- will case first and then second term on each iteration.
+pairFstSnd :: Integer -> Term
+pairFstSnd i =
+  debruijnTermUnsafe $
+    foldr (const comp) (constant () $ someValueOf DefaultUniInteger 0) [1..i]
+  where
+    intTy = PLC.mkTyBuiltin @_ @Integer ()
+    pairVal =
+      constant () $
+        someValueOf (PLC.DefaultUniPair DefaultUniInteger DefaultUniInteger) (42, 42)
```
effectfully: You can do `mkConstant @(Integer, Integer) () (42, 42)` (not that you need to).

plutus-benchmark/casing/src/PlutusBenchmark/Casing.hs
```diff
@@ -98,3 +104,114 @@ casingListOneBranch i = debruijnTermUnsafe $ go i arg
...
+        (kase () intTy pairVal [lamAbs () fstL intTy $ lamAbs () fstR intTy $ var () fstR])
+
+-- | Generate a term that does a lot of casing on unit using 'ChooseUnit'.
+chooseUnit :: Integer -> Term
+chooseUnit i =
+  debruijnTermUnsafe $
+    foldr (const comp) (constant () $ someValueOf DefaultUniInteger 0) [1..i]
+  where
+    intTy = PLC.mkTyBuiltin @_ @Integer ()
+    unitVal = constant () $ someValueOf PLC.DefaultUniUnit ()
```
effectfully: And that would be just `mkConstant () ()`.

plutus-benchmark/script-contexts/test/V1/9.6/checkScriptContext1.pir.golden
```diff
@@ -47,8 +47,10 @@
            (d : data) ->
             let
               !tup : pair integer (list data) = unConstrData d
-              !index : integer = fstPair {integer} {list data} tup
-              !args : list data = sndPair {integer} {list data} tup
+              !index : integer
+                = case integer tup [(\(l : integer) (r : list data) -> l)]
+              !args : list data
+                = case (list data) tup [(\(l : integer) (r : list data) -> r)]
```
effectfully: That's one place where instead of binding those two separately, we should just pull the `case` to the outside and bind `index` and `args` in its branch. There's many more of these below.

If you grep for `BI.fst`, you'll find plenty of places where it appears together with `BI.snd`. Turning those into a single case would be a good optimization.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -532,22 +532,31 @@ outOfBoundsErr x branches = fold
 
 instance AnnotateCaseBuiltin DefaultUni where
     annotateCaseBuiltin ty branches = case ty of
+        TyBuiltin _ (SomeTypeIn DefaultUniUnit)    ->
+          Right $ map (, []) branches
```
effectfully: We don't check that there's only one branch?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -532,22 +532,31 @@ outOfBoundsErr x branches = fold
...
             [f, t] -> Right $ [(f, []), (t, [])]
-            _      -> Left $ "Casing on bool requires exactly one branch or two branches"
+            _      -> Left "Casing on bool requires exactly one branch or two branches"
         TyBuiltin _ (SomeTypeIn DefaultUniInteger) ->
           Right $ map (, []) branches
         listTy@(TyApp _ (TyBuiltin _ (SomeTypeIn DefaultUniProtoList)) argTy) ->
           case branches of
             [cons]      -> Right [(cons, [argTy, listTy])]
             [cons, nil] -> Right [(cons, [argTy, listTy]), (nil, [])]
             _           -> Left $ "Casing on list requires exactly one branch or two branches"
```
effectfully: Remove the dollar here as well then.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -532,22 +532,31 @@ outOfBoundsErr x branches = fold
...
+        (TyApp _ (TyApp _ (TyBuiltin _ (SomeTypeIn DefaultUniProtoPair)) lTyArg) rTyArg) ->
+          case branches of
+            [f] -> Right [(f, [lTyArg, rTyArg])]
+            _   -> Left "Casing on pair requires exactly one branch"
         _                 -> Left $ display (() <$ ty) <> " isn't supported in 'case'"
 
 instance CaseBuiltin DefaultUni where
     caseBuiltin someVal@(Some (ValueOf uni x)) branches = case uni of
+        DefaultUniUnit
+          | 0 < len   -> Right $ HeadOnly $ branches Vector.! 0
```
effectfully: Hm, why do we allow any number of branches here, if we only need one?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -568,6 +577,11 @@ instance CaseBuiltin DefaultUni where
                 []       -> Right $ HeadOnly $ branches Vector.! 1
                 (y : ys) -> Right $ headSpine (branches Vector.! 0) [someValueOf ty y, someValueOf uni ys]
             | otherwise            -> Left $ outOfBoundsErr someVal branches
```
effectfully: Hm, I thought we've discussed it, but perhaps it was something else. We should have a `CaseResult` data type that is properly strict and removes the `Either` indirection that we have here.

It'll mean adding a constructor for errors to `HeadSpine`, since we only use that for casing on constants nowadays anyway.

effectfully (approved):
Great stuff, thank you.

I'm only confused why we allow more than 1 branch for casing on units.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
SeungheonOh: I was thinking something else. forgot to change back.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
SeungheonOh: Will make a separate optimization PR

cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests/sorted.uplc.golden
SeungheonOh: > it makes sense nullList is itself a glorified case

Yeah, It can be literally just 
```
case <list>
  (\_ _ -> falseCase)
  trueCase
```

Which I'm not entirely sure if it will be faster than builtin. I'll put this on my list to benchmark. Although it would be easier to keep `nullList` for optimization, I think.

I'm not entirely sure if this will improve anything drastically. IMO, we should wait until we figure out if builtin `nullList` is faster than casing `nullList` and proceed accordingly.

cardano-constitution/test/Cardano/Constitution/Validator/GoldenTests/unsorted.uplc.golden
SeungheonOh:
```hs
  letrec
    ~matchData_go : list (pair data data) -> List (Tuple2 data data)
      = (let
            a = pair data data
          in
          /\r ->
            \(z : r) (f : a -> list a -> r) (xs : list a) -> case r xs [f, z])
          {List (Tuple2 data data)}
          (Nil {Tuple2 data data})
          (\(x : pair data data) (xs : list (pair data data)) ->
             Cons
               {Tuple2 data data}
               (Tuple2
                  {data}
                  {data}
                  (case data x [(\(l : data) (r : data) -> l)])
                  (case data x [(\(l : data) (r : data) -> r)]))
               (matchData_go XS))
```

This snippet is what generates that. This is caused by use of `PlutusTx.Builtin.unsafeDataAsMap`: https://github.com/IntersectMBO/plutus/blob/be3808366ee2bf7b14a6388b3c087584b4532829/cardano-constitution/src/Cardano/Constitution/Validator/Common.hs#L103.

SeungheonOh: fixed

plutus-benchmark/script-contexts/test/V1/9.6/checkScriptContext1.pir.golden
SeungheonOh: done

PR #7222, opened 2025-07-17
Autodelete latest haddock build on new deployments

zeme-wana:
Each haddock build weighs ~500MB.
Every 2 weeks a new version of plutus is released, and with it, a new version of the haddock. 
This PR deletes the oldest haddock build before deploying a new one. 
We keep the last 10 builds.


Unisay (approved):
Looks ok.
I don't quite understand the rate at which haddock builds are added vs. how often they are removed.

PR #7223, opened 2025-07-18
Enable SoPs and all builtins in PlutusV1/V2 at PV11.

kwxm:
This fixes https://github.com/IntersectMBO/plutus-private/issues/1655 and https://github.com/IntersectMBO/plutus-private/issues/1656.

This will enable all builtins in PlutusV1 and PlutusV2 at PV11, and also sums of products.  A protocol parameter update will also be required to update the cost model parameters before the new features become usable on the chain: the required parameters can be obtained using `dump-cost-model-parameters` (see PR #7171).


plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
```diff
@@ -153,44 +104,173 @@ ledgerLanguageIntroducedIn = \case
...
+  , LengthOfArray, ListToArray, IndexArray
+  ]
+
+{-| Given a ledger language, return a map indicating which builtin functions were
+  introduced in which 'MajorProtocolVersion'.  This __must__ be updated when new
+  builtins are added.  It is not necessary to add entries for protocol versions
+  where no new builtins are added.  See Note [New builtins/language versions and
+  protocol versions]
 -}
+builtinsIntroducedIn :: PlutusLedgerLanguage -> Map.Map MajorProtocolVersion (Set.Set DefaultFun)
```
kwxm, line 230: This (and other functions in this module) are used to perform validity checks during on-chain deserialisation, so it's important that they're efficient.  I think that things like `batch2 ++ batch3 ++ batch4 ++ batch5 ++ batch6` will be computed at compile time, but `Set.fromList` and `Map.fromList` won't, so this isn't as efficient as it could be.  It seems that this problem can be [solved](https://well-typed.com/blog/2020/06/th-for-static-data/) using Template Haskell, and I plan to look into this in a later PR: see [this issue](https://github.com/IntersectMBO/plutus-private/issues/1691).

Another possibility would be to cache the results: see [this comment](https://github.com/IntersectMBO/plutus/issues/6579#issuecomment-2997728816).

kwxm: I had to rework these functions because the previous versions relied on assumptions that are no longer valid (specifically, the predicate in `takeWhileAntitone` must satisfy a [certain property](https://hackage-content.haskell.org/package/containers-0.8/docs/Data-Map-Internal.html#v:takeWhileAntitone), and enabling new builtins in PlutusV1/V2 meant that this no longer held in the previous version of the function.

plutus-ledger-api/test/Spec/CostModelParams.hs
kwxm, line 1: The tests here are still not ideal (see [this issue](https://github.com/IntersectMBO/plutus/issues/6089)), so we should come back and take another look at them.

plutus-ledger-api/test/Spec/Versions.hs
kwxm, line 1: These tests needed substantial reworking because the previous versions relied on assumptions that are no longer true.  What's really being tested in many tests is that functions like `builtinsAvailableIn` behave properly.  These functions are used on the chain and it's important that they do what they're supposed to do, so I've tried to make the tests as comprehensive as possible.

plutus-ledger-api/src/PlutusLedgerApi/V1/ParamName.hs
kwxm, line 1: The changes to `builtinsAvailableIn` and related functions will make the builtins pass the deserialisation checks, making the builtins usable in principle on the chain at PV11, except that the cost model parameters also need to be updated in order to make the builtins (and SoPs) affordable .  The changes to these files enable the update to work properly.

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
```diff
@@ -153,44 +104,173 @@ ledgerLanguageIntroducedIn = \case
...
+ expensive because the proposal to update the relevant protocol parameters has
+ not (yet) been enacted.  This has left a "gap" in the cost model paramters: for
+ PlutusV3, the parameters for Batch 3 are followed those for 4a, then 4b, but
+ for PlutusV2 those for Batch3 are followed by those for Batch 4a, and those for
+ 4b aren't in use yet.  Since you can't actually use the 4b builtins in PlutusV2
+ at the moment, it's tempting to insert the 4a parameter before the 4b
+ parameters and enable them all at PV11 and with a suitable parameter update.
+ However, if we do do this there's a theoretical risk of turning a phase 2
+ failure into a phase 1 failure: would that be problematic?
+-}
```
kwxm, line 199: The way the three sets of cost model parameters are set up now means that the PlutusV1 parameters are a permutation of the PlutusV2 parameters (but the cost models they represent wll be exactly the same). This is unavoidable because the parameters for `serialiseData` were already present in the middle of the V2 parameters but had to be added to the end of the previous version of the V1 parameters (where `serialiseData` didn't exist).  The parameters for the two integer/bytestring conversion functions are also in different places, but the V2 ones are in the correct place for the as-yet-unenacted parameter update that will enable them.  It may just be easier to leave things as they are and not try to match the parameters up.  See the discussion [here](https://github.com/IntersectMBO/plutus-private/issues/1673).

kwxm:
^ Not sure if that's benchmarking what I want; I'll check later.

plutus-ledger-api/changelog.d/20250718_072102_kenneth.mackenzie_enable_all_builtins_2.md
```diff
@@ -0,0 +1,7 @@
+<!--
```
zliu41: uncomment the changelog

plutus-ledger-api/src/PlutusLedgerApi/Common/ProtocolVersions.hs
```diff
@@ -86,16 +94,26 @@ knownPVs =
...
---
--- We should not assign names to future protocol versions until it's
--- confirmed that they are correct, otherwise we could accidentally
--- associate something with the wrong protocol version.
+-- We're sometimes in an intermediate state where we've added new builtins but
+-- not yet released them (but intend to). This is used by some of the tests to
+-- decide what PVs the test should include.  UPDATE THIS when we're expecting to
+-- release new builtins in a forthcoming PV.
+newestPV :: MajorProtocolVersion
+newestPV = anonPV
```
zliu41: If new builtins are assigned to `futurePV` first, then changed to the appropriate PV when ready, then I don't see why we need this.

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
zliu41: Does this cause any performance regression? Previously `builtinsIntroducedIn` was a top-level Map that can be cached, but it is now a function.

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
```diff
@@ -153,44 +104,173 @@ ledgerLanguageIntroducedIn = \case
...
-See Note [New builtins/language versions and protocol versions]
+{- It's tempting to try something like `fmap toEnum [0..50]` here, but that's
+   dangerous because the order of the constructors in DefaultFun doesn't
+   precisely match the order that the builtins were introduced in.  A safer
+   alternative would be to use the flat tags, but they're not directly
+   accessible at the moment.
+-}
+batch1 :: [DefaultFun]
+batch1 =
+  [ AddInteger, SubtractInteger, MultiplyInteger, DivideInteger, QuotientInteger
```
zliu41, line 149: It's tedious to check the correctness of these batches manually. How about keeping the old version of `builtinsIntroducedIn`, and adding some property tests to verify that the old and new versions always agree (up to PV11)?

plutus-ledger-api/changelog.d/20250718_072102_kenneth.mackenzie_enable_all_builtins_2.md
kwxm: Oops.  Done.

plutus-ledger-api/src/PlutusLedgerApi/Common/ProtocolVersions.hs
kwxm: > If new builtins are assigned to `futurePV` first, then changed to the appropriate PV when ready, then I don't see why we need this.

The preceding comment attempts to explain that.  Some of the tests ([here](https://github.com/IntersectMBO/plutus/blob/8593065e8a032b00e8601c928d301f10b3366000/plutus-ledger-api/test/Spec/Versions.hs#L36) for instance) iterate over all of the deployed (or just-about-to-be-deployed) PVs and use `newestPV` to say where to stop.  We can't use `futurePV` for that because it's `maxBound`.  I added `newestPV` so that we wouldn't have to go through the tests and replace the actual most recent PV everywhere when we update it.

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
kwxm: I ran the `validation-decode` benchmarks a few times and the results suggest that there's a slowdown of about 1.5-3% in deserialisation times.  An [earlier comment](https://github.com/IntersectMBO/plutus/pull/7223/files/38936222241e72c974f4be72273091e844e8bab7#r2215112646) mentions this and some ways to deal with (like evaluate as much as possible at compile time) it and I plan to come back and look at it again in the near future.

Superficially this looks as if it could be more efficient than the previous version because that had to look at all (LL,PV) combinations and the new version only has to look at the PV values once we've callled the function to get everything for the relevant LL version.  It does appear that it's not more efficient and I want to fix that.  Also, the previous version no longer works because it assumes that if something's available in a praticular LL and PV then it's avaliable for all earlier LLs and PVs, and that no longer holds (we're enabling things in PV11 that aren't available in PV10, for example).

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
kwxm: I'm not sure what you mean.  Do you want to keep both the old and the new versions?  The batches were implcitly [present](https://github.com/IntersectMBO/plutus/blob/2a70305402e563493d4cc6c195a573f6c5f9c529/plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs#L90) in the previous version and here I've pulled them out to the top level to make things a bit more explicit and easier to relate to the specification; also, they're used in the tests so it helps to be able to export them.  Of course, you should only ever add anything to the most recent batch (or add a new batch); maybe I should add a prominent comment to point that out.

zliu41: I mean adding a test verifying that the new `builtinsAvailableIn ll pv` and the old `builtinsAvailableIn ll pv` always return the exact same set. This is especially important for pv=10. Or is this already covered in existing tests?

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
zliu41: Like you mentioned on Slack, this check can be significantly simplified from PV11. So perhaps you'd want to implement a separate check just for PV11?

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
kwxm: So we'd have to keep the old `builtinsAvailableIn` function around specifically for the tests, even though it's not used?  It gives incorrect results for PV11, so we'd have to modify it anyway: that's why I changed the implementation!

kwxm: [This test](https://github.com/IntersectMBO/plutus/blob/8593065e8a032b00e8601c928d301f10b3366000/plutus-ledger-api/test/Spec/Versions.hs#L192) should test that deserialisation succeeds or fails as expected for every single builtin in every LL/PV combination, and it's much more thorough than the [previous version](https://github.com/IntersectMBO/plutus/blob/5634bfe5c5a55c3a46da07dfd1870ff3e742208c/plutus-ledger-api/test/Spec/Versions.hs#L63).  It depends on `batch1`, `batch2`, etc being correct, but it's hard to see how to test the correctness of `bultinsAvailableIn` without using those since we don't have any independent descrioption of which builtins are available.

The batches in the new version start [here](https://github.com/IntersectMBO/plutus/blob/8593065e8a032b00e8601c928d301f10b3366000/plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs#L145) and they're essentially just lifted from the [previous version](https://github.com/IntersectMBO/plutus/blob/5634bfe5c5a55c3a46da07dfd1870ff3e742208c/plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs#L91) of `builtinsAvailableIn`, with batch4 having to be split into two to handle `integerToByteString` and `byteStringToInteger` in PlutusV2 at Plomin.

zliu41: > So we'd have to keep the old builtinsAvailableIn function around specifically for the tests, even though it's not used?

Yeah, why not, we can move it to a test package. But if there's already a test that verifies that the new `builtinsAvailableIn` gives correct results for all ll/pv (especially PV10 and PV11), then we are good - that coverage is all we need.

kwxm: OK.  The old version works correctly up to PV10, so I can move it into the test code somewhere and add a test that the new implementation matches up to PV10 (but not PV11, where it breaks).

plutus-ledger-api/src/PlutusLedgerApi/V2/ParamName.hs
```diff
@@ -200,5 +202,117 @@ data ParamName =
...
+  | CountSetBits'cpu'arguments'intercept
+  | CountSetBits'cpu'arguments'slope
+  | CountSetBits'memory'arguments
+  | FindFirstSetBit'cpu'arguments'intercept
+  | FindFirstSetBit'cpu'arguments'slope
+  | FindFirstSetBit'memory'arguments
+  | Ripemd_160'cpu'arguments'intercept
+  | Ripemd_160'cpu'arguments'slope
+  | Ripemd_160'memory'arguments
+  -- Not yet deployed
```
zliu41: Aren't these also to be deployed in PV11? Why are they "Not yet deployed"?

kwxm: I've changed that to "To be deployed in PV11", but I can take it out if you want.  I may have copied it from the PlutusV3 parameters, where it marks the boundaries between the parameters for the things that are on the chain now and those that aren't.  

A lot of the code in plutus-ledger-api is a bit weird because in a sort of in-between state where it's about stuff that's waiting to happen.

plutus-ledger-api/src/PlutusLedgerApi/Common/ProtocolVersions.hs
```diff
@@ -68,16 +72,20 @@ valentinePV = MajorProtocolVersion 8
...
+-- | The Plomin HF was an intra-era HF where some new builtin functions were
+-- introduced in Plutus V2 and V3.
 plominPV :: MajorProtocolVersion
 plominPV = MajorProtocolVersion 10
 
+-- | Not sure what this is going to be called yet
+pv11PV :: MajorProtocolVersion
+pv11PV = MajorProtocolVersion 11
+
 -- | The set of protocol versions that are "known", i.e. that have been released
```
zliu41, line 84: nit: I'm not quite sure what it means for a PV to be "released". It's probably more accurate to say: these are the PVs in which some Plutus primitives are enabled.

plutus-ledger-api/src/PlutusLedgerApi/Common/ProtocolVersions.hs
```diff
@@ -86,16 +94,26 @@ knownPVs =
...
--- be used for something. It's a very high protocol version that should never
--- appear in reality.  New builtins should always be given this protocol version
--- until they've been finalised.
---
--- We should not assign names to future protocol versions until it's
--- confirmed that they are correct, otherwise we could accidentally
--- associate something with the wrong protocol version.
+-- We're sometimes in an intermediate state where we've added new builtins but
+-- not yet released them (but intend to). This is used by some of the tests to
+-- decide what PVs the test should include.  UPDATE THIS when we're expecting to
```
zliu41, line 102: Rather than saying "UPDATE THIS" in a comment here - which is extremely easy to overlook - it would be nice to have a checklist somewhere, that documents everything we need to do when we are ready to release new builtins.

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
```diff
@@ -153,44 +104,181 @@ ledgerLanguageIntroducedIn = \case
...
+ #6056 and #6065.  They are available on the chain, but they're prohibitively
+ expensive because the proposal to update the relevant protocol parameters has
+ not (yet) been enacted.  This has left a "gap" in the cost model paramters: for
+ PlutusV3, the parameters for Batch 3 are followed those for 4a, then 4b, but
+ for PlutusV2 those for Batch3 are followed by those for Batch 4a, and those for
+ 4b aren't in use yet.  Since you can't actually use the 4b builtins in PlutusV2
+ at the moment, it's tempting to insert the 4a parameter before the 4b
+ parameters and enable them all at PV11 and with a suitable parameter update.
+ However, if we do do this there's a theoretical risk of turning a phase 2
+ failure into a phase 1 failure: would that be problematic?
```
zliu41, line 198: Good point - different node versions disagreeing on whether a failure is phase 1 or phase 2 can indeed be _very_ problematic.

plutus-ledger-api/src/PlutusLedgerApi/MachineParameters.hs
```diff
@@ -16,7 +15,7 @@ machineParametersFor
   -> DefaultMachineParameters
 machineParametersFor ledgerLang majorPV =
   MachineParameters
-      (if majorPV < futurePV
+      (if majorPV < pv11PV
```
zliu41, line 18: :+1:

plutus-ledger-api/test/Spec/CostModelParams.hs
```diff
@@ -10,13 +10,14 @@ import PlutusLedgerApi.Test.V3.EvaluationContext qualified as V3
 import PlutusLedgerApi.V1 qualified as V1
 import PlutusLedgerApi.V2 qualified as V2
 import PlutusLedgerApi.V3 qualified as V3
+-- import PlutusCore.Evaluation.Machine.ExBudgetingDefaults (defaultCostModelParamsForTesting)
```
zliu41: Remove?

plutus-ledger-api/src/PlutusLedgerApi/Common/ProtocolVersions.hs
kwxm: Ah, I have an [issue](https://github.com/IntersectMBO/plutus-private/issues/1495) to do that: thanks for reminding me. I'll put the issue in the current sprint so I can do it while the process is still fresh in my mind.   I had in mind to write some kind of textual document, but maybe an issue template would be a better way to do it.

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
effectfully: > I think that things like batch2 ++ batch3 ++ batch4 ++ batch5 ++ batch6 will be computed at compile time

You don't know that, unless you've checked it. These are not small definitions and they're not marked as INLINE, GHC can easily avoid inlining them, in which case fusion doesn't kick in, in which case they don't get computed statically.

> Previously builtinsIntroducedIn was a top-level Map that can be cached, but it is now a function.

All we need to make it efficient again is just to put

```haskell
      Map.fromList
      [ (alonzoPV, Set.fromList batch1)
      , (pv11PV,   Set.fromList (batch2 ++ batch3 ++ batch4 ++ batch5 ++ batch6))
      ]
```

etc into their own top-level definitions marked as `OPAQUE`.

plutus-ledger-api/test/Spec/CostModelParams.hs
```diff
@@ -28,9 +28,9 @@ tests =
     "CostModelParams"
     "costModelParams"
     [ embed $ testCase "length" do
-        166 @=? length v1_ParamNames
-        185 @=? length v2_ParamNames
-        297 @=? length v3_ParamNames
+        295 @=? length v1_ParamNames
+        295 @=? length v2_ParamNames
+        313 @=? length v3_ParamNames
```
effectfully, line 33: We should make those golden tests.

effectfully (commented):
OK, there's no way I can make "I've spent hours reviewing Kenneth's PR and it turned out to be all good as expected" look good on a performance review, so here's a couple of comments that I had from a cursory reading.

plutus-ledger-api/test/Spec/CostModelParams.hs
kwxm: > We should make those golden tests.

Not sure about that.  It might be a bit too easy just to use `--accept` to change them.  This forces you to enter the correct numbers yourself.

effectfully: Good point, I take my "should" back.

PR #7224, opened 2025-07-21
Fix certifier builtin unparsing; retain certificates for failing tests

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1695

### DIFF
--- a/plutus-executables/test/certifier/Test/Certifier/Executable.hs
+++ b/plutus-executables/test/certifier/Test/Certifier/Executable.hs
@@ -132,8 +132,9 @@ agdaTestCert name = do
     setCurrentDirectory certDir
     (resCode, resText) <- runAgda ("src" </> fstToUpper name <> ".agda")
     setCurrentDirectory oldDir
-    removeDirectoryRecursive certDir
-    assertBool (name ++ " creates an invalid certificate:" ++ resText) (resCode == ExitSuccess)
+    if resCode == ExitSuccess
+      then removeDirectoryRecursive certDir
+      else assertFailure $ name ++ " creates an invalid certificate: " ++ resText
 
 {-
 agdaExampleCert :: String -> Assertion
@@ -157,6 +158,9 @@ srcTests =
   , "MinBS"
   , "AA2-CSE"
   , "forceDelayIfThenElse"
+  , "listUnparse"
+  , "pairUnparse"
+  , "nestedUnparse"
   ]
 
 makeExampleTests :: [ String ] -> [ TestTree ]
--- a/plutus-metatheory/src/FFI/AgdaUnparse.hs
+++ b/plutus-metatheory/src/FFI/AgdaUnparse.hs
@@ -4,8 +4,10 @@ module FFI.AgdaUnparse where
 
 import Data.ByteString (ByteString)
 import Data.Functor.Identity
+import Data.Proxy
 import Data.Text (Text)
 import Data.Text qualified as T
+import Data.Vector.Strict (Vector)
 import FFI.Untyped qualified as AgdaFFI
 import PlutusCore qualified as PLC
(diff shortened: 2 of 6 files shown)


plutus-executables/test/certifier/Test/Certifier/Executable.hs
```diff
@@ -157,6 +158,9 @@ srcTests =
   , "MinBS"
   , "AA2-CSE"
   , "forceDelayIfThenElse"
+  , "listUnparse"
+  , "pairUnparse"
+  , "nestedUnparse"
```
ramsay-t: If these are going to run as full serialisation tests could we roll them into one file? There is about a 6s overhead to running the serialisation and then calling Agda at the command line, so one file with lots of types in it might be better?

plutus-metatheory/src/FFI/AgdaUnparse.hs
```diff
@@ -120,51 +124,49 @@ instance AgdaUnparse (UPLC.DefaultUni (PLC.Esc a)) where
...
+      PLC.ValueOf PLC.DefaultUniUnit _ ->
         "unit " ++ agdaUnparse ()
-      PLC.ValueOf PLC.DefaultUniData _ :=> Identity val ->
+      PLC.ValueOf PLC.DefaultUniData val ->
         "pdata " ++ agdaUnparse val
-      PLC.ValueOf (PLC.DefaultUniList elemType) _ :=> Identity val ->
-        "(list " ++ agdaUnparse elemType ++ ") " ++ agdaUnparseDList elemType val
-      PLC.ValueOf (PLC.DefaultUniPair type1 type2) _ :=> Identity val ->
+      PLC.ValueOf (PLC.DefaultUniList elemType) val ->
+        "(list "
```
zliu41: You can just call `agdaUnparse (PLC.DefaultUniList elemType)` here?

plutus-metatheory/src/FFI/AgdaUnparse.hs
```diff
@@ -120,51 +124,49 @@ instance AgdaUnparse (UPLC.DefaultUni (PLC.Esc a)) where
...
+        ++ agdaUnparse elemType
+        ++ ") "
+        ++
+          ( PLC.bring (Proxy @AgdaUnparse) elemType
+          $ agdaUnparse val
+          )
+      PLC.ValueOf (PLC.DefaultUniPair type1 type2) val ->
         "(pair "
```
zliu41: ditto

plutus-metatheory/src/FFI/AgdaUnparse.hs
```diff
@@ -120,51 +124,49 @@ instance AgdaUnparse (UPLC.DefaultUni (PLC.Esc a)) where
   agdaUnparse (PLC.DefaultUniArray _) = error "Arrays are currently not supported."
   agdaUnparse (PLC.DefaultUniApply _ _) = error "Application of an unknown type is not supported."
 
-agdaUnparseValue :: DSum (PLC.ValueOf UPLC.DefaultUni) Identity -> String
-agdaUnparseValue dSum =
+agdaUnparseValue :: PLC.Some (PLC.ValueOf UPLC.DefaultUni) -> String
+agdaUnparseValue (PLC.Some dSum) =
```
zliu41: The name `dSum` is no longer appropriate since you aren't using `DSum` any more.

plutus-metatheory/src/FFI/AgdaUnparse.hs
ana-pantilie: That's right, good catch. Thanks!

plutus-executables/test/certifier/Test/Certifier/Executable.hs
ana-pantilie: Good idea, moved them to one file.

PR #7225, opened 2025-07-21
Add BuiltinValue type to plutus-core

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1690

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
@@ -53,6 +53,7 @@ import PlutusCore.Data (Data)
 import PlutusCore.Evaluation.Machine.ExMemoryUsage (IntegerCostedLiterally (..),
                                                     NumBytesCostedAsNumWords (..))
 import PlutusCore.Pretty.Extra (juxtRenderContext)
+import PlutusCore.Value qualified as PLC
 
 import Control.Monad.Except (throwError)
 import Data.ByteString (ByteString)
@@ -110,6 +111,7 @@ data DefaultUni a where
     DefaultUniString :: DefaultUni (Esc Text)
     DefaultUniUnit :: DefaultUni (Esc ())
     DefaultUniBool :: DefaultUni (Esc Bool)
+    DefaultUniValue :: DefaultUni (Esc PLC.Value)
     DefaultUniProtoArray :: DefaultUni (Esc Strict.Vector)
     DefaultUniProtoList :: DefaultUni (Esc [])
     DefaultUniProtoPair :: DefaultUni (Esc (,))
@@ -156,6 +158,9 @@ instance GEq DefaultUni where
         geqStep DefaultUniBool a2 = do
             DefaultUniBool <- Just a2
             Just Refl
+        geqStep DefaultUniValue a2 = do
+            DefaultUniValue <- Just a2
+            Just Refl
         geqStep DefaultUniProtoList a2 = do
             DefaultUniProtoList <- Just a2
             Just Refl
@@ -198,6 +203,7 @@ instance ToKind DefaultUni where
     toSingKind DefaultUniString               = knownKind
     toSingKind DefaultUniUnit                 = knownKind
     toSingKind DefaultUniBool                 = knownKind
+    toSingKind DefaultUniValue                = knownKind
     toSingKind DefaultUniProtoList            = knownKind
     toSingKind DefaultUniProtoArray           = knownKind
     toSingKind DefaultUniProtoPair            = knownKind
@@ -223,6 +229,7 @@ instance PrettyBy RenderContext (DefaultUni a) where
         DefaultUniString               -> "string"
         DefaultUniUnit                 -> "unit"
         DefaultUniBool                 -> "bool"
+        DefaultUniValue                -> "value"
         DefaultUniProtoList            -> "list"
         DefaultUniProtoArray           -> "array"
         DefaultUniProtoPair            -> "pair"
@@ -263,6 +270,8 @@ instance DefaultUni `Contains` () where
     knownUni = DefaultUniUnit
 instance DefaultUni `Contains` Bool where
     knownUni = DefaultUniBool
+instance DefaultUni `Contains` PLC.Value where
+    knownUni = DefaultUniValue
 instance DefaultUni `Contains` [] where
     knownUni = DefaultUniProtoList
 instance DefaultUni `Contains` Strict.Vector where
@@ -288,6 +297,8 @@ instance KnownBuiltinTypeAst tyname DefaultUni () =>
     KnownTypeAst tyname DefaultUni ()
 instance KnownBuiltinTypeAst tyname DefaultUni Bool =>
     KnownTypeAst tyname DefaultUni Bool
+instance KnownBuiltinTypeAst tyname DefaultUni PLC.Value =>
+    KnownTypeAst tyname DefaultUni PLC.Value
 instance KnownBuiltinTypeAst tyname DefaultUni [a] =>
     KnownTypeAst tyname DefaultUni [a]
 instance KnownBuiltinTypeAst tyname DefaultUni (Strict.Vector a) =>
@@ -313,6 +324,8 @@ instance KnownBuiltinTypeIn DefaultUni term () =>
     ReadKnownIn DefaultUni term ()
 instance KnownBuiltinTypeIn DefaultUni term Bool =>
     ReadKnownIn DefaultUni term Bool
+instance KnownBuiltinTypeIn DefaultUni term PLC.Value =>
+    ReadKnownIn DefaultUni term PLC.Value
 instance KnownBuiltinTypeIn DefaultUni term Data =>
     ReadKnownIn DefaultUni term Data
 instance KnownBuiltinTypeIn DefaultUni term [a] =>
@@ -338,6 +351,8 @@ instance KnownBuiltinTypeIn DefaultUni term () =>
     MakeKnownIn DefaultUni term ()
 instance KnownBuiltinTypeIn DefaultUni term Bool =>
     MakeKnownIn DefaultUni term Bool
+instance KnownBuiltinTypeIn DefaultUni term PLC.Value =>
+    MakeKnownIn DefaultUni term PLC.Value
 instance KnownBuiltinTypeIn DefaultUni term Data =>
     MakeKnownIn DefaultUni term Data
 instance KnownBuiltinTypeIn DefaultUni term [a] =>
@@ -588,6 +603,7 @@ instance Closed DefaultUni where
         , constr `Permits` Text
         , constr `Permits` ()
         , constr `Permits` Bool
+        , constr `Permits` PLC.Value
         , constr `Permits` []
         , constr `Permits` Strict.Vector
         , constr `Permits` (,)
@@ -612,6 +628,7 @@ instance Closed DefaultUni where
     encodeUni DefaultUniBLS12_381_G2_Element = [10]
     encodeUni DefaultUniBLS12_381_MlResult   = [11]
     encodeUni DefaultUniProtoArray           = [12]
+    encodeUni DefaultUniValue                = [13]
 
     -- See Note [Decoding universes].
     -- See Note [Stable encoding of tags].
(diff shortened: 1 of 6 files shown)


zliu41:
Ready for review. This is a naive implementation, where `Value` is simply a newtype wrapper around `Map ByteString (Map ByteString Integer)`. The problem is that this doesn't allow efficient access to the size of the largest inner map - it takes `O(m)` time to do so where `m` is the size of the outer map. This is likely a problem for costing operations like `insertCoin` and `deleteCoin`, which only take `O(logm) + O(logk)`, where `k` is the size of the largest inner map, since size lookup is more expensive than the operations themselves.

Nonetheless wen can merge this first, and if we need a different representation of `Value` we can modify it in a separate PR.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -110,6 +111,7 @@ data DefaultUni a where
     DefaultUniString :: DefaultUni (Esc Text)
     DefaultUniUnit :: DefaultUni (Esc ())
     DefaultUniBool :: DefaultUni (Esc Bool)
+    DefaultUniValue :: DefaultUni (Esc Value)
```
SeungheonOh: I think it's better to follow order defined in `encodeUni`, i.e. haven `DefaultUniValue` at the end.

SeungheonOh: There's more instances of this below as well.

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
```diff
@@ -86,6 +88,18 @@ conList uniA = trailingWhitespace . inBrackets $
 conArray :: DefaultUni (Esc a) -> Parser (Vector a)
 conArray uniA = Vector.fromList <$> conList uniA
 
+-- | Parser for values.
+conValue :: Parser PLC.Value
+conValue =
+  Value.fromList
+  <$> conList
+        (DefaultUniPair
```
SeungheonOh: we can probably use some better syntax for value. Just using 3tuple for each entry would be more readable.

SeungheonOh (approved):
Everything looks reasonable besides some nitpicks.

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
zliu41: That may be slightly better-looking, but then we won't be able to reuse the built-in list parser since a built-in list can't contain a 3-tuple. I'll leave this as future work.

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
```diff
@@ -90,15 +90,7 @@ conArray uniA = Vector.fromList <$> conList uniA
...
-conValue =
-  Value.fromList
-  <$> conList
-        (DefaultUniPair
-          DefaultUniByteString
-          (DefaultUniList
-            (DefaultUniPair DefaultUniByteString DefaultUniInteger)
-          )
-        )
+conValue = Value.fromList <$> conList knownUni
```
ana-pantilie, line 93: Can you explain this change? I remember I wasn't sure about the former implementation, so I'd like to understand how this all works.

zliu41: `` knownUni :: uni `Contains` a => uni (Esc a) ``, so it can be instantiated to any `uni (Esc a)` as long as `` uni `Contains` a ``. So there's no need to specify the concrete type.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -118,6 +119,7 @@ data DefaultUni a where
     DefaultUniBLS12_381_G1_Element :: DefaultUni (Esc BLS12_381.G1.Element)
     DefaultUniBLS12_381_G2_Element :: DefaultUni (Esc BLS12_381.G2.Element)
     DefaultUniBLS12_381_MlResult :: DefaultUni (Esc BLS12_381.Pairing.MlResult)
+    DefaultUniValue :: DefaultUni (Esc Value)
```
effectfully, line 122: I thought we agreed to call it `MaryValue`? Let's not repeat ledger folks' mistakes
image

zliu41: I don't recall the discussion - do you have a pointer? It's always called Value in plutus, and I'm hesitant to make a name 80% longer unless there's a clear advantage.

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
kwxm: I think that for the parser to work it'll need a case for `value` [here](https://github.com/IntersectMBO/plutus/blob/ad408e9af4127b92356862ab45f441697cebcb44/plutus-core/plutus-core/src/PlutusCore/Parser/Type.hs#L147) as well.

kwxm: I tried adding "value" and it still didn't work.  I'm not sure if that's because there's something wrong with the parser or the prettyprinter.  With "value" added in `Type.hs`, `uplc` can parse

```
(program 1.1.0
 (con value [(#1234,[(#5678, 99)])])
)
```
but prints out
```
(con value [(#1234, [])])
``` 

Maybe none of this matters until the internal representation's finalised.

plutus-metatheory/src/FFI/AgdaUnparse.hs
kwxm, line 1: Do we have to extend the metatheory when we add a new type now?  I wasn't sure if that was the case: we used to be able to leave it until later, but maybe the dependency of the executables on the metatheory means that we can't get away with that any more.

plutus-tx/src/PlutusTx/Lift/Class.hs
```diff
@@ -222,10 +224,18 @@ instance
 instance (uni `PLC.HasTypeLevel` Data) => Typeable uni BuiltinData where
   typeRep _ = typeRepBuiltin (Proxy @Data)
 
+-- FIXME (https://github.com/IntersectMBO/plutus-private/issues/1703)
+instance Typeable uni BuiltinValue where
+  typeRep = Haskell.error "Not Implemented"
```
kwxm, line 229: Is this particularly difficult to implement, or are we waiting until we decide the final implementation of the type?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
kwxm: (Following on from @effectfully's coment.) I don't think I've ever seen `MaryValue` before.  I suppose `Value` may be potentially confusing: we also have things like `CekValue`, which is something totally different.  I don't care massively though.

ana-pantilie: I don't recall that discussion either, and yes everywhere in our repo it's called `Value`. We have the current implementation of `Value` in the plutus ledger api, and I suppose that will be deprecated once the builtin is available.

> I suppose Value may be potentially confusing: we also have things like CekValue, which is something totally different.

Aren't the two "values" in this case part of separate languages? In on-chain code, it should be clear that `Value` has a specific meaning, and as a builtin it will be exposed as `BuiltinValue`. I think `BuiltinMaryValue` is going to be too long. I think it may only be confusing in off-chain code, where as you said there's `CekValue`, but do you even use the on-chain `Value` in off-chain code?

effectfully: It's literally in the [CIP](https://github.com/cardano-foundation/CIPs/pull/1011) name: "Plutus Core `MaryEraValue` Builtins". Some of us also discussed it on slack, see the last three messages starting [here](https://input-output-rnd.slack.com/archives/CCRB7BU8Y/p1730332283168279?thread_ts=1730305206.996379&cid=CCRB7BU8Y).

zliu41: But _only_ in the CIP name - not in the proposed types and signatures. There's currently no plan to introduce a new kind of Value, and the name is changeable later, so I see no point of speculatively doing so.

plutus-metatheory/src/FFI/AgdaUnparse.hs
zliu41: We must not require that - adding new types/functions to the metatheory can be a lot of work, and making it a blocker for the actual implementation is unacceptable imo.

plutus-tx/src/PlutusTx/Lift/Class.hs
zliu41: It does depend on the final implementation, but also, this is Plinth specific, and imo should be done in a separate step together with all other Plinth-side integration.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: > There's currently no plan to introduce a new kind of Value, and the name is changeable later, so I see no point of speculatively doing so.

`MaryValue` is what it's called in the ledger and what Alexey explicitly advised us to call it. Why introduce this inconsistency?

>  the name is changeable later, so I see no point of speculatively doing so.

Well, if you're ok with breaking pretty-printed programs and user code (in Plinth), then sure. But why?

I don't really care, `Value` is ok, but it's inconsistent with what the Ledger calls it and it's not a future-proof name, hence I strongly believe this outweighs the benefits of saving 4 characters.

zliu41: > MaryValue is what it's called in the ledger and what Alexey explicitly advised us to call it. Why introduce this inconsistency?

Because I disagree with the advice. Renaming something comes with a cognitive cost. `Value` has existed in plutus for what, at least five years? It's been talked about everywhere - papers, docs, discord, reddit... Everyone knows what `Value` is, and this is not a strong enough reason to rename it. Plutus is more user facing. Ledger is more internal.

In the near future, as Plutus gets more product support, we can revisit this, and a product manager can make the call.

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
zliu41: Updated `defaultUni`. Will revisit this and add a test in the next PR.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: > Value has existed in plutus for what, at least five years? It's been talked about everywhere - papers, docs, discord, reddit... Everyone knows what Value is, and this is not a strong enough reason to rename it.

Yes, and it's been a different thing! It used to be a Plinth data type and now it's a built-in type, to me this sounds like another reason to introduce a descriptive name.

In any case, I don't believe it's very important and I wasn't trying to block this PR or anything. `Value` is fine.

plutus-metatheory/src/FFI/AgdaUnparse.hs
kwxm: ^ I'm not sure what the situation is.  I was wondering if you'd updated the Agda code because you couldn't get your PR into a mergeable state without doing so.

plutus-tx/src/PlutusTx/Lift/Class.hs
kwxm: I wasn't sure if you'd been forced to start altering the plutus-tx code after extending the universe and had just added a stub to get past that.

kwxm: ^ I tried adding `Float` to the universe and I think things have changed since I last added a new type.  As the PR indicates, you need to add an instance for `AgdaUnparse` in `plutus-metatheory` and instances for `Lift` and `Typeable` in `plutus-tx`.  I'll try to document that somehwere.  I think you used to be able to add a new type to UPLC without modifying anything outside `plutus-core`, but that's changed now, which is a bit annoying because (as @zliu41 points out) it makes it a bit more difficult to add new built-in functions and types in a sequence of small PRs.

effectfully: > instances for `Lift` and `Typeable` in `plutus-tx`

I intentionally made this mandatory. We can just turn off the check in `PlutusTx.Lift.TestInstances` if we don't like it. Or we can turn it off when adding a new built-in type and then turn it back on once it's handled in Plinth.

PR #7231, opened 2025-07-23
Make plinth use builtin bool

### LINKED ISSUE #7219: Compile Plinth `Bool` to built-in `Bool`
Currently we compile Plinth `Bool` (and its values) to a SOP `Bool`, which I believe is far less efficient than using the newly available casing on built-in `Bool`. This issue for switching from the former to the latter.

I don't have any ideas on how to do that.

SeungheonOh:
closes #7219 

I have tried few(a bit too much) other approaches as well here--for example using `datatypeDefs` from `DefState` instead of adding an extra field--but what I have here is cleanest. 

Currently, it's actually slower than how it was before. This is because instead of using `case` directly, current implementation defines extra function that will apply branches to `case`; this is not optimal. However, in order to make it work efficiently as intended, we need to make some more bigger structural changes. 

There's few ways to go about here. We can add builtin casing primops to PIR; PIR will have primop named `Bool_match` and when it is called, it will inline case. This will allow us to not change anything on plutus-tx-plugin side which is nice. 

Other option is to handle case matching on the plugin side. Currently, plugin simply instantiates and applies branches to defined deconstructor. This can be changed so that deconstructor itself would define how a value is matched. Deconstructor will have type `:: Term -> [Term] -> Term` where first is scrutinee and second is branches instead of current `:: Term`.

I see latter more favorably as it won't require any "default" behaviors added to PIR, although it ~~might be a slightly bigger change~~ probably not, I read a bit more code and it seems pretty contained. I'd like to hear thoughts here. @zliu41 @effectfully

### DIFF
--- a/.github/workflows/compare-golden-budgets.yaml
+++ b/.github/workflows/compare-golden-budgets.yaml
@@ -1,0 +1,82 @@
+name: "🏆 Compare Golden Budgets"
+
+on:
+  pull_request:
+    paths:
+      - '**.eval.golden'
+
+jobs:
+  run-script:
+    runs-on: ubuntu-latest
+
+    permissions:
+      pull-requests: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@main
+        with:
+          fetch-depth: 5
+
+      - name: Make script executable
+        run: chmod +x scripts/eval-percent-diff.sh
+
+      - name: Run the bash script and capture output
+        run: ./scripts/eval-percent-diff.sh -t ${{ github.event.pull_request.base.sha }} 2>&1 | tee output.log
+
+      - name: Comment script output on PR
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const fs = require('fs');
+
+            const botCommentIdentifier = "<!-- GITHUB_ACTION_GOLDEN_FILE_BOT -->";
+
+            const scriptOutput = fs.readFileSync('output.log', 'utf8');
+
+            const headSha = context.payload.pull_request.head.sha;
+            const baseSha = context.payload.pull_request.base.sha;            
+            const commentBody = `
+              ${botCommentIdentifier}            
+              ## Execution Budget Golden Diff
+
+              ${baseSha.substring(0, 7)} (master) vs ${headSha.substring(0, 7)}
+
+              <details>
+              <summary>output</summary>
+              
+              ${scriptOutput}              
+
+              </details>
+
+              This comment will get updated when changes are made.
+            `;
+
+            const { data: comments } = await github.rest.issues.listComments({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+            });
+
+            const botComment = comments.find(comment => comment.body.includes(botCommentIdentifier));
+
+            if (botComment) {
+              console.log('Found existing comment. Updating...');
+              await github.rest.issues.updateComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                comment_id: botComment.id,
+                body: commentBody
+              });
+              console.log('Comment updated successfully.');
+            } else {
+              console.log('No existing comment found. Creating a new one...');
+              await github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.issue.number,
+                body: commentBody
+              });
+              console.log('Comment posted successfully.');
+            }
--- a/.github/workflows/slack-message-broker.yml
+++ b/.github/workflows/slack-message-broker.yml
@@ -21,6 +21,7 @@ on:
       - "🌘 Nightly Testsuite"
       - "📝 Papers & Specs"
       - "📊 Code Coverage Report"
+      - "🏆 Compare Golden Budgets"
       - "pages-build-deployment"
 
 jobs:
--- a/scripts/eval-percent-diff.sh
+++ b/scripts/eval-percent-diff.sh
@@ -1,0 +1,132 @@
+#!/usr/bin/env bash
+# diff‑budget.sh – show percent deltas for *.eval.golden metrics
+# Usage:  diff‑budget.sh [-t] [<base-commit>]
+#    -t    output as Markdown tables instead of colored text
+
+set -euo pipefail
+
+usage() {
+  cat <<EOF
+Usage: $0 [-t] [<base-commit>]
+
+  -t             output as Markdown tables
+  <base-commit>  if given, diff that commit against your working tree;
+                 otherwise diff working tree vs HEAD.
+EOF
+  exit 1
+}
+
+# --- flag parsing ------------------------------------------------------------
+TFLAG=0
+while getopts ":t" opt; do
+  case $opt in
+    t) TFLAG=1 ;;
+    *) usage ;;
+  esac
+done
+shift $((OPTIND - 1))
+
+# --- positional argument (base commit) --------------------------------------
+if [[ $# -gt 1 ]]; then
+  usage
+elif [[ $# -eq 1 ]]; then
+  BASE="$1"
+  GIT_DIFF_CMD=(git diff "$BASE" -- '*.eval.golden')
+else
+  GIT_DIFF_CMD=(git diff -- '*.eval.golden')
+fi
+
+# --- run diff and post‑process ----------------------------------------------
+if [[ $TFLAG -eq 1 ]]; then
+  # Markdown table output
+  "${GIT_DIFF_CMD[@]}" | gawk '
+  function strip_us(num,   t) { gsub(/_/, "", num); return num+0 }
+  function pct(old, new) {
+    if (old==0) return "N/A";
+    return sprintf("%+.2f%%", 100*(new-old)/old)
+  }
+  BEGIN {
+    known["CPU"]=1; known["Memory"]=1;
+    known["Term Size"]=1; known["Flat Size"]=1;
+    file=""; seen=0;
+  }
+  # new file header: reset seen flag, store filename
+  /^--- a\/.*\.eval\.golden/ {
+    file = substr($2,3)
+    seen = 0
+    next
+  }
+  # capture old values
+  /^-[A-Za-z ]+:[[:space:]]+[0-9_]+/ {
+    match($0,/^-([A-Za-z ]+):[[:space:]]+([0-9_]+)/,m)
+    label=m[1]; if (!(label in known)) next
+    old[label]=strip_us(m[2]); old_raw[label]=m[2]
+    next
+  }
+  # capture new values and print row (with header if first)
+  /^\+[A-Za-z ]+:[[:space:]]+[0-9_]+/ {
+    match($0,/^\+([A-Za-z ]+):[[:space:]]+([0-9_]+)/,m)
+    label=m[1]; if (!(label in known)) next
+    newval=strip_us(m[2]); raw_new=m[2]
+    if (label in old) {
+      delta=pct(old[label],newval)
+      if (!seen) {
+        print ""
+        print "### " file
+        print ""
+        print "| Metric     | Old     | New     | Δ%      |"
+        print "|------------|---------|---------|---------|"
+        seen=1
+      }
+      printf("| %-10s | `%7s` | `%7s` | %7s |\n",
+             label, old_raw[label], raw_new, delta)
+      delete old[label]
+    }
+  }
+  '
+else
+  # Colored text output
+  "${GIT_DIFF_CMD[@]}" | gawk -v ESC="$(printf '\033')" '
+  function strip_us(num,   t) { gsub(/_/, "", num); return num+0 }
+  function pct(old, new) {
+    if (old==0) return "N/A";
+    return sprintf("%+.2f%%", 100*(new-old)/old)
+  }
+  BEGIN {
+    known["CPU"]=1; known["Memory"]=1;
+    known["Term Size"]=1; known["Flat Size"]=1;
+    red   = ESC "[31m"; green = ESC "[32m"; reset = ESC "[0m";
+    file=""; seen=0;
+  }
+  # new file header: reset seen flag, store filename
+  /^--- a\/.*\.eval\.golden/ {
+    file = substr($2,3)
+    seen = 0
+    next
+  }
+  # capture old values
+  /^-[A-Za-z ]+:[[:space:]]+[0-9_]+/ {
+    match($0,/^-([A-Za-z ]+):[[:space:]]+([0-9_]+)/,m)
+    label=m[1]; if (!(label in known)) next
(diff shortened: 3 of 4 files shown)


SeungheonOh:
Okay, I proceeded with the second option, and it turned out much nicer than I originally anticipated. I force-pushed the change and it works. I think this setup can be used for other things as well; this gives complete control over how constructors and matchers operate.

In terms of performance difference, I can say there is none besides a few bytes. This makes sense since we are using the same `case` statement on either case. But for cleanliness, I think it's still better.

plutus-tx-plugin/src/PlutusTx/Compiler/Type.hs
```diff
@@ -223,6 +223,7 @@ clear natural ordering which we will assume GHC respects.
 -}
 
 {- Note [Ensuring compatibility with spec and stdlib types]
+TODO: ask question and update.
```
SeungheonOh, line 226: Is there any other places that uses constructor order of `Bool`? Now that we are using builtin bool, we can just change order when constructor and matcher are defined if this is only used for defining datatype.

SeungheonOh:
I removed all uses of old SOP based bools. This turned out to bring some improvements on budgets since all of the builtin functions can now return builtin bool directly without constructing SOP bool. 

However, some scripts suffered some degradation. I'm not entirely sure what's happening. My guess is we are missing out on some PIR optimizations since now we are not using PIR data and we have CaseOfCase turned off.

plutus-tx-plugin/test/AsData/Budget/9.6/destructSum-manual.eval.golden
```diff
@@ -1,6 +1,6 @@
-CPU:                 6_821_121
-Memory:                 24_722
-Term Size:                 212
-Flat Size:                 218
+CPU:                 7_253_244
+Memory:                 25_387
```
SeungheonOh: performance degradation here and on few other cases, but I think this is most severe

plutus-tx-plugin/test/AsData/Budget/9.6/destructSum-manual.pir.golden
```diff
@@ -34,30 +34,21 @@ in
...
     [ (/\dead ->
          let
            !tup : pair integer (list data) = unConstrData d
          in
          case
            (all dead. data)
-           (case
-              bool
-              (equalsInteger 0 (fstPair {integer} {list data} tup))
-              [False, True])
```
SeungheonOh: this should be `equals 1`. A bug on test case.

SeungheonOh: fixed

plutus-tx-plugin/test/AsData/Budget/9.6/destructSum.uplc.golden
```diff
@@ -2,54 +2,49 @@
...
-                                        (force headList (force tailList l))
-                                        (\x y z w ->
-                                           constrData
-                                             0
+         case
+           (equalsInteger 0 (force (force fstPair) cse))
+           [ (case
+                (equalsInteger 1 (force (force fstPair) cse))
+                [ (case
+                     (equalsInteger 2 (force (force fstPair) cse))
```
SeungheonOh: So, it's not pulling out `(force (force fstPair) cse)`

plutus-tx-plugin/test/AsData/Budget/9.6/destructSum-manual.eval.golden
SeungheonOh: nvm, we just needed more CSE iterations

plutus-tx-plugin/test/AsData/Budget/9.6/destructSum-manual.eval.golden
```diff
@@ -1,6 +1,6 @@
-CPU:                 6_821_121
-Memory:                 24_722
-Term Size:                 212
-Flat Size:                 218
+CPU:                 6_889_454
```
SeungheonOh, line 1: This increase is from fixing bug in the original test case.

SeungheonOh:
I think most(probably all) of performance degradation is due to CSE not having enough iteration.

plutus-tx-plugin/test/AsData/Budget/Spec.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE TemplateHaskell #-}
 {-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:context-level=0 #-}
 {-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:defer-errors #-}
+{-# OPTIONS_GHC -fplugin-opt PlutusTx.Plugin:max-cse-iterations=40 #-}
```
zliu41: why so many rounds of CSE?

plutus-core/plutus-ir/src/PlutusIR/Compiler/Definitions.hs
```diff
@@ -311,3 +350,27 @@ lookupDestructor x name = do
   pure $ case Map.lookup name ds of
     Just (PLC.Def{PLC.defVal = (Datatype _ _ _ destr _)}, _) -> Just $ Var x destr
     Nothing                                                  -> Nothing
+
+lookupManualDestructor ::
```
zliu41: Since this looks up both manual destructors and regular destructors, you can combine this with `lookupDestructor`, and keep the `lookupDestructor` name.

plutus-tx/src/PlutusTx/IsData/Class.hs
```diff
@@ -102,6 +103,37 @@ instance UnsafeFromData Builtins.BuiltinByteString where
   {-# INLINEABLE unsafeFromBuiltinData #-}
   unsafeFromBuiltinData = BI.unsafeDataAsB
 
+instance ToData Bool where
```
zliu41: What's the difference between this instance and the one generated by `makeIsDataSchemaIndexed`?

plutus-tx-plugin/src/PlutusTx/Compiler/Type.hs
```diff
@@ -325,14 +326,14 @@ getMatch tc = do
 {-| Get the matcher of the given 'Type' (which must be equal to a type constructor application)
 as a PLC term instantiated for the type constructor argument types.
 -}
-getMatchInstantiated :: (CompilingDefault uni fun m ann) => GHC.Type -> m (PIRTerm uni fun)
+getMatchInstantiated :: (CompilingDefault uni fun m ann) => GHC.Type -> m (PIR.Term PIR.TyName PIR.Name uni fun Ann -> PIR.Type PIR.TyName uni Ann -> [PIR.Term PIR.TyName PIR.Name uni fun Ann] -> PIR.Term PIR.TyName PIR.Name uni fun Ann)
```
zliu41: Please format this better

plutus-tx/src/PlutusTx/Builtins/Internal.hs
```diff
@@ -113,18 +113,10 @@ BOOL
 -}
 
 -- See Note [Opaque builtin types]
```
zliu41: Remove this as it no longer applies here

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -253,10 +253,10 @@ instance HasFromOpaque BuiltinUnit () where
   {-# INLINEABLE fromOpaque #-}
 
 instance HasToOpaque Bool BuiltinBool where
-  toOpaque b = if b then true else false
+  toOpaque b = b
```
zliu41: This is redundant - it's covered by the default method

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -253,10 +253,10 @@ instance HasFromOpaque BuiltinUnit () where
...
   {-# INLINEABLE toOpaque #-}
 instance HasFromOpaque BuiltinBool Bool where
-  fromOpaque b = ifThenElse b True False
+  fromOpaque b = b
```
zliu41: ditto

plutus-tx-plugin/test/Budget/9.6/lte0.pir.golden
```diff
@@ -1,39 +1,31 @@
...
     = \(n : integer) ->
-        Bool_match
-          (case Bool (lessThanEqualsInteger n 0) [False, True])
-          {all dead. List integer}
-          (/\dead -> Nil {integer})
-          (/\dead -> Cons {integer} 0 (go (subtractInteger n 1)))
+        case
+          (all dead. List integer)
+          (lessThanEqualsInteger n 0)
+          [ (/\dead -> Cons {integer} 0 (go (subtractInteger n 1)))
```
zliu41, line 27: The `/\dead ->` is no longer needed since `case` is already non-strict on the branches. It's Ok to leave this for a separate PR.

plutus-benchmark/nofib/test/9.6/queens4-bt.eval.golden
```diff
@@ -1,7 +1,7 @@
-CPU:             3_372_957_578
-Memory:             18_765_469
-Term Size:               1_781
-Flat Size:               1_760
+CPU:             3_363_965_578
+Memory:             18_709_269
+Term Size:               1_791
+Flat Size:               1_764
```
zliu41, line 4: Any idea why sometimes both term size and flat size get bigger?

plutus-tx-plugin/test/AsData/Budget/Spec.hs
SeungheonOh: Ops, this isn't meant to be pushed.

plutus-benchmark/nofib/test/9.6/queens4-bt.eval.golden
SeungheonOh: having extra CSE passes seems to increase the term size. I'm not entirely sure.

plutus-tx/src/PlutusTx/IsData/Class.hs
SeungheonOh: There's not. Thanks. I'll remove

SeungheonOh:
> I had hoped to merge datatypeDefs and manualDatatypeDefs, but I'm not surprised that it's not straightforward.

This can be done, and frankly, not too difficultly. It's just that we don't really need that amount of abstraction here since the only intended use cases for `manualDatatypeDefs` are just builtin types.

plutus-tx-plugin/test/Budget/9.6/lte0.pir.golden
SeungheonOh: I tried, this but it doesn't improve much. And actually hurt in some cases. It's probably messing with optimization. We can come back to this later

plutus-tx/src/PlutusTx/IsData/Class.hs
SeungheonOh: What I had was slightly more efficient. But I think I prefer consistency here.

plutus-ledger-api/test-plugin/Spec/Budget/9.6/geq5.eval.golden
```diff
@@ -1,6 +1,6 @@
...
-Term Size:                 562
-Flat Size:                 933
+CPU:                86_838_948
+Memory:                519_419
+Term Size:                 556
+Flat Size:                 925
 
-(constr 0)
\ No newline at end of file
+(con bool True)
```
effectfully: `constr 0` used to mean `True`? That's unfortunate.

plutus-ledger-api/test-plugin/Spec/Data/ScriptContext/9.6/succeedsIfHasDatum.uplc.golden
```diff
@@ -26,11 +26,11 @@
...
                                      (force (force sndPair) tup)))))
-                         [(\ds -> constr 0 []), (constr 1 [])]) ])
+                         [(\ds -> True), False]) ])
                  (unConstrData
                     (force headList
                        (force tailList
                           (force tailList
                              (force (force sndPair) (unConstrData d)))))))
-              [(delay ()), (delay (traceError "PT5"))])
+              [(delay (traceError "PT5")), (delay ())])
```
effectfully, line 35: This is funny. The reason why those `delay`s are not optimized away is that `force` can't see past a function application that we have there. The force-delay optimization understands that, but the force-case-delay one doesn't. This is one reason why I prefer to merge optimization together instead of keeping them separate.

plutus-tx-plugin/test/AsData/Budget/9.6/destructSum-manual.uplc.golden
```diff
@@ -3,62 +3,56 @@
...
-                                              (\x y z w ->
-                                                 constrData
-                                                   0
+            case
+              (equalsInteger 0 cse)
+              [ (case
+                   (equalsInteger 1 cse)
+                   [ (case
+                        (equalsInteger 2 cse)
+                        [ ((\cse -> case cse [cse]) error)
```
effectfully, line 12: We should have an optimization that rewrites this (and anything else that forces `error`) to `error`.

plutus-tx-plugin/test/Budget/9.6/lte0.pir.golden
effectfully: It doesn't really matter, because it's optimized away in UPLC anyway.

effectfully (approved):
Not gonna lie, I expected this to have a much bigger impact. Well, that's a shame. Thanks for doing it!

Didn't review any of the plugin code, just the resulting UPLC, which didn't tell me anything about whether it's possible to do this any better.

plutus-tx-plugin/src/PlutusTx/Compiler/Type.hs
zliu41: > Is there any other places that uses constructor order of Bool?

I doubt it.

plutus-benchmark/nofib/test/9.6/queens4-bt.eval.golden
zliu41: How about leaving the number of CSE passes as it is, for an apple to apple comparison? You can increase the number, if it is overall beneficial, in a separate PR.

plutus-tx-plugin/test/Budget/9.6/lte0.pir.golden
zliu41: > It doesn't really matter, because it's optimized away in UPLC anyway.

Which UPLC optimization?

effectfully: `UntypedPlutusCore.Transform.ForceCaseDelay`

zliu41: Ah right, the latest optimization!

PR #7232, opened 2025-07-24
builtinsAvailableIn: revert from function to map

kwxm:
Experimenting with efficiency of `builtinsAvailableIn` after #7223.


plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
```diff
@@ -227,58 +228,42 @@ batch6 =
...
+  [ ((PlutusV1, alonzoPV),    Set.fromList batch1)
+  , ((PlutusV1, pv11PV),      Set.fromList (batch2 ++ batch3 ++ batch4 ++ batch5 ++ batch6))
+  , ((PlutusV2, vasilPV),     Set.fromList (batch1 ++ batch2))
+  , ((PlutusV2, valentinePV), Set.fromList batch3)
+  , ((PlutusV2, plominPV),    Set.fromList batch4b)
+  , ((PlutusV2, pv11PV) ,     Set.fromList (batch4a ++ batch5 ++ batch6))
+  , ((PlutusV3, changPV),     Set.fromList (batch1 ++ batch2 ++ batch3 ++ batch4))
+  , ((PlutusV3, plominPV),    Set.fromList batch5)
+  , ((PlutusV3, pv11PV),      Set.fromList batch6)
+  ]
```
effectfully, line 243: It's fine, but I kinda liked previous version more, which you can make efficient if you follow [this](https://github.com/IntersectMBO/plutus/pull/7223#discussion_r2229334293).

kwxm: I had to lift the separate batches out to the top level so that I could export them for use in the tests.  I'll come back and try to improve the efficiency in a separate PR: I merged the less efficient version first so that I could experiment separately.

Oh wait, do you mean the version with the funciton rather than the map?  It was originally a map but I switched to a function and that seems to have made it less efficient, so I was trying to confirm that and it seesm to be true.  I'm experimenting with other ways to improve efficency in other branches.

effectfully: Yeah, I mean you can keep it a function and have it efficient too.

kwxm:
I'm closing this because it's stale, but may come back and have another look at the original issue.

PR #7233, opened 2025-07-24
builtinsAvailableIn: try to use TH to ensure maps computed at compile time

kwxm:
This carries on from #7232 by trying to use Template Haskell to compute some maps statically.


kwxm:
I'm closing this because it's stale, but may come back and have another look at the original issue.

PR #7235, opened 2025-07-24
Use Data backed ledger types for COOP

SeungheonOh:
Changes COOP to use Data backed ledger types. 

Some scripts have reduced cost, some increased(!!). This is probably because how scripts are written originally; original scripts pass around many un-deconstructed data and deconstruct those when needed, perhaps this is very expensive here. 

@zliu41 Turns out `RecordWildCards` works perfectly fine with datatypes defined with `asData`. I was just being delusional.


plutus-benchmark/coop/test/9.6/fsMpMinting.eval.golden
```diff
@@ -1,6 +1,6 @@
-CPU:               783_914_853
-Memory:              3_572_506
-Term Size:               4_267
-Flat Size:              10_061
+CPU:             1_189_413_270
```
SeungheonOh: 🤯

plutus-benchmark/coop/test/9.6/mustBurnOwnSingleton.eval.golden
```diff
@@ -1,6 +1,3 @@
-CPU:               189_313_355
-Memory:                733_370
-Term Size:               1_687
-Flat Size:               4_341
-
-(con unit ())
\ No newline at end of file
+An error has occurred:
+The machine terminated because of an error, either from a built-in function or from an explicit use of 'error'.
+Caused by: error
```
SeungheonOh, line 3: No clue why this is. No script logic was changed. 🫠

plutus-benchmark/coop/src/PlutusBenchmark/Coop/Scripts.hs
```diff
@@ -210,7 +204,7 @@ authMp'
            "Must mint at least one $AUTH token:\n"
            <> "Must have a specified CurrencySymbol in the Value"
        Just tokenNameMap ->
-         case AssocMap.toList tokenNameMap of
+         case AssocMap.toSOPList tokenNameMap of
```
ana-pantilie: This may be one of the reasons there is a performance decrease. `Data.AssocMap.toSOPList` has to traverse the whole map, while the `AssocMap.toList` doesn't have any runtime cost. Can you avoid using `toSOPList` here?

plutus-benchmark/coop/src/PlutusBenchmark/Coop/Gen.hs
```diff
@@ -724,17 +725,17 @@ _doNothing = id
...
 normalizeValue v =
-  Value . AssocMap.safeFromList . Map.toList . (AssocMap.safeFromList . Map.toList <$>) $
+  Value . AssocMap.safeFromSOPList . Map.toList . (AssocMap.safeFromSOPList . Map.toList <$>) $
     Map.unionsWith
       (Map.unionWith (+))
       ( [ Map.singleton cs (Map.singleton tn q)
-        | (cs, tokens) <- AssocMap.toList . getValue $ v
-        , (tn, q) <- AssocMap.toList tokens
+        | (cs, tokens) <- AssocMap.toSOPList . getValue $ v
+        , (tn, q) <- AssocMap.toSOPList tokens
```
ana-pantilie, line 733: Again, Data <-> SOP conversions are expensive. Can this be avoided here? You won't be able to use the list comprehension, unfortunately, but there should be a way to write this code using builtins. Let me know if you need help with this and we can take a look together.

ana-pantilie (commented):
I think you need to try to write the code where you convert between Data and SOP without the conversions. There are many instances of this in the PR, I didn't comment on all of them. Let me know if you need any help with that.

plutus-benchmark/coop/src/PlutusBenchmark/Coop/Gen.hs
SeungheonOh: It doesn't really matter on this module because this module generates script context as Data. So all the inefficient computation is only on the haskell side.

SeungheonOh:
I can't really figure out why some of the scripts(fsMpMinting, fsMpBurning mainly) have slowed down. 

All that's changed from fsMp script is changing SOP list functions to Builtin function, but no clue.. I'm not gonna burn too much time here.

plutus-benchmark/coop/test/9.6/mustBurnOwnSingleton.golden.eval
```diff
@@ -1,6 +1,3 @@
-CPU:               114_113_035
-Memory:                579_805
-AST Size:                1_274
-Flat Size:               3_833
-
-(con unit ())
\ No newline at end of file
+An error has occurred:
+The machine terminated because of an error, either from a built-in function or from an explicit use of 'error'.
+Caused by: error
```
ana-pantilie, line 3: Looks like you've introduced a bug somewhere

ana-pantilie (changes requested):

plutus-benchmark/coop/test/9.6/mustBurnOwnSingleton.golden.eval
SeungheonOh: yep, can't really figure this out either. Only thing changed is using builtin list utilities instead of SOP one

PR #7239, opened 2025-07-25
Remove builtin check during deserialisation

kwxm:
This removes the check for allowed builtins that happens during deserialisation just so that I can see how much overhead it actually causes.


kwxm:
It's difficult to tell exactly, but it looks like the builtin availability check accounts for about 3-5% of the total deserialisation time in the current state of `master`.  I'll close this, but may come back after trying to improve `builtinsAvailableIn`.

PR #7242, opened 2025-07-28
Pretty-print UPLC programs with unicode chars

zeme-wana:


plutus-conformance/test-cases/uplc/evaluation/builtin/constant/string/string-07/string-07.uplc.expected
```diff
@@ -0,0 +1,17 @@
...
+  (con
+    (list string)
+    [ "z ∈ ℝ ⇒ x² ≥ 0; z ∈ ℂ"
+    , "x ∈ ℝ ⇒ x² ≥ 0; z ∈ ℂ\\ℝ ⇒ z² ∉ {x ∈ ℝ: x ≥ 0}."
+    , "\\"
+    , "\""
+    , "--- \"\"\"\" \\ \\ \\\\ \\\\\\ \""
+    , "\b\n\a\t"
+    , "\DEL"
+    , "----- \\"
```
zeme-wana, line 12: There is a problem with our UPLC parser: 
This string: 
```
"----- \\" -----"
```
Should be invalid and return a parse error, however it is truncated instead: 
```
"----- \\"
```
Is this behaviour expected?

plutus-conformance/test-cases/uplc/evaluation/builtin/constant/string/string-08/string-08.uplc
```diff
@@ -0,0 +1,3 @@
+(program 1.0.0
+	(con string "----- \\" -----"
```
zeme-wana: There is a problem with our UPLC parser:
This string:
```
"----- \\" -----"
```
Should be invalid and return a parse error, however it is truncated instead:
```
"----- \\"
```
Is this behaviour expected?

SeungheonOh:
```
(program 1.1.0 
  (case (con bool True) (con string "----- \\" hello") (con string "foobar")))
```
This fails. 

~~I think the current parser passes because it sees `-----"` this part as a comment.~~

~~We should make parser handle `\\"` correctly.~~

edit: This is a correct behavior. `\\"` should fail; only backslash is escaped and `"` is not.

SeungheonOh:
~~Actually, this might be better off fixing it on the parser side. Let me test a few things.~~

```
{- | Parser for string constants (wrapped in double quotes).  Note that
 Data.Text.pack "performs replacement on invalid scalar values", which means
 that Unicode surrogate code points (corresponding to integers in the range
 0xD800-0xDFFF) are converted to the Unicode replacement character U+FFFD
 (decimal 65533).  Thus `(con string "X\xD800Z")` parses to a `Text` object
 whose second character is U+FFFD.
-}
conText :: Parser T.Text
conText = lexeme . fmap T.pack $ char '\"' *> manyTill Lex.charLiteral (char '\"')
```

Discard this, `charLiteral` does take care of unicode and escape sequences.

plutus-conformance/test-cases/uplc/evaluation/builtin/constant/string/string-08/string-08.uplc
zeme-wana: Wait then it's working correctly, no?
```
"----- \\" hello"
```
The above is supposed to fail paring.

SeungheonOh:
I fixed the one of the test cases, but I don't know how `parser captures ending positions correctly` can be fixed. Frankly, I don't even understand why this broken in the first place. I'll continue tomorrow.

edit: okay, dug around a bit more. seems like megaparsec just always simply see 1 byte as one character, so it doesn't correctly generate srcspan for multi-byte unicode characters like 😄

SeungheonOh:
Consider:
```
ghci> (runQuoteT $ parse (withSpan $ \s -> (s,) <$> symbol "\128512\128512\128512") "mystring" "\128512\128512\128512") :: Either ParserErrorBundle (SrcSpan, Text)
Right (mystring:1:1-1:6,"\128512\128512\128512")
```

Look how we only processed 3 characters, but SrcSpan is lying and giving us 6. Actually, this is very odd; all haskell string types should correctly handle wide characters automatically

SeungheonOh:
I fixed the other issue as well. I had to vendor some megaparsec code, which isn't optimal :/ 

Problem was: currently, megaparsec will count unicode wide character as 2 column wide; however, we don't count wide characters as two columns in other places. I changed Parser behavior to count every character as 1 column. Rationale being: emacs counts the emoji "😄" as 1 column. 
image

Instead, I could've made length reporting on the test side to count wide characters as 2 column, but I don't know what we'd prefer more. Perhaps you might have better insight here @zliu41

plutus-core/plutus-core/src/PlutusCore/Parser/ParserCommon.hs
```diff
@@ -87,6 +91,69 @@ leadingWhitespace = (whitespace *>)
...
+         in if
+              | ch == newlineTok ->
+                  SourcePos n (l <> pos1) pos1
+              | ch == tabTok ->
+                  SourcePos n l (mkPos $ c' + w - ((c' - 1) `rem` w))
+              | otherwise ->
+                  SourcePos n l (c <> columnIncrement ch)
+{-# INLINE reachOffsetNoLine' #-}
+
+getSourcePos' :: MonadParsec e Text m => m SourcePos
```
SeungheonOh, line 142: I made a mess here. Please move/change appropriately

plutus-core/plutus-core/src/PlutusCore/Pretty/PrettyConst.hs
```diff
@@ -104,9 +105,20 @@ instance Show a => DefaultPrettyBy ConstConfig (PrettyAny a) where
...
+-- but doesn't escape Unicode characters (\8704 and so on).
+instance NonDefaultPrettyBy ConstConfig T.Text where
+    nonDefaultPrettyListBy conf = Prettyprinter.list . Prelude.map (nonDefaultPrettyBy conf)
+    nonDefaultPrettyBy = inContextM $ \t -> pure $ pretty $ "\"" <> escape t <> "\""
+        where
+            escape t = snd $ T.foldl' prettyChar (False, "") t
+            prettyChar (needSep, acc) c
+                | c == '"' = (False, acc <> "\\\"") -- Not handled by 'showLitChar'
+                | c == '\\' = (False, acc <> "\\\\") -- Not handled by 'showLitChar'
+                | Char.isNumber c && needSep = (False, acc <> "\\&" <> T.singleton c)
```
SeungheonOh: We have to do this silly thing because sometimes we want to have `\128\&0` that is `[128, 48]`. So we insert `\&` when we have number followed right after `\[0-9]`.

kwxm: Why do we need to do this?  `Show` for `Text` seems to do the right thing, so can we not use that to prettyprint strings/text instead of writing our own version?  
```
ghci> :m + Data.Char
ghci> :m + Data.Text
ghci> s = [chr 128, chr 48]  -- String
ghci> s
"\128\&0"
ghci> putStrLn $ show s
"\128\&0"
ghci> t = pack s  -- Text
ghci> t
"\128\&0"
ghci> putStrLn $ show t
"\128\&0"
```

SeungheonOh: Show instance for Text handles control sequences correctly, but it won't print Unicode characters correctly. 

So, `show "\128512" ` will be `\128512` instead of `😀`

zeme-wana: > Show instance for Text handles control sequences correctly, but it won't print Unicode characters correctly.

@kwxm this is the whole reason we are doing this. Current solutions seems messy, I'm looking into [this package](https://hackage.haskell.org/package/unicode-show-0.1.1.1/docs/Text-Show-Unicode.html) now

SeungheonOh: Okay, I fixed it and it's not messy anymore.

PR #7243, opened 2025-07-28
Add useful bash scripts

SeungheonOh:
`eval-percent-diff.sh` will print percent change for golden budgets. It can either compare it against current git changes or current git head against some target git commit hash provided as cli argument.  

`regen-goldens.sh` will run and regenerate all test cases that have goldens.

samples:
```
File: plutus-benchmark/coop/test/9.6/certMpBurning.eval.golden
  CPU:         -1.48%
  Memory:      -1.57%
  Term Size:   +0.00%
  Flat Size:   +0.00%

File: plutus-benchmark/coop/test/9.6/certMpMinting.eval.golden
  CPU:         -0.09%
  Memory:      -0.01%
  Term Size:   +0.00%
  Flat Size:   +0.00%

File: plutus-benchmark/coop/test/9.6/fsMpBurning.eval.golden
  CPU:         +0.04%
  Memory:      +0.18%
  Term Size:   +0.00%
  Flat Size:   +0.00%
```

### DIFF
--- a/scripts/regen-goldens.sh
+++ b/scripts/regen-goldens.sh
@@ -13,40 +13,31 @@ projects=(
 )
 
 tests_plutus_conformance=(
-  "cabal run haskell-conformance -- --accept"
-  "cabal run haskell-steppable-conformance -- --accept"
-  "cabal run agda-conformance -- --accept"
+  "cabal run haskell-conformance -- --accept --hide-successes"
+  "cabal run haskell-steppable-conformance -- --accept --hide-successes"
+  "cabal run agda-conformance -- --accept --hide-successes"
 )
 tests_plutus_tx_plugin=(
-  "cabal run plutus-tx-plugin-tests -- --accept"
-  "cabal run size -- --accept"
+  "cabal run plutus-tx-plugin-tests -- --accept --hide-successes"
+  "cabal run size -- --accept --hide-successes"
 )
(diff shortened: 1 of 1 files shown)


scripts/regen-goldens.sh
```diff
@@ -0,0 +1,71 @@
...
+tests_plutus_tx_plugin=(
+  "cabal run plutus-tx-plugin-tests -- --accept"
+  "cabal run size -- --accept"
+)
+tests_plutus_ledger_api=(
+  "cabal run plutus-ledger-api-test -- --accept"
+  "cabal run plutus-ledger-api-plugin-test -- --accept"
+)
+tests_plutus_benchmark=(
+  "cabal run plutus-benchmark-nofib-tests -- --accept"
```
zliu41: How about `cabal test plutus-benchmark --accept`, so we don't need to maintain this list?

zliu41 (approved):
This is super cool - nicely done! Could you make an issue to run `eval-percent-diff.sh` in a GH action on every push to a PR that updates `*.eval.golden`?

scripts/regen-goldens.sh
SeungheonOh: Good point, I'll change this and other packages as well

SeungheonOh:
#7248 supersedes this PR

PR #7246, opened 2025-07-29
Use `unicode-show` package to pretty print UPLC constants

zeme-wana:


SeungheonOh:
Looks like this is duplicate of #7242?

zeme-wana:
Yeah I just wanted to see if using the `unicode-show` package would get the job done, but it's better to do https://github.com/IntersectMBO/plutus/pull/7242 instead

PR #7247, opened 2025-07-29
Change case order for two branch builtin list casing

SeungheonOh:
Change case order for two bench builtin list, so how it's 
```
case <builtin list> 
  <cons branch>
  <nil branch>
```

@ana-pantilie We need this merged before we proceed with the release!

### DIFF
--- a/plutus-core/changelog.d/20250729_124809_me_reverse_list_case.md
+++ b/plutus-core/changelog.d/20250729_124809_me_reverse_list_case.md
@@ -1,0 +1,4 @@
+### Changed
+
+- When `Case`ing on builtin list with two branches, now the "cons" branch comes first, followed by "nil" branch. Previously, it was "nil" than "cons" branch. This was changed to give better consistency against the one branch casing which will just have "cons" branch.
+


kwxm (approved):
This all looks good.  I didn't check every single detail of the golden files, but I did scan through them and all of the changes that I saw were just things being rearranged as you'd expect.

ana-pantilie:
Can you explain the reason we're making this change?

kwxm:
> Can you explain the reason we're making this change?

@ana-pantilie It's because you're allowed either one or two branches and if you only have one branch and it's not taken then you get `error`.  In the two-branch case we have (after this change)

```
case xs
  <consBranch>
  <nilBranch>
```
but you can leave out the `<nilBranch>` just have

```
case xs
  <consBranch>
```
This fails if it gets an empty list, so it's equivalent to
```
case xs
  <consBranch>
  (error)
```
This lets you handle the case of failing on `[]` more efficiently.

Before the change, the nil case came first, so you had 
```
case xs
  <nilBranch>
  <consBranch>
```
and
```
case xs
  <consBranch>
```
with the latter being equiavalent to
```
case xs
  (error)
  <consBranch>
```
which was a bit strange.

This PR makes it so that you can leave out the branch at the end and get `(error)`, which also makes it more similar to casing on other builtins, like for `integer`, where you have
```
case x
  <zeroBranch>
  <oneBranch>
  <twoBranch>
  ...
```
You can have any number of branches there and if `x` is out of range you get an error.

See [this Slack thread](https://input-output-rnd.slack.com/archives/G01EE4NQ9U0/p1753176576053089).

PR #7252, opened 2025-07-31
Throw compilation error when unsupported extension is enabled

### LINKED ISSUE #7238: Better compiler error message for GADTs
```haskell
data Foo a where
  Foo1 :: Integer -> Foo String
  Foo2 :: Bool -> Foo Int

foo :: Foo String -> Bool
foo (Foo1 x) = x PlutusTx.== 42
{-# INLINABLE foo #-}
```

Compiling `foo` leads to the following error:

```
Error: Unsupported feature: Kind: forall k0 k1.
             k0 -> k1 -> GHC.Prim.CONSTRAINT GHC.Types.ZeroBitRep
```

which is not very informative.

SeungheonOh:
closes #7238 

This will make plugin error out compilation when it detects unsupported language turned on. 

Currently, it look for `GADTs`, `PolyKinds`, ~~`DataKinds`~~, `DerivingVia`, and ~~`GeneralizedNewtypeDeriving`~~ but we can add more if needed; I'm not sure if this is the full list.

### DIFF
--- a/plutus-tx-plugin/changelog.d/20250818_111253_me_7238.md
+++ b/plutus-tx-plugin/changelog.d/20250818_111253_me_7238.md
@@ -1,0 +1,4 @@
+### Changed
+
+- Compiliation will fail if `GADTs` or `PolyKinds` is used in the module. This is to providing better compilation error message. This also encourages clearer onchain/offchain separation by encouraging offchain components that uses `GADTs` or `PolyKinds` to be separated into a new module outside of onchain module.
+


SeungheonOh:
hmm this might not be the best idea. This is slightly more picky and will error out when the extensions are used in same Haskell file but not in the plugin placeholder

plutus-tx-plugin/src/PlutusTx/Plugin.hs
```diff
@@ -488,6 +491,27 @@ runCompiler
...
+    enabledExtensions =
+      mapMaybe
+        (\case
+            GHC.On a -> Just a
+            GHC.Off _ -> Nothing)
+        extensions
+    extensionBlacklist =
+      [ GADTs
+      , PolyKinds
+      , DerivingVia
```
zliu41: Why is DerivingVia problematic?

SeungheonOh: This list came from https://github.com/IntersectMBO/plutus/tree/master/plutus-tx#haskell-language-support

I'm still not sure if this a correct approach. This change will prevent use of all aforementioned extensions in a given module even if they are not part of what's being compiled into UPLC using the plugin. For example, GADTs for off chain purposes cannot be on the same file that will invoke the plugin. 

I can see this being both positive and negative. This is good since it will encourage more stricter separation between on and off chain, but I also see this being slightly too restrictive.

zliu41: > I'm still not sure if this a correct approach. This change will prevent use of all aforementioned extensions in a given module even if they are not part of what's being compiled into UPLC using the plugin.

Initially I thought so too, but I think it's better than the cryptic error messages that people currently get. If one has both Plinth code and regular Haskell code, it should be easy enough to separate them into different modules.

PR #7253, opened 2025-07-31
Benchmarking consistency

kwxm:
Let's see if there's any improvement after #7249.


kwxm:
Several jobs waited for a runner for 24 hours then timed out.

kwxm:
The results did seem to be pretty consistent, with occasaional runs uniformly slowewr than usual.  It doesn't seem worse than usual.

PR #7255, opened 2025-08-01
Value benchmarking

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1731


plutus-benchmark/values/bench/Bench.hs
```diff
@@ -0,0 +1,241 @@
...
+                [ bench "union - two different values"
+                    $ nf (uncurry NestedValue.union) vals
+                , bench "valueContains - is not sub-value"
+                    $ nf (uncurry NestedValue.valueContains) vals
+                ]
+        , env setupEnvFlattened2 $ \ ~vals ->
+            bgroup "FlattenedValue"
+                [ bench "union - two different values"
+                    $ nf (uncurry FlattenedValue.union) vals
+                , bench "valueContains - is not sub-value"
```
ana-pantilie, line 208: I just realised `contains` is broken for all these implementations, redoing this.

ana-pantilie: Fixed ^ and realized I forgot to push latest changes! This is updated now, I should take another look and regenerate the benchmark report

plutus-benchmark/values/bench/Bench.hs
```diff
@@ -0,0 +1,239 @@
...
+     in pure (mmValue1, mmValue1Inv)
+
+main :: IO ()
+main = do
+    config <- getConfig 30
+    defaultMainWith config $
+        [ env setupEnvNested1 $ \ ~nValue ->
+            bgroup "NestedValue"
+                [ bench "insertCoin - new coin"
+                    $ nf (NestedValue.insertCoin polId500 tokN10000 200) nValue
```
effectfully: You can't do this. `nf` is O(n) and `insertCoin` is what, O(log n)? The cost of your measurements is asymptotically higher than the cost of the operation whose cost you're trying to measure.

PR #7262, opened 2025-08-07
Make our repo buildable again with `-O0`.

bezirg:
I am developing with -O0 which can cut build times in most cases. This makes our repo buildable with -O0.

Also this patch might help with hls and lsp, not 100% sure.

If you don't likek this patch, I can drop it and only keep it locally.

### DIFF
--- a/doc/docusaurus/docusaurus-examples.cabal
+++ b/doc/docusaurus/docusaurus-examples.cabal
@@ -18,6 +18,7 @@ common lang
     -Wall -Wnoncanonical-monad-instances -Wincomplete-uni-patterns
     -Wincomplete-record-updates -Wredundant-constraints -Widentities
     -Wunused-packages -Wmissing-deriving-strategies
+    -fobject-code -fno-ignore-interface-pragmas -fno-omit-interface-pragmas
 
 common ghc-version-support
   -- See the section on GHC versions in CONTRIBUTING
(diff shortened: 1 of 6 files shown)


Unisay:
I don't quite understand:
- Why couldn't we use `-00` before.
- How commits in this PR solve the problem.

bezirg:
> I don't quite understand:
> 
> * Why couldn't we use `-00` before.
> * How commits in this PR solve the problem.

There were some compiler options missing on the cabal file.
And there were some GHC -O1 compiler optimisations that happened to fire and make the examples work (e.g. using haskell's div instead of plutus-prelude divide, etc)

bezirg:
Can somebody accept this? It is not a major change

Unisay:
The build is failing for me (`aarch64-darwin`) with errors like this one:
```
[60 of 66] Compiling ShortCircuit.Spec ( test/ShortCircuit/Spec.hs, /Users/yura/Projects/Cardano/plutus/dist-newstyle/build/aarch64-osx/ghc-9.6.7/plutus-tx-plugin-1.54.0.0/t/plutus-tx-plugin-tests/build/plutus-tx-plugin-tests/plutus-tx-plugin-tests-tmp/ShortCircuit/Spec.o )

<no location info>: error:
    GHC Core to PLC plugin: Error: Reference to a name which is not a local, a builtin, or an external INLINABLE function: Variable PlutusTx.Builtins.error
                               No unfolding
```

bezirg:
> I tried `cabal clean && cabal build all --ghc-option -O0` and it fails.

It used to work  :(

bezirg:
Actually I give up on it, the unfoldings are not there with -O0 .

PR #7263, opened 2025-08-08
Fix natural number flat encoding on the spec

SeungheonOh:
Spec fix.

### DIFF
--- a/doc/plutus-core-spec/flat-serialisation.tex
+++ b/doc/plutus-core-spec/flat-serialisation.tex
@@ -195,16 +195,17 @@ decoded list $l$ (from the remainder $s^{(k)}$) forms the final result.
 
 \subsection{Natural numbers}
 We encode natural numbers by splitting their binary representations into
-sequences of 7-bit blocks, then emitting these as a list with the \textbf{least
-  significant block first}:
+sequences of 7-bit blocks, then emitting all but last chunks as a list with
+the \textbf{least significant block first}. The last chunk(most significant chunk)
+will be encoded as regular 7-bit block and appended right after the list of chunks.
 
 $$
-\E_{\N} (s, \sum_{i=0}^{n-1}k_i2^{7i}) = \Elist_7(s, [k_0, \ldots, k_{n-1}])
+\E_{\N} (s, \sum_{i=0}^{n-1}k_i2^{7i}) = \E_7(\Elist_7(s, [k_0, \ldots, k_{n-2}]), k_{n-1})
 $$
 \noindent(where $k_i \in \Z$ and $0 \leq k_i \leq 127$).
 \noindent The decoder is
 $$
-\D_{\N}(s) = (s', \sum_{i=0}^{n-1}k_i2^{7i}) \quad \text{if $\Dlist_7(s) = (s', [k_0, \ldots, k_{n-1}])$}.
+\D_{\N}(s) = (s'', \sum_{i=0}^{n-1}k_i2^{7i}) \quad \text{if $\Dlist_7(s) = (s', [k_0, \ldots, k_{n-2}])$ and $\D_7(s') = (s'', k_{n-1})$}.
 $$
 
 \subsection{Integers}


doc/plutus-core-spec/flat-serialisation.tex
```diff
@@ -195,16 +195,17 @@ \subsection{Arrays}
 
 \subsection{Natural numbers}
 We encode natural numbers by splitting their binary representations into
-sequences of 7-bit blocks, then emitting these as a list with the \textbf{least
-  significant block first}:
+sequences of 7-bit blocks, then emitting all but last chunks as a list with
+the \textbf{least significant block first}. The last chunk(most significant chunk)
+will be encoded as regular 7-bit block and appended right after the list of chunks.
```
kwxm, line 198: I think this needs a sentence like "Every natural number can be written uniquely in the form `$\sum_{k=0}^nk_i2^{7i} with n \geq 0 and k_n \neq 0$`. Using this fact, we define <definition of the encoder> ... ", the point being that we want to (a) disallow an empty sum, and (b) also disallow leading zeros.   The decoder does accept natural numbers encoded with extera leasing zeros (ie, at the end of the list of blocks), but we probably want to specify a canonical way of encoding everything.

doc/plutus-core-spec/flat-serialisation.tex
```diff
@@ -195,16 +195,17 @@ \subsection{Arrays}
...
 
 $$
-\E_{\N} (s, \sum_{i=0}^{n-1}k_i2^{7i}) = \Elist_7(s, [k_0, \ldots, k_{n-1}])
+\E_{\N} (s, \sum_{i=0}^{n-1}k_i2^{7i}) = \E_7(\Elist_7(s, [k_0, \ldots, k_{n-2}]), k_{n-1})
```
kwxm: To deal with making the top index n:
```suggestion
\E_{\N} (s, \sum_{i=0}^{n}k_i2^{7i}) = \E_7(\Elist_7(s, [k_0, \ldots, k_{n-1}]), k_{n})
```

doc/plutus-core-spec/flat-serialisation.tex
```diff
@@ -195,16 +195,17 @@ \subsection{Arrays}
...
 $$
 \noindent(where $k_i \in \Z$ and $0 \leq k_i \leq 127$).
 \noindent The decoder is
 $$
-\D_{\N}(s) = (s', \sum_{i=0}^{n-1}k_i2^{7i}) \quad \text{if $\Dlist_7(s) = (s', [k_0, \ldots, k_{n-1}])$}.
+\D_{\N}(s) = (s'', \sum_{i=0}^{n-1}k_i2^{7i}) \quad \text{if $\Dlist_7(s) = (s', [k_0, \ldots, k_{n-2}])$ and $\D_7(s') = (s'', k_{n-1})$}.
```
kwxm:
```suggestion
\D_{\N}(s) = (s'', \sum_{i=0}^{n}k_i2^{7i}) \quad \text{if $\Dlist_7(s) = (s', [k_0, \ldots, k_{n-1}])$ and $\D_7(s') = (s'', k_{n})$}.
```

doc/plutus-core-spec/flat-serialisation.tex
```diff
@@ -195,16 +195,17 @@ \subsection{Arrays}
 
 \subsection{Natural numbers}
 We encode natural numbers by splitting their binary representations into
-sequences of 7-bit blocks, then emitting these as a list with the \textbf{least
-  significant block first}:
+sequences of 7-bit blocks, then emitting all but last chunks as a list with
+the \textbf{least significant block first}. The last chunk(most significant chunk)
```
kwxm:
```suggestion
the \textbf{least significant block first}. The last chunk (most significant chunk)
```

kwxm: We're calling things both blocks and chunks.  We should stick to one, probably "blocks" since I think we use "chunk" elsewhere to refer to things whose length may vary.

doc/plutus-core-spec/flat-serialisation.tex
kwxm: Now I'm wondering what happens if there's only one block, say if your number is 17.  In that case you get

D_list(s) = (s', []) and D_7(s') = (s'', 17)

so you'll get [k_0, ..., k_{n-1}] = [] and k_n = 17.  I suppose we could claim that [k_0, ..., k_{n-1}] = [] forces n=0 (but why not n=-5 or something, which would also give you the empty list?), and then the result is k_{0}2^{7*0}=17.  Maybe we need a case to deal specially with the empty list. We could also bypass the list decoder altogether and write down the full details of how the decoder for N works, but that might need quite a lot of extra notation.

doc/plutus-core-spec/flat-serialisation.tex
```diff
@@ -195,16 +195,17 @@ \subsection{Arrays}
...
+will be encoded as regular 7-bit block and appended right after the list of chunks.
```
kwxm: You could maybe add a sentence saying that this amounts to all of the leading blocks being prefixed by 1 and the final one by 0.  That seems a bit easier to grasp (but that might just be me).

kwxm (commented):
Thanks for noticing this.  I think a couple of things need to be fixed though.

SeungheonOh:
I tried to make it clearer, both on the "7 bit block sequence" part and single block cases. 

Rendered:
image

doc/plutus-core-spec/flat-serialisation.tex
SeungheonOh: I made explicit case for when there is only one block to make this clearer

doc/plutus-core-spec/flat-serialisation.tex
kwxm: > "Every natural number can be written uniquely in the form `$\sum_{k=0}^nk_i2^{7i} with n \geq 0 and k_n \neq 0$`.

Actually I got that wrong. For that representation to be unique we need 0 <= k_i <= 127 for 0 <= i <= n (so you rule out stuff like 129 = 129*2^0 = 1*2^0 + 1*2^7)).  You do say later that each k_i is a 7-bit binary block, so maybe that covers it.

SeungheonOh: > that each k_i is a 7-bit binary block

this was the idea, but I'll clarify

kwxm:
Two final things:
  *  Could you change `ks` to `l` or something please?  Everyhwere else we use one-character variable names and juxtaposition to represent multiplication, so `ks` looks like it means `k×s`.
  * Can you update the date [here](url) as well please?

kwxm:
> Two final things:

A third final thing: could you add `\noindent` before intuitively?

Intuitively

doc/plutus-core-spec/flat-serialisation.tex
```diff
@@ -194,19 +194,34 @@ \subsection{Arrays}
...
-\D_{\N}(s) = (s', \sum_{i=0}^{n-1}k_i2^{7i}) \quad \text{if $\Dlist_7(s) = (s', [k_0, \ldots, k_{n-1}])$}.
+\D_{\N}(s) =
+\begin{cases}
+  \text{let } \Dlist_7(s) = (s', ks) \\
+  D_7(s') & \text{if } ks = [] \\
+  (s'', \sum_{i=0}^{n}k_i2^{7i}) & \text{if } ks = [k_0, \ldots, k_{n-1}] \text{ and } \D_7(s') = (s'', k_{n})
+\end{cases}
 $$
 
+Intuitively, every 7 bit blocks except the last block are prefixed by 1 and the last block is prefixed by 0.
```
kwxm:
```suggestion
\noindent Intuitively, every 7 bit blocks except the last block are prefixed by 1 and the last block is prefixed by 0.
```

kwxm: ^ I did this to see if contributing to the PR would affect my ability to approve and/or merge it.

kwxm (approved):
I think we can merge this once you've made the two changes mentioned [here](https://github.com/IntersectMBO/plutus/pull/7263#issuecomment-3180353997).

kwxm:
I don't think you made the changes I asked for [here](https://github.com/IntersectMBO/plutus/pull/7263#issuecomment-3180353997).

PR #7271, opened 2025-08-13
Add Plinth builtin for integer casing

SeungheonOh:
Adds `caseInt :: forall a. Integer -> [a] -> a` to Plinth builtin; `caseInt` will directly insert `Case` node and use builtin integer casing to select branch. 

`IsData.TH` derivation, specifically one for unsafeFromData, is modified so that it will use `caseInt` when possible.


cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests/unsorted.pir.golden
```diff
@@ -423,20 +423,12 @@ program
...
-                            {a}
-                            (`$dUnsafeFromData` (headList {data} args))) ]
-                     {all dead. dead})
-              , (/\dead -> Nothing {a}) ]
-              {all dead. dead}
+              (list data -> Maybe a)
+              index
+              [ (\(ds : list data) ->
+                   Just {a} (`$dUnsafeFromData` (headList {data} ds)))
+              , (\(ds : list data) -> Nothing {a}) ]
```
effectfully, line 430: Would it make sense to pull `\(ds : list data) -> <...>` out of `case`, given that it appears in both the branches? If only to save on size. Not as a general optimization, because it's semantics changing, since `case` may fail, but can we just change the relevant Plinth code or whatever?

That'd also play better with inlining, case-of-case and whatnot.

plutus-tx/src/PlutusTx/Builtins/Internal.hs
```diff
@@ -1058,3 +1058,7 @@ expModInteger b e m =
     BuiltinSuccess bs -> toInteger bs
     BuiltinSuccessWithLogs logs bs -> traceAll logs $ toInteger bs
 {-# OPAQUE expModInteger #-}
+
+caseInteger :: Integer -> [a] -> a
+caseInteger i b = b !! fromIntegral i
+{-# OPAQUE caseInteger #-}
```
effectfully, line 1064: This has an issue that `fromIntegral` can under/overflow, unlike the UPLC `case`. Probably doesn't matter, but worth mentioning.

Thanks for calling it `caseInteger` and not `caseInt` by the way.

effectfully (approved):
Didn't review the TH code, just the compilation results.

Great optimization overall.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -852,6 +854,35 @@ entryExitTracing lamName displayName e ty =
...
+   . (CompilingDefault uni fun m ann)
+  => GHC.CoreExpr
+  -> m [PIRTerm uni fun]
+takeHaskellList = withOrWithoutTick buildList
+  where
+    err =
+      throwPlain $
+        CompilationError "Unexpected expression is given where to be Haskell list constructor is expected"
+
+    withOrWithoutTick :: (GHC.CoreExpr -> a) -> GHC.CoreExpr -> a
```
zliu41: There's a helper function `strip` for removing the ticks

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -852,6 +854,35 @@ entryExitTracing lamName displayName e ty =
    an unfolding.
 -}
 
+takeHaskellList
```
zliu41: This is a strange name. Perhaps "compileCaseIntegerBranches"?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -852,6 +854,35 @@ entryExitTracing lamName displayName e ty =
...
+    err =
+      throwPlain $
+        CompilationError "Unexpected expression is given where to be Haskell list constructor is expected"
+
+    withOrWithoutTick :: (GHC.CoreExpr -> a) -> GHC.CoreExpr -> a
+    withOrWithoutTick f (GHC.Tick _ e) = f e
+    withOrWithoutTick f e              = f e
+
+    buildList (GHC.App (GHC.App (GHC.App (GHC.Var _con) _ty) e) _) = traverse compileExpr [e]
+    buildList (GHC.App (GHC.App _build _ty) (GHC.Lam _tyArg (GHC.Lam con (GHC.Lam nil li)))) =
```
zliu41, line 875: Add some comments, showing examples of what the CoreExpr looks like in different cases (empty list, singleton, multiple elements).

zliu41 (changes requested):
Need a changelog entry. This should also be added to the Plinth user guide (can be a separate PR).

Also, it would be great if we can identify code that doesn't use this new primitive, but can still be compiled to casing on integers. Could you make an issue?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
SeungheonOh: Naming was intentionally made a bit generic because this can be used for other things as well; For example, 

`ands :: [Bool] -> Bool`

that compiles given entries into `p1 && p2 && p3 ... && pn`. 

I'll see if I can come up with a better name

plutus-tx/src/PlutusTx/Builtins/Internal.hs
SeungheonOh: > Thanks for calling it caseInteger and not caseInt by the way.

What difference does this make?

effectfully: `Int` is a separate type, I don't think that saving 4 characters would justify the potential for confusion.

cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests/unsorted.pir.golden
SeungheonOh: This is quite a bit of more work. What we generate using TH is essentially a case matching with a ton of view patterns attached. Which is useful for the "safe" fromData, but not so much for unsafe one. 

Unfortunately, I'd had to redo pretty big chunk of TH to do this since many part of it is very intertwined. I prefer to have this as a separate PR, if we decide that we want this.

effectfully: Fair enough.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -852,6 +854,37 @@ entryExitTracing lamName displayName e ty =
...
+compileHaskellList
+  :: forall uni fun m ann
+   . (CompilingDefault uni fun m ann)
+  => GHC.CoreExpr
+  -> m [PIRTerm uni fun]
+compileHaskellList = buildList . strip
+  where
+    err =
+      throwPlain $
+        CompilationError "Unexpected expression is given where to be Haskell list constructor is expected"
```
zliu41: This sentence isn’t grammatically correct. Also, I would separate the two cases:
- The CoreExpr represents an empty list. This is a user error.
- The CoreExpr doesn't represent a list at all. This is a plugin bug.

SeungheonOh: > The CoreExpr doesn't represent a list at all

This isn't a plugin bug, we expect this to fail if something like `(id [1, 2, 3])` is given as well.

Issue #6816, opened 2025-01-31
Or-patternify the codebase


effectfully:
It's extremely easy to introduce a bug by adding a new binder to any of the ASTs because of functions like

```haskell
termSubstClosedTerm
    :: Eq name
    => name
    -> Term tyname name uni fun a
    -> Term tyname name uni fun a
    -> Term tyname name uni fun a
termSubstClosedTerm varFor new = go where
    go = \case
         Var    a var         -> if var == varFor then new else Var a var
         LamAbs a var ty body -> LamAbs a var ty (goUnder var body)
         t                    -> t & over termSubterms go
    goUnder var term = if var == varFor then term else go term
```

I added `Fix` to the AST and this function didn't give me any warning or error, it just started silently doing the wrong thing. And we have a number of those scattered over the entire codebase. This means that adding any kind of binder to any kind of AST is now blocked on us cleaning up this mess, because otherwise we'll just end up introducing bugs.

What we should do is rewrite the last clause to

```haskell
termSubstClosedTerm
    :: Eq name
    => name
    -> Term tyname name uni fun a
    -> Term tyname name uni fun a
    -> Term tyname name uni fun a
termSubstClosedTerm varFor new = go where
    go = \case
         Var a var -> if var == varFor then new else Var a var
         LamAbs a var ty body -> LamAbs a var ty (goUnder var body)
         t@(Apply{}; TyAbs{}; ... Error{}) -> t & over termSubterms go
    goUnder var term = if var == varFor then term else go term
```

i.e. explicitly list all the constructors in an Or-pattern (potentially pattern-synonym'd), so that adding new constructors forces the programmer to make a choice regarding how to handle the new constructor.

Or-patterns are only available since GHC-9.12 though.

Issue #3876, opened 2021-09-06
'unsafeEvaluateCek' is overly lazy


effectfully:
## Area

[x] *Plutus Foundation* Related to the GHC plugin, Haskell-to-Plutus compiler, on-chain code
[] *Plutus Application Framework* Related to the Plutus application backend (PAB), emulator, Plutus libraries
[] *Marlowe* Related to Marlowe
[] *Other* Any other topic (Playgrounds, etc.)

So we have this issue with `unsafeEvaluateCek` not forcing the final result of a builtin application when the result of `unsafeEvaluateCek` is forced. I thought it might be some strictness problem within the definition of the CEK machine, however after inspecting and reinspecting all of its internals, I wasn't able to spot any problem, everything there looks exactly like it's supposed to. Then I looked into `runCekM` and nothing there seemed suspicious as well. So I started inserting bangs everywhere until I made the CEK machine so stupidly strict that it started forcing the result of a builtin application before doing budgeting for it and yet it still wasn't strict enough somehow. At this point I was convinced that the problem is not in the definition of the CEK machine, but rather how it was invoked. And indeed, `unsafeEvaluateCek` is defined like this:

```haskell
unsafeEvaluateCek emitTime params = first unsafeExtractEvaluationResult . evaluateCek emitTime params
```

and this seemingly innocuous definition is our culprit. `first` uses an irrefutable match (via [`bimap`](https://hackage.haskell.org/package/base-4.14.0.0/docs/src/Data.Bifunctor.html#local-6989586621679537087)) for `(,)` (i.e. matches lazily on the spine of the tuple), which causes all subsequent forcing to WHNF not to force anything. Compare

```
>>> (error "uh-oh" :: (Int, Char)) `seq` True
*** Exception: uh-oh
```

with

```
>>> first succ (error "uh-oh" :: (Int, Char)) `seq` True
True
```

So in our case forcing the result of `unsafeEvaluateCek` to WHNF does not do anything AT ALL, no evaluation happens whatsoever.

Solution? Never use `bimap`, `first` and `second` over `(,)`: if laziness is required, use something with a more suggestive name, if it's not, `bimap` will do the wrong thing and so you should use the strict version of the function.

And yeah, in case you're curious, the `Functor` instance for `(,) a` matches [strictly](https://hackage.haskell.org/package/base-4.14.0.0/docs/src/GHC.Base.html#line-983) on the spine of the tuple, just for some extra confusion.

[...]

PR #7272, opened 2025-08-13
[Evaluation] Add 'CekResult' and improve 'dischargeCekValue'

effectfully:
This is something I've been meaning to do for a long time. It adds `CekResult` and `CekReport` so that we never run into very confusing issues with excessive laziness invalidating benchmarking results like it happened in #3876. Plus, 3-tuples are an anti-pattern anyway, so it's good to move from them.

I also fixed a bug in `dischargeCekValue` where it previously wouldn't discharge under `Constr` or `Case` due to a catch-all clause that we forgot to update when introducing SOPs. It doesn't matter on-chain (we only care there if something is a `()` or success in general, not whether `Constr` was discharges correctly), but since we use the CEK machine as a sorta normalizer in tests, we still want it to return the correct answer.

Catch-all clauses are cancer, this is not the first time they're causing problems, see #6816. I'm gonna remove as many of them from the codebase as possible, but in the meantime please be wary not to add more of this cancer to the codebase.

### DIFF
--- a/cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests.hs
+++ b/cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests.hs
@@ -1,4 +1,6 @@
 -- editorconfig-checker-disable-file
+{-# LANGUAGE GADTs #-}
+
 module Cardano.Constitution.Validator.Data.GoldenTests
     ( tests
     ) where
@@ -7,6 +9,7 @@ import Cardano.Constitution.Config
 import Cardano.Constitution.Data.Validator
 import Cardano.Constitution.Validator.TestsCommon
 import Helpers.TestBuilders
+import PlutusCore.Default as UPLC
 import PlutusCore.Evaluation.Machine.ExBudget
 import PlutusCore.Evaluation.Machine.ExBudgetingDefaults
 import PlutusCore.Pretty (prettyPlcReadableSimple)
@@ -89,5 +92,5 @@ runForBudget v ctx =
         in case UPLC.runCekDeBruijn defaultCekParametersForTesting counting noEmitter vPs of
                 -- Here, we guard against the case that a ConstitutionValidator **FAILS EARLY** (for some reason),
                 -- resulting in misleading low budget costs.
-                (Left _, _, _) -> error "For safety, we only compare budget of succesful executions."
-                (Right _ , UPLC.CountingSt budget, _) -> budget
+                UPLC.CekReport (UPLC.CekSuccessConstant (UPLC.Some (UPLC.ValueOf UPLC.DefaultUniUnit ()))) (UPLC.CountingSt budget) _ -> budget
+                _ -> error "For safety, we only compare budget of succesful executions."
--- a/plutus-benchmark/common/PlutusBenchmark/Common.hs
+++ b/plutus-benchmark/common/PlutusBenchmark/Common.hs
@@ -119,10 +119,10 @@ evaluateCekLikeInProd
     -> Either
             (UPLC.CekEvaluationException UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun)
             (UPLC.Term UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun ())
-evaluateCekLikeInProd evalCtx term = do
+evaluateCekLikeInProd evalCtx term =
     let -- The validation benchmarks were all created from PlutusV1 scripts
         pv = LedgerApi.ledgerLanguageIntroducedIn LedgerApi.PlutusV1
-    Cek.cekResultToEither . Cek._cekReportResult $
+    in Cek.cekResultToEither . Cek._cekReportResult $
         LedgerApi.evaluateTerm UPLC.restrictingEnormous pv LedgerApi.Quiet evalCtx term
 
 -- | Evaluate a term and either throw if evaluation fails or discard the result and return '()'.
--- a/plutus-benchmark/common/PlutusBenchmark/Common.hs
+++ b/plutus-benchmark/common/PlutusBenchmark/Common.hs
@@ -79,7 +79,7 @@ getConfig limit = do
 getCostsCek :: UPLC.Program UPLC.NamedDeBruijn DefaultUni DefaultFun () -> (Integer, Integer)
 getCostsCek (UPLC.Program _ _ prog) =
     case Cek.runCekDeBruijn PLC.defaultCekParametersForTesting Cek.tallying Cek.noEmitter prog of
-      (_res, Cek.TallyingSt _ budget, _logs) ->
+      Cek.CekReport _res (Cek.TallyingSt _ budget) _logs ->
           let ExBudget (ExCPU cpu)(ExMemory mem) = budget
           in (fromSatInt cpu, fromSatInt mem)
 
@@ -120,11 +120,10 @@ evaluateCekLikeInProd
             (UPLC.CekEvaluationException UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun)
             (UPLC.Term UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun ())
 evaluateCekLikeInProd evalCtx term = do
-    let (getRes, _, _) =
-            let -- The validation benchmarks were all created from PlutusV1 scripts
-                pv = LedgerApi.ledgerLanguageIntroducedIn LedgerApi.PlutusV1
-            in LedgerApi.evaluateTerm UPLC.restrictingEnormous pv LedgerApi.Quiet evalCtx term
-    getRes
+    let -- The validation benchmarks were all created from PlutusV1 scripts
+        pv = LedgerApi.ledgerLanguageIntroducedIn LedgerApi.PlutusV1
+    Cek.cekResultToEither . Cek._cekReportResult $
+        LedgerApi.evaluateTerm UPLC.restrictingEnormous pv LedgerApi.Quiet evalCtx term
 
 -- | Evaluate a term and either throw if evaluation fails or discard the result and return '()'.
 -- Useful for benchmarking.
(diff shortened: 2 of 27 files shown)


plutus-core/plutus-core/src/PlutusCore/Evaluation/Result.hs
```diff
@@ -21,8 +21,6 @@ import PlutusCore.Pretty
 import Control.Monad.Except (MonadError, catchError, throwError)
 
 -- | The parameterized type of results various evaluation engines return.
--- On the PLC side this becomes (via @makeKnown@) either a call to 'Error' or
--- a value of the PLC counterpart of type @a@.
```
effectfully, line 25: I disabled `makeKnown` for `EvaluationSuccess` a long time ago, so this comment has been a lie for quite a while.

plutus-tx-plugin/test/CallTrace/9.6/func05.eval.golden
```diff
@@ -7,4 +7,4 @@ Trace:
 -> $cmyClassFunc (test/CallTrace/Spec.hs:99:3-99:13)
 -> functionFromOtherModule (test/CallTrace/Spec.hs:93:1-93:23)
 -> $
--> errorWhenTrue (test/CallTrace/OtherModule.hs:9:1-9:13)
\ No newline at end of file
+-> errorWhenTrue
```
effectfully: I don't know what this is. I've rebased on `master` and currently rerunning the tests.

cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests.hs
```diff
@@ -89,5 +92,5 @@ runForBudget v ctx =
         in case UPLC.runCekDeBruijn defaultCekParametersForTesting counting noEmitter vPs of
                 -- Here, we guard against the case that a ConstitutionValidator **FAILS EARLY** (for some reason),
                 -- resulting in misleading low budget costs.
-                (Left _, _, _) -> error "For safety, we only compare budget of succesful executions."
-                (Right _ , UPLC.CountingSt budget, _) -> budget
+                UPLC.CekReport (UPLC.CekSuccessConstant (UPLC.Some (UPLC.ValueOf UPLC.DefaultUniUnit ()))) (UPLC.CountingSt budget) _ -> budget
+                _ -> error "For safety, we only compare budget of succesful executions."
```
zliu41:
```suggestion
                _ -> error "For safety, we only compare budgets of successful executions."
```

plutus-tx-plugin/test/CallTrace/9.6/func05.eval.golden
effectfully: Seems like it resolved itself.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -163,6 +170,38 @@ this can make a surprisingly large difference.
...
+data CekReport cost name uni fun = CekReport
+    { _cekReportResult :: CekResult name uni fun  -- ^ The result of evaluation.
+    , _cekReportCost   :: cost                    -- ^ The final @cost@ value.
+    , _cekReportLogs   :: [Text]                  -- ^ Logs emitted during evaluation.
+    }
+
+-- | Convert the given 'CekResult' into an 'Either'.
+-- This is useful, because in the ledger API we care whether the result is a constant or not, but in
+-- tests, executables etc we don't and so handling an either-error-or-term is more natural.
+cekResultToEither
```
zliu41, line 189: put this in testlib?

zliu41 (approved):
Looks good. I don't know why `dischargeCekValue` shouldn't just preserve the `ann` in `CekValue`, but that's orthogonal to this PR.

zliu41:
Also I suppose you'll run the benchmarks before merging

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: It's used for convenient APIs, so not just in tests:

image

plutus-benchmark/common/PlutusBenchmark/Common.hs
```diff
@@ -120,11 +120,10 @@ evaluateCekLikeInProd
             (UPLC.CekEvaluationException UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun)
             (UPLC.Term UPLC.NamedDeBruijn UPLC.DefaultUni UPLC.DefaultFun ())
 evaluateCekLikeInProd evalCtx term = do
-    let (getRes, _, _) =
-            let -- The validation benchmarks were all created from PlutusV1 scripts
-                pv = LedgerApi.ledgerLanguageIntroducedIn LedgerApi.PlutusV1
-            in LedgerApi.evaluateTerm UPLC.restrictingEnormous pv LedgerApi.Quiet evalCtx term
-    getRes
+    let -- The validation benchmarks were all created from PlutusV1 scripts
```
kwxm, line 123: [Just for the sake of saying something] 

You could remove the `do`  and have `let ... in ...` instead.  That would save me from having to remember what's monadic about `Either`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -534,57 +573,77 @@ instance Pretty CekUserError where
...
+            -- The index on the binder is meaningless, we put `0` by convention, see 'Binder'.
+            LamAbs () (NamedDeBruijn n deBruijnInitIndex) $ goValEnv env 1 body
+        -- We only return a discharged builtin application when (a) it's being returned by the
+        -- machine, or (b) it's needed for an error message.
+        -- @term@ is fully discharged, so we can return it directly without any further discharging.
+        VBuiltin _ term _ -> term
+        VConstr ind args -> Constr () ind . map goValue $ argStackToList args
+
+    -- Instantiate all the free variables of a term by looking them up in an environment.
+    -- Mutually recursive with dischargeCekVal.
```
SeungheonOh: mutually recursive with `goValue`?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -534,57 +573,77 @@ instance Pretty CekUserError where
...
-               dischargeCekValue
-               -- index relative to (as seen from the point of view of) the environment
-               (Env.indexOne valEnv $ idx - lamCnt)
-    Apply ann fun arg    -> Apply ann (go lamCnt fun) $ go lamCnt arg
-    Delay ann term       -> Delay ann $ go lamCnt term
-    Force ann term       -> Force ann $ go lamCnt term
-    t -> t
+-- | Convert the given 'ArgStack' to a list by reversing it.
+argStackToList :: ArgStack uni fun ann -> [CekValue uni fun ann]
+argStackToList = go [] where
```
kwxm, line 578: Would making this strict make any difference?  I've been in the habit of using accumulators in OCaml, but I'm never quite sure how they interact with laziness in Haskell.

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Common.hs
```diff
@@ -128,13 +128,13 @@ type PlcError = TypeErrorPlc TPLC.DefaultUni TPLC.DefaultFun ()
 type UplcTerm = UPLC.Term TPLC.Name TPLC.DefaultUni TPLC.DefaultFun ()
 
 -- Possible CEK evluation results, flattened out
-data CekResult =
+data TypeCekResult =
```
kwxm: I'm not convinced that `TypeCekResult` accurately conveys what this is for.  It sounds like it's synthesising a type for the result of a CEK evaluation or something.  The only alternative I can come up with at the moment is `TypeErrorOrCekResult`, which is pretty unwieldy.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: Yes, thank you.

effectfully:
@zliu41 

> Looks good. I don't know why dischargeCekValue shouldn't just preserve the ann in CekValue, but that's orthogonal to this PR.

You evaluate a builtin application, where do you get an annotation for the result? From the `Builtin` node? Well, then you also need to propagate it through the whole builtins machinery or postprocess the result -- slowing everything down just to use the same `()` in the end, because we don't care about annotations on the chain or pretty much anywhere (unless it's the steppable machine, but even there we don't use annotations in that place).

The real question is "why do we have annotations for the on-chain AST at all?". And the real answer is "we shouldn't". 

@kwxm 

> I'm wondering if the apparent slowdown in the validation benchmarks is real.

Given that

1. our benchmarks notoriously capture things they aren't supposed to be capturing

2. 4 runs of the same scripts give 4 different readings:
Screenshot from 2025-08-13 21-19-19
Screenshot from 2025-08-13 21-19-05

3. I'm only changing the returned value and not anything related to how the CEK machine operates

4. I've checked the Core and didn't see anything suspicious about it, it's still proper recursive join points:

image

I think it's fair for me not to lose sleep over a +1-2% change.

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Common.hs
effectfully: > I'm not convinced 

Me neither, but it's a bunch of tests stuff, so I personally don't care.

Anyway, renamed to `TypeErrorOrCekResult`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
effectfully: Nope, reversing a list is one of the few cases where a lazy left fold is appropriate, it's how it's [defined](https://hackage.haskell.org/package/ghc-internal-9.1201.0/docs/src/GHC.Internal.List.html#reverse) in `base` too.

PR #7278, opened 2025-08-14
[Evaluation] Add 'PartialBuiltinApp'

effectfully:
Instead of relying on laziness we can just create a special data type to handle partial builtin applications in the CEK machine. Let's see how this performs.


effectfully:
OK, this is clearly not an improvement.

PR #7281, opened 2025-08-15
Make `FrameAwaitFunValue` handle multiple arguments at once

SeungheonOh:
This is part of effort to add multi-application as per https://github.com/IntersectMBO/plutus/issues/6225.

`FrameAwaitFunValue` is very often nested and used iteratively; this PR makes all the nested structures stored in a single frame.

### DIFF
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
@@ -36,7 +36,7 @@ module UntypedPlutusCore.Evaluation.Machine.Cek.Internal
     , CekValue(..)
     , DischargeResult(..)
     , dischargeResultToTerm
-    , ArgStack(..)
+    , ArgStack
     , transferArgStack
     , CekUserError(..)
     , CekEvaluationException
@@ -111,6 +111,7 @@ import Data.Word
 import GHC.Generics
 import GHC.TypeLits
 import Prettyprinter
+import Queue qualified as Queue
 import Universe
 
 {- Note [Compilation peculiarities]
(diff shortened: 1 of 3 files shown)


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -710,8 +711,7 @@ directly to the head of the application. Which is why 'transferSpine' is a right
 -- See Note [ArgStack vs Spine].
 -- | Transfers an 'ArgStack' to a series of 'Context' frames.
 transferArgStack :: ArgStack uni fun ann -> Context uni fun ann -> Context uni fun ann
-transferArgStack EmptyStack c           = c
-transferArgStack (ConsStack arg rest) c = transferArgStack rest (FrameAwaitFunValue arg c)
+transferArgStack = FrameAwaitFunValues
```
zliu41: remove?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -847,15 +841,26 @@ enterComputeCek = computeCek
...
+        let
+          reverseArgStack = go EmptyStack
+            where
+              go acc EmptyStack       = acc
+              go acc (ConsStack x xs) = go (ConsStack x acc) xs
+          done' = ConsStack e done
         case todo of
           (next : todo') -> computeCek (FrameConstr env i todo' done' ctx) env next
-          []             -> returnCek ctx $ VConstr i done'
+          []             -> returnCek ctx $ VConstr i (reverseArgStack done')
```
effectfully, line 863: I feel like we can make this a tad more efficient by moving things around. I'll play with it once this PR is merged.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -671,6 +669,8 @@ data Context uni fun ann
     -- ^ @[_ N]@
     | FrameAwaitFunValue !(CekValue uni fun ann) !(Context uni fun ann)
     -- ^ @[_ V]@
+    | FrameAwaitFunValues !(ArgStack uni fun ann) !(Context uni fun ann)
+    -- ^ @[[[_ V_1] ...] V_n]@
```
effectfully: Could you please rename the frame to `FrameAwaitFunValueN` or something? The current name suggests that we're expecting multiple function values, but we're expecting a single function value applied to multiple argument values.

Also, could you please use iterated application `[_ V_1 ... V_n]`, that would be more readable. And below too.

I'd also prefer us to rename `V_1`/`V_n` to `V1`/`Vn`, but given that it's existing notation, we can do that separately if we agree that it'd read better given that a standalone `_` has its own different meaning.

effectfully (approved):
Good stuff.

Pure nitpicking:

PR #7287, opened 2025-08-21
[OPTIMIZATION] Use spine directly for constant casing

SeungheonOh:
don't merge yet

### DIFF
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
@@ -667,7 +667,7 @@ data Context uni fun ann
     -- ^ @[V _]@
     | FrameAwaitFunTerm !(CekValEnv uni fun ann) !(NTerm uni fun ann) !(Context uni fun ann)
     -- ^ @[_ N]@
-    | FrameAwaitFunValue !(CekValue uni fun ann) !(Context uni fun ann)
+    | FrameAwaitFunConN !(Spine (Some (ValueOf uni))) !(Context uni fun ann)
     -- ^ @[_ V]@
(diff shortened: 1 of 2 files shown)


SeungheonOh:
I don't believe previous result. Casings that have branches(Int, bool, etc) with no argument shouldn't change at all!

/benchmark casing

effectfully:
I think I have a better idea, hold on.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -877,7 +871,7 @@ enterComputeCek = computeCek
         VCon val -> case unCaserBuiltin ?cekCaserBuiltin val cs of
             Left err          -> throwErrorDischarged (OperationalError $ CekCaseBuiltinError err) e
             Right (HeadOnly fX) -> computeCek ctx env fX
```
SeungheonOh, line 905: this is why the first benchmark result seems nonsensical to me. Any casing with branches that doesn't have argument shouldn't change at all

SeungheonOh: or am I delusional... ?

SeungheonOh:
yeah benchmark doesn't makes sense. This is local run: https://gist.github.com/SeungheonOh/d4476c664315127f9c1fa188714bd071

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
SeungheonOh: Yeah no, benchmark machine is finicky. ignore this.

effectfully (approved):
Great stuff, thank you for doing it.

PR #7289, opened 2025-08-22
Add golden test for `validation` benchmark suite

SeungheonOh:
It's not _that_ useful to have golden cases for validation since all of the scripts in `validation` suite is smart contract execution, which would evaluate into some small value, typically unit or `id`. 

However, this is useful for detecting any irregularities in costing and for detecting cases when CEK changes causes existing code in `validation` to fail entirely.

### DIFF
--- a/plutus-benchmark/plutus-benchmark.cabal
+++ b/plutus-benchmark/plutus-benchmark.cabal
@@ -280,7 +280,7 @@ benchmark casing
 ---------------- validation ----------------
 
 library validation-internal
-  import:          lang, ghc-version-support, os-support
+  import:          lang, os-support
   hs-source-dirs:  validation/src
   exposed-modules: PlutusBenchmark.Validation.Common
   build-depends:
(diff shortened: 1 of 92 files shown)


plutus-benchmark/plutus-benchmark.cabal
```diff
@@ -279,12 +279,10 @@ benchmark casing
...
-benchmark validation
-  import:         lang, os-support
-  type:           exitcode-stdio-1.0
-  main-is:        BenchCek.hs
-  hs-source-dirs: validation/bench
-  other-modules:  Common
+library validation-internal
+  import:          lang, os-support
+  hs-source-dirs:  validation/src
+  exposed-modules: PlutusBenchmark.Validation.Common
```
effectfully, line 285: That's much better indeed.

effectfully (approved):
Thank you for doing it! I can sleep a little better now.

PR #7292, opened 2025-08-26
fix pragmas, indentation and expose `plutus-benchmark-common`

perturbing:
I often like to depend on this lib outside of plutus to experiements and test how costly things are. Like I did [here](https://github.com/perturbing/plutus-MiMC/blob/a29963d0e8092d4f3c04bf80e805e811201ad17c/plutus-mimc/bench/Bls/Run.hs#L18) (note that this depends on a custom fork already exposing this lib).

Can we permanently expose this lib in plutus?

### DIFF
--- a/plutus-benchmark/common/PlutusBenchmark/Common.hs
+++ b/plutus-benchmark/common/PlutusBenchmark/Common.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE BangPatterns        #-}
-{-# LANGUAGE ImportQualifiedPost #-}
 {-# LANGUAGE LambdaCase          #-}
 
 {- | Miscellaneous shared code for benchmarking-related things. -}
--- a/plutus-benchmark/common/PlutusBenchmark/Common.hs
+++ b/plutus-benchmark/common/PlutusBenchmark/Common.hs
@@ -1,5 +1,6 @@
-{-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE LambdaCase   #-}
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE LambdaCase          #-}
 
 {- | Miscellaneous shared code for benchmarking-related things. -}
 module PlutusBenchmark.Common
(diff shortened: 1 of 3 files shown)


plutus-benchmark/common/PlutusBenchmark/NaturalSort.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE DerivingStrategies #-}
```
effectfully: It's enabled globally in `plutus-benchmark.cabal`, so it shouldn't be needed here.

effectfully (approved):
I have no issue with `plutus-benchmark` being exposed publicly as long as folks understand that we may break backwards compatibility at will.

kwxm:
There's a question about whether this would mean that we'd have to include `plutus-benchmark` in the Plutus release process and release it to CHaP. An alternative would be to move the relevant code to `plutus-ledger-api` (which it depends on anyway, although maybe that's not really the right place for it).  

@zliu41: what do you think?

zliu41:
> Can we permanently expose this lib in plutus?

If adding `visibility:      public` is all you need, that's totally fine.

plutus-benchmark/common/PlutusBenchmark/Common.hs
```diff
@@ -1,5 +1,6 @@
-{-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE LambdaCase   #-}
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE ImportQualifiedPost #-}
```
kwxm:
```suggestion
```

plutus-benchmark/common/PlutusBenchmark/NaturalSort.hs
kwxm:
```suggestion
```

kwxm: @perturbing Did you have some reason for adding that, like you needed it to use a copy of the code outside our repository?  I've removed this and another one that don't seem to be needed.

PR #7298, opened 2025-08-29
Build with -Werror by default

ana-pantilie:
Saves my sanity.


SeungheonOh (approved):
I think it would save mine too. Thanks

PR #7299, opened 2025-09-01
Add SrcSpans of Vars in the compilation trace

zliu41:
The effect can be seen both in the full compilation trace (via `dump-compilation-trace`), or in the error message when the compilation fails.

Unfortunately most `Var`s don't have `SrcSpan`, but this is better than nothing.

### DIFF
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
@@ -637,24 +637,30 @@ hoistExpr var t = do
   -- See Note [Dependency tracking]
   modifyCurDeps (Set.insert lexName)
   maybeDef <- PIR.lookupTerm lexName
-  let addSpan = case getVarSourceSpan var of
+  let varSpan = getVarSourceSpan var
+      addSpan = case varSpan of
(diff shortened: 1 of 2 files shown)


plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -637,24 +637,30 @@ hoistExpr var t = do
...
-        (PIR.Def var' (PIR.mkVar var', PIR.Strict))
-        mempty
-
-      t' <- maybeProfileRhs var var' =<< addSpan (compileExpr t)
-      -- See Note [Non-strict let-bindings]
-      PIR.modifyTermDef lexName (const $ PIR.Def var' (t', PIR.NonStrict))
-      pure $ PIR.mkVar var'
+    Nothing -> withCurDef lexName
+      . traceCompilation 1 ("Compiling definition of:" GHC.<+> GHC.ppr var GHC.<> varSpanMsg)
+      $ do
```
zliu41, line 652: No change in the `do` block, only indentation changed.

SeungheonOh (approved):
looks good.

edit: need to sign commits

PR #7305, opened 2025-09-04
Remove unused test inputs

kwxm:
In #7223 I refactored the `builtinsAvailableIn` code quite substantially and added a check to make sure that it did exactly the same as the previous version (see [this comment](https://github.com/IntersectMBO/plutus/pull/7223#discussion_r2218084218)).  I've just noticed that this contained some data that was never used so I've removed it here.  I've also updated some comments.


plutus-ledger-api/test/Spec/Data/Versions.hs
kwxm, line 1: This test is only supposed to make sure that the two versions of the test agree up to Plomin.  It also included some new builtins in an entry that the [comprehension](https://github.com/IntersectMBO/plutus/blob/a2fb5260fe73f4102f88cba6c638663a4e652ba0/plutus-ledger-api/test/Spec/Data/Versions.hs#L326) a bit later didn't (and shouldn't) cover, so I've removed those.

zliu41 (approved):

PR #7308, opened 2025-09-04
Vendor the `flat` package

zeme-wana:
Vendor the [flat](https://hackage.haskell.org/package/flat) library in plutus:
- Original sources kept as-is except for: 
- Removed `benchmarks` folder
- Removed `DocTest` test-suite
- Changed `cabal-version: 3.0` in `flat.cabal`
- Added `default-extensions: ImportQualifiedPost` to `flat.cabal` and applied `stylish-haskell`
- Replaced instances of `import Data.ByteString.Internal.memcpy` with `import Foreign.Marshal.Utils (copyBytes)` as per deprecation errors.
- Original `PlutusCore.Flat` module in plutus-core has been renamed to `PlutusCore.FlatInstances`


cabal.project
```diff
@@ -25,6 +25,7 @@ packages: cardano-constitution
           plutus-benchmark
           plutus-conformance
           plutus-core
+          plutus-core/flat
```
kwxm: Does this mean that references to `flat` in the .cabal files will get our version now?  I'm not sure how things get resolved if there are multiple libraries with the same name floating about.

plutus-core/flat/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright (c) 2016, Pasqualino `Titto` Assini
```
kwxm, line 1: I wonder if we should remove this license if we're going to start modifying the code.  Maybe there's some company policy about this kind of situation? I see that the license does say that we're supposed to retain it, but what happens if we change the source code?

plutus-core/flat/flat.cabal
```diff
@@ -0,0 +1,171 @@
+cabal-version:      3.0
```
kwxm: The flat code is independent of the rest of the Plutus code at the moment.  I wonder if we should put it in a separate package, like `plutus-flat` or something.  I don't know if that would be sensible or not, but it seems like the sort of thing that people mighr have an opinion about.

kwxm (commented):
This all looks OK, and it appears to work as it should.  I'm not familiar with licensing issues and the details of how one deals with incorporating someone else's code into the codebase though, so we should probabaly get someone else to review it too.

cabal.project
zeme-wana: Yes. Cabal first resolves `flat` via `cabal.project`'s `packages` and S.R.P stanzas and if that fails it looks into `hackage` and `CHaP`

plutus-core/flat/LICENSE
zeme-wana: The license says that ridistribution is permitted so long as the original LICENSE file is retained, AFAIU. 
But we should get a second opinion.

plutus-core/flat/flat.cabal
zeme-wana: A top-level `plutus-flat` folder is just as good, probably even a better idea.
We should vote: who's in favour of a top-level `plutus-flat` as opposed to `plutus-core/flat`? React with 👍 @IntersectMBO/plutus-core-maintainers

Unisay: Without having a strong preference I'd chose plutus-core/flat since we already keep other libs like satint there and plutus-core package serves as a namespace. But top level isn't bad too.

ana-pantilie: Why are the other libs like `satint` in `plutus-core` instead of separate packages?

zliu41: Top-level with its own `.cabal` file is better - so that it is just a mirror of the original project's repository. Easier to upstream changes this way.

> Why are the other libs like satint in plutus-core instead of separate packages?

Is `satint` a vendored library? Seems not?

kwxm: I think I lean slightly towards putting it at the top level because it'd be a bit easier to work on, but it's not a big deal.

plutus-core/plutus-core.cabal
```diff
@@ -318,7 +318,7 @@ library
     , exceptions
     , extra
     , filepath
-    , flat                        ^>=0.6
+    , flat
```
zliu41: plutus-core:flat?

plutus-core/flat/flat.cabal
ana-pantilie: > Top-level with its own .cabal file is better - so that it is just a mirror of the original project's repository. Easier to upstream changes this way.

I agree with this.

PR #7309, opened 2025-09-04
Change default option for datatype compilation

basetunnel:
This changes the default datatype compilation option from `BuiltinCase` to `SumsOfProducts`. Tests that require `SumsOfProducts` have been adapted and docs have been regenerated.

I've checked the tests in `plutus-core` and `plutus-ledger-api`. Are there other tests I should look at?

### DIFF
--- a/plutus-core/changelog.d/20250911_104659_4022046+basetunnel_default_datatype_opt.md
+++ b/plutus-core/changelog.d/20250911_104659_4022046+basetunnel_default_datatype_opt.md
@@ -1,0 +1,3 @@
+### Changed
+
+- Change default compiler option `datatypes` from `BuiltinCasing` back to `SumsOfProducts`


doc/docusaurus/docs/delve-deeper/plinth-compiler-options.md
```diff
@@ -20,11 +20,13 @@ For each boolean option, you can add a `no-` prefix to switch it off, such as `n
 
 |Option|Value Type|Default|Description|
 |-|-|-|-|
+|`certify`|Maybe [Char]||Produce a certificate for the compiled program, with the given name. This certificate provides evidence that the compiler optimizations have preserved the functional behavior of the original program. Currently, this is only supported for the UPLC compilation pipeline. Warning: this is an experimental feature and may not work as expected.|
```
basetunnel, line 23: Not sure if this should be in there.

basetunnel:
I missed some tests, looking into it.

doc/docusaurus/docs/delve-deeper/plinth-compiler-options.md
zliu41: Yes - CI doesn't check that this page is up to date, so someone might have forgotten to update it.

zliu41 (approved):
Looks good, just need to fix the remaining tests.

basetunnel:
Looks like all tests are going through. I've updated the golden files that have annotations with line numbers, because those changed.

PR #7312, opened 2025-09-05
Kwxm/doc/release builtins

kwxm:
This describes how to update `plutus-ledger-api` when new PLC builtins are added.  The process is quite complicated and I ended up with rather a long document.  I'm not sure where the best place to put it is: it's not very discoverable where it is now.  Maybe we should make an issue template that points to it and has checkboxes for the important steps.

### DIFF
--- a/doc/notes/release-builtins/release-builtins.md
+++ b/doc/notes/release-builtins/release-builtins.md
@@ -1,0 +1,219 @@
+# Releasing new built-in functions in Plutus Core
+
+The process of releasing new builtins (or other Plutus Core features such as new
+AST nodes, like `constr` and `case`) is quite complex and has to take place in
+several stages.
+
+1. We add new builtins to the Plutus codebase and test and cost them, and
+eventually merge the code with the main branch. At this point the new builtins
+are available for use locally via eg `uplc` or Plinth.
+
+2. We also have to expose them to the ledger by adding code in
+`plutus-ledger-api`.  This process will be described in greater detail later.
+
+3. Eventually a Plutus release including the new builtins will be integrated
+   with the Cardano node, and at some point a new version of the node software will
+   be released for public use and will be adopted by node operators.  It may take
+   some time until a majority of the nodes in the network are using the new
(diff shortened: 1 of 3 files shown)


zliu41 (approved):
I would prefer this to be written as a checklist, consisting of steps to do when adding a new builtin, as in
1. Add the builtin type (if any) to `DefaultUni` and define necessary instances 
2. Add the builtin function to `DefaultFun` along with the implementation
3. Add the builtin function to `futurePV`
4. Cost it and add parameters to `ParamName`
...the rest...

But this looks OK.

Btw I remember you wrote another doc before about adding new builtins, but I can't find it. It would be good to put that and this one in the same directory.

kwxm:
> I would prefer this to be written as a checklist, consisting of steps to do when adding a new builtin

That's what I was planning to do, but it turned out to be too complicated: you have to initially put things into `futurePV`, then when you're getting close to releasing things you have to create some new PV for that and move the builtins you're really going to release into that, and sometimes create a new PV and so on. We've already got 8 or 9 issue templates for adding a new builtin, so maybe I can add a pointer to this to one of those and maybe add some checkboxes.  Writing the document made me pay attention to exactly what you have to do. so that should make it easier to make a list.

kwxm:
> I would prefer this to be written as a checklist, consisting of steps to do when adding a new builtin, as in
> 
>     1. Add the builtin type (if any) to `DefaultUni` and define necessary instances
> 
>     2. Add the builtin function to `DefaultFun` along with the implementation
>   ...

A lot of that's in the document you mentioned, which is hidden away [here](https://github.com/IntersectMBO/plutus/blob/master/plutus-core/cost-model/CostModelGeneration.md).  That may be a bit out of date now though.  I think you have to add things to the metatheory at a much earlier stage now because of some new dependencies, for example.

kwxm:
I'll merge this for now, but I'll try to collect it together with other relevant documents in a later PR and also take a look at the issue templates once that's done.

Issue #5777, opened 2024-02-15
A builtin converting the outer `Data.Constr` to a SOP for fast `case`?


effectfully:
## Intro

There exist multiple ways to create data types in Plutus Core:

1. Scott-encoding, long retired by now
2. sums-of-products (SOPs)
3. via the `Data` built-in type

SOPs have by far more [efficient](https://github.com/michaelpj/CIPs/blob/192e14988bdccb600450e6db09977d9ed801ebd2/CIP-0085/README.md#datatypes-using-data) pattern matching than `Data` (even with the latter properly [sped-up](https://github.com/IntersectMBO/plutus/issues/5711)). However at the script interface level we have `Data` and hence to get that fast pattern matching on SOPs we have to convert `Data` to a SOP first and this is such an expensive operation that in a lot of cases it completely defeats the purpose as the cost of conversion may significantly outweigh the benefit of having faster pattern matching.

A prime example of `Data` being generally preferable over SOPs is `ScriptContext`, since values of this type are huge and a script doesn't need to access all of them. This problem is discussed in detail in [this](https://github.com/michaelpj/CIPs/blob/mpj/script-context-sop/CIP-%3F%3F%3F%3F/README.md) CIP, which proposes to make `ScriptContext` into a SOP instead of the current `Data`. This is however problematic in a number of ways:

1. it's a breaking change to the API
2. alternative toolchains may not (and some likely do not) deal with SOPs in any way at all as they may find only using `Data` sufficient, particularly since using `Data` is simply faster in many cases

What if instead of trying to force SOPs upon everybody whereever possible we simply make pattern matching on `Data`-encoded data types more efficient, making it competitive with pattern matching on SOPs?

## High-level idea

There's one deceptively simple way of making pattern matching on `Data`-encoded data types faster: just convert a `Data.Constr` node into the respective `Term.Constr` node and case on the latter instead using the fast `case` primitive.

This is how `Constr` is defined for `Data`:

```haskell
data Data
    = Constr Integer [Data]
    | <...>
```

This is how `Constr` is defined for UPLC `Term`:

```haskell
data Term name uni fun ann
    = Constr !ann !Word64 ![Term name uni fun ann]
    | <...>
```

Any `Data.Constr` is embeddable into `Term` via

[...]

PR #7315, opened 2025-09-06
[Builtin] Casing on `Data.Constr`

SeungheonOh:
Implements 
```
case (con Data (Constr n [x, y, z])) 
  <branch_0 :: Data -> Data -> Data -> r>
  <branch_1 :: Data -> r>
  <branch_2 :: r>
```

where each branches get each elements of the "field" part of `Data.Constr` as arguments. This allows deconstruction of `Data.Constr` value **in single CEK step**.


plutus-core/plutus-core/src/PlutusCore/TypeCheck/Internal.hs
```diff
@@ -577,10 +578,28 @@ inferTypeM (Case ann resTy scrut branches) = do
             -- for the number of branches
             Nothing -> throwError (TypeMismatch ann (void scrut) expectedSop vScrutTy)
         vTy -> case annotateCaseBuiltin vTy branches of
-            Right branchesAndArgTypes -> for_ branchesAndArgTypes $ \(c, argTypes) -> do
+            Right branchesAndArgTypes -> for_ branchesAndArgTypes $ \case
+              FixedArityBranch c argTypes -> do
                 vArgTypes <- traverse (fmap unNormalized . normalizeTypeM) argTypes
                 -- made of sub-parts of a normalized type, so normalized
                 checkTypeM ann c (Normalized $ mkIterTyFun () vArgTypes (unNormalized vResTy))
+              VariableArityBranch c argType -> do
```
SeungheonOh, line 586: Not really happy here, but don't think there's any better way to go about without bigger changes. The fact that each branch can have arbitrary number of arguments introduces this dynamic behavior. So, it is only possible to check the return type and to check if are inputs(of arbitrary number) have correct type.

plutus-tx/src/PlutusTx/Builtins/Internal.hs
```diff
@@ -1068,3 +1071,18 @@ caseInteger i b = b !! fromIntegral i
...
+  let
+    constr = unsafeDataAsConstr d
+    idx = PlutusTx.Builtins.Internal.fst constr
+    ds = PlutusTx.Builtins.Internal.snd constr
+    unwrap (CaseDataConstrBranch x) = unsafeCoerce x
+    applyToBranch :: r -> BuiltinList BuiltinData -> r
+    applyToBranch body =
+      caseList' body (\arg -> applyToBranch (unsafeCoerce body $ arg))
+  in applyToBranch (unwrap $ branches !! fromIntegral idx) ds
+{-# OPAQUE caseDataConstr #-}
```
SeungheonOh: This is horrendous; needed that existential type since number of argument each branch handler takes is arbitrary. This works, but it's impossible to make it also work for when we disable builtin-casing compilation option. 

I'm going to try other ways. Namely, using `\[a, b, c] -> ...`

plutus-benchmark/script-contexts/test/V2/9.6/checkScriptContext2-20.eval.golden
```diff
@@ -1,6 +1,6 @@
-CPU:               117_015_830
-Memory:                574_864
-Term Size:                 954
-Flat Size:               1_499
+CPU:                46_154_228
```
SeungheonOh, line 1: Nice

SeungheonOh (commented):
CEK part is more or less set, but plinth part needs some rework. Feel free to just review plutus-core part only.

plutus-tx/src/PlutusTx/Builtins/Internal.hs
SeungheonOh: Made some changes. It works good now without this atrocity

effectfully (commented):
I don't like this design. See the "What exactly breaks if we add the unsafe constrTermFromConstrData?" section of #5777 for why.

If the team thinks this is what we should do, I won't fight it. But I believe it is not what we should do.

PR #7319, opened 2025-09-08
Builtin Value: efficient retrieval of total size and max inner map size

zliu41:
This PR changes the underlying data type of builtin `Value`, to keep track of the total size, and the size of the max inner map.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
@@ -13,7 +13,7 @@ module PlutusCore.Evaluation.Machine.ExMemoryUsage
     , NumBytesCostedAsNumWords(..)
     , IntegerCostedLiterally(..)
     , ValueTotalSize(..)
-    , ValueOuterPlusMaxInner(..)
+    , ValueOuterOrMaxInner(..)
     ) where
 
 import PlutusCore.Crypto.BLS12_381.G1 as BLS12_381.G1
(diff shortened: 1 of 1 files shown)


plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -14,33 +20,77 @@ import Data.Bifunctor
...
+
+{-| Pack a map from (currency symbol, token name) to amount into a `Value`.
+
+The map will be filtered so that it does not contain empty inner map or zero amount.
+-}
+pack :: NestedMap -> Value
+pack (normalize -> v) = Value v sizes size
+ where
+  sizes = Map.foldr' (IntMap.alter (maybe (Just 1) (Just . (+ 1))) . Map.size) mempty v
+  size = Map.foldr' ((+) . Map.size) 0 v
```
effectfully, line 68: Out of curiosity, is there any specific reason why you're using `foldr'` and not `foldl'`? It probably doesn't matter either way, but if it does, then I wanna know.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
```diff
@@ -244,7 +244,9 @@ instance Arbitrary Data where
     arbitrary = arbitraryBuiltin
     shrink = shrinkBuiltin
 
-deriving via Map ByteString (Map ByteString Integer) instance Arbitrary Value
+instance Arbitrary Value where
+    arbitrary = Value.pack <$> arbitrary
```
effectfully, line 248: This is a bad generator. In particular, you're very unlikely to generate two different `Value`s that share the same non-empty token name or currency symbol. See [this](https://github.com/IntersectMBO/plutus/blob/8d810d47112a6ca6a995feb968ffab99c622f217/plutus-ledger-api/testlib/PlutusLedgerApi/Test/V1/Data/Value.hs) for how we do it in `plutus-ledger-api`.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
```diff
@@ -244,7 +244,9 @@ instance Arbitrary Data where
...
+    shrink = fmap Value.pack . shrink . Value.unpack
```
effectfully, line 249: This isn't a good shrinker, I have a blog post about [that](https://github.com/effectfully-ou/sketches/tree/master/better-counterexample-minimization). Not too important though.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
```diff
@@ -422,6 +424,7 @@ instance KnownKind k => Arbitrary (MaybeSomeTypeOf k) where
                , JustSomeType DefaultUniBLS12_381_G1_Element
                , JustSomeType DefaultUniBLS12_381_G2_Element
                , JustSomeType DefaultUniBLS12_381_MlResult
+               , JustSomeType DefaultUniValue
```
effectfully, line 427: Eek, too bad this has to be remembered instead of giving us a warning. On the one hand it'd be great to fix that, on the other hand this whole stuff needs to be rewritten anyway if we ever split universes.

effectfully (approved):

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: The only reason is that foldr's step function type, `a -> b -> b`, is often more convenient than `b -> a -> b`. I don't think there's a performance difference in this particular case.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
zliu41: Agreed - I'm going to create a different (hedgehog) generator in the next PR.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
zliu41: Exactly. The only reason I updated here is because @kwxm found a bug in the previous PR, and I wondered why CI was green.

PR #7322, opened 2025-09-09
Add insertCoin and unionValue implementations and tests

zliu41:
As titled.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Value.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Value.hs
@@ -12,6 +12,8 @@ module PlutusCore.Value (
   toList,
   totalSize,
   maxInnerSize,
+  insertCoin,
+  unionValue,
 ) where
 
 import Codec.Serialise (Serialise)
@@ -91,8 +93,42 @@ fromList =
 
 normalize :: NestedMap -> NestedMap
 normalize = Map.filter (not . Map.null) . Map.map (Map.filter (/= 0))
+{-# INLINEABLE normalize #-}
 
 instance Pretty Value where
   pretty = pretty . fmap (bimap toText (fmap (first toText))) . toList
    where
     toText = Text.decodeLatin1 . Base64.encode
+
+{-| \(O(\log \max(m, k))\), where \(m\) is the size of the outer map, and \(k\) is
+the size of the largest inner map.
+-}
+insertCoin :: ByteString -> ByteString -> Integer -> Value -> Value
+insertCoin currency token amt (Value outer sizes size) = case Map.lookup currency outer of
+  Nothing ->
+    let inner = Map.singleton token amt
+        outer' = Map.insert currency inner outer
+        sizes' = updateSizes 0 sizes
+        size' = size + 1
+     in Value outer' sizes' size'
+  Just inner ->
+    let exists = Map.member token inner
+        inner' = Map.insert token amt inner
+        outer' = Map.insert currency inner' outer
+        sizes' = if exists then sizes else updateSizes (Map.size inner) sizes
+        size' = if exists then size else size + 1
+     in Value outer' sizes' size'
+{-# INLINEABLE insertCoin #-}
+
+-- | \(O(n)\), where \(n\) is the total size of the maps.
+unionValue :: Value -> Value -> Value
+unionValue v v' = pack (Map.unionWith (Map.unionWith (+)) (unpack v) (unpack v'))
+{-# INLINEABLE unionValue #-}
+
+-- | Decrement bucket @oldSize@, and increment bucket @oldSize + 1@.
+updateSizes :: Int -> IntMap Int -> IntMap Int
+updateSizes oldSize = dec . inc
+ where
+  inc = IntMap.alter (maybe (Just 1) (Just . (+ 1))) (oldSize + 1)
+  dec = IntMap.update (\n -> if n <= 1 then Nothing else Just (n - 1)) oldSize
+{-# INLINEABLE updateSizes #-}
--- a/plutus-core/plutus-core/test/Value/Spec.hs
+++ b/plutus-core/plutus-core/test/Value/Spec.hs
@@ -1,0 +1,150 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications  #-}
+{-# LANGUAGE ViewPatterns      #-}
+
+module Value.Spec (tests) where
+
+import Data.ByteString (ByteString)
+import Data.ByteString.Char8 qualified as BC
+import Data.Foldable
+import Data.Int
+import Data.Map.Strict qualified as Map
+import Data.Maybe
+import Hedgehog (Gen, MonadTest, Property, assert, forAll, property, (===))
+import Hedgehog.Gen qualified as Gen
+import Hedgehog.Range qualified as Range
+import PlutusCore.Value (Value)
+import PlutusCore.Value qualified as V
+import Safe.Foldable (maximumMay)
+import Test.Tasty
+import Test.Tasty.Hedgehog
+
+genCurrency :: Gen ByteString
+genCurrency = Gen.element $ fmap (\(n :: Int) -> "currency" <> BC.pack (show n)) [1 .. 100]
+
+genToken :: Gen ByteString
+genToken = Gen.element $ fmap (\(n :: Int) -> "token" <> BC.pack (show n)) [1 .. 100]
+
+genAmount :: Gen Integer
+genAmount = toInteger @Int16 <$> Gen.enumBounded
+
+genValue :: Gen Value
+genValue = do
+  n <- Gen.int (Range.constant 0 100)
+  let f :: Int -> Value -> Gen Value
+      f _ v = V.insertCoin <$> genCurrency <*> genToken <*> genAmount <*> pure v
+  foldrM f V.empty [1 .. n]
+
+genNestedMap :: Gen V.NestedMap
+genNestedMap = do
+  n <- Gen.int (Range.constant 0 100)
+  let f :: Int -> V.NestedMap -> Gen V.NestedMap
+      f _ nm = do
+        currency <- genCurrency
+        token <- genToken
+        amt <- genAmount
+        pure $ Map.insertWith (Map.unionWith (+)) currency (Map.singleton token amt) nm
+  foldrM f mempty [1 .. n]
+
+prop_packUnpackRoundtrip :: Property
+prop_packUnpackRoundtrip = property $ do
+  v <- forAll genValue
+  v === V.pack (V.unpack v)
+
+-- | Verifies that @pack@ correctly updates the sizes
+prop_packBookkeeping :: Property
+prop_packBookkeeping = property $ do
+  nm <- forAll genNestedMap
+  checkSizes (V.pack nm)
+
+{-| Verifies that @pack@ preserves @Value@ invariants, i.e.,
+no empty inner map or zero amount.
+-}
+prop_packPreservesInvariants :: Property
+prop_packPreservesInvariants = property $ do
+  nm <- forAll genNestedMap
+  checkInvariants (V.pack nm)
+
+-- | Verifies that @insertCoin@ correctly updates the sizes
+prop_insertCoinBookkeeping :: Property
+prop_insertCoinBookkeeping = property $ do
+  v <- forAll genValue
+  currency <- forAll genCurrency
+  token <- forAll genToken
+  amt <- forAll genAmount
+  let v' = V.insertCoin currency token amt v
+  checkSizes v'
+
+-- | Verifies that @insertCoin@ preserves @Value@ invariants
+prop_insertCoinPreservesInvariants :: Property
+prop_insertCoinPreservesInvariants = property $ do
+  v <- forAll genValue
+  currency <- forAll genCurrency
+  token <- forAll genToken
+  amt <- forAll genAmount
+  let v' = V.insertCoin currency token amt v
+  checkInvariants v'
+
+prop_unionCommutative :: Property
+prop_unionCommutative = property $ do
+  v <- forAll genValue
+  v' <- forAll genValue
+  V.unionValue v v' === V.unionValue v' v
+
+prop_insertCoinIdempotent :: Property
+prop_insertCoinIdempotent = property $ do
+  v <- forAll genValue
+  let fm = toFlatMap v
+  v === foldl' (\acc (c, t, a) -> V.insertCoin c t a acc) v fm
+
+checkSizes :: (MonadTest m) => Value -> m ()
+checkSizes v =
+  (expectedMaxInnerSize === actualMaxInnerSize)
+    >> (expectedSize === actualSize)
+ where
+  expectedMaxInnerSize = fromMaybe 0 . maximumMay $ Map.map Map.size (V.unpack v)
+  actualMaxInnerSize = V.maxInnerSize v
+  expectedSize = sum $ Map.map Map.size (V.unpack v)
+  actualSize = V.totalSize v
+
+checkInvariants :: (MonadTest m) => Value -> m ()
+checkInvariants (V.unpack -> v) = do
+  assert $ (not . any Map.null) v
+  assert $ (not . any (elem 0)) v
+
+toFlatMap :: Value -> [(ByteString, ByteString, Integer)]
+toFlatMap (V.toList -> xs) = [(c, t, a) | (c, ys) <- xs, (t, a) <- ys]
+
+tests :: TestTree
+tests =
+  testGroup
+    "Value"
+    [ testPropertyNamed
+        "packUnpackRoundtrip"
+        "prop_packUnpackRoundtrip"
+        prop_packUnpackRoundtrip
+    , testPropertyNamed
+        "packBookkeeping"
+        "prop_packBookkeeping"
+        prop_packBookkeeping
+    , testPropertyNamed
+        "packPreservesInvariants"
+        "prop_packPreservesInvariants"
+        prop_packPreservesInvariants
+    , testPropertyNamed
+        "insertCoinBookkeeping"
+        "prop_insertCoinBookkeeping"
+        prop_insertCoinBookkeeping
+    , testPropertyNamed
+        "insertCoinPreservesInvariants"
+        "prop_insertCoinPreservesInvariants"
+        prop_insertCoinPreservesInvariants
+    , testPropertyNamed
+        "unionCommutative"
+        "prop_unionCommutative"
+        prop_unionCommutative
+    , testPropertyNamed
+        "insertCoinIdempotent"
+        "prop_insertCoinIdempotent"
+        prop_insertCoinIdempotent
+    ]
--- a/plutus-core/changelog.d/20250909_084412_unsafeFixIO_insertCoin_unionValue.md
+++ b/plutus-core/changelog.d/20250909_084412_unsafeFixIO_insertCoin_unionValue.md
@@ -1,0 +1,4 @@
+
+### Added
+
+- Built-in type `Value`, and the implementations of `insertCoin` and `unionValue`.
--- a/plutus-core/plutus-core.cabal
+++ b/plutus-core/plutus-core.cabal
@@ -378,6 +378,7 @@ test-suite plutus-core-test
     Parser.Spec
     Pretty.Readable
     TypeSynthesis.Spec
+    Value.Spec
 
   default-language: Haskell2010
   build-depends:
@@ -396,6 +397,7 @@ test-suite plutus-core-test
     , plutus-core                      ^>=1.53
     , plutus-core:plutus-core-testlib
     , prettyprinter
+    , safe
     , serialise
     , tasty
     , tasty-golden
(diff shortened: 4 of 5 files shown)


plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -0,0 +1,150 @@
...
+genCurrency :: Gen ByteString
+genCurrency = Gen.element $ fmap (\(n :: Int) -> "currency" <> BC.pack (show n)) [1 .. 100]
+
+genToken :: Gen ByteString
+genToken = Gen.element $ fmap (\(n :: Int) -> "token" <> BC.pack (show n)) [1 .. 100]
+
+genAmount :: Gen Integer
+genAmount = toInteger @Int16 <$> Gen.enumBounded
+
+genValue :: Gen Value
```
zliu41: @effectfully what do you think of this generator?

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -91,8 +104,48 @@ fromList =
...
+  Nothing ->
+    let inner = Map.singleton token amt
+        outer' = Map.insert currency inner outer
+        sizes' = updateSizes 0 1 sizes
+        size' = size + 1
+     in Value outer' sizes' size'
+  Just inner ->
+    let exists = Map.member token inner
+        inner' = Map.insert token amt inner
+        outer' = Map.insert currency inner' outer
```
effectfully: This is so inefficient with all the lookup-member-inserts. Instead of using `lookup` + `insert` or `member` + `insert` you can probably do the same thing with just `alterF`.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -91,8 +104,48 @@ fromList =
...
+        sizes' = if exists then sizes else updateSizes (Map.size inner) (Map.size inner') sizes
+        size' = if exists then size else size + 1
+     in Value outer' sizes' size'
+{-# INLINEABLE insertCoin #-}
+
+-- | \(O(n)\), where \(n\) is the total size of the maps.
```
effectfully: That's not the complexity of `union`?

plutus-core/plutus-core/test/Value/Spec.hs
effectfully: Why use `hedgehog` over `QuickCheck` at all?

Why not reuse this logic:

```haskell
-- | Return how many candidates to randomly choose from to fill the given number of cells. For
-- example, if we only need to fill a single cell, we choose from 6 different candidates, and if we
-- need to fill 5 cells, we choose from 11 candidates.
--
-- >>> map (\i -> (i, toCellCandidatesNumber i)) [1..13]
-- [(1,6),(2,6),(3,6),(4,8),(5,11),(6,14),(7,18),(8,22),(9,27),(10,31),(11,36),(12,41),(13,46)]
toCellCandidatesNumber :: Int -> Int
toCellCandidatesNumber i = max 6 . floor @Double $ fromIntegral i ** 1.5

-- | Generate a 'BuiltinByteString' by picking one of the predetermined ones, given a number of
-- cells to fill (see 'toCellCandidatesNumber'). The idea is that we want to occasionally generate
-- the same 'CurrencySymbol' or 'TokenName' for different 'Value's to have decent test coverage,
-- hence to make name clashing more likely we pick from a predetermined set of
-- 'BuiltinByteString's. Otherwise the chance of generating the same 'BuiltinByteString' for two
-- different 'Value's would be virtually zero.
genShortHex :: Int -> Gen BuiltinByteString
genShortHex i =
    toBuiltin . Base16.encode . BS8.pack . show <$> elements [0 .. toCellCandidatesNumber i]
```

? You can even just copy-paste it here or pull it into some common test library. Although it's for `QuickCheck`. But you can use `QuickCheck` within `hedgehog` and vice versa.

We need to hardcode ADA, because it's hardcoded in the ledger, otherwise it's an important edge case that we're never going to hit.

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -0,0 +1,158 @@
...
+import Test.Tasty.Hedgehog
+
+genCurrency :: Gen ByteString
+genCurrency = Gen.element $ fmap (\(n :: Int) -> "currency" <> BC.pack (show n)) [1 .. 100]
+
+genToken :: Gen ByteString
+genToken = Gen.element $ fmap (\(n :: Int) -> "token" <> BC.pack (show n)) [1 .. 100]
+
+genAmount :: Gen Integer
+genAmount = toInteger @Int16 <$> Gen.enumBounded
```
effectfully: You're very unlikely to generate small ints this way as far as I can tell. And you're never gonna generate anything above 2^15, which seems overly restrictive too. 

You're also not gonna hit interesting numbers like 2^15-1, 2^10 etc.

We have all these issues sorted out for QuickCheck [generators](https://github.com/IntersectMBO/plutus/blob/13ff9940d4be9707b63af42bbd45fe5788deaf2e/plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs).

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -91,8 +104,48 @@ fromList =
...
 instance Pretty Value where
   pretty = pretty . fmap (bimap toText (fmap (first toText))) . toList
    where
     toText = Text.decodeLatin1 . Base64.encode
+
+{-| \(O(\log \max(m, k))\), where \(m\) is the size of the outer map, and \(k\) is
+the size of the largest inner map.
+-}
+insertCoin :: ByteString -> ByteString -> Integer -> Value -> Value
+insertCoin currency token amt (Value outer sizes size) = case Map.lookup currency outer of
```
effectfully: What if `amt` is `0`?

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -0,0 +1,158 @@
...
+
+genValue :: Gen Value
+genValue = do
+  n <- Gen.int (Range.constant 0 100)
+  let f :: Int -> Value -> Gen Value
+      f _ v = V.insertCoin <$> genCurrency <*> genToken <*> genAmount <*> pure v
```
effectfully: In particular, you want `genAmount` to generate `0` more often than with 1/2^16 chance.

plutus-core/plutus-core/test/Value/Spec.hs
effectfully: I had this written in the comment:

> Why limit integer to be below `100`? Why only positive ones?

I realized I looked at it wrong, but forgot to remove it from the comment before submitting. Removed now.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: Right, good catch!

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: > This is so inefficient with all the lookup-member-inserts.

Yeah, by a constant factor, due to the need to maintain sizes. 

> you can probably do the same thing with just alterF.

What if the `f` you have in mind?

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: It is the complexity of `pack`. To avoid `pack` would probably be prohibitively complex and I don't think it's worth going there.

plutus-core/plutus-core/test/Value/Spec.hs
zliu41: > Why use hedgehog over QuickCheck at all?

I have no preference, but it seems most other tests in `plutus-core-test` use hedgehog. 

> Why not reuse this logic:

I had no idea this exists (we need an AI bot to help us find such things). Will look into it

SeungheonOh: I prefer Quickcheck. Hedgehog have very slow generator for `Int`, meaning its generators are generally much slower. Slower generator means we can't run many cases, which defeats the purpose of property testing.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -66,22 +70,31 @@ pack (normalize -> v) = Value v sizes size
  where
   sizes = Map.foldr' (IntMap.alter (maybe (Just 1) (Just . (+ 1))) . Map.size) mempty v
   size = Map.foldr' ((+) . Map.size) 0 v
```
SeungheonOh:
```suggestion
  (sizes, size) = 
    Map.foldr' 
      (\e (ss, s) -> 
        ( IntMap.alter (maybe (Just 1) (Just . (+ 1))) (Map.size e) ss
        , s + (Map.size e)
        )
      ) 
      (mempty, 0) 
      v
```

does GHC know how to combine `Map.foldr'` together?

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: Wait, yeah this is totally wrong 🤦

plutus-core/plutus-core/test/Value/Spec.hs
effectfully: > I have no preference, but it seems most other tests in `plutus-core-test` use hedgehog.

Not for any good reason. I've [posted](https://input-output-rnd.slack.com/archives/G01EE4NQ9U0/p1714440863325549) several times on Slack that `hedgehog` shouldn't be used unless there's a specific reason for that. If you want to expose yourself to the horrors of shrinking in `hedgehog`, you can read [this](https://well-typed.com/blog/2019/05/integrated-shrinking/) post.

`QuickCheck` sucks, but `hedgehog` sucks even more.

> I had no idea this exists (we need an AI bot to help us find such things)

I [did](https://github.com/IntersectMBO/plutus/pull/7319#discussion_r2333055897) point you towards this file.

plutus-core/plutus-core/test/Value/Spec.hs
zliu41: > You're very unlikely to generate small ints this way as far as I can tell.

I thought `enumBounded` isn't uniform, but prefers numbers near the origin (0). I'd be surprised if it was uniform.

> You're also not gonna hit interesting numbers like 2^15-1, 2^10 etc.

It's `Integer`. How are large values more interesting than small values?

plutus-core/plutus-core/src/PlutusCore/Value.hs
effectfully: > What if the `f` you have in mind?

`(,) Bool` or whatever, something allowing you to return additional info together with the result of `insert`.

plutus-core/plutus-core/src/PlutusCore/Value.hs
ana-pantilie: You also need to check if `amt == 0`. In that case, the operation is equivalent to `deleteCoin currency token`.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -91,8 +104,48 @@ fromList =
...
+        inner' = Map.insert token amt inner
+        outer' = Map.insert currency inner' outer
+        sizes' = if exists then sizes else updateSizes (Map.size inner) (Map.size inner') sizes
+        size' = if exists then size else size + 1
+     in Value outer' sizes' size'
+{-# INLINEABLE insertCoin #-}
+
+-- | \(O(n)\), where \(n\) is the total size of the maps.
+unionValue :: Value -> Value -> Value
+unionValue v v' = pack (Map.unionWith (Map.unionWith (+)) (unpack v) (unpack v'))
```
ana-pantilie: See https://github.com/IntersectMBO/plutus/pull/7255/files#diff-0718f41c4fd914c8274438e1d1af10d48fe0e6fd738399c03a886f4ae85d474e. This was the best implementation of `union` I could get, maybe it can be modified to also take care of the sizes. The idea is that `unionWith` doesn't work, unless you also add a separate pass which normalizes the map (removes those elements of `0` amount, and the empty inner maps). That was obviously worse because of the separate pass, and using `merge` you can avoid the separate pass.

ana-pantilie: In case the link doesn't work, the idea is to use `Map.merge` with `Map.preserveMissing` and `Map.zipWithMaybeMatched`, see the implementation in `NestedMap.hs`.

plutus-core/plutus-core/test/Value/Spec.hs
effectfully: Looking at

```haskell
enum :: (MonadGen m, Enum a) => a -> a -> m a
enum lo hi =
  fmap Int -> a
forall a. Enum a => Int -> a
toEnum . integral $
    Range.constant (fromEnum lo) (fromEnum hi)
```

I think it is uniform?

In any case, that's besides the point, the origin here is not `0`, it's `INT_MIN`, if I'm reading the code correctly.

> It's `Integer`. How are large values more interesting than small values?

They're all interesting. You want small, you want large, you want a bit in the middle, some at the boundaries of integer types etc. Perhaps some at 10^n boundaries too, although we don't generate that at the moment.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: Good idea. Yeah, that should work

zliu41:
[d925bff](https://github.com/IntersectMBO/plutus/pull/7322/commits/d925bff449d35af9b9458b9d04015fd08bc2a6fd)  makes `insertCoin` faster by applying @effectfully 's idea of using `alterF`.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: Updated.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: Updated using your implementation. 

> maybe it can be modified to also take care of the sizes. 

This seems prohibitively complex and error prone. I don't think it's worth doing so. Re-calculating the size only takes time proportional to the size of the outermaps, which is lower than the cost of union itself.

plutus-core/plutus-core/test/Value/Spec.hs
zliu41: > You want small, you want large, you want a bit in the middle, some at the boundaries of integer types etc

I don't think I agree with this. The `Integer` in Values are very boring: the only thing we do is adding them. It doesn't matter if we add 10^100 to 10^100, or 2 to 2.

The only interesting thing is adding a number with its negation. Since the result is 0, it should be removed from the Value. To hit this more often, we should in fact keep the range _small_ (say within the bounds of Int8).

plutus-core/plutus-core/test/Value/Spec.hs
zliu41: We in fact do not need to generate `0` at all. There's no zero amount in `Value`!

The only reaso to generate `0` is as an argument to `insertCoin`. But that should be a separate generator, not the generator for `Value` itself.

zliu41:
Ok, I've added the `Arbitrary` instances for the built-in Value, which is similar to the instance for the non-builtin Value in plutus-ledger-api.

There are some difference though:
- for built-in value it doesn't make sense to generate zero amount often, because there's no zero amount in built-in value.
- for built-in value it makes sense to favor small-absolute-value amounts, so we hit interesting cases like `a + (-a)` often. In the non-builtin value, `a + (-a)` isn't an interesting case.

zliu41:
Ok, ported the tests to QC. It reminded me that while QC may be better at shrinking, Hedgehog does have a much nicer API.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: good point, updated

zliu41:
Merged to enable costing and conformance testing. Additional comments are welcome and will be addressed separately.

plutus-core/plutus-core/test/Value/Spec.hs
effectfully: > We in fact do not need to generate 0 at all. There's no zero amount in `Value`!

That's not how property-testing works, you're not supposed to reduce coverage just because you're sure that some number is not going to trigger a bug. In fact, if you didn't do that here, you would've noticed that `insertCoin` is buggy, because the generator for amounts is used in `prop_insertCoinPreservesInvariants`.

zliu41: I agree. I thought we were talking about the amounts inside a `Value`, which must be non-zero. The amount argument to `insertCoin` certainly can be zero. I had an impression that at least the Hedgehog generators will try to generate corner cases like 0 and maxBound without explicitly requesting it to do so. Is the default QuickCheck generator just completely uniform?

plutus-core/plutus-core/test/Value/Spec.hs
effectfully: > It doesn't matter if we add 10^100 to 10^100, or 2 to 2.

That's not how property-testing works, you're not supposed to reduce coverage just because you're sure that some number is not going to trigger a bug.

> Since the result is 0, it should be removed from the Value. To hit this more often, we should in fact keep the range small (say within the bounds of Int8).

No, it means that we need to generate

1. `Int8` numbers fairly often
2. specific `n` and `-n` occasionally
3. cover large numbers as well for general coverage and in case there can be an unexpected overflow somewhere etc

The `ArbitraryBuiltin` generator for integers gives you all of those. It generates numbers in these ranges with the specified probabilities:

```
        13.01% (1, 127)
        12.13% (-1, -127)
        10.91% (128, 32767)
        10.66% (-128, -32767)
         9.12% (-9223372036854775808, -147573952589676412912)
         9.02% (9223372036854775808, 147573952589676412912)
         8.16% (2147483648, 9223372036854775807)
         8.07% (-2147483648, -9223372036854775807)
         7.53% (32768, 2147483647)
         7.35% (-32768, -2147483647)
         4.04% 0
```

some of which are specific corner cases:

```
         4.04% 0
         1.87% 1
         1.85% 127
         1.81% -1
         1.67% -127
         1.57% -128
         1.54% -32767
         1.54% 32767
         1.49% 128
         1.36% -9223372036854775808
         1.29% -2147483648
         1.26% 9223372036854775807
         1.26% 9223372036854775808
         1.24% -147573952589676412912
         1.17% 2147483647
         1.14% -32768
         1.11% 147573952589676412912
         1.10% -9223372036854775807
         1.08% -2147483647
         1.06% 2147483648
         0.95% 32768
```

Not only does it have uncastrated coverage, but you're guaranteed to hit `n + (-n)` more often than by naively generating uniformly distributed `Int8`s.

The generator is not perfect, I do believe that we should consider `(-10, 10)` (or something) as a separate range or make use of the size parameter within each range (this is perhaps the best solution), but it's still vastly better than naively generating `Int8`s.

zliu41: Makes sense - I'll use `ArbitraryBuiltin @Integer` then.

plutus-core/plutus-core/src/PlutusCore/Value.hs
effectfully: Turning two traversals into one isn't necessarily an improvement, particularly when you introduce a space leak in the process (it's a classic one, see [1](https://h2.jaguarpaw.co.uk/posts/nested-strict-data/), [2](https://neilmitchell.blogspot.com/2015/09/three-space-leaks.html), [3](https://www.tweag.io/blog/2021-11-04-haskell-profiling-cachegrind/), [4](https://www.haskellforall.com/2013/08/composable-streaming-folds.html)), like in this instance (unless GHC is smart enough to fix it, which it usually is not).

Without looking at the GHC Core here you can't have any idea if you made things better or worse.

plutus-core/plutus-core/test/Value/Spec.hs
effectfully: > I thought we were talking about the amounts inside a `Value`

It's not the job of a test to even consider that -- it's the job of the `normalize` function. The test should make no assumptions about it or even its existence at all. Just generate `0` occasionally and let the function converting a list to a `Value` take care of it.

> I had an impression that at least the Hedgehog generators will try to generate corner cases like 0 and maxBound without explicitly requesting it to do so. 

Nope, `hedgehog` will certainly generate `0`, but its whole design is "create your own generators and make sure they're exactly what you want", which is bad design, because few people ever create generators that are not complete trash. `QuickCheck` has a better design (one `Arbitrary` instance for each type incentivizes to cover as many corner cases in it as possible), but when it comes to actual generators provided by the library, or the advice on how to create one, they're trash too.

plutus-core/plutus-core/test/Value/Spec.hs
effectfully: If you ever feel like it doesn't cover some edge case that would be nice to cover, chances are, it may be nice for other tests as well, and so please consider amending the generator. We don't care how complex, unprincipled or silly it is -- we only care that it hits all the right spots. So we can make it a completely unreadable mess of spaghetti code with magical numbers and overlapping logic -- and still call it a good generator. That's where it's currently going anyway.

PR #5061, opened 2023-01-21
PLT-1432: [Builtins] Make 'geq' inlinable

effectfully:
@michaelpj made me learn about loop-breakers, hence this PR achieving more or less the same as #4462 and #4463 except without as much burden.

My quick-and-dirty benchmarking suggests that this does give us a speedup, plus the Core definitely looks better.

Manual benchmarking is a pain in the ass, so let's wait for automated benchmarking to arise from the ashes, hopefully that'll happen soonish. Unless of course somebody wants to volunteer.

PR #4463, opened 2022-03-10
[Builtins] Unfold 'geq' for the default universe

effectfully:
Don't look here yet.

PR #4462, opened 2022-03-10
[Builtins] Add an inlinable version of 'GEq'

effectfully:
Don't look here yet.

PR #7323, opened 2025-09-09
[Builtins] Add an inlinable version of 'geq'

effectfully:
At long last I think I've finally discovered a way to create an inlinable version of `geq`.

Previous attempts: #4462, #4463, #5061, #6352.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
@@ -26,7 +26,6 @@ import PlutusCore.Evaluation.Machine.ExBudgetStream (ExBudgetStream)
 import PlutusCore.Evaluation.Machine.ExMemoryUsage (ExMemoryUsage, IntegerCostedLiterally (..),
                                                     NumBytesCostedAsNumWords (..), memoryUsage,
                                                     singletonRose)
-import PlutusCore.Evaluation.Result (EvaluationResult (..))
 import PlutusCore.Pretty (PrettyConfigPlc)
 import PlutusCore.Value (Value)
 import PlutusCore.Value qualified as Value


effectfully:
But how...

effectfully:
Numbers are a little over the place, but they're consistently negative across the three runs

image


I'm gonna call this a 4.3% performance improvement of `Constr`, which is quite great, given how ubiquitous this builtin is.

effectfully:
Damn. Brb.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -131,68 +134,146 @@ pattern DefaultUniArray uniA =
 pattern DefaultUniPair uniA uniB =
     DefaultUniProtoPair `DefaultUniApply` uniA `DefaultUniApply` uniB
 
+-- Removing 'LoopBreaker' didn't change anything at the time this comment was written, but we kept
+-- it, because it hopefully provides some additional assurance that 'geqL' will not get elaborated
+-- as a recursive definition.
+instance AllBuiltinArgs DefaultUni (GEqL DefaultUni) a => GEqL DefaultUni a where
```
zliu41, line 137: What does the superclass constraint, `AllBuiltinArgs...`, do? It doesn't seem to provide any method needed to implement `geqL`?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -131,68 +134,146 @@ pattern DefaultUniArray uniA =
...
+-- lazy 'Maybe'.
+geqLDefaultUni
+    :: forall (a :: GHC.Type) (b :: GHC.Type).
+       DefaultUni (Esc a)
+    -> DefaultUni (Esc b)
+    -> EvaluationResult (a :~: b)
+-- It's not clear why GHC decides to inline 'bring' and how exactly that happens given that it's a
+-- recursive function. So that can easily break in future, however it would introduce a very small
+-- slowdown and only for builtins that do runtime type tag equality checks, of which at the time
+-- this comment was written we had only 'MkCons'.
```
effectfully: I've tried playing with `bring`, but man is it incredibly annoying to optimize.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -131,68 +134,146 @@ pattern DefaultUniArray uniA =
...
+        pure Refl
+    geqL DefaultUniUnit a2 = do
+        DefaultUniUnit <- pure a2
+        pure Refl
+    geqL DefaultUniBool a2 = do
+        DefaultUniBool <- pure a2
+        pure Refl
+    geqL (DefaultUniProtoList `DefaultUniApply` a1) listA2 = do
+        DefaultUniProtoList `DefaultUniApply` a2 <- pure listA2
+        Refl <- geqL (LoopBreaker a1) (LoopBreaker a2)
```
zliu41, line 155: Can you not apply the same `LoopBreaker` trick to `geq`?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -131,68 +134,146 @@ pattern DefaultUniArray uniA =
...
+        -- Splitting the definition in two allows GHC to inline the initial step that appears
+        -- non-recursive to GHC, because recursion is hidden in the other function that is marked as
+        -- @NOINLINE@ and is chosen by GHC as a loop-breaker, see
+        -- https://wiki.haskell.org/Inlining_and_Specialisation#What_is_a_loop-breaker (we're not
+        -- really sure if this is a reliable solution, but if it stops working, we won't miss very
+        -- much and we've failed to settle on any other approach).
+        --
+        -- We use @NOINLINE@ instead of @OPAQUE@, because we don't actually care about the recursive
+        -- definition not being inlined, we just want it to be chosen as a loop breaker.
+        goStep, goRec :: DefaultUni a1 -> DefaultUni a2 -> Maybe (a1 :~: a2)
```
zliu41, line 202: Is there actual meaningful change here, or just renaming?

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: It's there so that the recursive call instead of being elaborated as a recursive call becomes a lookup in the `AllBuiltinArgs` dictionary. You can see `GEqL` being the constraint that `AllBuiltinArgs` is applied to. Now instead of having term-level recursion, you have type-level recursion -- and that one is inlining-friendly.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: It's not `LoopBreaker` that is the trick. It's turning term-level recursion into type-level recursion. `LoopBreaker` is just to make it harder for GHC to see through the actual trick. You can see a comment above this instance, which says that `LoopBreaker` doesn't actually seem to impact anything, I just put it in there to increase the odds GHC won't rebel.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: Just renaming, I was playing with other stuff and decided that there should be a convention to call such functions `goStep` and `goRec`.

effectfully:
OK, it's back to being good

image

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
zliu41: I probably should re-review this then. I thought by having `LoopBreaker`, the `geqL` here is no longer a recursive call, but it calls the `geqL` of a different instance - `LoopBreaker`'s, hence it makes `geqL` inlinable. But that's not how it works?

effectfully:
OK, it's back to being good.

image

effectfully:
OK, it's back to being good.

image

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
SeungheonOh: Just out of curiocity, what happens if you give `AllBuiltinArgs` constraint to original `GEq` instance definition? I reckon it will break stuff that requires eq-ing statically unknown value, but will it at least inline everything when builtin type is known?

If so, why can't we just have `instance AllBuiltinArgs ... => GEq (LoopBreaker DefaultUni) a` instead of having a whole new `GEqL` typeclass?

effectfully: > If so, why can't we just have instance `AllBuiltinArgs ... => GEq (LoopBreaker DefaultUni) a` instead of having a whole new GEqL typeclass?

You can't, `AllBuiltinArgs` expects an `a` as its last argument, but `GEq` doesn't fix one in place, because it's `GEq DefaultUni`, not `GEq DefaultUni a`.

Maaaybe we could do something like ``DefaultUni `Everywhere` AllBuiltinArgs DefaultUni (GEq DefaultUni) => GEq DefaultUni` but note how the constraint doesn't have any type variables whatsoever, nothing is stopping GHC from immediately resolving it or even discarding it completely. We want the constraint to be resolved recursively while `a` is still statically known.

plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
effectfully: Here's what 

```haskell
instance GEq uni => GEq (LoopBreaker uni) where
    geq (LoopBreaker u1) (LoopBreaker u2) = geq u1 u2
    {-# INLINE geq #-}

instance GEq DefaultUni where
    <...>
    geq DefaultUniProtoPair a2 = do
        DefaultUniProtoPair <- pure a2
        pure Refl
    geq (DefaultUniApply f1 x1) a2 = do
        DefaultUniApply f2 x2 <- pure a2
        Refl <- geq (LoopBreaker f1) (LoopBreaker f2)
        Refl <- geq (LoopBreaker x1) (LoopBreaker x2)
        pure Refl
    geq DefaultUniData a2 = do
        DefaultUniData <- pure a2
        pure Refl
    <...>
    {-# INLINE geq #-}
```

compiles to:

```haskell
Rec {
$fGEqTYPEDefaultUni_$cgeq
  :: forall a b. DefaultUni a -> DefaultUni b -> Maybe (a :~: b)
$fGEqTYPEDefaultUni_$cgeq
  = \ (@a_akgO)
      (@b_akgP)
      (ds_do1w :: DefaultUni a_akgO)
      (a2_aiqJ :: DefaultUni b_akgP) ->
      case ds_do1w of {
        <...>
        DefaultUniProtoPair co_akjq ->
          case a2_aiqJ of {
            __DEFAULT -> Nothing;
            DefaultUniProtoPair co1_akjz -> Just ($WRefl `cast` <Co:7> :: ...)
          };
        DefaultUniApply @k_akjM @k1_akjN @f1_akjO @a1_akjP co_akjQ f2_aiqR
                        x1_aiqS ->
          case a2_aiqJ of {
            __DEFAULT -> Nothing;
            DefaultUniApply @k2_akjZ @k3_akk0 @f3_akk1 @a3_akk2 co1_akk3
                            f4_aiqU x2_aiqV ->
              case $fGEqTYPEDefaultUni_$cgeq f2_aiqR f4_aiqU of {
                Nothing -> Nothing;
                Just x_ippB ->
                  let! { Refl co2_akkk ~ <- x_ippB } in
                  case $fGEqTYPEDefaultUni_$cgeq x1_aiqS x2_aiqV of {
                    Nothing -> Nothing;
                    Just x4_X5 ->
                      let! { Refl co3_akkB ~ <- x4_X5 } in
                      Just ($WRefl `cast` <Co:48> :: ...)
                  }
              }
          };
        DefaultUniData co_akkQ ->
          case a2_aiqJ of {
            __DEFAULT -> Nothing;
            DefaultUniData co1_akkZ -> Just ($WRefl `cast` <Co:7> :: ...)
          };
        <...>
      }
end Rec }
```

And even if I make it `OPAQUE`:

```haskell
instance GEq uni => GEq (LoopBreaker uni) where
    geq (LoopBreaker u1) (LoopBreaker u2) = geq u1 u2
    {-# OPAQUE geq #-}
```

the `GEq uni` constraint gets resolved by passing the `geq` method of the `GEq DefaultUni` instance creating a recursive definition, so even in that case nothing inlines.

The comment above the `GEqL DefaultUni a` instance says

> -- Removing 'LoopBreaker' didn't change anything at the time this comment was written, but we kept
> -- it, because it hopefully provides some additional assurance that 'geqL' will not get elaborated
> -- as a recursive definition.

so yeah, `LoopBreaker` is there "just in case", it's neither necessary nor sufficient right now.

I did check that turning `geqL` into a top-level definition creates a recursive function without `LoopBreaker`, so I do believe it's worth keeping the latter, because none of this is reliable and we don't know how GHC will choose to stab us in the back in future.

Issue #7130, opened 2025-06-04
Rename UPLC goldens to have file extension `.uplc`


SeungheonOh:
Thanks to @rvcas, we have language support to uplc now natively on github! We can change all golden files with extension `*.uplc.golden` to `*.golden.uplc` to take advantage of github syntax highlighting and language recognition. 

We also want other files like `*.pir.golden` to also be renamed to `*.golden.pir` for consistency.

Unisay:
We have a few ways to render UPLC to text:
|Name|Description|
|---|---|
|classic|Lisp-like syntax with unique variable names  (default)|
|classic-simple|Lisp-like syntax with ambiguous (no unique) variable names|
|readable|Succinct syntax with unique variable names|
|readable-simple|Succinct syntax with ambiguous (no unique) variable names|

What exactly is supported by Github?

SeungheonOh:
It appears it's classic, which is what's used in the UPLC spec itself

PR #7332, opened 2025-09-11
Rename golden files so that .golden is not the final extension

basetunnel:
This changes all `file.ext.golden` to `file.golden.ext` (#7130). This means that `.uplc` golden files can be properly highlighted on github. In case of multiple extensions `file.A.B.golden` changes to `file.golden.A.B`.


SeungheonOh (approved):
Thanks. Looks good to me

PR #7334, opened 2025-09-13
Add `insertCoin` and `unionValue` to `DefaultFun` and Plinth

zliu41:
As titled


plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -2256,6 +2281,8 @@ instance Flat DefaultFun where
               go 89 = pure LengthOfArray
               go 90 = pure ListToArray
               go 91 = pure IndexArray
+              go 94 = pure InsertCoin
```
kwxm, line 2310: Why the gap between 94 and 97?  Are we expecting it to be filled by some more Value-related functions?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -2015,6 +2021,22 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
           {-# INLINE indexArrayDenotation #-}
         in makeBuiltinMeaning indexArrayDenotation (runCostingFunTwoArguments . paramIndexArray)
 
+    toBuiltinMeaning _semvar InsertCoin =
+      let insertCoinDenotation :: ByteString -> ByteString -> Integer -> Value -> Value
+          insertCoinDenotation = Value.insertCoin
+          {-# INLINE insertCoinDenotation #-}
+       in makeBuiltinMeaning
+            insertCoinDenotation
+            (runCostingFunFourArguments . unimplementedCostingFun)
```
kwxm, line 2051: This will be the first four-argument builtin.  Luckily the basic infrastructure for four-argument costing functions is already in place, but we'll probably have to extend [this type](https://github.com/IntersectMBO/plutus/blob/ddc27116cfbb8926f933a2d5c13f9be9085d0da0/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs#L741) and related functions to deal with `insertValue`.  We probably don't want to do that until the final implementation of the `Value` type is decided though, although if that decision depends on costs we might have to add costing machinery for the different candidates.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
zliu41: Yes. I will add them in the order written in the CIP.

kwxm (approved):
This looks fine.  The process of adding new built-in functions and types to `plutus-tx` is a bit boilerplatey, so I wonder if it would be worth having a document or chekclist saying exactly what you have to do.  We've got such things for `plutus-core` and `plutus-ledger-api` but there's a gap for `plutus-tx`.

I got a bit worried whether there was a danger of confusion with [the `Value` type](https://github.com/IntersectMBO/plutus/blob/ddc27116cfbb8926f933a2d5c13f9be9085d0da0/plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs#L268) (and the other version [here](https://github.com/IntersectMBO/plutus/blob/ddc27116cfbb8926f933a2d5c13f9be9085d0da0/plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs#L290)) in `plutus-ledger-api`, but I don't think we ever have to talk about built-in types there so maybe it doesn't matter.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
zliu41: This will be a four argument builtin regardless of the underlying implementation.

zliu41:
> I got a bit worried whether there was a danger of confusion with [the Value type](https://github.com/IntersectMBO/plutus/blob/ddc27116cfbb8926f933a2d5c13f9be9085d0da0/plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs#L268) (and the other version [here](https://github.com/IntersectMBO/plutus/blob/ddc27116cfbb8926f933a2d5c13f9be9085d0da0/plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs#L290)) in plutus-ledger-api

The non-built-in `Value` type will be gone in the near future.

kwxm:
> This looks fine.

~I spoke too soon!  There are some failing tests in `cardano-constitution-test`.  I think that's just UPLC code getting rearranged though.~

[Never mind: I think you fixed that when I was typing.]

kwxm:
> > I got a bit worried whether there was a danger of confusion with [the Value type](https://github.com/IntersectMBO/plutus/blob/ddc27116cfbb8926f933a2d5c13f9be9085d0da0/plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs#L268) (and the other version [here](https://github.com/IntersectMBO/plutus/blob/ddc27116cfbb8926f933a2d5c13f9be9085d0da0/plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs#L290)) in plutus-ledger-api
> 
> The non-built-in `Value` type will be gone in the near future.

Will we not need that for old scripts?

zliu41:
> Will we not need that for old scripts?

You are right, we do. Then we should rename it to avoid the confusion (`PoorMan'sValue`? :smile: )

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
kwxm: > This will be a four argument builtin regardless of the underlying implementation.

Yes, but we'll have to add an extra constructor [here](https://github.com/IntersectMBO/plutus/blob/e01947b412acf8037876c48121a11ebd5489f638/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs#L741) and code to handle it [here](https://github.com/IntersectMBO/plutus/blob/e01947b412acf8037876c48121a11ebd5489f638/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs#L759) (and maybe elsewhere as well), and exactly what we add will depend on how the implementation behaves.  It's probaby not a big deal, but we won't be able to re-use any existing code.

PR #7336, opened 2025-09-15
Add `lookupCoin` and `valueContains`

zliu41:
as titled


plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -148,12 +152,38 @@ insertCoin currency token amt v@(Value outer sizes size)
...
+lookupCoin currency token (unpack -> outer) = case Map.lookup currency outer of
+  Nothing    -> 0
+  Just inner -> Map.findWithDefault 0 token inner
+
+-- | \(O(n_{2}\log \max(m_{1}, k_{1}))\).
+--
+-- @a@ contains @b@ only if all amounts in @b@ are positive.
+valueContains :: Value -> Value -> Bool
+valueContains v =
+  all (\(c, t, a2) -> let a1 = lookupCoin c t v in a2 > 0 && a1 >= a2) . toFlatList
```
ana-pantilie: Is it really necessary to use `toFlatList`? Why not fold over the original (nested) map and this way you can avoid traversing the whole thing twice.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -148,12 +152,38 @@ insertCoin currency token amt v@(Value outer sizes size)
...
+
+-- | \(O(\log \max(m, k))\)
+lookupCoin :: ByteString -> ByteString -> Value -> Integer
+lookupCoin currency token (unpack -> outer) = case Map.lookup currency outer of
+  Nothing    -> 0
+  Just inner -> Map.findWithDefault 0 token inner
+
+-- | \(O(n_{2}\log \max(m_{1}, k_{1}))\).
+--
+-- @a@ contains @b@ only if all amounts in @b@ are positive.
```
ana-pantilie: Someone might understand that `valueContains` implements only `@a@ contains @b@ only if all amounts in @b@ are positive.`, when that's not the full picture, since `b` should only contain coins which also appear in `a`.

Nevertheless, we should document this function, and the others as well.

cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests/sorted.cbor.size.golden
```diff
@@ -1 +1 @@
-2042
\ No newline at end of file
+2047
```
ana-pantilie: Why did these tests change? Are scripts now larger because we introduced new builtins?

zliu41: It just somehow caused GHC to generate slightly different Core.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: Good point, I meant to change it but forgot.

cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests/sorted.cbor.size.golden
kwxm: This happens quite a lot and it generally looks as if it's caused by changes in code that's been subjected to CSE (or at least if I look at the UPLC I see changes involving `\cse ->`).  I wonder if it indicates that we're missing some opportunity for optimising a little more efficiently?

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -148,12 +152,43 @@ insertCoin currency token amt v@(Value outer sizes size)
...
-  pack $ case Map.lookup currency outer of
-    Nothing    -> outer
-    Just inner -> Map.insert currency (Map.delete token inner) outer
+deleteCoin currency token (Value outer sizes size) = Value outer' sizes' size'
+ where
+  (mold, outer') = Map.alterF f currency outer
+  (sizes', size') = case mold of
+    Just old -> (updateSizes old (old - 1) sizes, size - 1)
+    Nothing  -> (sizes, size)
+  f
```
kwxm, line 167: The formatting of the signature's a bit weird. Maybe it just looks unusual because the function name only has one letter.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -148,12 +152,43 @@ insertCoin currency token amt v@(Value outer sizes size)
...
+
+-- | \(O(\log \max(m, k))\)
+lookupCoin :: ByteString -> ByteString -> Value -> Integer
+lookupCoin currency token (unpack -> outer) = case Map.lookup currency outer of
+  Nothing    -> 0
+  Just inner -> Map.findWithDefault 0 token inner
+
+{-| \(O(n_{2}\log \max(m_{1}, k_{1}))\).
+
+@a@ contains @b@ if all amounts in @b@ are positive, and each token in @b@ exists
```
kwxm: The positivity requirement differs from what's in the CIP (and I see that that's been discussed at some length on Slack), so maybe there should be an explanatory comment.

kwxm: You might expect `valueContains` to make `Value` into a partially ordered monoid/group, but the positivity requirement breaks that completely.  For instance if you have any value `v` containing a negative quantity, `valueContains v v` is `False,` which is kind of unusual.  It might be better to fail if the second argument contains a negative quantity: I think that would make `Value` obey sensible laws in all non-error cases (although thinking a bit more, matters may be complicated by the fact that there's an asymmetry because `a` _is_ allowed to contain negative values)

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -61,6 +61,33 @@ prop_insertCoinIdempotent = forAll arbitrary $ \v ->
...
+       in V.lookupCoin currency token v' === 0
+
+prop_deleteCoinIdempotent :: Property
+prop_deleteCoinIdempotent = forAll (arbitrary `suchThat` (\v -> V.totalSize v > 0)) $ \v ->
+  let fm = V.toFlatList v
+   in forAll (elements fm) $ \(c, t, _) ->
+        let v' = V.deleteCoin c t v
+         in v' === V.deleteCoin c t v'
+
+-- `valueContains` is reflexive for values that only contains positive amounts,
```
kwxm: It might be worth adding more tests here, for example that `valueContains` is a partial order on values where all of the quantities are nonnegative.   The definition of `valueContains` is moderately complicated so we should be sure that it's correct.  In a property test it might be difficult to hit meaningful cases like `a >= b && b >= c => a >= c` though (I haven't looked closely at the generator yet, so I could be wrong about that). Maybe we could have a tediously large number of unit tests for that kind of thing in the conformance tests?

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -148,12 +152,43 @@ insertCoin currency token amt v@(Value outer sizes size)
...
+      let (amt, inner') = Map.updateLookupWithKey (\_ _ -> Nothing) token inner
+       in (amt $> Map.size inner, if Map.null inner' then Nothing else Just inner')
+
+-- | \(O(\log \max(m, k))\)
+lookupCoin :: ByteString -> ByteString -> Value -> Integer
+lookupCoin currency token (unpack -> outer) = case Map.lookup currency outer of
+  Nothing    -> 0
+  Just inner -> Map.findWithDefault 0 token inner
+
+{-| \(O(n_{2}\log \max(m_{1}, k_{1}))\).
```
kwxm: What do  the subscripts mean?  (I know the answer to that, but maybe the comment should spell it out, or there could be a meta-comment about complexity bounds earlier on.)

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -148,12 +152,43 @@ insertCoin currency token amt v@(Value outer sizes size)
...
+
+@a@ contains @b@ if all amounts in @b@ are positive, and each token in @b@ exists
+in @a@ with at least the same amount.
+-}
+valueContains :: Value -> Value -> Bool
+valueContains v = Map.foldrWithKey' go True . unpack
+ where
+  go c inner = (&&) (Map.foldrWithKey' goInner True inner)
+   where
+    goInner t a2 = (&&) (let a1 = lookupCoin c t v in a2 > 0 && a1 >= a2)
```
kwxm: I found this a little confusing, possibly because I'm thinking too strictly.  Maybe 

```
    goInner t a2 = (&&) (a2 > 0 && lookupCoin c t v >= a2)
```
would be a little clearer.

kwxm (approved):
This looks OK to me.  Costing's going to be fun because we'll have to generate worst-case inputs for all of these functions.

plutus-core/plutus-core/src/PlutusCore/Value.hs
ana-pantilie: Good point. What does this function mean if `a` contains negative quantities? Should both inputs to `valueContains` be from `Value+` (meaning `Value` restricted to positive quantities) where it'd form an additive monoid with a partial ordering?

cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests/sorted.cbor.size.golden
zliu41: Perhaps, but there's a lot of re-ordering of things. Very difficult tell what actually changed.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: Yeah this is something we can tweak when we fork fourmolu.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: > Should both inputs to valueContains be from Value+ (meaning Value restricted to positive quantities) where it'd form an additive monoid with a partial ordering?

No, because to check the condition requires a linear traversal of the first Value. I'm fine with other sensible semantics, as long as it permits an efficient implementation. Suggestions welcome.

plutus-core/plutus-core/test/Value/Spec.hs
zliu41: I can indeed add some more properties for `valueContains`, though not necessarily this specific property.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: This is indeed very unsatisfying, mathematically. I don't have a better idea, so I may actually implement what @ana-pantilie suggested - requiring both values to be positive-only, otherwise error.

kwxm: > requiring both values to be positive-only, otherwise error.

As you point out, that could be expensive. Once the costing benchmarks are in place you could try it and see if the extra expense is a problem.

zliu41: Not necessarily, I can keep track of positiveness with a boolean flag.

effectfully (approved):

plutus-core/plutus-core/test/Value/Spec.hs
zliu41: Added another property, `prop_containsAfterDeletion`.

PR #7338, opened 2025-09-16
Implement ValueData and UnValueData

zliu41:
as titled


plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
```diff
@@ -725,6 +727,8 @@ defineBuiltinTerms = do
           -- Value
           PLC.InsertCoin -> defineBuiltinInl 'Builtins.insertCoin
           PLC.UnionValue -> defineBuiltinInl 'Builtins.unionValue
+          PLC.ValueData -> defineBuiltinInl 'Builtins.mkValue
```
Unisay, line 730: Is there a reason to have different names `ValueData` and `mkValue` ?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -192,6 +192,8 @@ data DefaultFun
     -- Values
     | InsertCoin
     | UnionValue
+    | ValueData
+    | UnValueData
```
Unisay, line 196: I like these names more:
```suggestion
    | ValueAsData
    | DataAsValue
```

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -185,6 +189,38 @@ unionValue (unpack -> vA) (unpack -> vB) =
       vB
 {-# INLINEABLE unionValue #-}
 
+-- | \(O(n)\). Encodes `Value` as `Data`, in the same way as non-builtin @Value@.
```
Unisay, line 192: > in the same way as non-builtin @Value@

I don't understand what this phrase means, given that this function is working with the non-builtin `Value`.

plutus-tx/src/PlutusTx/Builtins/Internal.hs
```diff
@@ -1092,6 +1092,19 @@ unionValue :: BuiltinValue -> BuiltinValue -> BuiltinValue
 unionValue (BuiltinValue v1) (BuiltinValue v2) = BuiltinValue $ Value.unionValue v1 v2
 {-# OPAQUE unionValue #-}
 
+mkValue :: BuiltinValue -> BuiltinData
+mkValue (BuiltinValue v) = BuiltinData $ Value.valueData v
```
Unisay, line 1096: `valueAsData` name is more consistent with `unsafeDataAsValue` below (modulo `unsafe` prefix)

Unisay (commented):
Looks good to me except the choise of names.

effectfully (approved):

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
zliu41: The names are consistent with other Data builtins (IData, UnIData etc.)

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: No, this function works with builtin Value.

plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs
zliu41: Not sure, but that's how other Data primitives are named (mkI, mkB etc.)

plutus-tx/src/PlutusTx/Builtins/Internal.hs
zliu41: The reason is for consistency with the naming of existing Data primitives.

PR #7339, opened 2025-09-16
Add BLS MSM builtins to metatheory

kwxm:
This adds the [BLS12-381 multi-scalar multiplication builtins](https://github.com/IntersectMBO/plutus/pull/7283) to the metatheory.


SeungheonOh (approved):
LGTM, although I'm not the most qualified person to review metatheory changes

plutus-metatheory/src/Builtin.lagda.md
```diff
@@ -118,7 +118,7 @@ data Builtin : Set where
   mkPairData                      : Builtin
   mkNilData                       : Builtin
   mkNilPairData                   : Builtin
-  -- BLS12_381
+  -- BLS12-381
```
ana-pantilie: I think this needs to be removed now.

kwxm: Good point.  I'm not sure what to do there.  It might make sense to put the two new functions here, but it might also be sensbile to keep them in the order that they were added.  I've left the new functions where they are but changed the comment to say "Initial BLS12-381 functions".  Things are a bit confused now that we keep adding new functions.  Maybe we could just use the batch numbers that we use in the specification and [in plutus-ledger-api](https://github.com/IntersectMBO/plutus/blob/fa40064e639b97df1559b1947299f2193f81fa6a/plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs#L147)  that group together all the builtins that were released at a particular HF.

kwxm:
I have no idea why this is in a branch called `kwxm/spec/add-bls-msm` when it's supposed to just be about that metatheory.  I don't think it's interfering with anything else though, so it should be OK to merge it.

PR #7343, opened 2025-09-18
Property tests for BLS12-381 multi-scalar multiplication

### LINKED ISSUE #7324: BLS12-381 MSM tests
We need some tests, including [property tests](https://github.com/IntersectMBO/plutus/blob/master/plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/BLS12_381.hs) for the two BLS MSM functions.

kwxm:
Closes #7324.
Closes https://github.com/IntersectMBO/plutus-private/issues/1651.

### DIFF
--- a/plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/BLS12_381.hs
+++ b/plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/BLS12_381.hs
@@ -29,6 +29,8 @@ import Test.QuickCheck hiding (Some (..))
 import Test.Tasty
 import Test.Tasty.QuickCheck hiding (Some (..))
 
+import PlutusCore.MkPlc (mkConstant)
+
 -- QuickCheck utilities
 
 mkTestName :: forall g. TestableAbelianGroup g => String -> String
@@ -39,11 +41,29 @@ withNTests = withMaxSuccess 200
 
 -- QuickCheck generators for scalars and group elements as PLC terms
 
-arbitraryConstant :: forall g. TestableAbelianGroup g => Gen PlcTerm
-arbitraryConstant = toTerm <$> (arbitrary @g)
+-- An arbitrary nonzero group element.  For the BLS12-381 groups it's highly
+-- unlikely that we'll get the zero element, but let's make sure.
+arbitraryNonZeroConstant :: forall g. TestableAbelianGroup g => Gen PlcTerm
+arbitraryNonZeroConstant = (toTerm <$> (arbitrary @g)) `suchThat` ((/=) (zeroTerm @g))
 
+-- For most tests we want to make sure that we get the zero point reasonably often.
+arbitraryConstant :: forall g. TestableAbelianGroup g => Gen PlcTerm
+arbitraryConstant =
+  frequency [ (9, arbitraryNonZeroConstant @g)
+            , (1, pure (zeroTerm @g))
+            ]
+
+{- Generate an arbitrary scalar.  Scalars map on to elements of Z_p where p is the
+   381-bit prime involved in BLS12_381.  This generator supplies integers up to
+   10000 bits long to give us some confidence that the reduction is handled
+   properly.
+-}
 arbitraryScalar :: Gen PlcTerm
-arbitraryScalar = integer <$> (arbitrary @Integer)
+arbitraryScalar =
+  integer <$>
+  frequency [ (1, arbitrary @Integer)
+            , (4, choose (-b, b))]
+  where b = (2::Integer)^(10000::Integer)
 
 -- Constructing pairing terms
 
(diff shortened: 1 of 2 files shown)


plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/BLS12_381.hs
```diff
@@ -39,11 +41,29 @@ withNTests = withMaxSuccess 200
...
+{- Generate an arbitrary scalar.  Scalars map onto elements of Z_p where p is the
+   255-bit prime order of the groups involved in BLS12_381.  This generator
+   supplies integers up to 10000 bits long to give us some confidence that the
+   reduction is handled properly.
+-}
 arbitraryScalar :: Gen PlcTerm
-arbitraryScalar = integer <$> (arbitrary @Integer)
+arbitraryScalar =
+  integer <$>
+  frequency [ (1, arbitrary @Integer)
```
effectfully: I'm really not a fan of `arbitrary @Integer` for reasons specified in this Note:

```
{- Note [QuickCheck and integral types]
The 'Arbitrary' instances for 'Integer' and 'Int' only generate small integers:

    >>> :set -XTypeApplications
    >>> fmap (any ((> 30) . abs) . concat . concat . concat) . sample' $ arbitrary @[[[Integer]]]
    False
    >>> fmap (any ((> 30) . abs) . concat . concat . concat) . sample' $ arbitrary @[[[Int]]]
    False

We want to generate larger ones, including converted-to-Integer 'minBound' and 'maxBound' of various
integral types. Hence we declare 'nextInterestingBound' and 'highInterestingBound' to specify the
"interesting" ranges to generate positive integers within. We also make it likely to hit either end
of each of those ranges.
-}
```

Use `arbitraryBuiltin @Integer`?

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/BLS12_381.hs
```diff
@@ -239,6 +259,74 @@ test_Z_action_good =
...
+
+{- Check that multiScalarMul [s_1, ..., s_m] [p_1, ..., p_n] =
+      0 + (s_1*p_1) + ... + (s_r*p_r) where r = min m n
+-}
+test_multiScalarMul_correct :: forall g. TestableAbelianGroup g => TestTree
+test_multiScalarMul_correct =
+  testProperty
+  (mkTestName @g "multiScalarMul_is_iterated_mul_and_add") .
+  withNTests $ do
+     scalars <- listOf (arbitrary @Integer)
```
effectfully: And in other places.

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/BLS12_381.hs
```diff
@@ -239,6 +259,74 @@ test_Z_action_good =
...
+     points  <- listOf (arbitrary @g)
+     let e1 = multiScalarMulTerm @g (mkConstant () scalars) (mkConstant () points)
+         mkMulAdd acc (s, x) = addTerm @g acc (scalarMulTerm @g s x)
+         scalarTerms = fmap (mkConstant ()) scalars
+         pointTerms  = fmap (mkConstant ()) points
+         e2 = foldl mkMulAdd (zeroTerm @g) (zip scalarTerms pointTerms)
```
effectfully: Nit: `foldl'` is more appropriate here.

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/BLS12_381.hs
```diff
@@ -437,16 +525,17 @@ test_pairing_balanced =
           e3 = finalVerifyTerm e1 e2
       pure $ evalTerm e3 === cekSuccessTrue
 
--- finalVerify returns False for random inputs
+-- Cheack that `finalVerify` returns False for random inputs.  We exclude the
```
effectfully: Nit: "Check" instead of "Cheack".

effectfully (approved):

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/BLS12_381.hs
kwxm: > I'm really not a fan of `arbitrary @Integer` for reasons specified in this Note:

That's what I was trying to fix!  In this case, we really want to choose an arbitrary integer in the range [0..p-1], where p is a 255-bit prime, and in fact we want bigger ones to make sure that the reduction modulo p is handled correctly, so most of the time the big generator a couple of lines below is used (which I hope generates numbers uniformly, although I haven't checked).  I left the standard generator there so that we'd get things like -1,0,and 1 reasonably often.

effectfully: > I left the standard generator there so that we'd get things like -1,0,and 1 reasonably often.

`arbitraryBuliltin @Integer` is better at that too, because it generates those regardless of the size parameter decently often.

This is `arbitraryBuiltin @Integer`:

```
         4.40% 0        
         1.74% -1
         1.60% 1
```

This is `arbitrary @Integer`:

```
         3.24% 0     
         2.42% -1
         2.23% 1
```

So quite similar, except the latter generates fewer `0`, `-1` and `1` as the size parameter grows while the former generates those at a consistent rate.

Let me know if you want to generate `0` slightly less often or `-1`/`1` slightly more often.

But OK, here's another reason:

> 255-bit prime, and in fact we want bigger ones to make sure that the reduction modulo p is handled correctly

QuickCheck isn't written with that in mind, the shrinker for `Integer` at most divides by 2 and you're gonna wait a long time before shrinking for a failing property stops (or it gives up way too early, I don't remember). The `arbitraryBuiltin` shrinker doesn't have this issue:

```
-- | Same as 'shrinkIntegral' except includes the square root of the given number (or of its
-- negative if the number is negative, in which case the square root is negated too). We need the
-- former because 'shrinkIntegral' at most divides the number by two, which makes the number smaller
-- way too slow, hence we add square root to speed up the process.
--
-- >>> shrinkIntegralFast (0 :: Integer)
-- []
-- >>> shrinkIntegralFast (1 :: Integer)
-- [0]
-- >>> shrinkIntegralFast (9 :: Integer)
-- [0,3,5,7,8]
-- >>> shrinkIntegralFast (-10000 :: Integer)
-- [0,10000,-100,-5000,-7500,-8750,-9375,-9688,-9844,-9922,-9961,-9981,-9991,-9996,-9998,-9999]
```

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/BLS12_381.hs
kwxm: Hmmm.  I did manage to use that in one of the earlier tests but then forgot to do it here: I think I got tripped up there because `foldl` is imported from the prelude.

kwxm:
I think I'll rework these a bit more before merging.  The `arbitraryConstant` function generates a _term_ that has a reasonable chance of being the zero element of the group, but a number of the tests use `arbitrary` for group elements not embedded in terms, and that doesn't make any attempt to hit zero.

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/BLS12_381.hs
kwxm: I used `ArbitraryBuiltin` for small integers, but I'm not too worried about shrinking the huge ones. These tests do pass and if a failure appears for some reason (like a bug in a library function) I don't think that having a not-very-good shrinker will stop us from working out what's wrong.

kwxm:
@effectfully I've reworked this so that the [`Arbitrary` instances for group elements ](https://github.com/IntersectMBO/plutus/blob/6d0e4450e8b097dc50af7fd80633aa3184a66bf3/plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/BLS12_381/TestClasses.hs#L63) now deliberately produce the zero element 10% of the time (the previous version only did that for constant terms containing group elements).  Once I did that a couple of tests that check the serialisation format started failing because zero is encoded specially and the old generator was never producing it (if it had we would essentially have a has collision).  We should probably move these generators somewhere else and use them in the[ `ArbitraryBuiltin` instances](https://github.com/IntersectMBO/plutus/blob/8d6bb1345acdd5594559b31ff5e559c7a856da57/plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs#L250) too.

PR #7344, opened 2025-09-18
Cost models for `LookupCoin`, `ValueContains`, `ValueData`, `UnValueData` builtins

Unisay:
# Costing Value Builtins with Worst-Case Benchmarking

## Overview

This PR implements costing for four Plutus Core Value builtins: `LookupCoin`, `ValueContains`, `ValueData`, and `UnValueData`. The implementation uses a **worst-case oriented benchmarking strategy** that ensures conservative cost estimates for adversarial on-chain scenarios.

Values in Plutus Core are implemented as nested Maps: `Map PolicyId (Map TokenName Quantity)`, backed by BST-based Data.Map. The benchmarking approach systematically explores BST worst-case behavior through careful test case generation.

### LookupCoin

**Cost Model Type**: `linear_in_z` (linear in sum of logarithms)
- **CPU**: `intercept + slope × (log(outerSize) + log(maxInnerSize))`
- **Memory**: constant (1 word)

**Size Measure**: `ValueLogOuterSizeAddLogMaxInnerSize`
- Computes `log₂(numPolicies) + log₂(maxTokensPerPolicy)`
- Reflects O(log m + log k) BST lookup cost through nested maps
- Based on experimental evidence showing lookup time scales with sum of depths, not their maximum

**Rationale**: Looking up a coin requires traversing the outer BST to find the policy, then traversing the largest inner BST to find the token. The sum of logarithms accurately models the total comparison cost.

### ValueContains

**Cost Model Type**: `multiplied_sizes` (product of dimensions)
- **CPU**: `intercept + slope × container_log_size × contained_total_entries`
- **Memory**: constant (1 word)

**Size Measures**:
- Container: `ValueLogOuterSizeAddLogMaxInnerSize` (same as LookupCoin)
- Contained: `ValueTotalSize` (total number of entries)

**Rationale**: ValueContains performs one LookupCoin operation per entry in the contained Value. The cost is the product of:
1. Per-lookup cost (proportional to container BST depth: log m + log k)
2. Number of lookups (contained Value size: n₂)

Result: O(n₂ × (log m₁ + log k₁)) complexity.

**Implementation Note**: Uses `Map.isSubmapOfBy` with optimized short-circuiting, providing 2-4x speedup over naive iteration.

### ValueData

**Cost Model Type**: `constant_cost`
- **CPU**: constant (194,713 steps)
- **Memory**: constant (1 word)

**Size Measure**: Raw Value (no wrapper needed)

**Rationale**: Wrapping a Value as Plutus Data is a constant-time pointer operation. The Data structure already exists in memory; valueData just changes the type tag. Benchmarks confirm minimal variance across Value sizes.

### UnValueData

**Cost Model Type**: `linear_in_x` (linear in Data size)
- **CPU**: `intercept + slope × data_size`
- **Memory**: constant (1 word)

**Size Measure**: Standard Data size (built-in)

**Rationale**: Deserializing Data to Value requires traversing the Data structure and validating the nested map structure. Cost scales linearly with Data size. The slope (43,200 steps per Data node) reflects validation overhead.

### ValueLogOuterSizeAddLogMaxInnerSize

```haskell
instance ExMemoryUsage ValueLogOuterSizeAddLogMaxInnerSize where
    memoryUsage (ValueLogOuterSizeAddLogMaxInnerSize v) =
      let outerSize = Map.size (Value.unpack v)          -- number of policies
          innerSize = Value.maxInnerSize v                -- max tokens in any policy
          logOuter = if outerSize > 0 then integerLog2 outerSize + 1 else 0
          logInner = if innerSize > 0 then integerLog2 innerSize + 1 else 0
      in singletonRose $ fromIntegral (logOuter + logInner)
```

**Purpose**: Models worst-case BST traversal depth through nested maps.

**Key Insight**: For a Value with m policies where the largest policy has k tokens, worst-case lookup requires:
- Traversing outer BST of depth ~log₂(m)
- Traversing largest inner BST of depth ~log₂(k)
- Total comparisons: proportional to log m + log k

**Why sum, not max?**: Experimental benchmarks showed lookup time scales linearly with the sum of depths. Both traversals must complete; they're not alternatives.

### ValueTotalSize

```haskell
instance ExMemoryUsage ValueTotalSize where
    memoryUsage = singletonRose . fromIntegral . Value.totalSize . unValueTotalSize
```

**Purpose**: Counts total number of (policyId, tokenName, quantity) entries across all policies.

**Usage**: Measures iteration count for operations like ValueContains that must check every entry in the contained Value.

## Worst-Case Benchmarking Strategy

The benchmarking methodology prioritizes **conservative cost estimates** through systematic worst-case generation.

### 1. Worst-Case BST Keys

**Problem**: Random ByteString keys typically differ in the first 1-2 bytes, making BST comparisons artificially cheap (short-circuit after 1-2 byte comparisons).

**Solution**: Generate keys with a common prefix:
```haskell
generateKey = do
  let prefix = BS.replicate 28 0xFF        -- 28 bytes of 0xFF
  suffix <- BS.pack <$> replicateM 4 (uniformRM (0, 255))
  return (prefix <> suffix)                 -- 32-byte key
```

**Result**: Forces full 32-byte comparisons during BST traversal, reflecting adversarial scenarios where an attacker crafts keys to maximize comparison cost.

### 2. Power-of-2 Size Grid

**Approach**: Test all combinations of sizes from the sequence:
```
2, 3, 4, 6, 8, 11, 16, 23, 32, 45, 64, 91, 128, 181, 256, 362, 512, 724, 1024, 1448
```

This sequence includes both powers of 2 (2ⁿ) and geometric means (2^(n+0.5) ≈ 2ⁿ × √2).

**Coverage**:
- **LookupCoin**: 20 × 20 = 400 test points spanning BST depths 2 to 21
- **ValueContains**: 10 × 10 = 100 container configurations, each tested with 10 contained sizes

**Rationale**: Power-of-2 sizing systematically explores different BST depths. The half-powers provide finer granularity between powers, ensuring no "gaps" in depth coverage.

### 3. Maximum Depth Targeting

For each Value generated, we track the **deepest entry** (rightmost in both outer and inner BSTs):

```haskell
generateConstrainedValueWithMaxPolicy numPolicies tokensPerPolicy = do
  policyIds <- replicateM numPolicies (generateKey g)
  tokenNames <- replicateM tokensPerPolicy (generateKey g)

  let sortedPolicyIds = sort policyIds
      sortedTokenNames = sort tokenNames
      maxPolicyId = last sortedPolicyIds      -- deepest in outer BST
      deepestToken = last sortedTokenNames    -- deepest in inner BST

  -- Structure: max policy gets ALL tokens, others get minimal (1 token each)
  pure (value, maxPolicyId, deepestToken)
```

**Key Optimization**: Only the max policy receives all tokens. Other policies get a single token each. This minimizes "off-path" costs while maximizing depth at the target lookup location.

**Lookup Keys**: Benchmarks always query `(maxPolicyId, deepestToken)`, forcing maximum BST traversal depth.

### LookupCoin: Exhaustive Depth Coverage

```haskell
lookupCoinArgs =
  [ (maxPolicyId, deepestToken, value)
  | numPolicies <- [2, 3, 4, ..., 1024, 1448]      -- 20 sizes
  , tokensPerPolicy <- [2, 3, 4, ..., 1024, 1448]   -- 20 sizes
  ]
```

**Result**: 400 test points systematically covering all depth combinations from (2,2) to (21,21).

**Lookup Strategy**: Every benchmark queries the deepest possible entry in the Value's BST structure.

### ValueContains: Subset with Worst-Case Entry

```haskell
valueContainsArgs =
  [ (container, contained)
  | numPolicies <- [2, 4, 8, ..., 512, 1024]       -- 10 sizes
  , tokensPerPolicy <- [2, 4, 8, ..., 512, 1024]   -- 10 sizes
  , containedSize <- [step, 2*step, ..., min(1000, totalEntries)]  -- 10 sizes
  ]
```

**Contained Value Construction**:
1. Generate container with worst-case BST structure
2. Extract all entries as flat list
3. Select `containedSize - 1` arbitrary entries
4. **Append the deepest entry last**

**Critical Detail**: Placing the deepest entry last ensures:
- All lookups succeed (no early exit from subset check)
- Maximum BST depth is traversed for at least one lookup
- Tests realistic "contained ⊆ container" relationships

**Result**: ~1000 systematic test cases exploring both container depth and iteration count dimensions.

### ValueData & UnValueData: Random Distribution

```haskell
generateTestValues = empty : replicateM 100 (randomValue 1 to 100_000 entries)
```

**Strategy**: Random sampling with uniform distribution across:
- Number of policies (1 to numEntries)
- Tokens per policy (distributed to reach numEntries total)
- Entry counts from 1 to 100,000

**Maximum Size**: 100,000 entries (up from original 416), reflecting **execution budget constraints** rather than ledger storage limits.

**Rationale**: Scripts can programmatically generate Values much larger than on-chain storage allows. The 100K limit represents what's achievable within maximum CPU execution budget (~10-15 billion picoseconds) while leaving room for actual script logic.

**Constant vs Linear Models**: ValueData shows constant cost (pointer wrapping), while UnValueData shows linear cost (structural validation), confirmed by benchmarks across this wide size range.

## Performance Impact

The `valueContains` implementation received a 2-4x speedup optimization:

**Before**: Manual iteration with early exit
```haskell
valueContains v1 v2 = all (\(p,t,q) -> lookupCoin p t v1 >= q) (toList v2)
```

**After**: Native Map.isSubmapOfBy
```haskell
valueContains v1 v2 = Map.isSubmapOfBy (Map.isSubmapOfBy (<=)) (unpack v2) (unpack v1)
```

**Benefit**: Leverages optimized Map internals with better short-circuiting and comparison batching. Cost model updated to reflect improved performance.

## Testing & Validation

- **Conformance tests**: Updated budget expectations across 100+ test cases
- **Ledger API tests**: Verified backward compatibility with existing script validation
- **Benchmark data**: 400+ data points per builtin ensuring robust cost model fitting
- **Cost model validation**: R² > 0.95 for all fitted models

## Visualization

Interactive cost model visualizations available at:
**https://plutus.cardano.intersectmbo.org/cost-models/**

To preview this PR's cost models, configure the data source to load from this branch:
1. Open the visualization page for the function (e.g., `/cost-models/valuecontains/`)
2. Update the data source URLs to point to this branch's raw files:
   - Benchmark data: `https://raw.githubusercontent.com/IntersectMBO/plutus/yura/costing-builtin-value/plutus-core/cost-model/data/benching-conway.csv`
   - Cost model: `https://raw.githubusercontent.com/IntersectMBO/plutus/yura/costing-builtin-value/plutus-core/cost-model/data/builtinCostModelC.json`
3. The visualization will render this PR's updated cost model parameters

Available visualizations: `lookupCoin`, `valueContains`, `valueData`, `unValueData`

## Summary

This PR establishes production-ready costing for Value builtins through:

1. **Accurate cost models** based on algorithmic complexity (BST depth, iteration count)
2. **Worst-case oriented benchmarking** ensuring conservative estimates for adversarial scenarios
3. **Systematic test coverage** across realistic and extreme Value sizes
4. **Performance optimization** (valueContains 2-4x speedup) reflected in updated costs

The worst-case focus—common-prefix keys, maximum-depth lookups, systematic size coverage—provides strong safety guarantees for on-chain execution budgeting.


plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -391,6 +392,29 @@ instance ExMemoryUsage ValueOuterOrMaxInner where
...
+-}
+newtype Logarithmic n = Logarithmic { unLogarithmic :: n }
+
+instance ExMemoryUsage n => ExMemoryUsage (Logarithmic n) where
+    memoryUsage (Logarithmic wrapped) =
+      case memoryUsage wrapped of
+        CostRose size _ ->
+          let sizeInteger :: Integer
+              sizeInteger = fromSatInt size
+              logSize = I# (integerLog2# sizeInteger)
```
kwxm: I see that there's a function `integerLog2 :: Integer -> Word` in GHC.Num that I don't remember seeing before.  Maybe it didn't exist when we started doing this.  I don't know if it'd be any better than doing this.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -391,6 +392,29 @@ instance ExMemoryUsage ValueOuterOrMaxInner where
...
+For example, @Logarithmic ValueOuterOrMaxInner@ can be used to model operations
+that are O(log max(m, k)) where m is the number of policies and k is the max tokens
+per policy.
+
+The memory usage is calculated as: @max 1 (floor (log2 size + 1))@ where size comes
+from the wrapped newtype's ExMemoryUsage instance.
+-}
+newtype Logarithmic n = Logarithmic { unLogarithmic :: n }
+
+instance ExMemoryUsage n => ExMemoryUsage (Logarithmic n) where
```
kwxm: You could also build the logarithm into the other wrappers if we never need to look at the un-logarithmic sizes.  I'm not sure if that's the case though, and this is probably clearer.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -391,6 +392,29 @@ instance ExMemoryUsage ValueOuterOrMaxInner where
...
+depends on log(n) where n is the size measure from the wrapped newtype.
+
+For example, @Logarithmic ValueOuterOrMaxInner@ can be used to model operations
+that are O(log max(m, k)) where m is the number of policies and k is the max tokens
+per policy.
+
+The memory usage is calculated as: @max 1 (floor (log2 size + 1))@ where size comes
+from the wrapped newtype's ExMemoryUsage instance.
+-}
+newtype Logarithmic n = Logarithmic { unLogarithmic :: n }
```
kwxm: Note that the denotations of the bultins in [Builtins.hs](https://github.com/IntersectMBO/plutus/blob/master/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs) need to use exactly the same size measures, and hence the same wrappers, as the budgeting benchmarks. See `replicateByte` and `shiftByteString` for examples.  Also we'll probably have to re-run the benchmarks once the final form of the costing functions is decided, to make sure that the cost of running the costing functions is included in the cost of running the builtin.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
Unisay: Updated to use new GHC.Num API

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
Unisay: I can see it both ways; this way it seems that logarithmic behavior is more apparent at usage sites.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
zliu41: Is this needed? I don't see any usage.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -0,0 +1,297 @@
...
+generateKey :: (StatefulGen g m) => g -> m K
+generateKey g = do
+  bs <- uniformByteStringM Value.maxKeyLen g
+  case Value.k bs of
+    Just key -> pure key
+    Nothing  -> error "Internal error: maxKeyLen key should always be valid"
+
+-- | Generate random key as ByteString (for lookup arguments)
+generateKeyBS :: (StatefulGen g m) => g -> m ByteString
+generateKeyBS = uniformByteStringM Value.maxKeyLen
```
zliu41: If the keys are completely random, then `lookupCoin` will probably never hit an existing entry, right?

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
Unisay: I've moved the logarithmic measurement from a dedicated `Logarithmic` newtype to the `LogValueOuterOrMaxInner`

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
kwxm: > lookupCoin will probably never hit an existing entry,

Maybe that's what we want?  Do we  know if finding out that something's not in the map is the worst case?  Naively you might think that the time taken to discover that some key is not in the map is always greater or equal to the time taken to find a key that is in the map.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -176,5 +176,10 @@ builtinMemoryModels = BuiltinCostModelBase
   , paramLengthOfArray                   = Id $ ModelOneArgumentConstantCost 10
   , paramListToArray                     = Id $ ModelOneArgumentLinearInX $ OneVariableLinearFunction 7 1
   , paramIndexArray                      = Id $ ModelTwoArgumentsConstantCost 32
+  -- Builtin values
+  , paramLookupCoin                      = Id $ ModelThreeArgumentsConstantCost 1
```
kwxm: This should probably be something bigger than 1.  The memory costs aren't terrribly meaningful, but  I thought we were using 32 for things that just return a pointer.  Looking at the other functions, we don't seem to be very consistent about that though.

plutus-core/cost-model/data/benching-conway.csv
```diff
@@ -12094,203 +12094,710 @@ IndexArray/42/1,1.075506579052359e-6,1.0748433439930302e-6,1.0762684407023462e-6
 IndexArray/46/1,1.0697135554442532e-6,1.0690902192698813e-6,1.0704133377013816e-6,2.2124820728450233e-9,1.8581237858977844e-9,2.6526943923047553e-9
 IndexArray/98/1,1.0700747499373992e-6,1.0693842628239684e-6,1.070727062396803e-6,2.2506114869928674e-9,1.9376849028666025e-9,2.7564941558204088e-9
 IndexArray/82/1,1.0755056682976695e-6,1.0750405368241111e-6,1.076102212770973e-6,1.8355219893844098e-9,1.5161640335164335e-9,2.4443625958006994e-9
-Bls12_381_G1_multiScalarMul/1/1,8.232134704712041e-5,8.228195390475752e-5,8.23582682466318e-5,1.224261187989977e-7,9.011720721178711e-8,1.843107342917502e-7
```
kwxm, line 12097: GitHub seeems to think that the data for all of the BLS functions has changed, but I don't think they have.

plutus-core/cost-model/data/builtinCostModelA.json
```diff
@@ -1,1209 +1,1258 @@
...
-                "intercept": 116711,
-                "slope": 1957
-            },
-            "type": "linear_in_x"
-        },
-        "memory": {
-            "arguments": 4,
-            "type": "constant_cost"
-        }
+  "addInteger": {
```
kwxm, line 2: The indentation seems to have changed from three spaces to two, which makes it hard to compare the files.  Any idea why?

kwxm: Yeah, the new functions have been added at the end, which suggests it's not the case that the whole thing has been regenerated.  Maybe an editor's re-indented everything.

plutus-core/cost-model/data/models.R
```diff
@@ -802,13 +806,26 @@ modelFun <- function(path) {
...
     indexArrayModel           <- constantModel ("IndexArray")
 
+    ## Values
+
+    # Z wrapped with `Logarithmic . ValueOuterOrMaxInner`
+    lookupCoinModel           <- linearInZ ("LookupCoin")    
+
+    # X wrapped with `Logarithmic . ValueOuterOrMaxInner`
+    # Y wrapped with `ValueTotalSize`
+    valueContainsModel        <- linearInY("ValueContains")
```
kwxm:
```suggestion
    valueContainsModel        <- linearInY ("ValueContains")         
```

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -391,6 +393,48 @@ instance ExMemoryUsage ValueOuterOrMaxInner where
...
+
+instance ExMemoryUsage n => ExMemoryUsage (Logarithmic n) where
+    memoryUsage (Logarithmic wrapped) =
+      case memoryUsage wrapped of
+        CostRose size _ ->
+          let sizeInteger :: Integer
+              sizeInteger = fromSatInt size
+              logSize = integerLog2 sizeInteger
```
kwxm: If you're going to do this here then you might as well do the same thing for `memoryUsageInteger` and then you can probably get rid of `MagicHash` and the imports at the top.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -391,6 +393,48 @@ instance ExMemoryUsage ValueOuterOrMaxInner where
...
+O(log max(m, k)) where m is the number of policies and k is the max tokens per policy.
+
+If this is used to wrap an argument in the denotation of a builtin then it *MUST* also
+be used to wrap the same argument in the relevant budgeting benchmark.
+-}
+newtype LogValueOuterOrMaxInner = LogValueOuterOrMaxInner { unLogValueOuterOrMaxInner :: Value }
+
+instance ExMemoryUsage LogValueOuterOrMaxInner where
+    memoryUsage (LogValueOuterOrMaxInner v) =
+      let size = Map.size (Value.unpack v) `max` Value.maxInnerSize v
```
kwxm: I'm still a little suspicious about this.  I'm wondering how the behaviour of the function changes when the maximum is the LHS instead of the RHS.   In theory it might make more sense to model the time by something of the form `a + bm + cn` where m is the size of the outer map and n is the maximum size of the inner maps, but that would be very difficult to do with the current costing framework.  The logarithms are never going to be very big anyway (you'd need a map with over 1,000,000 entries  to even get as far as 20), so we don't need to be ultra-precise.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
Unisay: Updated: 10 for Integer (lookup coin returns integer quantity). 32 for `BuiltinResult Bool` and `Data` and `Value`. I thought about re-using `boolMemModel` but rejected this idea because `BuiltinResult ...`.

plutus-core/cost-model/data/builtinCostModelA.json
Unisay: Yes, this is my editor. I will restore the original indentation (4 spaces)

plutus-core/cost-model/data/benching-conway.csv
Unisay: The file on master contains Windows-style line terminators (`\r\n`) for BLS lines:

```sh
git show master:plutus-core/cost-model/data/benching-conway.csv | grep "Bls12_381_G1_multiScalarMul/1/1" | od -c | grep -C1 "\r"
0000000   B   l   s   1   2   _   3   8   1   _   G   1   _   m   u   l
0000020   t   i   S   c   a   l   a   r   M   u   l   /   1   /   1   ,
0000040   8   .   2   3   2   1   3   4   7   0   4   7   1   2   0   4
--
0000200   8   7   1   1   e   -   8   ,   1   .   8   4   3   1   0   7
0000220   3   4   2   9   1   7   5   0   2   e   -   7  \r  \n
```

This PR changes `\r\n` to `\n` .

plutus-core/cost-model/data/builtinCostModelA.json
```diff
@@ -1050,26 +1050,26 @@
     "complementByteString": {
         "cpu": {
             "arguments": {
-               "intercept": 107878,
+                "intercept": 107878,
```
Unisay: This was a 3-space indent, inconsistent with 4-space used in all other lines.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -0,0 +1,186 @@
...
+    (lookupCoinArgs gen) -- the argument combos to generate benchmarks for
+
+lookupCoinArgs :: StdGen -> [(ByteString, ByteString, Value)]
+lookupCoinArgs gen = runStateGen_ gen \(g :: g) -> do
+  -- Add search keys to common test values
+  let testValues = generateTestValues gen
+  commonWithKeys <- mapM (withSearchKeys g . pure) testValues
+
+  -- Additional tests specific to lookupCoin
+  let valueSizes = [(100, 10), (500, 20), (1_000, 50), (2_000, 100)]
```
ana-pantilie: What is the reasoning behind picking these specific sizes? 

Most importantly, however, you're not actually generating `Value`s of these sizes, because you're not checking whether the keys you generate are unique per map or not.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
ana-pantilie: I don't agree. I think we should actively include both the case when the map contains the key and when it doesn't. Otherwise we're not really measuring this case, and that's the whole point of benchmarking, right? Otherwise we would just use the, analytically discovered, worst-time complexity of the algorithm and pick a function from that category for its cost, right?

ana-pantilie: Also, as I mentioned above, you won't have a good idea of the actual size of the `Value` if you don't enforce uniqueness of the keys.

kwxm:
> I think it's important we both generate random inputs and inputs which hit the various edge-cases the algorithm has. The benchmarking data we generate should be a sample which describes the algorithm's behavior as completely as possible.

In general we want the costing function to describe the worst-case behaviour of the builtin, which means that the benchmarks should be run with inputs which produce worst-case behaviour.  If we feed them totally random inputs then we'll fit a costing function which describes the average-case behaviour, and this could be dangerous.  For example if a function takes 1ms on average but there are particular inputs which cause it to take 20ms then the cost model has to charge 20ms for all inputs so that we don't undercharge for scripts that _do_ actually exercise the worst case (perhaps repeatedly) [An example of this kind of behaviour would be for `equalsByteString`, where we only benchmark the case where the two inputs are equal: if they're not equal (in particular if they have different lengths) then the function can return very quickly, but it they are equal the builtin will have to examine every byte of the input].  For some builtins the worst case won't be significantly worse than the average case, and then we miight be able to get away with random inputs, especially if it's hard to generate worst case inputs.

In the case of `lookupCoin` I'm not sure what the worst case is.  It may well be when the thing you're looking up isn't present in the map, but we should check whether this is actually true.

Addendum.  It's often necessary to do a lot of exploratory benchmarking to check that the builtin behaves as you're expecting, and this isn't generally reflected in the final costing benchmarks and costing function (for example, see the costing branches for `expModInteger` [here](https://github.com/IntersectMBO/plutus/pull/7071) and [here](https://github.com/IntersectMBO/plutus/pull/7080), where I ran dozens of different benchmarks with different inputs before settling on the final version). So we're not just choosing a costing function and b,lindly fitting branhcmark results to it, we're checking that our initial assumptions are correct and modifying them if necessary.  It might be useful to do this with the lookup/insertion/union costs as well.

zliu41:
> In the case of lookupCoin I'm not sure what the worst case is. It may well be when the thing you're looking up isn't present in the map, but we should check whether this is actually true.

Like what I said earlier, for `lookupCoin` the worst case would be: the currency hits the largest inner map, and the token does not exist in the inner map. Because this involves lookups in both the outer map and inner map. @Unisay Let's make sure that the case where the currency hits the largest inner map is well represented in the data.

kwxm:
> > In the case of lookupCoin I'm not sure what the worst case is. It may well be when the thing you're looking up isn't present in the map, but we should check whether this is actually true.
> 
> Like what I said earlier, for `lookupCoin` the worst case would be: the currency hits the largest inner map, and the token does not exist in the inner map. Because this involves lookups in both the outer map and inner map. @Unisay Let's make sure that the case where the currency hits the largest inner map is well represented in the data.

What about the case when the outer map is really big compared to the inner ones?  Also, we could make all of the inner maps the same size.  We probably don't want to be benchmarking anything that isn't the worst case.

zliu41:
> What about the case when the outer map is really big compared to the inner ones?

Isn't the worst case in that case still performing lookup in both the outer map and inner map? If you are concerned that the outer map key is not at the leaf, you can use either the smallest key or the largest key.

By the way, to make key comparisons more expensive, it would also be useful to fix first 30 or 31 bytes, and only vary the last byte.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -0,0 +1,243 @@
...
+
+-- | Generate Value with random budget between 1 and 30,000 bytes
+generateValue :: (StatefulGen g m) => g -> m Value
+generateValue g = do
+  maxEntries <- uniformRM (1, maxValueEntries maxValueInBytes) g
+  generateValueMaxEntries maxEntries g
+ where
+  -- Maximum budget for Value generation (30,000 bytes)
+  maxValueInBytes :: Int
+  maxValueInBytes = 30_000
```
zliu41: Why is 30000 bytes the max? A Value can be created programmatically, in which case it is not limited by bytes, but limited by execution units.

This limit seems way to small. We should benchmark Values with minimum 10k entries if not 100k or more.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -0,0 +1,243 @@
...
+
+{-| Helper: Create a worst-case ByteString key from an integer
+The key has maxKeyLen-4 bytes of 0xFF prefix, followed by 4 bytes encoding the integer
+-}
+mkWorstCaseKey :: Int -> ByteString
+mkWorstCaseKey n =
+  let prefixLen = Value.maxKeyLen - 4
+      prefix = BS.replicate prefixLen (0xFF :: Word8)
+      -- Encode the integer in big-endian format (last 4 bytes)
+      b0 = fromIntegral $ (n `shiftR` 24) .&. 0xFF
```
zliu41: you can simply generate 4 random bytes, instead of generating an integer and doing these bitwise operations.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -0,0 +1,243 @@
...
+
+lookupCoinBenchmark :: StdGen -> Benchmark
+lookupCoinBenchmark gen =
+  createThreeTermBuiltinBenchElementwiseWithWrappers
+    (id, id, ValueLogOuterOrMaxInner) -- Wrap Value argument to report outer/max inner size with log
+    LookupCoin -- the builtin fun
+    [] -- no type arguments needed (monomorphic builtin)
+    (lookupCoinArgs gen) -- the argument combos to generate benchmarks for
+
+lookupCoinArgs :: StdGen -> [(ByteString, ByteString, Value)]
```
kwxm, line 51: Maps are [implemented as balanced binary trees](https://hackage-content.haskell.org/package/containers-0.8/docs/src/Data.Map.Internal.html#Map) with (key, value) pairs at the internal nodes, so if you've got a node containing key `k` then the left subtree of the node will only contain keys less than `k` and the right subtree keys greater than `k`.  The functions that operate on maps are supposed to keepthe tree balanced, so the left subtree should be  (approximately) the same size as the right one.  When you've got `2^n-1` nodes the tree should be prefectly balanced and every path from the root to a leaf should have length `n` (ie, you should pass through `n` nodes as you travel from the root to an entry both of whose subtrees are empty (`Tip`)).

I think that to get the worst case behaviour for `lookupCoin` you can generate an outer map with `2^a - 1` unique keys for `a` in some range like 1..10 or 1..15, so you get a full tree. Looking for the entry with the largest key should then require searching all the way to the bottom of the tree (always branching to the right), which should be the worst case (and also make sure that all of the keys have a long common prefix to maximise the comparison time).  The inner map for this longest case should also be a full binary tree, with `2^b - 1` entries for some `b`; we want the worst case to search the inner map as well, and in this case that should happen when you look for a key that's bigger than all of the keys in the inner map, since again you'll have to search all the way down the right hand side of the tree; I think that if you search for the biggest key it'll take pretty much the same time though. The total time taken should be proportional to `a+b`, since you'll have so examine `a` nodes in the outer map and `b` in the inner map, and since the keys are of the same type for inner and outer maps the time taken per node should be about the same for both (if the key types in the inner and outer maps were different then you might be looking at something of the form `ra+sb`, but here `r` and `s` should be the same, so you've got `r(a+b)`).

Note that `a` and `b` are the _depths_ of the trees, which wil be `integerLog2` of the number of entries, so I think here you want to use a size measure which is `integerLog2(outer size) + integerLog2(maximum inner size)`.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -0,0 +1,243 @@
...
+lookupCoinArgs gen = runStateGen_ gen \(g :: g) -> do
+  -- Add search keys to common test values
+  let testValues = generateTestValues gen
+  commonWithKeys <- mapM (withSearchKeys g . pure) testValues
+
+  -- Additional tests specific to lookupCoin
+  let valueSizes = [(100, 10), (500, 20), (1_000, 50), (2_000, 100)]
+  additionalTests <-
+    sequence $
+      -- Value size tests (number of policies × tokens per policy)
```
kwxm: Why "policy"?  I though that meant something completely different.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
kwxm: I guess you'll need to benchmark this over some set of pairs of depths `(a,b)` where `a` and `b` both vary over [1..15] or something, but that might take a long time since for that range you'd be running 225 benchmarks.  I think that in fact the time will only depend on `a+b`, but initially we should check that that's true by looking at different values of `a`  and `b` with the same sum.

zliu41: It might be an overkill to bother generating full trees. In any balanced binary tree the depths of any two leaves differ by at most 1, so as long as we make sure that we hit a leaf node (using either the smallest key or the largest key), 🤷 

It's more important to make sure the outer key hits the largest inner map. I _think_ the worst case is: there's a large inner map with N/2 keys, together with N/2 singleton inner maps, and the outer key hits that large inner map. Whether the outer map and the inner map are full trees shouldn't matter that much.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
zliu41: Policy is the same thing as currency symbol.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
zliu41: So in summary, I would do this:
- Given total size `N`, let there be roughly `N/2` inner maps: one big inner map whose size is roughly `N/2`, and the rest are singletons.
- Outer key should hit the big inner map.
- Both the outer key and the inner key should hit leaf nodes. So both keys should be either the min or the max key in the respective map (or the inner key may be absent in the inner map).

This should be very close to the worst case, if not _the_ worst case. I wouldn't bother with varying `a` and `b`, or generating full trees, or anything like that.

kwxm: >  I wouldn't bother with varying a and b

I think it's worth doing that at least once just to make sure that the time taken depends only on the sum of `a` and `b` .  If we can show that it does then we can just restrict the benchmarking to the case when `a = 1`, so you only have one entry in the outer map; alternatively, you can take `b = 1` and just worry about the size of the outer map.  I think we can effectively regard the entire map as one big tree: when you get to the tip of the outer map you move into one of the inner maps and continue searching there, so it's like the inner maps are glued onto the leaves of the outer map.  Then all that matters is the total depth, `a+b`.  I'd like to check that assumption before going any further though: it's better to have some evidence than just to guess what's going on.

zliu41: > If we can show that it does then we can just restrict the benchmarking to the case when a = 1, so you only have one entry in the outer map; alternatively, you can take b = 1 and just worry about the size of the outer map.

You can't. `lookupCoin` is `O(log max(m, k))`. So it's better to balance the size of the outer map and the size of the largest inner map.

zliu41: In other words, given a size measure of 10, you can:
- Make the outer map have 1024 entries, and the largest inner map have 1024 entries
- Or make the outer map have 1024 entries, and the largest inner map have 1 entry
- Or make the outer map have 1 entry, and the largest inner map have 1024 entries

Obviously the first is the worst case

kwxm: > You can't. lookupCoin is O(log max(m, k))

No, I don't think it is.  I think it's `O(log a + log b)`, or at least that `log a + log b` is a size measure that will give a more precise result than `log(max(a, b)` , and that's  what my proposed experiment is trying to confirm.  The O's are obscuring what's actually going on since they're hiding the details of the constants.    For the sum the actual costing function will be of the form `r + s(log a+ log b)` and for the maximum it'll be of the form `u + v*log(max(a, b))`.  Now `log a+log b <= 2*max(log a, log b)`, with equality when `a=b`, so `r+s*(log a + log b) <= r + 2*s*max(log a, log b)`, and when you put an `O` round them they become the same, but the right hand one can actually be almost twice the left hand one.

> In other words, given a size measure of 10, you can:

>    Make the outer map have 1024 entries, and the largest inner map have 1024 entries
>    Or make the outer map have 1024 entries, and the largest inner map have 1 entry
>    Or make the outer map have 1 entry, and the largest inner map have 1024 entries

> Obviously the first is the worst case

With the sum size measure the first map has a size of 20 and the other ones have a size of 11, but with the maximum they all have size 10, so the sum version picks out the worst case and the maximum doesn't. A stragiht line fitted to the benchmarking results using the sum measure should give us a more precise bound for the execution time than if we use the maximum measure.  The sum is a more accurate measure of the total depth of a value than the maximum, and I think that the total depth is what we need to worry about.

Unisay: I made benchmarking and here is what it shows (LLM summary):
```
Two-Level Map Lookup Performance: Experimental Analysis

  Experimental Setup

  Data structure: Map ByteString (Map ByteString Natural)

  Key characteristics:
  - 32-byte ByteStrings with 28-byte common prefix (simulating hash-like keys)
  - Perfect binary tree structure (keys = 2^n - 1)
  - Worst-case lookup: deepest key in both outer and inner maps

  Test parameters:
  - Outer map depth: a (containing 2^a - 1 keys)
  - Inner map depth: b (containing 2^b - 1 keys)
  - Total depth: N = a + b

  ---
  Experiment 1: Distribution Impact (Constant Total Depth)

  Hypothesis: When N is held constant, how does the distribution of depth between outer and inner maps
  affect performance?

  Test configuration: N = 17, varying distributions from (1,16) to (16,1)

  Results: Distribution Has Minimal Impact

  | Distribution   | Outer Keys | Inner Keys | Lookup Time | Deviation from Mean |
  |----------------|------------|------------|-------------|---------------------|
  | (10,7)         | 1,023      | 127        | 1.657 μs    | +1.5%               |
  | (6,11)         | 63         | 2,047      | 1.655 μs    | +1.4%               |
  | (8,9) balanced | 255        | 511        | 1.644 μs    | +0.7%               |
  | (1,16) extreme | 1          | 65,535     | 1.607 μs    | -1.6%               |
  | (16,1) extreme | 65,535     | 1          | 1.598 μs    | -2.1%               |

  Range: 1.598 μs to 1.657 μs (3.6% variation)

  Key finding: Distribution choice has negligible impact on performance when total depth is constant.

  ---
  Experiment 2: Linear Scaling with Total Depth

  Hypothesis: Lookup time scales linearly with total depth N = a + b, regardless of distribution.

  Test configuration: N ∈ {10, 12, 14, 16, 18, 20}, three representative distributions per N

  Results: Strong Linear Relationship

  | N (Total Depth) | Representative Samples  | Min Time | Max Time | Average Time | Cost per Level |
  |-----------------|-------------------------|----------|----------|--------------|----------------|
  | 10              | (1,9), (5,5), (9,1)     | 125.6 ns | 128.3 ns | 127.0 ns     | —              |
  | 12              | (1,11), (6,6), (11,1)   | 132.9 ns | 142.9 ns | 136.9 ns     | +4.95 ns       |
  | 14              | (1,13), (7,7), (13,1)   | 140.1 ns | 154.0 ns | 144.5 ns     | +3.80 ns       |
  | 16              | (1,15), (8,8), (15,1)   | 145.3 ns | 160.4 ns | 151.1 ns     | +3.30 ns       |
  | 18              | (1,17), (9,9), (17,1)   | 152.7 ns | 167.6 ns | 158.6 ns     | +3.75 ns       |
  | 20              | (1,19), (10,10), (19,1) | 158.8 ns | 173.9 ns | 168.8 ns     | +5.10 ns       |

  Average cost per level: 4.18 ns

  Linear model: Time ≈ 80 ns + 4.2 ns × N (R² ≈ 0.99)

  ---
  Scaling Analysis

  | Test                   | Expected (if linear) | Observed    | Result     |
  |------------------------|----------------------|-------------|------------|
  | N=10 → N=20 (2x depth) | 2.00x time           | 1.33x time  | Sub-linear |
  | Per-level increment    | Constant             | 4.18 ns avg | ✅ Constant |

  Note: The sub-linear scaling (1.33x instead of 2x) suggests the baseline overhead is significant
  relative to per-level cost at small N values.

  ---
  Key Experimental Tendencies

  1. Distribution Independence (Constant N)
    - Variation between distributions: <4%
    - Expensive key comparisons dominate performance
    - Tree shape/cache effects are negligible
  2. Linear Depth Scaling
    - Each additional tree level: +4.2 ns
    - Baseline overhead: ~80 ns
    - Strong linear correlation (R² ≈ 0.99)
  3. Distribution Variation by Depth

  | Total Depth (N) | Max-Min Spread | % Variation |
  |-----------------|----------------|-------------|
  | 10              | 2.7 ns         | 2.1%        |
  | 12              | 10.0 ns        | 7.3%        |
  | 14              | 13.9 ns        | 9.6%        |
  | 16              | 15.1 ns        | 10.0%       |
  | 18              | 14.9 ns        | 9.4%        |
  | 20              | 15.1 ns        | 8.9%        |

  Pattern: Variation increases slightly at deeper depths but remains <10%

  ---
  Conclusions

  1. For expensive key comparisons (ByteStrings with common prefixes):
    - Lookup time is primarily determined by total depth (a + b)
    - Distribution choice (split between outer/inner) has minimal impact
  2. Performance model:
  Lookup Time ≈ 80 ns + 4.2 ns × (a + b)
  3. Practical implication:
    - When designing nested map structures with expensive keys, optimize for total depth minimization
  rather than specific distribution patterns
    - Tree balancing and cache optimization are secondary concerns
  4. Cost breakdown:
    - ~80 ns: Fixed overhead (function calls, setup)
    - ~4.2 ns per level: Key comparison + tree traversal
```

zliu41: I'm fine with using `log m + log k` instead of `log (max m k)`. I don't think it really matters either way, but let's just make a decision and move forward with getting the costing done.

We need to stay on schedule for HF by EOY, and it is a firm deadline. If both @kwxm and @Unisay prefers `log m + log k` then go ahead with it.

zliu41:
@Unisay I still need a summary of how the main recent discussion points were addressed (or why if not addressed), so reviewers know where to look.

zliu41:
It would also be helpful if you reply to each unresolved comment above to indicate if it has been addressed or why it hasn't.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
Unisay: I've implemented the `log m + log k` size measure as a newtype `ValueLogOuterSizeAddLogMaxInnerSize` and it yields compelling results IMO.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
Unisay: I've implemented your suggestion.

zliu41 (approved):
For `ValueData` and `UnValueData`, you tested "randomValue 1 to 100_000 entries", but for `LookupCoin` and `ValueContains`, why is the max value size only 1448? Or is the description not up to date?

Otherwise LGTM - nice work!

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -176,5 +176,10 @@ builtinMemoryModels = BuiltinCostModelBase
   , paramLengthOfArray                   = Id $ ModelOneArgumentConstantCost 10
   , paramListToArray                     = Id $ ModelOneArgumentLinearInX $ OneVariableLinearFunction 7 1
   , paramIndexArray                      = Id $ ModelTwoArgumentsConstantCost 32
+  -- Builtin values
+  , paramLookupCoin                      = Id $ ModelThreeArgumentsConstantCost 10
+  , paramValueContains                   = Id $ ModelTwoArgumentsConstantCost 32
```
kwxm, line 181:
```suggestion
  , paramValueContains                   = Id $ boolMemModel
```

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -176,5 +176,10 @@ builtinMemoryModels = BuiltinCostModelBase
...
+  , paramValueData                       = Id $ ModelOneArgumentConstantCost 32
```
kwxm, line 182: I think the memory models for `valueData` and `unValueData` need to be much bigger, since they're supposed to represent the total amount of memory used by the returned value.  Experimenting with the results of `generateTestValues` in `Benchmarks.Values`, I got a list of Values with thefollowing memory usages:
```
[0,55539,12118,10211,45715,8631,25078,1706,13340,24360,17529,11374,7681,71229,7345,14258,9161,14034,1339,48068,23206,41314,6950,16799,15401,14397,349,6205,4611,28034,34924,9816,11709,36200,2539,6722,53631,22384,32041,60206,15751,6760,94287,12000,37360,10870,35535,9649,6938,3891,57221,23825,16219,51830,3712,29569,3065,50249,9171,82416,42921,32171,1899,58222,17522,32561,30366,1596,5008,17914,5177,10016,9206,7188,93911,63802,8962,13202,8621,13884,80,43194,8112,54225,1077,1036,45364,31703,1872,24615,48316,9248,40840,8876,344,18905,2591,19916,1295,10229,18246]
```
and converting these into `Data` gave a list of objects with the following memory usages:
```
[4,1388479,302906,255279,1142879,215659,626942,42498,333432,608992,438217,284330,192029,1780729,183629,356454,229017,350854,33443,1201704,580142,1032854,173682,419955,385029,359893,8729,155129,115051,700854,873092,245368,292645,904992,63479,168018,1340779,559592,801017,1505154,393779,168932,2357179,300004,934004,271754,888379,241205,173322,97279,1430529,595629,405455,1295754,92756,739229,76629,1256229,229279,2060404,1073029,804267,47095,1455554,438042,814017,759154,39904,125204,447830,129345,250404,230154,179608,2347779,1595054,224030,330030,215505,347092,1944,1079842,202804,1355629,26149,25268,1134104,792579,46768,615343,1207904,231168,1021004,221844,8292,472605,64647,497868,32379,255693,456130]
```
Zipping with `div`, it looks as if the memory usages of the `Data` objects are generally 24-25 times the memory usages of the corresponding `Value` objects, so on the face of it the memory model for `valueData` should probably multiply by 25 and the memory model for `unValueData` should probably divide by 25 (which we can't currently do).  However this is misleading because the "memory usage" for a `Value` object is the `totalSize`,ie the total number of nodes in inner maps.  We really want the total amount of memory occupied by the value, which will be approximated by something like (`size of outer map) * (size of currency name ) + totalSize * (size of token name + size of quantity)` (although we'll just be generating pointers to the existing names and quantities, not . Unfortunately we have to use the same size measure as the  denotation, so that we can't feed the actualy memory usage to the memory costing function.  If we look at the memory usage function for `Data` then we may be able to work out how it relates to the actual memory usage of the corresponding `Value`, and if we're lucky it may turn out that 25 times the number of nodes in the outer map.  This will need a bit of investigation though.

plutus-core/cost-model/data/builtinCostModelA.json
kwxm, line 2: The indentation seemsto have changed in these files, which makes it tricky to see what the important differences are.  Did they get reformatted by an editor or something?

plutus-core/cost-model/data/models.R
```diff
@@ -802,13 +806,33 @@ modelFun <- function(path) {
...
+    lookupCoinModel           <- linearInZ ("LookupCoin")    
+
+    # X wrapped with `ValueLogOuterSizeAddLogMaxInnerSize` (sum of logarithmic sizes)
+    # Y wrapped with `ValueTotalSize` (contained value size)
+    valueContainsModel        <- {
+        fname <- "ValueContains"
+        filtered <- data %>%
+            filter.and.check.nonempty(fname) %>%
+            discard.overhead ()
+        m <- lm(t ~ I(x_mem * y_mem), filtered)
```
kwxm, line 828: I think this model may be inaccurate since we changed the implementation of `valueContains`.  I'll think about that, but for the time being I think the predictions actually look pretty close to the benchmark results, so it should be safe to merge this so that we can move on, but come back to it later.

plutus-core/cost-model/data/models.R
```diff
@@ -802,13 +806,33 @@ modelFun <- function(path) {
...
+        mk.result(m, "multiplied_sizes")
+    }         
+
+    # Sizes of parameters are used as is (unwrapped):
+    valueDataModel            <- constantModel ("ValueData")
```
kwxm, line 833: I'm still a bit mystified about why this  is constant cost, but I think the benchmarks are doing the right thing and the results do in fact seem to be pretty constant.   Maybe we could make in `linearInX` with a zero (or at least small) slope in case we have to  change it later (it's safe to use a linear function to represent a constant one, but difficult to change from constant to linear later).

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -382,14 +378,36 @@ newtype ValueTotalSize = ValueTotalSize { unValueTotalSize :: Value }
...
+scales linearly with the sum of depths (log m + log k), not their maximum.
+
+Used for builtins like lookupCoin where worst-case performance requires
+traversing both the outer map to find the policy AND the largest inner map
+to find the token.
+
+If this is used to wrap an argument in the denotation of a builtin then it *MUST* also
+be used to wrap the same argument in the relevant budgeting benchmark.
+-}
+newtype ValueLogOuterSizeAddLogMaxInnerSize =
```
kwxm, line 400: This name's a bit unwieldy!  Maybe we could call it `MaxTotalDepth` or something since that's what it's measuring (the length of the longest path to a leaf in the outer map plus the length of the longest path from the root of an inner map to a leaf).

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -235,12 +231,12 @@ instance ExMemoryUsage () where
...
 memoryUsageInteger :: Integer -> CostingInteger
--- integerLog2# is unspecified for 0 (but in practice returns -1)
+-- integerLog2 is unspecified for 0 (but in practice returns -1)
 -- ^ This changed with GHC 9.2: it now returns 0.  It's probably safest if we
 -- keep this special case for the time being though.
 memoryUsageInteger 0 = 1
--- Assume 64 Int
-memoryUsageInteger i = fromIntegral $ I# (integerLog2# (abs i) `quotInt#` integerToInt 64) + 1
+-- Assume 64-bit words
+memoryUsageInteger i = fromIntegral (integerLog2 (abs i) `div` 64 + 1)
```
kwxm, line 239: :+1:

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -176,5 +176,10 @@ builtinMemoryModels = BuiltinCostModelBase
   , paramLengthOfArray                   = Id $ ModelOneArgumentConstantCost 10
   , paramListToArray                     = Id $ ModelOneArgumentLinearInX $ OneVariableLinearFunction 7 1
   , paramIndexArray                      = Id $ ModelTwoArgumentsConstantCost 32
+  -- Builtin values
+  , paramLookupCoin                      = Id $ ModelThreeArgumentsConstantCost 10
```
kwxm, line 180: I guess this is OK.  It'll actually return a pointer to an already-allocated quantity in the heap and I think we've used 10 for that elsewhere.  That's probably not totally accurate, but the numbers in here don't bear much of a realationship to reality anyway.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -0,0 +1,343 @@
...
+  {- ValueContains performs multiple LookupCoin operations (one per entry in contained).
+     Worst case: All lookups succeed at maximum depth with many entries to check.
+
+     Strategy:
+     1. Generate container with worst-case BST structure (uniform, power-of-2 sizes)
+     2. Select entries FROM container (maintain subset relationship for no early exit)
+     3. Include deepest entry to force maximum BST traversal
+     4. Test multiple contained sizes to explore iteration count dimension
+
+     Result: ~1000 systematic worst-case benchmarks vs 100 random cases previously
```
kwxm, line 117: I think this is maybe a bit too much.  It takes about 2½ hours to benchmark just this function, which is much  longer than any of the other builtins.  Here's a list of the numbers of datapoints in the CSV file for the most intensively benchmarked builtins, and `valueContains` is much bigger than anything else.  Maybe we could reduce it to 15x15 or something.
```
    202 EqualsString
    225 AddInteger
    256 DivideInteger
    256 ExpModInteger
    256 MultiplyInteger
    300 EqualsByteString
    400 ConstrData
    400 EqualsData
    400 LookupCoin
    400 MkPairData
    400 SerialiseData
    441 AppendByteString
    441 AppendString
    500 ChooseList
    625 AndByteString
   1052 ValueContains
```

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -0,0 +1,343 @@
...
+    -- Random tests for parameter spread (100 combinations)
+    replicateM 100 (generateValue g)
+
+-- | Generate Value with random number of entries between 1 and maxValueEntries
+generateValue :: (StatefulGen g m) => g -> m Value
+generateValue g = do
+  numEntries <- uniformRM (1, maxValueEntries) g
+  generateValueMaxEntries numEntries g
+
+-- | Maximum number of (policyId, tokenName, quantity) entries for Value generation.
```
kwxm, line 231: I think these are a bit big. There's a danger that if you benchmark with very large inputs it'll be inaccurate for smaller (and more realistic) ones. If a function is constant cost then it doesn't matter too much what the input sizes are, and for linear costing functions we can maybe trade a bit of inaccuracy for bigger numbers in favour of accurate costing for smaller ones.  The current CPU costing function for `unValueData` is `1000 + 43200*(total size)`, which grows pretty quickly.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -0,0 +1,343 @@
...
+  createTwoTermBuiltinBenchElementwiseWithWrappers
+    (ValueLogOuterSizeAddLogMaxInnerSize, ValueTotalSize)
+    -- Container: sum of log sizes, Contained: totalSize
+    ValueContains -- the builtin fun
+    [] -- no type arguments needed (monomorphic builtin)
+    (valueContainsArgs gen) -- the argument combos to generate benchmarks for
+
+valueContainsArgs :: StdGen -> [(Value, Value)]
+valueContainsArgs gen = runStateGen_ gen \g -> do
+  {- ValueContains performs multiple LookupCoin operations (one per entry in contained).
```
kwxm, line 108: I think this is no longer accurate: it's not always searching from the root of the containing value.  I'll try to say more about this later.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -0,0 +1,343 @@
...
+lookupCoinBenchmark gen =
+  createThreeTermBuiltinBenchElementwiseWithWrappers
+    (id, id, ValueLogOuterSizeAddLogMaxInnerSize) -- Wrap Value argument to report sum of log sizes
+    LookupCoin -- the builtin fun
+    [] -- no type arguments needed (monomorphic builtin)
+    (lookupCoinArgs gen) -- the argument combos to generate benchmarks for
+
+lookupCoinArgs :: StdGen -> [(ByteString, ByteString, Value)]
+lookupCoinArgs gen = runStateGen_ gen \(g :: g) -> do
+  {- Exhaustive power-of-2 combinations for BST worst-case benchmarking.
```
kwxm, line 53: I don't think this is covering the worst case.  The attached plot shows the raw benchmark results for `lookupCoin` with a regression line fitted.  Above every size there are a number of points that take different times, so I think this is benchmarking the average case.  Ideally we'd just get the top point of each column of points and fit a line through those.  However, it probably doesn't matter too much. The vertical columns look quite big, but in fact the difference from the regression line is only about 3-4%, which seems pretty acceptable.

lookupCoin

kwxm:
Maybe some of the stuff about the benchmarking strategy in the intiial PR comment could go in the file containing the benchmarks so that we can find it when we look at the file in a few years and wonder why the benchmarks are like they are.  I think there's some overlap with the existing comments, but there's stuff that I don't think is covered in the file.

plutus-core/cost-model/data/builtinCostModelA.json
Unisay: This was an unintended change! I was planning to fix indentation in a separate PR...

PR #7347, opened 2025-09-18
Improve Deployments of Docusaurus Site & Papers

zeme-wana:
Only deploy a preview of the docusaurus site when the relevant files have changed.
Also deploy a preview for the papers and specs.

### DIFF
--- a/.github/workflows/papers-and-specs.yml
+++ b/.github/workflows/papers-and-specs.yml
@@ -1,5 +1,8 @@
 # This workflow builds and deploys various papers.
 # 
+# On pull requests, this workflow builds the papers and deploys a temporary preview
+# to gh-pages, but only if the relevant files have changed.
+#
 # This workflow runs on all push to master and can also be triggered manually.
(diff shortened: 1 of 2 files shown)


.github/workflows/papers-and-specs.yml
```diff
@@ -42,8 +49,16 @@ jobs:
...
+        if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
         uses: JamesIves/github-pages-deploy-action@v4.7.3
         with:
           folder: _resources
           target-folder: resources
-          single-commit: true
\ No newline at end of file
+          single-commit: true
+
+      - name: Publish Preview Papers
```
zliu41, line 59: Is this what enables the preview of the spec? I'm not seeing any mention of `plutus-core-spec` here.

zeme-wana: Yes. `plutus-core-spec` is build on line 37

PR #7348, opened 2025-09-18
Use `arbitraryBuiltin @Integer` to generate quantities in `Value`

zliu41:
to get better coverage.


plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -34,19 +33,21 @@ prop_packPreservesInvariants = forAll arbitrary $ \nm ->
...
 
 -- | Verifies that @insertCoin@ preserves @Value@ invariants
 prop_insertCoinPreservesInvariants :: Property
-prop_insertCoinPreservesInvariants = forAll arbitrary $ \(v, amt) ->
+prop_insertCoinPreservesInvariants = forAll arbitrary $ \v ->
   forAll (genShortHex (V.totalSize v)) $ \currency ->
     forAll (genShortHex (V.totalSize v)) $ \token ->
-      let v' = V.insertCoin currency token amt v
-       in checkInvariants v'
+      forAll arbitraryBuiltin $ \amt ->
```
effectfully: Don't use `forAll` unless you absolutely have to, it's another footgun helpfully provided by QuickCheck with no commentary. It doesn't perform any shrinking. Just bind `v` and `amt` directly:

```haskell
prop_insertCoinPreservesInvariants :: Value -> Integer -> Property
prop_insertCoinPreservesInvariants = <...>
```

Alternatively, you can use `forAllShrink`, but IIRC that makes shrinking sequential, i.e. once shrinking the first argument stops failing the test, you start shrinking the second and never come back to try shrinking the first again.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
```diff
@@ -245,7 +248,71 @@ instance Arbitrary Data where
...
+{-| Generate a 'ByteString' by picking one of the predetermined ones, given a number of
+cells to fill (see 'toCellCandidatesNumber'). The idea is that we want to occasionally generate
+the same 'CurrencySymbol' or 'TokenName' for different 'Value's to have decent test coverage,
+hence to make name clashing more likely we pick from a predetermined set of
+'ByteString's. Otherwise the chance of generating the same 'ByteString' for two
+different 'Value's would be virtually zero.
+-}
+genShortHex :: Int -> Gen ByteString
+genShortHex i =
+  Base16.encode . BC.pack . show <$> elements [0 .. toCellCandidatesNumber i]
```
effectfully, line 272: This was naive on my part BTW, as you noted we should generate large bytestrings too and as Yura noted we should generate regular token and currency names. So this generator that I created is no good at all.

effectfully (approved):

plutus-core/plutus-core/test/Value/Spec.hs
zliu41: >  Just bind v and amt directly

Does that use `arbitrary @Integer`? If so does that mean I'll need to wrap it in a newtype?

effectfully: Oh yes, sorry, I'm stupid, you need the newtype wrapper.

PR #7349, opened 2025-09-18
Keep track of currency/token lengths in builtin Value

zliu41:
The memory usage of Values are updated to include a factor of max key length in the Value.


plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -374,22 +375,37 @@ instance ExMemoryUsage Data where
...
--- | Measure the size of a `Value` by taking the max of
--- (size of the outer map, size of the largest inner map).
-newtype ValueOuterOrMaxInner = ValueOuterOrMaxInner { unValueOuterOrMaxInner :: Value }
+instance ExMemoryUsage ValueTotalSize where
+  memoryUsage =
+    singletonRose
+      . fromInteger
+      . uncurry (*)
+      . (toInteger . Value.totalSize &&& toInteger . (+1) . Value.maxKeyLength)
+      . unValueTotalSize
```
effectfully, line 392: Nit: at this point I'd just use a `where` or a `let`.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -82,11 +88,19 @@ pack = pack' . normalize
 
 -- | Like `pack` but does not normalize.
 pack' :: NestedMap -> Value
-pack' (normalize -> v) = Value v sizes size
+pack' (normalize -> v) = Value v sizes lens size
```
effectfully: The comment says it doesn't normalize, but it does normalize.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -82,11 +88,19 @@ pack = pack' . normalize
...
  where
-  (sizes, size) = Map.foldl' alg (mempty, 0) v
-  alg (ss, s) inner =
+  (sizes, lens, size) = Map.foldrWithKey' alg (mempty, mempty, 0) v
+  alg currency inner (ss, ls, s) =
     ( IntMap.alter (maybe (Just 1) (Just . (+ 1))) (Map.size inner) ss
+    , IntMap.unionWith
+        (+)
+        (IntMap.alter (maybe (Just 1) (Just . (+ 1))) (B.length currency) ls)
```
effectfully: This `IntMap.alter <...>` pattern occurs a lot, give it a general name like `incOne` or whatever (and use that instead of `incLen`)?

effectfully (approved):

PR #7350, opened 2025-09-18
[Builtins] Implement 'ReadKnownIn' and 'MakeKnownIn' instances manually

effectfully:
This removes all `readKnown` and `makeKnown` calls from the GHC Core of builtins (I did check).


plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
```diff
@@ -245,42 +244,36 @@ class KnownTypeAst tyname uni x where
     typeAst :: Type tyname uni ()
     default typeAst :: KnownTypeAst tyname uni (ElaborateBuiltin uni x) => Type tyname uni ()
     typeAst = toTypeAst $ Proxy @(ElaborateBuiltin uni x)
-    {-# INLINE typeAst #-}
```
zliu41, line 248: What's the reason for these?

zliu41 (approved):

plutus-core/plutus-core/src/PlutusCore/Builtin/KnownTypeAst.hs
effectfully: It's types stuff, we don't have it anywhere on the evaluation path. Originally I added `INLINE` there just so that the person editing this code doesn't need to think whether to add `INLINE` or not, but now that `INLINE` pragmas don't get generated for newtype-derived instances, I thought we can as well remove them from here for consistency, since they don't do anything useful anyway.

effectfully:
The cost of not getting inlining right:

image
image
image

PR #7355, opened 2025-09-21
Benchmarking test 2

kwxm:
Let's try that again.


kwxm:
The [first attempt](https://github.com/IntersectMBO/plutus/pull/7354) produced some weird results, but this looks fine now.  Anyway, the benchmarking action works OK, which is what I wanted to make sure of.

PR #7357, opened 2025-09-23
Experimenting with an optimised list decoder

kwxm:
I just want to see if a bit of refactoring can speed things up.

Later: the [version](https://github.com/IntersectMBO/plutus/pull/7358) that uses `let` instead of `where` seems to be a bit faster, so we should probably abandon this one in favour of that.


kwxm:
I'm going to close this in favour of #7358.

PR #7358, opened 2025-09-23
Experimenting with an optimised list decoder (2)

kwxm:
Let's try again with `let` instead of `where`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -115,3 +115,4 @@ plutus-pab/test-node/alonzo-purple/db
 .nvimrc
 release-*
 TestCert.*
+worktrees
--- a/nix/agda-tools.nix
+++ b/nix/agda-tools.nix
@@ -143,7 +143,6 @@ let
 
   # Path to the stdlib .agda-lib file for shell export.
   NIX_AGDA_STDLIB = "${agda-stdlib}/standard-library.agda-lib";
-
 in
 
 {
--- a/nix/metatheory.nix
+++ b/nix/metatheory.nix
@@ -105,7 +105,6 @@ let
       exit 1
     fi
   '';
-
 in
 
 {
--- a/nix/outputs.nix
+++ b/nix/outputs.nix
@@ -97,9 +97,20 @@ let
     ghc910 = mkShell project.projectVariants.ghc910;
   };
 
+  # The default shell contains the agda-with-stdlib-and-metatheory package which will
+  # break on `nix develop` if the .lagda files are broken. In order to escape this
+  # situation we introduce a shell that doesn't contain that executable.
+  metatheory-jailbreak-shell = non-profiled-shells.default.overrideAttrs (attrs: {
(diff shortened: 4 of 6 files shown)


kwxm:
There's a bit of fluctuation in the benchmark results, but there's definitely an improvement, and it looks as if this is better than [the other version](https://github.com/IntersectMBO/plutus/pull/7357).  I say we merge this one.

plutus-core/flat/src/PlutusCore/Flat/Decoder/Strict.hs
```diff
@@ -55,15 +55,36 @@ import Numeric.Natural (Natural)
...
+                         b <- dBool
+                         if b
+                           then (x:) <$> goZero'
+                           else pure []
+                       {-# INLINE goZero' #-}
+                   in goZero'
+                 {-# INLINE goZero #-}
+             in runGet ((h:) <$> goZero h) end s'
+    else pure []
+    where
```
zliu41: Redundant

kwxm: Oops!  That's left over from the other version.

plutus-core/flat/src/PlutusCore/Flat/Decoder/Strict.hs
```diff
@@ -55,15 +55,35 @@ import Numeric.Natural (Natural)
...
+         \end s -> do
+           GetResult s' h <- runGet dec end s
+           if s' /= s
+             then
+             let goNormal = do
+                   b <- dBool
+                   if b
+                     then (:) <$> dec <*> goNormal
+                     else pure []
+                 {-# INLINE goNormal #-}
```
effectfully, line 72: It's a recursive function, so unless it gets worker-wrapper-transformed (which it may be), it's not gonna inline.

plutus-core/flat/src/PlutusCore/Flat/Decoder/Strict.hs
```diff
@@ -55,15 +55,35 @@ import Numeric.Natural (Natural)
...
+  if b
+    then Get $
+         \end s -> do
+           GetResult s' h <- runGet dec end s
+           if s' /= s
+             then
+             let goNormal = do
+                   b <- dBool
+                   if b
+                     then (:) <$> dec <*> goNormal
```
effectfully, line 70: Incidentally, there's a blog [post](https://www.joachim-breitner.de/blog/620-Constructing_a_list_in_a_Monad) about exactly this sort of thing and it suggests that this is indeed the most optimal way of implementing this logic.

kwxm:
> If you make the implementation wrong, will the existing tests fail?

The existing tests seem pretty comprehensive.  I tried making a couple of random changes and six or seven tests failed.  It's hard to be sure if they'll catch every possible error though.

PR #7359, opened 2025-09-23
Conformance tests for built-in functions on Value

basetunnel:
This contains uplc conformance tests for

- built-in functions `insertCoin`, `unionValue`, `lookupCoin`, `valueContains`
- constants of builtin `Value` type

Test cases are based on the semantics described in https://github.com/cardano-foundation/CIPs/pull/1088, but it seems like the semantics of `valueContains` are still being discussed. When that is finalised, I can add some more cases if necessary.


basetunnel:
Tests are failing because the metatheory needs to be updated. Working on that now.

basetunnel:
I've ignored these tests in the `agda-conformance` test suite for now, because the formalisation of `Value` in agda is not yet finished.

Also added some new ones to test maximum key length.

PR #7361, opened 2025-09-24
Add new metatheory-jailbreak devShell

zeme-wana:
The default shell contains the `agda-with-stdlib-and-metatheory` package which will break on `nix develop` if the `.lagda` files are broken. 
In order to escape this situation and make development possible, we introduce a shell that doesn't contain that executable.


ana-pantilie (approved):
Tested it out and it works great! Thank you @zeme-wana !

basetunnel:
Also works for me, thanks!

PR #7362, opened 2025-09-24
Disable for `wasm` everything that is disabled for windows and address issues with `KnownTypeAst` and `wasm`

palas:
In the same way that they are disabled for windows, this PR disables tests and executables for `wasm`, since they currently don't work in `wasm`.

We also specify the bits of some `Int`s and `Word`s so that it works when they are 32 bit by default, and thus solve an issue with `wasm` and `KnownTypeAst`.

Pre-submit checklist:
- Branch
- PR
    - [x] (For external contributions) Corresponding issue exists and is linked in the description
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1197,28 +1198,28 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
           DefaultFunSemanticsVariantB -> consByteStringMeaning_V1
           DefaultFunSemanticsVariantC -> consByteStringMeaning_V2
   toBuiltinMeaning _semvar SliceByteString =
-    let sliceByteStringDenotation :: Int -> Int -> BS.ByteString -> BS.ByteString
-        sliceByteStringDenotation start n xs = BS.take n (BS.drop start xs)
+    let sliceByteStringDenotation :: Int64 -> Int64 -> BS.ByteString -> BS.ByteString
```
zliu41: @kwxm can we do these?

palas:
@zliu41 There is another issue that doesn't show in the tests but it is making reading the protocol parameters from wasm fail (because of cost models apparently), and I think it is related to this: https://github.com/IntersectMBO/plutus/blob/0ddb89848cfd7f23199e23b7d3cb5a090054bea4/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs#L67-L68

This is the error I get:
```
RuntimeError: Expected Right, got Left: "Error in $.costModels: CMInternalWriteError \"parsing Int failed, value is either floating or will cause over or underflow 9.223372036854775807e18\""
  rightOrError, called at src-lib/Cardano/Wasm/Api/Tx.hs:194:25 in cardano-wasm-10.0.0.0-inplace-cardano-wasi-lib:Cardano.Wasm.Api.Tx
  estimateMinFeeImpl, called at src-wasm
```

Not sure why we just got this error with the latest changes? Some new cost param that exceeds 32bit int?

zliu41:
> I think it is related to this

@palas That's not a recent change. Neither is any of the other usage of `Int` - they've been `Int` for years. Does 1.58 break your wasm build, but it's fine with 1.57? I'm not sure what your wasm build does, but if that's the case, we can do a bisect.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
effectfully: If that is going to be run on a 32-bit system where correctness matters, then no: `fromIntegral` is going to silently truncate the given `Int64`s into `Int32`s.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1803,8 +1804,8 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
   -- Bitwise operations
 
   toBuiltinMeaning _semvar ReadBit =
-    let readBitDenotation :: BS.ByteString -> Int -> BuiltinResult Bool
-        readBitDenotation = Bitwise.readBit
+    let readBitDenotation :: BS.ByteString -> Int64 -> BuiltinResult Bool
+        readBitDenotation bs = Bitwise.readBit bs . fromIntegral
```
effectfully: Likewise.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1946,11 +1947,11 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
         {-# INLINE listToArrayDenotation #-}
      in makeBuiltinMeaning listToArrayDenotation (runCostingFunOneArgument . paramListToArray)
   toBuiltinMeaning _semvar IndexArray =
-    let indexArrayDenotation :: SomeConstant uni (Vector a) -> Int -> BuiltinResult (Opaque val a)
+    let indexArrayDenotation :: SomeConstant uni (Vector a) -> Int64 -> BuiltinResult (Opaque val a)
         indexArrayDenotation (SomeConstant (Some (ValueOf uni vec))) n =
           case uni of
             DefaultUniArray arg -> do
-              case vec Vector.!? n of
+              case vec Vector.!? fromIntegral n of
```
effectfully: Likewise.

effectfully (commented):
It is not safe to run UPLC on a 32-bit machine. This is discussed in `Note [Integral types as Integer]`. There's also

```haskell
{-|
The old implementation relied on this function which is safe
*only* for 64-bit systems. There were previously safety checks to fail compilation
on other systems, but we removed them  since we only test on 64-bit systems afterall. -}
naturalToWord64Maybe :: Natural -> Maybe Word64
naturalToWord64Maybe n = fromIntegral <$> naturalToWordMaybe n
```

Not to mention that `SatInt` is `Int` and not `Int64`.

There was also something recent that breaks on a 32-bit system, but I don't remember what (UPD: it was [this](https://github.com/IntersectMBO/plutus/issues/7478)).

Basically, if you're doing this for shits and giggles, go ahead, but if you actually want to run UPLC on a 32-bit systems, you absolutely can't do that.

effectfully:
Oh, and the whole bitwise builtins business relies on the architecture being 64-bit, e.g.

```haskell
let bigSrcPtr :: Ptr Word64 = castPtr srcPtr
```

Can it be safe to still use it on a 32-bit platform? I dunno, but I bet it wasn't written with that in mind, so again, if you're going to use this for something that matters, then you can't.

effectfully:
> There was also something recent that breaks on a 32-bit system, but I don't remember what.

[this](https://github.com/IntersectMBO/plutus/issues/7478)

palas:
@effectfully @zliu41. In order to know whether `cardano-wasm` is feasible we would need to know which level of compatibility with 32bit can be guaranteed by Plutus. All of these are obviously possible technically, but we don't know how hard they are, so it may not be worth the effort. The options I see are, from more to less ideal:

- Full 32bit support

It seems for UPLC, using `Int64` would make the code inefficient, but there is the possibility of having conditional compilation so that it is only inefficient for 32bits, and that the code that gets compiled for 64bits doesn't change. We probably don't care about UPLC being ran efficiently in `cardano-wasm`, because we are not doing block production, but we do care about it being accurate, because the plan was to support transaction balancing.

- 32bit support for everything but UPLC code execution

At least compilation used to work fine until 1.57 (included). It is no longer the case as of 1.58. But we don't know if there may be runtime issues anywhere. Maybe you can advise on this. If we can get everything but UPLC execution to work in `cardano-wasm` there is still the possibility of just giving up transaction balancing and supporting the rest.

- No 32bit support

If no 32bit support can be guaranteed then we probably need to discontinue `cardano-wasm` because it is not even compiling right now.

---

In any case, I cannot really help with any of these because I don't have the expertise, this PR was just best-effort hoping it would be some small issue, but obviously it isn't.

So, which level of 32bit support can you feasibly guarantee?

Are there any other levels I am not considering?

effectfully:
> In order to know whether cardano-wasm is feasible we would need to know which level of compatibility with 32bit can be guaranteed by Plutus.

Right now, 32-bit is explicitly precluded and you simply cannot run UPLC on anything 32-bit if you at all care about getting the correct behavior (I'm not talking about efficiency here). This is an assumption that has been built-in for a very long time, so backtracking on that now would be a massive effort, someone will have to review basically every single line of critical code to verify that it's safe to run on a 32-bit system.

If you only want things to build, I think it's fine. The benchmark results are almost certainly nonsense, those `Int`s are already `Int64` and those `fromIntegral` calls should be no-op in GHC Core, but even if that somehow not the case, you still wouldn't get a slowdown in `validation` benchmarks across the board.

> So, which level of 32bit support can you feasibly guarantee?

Well, I no longer work on the project, but the answer is 0. "It builds but running it will result in undefined behavior"

zliu41:
> 32bit support for everything but UPLC code execution

What else is useful to you, if you can't use the UPLC evaluator in wasm?

> probably need to discontinue cardano-wasm

It can continue but you'll need a different evaluator implementation. There are already several other UPLC evaluator implementations in other languages (e.g., talk to Lucas Rosa), some of which may support 32-bit systems better.

> At least compilation used to work fine until 1.57 (included)

I'm surprised the compilation was working. I still don't know what it is in 1.58 that broke it. We can do a bisect, but given what Roman said above, do you think it's still worth doing? @palas

palas:
> What else is useful to you, if you can't use the UPLC evaluator in wasm?

I think it is mainly the fact that it is integrated in the types of ledger. So, inspecting UPLC code, parsing of cost models within protocol parameters. There may be more things. The risk is that it may mangle some of those values in 32bits, if it makes assumptions that only hold for 64bit.

> I'm surprised the compilation was working. I still don't know what it is in 1.58 that broke it. We can do a bisect, but given what Roman said above, do you think it's still worth doing?

Maybe in the long term it won't. But if we break the compilation and tests of `cardano-wasm` it will start to degrade and is tough getting back from there if we have any hope of it eventually working out. So if it is not too hard, it would be good to patch it so that it holds while we make a decision.

It is probably an issue with `SatInt` having an `Int` internally. I can try writing a workaround for that and see if the test passes, may be the easiest.

Is the current patch mergeable so far? I think the concerns about `fromIntegral` are not problematic because they talk about indexes in containers, and I really doubt there can be a container that is larger than `2^31 - 1` in the context of Plutus.

palas:
Nice! The last commit I pushed passes the `cardano-wasm` test. That is good because we can at least merge the latest plutus and not have `cardano-wasm` broken just yet.

I guess the implementation of multiplication may be less efficient than the previous one, but maybe even that can be fixed, there are several techniques to check for overflow, it would be a matter of measuring

plutus-core/satint/src/Data/SatInt.hs
```diff
@@ -111,57 +117,38 @@ kind of overflow we're facing, and pick the correct result accordingly.
...
-  case mulIntMayOflo# x# y# of
-    0# -> SI (I# (x# *# y#))
-    -- Overflow
-    _ ->
-      if isTrue# ((x# ># 0#) `andI#` (y# ># 0#))
+timesSI (SI a) (SI b) =
+  let a' = toInteger a
+      b' = toInteger b
+      r = a' * b'
+   in if r > maxBoundInteger
```
effectfully, line 233: Yeah you can't do this, it's very inefficient.

plutus-core/satint/src/Data/SatInt.hs
```diff
@@ -111,57 +117,38 @@ kind of overflow we're facing, and pick the correct result accordingly.
...
-    -- Overflow
-    _ ->
-      if isTrue# ((x# >=# 0#) `andI#` (y# <# 0#))
+minusSI (SI a) (SI b) =
+  let r = a - b
+   in if a >= 0 && b < 0 && r < 0
         then maxBound
         else
-          if isTrue# ((x# <=# 0#) `andI#` (y# ># 0#))
+          if a < 0 && b > 0 && r > 0
```
effectfully, line 223: This may also be significantly less efficient than before.

plutus-core/satint/src/Data/SatInt.hs
palas: I probably can do much better, this was just to know if it fixed the issue

SeungheonOh: I think it's okay:

original core:
```
timesSI :: SatInt -> SatInt -> SatInt
timesSI
  = \ (ds :: SatInt) (ds1 :: SatInt) ->
      case ds `cast` <Co:1> :: ... of { I# x# ->
      case ds1 `cast` <Co:1> :: ... of { I# y# ->
      case mulIntMayOflo# x# y# of {
        __DEFAULT ->
          case andI# (># x# 0#) (># y# 0#) of {
            __DEFAULT ->
              case andI# (># x# 0#) (<# y# 0#) of {
                __DEFAULT ->
                  case andI# (<# x# 0#) (># y# 0#) of {
                    __DEFAULT ->
                      case andI# (<# x# 0#) (<# y# 0#) of {
                        __DEFAULT -> overflowError;
                        1# -> maxInt `cast` <Co:2> :: ...
                      };
                    1# -> minInt `cast` <Co:2> :: ...
                  };
                1# -> minInt `cast` <Co:2> :: ...
              };
            1# -> maxInt `cast` <Co:2> :: ...
          };
        0# -> (I# (*# x# y#)) `cast` <Co:2> :: ...
      }
      }
      }
```

new core:
```
timesSI :: SatInt -> SatInt -> SatInt
timesSI
  = \ (ds :: SatInt) (ds1 :: SatInt) ->
      case ds `cast` <Co:1> :: ... of { I64# a ->
      case ds1 `cast` <Co:1> :: ... of { I64# b ->
      let {
        b_native :: Int#
        b_native = int64ToInt# b } in
      let {
        a_native :: Int#
        a_native = int64ToInt# a } in
      case mulIntMayOflo# a_native b_native of {
        __DEFAULT ->
          case ==# (># a_native 0#) (># b_native 0#) of {
            __DEFAULT -> $fBoundedInt64_$cminBound `cast` <Co:2> :: ...;
            1# -> $fBoundedInt64_$cmaxBound `cast` <Co:2> :: ...
          };
        0# ->
          (I64# (intToInt64# (*# a_native b_native))) `cast` <Co:2> :: ...
      }
      }
      }
```

SeungheonOh:
I went over GHC core changes, it seems most things are okay, `fromSatInt` seems to have gotten slower; I'll see if there's a way to make it generate better core for `fromSatInt`.

SeungheonOh:
no clue where +10.1% bench result is coming from. It's about 5% faster on my local machine

palas:
let's see if the rebased version does better

SeungheonOh:
I think the culprit was `plusSI` and `minusSI` where GHC doesn't know to fully collapse bound checks. I made it so that it does old Int# operations directly when using 64 bit machine. This is slightly ugly but it retains original performance fully

effectfully:
Still hate this PR. It pretends that UPLC can run on a 32-bit platform, which it absolutely cannot, see [this](https://github.com/IntersectMBO/plutus/pull/7362#issuecomment-3867287046) message and the one below.

What's the point to build on a 32-bit platform if you can't run there? And complicate actual critical-path code for that. Like, this project isn't a property of Intersect, it belongs to the Cardano community, and so I as a member ask: why are you complicating critical-path code to make unrunnable thing buildable? What's the motivation there?

SeungheonOh:
Honestly, I think it's best to just remove `cardano-wasm`'s dependency to this codebase altogether and implement a separate implementation (or use existing ones that can be compiled to wasm -- check https://github.com/utxo-company/plutuz). Cardano node itself doesn't target 32bit machines so I agree with Roman's point that it is best to keep critical pass on this implementation of CEK clean and free of 32bit-specific logic.

palas:
> Still hate this PR. It pretends that UPLC can run on a 32-bit platform, which it absolutely cannot, see https://github.com/IntersectMBO/plutus/pull/7362#issuecomment-3867287046 message and the one below.
> 
> What's the point to build on a 32-bit platform if you can't run there? And complicate actual critical-path code for that. Like, this project isn't a property of Intersect, it belongs to the Cardano community, and so I as a member ask: why are you complicating critical-path code to make unrunnable thing buildable? What's the motivation there?

Even if wasm doesn't support UPLC, there is a lot more to Cardano than UPLC, and we cannot use any of it if Plutus doesn't even compile to 32 bit. I am no longer asking for full 32 bit compatibility.

Also Plutus is the only component in Cardano that doesn't compile to 32 bit. Not sure why it doesn't too, because Cardano is meant to be a universal platform, so it seems a very arbitrary decision to assume 64 bit words for the implementation of an abstract machine, but never mind that.

Haskell is a platform independent language and the only reason I see why this project is not working in 32 bits (at least for basic functionalities) is because it is using a machine dependent type `Int` (well, also `Word`) which is equivalent to `Int64` in 64 bit machines. So I don't understand why, if the type is assumed to be 64 bits, it is not represented as `Int64`? That is what `Int64` is for. Instead, the choice was using "`Int` but it doesn't work if it is not 64 bits", I am not sure what is the point of making this decision nor the point of not rectifying it.

Also, it should be possible to have all the changes necessary compiled conditionally to 32bit and this shouldn't affect the 64bit compilation at all. The changes I did in the 64bit `SatInt` path, I only proposed them because I thought they were improvement but I can remove them, and I am going to do that next.

Ultimately, it is not my choice, but not allowing Plutus to acknowledge 32bit just gives me a lot of work each time a new release of Plutus comes out, because I have to keep rebasing this PR and resolving the conflicts. And, like I say, Plutus is the only component in Cardano that has this problem.

Unfortunately, your suggestion of using an alternative implementation of Plutus doesn't help at all with compiling `cardano-api` to wasm, because all the components in the stack are statically linked.

palas:
> and I am going to do that next.

Ah, it seems @SeungheonOh already basically did that: so basically  calling `int64ToInt#` before and `intToInt64#` afterwards, the rest is the same.

I could go even further and do the types conditional too, but that would clutter much more 🤷‍♂️

effectfully:
> Even if wasm doesn't support UPLC, there is a lot more to Cardano than UPLC, and we cannot use any of it if Plutus doesn't even compile to 32 bit.

Swap the Plutus dependency on your build entirely if you don't need it? Just make a fake Plutus that always throws errors.

> Not sure why it doesn't too, because Cardano is meant to be a universal platform, so it seems a very arbitrary decision to assume 64 bit words for the implementation of an abstract machine, but never mind that.

It's for the builtins mostly, not the machine itself.

> Instead, the choice was using "Int but it doesn't work if it is not 64 bits", I am not sure what is the point of making this decision nor the point of not rectifying it.

`Int` is a bit more ergonomic than `Int64` and as far as I understand the original idea was to only support 64 bits anyway, so why not use `Int` and error out on non-64-bit systems.

> Also, it should be possible to have all the changes necessary compiled conditionally to 32bit and this shouldn't affect the 64bit compilation at all.

Sure, you can have a buggy 32-bit version of Plutus at the cost of adding more logic to the critical path. I don't think this is a good idea at all.

> Ultimately, it is not my choice, but not allowing Plutus to acknowledge 32bit just gives me a lot of work each time a new release of Plutus comes out, because I have to keep rebasing this PR and resolving the conflicts.

I still don't really know what you do or why.

> I am not sure what is the point of making this decision nor the point of not rectifying it.

If you want to rectify that decision, you have my blessing for as long as you're gonna guarantee that the 32-bit is no more buggy than the 64-bit one, including all builtins and everything on the critical path.

This is the PR where it was decided to abandon 32-bit systems: #4413. The only reason specified is "32-bit systems keep causing us trouble" and I don't know what that means.

palas:
> Swap the Plutus dependency on your build entirely if you don't need it? Just make a fake Plutus that always throws errors.

`cardano-wasm` depends on `cardano-api` which depends on `plutus-core`. `cardano-wasm` cannot swap the `plutus-core` dependency, also because the transaction and the protocol parameters formats include plutus types.

> It's for the builtins mostly, not the machine itself.

But that makes the whole Plutus machine dependent, and it is meant to be a language that can be executed the same everywhere, that is the whole point, so that you always get the same result, otherwise we could have just used x86_64 machine code instead of UPLC.

Also, Cardano should run in a diverse set of machines so that it is not vulnerable to issues with a particular type of machine. I understand that we cannot **actively** support every type of machine, but just writing the code (at this level of abstraction) purposefully wrong so that it doesn't work for other bit-sizes seems completely unnecessary.

> Int is a bit more ergonomic than Int64 and as far as I understand the original idea was to only support 64 bits anyway, so why not use Int and error out on non-64-bit systems.

What do you mean by ergonomic? What are the differences between `Int` and `Int64` in 64 bit? Using a primitive that is machine dependent to represent a 64 bit integer is not a consistent usage of types, and I don't see the advantage.

> Sure, you can have a buggy 32-bit version of Plutus at the cost of adding more logic to the critical path. I don't think this is a good idea at all.

We are not adding any logic, just conversion functions that do nothing if the platform is 64bits. Also, would you prefer a cellphone that is a little water proof or not water proof at all? We can have a disclaimer saying that 32bit is not supported, or require an unstable/unsafe flag. The all-or-nothing is an unnecessary requirement.

Still, it would be possible to test the 32bit version with existing Plutus contracts and see if they work fine, and probably it is not **that** difficult to get them to work (although not efficiently), but I understand that that is a lot to ask for, and we all have to prioritise our resources.

> I still don't really know what you do or why.

I spent something like 6 months working on a prototype of a library that exposes the functionalities of `cardano-api` as a wasm module, so that these functionalities can be accessed from anywhere, for example, from JS inside a browser, but also from other languages. Basically trying to make Cardano more accessible, and that dApp creators don't have to rely on libraries created by the community and can use the "official" implementation directly, which tends to be up to date, and also is very well tested and robust because it is type driven.

And that works fine, the only issue is that I had to patch a lot of the stack of `cardano-api` (including Plutus) for it to work.

And, since you say that Plutus could never work like this. We have pretty much decided to freeze this work, since most dApps need support for UPLC, and it doesn't make that much sense to have a library for Cardano that doesn't support it.

But even then, I would rather not trash the work of 6 months, and I would like the part that is already done to keep working. If we ever want to go that route, then it won't be necessary to spend those 6 months again. And also the project is a very good test for GHC's wasm target.

Again, it is not my call, and if the Plutus team decides I am asking for too much, then I'll just trash it and move on. But I think it is a real pity.

> If you want to rectify that decision, you have my blessing for as long as you're gonna guarantee that the 32-bit is no more buggy than the 64-bit one, including all builtins and everything on the critical path.

I would love to, but I won't be paid to do that nor have I the expertise needed to maintain Plutus, and my days only have so many hours.

> This is the PR where it was decided to abandon 32-bit systems: https://github.com/IntersectMBO/plutus/pull/4413. The only reason specified is "32-bit systems keep causing us trouble" and I don't know what that means.

Thanks, that is very useful. My understanding is that it is difficult to support builtins from 32-bit systems. And that makes sense and it is fine, but that doesn't mean that the rest of stuff, which is easy to support, needs to break too.

Curiously, it looks like 32-bit was being supported because of GHCJS, and because GHCJS didn't go forward, then it was decided there were no reasons to support 32-bit. And now that GHC to WASM is a viable alternative, it cannot be done because 32-bit is not supported.

ana-pantilie:
@palas These changes may affect the 64 bit version of the CEK machine in unpredictable ways, which is a much bigger problem for `plutus` compared to other projects because we need to make sure costing is correct. Introducing new dependencies in the CEK machine may break things now (so it would require re-costing), and they may break things in the future once we upgrade those dependencies making the CEK machine even more delicate than it already is. @kwxm please correct me if I'm mistaken about this.

I think we would need separate denotations for builtins and a separate costing machinery for the 32 bit version, so that would need to be added to this PR in order to make this mergeable. Otherwise we'd be merging a broken feature into master which is not something we want to do regardless of the feature, especially not one which may break the most security-critical part of the codebase.

Adding and maintaining 32 bit support is definitely a significant undertaking. If you're not interested in having the evaluator work then I'd also suggest forking the project and finding solutions to whatever problems there are with doing that.

palas:
@ana-pantilie, to answer your concerns:

- No new dependencies are introduced. The 64-bit code path uses the same primops (`addIntC#`, `subIntC#`, `timesInt2#`). The only difference is `int64ToInt#`/`intToInt64#` conversions that are zero-cost on 64-bit.
- Costing is unaffected. The latest benchmark results show no regression (apart from noise). And the generated assembly code on 64-bit is identical, @SeungheonOh can confirm.
- I'm not asking for 32-bit support. UPLC evaluation on 32-bit would remain unsupported. This PR only makes small changes which are needed because `cardano-ledger` depends on `plutus-core` for some types and serialization.

Could you point me to a specific change in this PR that you believe could affect costing? I'd like to address any concrete concern.

effectfully:
> UPLC evaluation on 32-bit would remain unsupported.

Why not just make the build-breaking `ReadKnownIn` and `MakeKnownIn` instances throw on 32-bit systems then instead of breaking the build? It'll be as correct as the current 32-bit code and you won't pollute anything else as far as I understand. Or is that wrong?

palas:
> Why not just make the build-breaking ReadKnownIn and MakeKnownIn instances throw on 32-bit systems then instead of breaking the build? It'll be as correct as the current 32-bit code and you won't pollute anything else as far as I understand. Or is that wrong?

@effectfully, do you mean like this? [Commit 6f26907](https://github.com/IntersectMBO/plutus/pull/7362/changes/6f2690703076b830caeb20a36a161defe65c2036#diff-4c05db6daf04fdffe5653ce0426d2707197302397c7382e97f1b1c745973be2a)

That works for me 👍

effectfully:
Yeah, something like that. I'm not sure if it should be a hard `error` or a soft `fail`, but whatever. Ideally, the evaluator would check if it's on a 64-bit system at startup and throw if it's not.

SeungheonOh (approved):
Looks good now after making 32bit route failing

effectfully:
FYI https://discourse.haskell.org/t/rfc-sunsetting-i386/14149

PR #7370, opened 2025-09-26
Revert "Keep track of currency/token lengths in builtin Value (#7349)"

zliu41:
@Unisay noted that currency symbols and token names are capped at 32 bytes, which means we don't need to take map key length into account when costing, hence reverting the PR for keeping track of the lengths.


effectfully (approved):
Well, this needs to be reflected in the builtins, documented and tested then. It's fine to do that in follow-ups though.

zliu41:
> Do you also plan to change InsertCoin and UnValueData to fail if applied to overly long currency symbols (> 28) and token names (> 32)?

Yep

PR #7371, opened 2025-09-28
Value decoding/parsing: reject currencies/tokens longer than 32 bytes

zliu41:
This PR:
- Restricts currency symbol and token name lengths in builtin Value to be <=32 bytes
- Rejects invalid currency symbols and token names when decoding Value from CBOR, decoding Value from Flat, and parsing Value.

This PR does _not_ update `insertCoin` or `unValueData` to reject invalid currency symbols and token names. This will be done separately.


plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
```diff
@@ -90,7 +90,16 @@ conArray uniA = Vector.fromList <$> conList uniA
...
+conValue = do
+  Value.fromList <$> (traverse validateKeys =<< conList knownUni)
+ where
+  validateToken (token, amt) = do
+    tk <- maybe (fail $ "Invalid token: " <> show (unpack token)) pure (Value.k token)
+    pure (tk, amt)
+  validateKeys (currency, tokens) = do
+    ck <- maybe (fail $ "Invalid currency: " <> show (unpack currency)) pure (Value.k currency)
+    tks <- traverse validateToken tokens
+    pure (ck, tks)
```
Unisay, line 102: I couldn't find corresponding changes in tests that prove this logic is correct.

zliu41: Added tests

kwxm: This would be a reasonable thing to add to the conformance tests as well.  Unfortunately we don't have any conformance tests for `flat`...

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -110,19 +161,20 @@ empty :: Value
 empty = Value mempty mempty 0
 {-# INLINE empty #-}
 
-toList :: Value -> [(ByteString, [(ByteString, Integer)])]
+toList :: Value -> [(K, [(K, Integer)])]
 toList = Map.toList . Map.map Map.toList . unpack
 {-# INLINEABLE toList #-}
 
-toFlatList :: Value -> [(ByteString, ByteString, Integer)]
+toFlatList :: Value -> [(K, K, Integer)]
```
kwxm, line 168: For a moment I wondered what on earth this had to do with `flat`...

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -5,6 +5,10 @@
 
 module PlutusCore.Value (
   Value, -- Do not expose data constructor
+  K, -- Do not expose data constructor
+  k,
```
kwxm, line 9: Will users need to use this in Plinth?  If so, `k` might be a bit short: it could clash with a variable name, for instance.

kwxm (approved):
This seems fine.  If we didn't have this, what would the ledger do if we managed to produce an output with an invalid key?

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: Nope. And this module is intended to be imported qualified anyway.

PR #7374, opened 2025-09-30
Cover some missing cases in the conformance tests for writeBits

kwxm:
Closes https://github.com/IntersectMBO/plutus-private/issues/1852

The [conformance tests ](https://github.com/IntersectMBO/plutus/pull/7369)for the BLS12-281 multi-scalar multiplication builtins revealed a small [mistake in the metatheory](https://github.com/IntersectMBO/plutus/pull/7369#discussion_r2381226534), which in turn led me to realise that the tests for `writeBits` weren't covering the case of an empty list of updates, which should do nothing (`writeBits` takes a bytestring, a boolean value `b` and a list of bit positions to set or clear according to the value of `b`).  This adds some extra cases to cover that, and also one that checks that you're not allowed to write to a negative position in the empty bytestring (we already have tests that updates to the empty bytestring fail in `case-01`-`case-04`, but those only check positive positions).


ana-pantilie (approved):
LGTM!

PR #7376, opened 2025-09-30
valueContains: enforce no negative amounts in either Value

zliu41:
Per discussion and consensus in https://github.com/cardano-foundation/CIPs/pull/1088


plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -103,17 +104,25 @@ data Value
       {- ^ Total size, i.e., sum total of inner map sizes. This avoids recomputing
       the total size during the costing of operations like `unionValue`.
       -}
+      {-# UNPACK #-} !Int
```
ana-pantilie, line 107: Ah, neat idea!

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -189,52 +203,70 @@ instance Pretty Value where
...
       (_, Nothing) -> fail $ "insertCoin: invalid token: " <> show (B.unpack token)
       (Just ck, Just tk) ->
         let f
               :: Maybe (Map K Integer)
-              -> ( -- Just (old size of inner map) if the total size grows by 1,
-                   -- otherwise Nothing
-                   Maybe Int
+              -> ( -- Left (old size of inner map) if the total size grows by 1,
+                   -- otherwise, Right (old amount)
+                   Either Int Integer
```
ana-pantilie, line 216: Similar to the concept of "boolean blindness", I prefer to define my own, equivalent, datatypes in these situations.

So, in this case, instead of writing a comment which documents what `Either Int Integer` represents, you can have the code document itself if you define a type `data Growth = OldInner Int | OldAmount Integer` (very bad names, but you get the idea).

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -189,52 +203,70 @@ instance Pretty Value where
...
-                let (isJust -> exists, inner') =
+                let (moldAmt, inner') =
                       Map.insertLookupWithKey (\_ _ _ -> amt) tk amt inner
-                 in (if exists then Nothing else Just (Map.size inner), Just inner')
-            (mold, outer') = Map.alterF f ck outer
-            (sizes', size') = case mold of
-              Just old -> (updateSizes old (old + 1) sizes, size + 1)
-              Nothing  -> (sizes, size)
-         in pure $ Value outer' sizes' size'
+                 in (maybe (Left (Map.size inner)) Right moldAmt, Just inner')
```
ana-pantilie, line 224: Similar to the above, this combination of Maybe, Either and tuples makes it hard to understand what is going on without running the whole function in your head.

Unisay (approved):
LGTM

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: This is very local - used only in a local function inside a function, so I'd very much rather not add a dedicated top-level data type just for this.

PR #7378, opened 2025-10-02
Enable BLS12-381 MSM builtins at PV11

### LINKED ISSUE #7330: Complete BLS12-381 MSM integration with `plutus-ledger-api`.
The two MSM builtins have been added to the cost model parameters, but there is still work to be done to complete the `plutus-ledger-api` integration, as described in [this document](https://github.com/IntersectMBO/plutus/blob/master/doc/notes/release-builtins/release-builtins.md).

**THIS WILL ENABLE THE NEW BUILTINS ON THE CHAIN** at the next HF, and should only be done after we've completed most of the other MSM tasks, specifically #7324, #7325, #7327 #7328, and #7329.  We'll also need to do #7326 , but that can wait.

kwxm:
Closes #7330.

Closes https://github.com/IntersectMBO/plutus-private/issues/1812

This enables the two BLS12-381 MSM buitins at PV11 by adding them to `batch6` of the builtins, which makes them visible to the `builtinsAvailableIn` check that gets performed during deserialisation.  The cost model parameter names were added in an earlier PR, so there's no need to do anything for them here.

### DIFF
--- a/plutus-ledger-api/testlib/PlutusLedgerApi/Test/V3/Data/EvaluationContext.hs
+++ b/plutus-ledger-api/testlib/PlutusLedgerApi/Test/V3/Data/EvaluationContext.hs
@@ -88,6 +88,8 @@ clearBuiltinCostModel r = r
                , paramLengthOfArray = mempty
                , paramListToArray = mempty
                , paramIndexArray = mempty
+               , paramBls12_381_G1_multiScalarMul = mempty
+               , paramBls12_381_G2_multiScalarMul = mempty
                }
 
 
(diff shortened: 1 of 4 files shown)


plutus-ledger-api/testlib/PlutusLedgerApi/Test/V3/Data/EvaluationContext.hs
```diff
@@ -88,6 +88,8 @@ clearBuiltinCostModel r = r
                , paramLengthOfArray = mempty
                , paramListToArray = mempty
                , paramIndexArray = mempty
+               , paramBls12_381_G1_multiScalarMul = mempty
```
kwxm, line 91: This isn't strictly necessary, but these tests will have to be reworked at some point (see #6089) so I'm adding them just in case.

plutus-ledger-api/testlib/PlutusLedgerApi/Test/V3/EvaluationContext.hs
```diff
@@ -95,6 +95,8 @@ clearBuiltinCostModel r = r
                , paramLengthOfArray = mempty
                , paramListToArray = mempty
                , paramIndexArray = mempty
+               , paramBls12_381_G1_multiScalarMul = mempty
```
kwxm, line 98: Same here.

PR #7389, opened 2025-10-15
Restrict Value quantities to signed 128-bit integer range

Unisay:
## Summary

This PR introduces a `Quantity` newtype to enforce bounds checking for Value amounts, restricting them to the signed 128-bit integer range (-2^127 to 2^127-1).

### Core Changes
- Add `Quantity` newtype in `PlutusCore.Value` with smart constructor and validators
- Update `unionValue` to return `BuiltinResult Value` to handle overflow detection
- Add validation for currency symbols and token names (max 32 bytes)
- Update all Value operations to use `Quantity` instead of raw `Integer`
- Add `ArbitraryBuiltin` instance for `Quantity` with improved value distribution

### Test Updates
- Update golden test for `unionValue` signature change (`Value -> Value -> Value` → `Value -> Value -> BuiltinResult Value`)
- Fix test files to handle `BuiltinResult` and `Quantity` types
- Migrate property test generators to use `arbitraryBuiltin` for better test coverage
- All existing tests pass with the new implementation

## Implementation Details

- `Quantity` uses `UnsafeQuantity` constructor (not exported)
- Smart constructor `quantity :: Integer -> Maybe Quantity` validates bounds
- All entry points validate bounds and return `BuiltinResult` for operations that can fail
- Internal operations use unsafe functions with validation afterward
- Consistent terminology: "quantity" instead of "amount" throughout
- `arbitraryBuiltin` provides better distribution of test values than stock `Arbitrary`, with improved coverage of edge cases near bounds

Closes https://github.com/IntersectMBO/plutus-private/issues/1881

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
@@ -2063,7 +2063,7 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
             (runCostingFunThreeArguments . unimplementedCostingFun)
 
     toBuiltinMeaning _semvar UnionValue =
-      let unionValueDenotation :: Value -> Value -> Value
+      let unionValueDenotation :: Value -> Value -> BuiltinResult Value
           unionValueDenotation = Value.unionValue
           {-# INLINE unionValueDenotation #-}
        in makeBuiltinMeaning
--- a/plutus-core/plutus-core/test/TypeSynthesis/Golden/Signatures/DefaultFun/UnionValue.golden.sig
+++ b/plutus-core/plutus-core/test/TypeSynthesis/Golden/Signatures/DefaultFun/UnionValue.golden.sig
@@ -1,1 +1,1 @@
-Value -> Value -> Value
+Value -> Value -> BuiltinResult Value
--- a/plutus-tx/src/PlutusTx/Builtins/Internal.hs
+++ b/plutus-tx/src/PlutusTx/Builtins/Internal.hs
@@ -1103,16 +1103,22 @@ lookupCoin (BuiltinByteString c) (BuiltinByteString t) (BuiltinValue v) =
 {-# OPAQUE lookupCoin #-}
 
 unionValue :: BuiltinValue -> BuiltinValue -> BuiltinValue
-unionValue (BuiltinValue v1) (BuiltinValue v2) = BuiltinValue $ Value.unionValue v1 v2
+unionValue (BuiltinValue v1) (BuiltinValue v2) =
+  case Value.unionValue v1 v2 of
+    BuiltinSuccess v -> BuiltinValue v
+    BuiltinSuccessWithLogs logs v -> traceAll logs (BuiltinValue v)
+    BuiltinFailure logs err ->
+      traceAll (logs <> pure (display err)) $ Haskell.error "unionValue errored."
 {-# OPAQUE unionValue #-}
 
 valueContains :: BuiltinValue -> BuiltinValue -> Bool
-valueContains (BuiltinValue v1) (BuiltinValue v2) = case Value.valueContains v1 v2 of
-  BuiltinSuccess r -> r
-  BuiltinSuccessWithLogs logs r -> traceAll logs r
-  BuiltinFailure logs err ->
-    traceAll (logs <> pure (display err)) $
-      Haskell.error "valueContains errored."
+valueContains (BuiltinValue v1) (BuiltinValue v2) =
+  case Value.valueContains v1 v2 of
+    BuiltinSuccess r -> r
+    BuiltinSuccessWithLogs logs r -> traceAll logs r
+    BuiltinFailure logs err ->
+      traceAll (logs <> pure (display err)) $
+        Haskell.error "valueContains errored."
 {-# OPAQUE valueContains #-}
 
 mkValue :: BuiltinValue -> BuiltinData
@@ -1120,12 +1126,13 @@ mkValue (BuiltinValue v) = BuiltinData $ Value.valueData v
 {-# OPAQUE mkValue #-}
 
 unsafeDataAsValue :: BuiltinData -> BuiltinValue
-unsafeDataAsValue (BuiltinData d) = case Value.unValueData d of
-  BuiltinSuccess v -> BuiltinValue v
-  BuiltinSuccessWithLogs logs v -> traceAll logs (BuiltinValue v)
-  BuiltinFailure logs err ->
-    traceAll (logs <> pure (display err)) $
-      Haskell.error "Data to Value conversion errored."
+unsafeDataAsValue (BuiltinData d) =
+  case Value.unValueData d of
+    BuiltinSuccess v -> BuiltinValue v
+    BuiltinSuccessWithLogs logs v -> traceAll logs (BuiltinValue v)
+    BuiltinFailure logs err ->
+      traceAll (logs <> pure (display err)) $
+        Haskell.error "Data to Value conversion errored."
 {-# OPAQUE unsafeDataAsValue #-}
 
 caseInteger :: Integer -> [a] -> a


plutus-core/plutus-core/src/PlutusCore/Builtin/Result.hs
```diff
@@ -1,12 +1,10 @@
 -- editorconfig-checker-disable-file
-{-# LANGUAGE FlexibleInstances      #-}
-{-# LANGUAGE FunctionalDependencies #-}
```
Unisay, line 3: Dropped unused pragmas: TemplateHaskell, FunctionalDependencies.

plutus-core/plutus-core/src/PlutusCore/Pretty/PrettyConst.hs
```diff
@@ -113,7 +113,7 @@ instance NonDefaultPrettyBy ConstConfig T.Text where
     nonDefaultPrettyListBy conf = Prettyprinter.list . Prelude.map (nonDefaultPrettyBy conf)
     nonDefaultPrettyBy = inContextM $ \t -> pure $ pretty $ "\"" <> escape t <> "\""
         where
-            escape t = T.foldr' prettyChar "" t
```
Unisay, line 116: There was an HLint warning.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -249,74 +340,75 @@ deleteCoin (UnsafeK -> currency) (UnsafeK -> token) (Value outer sizes size neg)
...
+@a@ contains @b@ if for each @(currency, token, quantity)@ in @b@, if @quantity > 0@, then
+@lookup currency token a >= quantity@, and if @quantity < 0@, then
+@lookup currency token a == quantity@.
 -}
 valueContains :: Value -> Value -> BuiltinResult Bool
 valueContains v1 v2
   | negativeAmounts v1 > 0 = fail "valueContains: first value contains negative amounts"
   | negativeAmounts v2 > 0 = fail "valueContains: second value contains negative amounts"
-  | otherwise = BuiltinSuccess . getAll $ Map.foldrWithKey' go mempty (unpack v2)
+  | otherwise = BuiltinSuccess . getAll $ Map.foldrWithKey' go (All True) (unpack v2)
```
zliu41: ? `mempty` is much better than `All True` because you don't have to decide whether it should be `True` or `False`. That's the whole point of `All`.

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -86,12 +128,31 @@ prop_insertCoinValidatesToken v =
           BuiltinFailure{} -> property True
           _                -> property False
 
-prop_lookupAfterInsertion :: Value -> ValueAmount -> Property
-prop_lookupAfterInsertion v (ValueAmount amt) =
+prop_insertCoinValidatesQuantityMin :: Value -> Property
+prop_insertCoinValidatesQuantityMin v =
+  forAll gen32BytesOrFewer $ \c ->
+    forAll gen32BytesOrFewer $ \t ->
+      let amt = V.minQuantity - 1
```
zliu41: This isn't really a property-based test, since the random parts are the currencies and tokens, while the quantity is fixed at `minBound - 1`. Try generating more quantities.

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -124,7 +185,12 @@ prop_deleteCoinPreservesInvariants v =
   vs = scanr (\(c, t, _) -> V.deleteCoin (V.unK c) (V.unK t)) v fl
 
 toPositiveValue :: Value -> Value
-toPositiveValue = V.pack . fmap (fmap abs) . V.unpack
+toPositiveValue =
+  V.pack . fmap (Map.map (\q ->
+    case V.quantity (abs (V.unQuantity q)) of
+      Just absQ -> absQ
+      Nothing   -> error $ "toPositiveValue: abs quantity out of bounds: " <> show (V.unQuantity q)
```
zliu41: Doesn't this mean that tests using `toPositiveValue` may randomly fail? It's better to use `maxBound` in this case.

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -191,32 +257,79 @@ checkBookkeeping v =
...
+      let d = Map [(B c, Map [(B t, I (V.unQuantity quantity))])]
+       in case V.unValueData d of
+            BuiltinFailure{} -> property True
+            _                -> property False
+
+prop_unValueDataValidatesQuantityMin :: Property
+prop_unValueDataValidatesQuantityMin =
+  forAll gen32BytesOrFewer $ \c ->
+    forAll gen32BytesOrFewer $ \t ->
+      let amt = V.minQuantity - 1
```
zliu41: Similarly, you can generate more quantities here. You can also generate a Value with multiple quantities (perhaps 90% valid and 10% invalid), and then check that `unValueData` succeeds if and only if all quantities are valid.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -83,15 +94,76 @@ instance CBOR.Serialise K where
     maybe (fail $ "Invalid Value key: " <> show (B.unpack b)) pure (k b)
   {-# INLINEABLE decode #-}
 
-type NestedMap = Map K (Map K Integer)
+----------------------------------------------------------------------------------------------------
+-- Quantity: Signed 128-bit Integer ----------------------------------------------------------------
+
+-- | Minimum value for a quantity in a Value (signed 128-bit integer minimum).
+minQuantity :: Integer
```
zliu41: Define a `Bounded` instance for `Quantity`, and then you can use `minBound` and `maxBound`.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -175,23 +247,41 @@ empty :: Value
 empty = Value mempty mempty 0 0
 {-# INLINE empty #-}
 
-toList :: Value -> [(K, [(K, Integer)])]
+toList :: Value -> [(K, [(K, Quantity)])]
```
zliu41, line 246: Given we are using a single letter `K` for keys, it would be more consistent to use a single letter - `Q` or `V` for value.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -249,74 +340,75 @@ deleteCoin (UnsafeK -> currency) (UnsafeK -> token) (Value outer sizes size neg)
...
+    Just inner -> unQuantity $ Map.findWithDefault zeroQuantity token inner
 
 {-| \(O(n_{2}\log \max(m_{1}, k_{1}))\), where \(n_{2}\) is the total size of the second
 `Value`, \(m_{1}\) is the size of the outer map in the first `Value` and \(k_{1}\) is
 the size of the largest inner map in the first `Value`.
 
-@a@ contains @b@ if for each @(currency, token, amount)@ in @b@, if @amount > 0@, then
-@lookup currency token a >= amount@, and if @amount < 0@, then
-@lookup currency token a == amount@.
+@a@ contains @b@ if for each @(currency, token, quantity)@ in @b@, if @quantity > 0@, then
```
zliu41: Do you find "quantity" to be a better word than "amount"? Why?

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -249,74 +340,75 @@ deleteCoin (UnsafeK -> currency) (UnsafeK -> token) (Value outer sizes size neg)
...
 valueContains :: Value -> Value -> BuiltinResult Bool
 valueContains v1 v2
   | negativeAmounts v1 > 0 = fail "valueContains: first value contains negative amounts"
   | negativeAmounts v2 > 0 = fail "valueContains: second value contains negative amounts"
-  | otherwise = BuiltinSuccess . getAll $ Map.foldrWithKey' go mempty (unpack v2)
+  | otherwise = BuiltinSuccess . getAll $ Map.foldrWithKey' go (All True) (unpack v2)
  where
-  go c inner = (<>) (Map.foldrWithKey' goInner mempty inner)
+  go :: K -> Map K Quantity -> All -> All
+  go c inner = (<> Map.foldrWithKey' goInner (All True) inner)
```
zliu41: ditto

plutus-core/plutus-core/src/PlutusCore/Value.hs
Unisay: I find `K` confusing. I'd rather use `Key` instead.

plutus-core/plutus-core/src/PlutusCore/Value.hs
Unisay: My goal is to use whichever terminology we chose consistently, and since 
[we already started using](https://github.com/IntersectMBO/plutus/blob/master/plutus-ledger-api/src/PlutusLedgerApi/V3/MintValue.hs#L114-L127) "quantity" to name the same concept in the plutus-ledger-api I decided to apply the same naming in this new builtin-value related functionality too.

Also, ["quantity" is countable, "amount" is un-countable](https://chatgpt.com/share/68f0b471-f2f4-8005-bfe2-449942876f2a).

plutus-core/plutus-core/src/PlutusCore/Value.hs
Unisay: 👌🏼

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
```diff
@@ -91,13 +92,22 @@ conArray uniA = Vector.fromList <$> conList uniA
 -- | Parser for values.
 conValue :: Parser PLC.Value
 conValue = do
-  Value.fromList <$> (traverse validateKeys =<< conList knownUni)
+  keys <- traverse validateKeys =<< conList knownUni
+  case Value.fromList keys of
+    PlutusCore.Builtin.Result.BuiltinSuccess v -> pure v
+    PlutusCore.Builtin.Result.BuiltinSuccessWithLogs _logs v -> pure v
+    PlutusCore.Builtin.Result.BuiltinFailure logs _trace ->
```
effectfully, line 99: Nit: why not just import `BuiltinResult (..)`.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -83,15 +92,72 @@ instance CBOR.Serialise K where
...
+  pretty (UnsafeQuantity i) = pretty i
+
+instance Bounded Quantity where
+  minBound = UnsafeQuantity (-(2 ^ (127 :: Integer)))
+  {-# INLINE minBound #-}
+  maxBound = UnsafeQuantity (2 ^ (127 :: Integer) - 1)
+  {-# INLINE maxBound #-}
+
+-- | Smart constructor for Quantity that validates bounds.
+quantity :: Integer -> Maybe Quantity
```
effectfully, line 135: We should perhaps use `EvaluationResult` instead of `Maybe` as the former is properly strict.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -83,15 +92,72 @@ instance CBOR.Serialise K where
...
+  {-# INLINE encode #-}
+  decode = do
+    i <- CBOR.decode
+    case quantity i of
+      Just q  -> pure q
+      Nothing -> fail $ "Quantity out of signed 128-bit integer bounds: " <> show i
+  {-# INLINEABLE decode #-}
+
+instance Flat.Flat Quantity where
+  encode (UnsafeQuantity i) = Flat.encode i
```
effectfully, line 116: I think we should probably create an appropriate [`ZigZag`](https://hackage.haskell.org/package/flat-0.6/docs/src/Data.ZigZag.html#ZigZag) instance and make use of the fact that we know the maximum width of the data type. It should be a tad faster but I don't know if it actually produces smaller encodings. If it doesn't, then maybe we don't care.

@kwxm do you happen to know?

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -175,23 +241,41 @@ empty :: Value
...
-    . fmap (second (Map.fromListWith (+)))
+fromList :: [(K, [(K, Quantity)])] -> BuiltinResult Value
+fromList xs = do
+  -- Use unchecked addition during construction
+  let outerMap =
+        Map.fromListWith
+          (Map.unionWith unsafeAddQuantity) -- combine inner maps with unchecked addition
+          (second (Map.fromListWith unsafeAddQuantity) <$> xs)
+  -- Validate all quantities are within bounds
+  pack <$> validateNestedMap "fromList" outerMap
```
effectfully: Instead of passing `fromList` explicitly you can do `HasCallStack => <...>`.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -175,23 +241,41 @@ empty :: Value
...
 
+-- | Unsafe addition of quantities without bounds checking.
+unsafeAddQuantity :: Quantity -> Quantity -> Quantity
+unsafeAddQuantity (UnsafeQuantity x) (UnsafeQuantity y) = UnsafeQuantity (x + y)
+{-# INLINE unsafeAddQuantity #-}
+
+-- | Validate all quantities in a nested map are within bounds.
+validateNestedMap :: String -> NestedMap -> BuiltinResult NestedMap
+validateNestedMap context =
+  traverse $ traverse \(UnsafeQuantity i) ->
```
effectfully: You shouldn't use `traverse` as it pointlessly reallocates the entire map. Just check the condition using a fold and return the original map if the check is successful or `fail` otherwise.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -203,39 +287,40 @@ instance Pretty Value where
...
-      (Nothing, _) -> fail $ "insertCoin: invalid currency: " <> show (B.unpack currency)
-      (_, Nothing) -> fail $ "insertCoin: invalid token: " <> show (B.unpack token)
-      (Just ck, Just tk) ->
+insertCoin unsafeCurrency unsafeToken unsafeAmount v@(Value outer sizes size neg)
+  | unsafeAmount == 0 = pure $ deleteCoin unsafeCurrency unsafeToken v
+  | otherwise = case (k unsafeCurrency, k unsafeToken, quantity unsafeAmount) of
+      (Nothing, _, _) -> fail $ "insertCoin: invalid currency: " <> show (B.unpack unsafeCurrency)
+      (_, Nothing, _) -> fail $ "insertCoin: invalid token: " <> show (B.unpack unsafeToken)
+      (_, _, Nothing) -> fail $ "insertCoin: quantity out of bounds: " <> show unsafeAmount
+      (Just currency, Just token, Just qty) ->
```
effectfully, line 304: Have you heard of this thing called `Monad`? :slightly_smiling_face: 

You can just make `k` and `quantity` run in `BuiltinResult` and perform all these checks using tidy do-notation.

Maybe we should distinguish between `currency` and `token` statically too,

Anyway, it's fine as is, just nitpicking.

Although I'm not a fan of the pointlessly lazy `Maybe`.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -249,74 +334,75 @@ deleteCoin (UnsafeK -> currency) (UnsafeK -> token) (Value outer sizes size neg)
...
 {-| \(O(n_{2}\log \max(m_{1}, k_{1}))\), where \(n_{2}\) is the total size of the second
 `Value`, \(m_{1}\) is the size of the outer map in the first `Value` and \(k_{1}\) is
 the size of the largest inner map in the first `Value`.
 
-@a@ contains @b@ if for each @(currency, token, amount)@ in @b@, if @amount > 0@, then
-@lookup currency token a >= amount@, and if @amount < 0@, then
-@lookup currency token a == amount@.
+@a@ contains @b@ if for each @(currency, token, quantity)@ in @b@, if @quantity > 0@, then
+@lookup currency token a >= quantity@, and if @quantity < 0@, then
+@lookup currency token a == quantity@.
```
effectfully: @zliu41 this comment is outdated and wrong, right?

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -249,74 +334,75 @@ deleteCoin (UnsafeK -> currency) (UnsafeK -> token) (Value outer sizes size neg)
...
   | negativeAmounts v2 > 0 = fail "valueContains: second value contains negative amounts"
   | otherwise = BuiltinSuccess . getAll $ Map.foldrWithKey' go mempty (unpack v2)
  where
-  go c inner = (<>) (Map.foldrWithKey' goInner mempty inner)
+  go :: K -> Map K Quantity -> All -> All
+  go c inner = (<> Map.foldrWithKey' goInner mempty inner)
    where
-    goInner t a2 = (<>) (All $ lookupCoin (unK c) (unK t) v1 >= a2)
+    goInner :: K -> Quantity -> All -> All
+    goInner t a2 = (<> All (lookupCoin (unK c) (unK t) v1 >= unQuantity a2))
```
effectfully: If this was `foldr` over lists, you would've turned a short-circuiting fold into a non-short-circuiting one.

However it's a strict `foldr` over `Map`. I.e. it wasn't short-circuiting in any way already. We can make it a lazy `foldr` to make it short-circuit, however it won't be immediate like with lists, because we'd have to reach a leaf first, which isn't too bad given that it's O(log n) work in the unhappy case instead of the current O(n).

I'm not sure if we care, it's not gonna be reflected in costing anyway, because of how hard it would be to do that.

plutus-core/plutus-core/src/PlutusCore/Pretty/PrettyConst.hs
```diff
@@ -162,6 +162,9 @@ instance PrettyBy ConstConfig Data where
 instance PrettyBy ConstConfig Value.K where
     prettyBy config = prettyBy config . Value.unK
 
+instance PrettyBy ConstConfig Value.Quantity where
+    prettyBy config = prettyBy config . Value.unQuantity
+
 instance PrettyBy ConstConfig Value where
     prettyBy config = prettyBy config . Value.toList
```
effectfully, line 169: Not that it matters much, but these should be newtype-derivable.

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -38,60 +38,121 @@ prop_packPreservesInvariants :: V.NestedMap -> Property
...
+    (BuiltinSuccess r1, BuiltinSuccess r2) -> r1 === r2
+    (BuiltinFailure{}, BuiltinFailure{})   -> property True
+    _                                      -> property False
 
 prop_unionAssociative :: Value -> Value -> Value -> Property
 prop_unionAssociative v1 v2 v3 =
-  V.unionValue v1 (V.unionValue v2 v3) === V.unionValue (V.unionValue v1 v2) v3
+  let succeeded (BuiltinSuccess _)           = True
+      succeeded (BuiltinSuccessWithLogs _ _) = True
+      succeeded (BuiltinFailure _ _)         = False
```
effectfully: A.k.a. `not . null`.

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -38,60 +38,121 @@ prop_packPreservesInvariants :: V.NestedMap -> Property
...
+          let v23 = extractValue r23
+              v12 = extractValue r12
+              r1 = V.unionValue v1 v23
+              r2 = V.unionValue v12 v3
+           in case (succeeded r1, succeeded r2) of
+                (True, True)   -> extractValue r1 === extractValue r2
+                (False, False) -> property True
+                -- If one final operation fails and the other succeeds, union is not
+                -- associative due to overflow detection. This is expected and correct.
+                _              -> property ()
```
effectfully: This all seems a tad inconsistent. Why explicitly match on `(False, True)` below and use a `_` here? Why `property True` vs `property ()`?

Anyway, I'm nitpicking again.

effectfully (approved):
LGTM, even though performance can be improved. My comments are mostly nitpicking, feel free to ignore.

I don't like the AI-generated OP, as it's mostly noise. Just the `Summary` part would be enough, the rest is easily inferable from the code changes.

I didn't review the tests fully, will do that another time.

plutus-core/plutus-core/src/PlutusCore/Value.hs
kwxm: > Do you find "quantity" to be a better word than "amount"? Why?

@zliu41 We used the word "quantity" in the original [EUTXO specification](https://github.com/IntersectMBO/plutus/tree/master/doc/extended-utxo-spec) (and see also [this paper](https://iohk.io/en/research/library/papers/utxomautxo-with-multi-asset-support/)).  I haven't checked, but that terminology may also have propagated into the ledger code and specification.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -83,15 +92,72 @@ instance CBOR.Serialise K where
...
+  | i >= unQuantity minBound && i <= unQuantity maxBound = Just (UnsafeQuantity i)
+  | otherwise = Nothing
+{-# INLINEABLE quantity #-}
+
+-- | The zero quantity.
+zeroQuantity :: Quantity
+zeroQuantity = UnsafeQuantity 0
+{-# INLINE zeroQuantity #-}
+
+-- | Safely add two quantities, checking for overflow.
```
kwxm, line 146: I think that `addQuantity` will be sort of non-associative, which might be unexpected.  You can have a situation where `addQuantity (addQuantity a b) c` fails, but `addQuantity a (addQuantity b c)` succeeds, for example if a=2^127-1, b = 3, and c=-5.  I suppose that that's unavoidable, but it might invalidate some potential optimisations.

plutus-core/plutus-core/src/PlutusCore/Pretty/PrettyConst.hs
Unisay: It would require data constructors to be in scope (they aren't).

plutus-core/plutus-core/src/PlutusCore/Value.hs
Unisay: We normally don't encode negative quantities, what problem would such instance solve?

plutus-core/plutus-core/src/PlutusCore/Value.hs
Unisay: 100%, adding quantities is partial and therefore not associative (witnessed by property test failures).

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: Correct, it should be updated.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: Yeah, just use `any`

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -175,23 +241,41 @@ empty :: Value
...
+  pack <$> validateNestedMap "fromList" outerMap
 {-# INLINEABLE fromList #-}
 
+-- | Unsafe addition of quantities without bounds checking.
+unsafeAddQuantity :: Quantity -> Quantity -> Quantity
+unsafeAddQuantity (UnsafeQuantity x) (UnsafeQuantity y) = UnsafeQuantity (x + y)
+{-# INLINE unsafeAddQuantity #-}
+
+-- | Validate all quantities in a nested map are within bounds.
+validateNestedMap :: String -> NestedMap -> BuiltinResult NestedMap
```
zliu41: Call it `validateQuantities`, since this specifically validates the quantities, rather than any other invariants.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: @effectfully I'm not sure if this is what you are saying, but I think we should change

```haskell
(<> All (lookupCoin (unK c) (unK t) v1 >= unQuantity a2))
```

to

```haskell
(<>) (All (lookupCoin (unK c) (unK t) v1 >= unQuantity a2))
```

because `(<>)` is lazy in the _second_ argument, and so `All ...` should be the first argument.

For instance, this does _not_ terminate:

```haskell
foldr (\x -> (|| x > 0)) False [1..]
```

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: I think we are fine. If quantities are within Int64 bounds, then it is practically associative. If you have enormous quantities (which isn't allowed by the ledger anyway), then :shrug:

plutus-core/plutus-core/src/PlutusCore/Value.hs
Unisay: Used `find`: it also avoids re-allocation and returns invalid quantity for the error message.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -175,23 +243,47 @@ empty :: Value
...
 
+-- | Unsafe addition of quantities without bounds checking.
+unsafeAddQuantity :: Quantity -> Quantity -> Quantity
+unsafeAddQuantity (UnsafeQuantity x) (UnsafeQuantity y) = UnsafeQuantity (x + y)
+{-# INLINE unsafeAddQuantity #-}
+
+-- | Validate all quantities in a nested map are within bounds.
+validateQuantities :: HasCallStack => NestedMap -> BuiltinResult NestedMap
+validateQuantities nestedMap =
+  case find isOutOfBounds allQuantities of
```
zliu41, line 273: `any (any isOutOfBounds . M.elems) (M.elems nestedMap)`

plutus-core/plutus-core/src/PlutusCore/Value.hs
effectfully: What you have there is a recursive [function](https://hackage.haskell.org/package/flat-0.6/docs/src/Flat.Decoder.Strict.html#dUnsigned) for both the encoder and the decoder. But you know the width of `Quantity`, you don't need any recursion there, you can do what they do for [`Word64`](https://hackage.haskell.org/package/flat-0.6/docs/src/Flat.Decoder.Strict.html#dWord64) if you don't want to encode negative quantities (which you do, hence you should sprinkle `ZigZag` on top).

I.e. negative quantities are irrelevant, you handle them already anyway and you will keep doing so. The point is that this implementation is not efficient (and is likely not compact either).

plutus-core/plutus-core/src/PlutusCore/Pretty/PrettyConst.hs
effectfully: Oh I see. Yeah I guess that makes sense.

plutus-core/plutus-core/src/PlutusCore/Value.hs
kwxm: > (and is likely not compact either)

I don't think it'll make any difference in terms of size.  Flat encodes natural numbers as a seqence of seven-bit blocks with the final block preceded by a 0 and all of the others preceded by a 1.  If you add the zigzag thing on top I think you're going to need eight bits for numbers in  [-64..63] and sixteen for numbers in [-128..--65] ∪[64..127].

GeorgeFlerovsky:
@Unisay @effectfully @zliu41 
Should the `Value` type in `PlutusLedgerApi` also be changed?
Currently, `PlutusLedgerApi.V1.Value.hs` (re-exported by `V2` and `V3`) defines `Value` like this:
```haskell
newtype Value = Value { getValue :: Map CurrencySymbol (Map TokenName Integer) }
```
https://github.com/IntersectMBO/plutus/blob/7ebd44f7c83ba4909a66ab3466802ad6b9714fd2/plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs#L268C71-L268C78

zliu41:
Yeah we'll do that later, likely after the HF

Issue #7383, opened 2025-10-12
Conform uplc parsing to spec


carloskiki:
From the specification:

> Note. At the time of writing the syntax accepted by IOG’s parser for textual Plutus Core differs slightly from that above in that subobjects of Constr, Map and List objects must not be parenthesised: for example one must write `(con data (Constr 1 [I 2, B #,Map []])`. This discrepancy will be resolved in the near future.

Is it possible to address this shortcoming?

carloskiki:
At least for cases where we have a primitive list of data, for example:
https://github.com/IntersectMBO/plutus/blob/b12c894833cae725bab11577c845701aeb05dc97/plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/listData/listData.uplc#L2

I would expect each data instance in the list to be wrapped in parens. I can understand if there are no parens in the case for `Data`'s `List` or `Map` variants (because we are already parsing a `Data` constant), but for a primitive `list` having `data` not wrapped in parens is inconsistent.

effectfully:
We use Haskell-like syntax for constants and I don't see any reason to be wrapping those subjects into parens.

Would you expect a list of integers to require parens like this: `(con (list integer) [(1), (2), (3)]`?

I don't mind supporting parens there though, why not. Except that somehow seems to make GHC loop lol:

```diff
diff --git plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
index a35178aaa..b68f94e0f 100644
--- plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
+++ plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
@@ -24,7 +24,7 @@ import Data.Text qualified as T
 import Data.Text.Internal.Read (hexDigitToInt)
 import Data.Vector.Strict (Vector)
 import Data.Vector.Strict qualified as Vector
-import Text.Megaparsec (customFailure, getSourcePos, takeWhileP)
+import Text.Megaparsec (customFailure, getSourcePos, takeWhileP, try)
 import Text.Megaparsec.Char (char, hexDigitChar, string)
 import Text.Megaparsec.Char.Lexer qualified as Lex
 
@@ -144,7 +144,7 @@ conBLS12_381_G2_Element = do
 
 -- | Parser for constants of the given type.
[...]

PR #7391, opened 2025-10-19
[Textual] Parse constants with parens

effectfully:
This makes the parse recognize constants with extra parens in them as requested in #7383.

I don't feel like we want to make any of those extra parens mandatory, but there's no harm in being lenient and parsing them.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
@@ -24,7 +24,7 @@ import Data.Text qualified as T
 import Data.Text.Internal.Read (hexDigitToInt)
 import Data.Vector.Strict (Vector)
 import Data.Vector.Strict qualified as Vector
-import Text.Megaparsec (customFailure, getSourcePos, takeWhileP)
+import Text.Megaparsec (customFailure, getSourcePos, takeWhileP, try)
 import Text.Megaparsec.Char (char, hexDigitChar, string)
 import Text.Megaparsec.Char.Lexer qualified as Lex
 
@@ -145,7 +145,7 @@ conBLS12_381_G2_Element = do
 -- | Parser for constants of the given type.
 constantOf :: ExpectParens -> DefaultUni (Esc a) -> Parser a
 constantOf expectParens uni =
-  case uni of
+  try (trailingWhitespace . inParens $ constantOf ExpectParensNo uni) <|> case uni of
     DefaultUniInteger                                                 -> conInteger
     DefaultUniByteString                                              -> conBS
     DefaultUniString                                                  -> conText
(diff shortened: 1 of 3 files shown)


carloskiki:
I'm not sure this is better. I would rather have an inconsistent syntax with a single way of doing things, than a partially consistent syntax with non-canonical ways of doing things.

effectfully:
I don't care either way.

My justification is that for constants we mostly reuse Haskell syntax and there parsing of redundant parens is mostly supported:

```
>>> read "[(True),(False)]" :: [Bool]
[True,False]
>>> read "[((1,2)),((3,4))]" :: [(Int, Int)]
[(1,2),(3,4)]
>>> read "[((1,2),(3,4))]" :: [((Int, Int), (Int, Int))]
[((1,2),(3,4))]
```

If any of the reviewers don't like this PR, feel free to close it.

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
```diff
@@ -145,7 +145,7 @@ conBLS12_381_G2_Element = do
 -- | Parser for constants of the given type.
 constantOf :: ExpectParens -> DefaultUni (Esc a) -> Parser a
 constantOf expectParens uni =
```
zliu41, line 147: I think the `expectParens` parameter is useless now?

zliu41 (approved):

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
effectfully: Nope, we still can't allow `(con data I 42)`, it has to be `(con data (I 42))`.
While `(con (list data) [I 42])` is fine.
So the distinction still exists.

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
```diff
@@ -145,7 +145,7 @@ conBLS12_381_G2_Element = do
...
-  case uni of
+  try (trailingWhitespace . inParens $ constantOf ExpectParensNo uni) <|> case uni of
```
kwxm, line 148: This now accepts 
```
(con (list data) [ (((((((((((I 5))))))))))), B #FF ])
```
and
```
(con unit (((((((((((((((((((()))))))))))))))))))))
```
Do we really want that?

kwxm, line 148: Also, the prettyprinter still prints list elements without parentheses. I suppose we have to choose one way or the other, so that seems fine.

kwxm (commented):
I'm not too sure about the arbitrarily nested parentheses.

Also, we should probably have some conformance tests [here](https://github.com/IntersectMBO/plutus/tree/master/plutus-conformance/test-cases/uplc/evaluation/builtin/constant/data).

PR #7396, opened 2025-11-03
docs(spec): allow optional numeric suffix in Name grammar

### LINKED ISSUE #7365: Minor Discrepancy in Plutus Spec and Implementation: Name Grammar
The lexical grammar specification doesn't match the implementation of `Name`. Names are allowed to include a `-` in the implementation, which is not included in the grammar (Section 2.1.1).

Current Defintion: `[a-zA-Z][a-zA-Z0-9_']*`
Suggested Fix: `[a-zA-Z][a-zA-Z0-9_'-]*` (or similar)

Unisay:
## Summary

Updates the Plutus Core specification to allow names with optional numeric suffixes (e.g., `name-42`, `x-0`, `variable-12345`). This resolves the discrepancy between the formal specification and the parser implementation.

## Changes

- Updated the Name grammar in the lexical specification to permit optional `-[0-9]+` suffix
- Added explanatory paragraph in the Notes section describing the purpose and format of name suffixes
- Verified LaTeX compilation produces correct output

## Motivation

Fixes #7365

The parser implementation has long supported unique name indexes encoded as dash-number suffixes, but the formal grammar specification did not reflect this capability. This created a gap between the documented language and its actual behavior.

## Grammar Change

**Before:**
```
Name ::= [a-zA-Z][a-zA-Z0-9_']*
```

**After:**
```
Name ::= [a-zA-Z][a-zA-Z0-9_']*(-[0-9]+)?
```

## Testing

- LaTeX document compiles successfully
- No implementation changes required (parser already supports this pattern)
- Purely documentation update

### DIFF
--- a/plutus-core/changelog.d/20251103_142610_Yuriy.Lazaryev_issue_7365_name_grammar_spec.md
+++ b/plutus-core/changelog.d/20251103_142610_Yuriy.Lazaryev_issue_7365_name_grammar_spec.md
@@ -1,0 +1,3 @@
+### Changed
+
+- Updated the Plutus Core specification to allow names with optional numeric suffixes (e.g., `name-42`) in #7396, resolving the discrepancy between the formal grammar and the parser implementation.


kwxm (approved):
This looks fine, but I see that there's some pre-existing inconsistency in the grammar:
Grammar

In one place we've got a standard `+` but in other places it's a superscript.  I'm not sure how to resolve this.  Elsewhere in the specification we use `X⁺` (that's a superscript `+`) to denote a nonempty sequence of elements from a set `X`, but the non-superscript version is probably more standard for EBNF-type regular expressions so maybe we should use that here.

I suppose the same goes for `*`.

PR #7398, opened 2025-11-03
Add `ScaleValue` primitive

SeungheonOh:
adds 
```hs
scaleValue :: Integer -> Value -> Value
```

where all of the quantities of the value (second argument) get multiplied by the given constant factor(first argument)


plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -470,3 +476,33 @@ updateSizes old new = dec . inc
...
+    BuiltinSuccess $ Value outer' sizes size neg
+  -- When scaling by negative factor, only need to "flip" negative amounts.
+  | c < 0 = do
+    outer' <- go outer
+    -- let
+    --   neg' = Map.foldl' alg 0 outer'
+    --   alg n inner = n + Map.size (Map.filter (< zeroQuantity) inner)
+
+    -- TODO: make sure (size - neg) is correct
+    BuiltinSuccess $ Value outer' sizes size (size - neg)
```
SeungheonOh, line 490: please check here!

Need to make sure if I can just `size - neg` here or if I need to count all negative quantities again.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
```diff
@@ -300,7 +300,7 @@ instance ArbitraryBuiltin Value.Quantity where
 
 instance Arbitrary Value.Quantity where
     arbitrary = arbitraryBuiltin
-    shrink = shrinkBuiltin
+    shrink = const [] -- shrinkBuiltin
```
SeungheonOh, line 303: `shrinkBuiltin` hangs forever shrinking `Value`. Needs to be fixed

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -188,6 +188,31 @@ prop_containsEnforcesPositivity v
       (BuiltinFailure{}, BuiltinFailure{}) -> property True
       _                                    -> property False
 
+prop_negateIsInverse :: Value -> Property
+prop_negateIsInverse v =
+  let
+    inverseUnion = do
+      vInv <- V.scaleValue (-1) v
```
zliu41, line 225: This will occasionally fail, right? You may want to use `suchThat` to ensure that `v` doesn't contain `minBound :: Quantity`.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: I think this is correct, but it's a good idea to add a `prop_scaleValueBookkeeping` test.

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -188,6 +188,31 @@ prop_containsEnforcesPositivity v
       (BuiltinFailure{}, BuiltinFailure{}) -> property True
       _                                    -> property False
 
+prop_negateIsInverse :: Value -> Property
```
zliu41, line 221: You can add more tests like `prop_negateValueInvolutive` and `prop_scaleBy1IsIdentity`

plutus-core/plutus-core/test/Value/Spec.hs
SeungheonOh: nice catch. Thanks

plutus-core/plutus-core/src/PlutusCore/Value.hs
SeungheonOh: Okay, added bookkeeping property as well. It seems to be fine.

plutus-core/plutus-core/test/Value/Spec.hs
kwxm: > This will occasionally fail, right? You may want to use `suchThat` to ensure that `v` doesn't contain `minBound :: Quantity`.

Can we avoid problems like this by making `minBound` equal to `-(2^127-1)` (ie, `-maxBound`) instead of -2^127?

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -148,6 +149,11 @@ addQuantity :: Quantity -> Quantity -> Maybe Quantity
 addQuantity (UnsafeQuantity x) (UnsafeQuantity y) = quantity (x + y)
 {-# INLINEABLE addQuantity #-}
 
+-- | Safely scale a quantitie with given integer, checking for overflow.
```
kwxm:
```suggestion
-- | Safely scale a quantity with given integer, checking for overflow.
```

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -470,3 +476,28 @@ updateSizes old new = dec . inc
       then id
       else IntMap.update (\n -> if n <= 1 then Nothing else Just (n - 1)) old
 {-# INLINEABLE updateSizes #-}
+
+-- | \(O(n)\). Scale each tokens by the given constant factor.
+scaleValue :: Integer -> Value -> BuiltinResult Value
+scaleValue c (Value outer sizes size neg)
```
kwxm, line 482: If the `Value` is nonempty then it must contain at least one nonzero `Quantity` `q`.  If you're scaling by the integer `n` and it's outside the allowable range for `Quantity` then `n*q` will also lie outside the allowable range and so `scaleValue` is bound to fail (although I'm wondering if `maxBound` or `minBound` might cause problems).  You could check at the start and fail immediately if `n` is too big or small; of course even if `n` is in range you might still get an overflow for some particular quantity.  

I don't know if it's actually worthwhile checking this up front, but the point still stands that `scaleValue` shoud only succeed for integers of size 2 or less, so we can probably ignore the size of `n` while doing costing and just base the cost on the total number of tokens.

plutus-core/plutus-core/src/PlutusCore/Value.hs
kwxm: I think that's right.

plutus-core/plutus-core/test/Value/Spec.hs
kwxm: Weird things might also happen if the scalar is close to the upper or lower bound for `Quantity`.  I think that [ArbitraryBuiltin for Integer](https://github.com/IntersectMBO/plutus/blob/8444bd00cc605d6e753aa617891897ce4b52cc66/plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs#L152) will generate values like that, or you could do it yourself.

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -188,6 +188,51 @@ prop_containsEnforcesPositivity v
       (BuiltinFailure{}, BuiltinFailure{}) -> property True
       _                                    -> property False
 
+scaleCorrectlyBound :: Integer -> Value -> Bool
+scaleCorrectlyBound factor val =
```
kwxm: The name's confusing me a bit here.  Isn't this returning `False` if anything overflows?

plutus-core/plutus-core/test/Value/Spec.hs
kwxm: Maybe also test that scaling by zero returns the empty value?

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -470,3 +476,28 @@ updateSizes old new = dec . inc
       then id
       else IntMap.update (\n -> if n <= 1 then Nothing else Just (n - 1)) old
 {-# INLINEABLE updateSizes #-}
+
+-- | \(O(n)\). Scale each tokens by the given constant factor.
```
kwxm:
```suggestion
-- | \(O(n)\). Scale each token by the given constant factor.
```

plutus-core/plutus-core/src/PlutusCore/Value.hs
SeungheonOh: Checking `minBound <= n <= maxbound` for n sounds useful. I'll add.

> the point still stands that scaleValue shoud only succeed for integers of size 2 or less

I'm not entirely sure "size 2" means here.

kwxm: > I'm not entirely sure "size 2" means here.

Oh, sorry.  The cost of a builtin is based on the sizes of its inputs, and usually the sizes are measured in units of 64-bit words (I really wish it was bytes, but we got stuck with words a long time ago).  This applies to the standard size measure for `Integer`, and a 128-bit integer fits into two words so the mulltiplier has size at most two here.

SeungheonOh: Thinking again, checking `n` upfront might not be as useful as I originally thought it would be. Since checking validity of `n` won't allow us to skip range check on `scaleQuantity` later on, this feels like extraneous work. Also, when `n` is out of bounds, it will fail right away on the first value it encounters on the nested map anyways.

kwxm: > Also, when `n` is out of bounds, it will fail right away on the first value it encounters on the nested map anyways.

Good point.  I withdraw my suggestion!

plutus-core/plutus-core/test/Value/Spec.hs
zliu41: > Can we avoid problems like this by making minBound equal to -(2^127-1) (ie, -maxBound) instead of -2^127?

I think that's more problematic (i.e., weird) in a different way.

plutus-core/testlib/PlutusCore/Generators/QuickCheck/Builtin.hs
zliu41: Please make an issue

SeungheonOh: made an issue: https://github.com/IntersectMBO/plutus/issues/7404

SeungheonOh: I think we can keep `const []` in the meantime so that properties don't randomly hang

zliu41: Sure, as long as we have an issue to capture this.

SeungheonOh:
For some reason, goldens changed ([a577ce9](https://github.com/IntersectMBO/plutus/pull/7398/commits/a577ce999aea6fc36e42f2fdb881753cd9ac01a8)). I'm trying to figure out why. 

Update:
I think somehow number of CSE passes changes. I'm not sure how having extra builtin changes CSE pass count. Any idea?

zliu41:
I don't think it's the number of CSE passes - unless of course you have clear evidence of that. I think it's likely that the additional definitions in `plutus-tx-plugin` caused certain things to be ordered slightly differently or certain optimizations to work slightly differently (IIRC the initial PIR contains all those definitions, and the redundant ones are then optimized away). This has happened before and is not entirely surprising. I wouldn't spend too much time on it.

SeungheonOh:
Okay. I saw some cse term not being pulled out and thought it was something to do with CSE. But, in that case, I'll ignore and merge when CI passes.

PR #7400, opened 2025-11-04
SOP vs Scott bench

SeungheonOh:
Compares SOP encoded values and Scott encoded values in terms of their ExBudget usage. 

[result](https://github.com/IntersectMBO/plutus/blob/sho/benchScottVsSop/plutus-benchmark/sop/README.md)


SeungheonOh:
As this is only an experimental benchmark and is never meant to be merged. I will close this.

PR #7402, opened 2025-11-04
NegateValue vs ScaleValue

SeungheonOh:
compares `NegateValue` and `ScaleValue` builtin. 

I benchmarked two implementation of `NegateValue`. First, with `(-1) * quantity` other with `negate quantity` and two results differ significantly.

This is result is from having `(-1) * quantity`

negatevsscalevalue

This one is with `negate quantity`

negatevsscalevalue-betternegate

where 

```hs
negateQuantity :: Quantity -> Maybe Quantity
negateQuantity (UnsafeQuantity x) =
  if x == -170141183460469231731687303715884105728
  then Nothing
  else Just $ UnsafeQuantity (negate x)
```

I didn't know `negate` would perform so much better than `(-1) *`


SeungheonOh:
negatevsscalevalue-betternegate-test

This is with slower `negateValue`:
```hs
negateQuantity :: Quantity -> Maybe Quantity
negateQuantity (UnsafeQuantity x) = quantity (negate x)
{-# INLINE negateQuantity #-}
```

zliu41:
Does any of these graphs compare two implementations that are otherwise exactly the same, except one uses `negateQuantity`, and the other uses `scaleQuantity`?

You seem to be comparing `negateValue` and `scaleValue`, but there might be other differences between them other than negation vs. multiplication?

zliu41:
It would also be nice to verify whether `0 - x` is indeed that much faster than `(-1) * x` (a simple criterion benchmark would do). Because according to their cost models, the differences aren't that big.

zliu41:
In any case, I think we should still add `scaleValue` instead of `negateValue`. We can always optimize for special cases like 1 or -1 later.

SeungheonOh:
> Does any of these graphs compare two implementations that are otherwise exactly the same, except one uses negateQuantity, and the other uses scaleQuantity?

Yes!

SeungheonOh:
We got all we needed from benchmark!

PR #7406, opened 2025-11-06
Add Enum Ratio to PlutusTx stdlib.

bezirg:
Added an Enum Rational instance that mimicks that of GHC and Haskell report.

I also removed the toGHC/fromGHC from plutustx since they don't work on-chain.
I think it is straightforward to re-create these two-functions for off-chain use, simply by a combination of numerator and denominator.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-tx/src/PlutusTx/Ratio.hs
+++ b/plutus-tx/src/PlutusTx/Ratio.hs
@@ -31,10 +31,11 @@ module PlutusTx.Ratio (
   recip,
   abs,
   negate,
-  half,
-  fromGHC,
-  toGHC,
   gcd,
+
+  -- * Conversion from/to Haskell
+  fromHaskellRatio,
+  toHaskellRatio,
 ) where
 
 import PlutusTx.Applicative qualified as P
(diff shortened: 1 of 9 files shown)


bezirg:
> Changes look good, but I think it's okay to keep `fromGHC` and `toGHC` as is. We have other helpers in plutus-tx and plutus-ledger-api that won't work on-chain but is quite useful.

I thought we are trying to keep plutus-tx on-chain only. But I am open to discussions about this and if so , bring it back.
If it is the case to keep it then I propose to change the name `fromGHC` to something more generic like `fromHS` or `fromHaskell`.

SeungheonOh:
> I thought we are trying to keep plutus-tx on-chain only.

I was mostly about keeping `plutus-core` out from packages that defines plinth types. I don't have a firm opinion on things that interfaces with regular Haskell as they don't often add any new dependencies anyways

bezirg:
I put the fromGHC / toGHC back and renamed them to from/toHaskellRatio

plutus-tx/test/Rational/Laws/Construction.hs
```diff
@@ -1,18 +1,24 @@
 -- editorconfig-checker-disable-file
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TupleSections     #-}
+{-# LANGUAGE NoImplicitPrelude     #-}
```
SeungheonOh: indentation

plutus-tx/src/PlutusTx/Ratio.hs
```diff
@@ -375,7 +365,53 @@ euclid x y
   | P.True = euclid y (x `Builtins.modInteger` y)
 {-# INLINEABLE euclid #-}
 
-$(makeLift ''Rational)
+instance Enum Rational where
+  {-# INLINEABLE succ #-}
+  succ (Rational n d) = Rational (n `Builtins.addInteger` d) d
```
SeungheonOh: You can just use `P.+` I think

PR #7407, opened 2025-11-06
Conformance tests for value built-ins taking into account 128-bit integers

basetunnel:


zliu41 (approved):

basetunnel:
> Can we have ones for `ScaleValue` as well?

I've added them, please check

PR #7409, opened 2025-11-07
Additional conformance tests for insertCoin and valueContains

basetunnel:
- Tests the following condition of `insertCoin` with a key of size > 32 bytes
> Both BuiltinCurrencySymbol and BuiltinTokenName must be no longer than 32 bytes **(unless the amount is zero).**

- Tests that `valueContains` fails when the first argument has negative value, even if the second value is empty. (also renamed existing `empty` test to `pos-empty`)


kwxm (approved):
I'd suggest also adding a test that it actually fails when the key is too big and you're inserting a nonzero quantity, and also tests for what happens when the second key is too big and the quantity is zero and nonzero.

basetunnel:
> I'd suggest also adding a test that it actually fails when the key is too big and you're inserting a nonzero quantity, and also tests for what happens when the second key is too big and the quantity is zero and nonzero.

Thanks, I've added these tests. They are called `key-too-long-1`, `key-too-long-2` (first and second key, inserting >0 quantity) and `long-key-zero-1` and `long-key-zero-2` (first and second key, inserting 0 quantity)

kwxm:
Thanks @basetunnel.  That all looks good, so I think we can merge it now.

Issue #7401, opened 2025-11-04
Standardize repetition notation in grammar specifications (EBNF)


Unisay:
### Problem

The Plutus Core specification contains inconsistent notation for repetition patterns in the lexical grammar (Figure 1 in untyped-grammar.tex):

- Line 19 (Name): Uses `\textsuperscript{*}` for zero-or-more repetition
- Line 23 (Version): Uses `\textsuperscript{+}` for one-or-more repetition  
- Line 24 (Natural): Uses standard `+` (no textsuperscript) for one-or-more repetition

### Context

The specification uses two different notational systems:

1. **Mathematical set notation**: $X^*$ and $X^+$ (defined in notation.tex:21-27) represent finite sequences and non-empty finite sequences respectively
2. **EBNF grammar notation**: Standard regex-style `*` and `+` for repetition patterns in formal grammars

The lexical grammar section (Figure 1) is an EBNF-style grammar, not mathematical set notation, yet it mixes both conventions.

### Proposed Solution

Standardize the lexical grammar to use consistent **standard EBNF notation** (non-superscript):

```latex
\textrm{Name}     & n  & ::= & \texttt{[a-zA-Z][a-zA-Z0-9\_\textquotesingle]*}   & \textrm{name}\\
\textrm{Version}  & v  & ::= & \texttt{[0-9]+.[0-9]+.[0-9]+}                     & \textrm{version}\\
\textrm{Natural}  & k  & ::= & \texttt{[0-9]+}                                   & \textrm{a natural number}\\
```

**Rationale:**
- EBNF grammars conventionally use standard `*` and `+` operators
- The mathematical superscript notation ($X^*$, $X^+$) should remain for set-theoretic contexts (as in notation.tex)
- This aligns with standard grammar notation across computer science literature
- Improves readability for readers familiar with EBNF/regex syntax

### Related

- Identified in PR #7396 review: https://github.com/IntersectMBO/plutus/pull/7396#pullrequestreview-3414692947
- Issue #7365 (which PR #7396 addresses) only updated Name grammar, leaving Version inconsistent

### Files to Update

- `doc/plutus-core-spec/untyped-grammar.tex` (lines 19, 23)

basetunnel:
I've opened #7414 that implements the suggested change.

basetunnel:
Closed with #7414

PR #7414, opened 2025-11-13
Standardize repetition notation in EBNF grammar of plutus specificati…

basetunnel:
Implements the suggested change in #7401 .


Unisay:
Thank you!

PR #7417, opened 2025-11-13
Apply fourmolu to the codebase

zeme-wana:
Our custom fork of fourmolu has the following behaviours which are different than upstream: 
### Multi-line Haddock has the closing `-}` on the last line, not on a separate newline.
```
{-| Make a 'NameInfo' mapping the given set of TH names to their
'GHC.TyThing's for later reference. 
-}
```
-->
```
{-| Make a 'NameInfo' mapping the given set of TH names to their
'GHC.TyThing's for later reference. -}
```
### No space before `|` or `^` in multiline haddock:
```
{- | Make a 'NameInfo' mapping the given set of TH names to their
'GHC.TyThing's for later reference. -}
```
--> 
```
{-| Make a 'NameInfo' mapping the given set of TH names to their
'GHC.TyThing's for later reference. -}
```
### No newline between comments and haddock
```
-- Normally fourmolu would add a newline after this
{- | Make a 'NameInfo' mapping the given set of TH names to their
'GHC.TyThing's for later reference. -}
```
--> 
```
-- Normally fourmolu would add a newline after this
{-| Make a 'NameInfo' mapping the given set of TH names to their
'GHC.TyThing's for later reference. -}
```

### DIFF
--- a/plutus-tx-plugin/test/Plugin/Coverage/9.6/coverageCode.golden.pir
+++ b/plutus-tx-plugin/test/Plugin/Coverage/9.6/coverageCode.golden.pir
@@ -11,7 +11,7 @@ let
           (/\dead ->
              trace
                {all dead a. Maybe a}
-               "CoverLocation (CovLoc {_covLocFile = \"test/Plugin/Coverage/Spec.hs\", _covLocStartLine = 42, _covLocEndLine = 42, _covLocStartCol = 26, _covLocEndCol = 33})"
+               "CoverLocation (CovLoc {_covLocFile = \"test/Plugin/Coverage/Spec.hs\", _covLocStartLine = 42, _covLocEndLine = 42, _covLocStartCol = 8, _covLocEndCol = 15})"
                (/\dead -> Nothing)
                {all dead. dead})
           {all dead. dead}
@@ -107,11 +107,11 @@ in
     (/\dead ->
        trace
          {all dead. Maybe bool}
-         "CoverLocation (CovLoc {_covLocFile = \"test/Plugin/Coverage/Spec.hs\", _covLocStartLine = 40, _covLocEndLine = 42, _covLocStartCol = 1, _covLocEndCol = 33})"
+         "CoverLocation (CovLoc {_covLocFile = \"test/Plugin/Coverage/Spec.hs\", _covLocStartLine = 40, _covLocEndLine = 42, _covLocStartCol = 1, _covLocEndCol = 15})"
          (/\dead ->
             trace
               {all dead. Maybe bool}
-              "CoverLocation (CovLoc {_covLocFile = \"test/Plugin/Coverage/Spec.hs\", _covLocStartLine = 40, _covLocEndLine = 42, _covLocStartCol = 9, _covLocEndCol = 33})"
+              "CoverLocation (CovLoc {_covLocFile = \"test/Plugin/Coverage/Spec.hs\", _covLocStartLine = 40, _covLocEndLine = 42, _covLocStartCol = 9, _covLocEndCol = 15})"
               (/\dead ->
                  Maybe_match
                    {integer}


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -530,195 +558,205 @@ But in our case this is okay, because:
...
-throwErrorDischarged
-  :: ThrowableBuiltins uni fun
-  => EvaluationError (MachineError fun) CekUserError
-  -> CekValue uni fun ann
-  -> CekM uni fun s x
+throwErrorDischarged ::
+  ThrowableBuiltins uni fun =>
+  EvaluationError (MachineError fun) CekUserError ->
+  CekValue uni fun ann ->
+  CekM uni fun s x
```
SeungheonOh: I prefer having arrows in the beginning of the line. I think there was some option in fourmolu to force this.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -757,110 +795,112 @@ runCekM
...
+
+    -- \| The returning phase of the CEK machine.
+    --    Returns 'EvaluationSuccess' in case the context is empty, otherwise pops up one frame
+    --    from the context and uses it to decide how to proceed with the current value v.
+    --
+    --      * 'FrameForce': call forceEvaluate
+    --      * 'FrameApplyArg': call 'computeCek' over the context extended with 'FrameApplyFun'
+    --      * 'FrameApplyFun': call 'applyEvaluate' to attempt to apply the function
+    --          stored in the frame to an argument.
+    --
```
SeungheonOh, line 878: I think `{- | ... -}` is better for multiline haddock. 

I think `haddock-style: multi-line` will do this.

SeungheonOh (commented):
Just few observations.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
zeme-wana: Done, but note that using `multi-line` or `multi-line-compact`, the ending `-}` is always put on a separate newline: I'm unsure if this behaviour is preferred.

cardano-constitution/src/Cardano/Constitution/Data/Validator.hs
```diff
@@ -1,26 +1,28 @@
...
-    ( module Export
-    , defaultValidators
-    , defaultValidatorsWithCodes
-    ) where
+
+module Cardano.Constitution.Data.Validator (
+  module Export,
+  defaultValidators,
+  defaultValidatorsWithCodes,
+) where
```
Unisay: Change from leading commas to trailing? 

Ok, I personally prefer leading, but don't mind trailing in the case of imports.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
basetunnel: Just curious, @SeungheonOh why do you prefer `{- | ...}` over `--| ...` for multi-line haddock? I don't have a strong opinion, but I think `-- |` is more common.

basetunnel:
I don't have objections to using fourmolu, there's a few things that I haven't seen often, such as `{- | -}` for multi-line haddock and aligning `let`/`in` like this:

```
let ...
 in ...
```

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
SeungheonOh: I like it because I don't have to add `--` on every newline.

cardano-constitution/src/Cardano/Constitution/Data/Validator.hs
zeme-wana: I've switched back to leading

PR #7422, opened 2025-11-15
Use correct "golden.eval" postfix for workflow and script

SeungheonOh:
I don't recall this name change. I think `.eval.golden` make more sense. Either way, this PR fixes "Compare Golden Budget" workflow in the meantime.

### DIFF
--- a/scripts/eval-percent-diff.sh
+++ b/scripts/eval-percent-diff.sh
@@ -51,7 +51,7 @@ if [[ $TFLAG -eq 1 ]]; then
     file=""; seen=0;
   }
   # new file header: reset seen flag, store filename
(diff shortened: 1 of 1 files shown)


Unisay:
One aspect to take into account:
Often syntax highlighters (including the UPLC one) are triggered by a file extension, i.e. `foo.golden.uplc` will be highlighted but `foo.uplc.golden` won't! This way uplc extension works like a media-type.
In case of the `foo.golden.eval` - `eval` isn't a recognized extension but IMO it makes sense to keep the same order for consistency.

SeungheonOh:
ah!, yes that was the reason. I remember now. In that case, we can just keep as is

PR #7432, opened 2025-11-19
Make the Tx.&& and Tx.|| short-circuit also in Haskell

bezirg:
This does not affect UPLC code, which should be already short-circuiting. This is about Haskell offchain code that uses the PlutusTx library. The PR provides a simple unit test.


bezirg:
Can somebody explain to me why this change affects the uplc code generation? See the failing tests

SeungheonOh:
It's probably because GHC changing up something because arguments being lazy. I'll take a look today

SeungheonOh:
Okay, none of the tests actually fails, just the golden files need to be updated. Interestingly, this change makes script inline the definitions of equality in multiple places, making it less efficient. I doubt there's anything we can do to actually fix it because at plugin level we don't have enough control

bezirg:
> Okay, none of the tests actually fails, just the golden files need to be updated. Interestingly, this change makes script inline the definitions of equality in multiple places, making it less efficient. I doubt there's anything we can do to actually fix it because at plugin level we don't have enough control

thanks for updating the golden!

SeungheonOh:
actually, it's within 0.5% increase in MEM and CPU. Maybe it's not that big of a deal if we just take this changes.

bezirg:
> actually, it's within 0.5% increase in MEM and CPU. Maybe it's not that big of a deal if we just take this changes.

agreed

bezirg:
freaking verified signatures on the commits, they break all the time. I have to squash sorry.

SeungheonOh (approved):

PR #7435, opened 2025-11-19
Value costing: insertCoin, unionValue, scaleValue

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1899
Fixes https://github.com/IntersectMBO/plutus-private/issues/1900
Fixes https://github.com/IntersectMBO/plutus-private/issues/2004

See https://plutus.cardano.intersectmbo.org/pr-preview/cost-models/pr-7435/ for cost model vizualizations

### DIFF
--- a/plutus-core/cost-model/print-cost-model/Main.hs
+++ b/plutus-core/cost-model/print-cost-model/Main.hs
@@ -152,7 +152,7 @@ renderModel =
     LinearInX f -> [renderLinearFunction f "x"]
     LinearInY f -> [renderLinearFunction f "y"]
     LinearInZ f -> [renderLinearFunction f "z"]
-    LinearInW f -> [renderLinearFunction f "w"]
+    LinearInU f -> [renderLinearFunction f "u"]
     QuadraticInY f -> [renderOneVariableQuadraticFunction f "y"]
     QuadraticInZ f -> [renderOneVariableQuadraticFunction f "z"]
     QuadraticInXAndY f -> [renderTwoVariableQuadraticFunction f "x" "y"]
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/SimpleJSON.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/SimpleJSON.hs
@@ -110,7 +110,7 @@ data Model
   | LinearInX LinearFunction
   | LinearInY LinearFunction
   | LinearInZ LinearFunction
-  | LinearInW LinearFunction
+  | LinearInU LinearFunction
   | LiteralInYOrLinearInZ LinearFunction
   | LinearInMaxYZ LinearFunction
   | LinearInYAndZ TwoVariableLinearFunction
@@ -158,7 +158,7 @@ instance FromJSON Model where
           "linear_in_x" -> LinearInX <$> parseJSON args
           "linear_in_y" -> LinearInY <$> parseJSON args
           "linear_in_z" -> LinearInZ <$> parseJSON args
-          "linear_in_u" -> LinearInW <$> parseJSON args
+          "linear_in_u" -> LinearInU <$> parseJSON args
           "quadratic_in_y" -> QuadraticInY <$> parseJSON args
           "quadratic_in_z" -> QuadraticInZ <$> parseJSON args
           "quadratic_in_x_and_y" -> QuadraticInXAndY <$> parseJSON args


ana-pantilie:
## InsertCoin

Screenshot 2025-11-19 at 23 10 13

f$t - time
f$u - size of Value
Black circles - amount inserted is zero (so insertCoin behaves like deleteCoin)
Red circles - amount inserted is maxBound

ana-pantilie:
## UnionValue

### Linear model in sum of total sizes: not very good.

Screenshot 2025-11-20 at 19 54 09
Screenshot 2025-11-20 at 19 57 46

-------

### Quadratic model in square of sum of total sizes: pretty good!

Screenshot 2025-11-20 at 19 55 14
Screenshot 2025-11-20 at 19 55 30
Screenshot 2025-11-20 at 19 58 04

ana-pantilie:
## ScaleValue

### With randomized quantities and scalars and randomized total size of value

Screenshot 2025-11-21 at 22 59 18

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/SimpleJSON.hs
```diff
@@ -93,12 +106,14 @@ data Model
...
   | LinearInY LinearFunction
   | LinearInZ LinearFunction
+  | LinearInW LinearFunction
   | LiteralInYOrLinearInZ LinearFunction
   | LinearInMaxYZ LinearFunction
   | LinearInYAndZ TwoVariableLinearFunction
   | QuadraticInY OneVariableQuadraticFunction
   | QuadraticInZ OneVariableQuadraticFunction
   | QuadraticInXAndY TwoVariableQuadraticFunction
+  | SquareOfSum SquareOfTwoVariableSumFunction
```
zliu41: This most likely won't be needed - `a + bx + cy` should be good enough.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -212,6 +216,143 @@ unValueDataBenchmark :: StdGen -> Benchmark
...
+  where
+    entries =
+      [ ( pId
+        , [ (tName, amt)
+          | tName <- tokenNames
+          ]
+        )
+      | pId <- policyIds
+      ]
+    amt = mkQuantity 10000000
```
zliu41: I don't think we need to use smaller integers at all - just always use `maxBound / 2`.

ana-pantilie: I keep getting errors when using integers close to the maxBound, not sure why yet. This is the same issue I ran into when working on the `scaleValue` generator. These errors appear despite taking precautions to not allow the additions/multiplications to overflow. Since for `unionValue` the size of the integers doesn't matter I have postponed the investigation for a bit later, but it's on my TODO list.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/SimpleJSON.hs
ana-pantilie: The `SquareOfSum` stuff is from before the recent experimentation for `unionValue`, so it's outdated.

> a + bx + cy should be good enough.

I would assume `a`, `b` and `c` are the coefficients we are trying to determine while `x` and `y` are the sizes of the inputs, right? Then, as far as I understand, this translates to running R's linear regression algorithm for the `t ~ x + y` formula. @kwxm am I understanding what @zliu41 is suggesting correctly?

ana-pantilie: Btw, just to be sure, `I(x + y)` means "find a coefficient for (x + y)" while `x + y` means "find a coefficient for x and a coefficient for y", right?

plutus-core/cost-model/data/builtinCostModelC.json
```diff
@@ -1272,5 +1272,32 @@
...
+      "arguments": {
+        "interaction_slope_x": 114467,
+        "interaction_slope_xy": 7,
+        "interaction_slope_y": 160561
+      },
+      "type": "with_interaction_of_x_and_y"
+    },
+    "memory": {
+      "arguments": 1,
+      "type": "constant_cost"
```
ana-pantilie: Is this supposed to be changed somehow? @kwxm

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -181,6 +181,9 @@ builtinMemoryModels =
     , paramValueContains = Id $ ModelTwoArgumentsConstantCost 32
     , paramValueData = Id $ ModelOneArgumentConstantCost 32
     , paramUnValueData = Id $ ModelOneArgumentConstantCost 32
+    , paramInsertCoin = Id $ ModelFourArgumentsConstantCost 1
+    , paramUnionValue = Id $ ModelTwoArgumentsConstantCost 1
+    , paramScaleValue = Id $ ModelTwoArgumentsConstantCost 1
```
ana-pantilie: Put something here!
- insertCoin is linear in its value argument
- unionValue is linear in (arg1 + arg2)
- scaleValue is linear in its value argument

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -261,6 +262,10 @@ fromList xs = do
   pack <$> validateQuantities outerMap
 {-# INLINEABLE fromList #-}
 
+-- | Left biased unsafe fromList.
+unsafeFromList :: [(K, [(K, Quantity)])] -> Value
```
zliu41: This is only useful for benchmarking so it should be placed in the benchmarking code.

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -1963,12 +1963,18 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
           bls12_381_G2_multiScalarMulDenotation
           (runCostingFunTwoArguments . paramBls12_381_G2_multiScalarMul)
   toBuiltinMeaning _semvar InsertCoin =
-    let insertCoinDenotation :: ByteString -> ByteString -> Integer -> Value -> BuiltinResult Value
-        insertCoinDenotation = Value.insertCoin
+    let insertCoinDenotation
+          :: ByteString
+          -> ByteString
+          -> Integer
+          -> ValueLogOuterSizeAddLogMaxInnerSize
```
zliu41: Please rename this type as @kwxm suggested.

plutus-core/cost-model/data/builtinCostModelC.json
```diff
@@ -1272,5 +1272,55 @@
...
+        "slopex": 172116,
+        "slopexy": 6,
+        "slopey": 183150
+      },
+      "type": "with_interaction_in_x_and_y"
+    },
+    "memory": {
+      "arguments": {
+        "intercept": 24,
+        "slope": 21
```
zliu41, line 1305: How are the slope and intercept determined?

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -2006,12 +2012,12 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
           unValueDataDenotation
           (runCostingFunOneArgument . paramUnValueData)
   toBuiltinMeaning _semvar ScaleValue =
-    let unValueDataDenotation :: Integer -> Value -> BuiltinResult Value
-        unValueDataDenotation = Value.scaleValue
+    let unValueDataDenotation :: Integer -> ValueTotalSize -> BuiltinResult Value
```
zliu41: unValueDataDenotation -> scaleValueDenotation

plutus-core/cost-model/data/builtinCostModelC.json
ana-pantilie: See the [note](https://github.com/IntersectMBO/plutus/blob/d529ba8daddb02d10e3b89954840f08568dd2e9b/plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs#L67).

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -257,16 +363,21 @@ generateConstrainedValueWithMaxPolicy
   -> Int -- Number of tokens per policy
   -> g
   -> m (Value, K, K) -- Returns (value, maxPolicyId, deepestTokenInMaxPolicy)
-generateConstrainedValueWithMaxPolicy numPolicies tokensPerPolicy g = do
+generateConstrainedValueWithMaxPolicy numPolicies tokensPerPolicy =
+  generateConstrainedValueWithMaxPolicyAndQuantity numPolicies tokensPerPolicy maxBound
+
+-- | Generate constrained Value with information about max-size policy and quantity
+generateConstrainedValueWithMaxPolicyAndQuantity
```
zliu41: Did you intend to use this somewhere, perhaps in `scaleValueArgs`? Otherwise there's not much point factoring this out.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -64,6 +64,60 @@ mlResultMemSize = toMemSize Pairing.mlResultMemSizeBytes
...
+reference it.
+An invariant of 'Value' is that the 'IntMap' contains exactly as many keys as there are unique
+sizes of inner maps in the nested 'Map'. In the worst-case scenario we are considering here,
+each inner map has the same size of 1, so the memory footprint of the 'IntMap' will be constant:
+  - 1 word for the 'Tip' constructor
+  - 2 words for the key pointer and value pointer
+  - 2 words for the key 'Int' closure
+  - 2 words for the value 'Int' closure
+Putting this all together, we arrive at the following memory usage function:
+    1*n + 4 + 7 + 1
```
zliu41: It's unclear what the "4", "7" and "1" refer to. Please make it more obvious.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -64,6 +64,60 @@ mlResultMemSize = toMemSize Pairing.mlResultMemSizeBytes
...
+This bookeeping data consists of two unboxed 'Int's and an 'Data.IntMap.IntMap Int'.
+At runtime, this results in the following memory layout:
+  - 1 word for the 'Value' constructor
+  - 1 word for each of the two unboxed 'Int's
+  - 1 word for the pointer to the 'IntMap'
+  - 1 word for the pointer to the nested 'Map'
+  ... and the allocations for the 'IntMap' and nested 'Map' themselves.
+The 'Map' and 'IntMap' are implemented as a self-balancing binary tree and a big-endian patricia tree,
+respectively. Since we are interested in the worst-case memory usage, we assume that
+for each application of the builtin function both of the trees have to be completely reallocated.
```
zliu41: This only applies to `unionValue`, since in `insertCoin` and `scaleValue`, there's only one tree.

Also, `insertCoin` never completely allocates the whole tree, even in the worst case.

plutus-core/cost-model/data/builtinCostModelC.json
```diff
@@ -1272,5 +1272,55 @@
...
+        "slope": 18413
+      },
+      "type": "linear_in_w"
+    },
+    "memory": {
+      "arguments": {
+        "intercept": 45,
+        "slope": 21
+      },
+      "type": "linear_in_w"
```
zliu41: I'd still like to know if `insertCoin` is CPU bound or memory bound.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -64,6 +64,60 @@ mlResultMemSize = toMemSize Pairing.mlResultMemSizeBytes
...
+Since we are dealing with a nested 'Map' with the structure 'Map CurrencySymbol (Map TokenName Integer)',
+we can consider only the worst-case scenario where the map is flattened, i.e., each currency symbol maps
```
zliu41: I don't know why this is the worst case memory-wise, and I highly doubt this is the worst case for `insertCoin`. When inserting a node into a Map, usually the deeper the Map is, the more allocation is needed. I imagine a deep Map also makes rebalancing trickier.

ana-pantilie: > I don't know why this is the worst case memory-wise, and I highly doubt this is the worst case for insertCoin.

It's not, I was mistaken. @kwxm, @Unisay and I had a meeting a few days ago where we discussed this. I think the structure of the map doesn't matter when it comes to how much memory is allocated (for one map), only the number of nodes does. But that doesn't really matter much because we care about the new memory being allocated after the builtin run, not the total memory.

> I imagine a deep Map also makes rebalancing trickier.

You're probably right, but it's tricky to figure out what the worst case is and how much memory is allocated.

I think we need to approach memory costing for the `Value` builtins similar to the way we approach cpu costing: by experimentally finding the worst case and generating the data which we can base predictions on. I don't think we currently do this for the memory costing of any of the other builtins, so we need to also add the required code infrastructure to generate the models using R.

At the same time, @kwxm said we gave up on doing accurate memory costing a while ago, but I don't know the reason behind this.

plutus-core/cost-model/data/builtinCostModelC.json
ana-pantilie: I don't know, I haven't had time to check this yet.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
ana-pantilie: I think this is a remnant of the experimentation which happened on this branch. I'll revert the change it if I have time.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
ana-pantilie: They come from me drawing out the closures from my understanding of ghc's memory allocation and counting the number of words (which are mentioned in the note). I don't think this methodology is the best, so we might end up removing this note in the end.

zliu41: The note is useful, just organize it better and make it easier to follow. There's a lot of "1 word for this", "2 words for that", etc etc., and it is non-obvious what the "4" "7" "1" correspond to.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
zliu41: It is almost always sufficient to determine the amount of allocation by understanding how the algorithm works. I don't even know how you can accurately measure allocations from a specific builtin call - seems quite tricky. 

For `insertCoin`, the key is to be aware that rebalancing allocates at most O(logn) nodes. I think it's more than enough to assume that it allocates `ValueMaxDepth` number of nodes and be done with it. If you want to be more precise, feel free to read the paper to determine the exact worst case (which should still be far easier than trying to measure it accurately), but for memory cost we don't need to be that precise.

plutus-core/cost-model/data/builtinCostModelC.json
zliu41: It's important to know this because if it exhausts memory 10 times faster than it exhausts CPU, it indicates that the memory cost is nonsensical and can severely limit the usefulness of Value.

ana-pantilie: I had time today to do this, these are the results. I created and ran the following test:
```
compiledInsertCoin :: CompiledCode BuiltinValue
compiledInsertCoin =
  $$(compile [||insertCoin||])
    `unsafeApplyCode` liftCodeDef "foo"
    `unsafeApplyCode` liftCodeDef "bar"
    `unsafeApplyCode` liftCodeDef 10
    `unsafeApplyCode` liftCodeDef (mkValueInData 1000000)

mkValueInData :: Int -> BuiltinValue
mkValueInData n =
  let currSymbols = (\x -> B.mkB $ "cS" <> B.encodeUtf8 (B.toBuiltin (Text.pack (show x)))) <$> take n [1 ..]
      tokN = B.mkB "tN"
      entries =
        B.mkMap
          [ ( cS
            , B.mkMap [(tokN, B.mkI 100)]
            )
          | cS <- currSymbols
          ]
   in B.unsafeDataAsValue entries
```

The resulting budget is:
```
CPU:                   935_697
Memory:                  1_786
AST Size:                   12
Flat Size:          17_888_928
```

I would say that's definitely not memory bound, unless I didn't construct the example properly.

zliu41: You can do a rough calculation like this: when you `insertCoin` into a `Value` of max depth 10:

```
CPU = (18413 * 10 + 356924) / 10000000000
= 0.00541054% of the transaction CPU limit

Memory = (21 * 10 + 45) / 14000000
= 0.001821428% of the transaction memory limit
```

So we are good.

zliu41 (approved):
Looks good but we need @kwxm to sign off on this.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -499,6 +500,32 @@ data ModelConstantOrTwoArguments = ModelConstantOrTwoArguments
   deriving stock (Show, Eq, Generic, Lift)
   deriving anyclass (NFData)
 
+-- | a*x + b*y + c*x*y + I
+data TwoVariableWithInteractionFunction = TwoVariableWithInteractionFunction
+  { intercept :: Intercept
+  , slopex :: Slope
```
kwxm: I'd suggest not including `x` and `y` in the parameter names here since it's conceivable that we could re-use this for something else where it's applied to other parameters.  Similar costing functions use `slope1` and `slope2` instead, and I suppose we could have `slope12` as well; alternatively, this is a type of polynomial (and the graph isn't a plane, so it doesn't really make sense to call the coefficients "slopes"), so we could use the already existing `Coefficient00`, `Coefficient01`, `Coefficient 10`, and `Coefficient11` instead, although that's a bit lengthy.  

This thing is in fact an instance of  the `TwoVariableQuadraticFunction` type (with the `x^2` and `y^2` coeccificents equal to zero), but it'd probably be overkill to use that instead.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -579,52 +607,52 @@ runTwoArgumentModel
 runTwoArgumentModel
   (ModelTwoArgumentsConstantCost c) = lazy $ \_ _ -> CostLast c
 runTwoArgumentModel
-  (ModelTwoArgumentsAddedSizes (OneVariableLinearFunction intercept slope)) =
+  (ModelTwoArgumentsAddedSizes (OneVariableLinearFunction intercept' slope)) =
```
kwxm: Are these all `intercept'` now because `TwoVariableWithInteractionFunction` contains a field called `intercept`?  It seems questionable to change a large number of variable names so that the simple  form of the name can be used in a small number of places.  Note that other similar types have unwieldy field names like `oneVariableLinearFunctionIntercept` to avoid clashes and leave the simple names available for use elsewhere.

plutus-core/cost-model/data/models.R
```diff
@@ -170,7 +173,7 @@ get.bench.data <- function(path) {
         comment.char="#"
     )
 
-    benchname <- regex("([[:alnum:]_]+)/(\\d+)(?:/(\\d+))?(?:/(\\d+))?")
+    benchname <- regex("([[:alnum:]_]+)/(\\d+)(?:/(\\d+))?(?:/(\\d+))?(?:/(\\d+))?")
     ## We have benchmark names like "AddInteger/11/22", the numbers representing the sizes of
     ## the inputs to the benchmark.  This extracts the name and up to three numbers, returning
     ## "NA" for any that are missing.  If we ever have builtins with more than three arguments
```
kwxm: This comment's outdated now because we do in fact have builtins with more than three arguments and have had for some time.  I think this is the first time we've had to base costing on an argument after the third one though.

plutus-core/cost-model/data/models.R
```diff
@@ -182,7 +185,7 @@ get.bench.data <- function(path) {
     ## the size of the first one because the others are terms (and the time is
     ## constant anyway).
 
-    numbercols = c("x_mem", "y_mem", "z_mem")
+    numbercols = c("x_mem", "y_mem", "z_mem", "w_mem")
```
kwxm: A very small quibble: when I've been working with benchmark results in R I've been in the habit of calling the arguments `x`, `y`, `z`, `u`, `v`, and `w`, as in the `toframe` script.  I don't think that notation's actually made it into the codebase anywhere before though.  I'm not sure what we'll do if we have to worry about a seventh argument at some time in the future.  It's a bit late now, but maybe we should have called the arguments `x1`. `x2`, `x3`, ...

plutus-core/cost-model/data/models.R
```diff
@@ -817,8 +828,21 @@ modelFun <- function(path) {
...
+    # W wrapped with `Logarithmic . ValueOuterOrMaxInner`
+    insertCoinModel           <- linearInW ("InsertCoin")    
+
+    # X and Y wrapped with `ValueTotalSize` (contained value size)
+    unionValueModel         <- {
+        fname <- "UnionValue"
+        filtered <- data %>%
+            filter.and.check.nonempty(fname) %>%
+            discard.overhead ()
+        filtered$off <- 1 # offset to force inclusion of intercept in the model
```
kwxm: What happens if you let `lm` infer the intercept? Do you get some huge number?  I think it's generally regarded as bad practice to force models to got through specific points since it can cause all kinds of distortion.  This will probably give you estimates smaller than the real value for small inputs, but maybe that's acceptable if it gives better estimates for bigger values.  Admittedly, letting `lm` infer the intercept means that you sometimes end up with a negative intercept, and there's code in this file to adjust it if that happens.

kwxm (commented):
Here are some initial comments; I'll continue to review the PR (in particular, the benchmarking code: it's going to take me a while to grasp the details fo that) and add more comments later.  It all looks pretty sensible so far though.

plutus-core/cost-model/print-cost-model/Main.hs
```diff
@@ -92,6 +92,26 @@ renderTwoVariableQuadraticFunction
...
+  :: TwoVariableWithInteractionFunction
+  -> String
+  -> String
+  -> String
+renderWithInteractionInXAndY
+  (TwoVariableWithInteractionFunction intercept cx cy cxy)
+  var1
+  var2 =
+    printf
+      "%d + %d + %d*%s + %d*%s + %d*%s*%s"
```
kwxm:
```suggestion
      "%d + %d*%s + %d*%s + %d*%s*%s"
```

There's an extra `%d` at the start.

plutus-core/cost-model/data/models.R
kwxm: Oh wait, that gets ignored anyway because it does infer a negative intercept: when you run `generate-cost-model` it says (amongst other things)
```
** WARNING: a negative coefficient -3216.982664 for (Intercept) occurred in the model for unValueData. This has been adjusted to 0.001.
** WARNING: a negative coefficient -764.798877 for (Intercept) occurred in the model for unionValue. This has been adjusted to 0.001.
** WARNING: a negative coefficient -2163.609705 for (Intercept) occurred in the model for scaleValue. This has been adjusted to 0.001.
```
So in the JSON file these all end up with an intercept of 1000 anyway.  I'm not sure what's going on here...

plutus-core/cost-model/data/models.R
```diff
@@ -832,6 +856,9 @@ modelFun <- function(path) {
     # Sizes of parameters are used as is (unwrapped):
     valueDataModel            <- constantModel ("ValueData")
     unValueDataModel          <- linearInX ("UnValueData")
+    
+    # Y wrapped with `TotalSize` (contained value size)
+    scaleValueModel           <- linearInY ("ScaleValue")
```
kwxm, line 859: I'm still puzzled by the shape of the data here, but making it linear in `y` does  give a good overapproximation for all values of `y`.  The downward bend at the start drags the regression line down quite a bit and you end up with a negavie intercept, but later that gets adjusted to 1000 and we end up with a line that lies just above all of the benchmark times.

plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs
```diff
@@ -291,6 +291,13 @@ batch6 =
   , IndexArray
   , Bls12_381_G1_multiScalarMul
   , Bls12_381_G2_multiScalarMul
+  , LookupCoin
+  , InsertCoin
+  , ValueContains
+  , UnionValue
+  , ScaleValue
+  , ValueData
+  , UnValueData
```
kwxm:
```suggestion
  , InsertCoin
  , LookupCoin
  , UnionValue
  , ValueContains
  , ValueData
  , UnValueData
  , ScaleValue
```

^ That's the [same order](https://github.com/IntersectMBO/plutus/blob/b559899fce0404ea7dbbe063dc74159e2642c03c/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs#L198C3-L205C15) as the constructors in `Builtins.hs`.  It probably doesn't matter, but let's be consistent.

plutus-ledger-api/src/PlutusLedgerApi/V1/ParamName.hs
```diff
@@ -329,5 +329,19 @@ data ParamName
   | UnValueData'cpu'arguments'intercept
   | UnValueData'cpu'arguments'slope
   | UnValueData'memory'arguments
+  | InsertCoin'cpu'arguments'intercept
```
kwxm: Maybe we should also put the parameter names in the same [order](https://github.com/IntersectMBO/plutus/blob/b559899fce0404ea7dbbe063dc74159e2642c03c/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs#L198C3-L205C15) as the constructors in `Builtins.hs` (and also for `V2` and `V3`).  Again, it probably doesn't matter, but it'd make it a bit easier to match up the parameters with the relevant builtins.

plutus-core/cost-model/budgeting-bench/Main.hs
```diff
@@ -61,7 +61,8 @@ main = do
       <> Benchmarks.Strings.makeBenchmarks gen
       <> Benchmarks.Tracing.makeBenchmarks gen
       <> Benchmarks.Unit.makeBenchmarks gen
-      <> Benchmarks.Values.makeBenchmarks gen
+      <> Benchmarks.Values.makeBenchmarks
+        gen
```
kwxm: Why is this on a new line?  Is fourmolu being awkward?

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -382,29 +382,43 @@ valueContains v1 v2
 {-# INLINEABLE valueContains #-}
 
 {-| \(O(n_{1}) + O(n_{2})\), where \(n_{1}\) and \(n_{2}\) are the total sizes
-(i.e., sum of inner map sizes) of the two maps. -}
+(i.e., sum of inner map sizes) of the two maps.
+
+Shortcircuits if either value is empty.
+
+Since 'unionValue' is commutative, we switch the arguments whenever the second
```
kwxm, line 389: This means that we effectively always evaluate `unionValue x y` with the total size of `x` greater than the total size of `y`, right?  In the costing benchmarks for `value` it [says](https://github.com/IntersectMBO/plutus/blob/c411ba4c2081a32cd55b4fc8b81d6cb04571467f/plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs#L257)
> therefore we consider two Values where the first is a sub-value of the second.

and that's the case where the total size of `x` is less than the total size of `y`.  For a horrible moment I thought that that was the wrong way round, but it doesn't matter because the implementation will swap them anyway. In fact maybe that's better, because it'll include the (very small) swapping time in the costs as well.

plutus-core/cost-model/budgeting-bench/Common.hs
```diff
@@ -85,13 +87,19 @@ benchWith
 -- gets back a 'Data' value it'll traverse all of it.
 benchWith params name term =
   bench name $
-    whnf (handleEvaluationErrors . evaluateCekNoEmit params) term
```
kwxm, line 88: I may have to change this to use something like `evaluateCekLikeInProd` (see https://github.com/IntersectMBO/plutus-private/issues/1722), in which case the log messages will disappear again. 

Also, I wonder if changing from a non-emitting evaluator to one which does emit will affect the benchmark performance.  I guess we can check that by running a few benchmarks with both versions of the evaluator.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -212,9 +216,111 @@ unValueDataBenchmark :: StdGen -> Benchmark
...
+{-| Maximum total size of Value arguments for ScaleValue benchmarking.
+Based on practical limits within execution budget constraints. -}
+maxScaleValueEntries :: Int
+maxScaleValueEntries = 90_000
+
+{-| Generate argument pairs for ScaleValue benchmarking.
+Since 'scaleValue' needs to traverse the entire Value, we may fix the structure
+of the Value to be a flattened map with a single token name per policy ID.
+To ensure worst-case performance, we fix the resulting scaled quantities to
+be 'maxBound'. -}
```
kwxm, line 294: I wonder if this really is the worst case.  All of the multiplications are `(maxBound/2) * 2`, and perhaps mulitplication by 2 will be extra-efficient.  Also, is there any danger that caching will speed things up, since we're doing the same multiplication each time?  If we're really paranoid we could try something like generating random numbers close to `maxBound` and setting the scalar and the  amount to be the square root of that, or something close.  I suspect that most of the work will be done traversing the values,  so the cost of the multiplications won't in fact make any significant difference (I could be wrong though!).

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -212,9 +216,111 @@ unValueDataBenchmark :: StdGen -> Benchmark
...
+    (id, id, id, ValueMaxDepth)
+    InsertCoin
+    []
+    (runBenchGen gen insertCoinArgs)
+
+insertCoinArgs :: StatefulGen g m => g -> m [(ByteString, ByteString, Integer, Value)]
+insertCoinArgs gen = do
+  lookupArgs <- lookupCoinArgs gen
+  let noOfBenchs = length lookupArgs
+  amounts <- genZeroOrMaxAmount gen noOfBenchs
```
kwxm: Just to be annoying, I'd go for "Benches" here instead of "Benchs".

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -212,9 +216,111 @@ unValueDataBenchmark :: StdGen -> Benchmark
...
+      value2 = buildValue policyIdsV2 [tokenName] (mkQuantity amt)
+  numPolicyIdsToKeep <- uniformRM (1, numPolicyIdsV2) gen
+  let policyIdsV1 = take numPolicyIdsToKeep policyIdsV2
+      value1 = buildValue policyIdsV1 [tokenName] (mkQuantity amt)
+  pure (value1, value2)
+
+----------------------------------------------------------------------------------------------------
+-- ScaleValue --------------------------------------------------------------------------------------
+
+scaleValueBenchmark :: StdGen -> Benchmark
```
kwxm, line 277: I'd still like to work oout why the graph is that odd shape (curved at first but then straightening out).  The benchmark inputs look perfectly reasonable, so it's a bit mystifying.  I don't think we need to let that hold us up though.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -64,6 +64,100 @@ mlResultMemSize = toMemSize Pairing.mlResultMemSizeBytes
 
 newtype Id a = Id {getId :: a}
 
+{- Note [Memory model for Value builtins]
+---------------------------------------
+
+## Overview
+
+The memory model for the Value builtin 'unionValue' estimates the heap allocation
```
kwxm: This isn't just about `unionValue`: it talks about other builtins later.  Presumably when this gets merged it'll need to be updated to deal with the other `value` builtins too, which might require significant rewriting.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -64,6 +64,100 @@ mlResultMemSize = toMemSize Pairing.mlResultMemSizeBytes
...
+  2. The map structure is flat, where each currency symbol maps to only one token name
+  3. This simplifies analysis while providing a reasonable approximation
+  4. Deeper trees would require more reallocation, but accounting for this is too complex
+
+## Per-Pair Allocation (Outer and Inner Maps)
+
+For each unique (currency symbol, token name) pair:
+
+Outer 'Map' (CurrencySymbol to inner 'Map'):
+  - 1 word for the 'Bin' closure
```
kwxm, line 109: This is talking about the implementation of the `Map` type: there's nothing to tell the reader what `Bin` and `Tip` are, and seraching the Plutus code won't help.  Maybe include the definition of `Map` so that we know what this is talking about.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -64,6 +64,100 @@ mlResultMemSize = toMemSize Pairing.mlResultMemSizeBytes
...
+number of unique (currency symbol, token name) pairs, based on worst-case allocation.
+
+## Value Structure
+
+Structurally, a 'Value' consists of:
+  - A nested 'Data.Map.Map': Map CurrencySymbol (Map TokenName Integer)
+  - Bookkeeping data: two unboxed 'Int's and a 'Data.IntMap.IntMap Int'
+
+Runtime memory layout breakdown:
+  - 1 word for the 'Value' constructor
```
kwxm, line 84: Is there a reference for where these numbers have come from?

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -64,6 +64,100 @@ mlResultMemSize = toMemSize Pairing.mlResultMemSizeBytes
...
+
+For 'insertCoin', the worst-case allocation occurs when a new pair is inserted into the map.
+Given the balanced tree representation, the memory allocation is based on 'ValueMaxDepth'
+which calculates the logarithmic depth of the tree. Thus, for 'insertCoin':
+
+    Memory = 21*log(n) + 12 + 21 + 12 = 21*log(n) + 45
+
+For 'unionValue', worst-case assumes disjoint sets of pairs in both 'Value's being united:
+
+    Memory = 21*n + 12 + 21*m + 12 = 21*(n + m) + 24
```
kwxm, line 158: It doesn't say what `m` is, but presumably `m` and `n` are the sizes of the arguments.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -64,6 +64,100 @@ mlResultMemSize = toMemSize Pairing.mlResultMemSizeBytes
...
+  - 1 word for the 'Value' constructor
+  - 1 word for each of the two unboxed 'Int's
+  - 1 word for the pointer to the 'IntMap'
+  - 1 word for the pointer to the nested 'Map'
+  - The allocations for the 'IntMap' and nested 'Map' themselves
+
+## Memory Analysis Assumptions
+
+We assume for each builtin application:
+  1. Both trees must be completely reallocated (worst-case)
```
kwxm, line 93: The reader needs to know that `Map`s are implemented as trees here.  If you look at the defintion of `Value` it's not obvious that trees are involved anywhere.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -64,6 +64,100 @@ mlResultMemSize = toMemSize Pairing.mlResultMemSizeBytes
...
+Runtime memory layout breakdown:
+  - 1 word for the 'Value' constructor
+  - 1 word for each of the two unboxed 'Int's
+  - 1 word for the pointer to the 'IntMap'
+  - 1 word for the pointer to the nested 'Map'
+  - The allocations for the 'IntMap' and nested 'Map' themselves
+
+## Memory Analysis Assumptions
+
+We assume for each builtin application:
```
kwxm, line 92: > each builtin application

This is just about `unionValue`, right?

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -64,6 +64,100 @@ mlResultMemSize = toMemSize Pairing.mlResultMemSizeBytes
...
+For 'unionValue', worst-case assumes disjoint sets of pairs in both 'Value's being united:
+
+    Memory = 21*n + 12 + 21*m + 12 = 21*(n + m) + 24
+
+For 'scaleValue', since there is no change in the number of unique pairs, the cost is
+estimated based on reallocating the 'Quantity' values for each pair in the input 'Value'.
+Remember, an integer is allocated as 3 words (pointer + closure data, where the data is max
+of two words size), so for 'scaleValue':
+
+    Memory = 3*n
```
kwxm: This seems incorrect.  If you're scaling a `value` with n nodes, in general the entire internal structure of the value will have to be re-created, so the total memory size of the output will be the same as the total memory size of the input, no?  That would mean that the memory used by `scaleValue` on a `value` with `n` nodes is the total memory usage of a value with `n` nodes, which I think is `21n+12`: that's a lot more than `3*n`.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -64,6 +64,100 @@ mlResultMemSize = toMemSize Pairing.mlResultMemSizeBytes
...
+The 'IntMap' contains exactly as many keys as there are unique inner map sizes.
+In our worst-case flat scenario (all inner maps have size 1), the footprint is constant:
+  - 1 word for the 'Tip' constructor
+  - 2 words for key and value pointers
+  - 2 words for key 'Int' closure (header + payload)
+  - 2 words for value 'Int' closure (header + payload)
+
+\**Total for IntMap: 7 words (constant)**
+
+## Final Formula
```
kwxm: The final formula for what?  I think this is the total memory usage of a `value` with `n` entries, but it'd be helpful if it said that explicilty.  I'd separate out this section and then use the formula in discussions of specific builtins later.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
ana-pantilie: Interesting point, thanks. I tested this out and there is indeed a slight increase in time for larger Values, which makes me think that when the amounts are all unique the extra time spent to compute the multiplications is accumulated and becomes visible when the `Value` is very big (over 40000 entries).

However, as you can see in the plot below, the extra time is insignificant when it comes to the current model we're using.

How to read the plot:
- g$y is total size of the `Value` argument
- g$t is the time
- green points are the _current_ version (which uses `(maxBound/2) * 2` for each multiplication)
- black points are the _proposed_ version (which uses `(x * x where x is in [sqrt(maxBound - 200), ..., sqrt(maxBound)]`)
- red line is the _current_ model (calculated for the, _current_, green points); this model is modified according to `models.R` where we force the intercept to be `1/1000` if the computed intercept is negative

To conclude, I believe the current model is good enough for both scenarios. I can change the generator to the new version and calculate the model based on that, but that would over-approximate things worse than this model. Or we can do some more investigating, if you think that this difference isn't negligible maybe we should define a new metric for integer sizes and take the scalar argument into account as well. Let me know what you think.

Screenshot 2026-01-08 at 14 19 29

Edit: I also checked with the new version being `(x * x where x is in [sqrt(maxBound - numPolicyIds), ..., sqrt(maxBound)]`, so instead of `200` I used `numPolicyIds`, which makes the interval for `x` much larger making duplicates even less possible, but the performance is basically the same.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
ana-pantilie: >  If you're scaling a value with n nodes, in general the entire internal structure of the value will have to be re-created, so the total memory size of the output will be the same as the total memory size of the input, no?

~Why would the entire structure have to be recreated? Wouldn't structural sharing avoid having to reallocate the whole `Value`, and the new `Value` could just point to new amounts?~ Edit: discussed in meeting, sharing doesn't work for the same reason modifying the last element in a list would not allow the tails of the lists to be shared.

plutus-core/cost-model/data/models.R
ana-pantilie: As discussed in the meeting: this is broken, also why is it `1`? R uses fractional numbers so `1` is way too big

plutus-core/cost-model/budgeting-bench/Main.hs
ana-pantilie: Fixed.

plutus-core/cost-model/budgeting-bench/Common.hs
ana-pantilie: I reverted this change to not impact your work and the current benchmark performance.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
kwxm: Thanks for checking!

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/SimpleJSON.hs
```diff
@@ -93,12 +110,14 @@ data Model
   | LinearInX LinearFunction
   | LinearInY LinearFunction
   | LinearInZ LinearFunction
+  | LinearInW LinearFunction
```
kwxm:
```suggestion
  | LinearInU LinearFunction
```

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/SimpleJSON.hs
```diff
@@ -139,13 +158,15 @@ instance FromJSON Model where
           "linear_in_x" -> LinearInX <$> parseJSON args
           "linear_in_y" -> LinearInY <$> parseJSON args
           "linear_in_z" -> LinearInZ <$> parseJSON args
+          "linear_in_u" -> LinearInW <$> parseJSON args
```
kwxm:
```suggestion
          "linear_in_u" -> LinearInU <$> parseJSON args
```

plutus-core/cost-model/print-cost-model/Main.hs
```diff
@@ -132,9 +152,11 @@ renderModel =
     LinearInX f -> [renderLinearFunction f "x"]
     LinearInY f -> [renderLinearFunction f "y"]
     LinearInZ f -> [renderLinearFunction f "z"]
+    LinearInW f -> [renderLinearFunction f "w"]
```
kwxm:
```suggestion
    LinearInU f -> [renderLinearFunction f "u"]
```

kwxm (approved):
This all looks good now.  Thanks!

PR #7438, opened 2025-11-20
Support `mkNil` for any builtin types

SeungheonOh:
You can now use `PlutusTx.Builtins.Opaque.mkNil @Bool` and `mkNil @BuiltinByteString`. 

This doesn't add support for arbitrarily nested array. Only adds support for `Bool` and `ByteString`.

I'm figuring out making this work with any subtype of default universe, so please don't merge yet

### DIFF
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
@@ -938,34 +938,34 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
       -> m (PIRTerm uni fun)
     compileMkNil ty =
       let
-        tyToUniverse :: GHC.Type -> Maybe (SomeTypeIn' PLC.DefaultUni)
+        tyToUniverse :: GHC.Type -> Maybe (SomeStarIn PLC.DefaultUni)
         tyToUniverse (GHC.TyConApp tyCon [tyArg1, tyArg2])
           | tyCon == builtinPairTyCon = do
-              (SomeTypeIn' a) <- tyToUniverse tyArg1
-              (SomeTypeIn' b) <- tyToUniverse tyArg2
-              pure $ SomeTypeIn' (PLC.DefaultUniPair a b)
+              (SomeStarIn a) <- tyToUniverse tyArg1
+              (SomeStarIn b) <- tyToUniverse tyArg2
+              pure $ SomeStarIn (PLC.DefaultUniPair a b)
           | otherwise = Nothing
         tyToUniverse (GHC.TyConApp tyCon [tyArg1])
           | tyCon == builtinListTyCon = do
-              (SomeTypeIn' a) <- tyToUniverse tyArg1
-              pure $ SomeTypeIn' (PLC.DefaultUniList a)
+              (SomeStarIn a) <- tyToUniverse tyArg1
+              pure $ SomeStarIn (PLC.DefaultUniList a)
           | tyCon == builtinArrayTyCon = do
-              (SomeTypeIn' a) <- tyToUniverse tyArg1
-              pure $ SomeTypeIn' (PLC.DefaultUniArray a)
+              (SomeStarIn a) <- tyToUniverse tyArg1
+              pure $ SomeStarIn (PLC.DefaultUniArray a)
           | otherwise = Nothing
         tyToUniverse (GHC.TyConApp tyCon [])
           | tyCon == GHC.integerTyCon || tyCon == builtinIntegerTyCon =
-            pure $ SomeTypeIn' PLC.DefaultUniInteger
-          | tyCon == builtinByteStringTyCon = pure $ SomeTypeIn' PLC.DefaultUniByteString
-          | tyCon == GHC.boolTyCon = pure $ SomeTypeIn' PLC.DefaultUniBool
-          | tyCon == builtinDataTyCon = pure $ SomeTypeIn' PLC.DefaultUniData
-          | tyCon == builtinBLS12_G1_TyCon = pure $ SomeTypeIn' PLC.DefaultUniBLS12_381_G1_Element
-          | tyCon == builtinBLS12_G2_TyCon = pure $ SomeTypeIn' PLC.DefaultUniBLS12_381_G2_Element
-          | tyCon == builtinValueTyCon = pure $ SomeTypeIn' PLC.DefaultUniValue
+            pure $ SomeStarIn PLC.DefaultUniInteger
+          | tyCon == builtinByteStringTyCon = pure $ SomeStarIn PLC.DefaultUniByteString
+          | tyCon == GHC.boolTyCon = pure $ SomeStarIn PLC.DefaultUniBool
+          | tyCon == builtinDataTyCon = pure $ SomeStarIn PLC.DefaultUniData
+          | tyCon == builtinBLS12_G1_TyCon = pure $ SomeStarIn PLC.DefaultUniBLS12_381_G1_Element
+          | tyCon == builtinBLS12_G2_TyCon = pure $ SomeStarIn PLC.DefaultUniBLS12_381_G2_Element
+          | tyCon == builtinValueTyCon = pure $ SomeStarIn PLC.DefaultUniValue
           | otherwise = Nothing
         tyToUniverse _ = Nothing
       in case tyToUniverse ty of
-           Just (SomeTypeIn' (ty' :: PLC.DefaultUni (PLC.Esc a))) ->
+           Just (SomeStarIn (ty' :: PLC.DefaultUni (PLC.Esc a))) ->
              pure $ PLC.constant annMayInline $ PLC.Some $ PLC.ValueOf (PLC.DefaultUniList ty') []
            Nothing -> throwPlain $ CompilationError "'mkNil' applied to an unknown type"
 
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Utils.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Utils.hs
@@ -27,10 +27,12 @@ import Data.Kind qualified as Kind
 import Data.Map qualified as Map
 import Data.Text qualified as T
 
-{- | Identical to `SomeTypeIn` but without existential kind.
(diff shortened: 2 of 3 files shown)


zliu41 (approved):

SeungheonOh:
I made it so that now `mkNil` will also work on any types.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -1090,19 +1130,9 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
-            | tyCon == builtinDataTyCon -> pure $ PLC.mkConstant annMayInline ([] @PLC.Data)
-            | tyCon == builtinBLS12_G1_TyCon ->
-                pure $ PLC.mkConstant annMayInline ([] @BLS12_381_G1.Element)
-            | tyCon == builtinBLS12_G2_TyCon ->
-                pure $ PLC.mkConstant annMayInline ([] @BLS12_381_G2.Element)
-          GHC.TyConApp tyCon [GHC.TyConApp tyArg1 [], GHC.TyConApp tyArg2 []]
-            | (tyCon, tyArg1, tyArg2) == (builtinPairTyCon, builtinDataTyCon, builtinDataTyCon) ->
-                pure $ PLC.mkConstant annMayInline ([] @(PLC.Data, PLC.Data))
-          _ -> throwPlain $ CompilationError "'mkNil' applied to an unknown type"
+      | GHC.getName n == mkNilOpaqueName -> compileMkNil ty
```
SeungheonOh, line 1133: even though we now match `mkNil` directly and have "OPAQUE" for `mkNil`, we still need `mkNilOpaque` because for some reason ghc would ignore OPAQUE and use `mkNilOpaque` anyways.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -929,6 +931,44 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
   listIndexId <- lookupGhcId '(PlutusTx.List.!!)
 
+  let
+    compileMkNil
+      :: (CompilingDefault uni fun m ann)
+      => GHC.Type
+      -> m (PIRTerm uni fun)
+    compileMkNil ty =
+      let
+        tyToUniverse :: GHC.Type -> Maybe (SomeTypeIn' PLC.DefaultUni)
```
SeungheonOh: is there a way to do this with existing `SomeTypeIn`? `SomeTypeIn` won't work because it has existential kind and it's impossible to force that to `Type`...

plutus-tx-plugin/src/PlutusTx/Compiler/Utils.hs
```diff
@@ -19,9 +23,15 @@ import Control.Monad.Reader (MonadReader, ask)
...
 import Language.Haskell.TH.Syntax qualified as TH
 
+import Data.Kind qualified as Kind
 import Data.Map qualified as Map
 import Data.Text qualified as T
 
+{- | Identical to `SomeTypeIn` but without existential kind.
+-}
+type SomeTypeIn' :: (Kind.Type -> Kind.Type) -> Kind.Type
+data SomeTypeIn' uni = forall a. SomeTypeIn' !(uni (PLC.Esc a))
```
effectfully: Could call it `SomeStarIn` or something. Just nitpicking obviously.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
effectfully: I think you can avoid existentials altogether and do something like

```haskell
withCompiledMkNil :: GHC.Type -> (forall a. PLC.DefaultUni (Esc a) -> m (PIRTerm uni fun))  -> m (PIRTerm uni fun)
```

but I also think that your version is simpler, so who cares.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
effectfully: I don't think we have `OPAQUE` for `mkNil`?
image

IIRC the point of this design was to make sure that some type class indirection doesn't get in the way, but I don't remember if it was necessary.

effectfully (approved):
You're my hero.

This was my original idea on how to make it work in the general case:

> The only more or less reliable way I could think of is serializing the type at the type level into a `Symbol` (to make sure it's fully done at compile time), requiring it to be `KnownSymbol`, then extracting the resulting `String` in the compiler and parsing it back to get a `DefaultUni` object. Sounds pretty crazy, but I can't think of another way of parsing a Haskell value back after embedding it into a `CoreExpr`. 

But that's a lot of effort for something rather insignificant and your solution should be enough in most cases (I don't wanna think when it's not enough).

I'll check if I can it statically detectable that an instance is missing.

plutus-tx-plugin/src/PlutusTx/Compiler/Utils.hs
```diff
@@ -19,9 +23,15 @@ import Control.Monad.Reader (MonadReader, ask)
 
 import Language.Haskell.TH.Syntax qualified as TH
 
+import Data.Kind qualified as Kind
 import Data.Map qualified as Map
 import Data.Text qualified as T
 
+{- | Identical to `SomeTypeIn` but without existential kind.
```
zliu41: Elaborate why the existential kind is problematic?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -929,6 +931,44 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
...
+        tyToUniverse (GHC.TyConApp tyCon [])
+          | tyCon == GHC.integerTyCon || tyCon == builtinIntegerTyCon =
+            pure $ SomeTypeIn' PLC.DefaultUniInteger
+          | tyCon == builtinByteStringTyCon = pure $ SomeTypeIn' PLC.DefaultUniByteString
+          | tyCon == GHC.boolTyCon = pure $ SomeTypeIn' PLC.DefaultUniBool
+          | tyCon == builtinDataTyCon = pure $ SomeTypeIn' PLC.DefaultUniData
+          | tyCon == builtinBLS12_G1_TyCon = pure $ SomeTypeIn' PLC.DefaultUniBLS12_381_G1_Element
+          | tyCon == builtinBLS12_G2_TyCon = pure $ SomeTypeIn' PLC.DefaultUniBLS12_381_G2_Element
+          | tyCon == builtinValueTyCon = pure $ SomeTypeIn' PLC.DefaultUniValue
+          | otherwise = Nothing
```
kwxm, line 965: This worries me a little because we can add a new built-in type and forget to add it here, and we won't get any kind of warning from the compiler.  I don't immediately see any way of avoiding that though.  Maybe we could have a warning comment somewhere that you'd see it when you're adding a new type, for example at the definition of `mkNil` or `DefaultUni` ?  Or could we have a test that iterates over everything in `DefaultUni` and checks that `mkNil` works for the Plinth version?

plutus-tx/src/PlutusTx/Builtins/HasOpaque.hs
```diff
@@ -268,6 +268,7 @@ class MkNil arep where
   mkNil = mkNilOpaque
   {-# INLINEABLE mkNil #-}
 instance MkNil BuiltinInteger
+instance MkNil BuiltinByteString
```
kwxm, line 278: I tried adding an instance for `BuiltinArray` here and then `mkNil` does indeed work for the improbable case of lists of arrays.  Also, since we're here, the first lot of parentheses can be deleted in 
```
instance (MkNil a) => MkNil (BuiltinList a)
```
a few lines later.

kwxm (approved):
This is great: it makes the behaviour of the compiler much more consistent.  I don't understand why it works, but I tried some examples and it was able to  cope with all of them.

It might be worth adding some tests for the new capabilities before merging. Also, is there any documentation that will need to be updated? Possibly we've said somehere that `mkNil` doesn't work in full generalilty, but I don't know where to look for that kind of thing.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
SeungheonOh: I refuse to use rank-n-types whenever possible

SeungheonOh:
> Possibly we've said somehere that mkNil doesn't work in full generalilty, but I don't know where to look for that kind of thing.

I've scanned over the repo, didn't find anything that mention `mkNil`.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
SeungheonOh: I don't really have a concern here since this won't really generate any unsafe code. Adding a comment on class definition of `MkNil` would be sufficient imo.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
SeungheonOh: Ah, right. I tried with Opaque, hoping I won't need to match on both `mkNilOpaque` and `mkNil` on the plugin side, which it wasn't the case. Now that plugin handles both `mkNilOpaque` and `mkNil`, it doesn't really matter.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
effectfully: > This worries me a little because we can add a new built-in type and forget to add it here

@SeungheonOh add

```haskell
{-| Test that 'mkNil' works for all types from 'PLC.DefaultUni'
-}
instance
  TestAllBuiltinsSatisfy
    (PLC.AllBuiltinArgs PLC.DefaultUni MkNil)
    MkNil
```

to `PlutusTx.Lift.TestInstances` and you'll find out that there are still built-in types missing. Also add a comment saying that if an instance for a built-in type is missing, one should go and update the `tyUniverse` thing.

PR #7452, opened 2025-11-25
[OPTIMIZATION] Combine `Either Text (HeadSpine ...)` into `HeadSpine`

SeungheonOh:
seeing if this is worthwhile optimization

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
@@ -338,7 +338,7 @@ data Spine a
 data HeadSpine err a b
     = HeadOnly a
     | HeadSpine a (Spine b)
-    | HeadError err
+    | HeadError ~err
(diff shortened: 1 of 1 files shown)


plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs
```diff
@@ -335,20 +335,22 @@ data Spine a
 -- it's empty -- and the no-spine case is by far the most common one, hence we want to optimize it).
 --
 -- Used in built-in functions returning function applications such as 'CaseList'.
-data HeadSpine a b
+data HeadSpine err a b
     = HeadOnly a
     | HeadSpine a (Spine b)
+    | HeadError err
```
effectfully: You may wanna try making it lazy, 'cause God knows what GHC is doing there.

effectfully (approved):
Cool, I knew it'd be an improvement, but I didn't realize it'd be that much of an improvement.

SeungheonOh:
cool

PR #7456, opened 2025-11-26
Add deriveEnum for Plinth

bezirg:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-tx/plutus-tx.cabal
+++ b/plutus-tx/plutus-tx.cabal
@@ -118,13 +118,13 @@ library
     PlutusTx.Utils
 
   other-modules:
+    PlutusTx.Enum.TH
     PlutusTx.IsData.Instances
     PlutusTx.IsData.TH
     PlutusTx.Lift.Instances
     PlutusTx.Lift.TestInstances
     PlutusTx.Lift.TH
     PlutusTx.Lift.THUtils
-    PlutusTx.Enum.TH
 
   build-depends:
     , aeson                  >=2.2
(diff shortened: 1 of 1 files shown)


SeungheonOh:
With regards to enums, it would also be nice to have a variant of `mkIsData` for enum types where it will encode each constructors as integer instead of `Constr n []`, but it's a bit off topic

bezirg:
> With regards to enums, it would also be nice to have a variant of `mkIsData` for enum types where it will encode each constructors as integer instead of `Constr n []`, but it's a bit off topic

Interesting idea. I don't know if the plutus core spec though has to be modified to account for this optimization.

bezirg:
> With regards to enums, it would also be nice to have a variant of `mkIsData` for enum types where it will encode each constructors as integer instead of `Constr n []`, but it's a bit off topic

I think I will add it in another PR as `deriveEnumData ''ADT`

plutus-tx/plutus-tx.cabal
```diff
@@ -123,6 +124,7 @@ library
     PlutusTx.Lift.TestInstances
     PlutusTx.Lift.TH
     PlutusTx.Lift.THUtils
+    PlutusTx.Enum.TH
```
Unisay: `cabal-fmt` re-orders modules alphabetically, does it?

bezirg: Is cabal-fmt part of CI? I will run it manually

SeungheonOh:
> Interesting idea. I don't know if the plutus core spec though has to be modified to account for this optimization.

This is completely independent from core spec since it just changes how we generate some code. 

On the side note we can probably make a smart version of current TH helper that decides most efficient representation like Constr vs List vs I

PR #7458, opened 2025-11-26
refactor(benchmark): use ValueTotalSize for both valueContains args

Unisay:
## Summary

Change `valueContains` benchmark to use `ValueTotalSize` wrapper for both container and contained arguments, instead of `ValueLogOuterSizeAddLogMaxInnerSize` for the container.

This addresses Kenneth's observation that using total size for both arguments will:
1. Straighten out the curved edge of the benchmark data
2. Make the early-exit boundary at n₁ = n₂ visible as a straight line

### Benchmark Wrapper Update
- Changed wrapper from `(ValueLogOuterSizeAddLogMaxInnerSize, ValueTotalSize)` to `(ValueTotalSize, ValueTotalSize)`
- Updated denotation signature in `Builtins.hs` to match
- Updated golden signature test file

### Cost Model Regeneration
- Regenerated benchmark data in `benching-conway.csv`
- Updated cost model parameters in all JSON files (A, B, C)
- New model: `multiplied_sizes` with updated coefficients

### Conformance Tests
- Updated budget expectations for `valueContains` tests
- Updated budget expectations for array builtin tests affected by rebase

## Technical Context

The `valueContains` builtin uses `Map.isSubmapOfBy` with a splitLookup-based divide-and-conquer algorithm. By measuring both arguments using `ValueTotalSize` (total entry count), the cost model better reflects the actual complexity bounds where early exit occurs when contained size exceeds container size.

## Test plan
- [x] Build compiles with `--ghc-options=-Werror`
- [x] Conformance tests updated with new budget expectations
- [x] Benchmark still generates ~1000 test cases


Unisay:
newplot
newplot1
newplot2

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -107,22 +107,31 @@ valueContainsBenchmark gen =
 
 valueContainsArgs :: StdGen -> [(Value, Value)]
 valueContainsArgs gen = runStateGen_ gen \g -> do
-  {- ValueContains performs multiple LookupCoin operations (one per entry in contained).
-     Worst case: All lookups succeed at maximum depth with many entries to check.
+  {- ValueContains uses nested Map.isSubmapOfBy with a splitLookup-based
+     divide-and-conquer algorithm. Each key in contained triggers a splitLookup
+     on container, then recursively checks subtrees (not independent root lookups).
+
+     Complexity: O(n₂ × log(max(m₁, k₁))) where:
```
kwxm: The bound given in the [documentation](https://hackage-content.haskell.org/package/containers-0.8/docs/Data-Map-Strict.html#g:28) is `O(m*(n/m+1))`, where `n` is the number of nodes in the smaller map and `m` is the number in the larger one.  You certainly have to examine all of the nodes in the small map and you potentially have to visit almost all of the nodes in the larger one to find the things you're looking for, so it makes sense that the bound should involve the total sizes and not the  depth.  Things are complicated here by the fact that we have a nested map, but I think we can pretend that it's one big map where the inner maps are attached to the nodes of the outer one.

zliu41:
> I think the model should probably be linear_in_x_and_y instead of multiplied_sizes

`linear_in_x_and_y` when `x >= y`, and constant otherwise.

Unisay:
Closing this PR in favor of https://github.com/IntersectMBO/plutus/pull/7476

PR #7468, opened 2025-11-28
perf(builtins): optimize listToArray implementation and costing

Unisay:
## Context

- **What**: Optimize `listToArray` builtin implementation and update cost model parameters
- **Why**: Improve performance and accuracy of list-to-array conversions in Plutus Core

## Approach

This PR makes two key optimizations to the `listToArray` builtin:

**1. Implementation Optimization**: Changed from `Vector.fromList` to `Vector.fromListN` which avoids repeated array resizing during conversion. Since we already traverse the list to calculate memory usage, we know the length upfront and can allocate the exact size needed.

**2. Memory Usage Costing**: Changed list memory costing from eager full-spine traversal to batched incremental costing. Instead of computing `length l` upfront (which forces the entire list spine), we now process the list in batches of 100 elements, producing `CostRose` nodes incrementally. This is more efficient for large lists while maintaining accurate costing.

The benchmark range was also expanded from 1-100 to 1-5000 elements to capture the linear cost behavior across a wider range of inputs, resulting in updated cost model parameters (lower intercept, higher slope).

### Core Implementation
- Use `Vector.fromListN` instead of `Vector.fromList` for more efficient array allocation
- Implement batched memory costing for lists using `CostRose` tree structure

### Cost Model Updates
- Updated `listToArray` CPU cost parameters based on new benchmarks
- Intercept reduced from 307802 to 1000
- Slope increased from 8496 to 24619
- Reflects the optimized implementation characteristics
- Regenerated `indexArray` and `lengthOfArray` CPU costs (were stale since 69cb3d9cc6)

### Cost Model Visualization Fix
- Fixed overhead calculation in JS to use `Nop*o` pattern (matching R's models.R)

### Cost Model presented visually
 
https://plutus.cardano.intersectmbo.org/pr-preview/cost-models/pr-7468/listtoarray/index.html?branch=yura%2Flist-to-array-improvements

### Benchmarking
- Expanded benchmark range to 1-5000 elements for better linear model fitting

### Documentation
- Added cost model visualization pages for all three array builtins:
  - `ListToArray` (linear cost model)
  - `LengthOfArray` (constant cost)
  - `IndexArray` (constant cost)
- Updated navigation across all existing visualization pages

### Implementation Testing
1. Run the test suite: `cabal test plutus-core-test`
2. Run conformance tests (expect budget differences due to cost model changes)

### Cost Model Verification
1. Start local server: `cd doc/cost-models && python -m http.server 8000`
2. Visit http://localhost:8000/listtoarray/
3. Verify benchmark data points align with model predictions
4. Check LengthOfArray and IndexArray pages show constant cost behavior

### Benchmark Reproduction
```bash
cabal run plutus-core:cost-model-budgeting-bench -- ListToArray
```

## Author's Checklist

- [x] Implementation uses more efficient `Vector.fromListN`
- [x] Memory costing uses batched incremental approach
- [x] Cost model parameters updated from new benchmarks
- [x] Visualization pages added for array builtins
- [x] Conformance test budget updates reviewed
- [x] Fixed stale `indexArray`/`lengthOfArray` cost models
- [x] Fixed JS visualization overhead calculation

### DIFF
--- a/plutus-core/cost-model/data/builtinCostModelA.json
+++ b/plutus-core/cost-model/data/builtinCostModelA.json
@@ -726,7 +726,7 @@
   },
   "indexArray": {
     "cpu": {
-      "arguments": 194922,
+      "arguments": 232010,
       "type": "constant_cost"
     },
     "memory": {
@@ -736,7 +736,7 @@
   },
   "lengthOfArray": {
     "cpu": {
-      "arguments": 198994,
+      "arguments": 231883,
       "type": "constant_cost"
     },
     "memory": {
--- a/plutus-core/cost-model/data/builtinCostModelB.json
+++ b/plutus-core/cost-model/data/builtinCostModelB.json
@@ -726,7 +726,7 @@
   },
   "indexArray": {
     "cpu": {
-      "arguments": 194922,
+      "arguments": 232010,
       "type": "constant_cost"
     },
     "memory": {
@@ -736,7 +736,7 @@
   },
   "lengthOfArray": {
     "cpu": {
-      "arguments": 198994,
+      "arguments": 231883,
       "type": "constant_cost"
     },
     "memory": {
--- a/plutus-core/cost-model/data/builtinCostModelC.json
+++ b/plutus-core/cost-model/data/builtinCostModelC.json
@@ -735,7 +735,7 @@
   },
   "indexArray": {
     "cpu": {
-      "arguments": 194922,
+      "arguments": 232010,
       "type": "constant_cost"
     },
     "memory": {
@@ -745,7 +745,7 @@
   },
   "lengthOfArray": {
     "cpu": {
-      "arguments": 198994,
+      "arguments": 231883,
       "type": "constant_cost"
     },
     "memory": {


plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -203,12 +203,18 @@ instance ExMemoryUsage (a, b) where
...
+ This avoids forcing the entire list upfront just to compute the length, instead
+ producing costs incrementally as CostRose children. Each batch of 100 elements
+ produces one cost node, which is more efficient than per-element costing while
+ still avoiding full spine traversal before the builtin executes. -}
 instance ExMemoryUsage [a] where
-  memoryUsage l = singletonRose . fromIntegral $ length l
+  memoryUsage = go
+    where
+      batchSize = 100
+      go xs = case splitAt batchSize xs of
```
effectfully: I meant literal pattern matching: `(_:_:_:_:_:_:_:_:_:_:_:_:xs) -> <...>`. It's stupid, but it's way faster than using `splitAt`.

If you don't like doing it manually, you can statically unroll loops like I did [here](https://github.com/IntersectMBO/plutus/pull/5265).

Maybe it doesn't matter due to not being a bottleneck.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -203,12 +203,18 @@ instance ExMemoryUsage (a, b) where
...
+        (batch, [])   -> singletonRose (fromIntegral (length batch))
+        (_, rest)     -> CostRose (fromIntegral batchSize) [go rest]
```
effectfully: I'd cache `fromIntegral batchSize` explicitly, because GHC is probably not gonna do it.

effectfully (approved):
The data [here](https://plutus.cardano.intersectmbo.org/pr-preview/cost-models/pr-7468/listtoarray/index.html) still only shows lists of lengths up to 100? And the formula is the old one?

zliu41:
> Intercept reduced from 307802 to 1000
Slope increased from 8496 to 24619

How could the slope increase so dramatically? These numbers imply that for `length > 20`, the new cost will be much more expensive than the old one. Is that what you expected?

Unisay:
> The data [here](https://plutus.cardano.intersectmbo.org/pr-preview/cost-models/pr-7468/listtoarray/index.html) still only shows lists of lengths up to 100? And the formula is the old one?

Its important to use proper branch in the data source:
image

The link in the PR description does specify the branch to use, so if you follow that link it should be set automatically.

Unisay:
> > Intercept reduced from 307802 to 1000
> > Slope increased from 8496 to 24619
> 
> How could the slope increase so dramatically? These numbers imply that for `length > 20`, the new cost will be much more expensive than the old one. Is that what you expected?

The old benchmark only tested lists up to 100 elements, so the high intercept (307802) was compensating for an underestimated slope. With 1-5000 elements, the regression captures the actual per-element cost. Break-even is around n≈19, so yes, longer lists cost more than before, but that's because the old model was undercharging by extrapolating from too little data.

kwxm:
> > The data [here](https://plutus.cardano.intersectmbo.org/pr-preview/cost-models/pr-7468/listtoarray/index.html) still only shows lists of lengths up to 100? And the formula is the old one?
> 
> Its important to use proper branch in the data source: image
> 
> The link in the PR description does specify the branch to use, so if you follow that link it should be set automatically.

I was confused by this as well, and the [link](https://plutus.cardano.intersectmbo.org/pr-preview/cost-models/pr-7468/listtoarray/index.html?branch=yura%2Flist-to-array-improvements) in the PR description is only displaying sizes up to 100.  It looks as if it's still displaying the old data because the new data doesn't have many points with x <= 100. Maybe I've missed out some step?

PR #7474, opened 2025-12-03
Make `Enum PlutusTx.Rational` plc compilable

bezirg:
The Enum Rational instance was accidentally using some Haskell code which was making it fail to compile via plc.

Locally tested and now it plc-compiles with this patch.

### DIFF
--- a/plutus-tx/src/PlutusTx/Ratio.hs
+++ b/plutus-tx/src/PlutusTx/Ratio.hs
@@ -61,7 +61,6 @@ import GHC.Generics
 import PlutusTx.Blueprint.Class (HasBlueprintSchema (..))
 import PlutusTx.Blueprint.Definition (HasBlueprintDefinition (..), HasSchemaDefinition)
 import Prettyprinter (Pretty (..), (<+>))
-import Prelude (Ord (..), Show, (*))
 import Prelude qualified as Haskell
 
 {-| Represents an arbitrary-precision ratio.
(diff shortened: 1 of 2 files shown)


plutus-tx/src/PlutusTx/Ratio.hs
```diff
@@ -94,12 +93,12 @@ instance P.Ord Rational where
   {-# INLINEABLE (>) #-}
   Rational n d > Rational n' d' = (n P.* d') P.> (n' P.* d)
 
-instance Ord Rational where
-  compare (Rational n d) (Rational n' d') = compare (n * d') (n' * d)
-  Rational n d <= Rational n' d' = (n * d') <= (n' * d)
-  Rational n d >= Rational n' d' = (n * d') >= (n' * d)
-  Rational n d < Rational n' d' = (n * d') < (n' * d)
-  Rational n d > Rational n' d' = (n * d') > (n' * d)
+instance Haskell.Ord Rational where
```
SeungheonOh, line 96: Just curious: if we can use haskell's `Ord` for plugin, why have `PlutusTx.Ord`? Can you not just use haskell's `Ord` everywhere?

bezirg:
> @bezirg I don't understand the change - why did you switch from Plinth Ord to Haskell Ord?

I didn't. The old instance was improting Ord from Prelude.

I now made it more explicit that this is the
Haskell's instance.

This is the exact confusion I run into very often, and the reason behind the bug that this PR tries to fix :)

bezirg:
> @bezirg I don't understand the change - why did you switch from Plinth Ord to Haskell Ord?

The plinth ord instance is the one above which remains unchanged

plutus-tx/src/PlutusTx/Ratio.hs
bezirg: The plugin cannot work with Haskell Prelude/base typeclasses because they don't have the inlinable flag: as a result we cannot take the ghc core unfoldings. Maybe there is also another reason.

PR #7476, opened 2025-12-04
refactor(cost-model): use linear costing for valueContains

Unisay:
## Summary

- Change `valueContains` costing from `multiplied_sizes` to `const_above_diagonal` with `linear_in_x_and_y` inner model
- Add `LinearInXAndY` constructor to SimpleJSON.hs for JSON parsing
- Update cost model visualization to support new model types
- Update V1/V2/V3 ParamName enums to match new cost model structure
- Improve benchmark data distribution (1000+ points → 570 points)

## Motivation

The true complexity of `valueContains` is `O(m*log(n/m+1))` where n is the container size and m is the contained size. Kenneth MacKenzie showed that this is bounded above by a linear function `(m+n)/k` for some constant k.

The new model better reflects this:
- **Above diagonal** (y > x): constant cost (early exit, returns False immediately)
- **Below diagonal** (x >= y): `intercept + slope1*x + slope2*y`

## Benchmark Improvements

Addressed data distribution issues identified in code review:
- **Reduced clustering**: Below-diagonal sampling (numPolicies >= tokensPerPolicy) reduces oversampling at small sizes
- **Symmetric ranges**: Both container and contained now reach size 2047 (removed 1000 cap)
- **Better coverage**: 55 below-diagonal + 2 above-diagonal combinations (57 total)
- **Fewer points**: ~570 benchmark points (down from ~1000) → faster runtime
- **Maintained structure**: All sizes remain powers of 2 for worst-case saturated BST branches

## Changes

| File | Change |
|------|--------|
| `models.R` | Update `valueContainsModel` to use `const_above_diagonal` with `linear_in_x_and_y` |
| `SimpleJSON.hs` | Add `LinearInXAndY` constructor for JSON parsing |
| `utils.js` | Add evaluators for `linear_in_x_and_y` and `const_above_diagonal` |
| `Benchmarks/Values.hs` | Improve valueContainsArgs distribution strategy |
| `benching-conway.csv` | Regenerated with improved benchmark data |
| `builtinCostModel{A,B,C}.json` | Regenerated with new model type and improved benchmarks |
| `V{1,2,3}/ParamName.hs` | Update ParamName enums from 2 to 4 CPU parameters to match new structure |
| Conformance tests | Updated budget expectations to match new cost parameters |

## New Cost Model Structure

```json
"valueContains": {
  "cpu": {
    "arguments": {
      "constant": 358864,
      "model": {
        "arguments": {
          "intercept": 706629,
          "slope1": 2015,
          "slope2": 28294
        },
        "type": "linear_in_x_and_y"
      }
    },
    "type": "const_above_diagonal"
  },
  "memory": {
    "arguments": 32,
    "type": "constant_cost"
  }
}
```

## Parameter Update

The new cost model structure required updating the ParamName enums in V1, V2, and V3:

**Old** (2 CPU parameters):
- `ValueContains'cpu'arguments'intercept`
- `ValueContains'cpu'arguments'slope`

**New** (4 CPU parameters):
- `ValueContains'cpu'arguments'constant`
- `ValueContains'cpu'arguments'model'arguments'intercept`
- `ValueContains'cpu'arguments'model'arguments'slope1`
- `ValueContains'cpu'arguments'model'arguments'slope2`

This fixes `extractCostModelParamsLedgerOrder` failures that occurred when the parameter count didn't match the expected structure.

## Test plan

- [x] Build passes
- [x] Ledger API plugin tests pass (V1/V2/V3 validators)
- [x] Conformance tests pass with updated budget expectations
- [x] [Cost model visualization](https://plutus.cardano.intersectmbo.org/pr-preview/cost-models/pr-7476/valuecontains/index.html?branch=yura/valuecontains-linear-costing) works with new model type

Closes https://github.com/IntersectMBO/plutus-private/issues/1984

### DIFF
--- a/plutus-metatheory/src/Cost/Raw.lagda.md
+++ b/plutus-metatheory/src/Cost/Raw.lagda.md
@@ -140,6 +140,7 @@ data RawModel : Set where
     LinearInU              : LinearFunction → RawModel
     LiteralInYOrLinearInZ  : LinearFunction → RawModel
     LinearInMaxYZ          : LinearFunction → RawModel
+    LinearInXAndY          : TwoVariableLinearFunction → RawModel
     LinearInYAndZ          : TwoVariableLinearFunction -> RawModel
     QuadraticInY           : OneVariableQuadraticFunction → RawModel
     QuadraticInZ           : OneVariableQuadraticFunction → RawModel
@@ -153,10 +154,10 @@ data RawModel : Set where
 
 {-# COMPILE GHC RawModel = data Model (ConstantCost | AddedSizes |
     MultipliedSizes | MinSize | MaxSize | LinearInX | LinearInY | LinearInZ |
-    LinearInU | LiteralInYOrLinearInZ | LinearInMaxYZ | LinearInYAndZ |
-    QuadraticInY | QuadraticInZ | QuadraticInXAndY | WithInteractionInXAndY | 
-    SubtractedSizes | ConstAboveDiagonal | ConstBelowDiagonal | 
-    ConstOffDiagonal | ExpModCost) #-}
+    LinearInU | LiteralInYOrLinearInZ | LinearInMaxYZ | LinearInXAndY |
+    LinearInYAndZ | QuadraticInY | QuadraticInZ | QuadraticInXAndY |
+    WithInteractionInXAndY | SubtractedSizes | ConstAboveDiagonal |
+    ConstBelowDiagonal | ConstOffDiagonal | ExpModCost) #-}
 
 record CpuAndMemoryModel : Set where
      constructor mkCpuAndMemoryModel
--- a/plutus-metatheory/src/MAlonzo/Code/Cost/Model.hs
+++ b/plutus-metatheory/src/MAlonzo/Code/Cost/Model.hs
@@ -667,7 +667,7 @@ d_convertRawModel_292 v0 v1
                 2 -> coe v2
                 1 -> coe v2
                 _ -> coe v2
-         MAlonzo.Code.Cost.Raw.C_LinearInYAndZ_168 v3
+         MAlonzo.Code.Cost.Raw.C_LinearInYAndZ_170 v3
            -> case coe v0 of
                 3 -> case coe v3 of
                        MAlonzo.Code.Cost.Raw.C_mkTwoVariableLinearFunction_74 v4 v5 v6
@@ -679,7 +679,7 @@ d_convertRawModel_292 v0 v1
                 2 -> coe v2
                 1 -> coe v2
                 _ -> coe v2
-         MAlonzo.Code.Cost.Raw.C_QuadraticInY_170 v3
+         MAlonzo.Code.Cost.Raw.C_QuadraticInY_172 v3
            -> case coe v0 of
                 _ | coe geqInt (coe v0) (coe (2 :: Integer)) ->
                     case coe v3 of
@@ -695,7 +695,7 @@ d_convertRawModel_292 v0 v1
                       _ -> MAlonzo.RTE.mazUnreachableError
                 1 -> coe v2
                 _ -> coe v2
-         MAlonzo.Code.Cost.Raw.C_QuadraticInZ_172 v3
+         MAlonzo.Code.Cost.Raw.C_QuadraticInZ_174 v3
            -> case coe v0 of
                 _ | coe geqInt (coe v0) (coe (3 :: Integer)) ->
                     case coe v3 of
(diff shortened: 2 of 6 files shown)


Unisay:
This PR replaces https://github.com/IntersectMBO/plutus/pull/7476

plutus-core/cost-model/data/builtinCostModelA.json
```diff
@@ -1222,10 +1222,17 @@
   "valueContains": {
     "cpu": {
       "arguments": {
-        "intercept": 1163050,
-        "slope": 1470
+        "constant": 278039,
+        "model": {
+          "arguments": {
+            "intercept": 386227,
+            "slope1": 1970,
```
kwxm: The  theoretical linear upper bound of `k*(m+n)` would have `slope1 = slope2` here, but here the two slopes are very different!  I think that's OK though: the theoretical bound is very conservative and this seems to fit the data very well.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/SimpleJSON.hs
```diff
@@ -145,6 +146,7 @@ instance FromJSON Model where
           "exp_mod_cost" -> ExpModCost <$> parseJSON args
           "literal_in_y_or_linear_in_z" -> LiteralInYOrLinearInZ <$> parseJSON args
           "linear_in_max_yz" -> LinearInMaxYZ <$> parseJSON args
+          "linear_in_x_and_y" -> LinearInXAndY <$> parseJSON args
```
kwxm, line 169: We also need a case for `LinearInXAndY` in the `print-cost-model` executable [here](https://github.com/IntersectMBO/plutus/blob/b31004dd2e372d0af0382bf56473a2d033d0ff40/plutus-core/cost-model/print-cost-model/Main.hs#L140).

plutus-core/cost-model/data/models.R
```diff
@@ -818,15 +818,28 @@ modelFun <- function(path) {
...
             filter.and.check.nonempty(fname) %>%
-            discard.overhead ()
-        m <- lm(t ~ I(x_mem * y_mem), filtered)
-        mk.result(m, "multiplied_sizes")
+            discard.overhead()
+
+        # Above diagonal (y > x): containment impossible, quick False return
+        # Use minimum observed time as constant for this case
+        above_diag <- filtered %>% filter(y_mem > x_mem)
+        constant <- if (nrow(above_diag) > 0) ceiling(mean(above_diag$t)) else min(filtered$t)
```
kwxm: I suggest adding some points (maybe 20 or so) above the diagonal in the benchmarking code to ensure that `above_diag` is always nonempty and then we won't need to have the two `if`s in this code.  There's redundancy here that can be avoided by making sure that the benchmark results always contain points on both sides of the diagonal.  If someone was later to remove the points above the diagonal then we'd get an error here anyway when we get to `mean(above_diag)`, so there's no real reason to have two different cases to deal with the presence or absence of points above the diagonal.

kwxm (changes requested):
This looks good,  but again I urge you to reduce the number of benchmarking inputs: see [this earlier comment.](https://github.com/IntersectMBO/plutus/pull/7344#discussion_r2536202699)  We have over 1000 points here and that makes the benchmark take a very long time (> 90 minutes for just `valueContains`) and doesn't contribute anything to the accuracy of the model: you can see  this by taking subsets of the existing data and fitting models to them.  Also, the data is very unevenly distributed, which is probably not a good idea.  There are 500 inputs with x<200 and y<200, so 1/25 of the space of inputs is contributing half of the  datapoints, which will bias the model.  Something like the below-diagonal part of a 25x25 or 30x30 grid of evenly spaced inputs (plus some points above the diagonal) would probably suffice and not lose any precision.

Also, the inputs aren't symmetric: the `x` values go up to 2047, but the `y` values only go up to 1000, so we're not always checking cases like a value containing itself, which might take some time.  Can you remind me if there's some reason for this?  We had earlier discussions about what the worst case would be, but I don't recall the details at the moment.

plutus-core/cost-model/data/models.R
kwxm: I don't think there's any need to have a branch here any more. The benchmarks generate points above the diagonal now,  so we can be certain that the condition will always be satisfied (and we'll get an error if `above-diag` is empty for some reason).

kwxm:
Did you update the contents of  `benching-conway.csv"?  It still contains 1052 results for `ValueContains` but the new version of the benchmarks only has 607.

plutus-core/cost-model/data/models.R
Unisay: I've fixed this before but then my fix got lost during the git conflict resolution, I'll re-apply the fix. Thanks for catching it!

Unisay:
> Did you update the contents of `benching-conway.csv"? It still contains 1052 results for `ValueContains` but the new version of the benchmarks only has 607

My bad, fixed now:
image

kwxm (commented):
Thanks: I think we can merge this now!  We'lll probably need @zliu41 to do it though.

PR #7481, opened 2025-12-09
Docs for casing on constants

basetunnel:


Unisay:
The doc seems useful! I propose to add Plinth examples to accompany UPLC examples, as this user guide is for Plinth and it makes sense to showcase its types/expressions too.

doc/docusaurus/docs/delve-deeper/casing-constants.md
```diff
@@ -0,0 +1,125 @@
...
+
+In the future, support for `data` is also planned.
+
+For each type, the allowed branches and their order differs. See [Supported
+Types](#supported-types) for more detail.
+
+## Compiling to `case` in Plinth
+
+When compiling Plinth code with the [option](./plinth-compiler-options)
+`datatypes=BuiltinCasing`, many standard library functions will be compiled into
```
zliu41: We need to mention that this option is going away post hard fork (see the comment on the `DatatypeStyle` data type). Also please make an issue to update this post hard fork.

doc/docusaurus/docs/delve-deeper/casing-constants.md
```diff
@@ -0,0 +1,125 @@
...
+
+:::info
+
+This use of `case` in UPLC was introduced with protocol version 11 and requires
+[Plutus Core version](../essential-concepts/versions#plutus-core-version)
+`1.1.0`.
+
+:::
+
+In UPLC, it is possible to branch on expressions of certain built-in types, like
```
zliu41, line 15: We should mention other ways of deconstructing builtin types (e.g., for `bool` one can use `ifThenElse`), and why using the new builtin casing feature is better.

doc/docusaurus/docs/delve-deeper/casing-constants.md
```diff
@@ -0,0 +1,125 @@
...
+A built-in pair expects a single branch: a function that takes both components
+of the pair.
+
+This example sums the two integer constants in a pair.
+
+```uplc
+lam x (case x (lam a (lam b [(builtin addInteger) a b])))
+```
+
+### Integer
```
zliu41, line 99: Mention that the `datatypes=BuiltinCasing` optoin is irrelevant to Integer. To take advantage of casing on integer, in Plinth one must use `PlutusTx.Builtins.caseInteger`.

zliu41 (changes requested):
Our convention for docs is to put each sentence on one line. No need to update this one but please keep it in mind in the future.

doc/docusaurus/docs/delve-deeper/casing-constants.md
basetunnel: Looking at the implementation, I think the `BuiltinCasing` option is necessary. Otherwise it compiles into `!!`.

https://github.com/IntersectMBO/plutus/blob/master/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs#L969

doc/docusaurus/docs/delve-deeper/casing-constants.md
basetunnel: https://github.com/IntersectMBO/plutus-private/issues/2014

doc/docusaurus/docs/delve-deeper/casing-constants.md
basetunnel: I've added some `:::info` about this to most of the types, please check if my understanding is correct here.

zliu41: Looks good!

doc/docusaurus/docs/delve-deeper/casing-constants.md
zliu41: You are right, good catch!

PR #7486, opened 2025-12-12
Metatheory: use Fin instead of nested Maybe for representing variables

basetunnel:
This changes the representation of variables in `Untyped` syntax to `Fin n` (see `Untyped.lagda.md`). The current AST type is parametrised by `X : Set` for variables, meant to be instantiated to `Maybe (Maybe (... ⊥))`, i.e. something isomorphic to `Fin n.` This underconstrained definition led to several work-arounds when @ramsay-t was working on the certifier. 

The original motivation for using `Maybe` was to accomodate for an early version of agda2hs, but the metatheory is actually compiled using the MAlonzo backend nowadays.

Resulting simplifications:

- No more need to parametrise definitions by DecEq instances to compare variables in decision procedures (see e.g. `UntypedTranslation.lagda.md`)
- Many uses of `Set₁` can now simply be `Set`
- Terms are now indexed by a natural number to represent the amount of variables that are in scope, e.g. `2 ⊢` instead of `(Maybe (Maybe ⊥)) ⊢`
- `{-# OPTIONS --injective-type-constructors #-}` is not needed anymore

### DIFF
--- a/plutus-metatheory/README.md
+++ b/plutus-metatheory/README.md
@@ -80,7 +80,7 @@ $ cabal test plutus-metatheory
 To build the documentation as a static site:
 
 ```
-$ agda --html --html-highlight=auto --html-dir=html src/index.lagda.md
+$ agda-with-stdlib --html --html-highlight=auto --html-dir=html src/index.lagda.md
 $ jekyll build -s html -d html/_site
 ```
 
--- a/plutus-metatheory/src/Algorithmic/Erasure/RenamingSubstitution.lagda.md
+++ b/plutus-metatheory/src/Algorithmic/Erasure/RenamingSubstitution.lagda.md
@@ -1,4 +1,4 @@
----)
+---
 title: Algorithmic.Erasure.RenamingSubstitution
 layout: page
 ---


plutus-metatheory/src/Untyped.lagda.md
```diff
@@ -19,10 +19,11 @@ open import Builtin using (Builtin;equals;decBuiltin)
 open Builtin.Builtin
 
 open import Agda.Builtin.String using (primStringFromList; primStringAppend; primStringEquality)
-open import Data.Nat using (ℕ;suc;zero)
+open import Data.Nat using (ℕ;suc;zero;_<_;_<?_)
+open import Data.Fin using (Fin;suc;zero;toℕ;fromℕ<)
 open import Data.Nat.Show using () renaming (show to showℕ)
 open import Data.Bool using (Bool;true;false;_∧_)
-open import Data.Integer using (_<?_;_+_;_-_;∣_∣;_≤?_;_≟_;ℤ) renaming (_*_ to _**_)
+open import Data.Integer using (_+_;_-_;∣_∣;_≟_;ℤ) renaming (_*_ to _**_) -- TODO: remove unnecessary imports
```
ana-pantilie, line 26: Unrelated to this PR, but perhaps there's a warning we can turn on to remove redundant imports. I have a PR where I turn on all warnings and lift them to error to see what can be improved: https://github.com/IntersectMBO/plutus/pull/7318

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
```diff
@@ -39,21 +40,21 @@ open import Untyped.Reduction using (iterApp)
 
 ```
 data CaseReduce : Relation where
-  casereduce : {X : Set} {{_ : DecEq X}} {x : X ⊢} { x' : X ⊢} {vs xs : List (X ⊢)} {i : ℕ}
+  casereduce : {X : ℕ} {x : X ⊢} { x' : X ⊢} {vs xs : List (X ⊢)} {i : ℕ}
```
ana-pantilie, line 43: Oh, interesting. Does this mean that the `DecEq` instance was there to be able to reason on the equality of terms of the type `Maybe (Maybe ... bottom)`?

ana-pantilie (approved):
Nice! I think you need to somehow restart CI since it timed-out. If CI is green, I think we can merge it.

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
basetunnel: Yes exactly, it seems like that was necessary in decision procedures. The situation is much simpler with `Fin`, since it only has a single `DecEq` instance.

plutus-metatheory/src/Untyped.lagda.md
basetunnel: Good idea, I've added an [issue](https://github.com/IntersectMBO/plutus-private/issues/2010) to track that (I couldn't find an existing one)

PR #7489, opened 2025-12-15
Improve UPLC parser error messages

zeme-wana:
See linked [private issue](https://github.com/IntersectMBO/plutus-private/issues/1840).

When using `choice` with multiple alternatives wrapped in `try`, megaparsec can produce misleading or uninformative error messages. This is a fundamental limitation of how megaparsec reports parse errors when backtracking is involved.

The `try` combinator fully backtracks on failure, restoring the input stream to its original position. This means that even if a parser consumed significant input before failing, the input position is reset.

Because `try` backtracks fully, the error message only reflects the outer parser's expectations, not the specific alternative that was attempted. All the detailed error information from deep within a failing alternative is lost.

In order to this fix, we need move the `try` deeper into each choice.

Additionally, Megaparsec uses the longest expected `SrcSpan` when reporting a failure caused by a `choice` with multiple try `branches`. As a result, when matching a list of symbols, the length of the longest string is used to construct the `SrcSpan`, which can make the error message misleading. We can mitigate this by adding a `fail` case with an explicit error message.

### DIFF
--- a/plutus-core/plutus-core.cabal
+++ b/plutus-core/plutus-core.cabal
@@ -489,6 +489,7 @@ library untyped-plutus-core-testlib
     , filepath
     , hedgehog
     , lens
+    , megaparsec
     , mtl
     , plutus-core                      ^>=1.56
     , plutus-core:flat


kwxm:
Note also #7391.

PR #7491, opened 2025-12-16
Costing for improved integer/bytestring conversions

kwxm:
This updates the cost model to reflect the improved performance of `integerToByteString` and `byteStringToInteger` due to #7439: the execution times of both functions are now linear in the sizes of their inputs instead of quadratic: see the `builtinCostModel` JSON files.

For small inputs values (less than size 2 or 3, so about 20 bytes) the new costs are slightly higher than the old ones (possibly due to randomness in the benchmarking process), but for larger inputs the new versions are dramatically faster: see the attached PNGs (blue points are benchmark times for the old versions, green ones are for the new versions). 

The costing functions in this PR are provisional and may be updated before the new cost model is deployed on the chain.

byteStringToInteger

integerToByteString


kwxm:
Update the comments in the budgeting benchmarks.

PR #7494, opened 2025-12-18
Find Use BuiltinList in findDatumHash

zeme-wana:


plutus-ledger-api/src/PlutusLedgerApi/V2/Data/Contexts.hs
```diff
@@ -244,9 +245,11 @@ findDatum dsh TxInfo {txInfoData} = lookup dsh txInfoData
 {-| Find the hash of a datum, if it is part of the pending transaction's
 hashes -}
 findDatumHash :: Datum -> TxInfo -> Maybe DatumHash
-findDatumHash ds TxInfo {txInfoData} = fst <$> List.find f (toSOPList txInfoData)
+findDatumHash ds TxInfo {txInfoData} = do
+  getHash <$> BuiltinList.find matchDatum (toBuiltinList txInfoData)
   where
-    f (_, ds') = ds' == ds
+    getHash = unsafeFromBuiltinData . Builtins.fst
```
bezirg, line 251: Isn't it better to use the safer `fromBuiltinData`?

bezirg: But now that I am thinking more about it, the data will be well-formed coming from the ledger

bezirg (approved):

PR #7496, opened 2025-12-18
Correct definition of psi in finalVerify

kwxm:
This corrects an inaccuracy in the description of the `finalVerify` builtin.  Thanks to Mark Petruska for spotting that.


mpetruska (approved):
:+1:, thank you.

PR #7502, opened 2025-12-22
Implement deriveEnumData

bezirg:
Add to deriveEnum a more efficient implementation of IsData classes.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-tx-plugin/plutus-tx-plugin.cabal
+++ b/plutus-tx-plugin/plutus-tx-plugin.cabal
@@ -115,6 +115,7 @@ test-suite plutus-tx-plugin-tests
   hs-source-dirs:     test
   main-is:            Spec.hs
   other-modules:
+    Enum.Spec
     Array.Spec
     AsData.Budget.Spec
     AsData.Budget.Types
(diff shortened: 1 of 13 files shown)


zliu41 (commented):
Why do we still need `deriveEnum` given `deriveEnumData`?

bezirg:
> Why do we still need `deriveEnum` given `deriveEnumData`?

Because there are existing datatypes like Bool, that have the old less-efficient encoding, and I didn't want to break compatibility with those.

If u think we can get away with breaking the encoding of Bool and () , then deriveEnum can be removed.

I am worried that the ledger-api relies on these old encodings

zliu41:
We can't break the encoding of `Bool` and `()`, but how many types need `deriveEnum`? If there are only a few, then it is far easier to write their instances manually than writing the TH code for `deriveEnum`.

bezirg:
> We can't break the encoding of `Bool` and `()`, but how many types need `deriveEnum`?

About the encoding, i think you are right, no context depends on it , so we can switch to deriveenumdata and get rid of deriveenum

PR #7511, opened 2026-01-07
Add --cse-which-subterms flag

basetunnel:
This allows CSE to consider all sub-expressions. Currently it excludes work-free subexpressions such as partially applied built-ins (https://github.com/IntersectMBO/plutus-private/issues/1761).

- factored out `isWorkFree'` from the logic of `cse`, makes it easier to understand

- instead of `Bool`, I've added a `CseWhichSubterms` type to avoid boolean blindness. I found that it makes the code more understandable. If Bool is preferred for consistency with other simplifier flags, I'm happy to change it.

### DIFF
--- a/plutus-tx-plugin/changelog.d/20260204_153536_4022046+basetunnel_cse_configurable.md
+++ b/plutus-tx-plugin/changelog.d/20260204_153536_4022046+basetunnel_cse_configurable.md
@@ -1,0 +1,3 @@
+### Added
+
+Added the compiler option `--cse-which-subterms` to control which terms the CSE pass of the UPLC simplifier considers


basetunnel:
I've taken a quick look at the five test cases that are there for CSE.

The `cseExpensive` test becomes 10x slower using `AllSubterms`. The other tests are too small to observe a difference:
```
  simplify
    cse
      AllSubterms
        cse1_AllSubterms:                    OK (0.00s)
        cse2_AllSubterms:                    OK (0.00s)
        cse3_AllSubterms:                    OK (0.00s)
        cseExpensive_AllSubterms:            OK (27.91s)
        csePlusTree_AllSubterms:             OK (0.00s)
        cseRepeatPlus_AllSubterms:           OK (0.00s)
      ExcludeWorkFree
        cse1_ExcludeWorkFree:                OK (0.00s)
        cse2_ExcludeWorkFree:                OK (0.00s)
        cse3_ExcludeWorkFree:                OK (0.00s)
        cseExpensive_ExcludeWorkFree:        OK (2.85s)
        csePlusTree_ExcludeWorkFree:         OK (0.01s)
        cseRepeatPlus_ExcludeWorkFree:       OK (0.00s)

```

Output size on disk (in bytes) is slightly better in some cases, but equal in others:

```
259 cse1_AllSubterms.golden.uplc
278 cse1_ExcludeWorkFree.golden.uplc
160 cse2_AllSubterms.golden.uplc
160 cse2_ExcludeWorkFree.golden.uplc
119 cse3_AllSubterms.golden.uplc
119 cse3_ExcludeWorkFree.golden.uplc
617455 cseExpensive_AllSubterms.golden.uplc
617455 cseExpensive_ExcludeWorkFree.golden.uplc
162 csePlusTree_AllSubterms.golden.uplc
162 csePlusTree_ExcludeWorkFree.golden.uplc
54 cseRepeatPlus_AllSubterms.golden.uplc
74 cseRepeatPlus_ExcludeWorkFree.golden.uplc
```
Size on disk is probably not a great measure, but I don't know of a simple way to measure AST size of the outputs.

I'll check why the expensive test gets so much slower, perhaps it's just considering way more sub-expressions.

basetunnel:
The tests from `plutus-tx-plugin:plutus-tx-plugin-tests`, `plutus-benchmark` and `cardano-constitution` resulted in 54 changed `golden.eval` files. On average, there is a slight improvement in CPU en Mem, although size goes up, somewhat surprisingly.

```
--- CPU ---
Mean pct:        -0.11%
Std deviation:   3.9%
Min / Max pct:   -11.11% / +22.50%
Improvements:    32
Regressions:     18
Unchanged:       4

--- MEM ---
Mean pct:        -0.22%
Std deviation:   3.8%
Min / Max pct:   -10.71% / +18.36%
Improvements:    32
Regressions:     18
Unchanged:       4

--- AST ---
Mean pct:        +0.18%
Std deviation:   5.0%
Min / Max pct:   -5.59% / +26.88%
Improvements:    34
Regressions:     13
Unchanged:       7

--- FLAT ---
Mean pct:        +0.75%
Std deviation:   4.1%
Min / Max pct:   -4.85% / +26.15%
Improvements:    11
Regressions:     36
Unchanged:       7
```

basetunnel:
<details>
<summary>CPU changes</summary>
cpu_changes

</details>

<details>
<summary>Mem changes</summary>
mem_changes

</details>

<details>
<summary>AST changes</summary>
ast_changes

</details>

<details>
<summary>FLAT changes</summary>
flat_changes

</details>

basetunnel:
Looking at the diff in UPLC of [8 queens](https://github.com/IntersectMBO/plutus/compare/basetunnel/cse-configurable...basetunnel/cse-configurable-tests?diff=split&w#diff-f14fcb30c640562834e9043ca8b24c404f9f6be76397eaaaaf1adfb5ebac0cf1) and [Ed225119](https://github.com/IntersectMBO/plutus/compare/basetunnel/cse-configurable...basetunnel/cse-configurable-tests?diff=split&w#diff-76e6d4460de0b6d0b4f88030ab64d2465f82fd335220efd3cb0925826b38abcd) I find it hard to see what happened and why. Perhaps a bad interaction between inlining and cse?

zliu41:
Thanks for the data. I'm surprised that there are CPU improvements when allowing workfree terms. As I noted before,

```
      -- We don't consider work-free terms for CSE, because doing so may or may not
      -- have a size benefit, but certainly doesn't have any cost benefit (the cost
      -- will in fact be slightly higher due to the additional application).
```


Do you know why we are seeing CPU improvements? If not, it's worth looking into the `or` test case, which is fairly small.

We should also update `Note [CSE]`. There are a few mentions of "workfree" there.

zliu41:
>  If not, it's worth looking into the `or` test case, which is fairly small.

Also, does the `or` test case really happen to have the same AST sizes before and after? It's hard to count by eye but that would be some coincidence.

basetunnel:
> Do you know why we are seeing CPU improvements?

CSE is interleaved with some simplifier passes, so it seems to be the result of an interaction between CSE and inlining. I don't think the speed-up is due to a CSE transformation only.

Looking at the `or` case, the only difference is that the simplifier produced different code for the Z combinator. The code that it used to produce for Z (using `let` for readability) was:

```haskell
\f ->
    let t = \s -> f (\x -> s s x)
    in
    f (\x ->
        f (\x ->
            f (\x ->
                f (\x ->
                    t t x
                  ) x
              ) x
          ) x
      )
```

Note that it was unrolled four times. Using CSE with work-free expressions we get:
```haskell
\f ->
    f (\x ->
        f (\x ->
            f (\x ->
                let t = \s -> f (\x -> s s x)
                in
                    f (\x -> t t x)
              ) x
          ) x
      )
```

It's still unrolled four times, but the `let` ended up deeper down in the recursion. Both terms have the same AST size.

The evaluation evaluation speed-up is due to the argument that `or` is tested on: `or [True, False, True, False]`. It doesn't have to recurse, because it is done after seeing the head. Therefore, the `let t = \s -> f (\x -> s s x)` is not evaluated anymore (which will only happen after at least 3 times of recursing).

I think `t` was pushed down compared to the original because it was inlined and then CSE'd again, more locally to its use sites.

So it seems that there is a slight improvement for the unrolled Z combinator, when the number of recursive calls is limited. But seeing the amount of slowdown on some of the larger tests like 8-queens makes me wonder if it is worth it to make work-free CSE the default. I can spend some time to investigate what is going on there.

By the way, is it intentional/desirable that each run of the inliner unrolls the Z combinator one more time?

zliu41:
> CSE is interleaved with some simplifier passes, so it seems to be the result of an interaction between CSE and inlining.

Makes sense.

> By the way, is it intentional/desirable that each run of the inliner unrolls the Z combinator one more time?

Yes, because the default value of `inline-callsite-growth` is 5, it allows the callsite inliner to slightly increase the script size at each callsite, in the hope that inlining can unlock further optimizations (in this case it doesn't).

By the way, this test has both PIR and UPLC simplifiers turned off, and I don't think it should. If PIR simplifier is on, then this behavior does not occur.

Please do the following:
- remove the `max-simplifier-iterations-pir=0` and `max-simplifier-iterations-uplc=0` flags in `BuiltinList.Budget.Spec`.
- In "Note [CSE]", there are two occurrences of "workfree", and both should be updated. We now have a flag that allows the pass to CSE workfree terms, so the Note is no longer up to date.
- Rebase on master so CI can run properly.

basetunnel:
> remove the max-simplifier-iterations-pir=0 and max-simplifier-iterations-uplc=0 flags in BuiltinList.Budget.Spec.

After doing this, the cse flag doesn't seem to have any effect except that more variables are now named `cse` [diff](https://github.com/IntersectMBO/plutus/compare/basetunnel/cse-configurable...basetunnel/cse-configurable-builtinlist), which suggests it's doing unnecessary CSE/inlining.

> In "Note [CSE]", there are two occurrences of "workfree", and both should be updated. We now have a flag that allows the pass to CSE workfree terms, so the Note is no longer up to date.

Done.

By the way, the comments have

> The first pass collects builtin arity information as described above.

But I don't see it described above. I also don't think there is a separate pass for that, right? I think we should mention that the first thing that happens is a renaming pass.

zliu41:
There are compilation errors that need to be fixed

basetunnel:
Opened https://github.com/IntersectMBO/plutus-private/issues/2059 to track 8-queens slowdown/increased memory usage.

PR #7513, opened 2026-01-09
Add abs to memoryUsage for `NumBytesCostedAsNumWords`

kwxm:
The `ExMemoryUsage` instance for `NumBytesCostedAsNumWords` was missing an `abs`, meaning that it could return a negative value for negative numbers.  This didn't actually matter because the only builtins using `NumBytesCostedAsNumWords` fail on negative inputs, but let's make sure that the value is always non-negative in case we use it for anything else later.  Tests for non-negativity will follow in a later PR.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
@@ -280,7 +280,7 @@ instance ExMemoryUsage Word8 where
 newtype NumBytesCostedAsNumWords = NumBytesCostedAsNumWords {unNumBytesCostedAsNumWords :: Integer}
 
 instance ExMemoryUsage NumBytesCostedAsNumWords where
-  memoryUsage (NumBytesCostedAsNumWords n) = singletonRose . fromIntegral $ ((n - 1) `div` 8) + 1
+  memoryUsage (NumBytesCostedAsNumWords n) = singletonRose . fromIntegral $ ((abs n - 1) `div` 8) + 1
   {-# INLINE memoryUsage #-}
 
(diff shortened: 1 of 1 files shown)


plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemoryUsage.hs
```diff
@@ -280,7 +280,7 @@ instance ExMemoryUsage Word8 where
 newtype NumBytesCostedAsNumWords = NumBytesCostedAsNumWords {unNumBytesCostedAsNumWords :: Integer}
 
 instance ExMemoryUsage NumBytesCostedAsNumWords where
-  memoryUsage (NumBytesCostedAsNumWords n) = singletonRose . fromIntegral $ ((n - 1) `div` 8) + 1
+  memoryUsage (NumBytesCostedAsNumWords n) = singletonRose . fromIntegral $ ((abs n - 1) `div` 8) + 1
```
zliu41, line 283: Why do we minus 1 before `` `div` 8 ``? And shouldn't be `abs (n - 1)`? Otherwise `memoryUsage` could return 0, but it should be at least 1.

kwxm: It already returns 0 for n=0, but that's OK because of [this](https://github.com/IntersectMBO/plutus/blob/7f9012dba5d044cdb8b32365a2705eee9928c84a/plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs#L150) and [this](https://github.com/IntersectMBO/plutus/blob/7f9012dba5d044cdb8b32365a2705eee9928c84a/plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs#L163).

PR #7517, opened 2026-01-12
Add new costing function to metatheory

SeungheonOh:


SeungheonOh:
If anyone can force merge, please feel free to. seems like CI is stuck again

PR #7467, opened 2025-11-28
Make `Builtins.unsafeUncons` use builtin list casing

sohang3112:
avoid seperate head, tail calls in `Builtins.unsafeUncons`. No Changelog Required. Fixes #7459 

Pre-submit checklist:
- Branch
    - [x] Tests are provided (if possible)
    - [x] Commit sequence broadly makes sense
    - [x] Key commits have useful messages
    - [x] Changelog fragments have been written (if appropriate)
    - [x] Relevant tickets are mentioned in commit messages
    - [x] Formatting, PNG optimization, etc. are updated
- PR
    - [x] (For external contributions) Corresponding issue exists and is linked in the description
    - [x] Targeting master unless this is a cherry-pick backport
    - [x] Self-reviewed the diff
    - [x] Useful pull request description
    - [x] Reviewer requested

PR #7519, opened 2026-01-13
Make `Builtins.unsafeUncons` use builtin list casing 

SeungheonOh:
Same as #7467 but new PR because I don't have push access to remote used in the original PR.


SeungheonOh:
More broken CI 😢

PR #7520, opened 2026-01-13
Add memory models for valueData and unValueData

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/2018

Depends on `DataNodeCount` from @kwxm 's branch: https://github.com/IntersectMBO/plutus/compare/kwxm/costing/fix-valueData

### DIFF
--- a/plutus-core/changelog.d/20260113_231029_ana.pantilie95_value_unvalue_data_mem_costing.md
+++ b/plutus-core/changelog.d/20260113_231029_ana.pantilie95_value_unvalue_data_mem_costing.md
@@ -1,0 +1,4 @@
+### Added
+
+- Added memory cost models for the `valueData` and `unValueData` builtins.
+


kwxm:
I think the easiest way to deal with this would just be  to extract the parts that I need and copy them into my branch.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -164,6 +164,97 @@ size must be allocated:
...
+    Memory = 21*((n - 3)/2) + 12 = 10.5*n - 19.5 --approx--> 11*n - 19
+
+where 'n' is the number of 'Data' nodes in the input structure.
+
+However, this formula can yield negative results for small 'n' (specifically n < 4).
+Due to limitations in our costing framework, we cannot case on the value of 'n' to handle
+such situations. To ensure non-negative memory costs, we further overapproximate the formula
+to remove the negative intercept:
+
+    Memory = 21*(n/2) + 12 = 10.5*n + 12 --approx--> 11*n + 12
```
zliu41, line 255: If the problem is negative intercept, why not just make intercept 0?

kwxm:
> I think the easiest way to deal with this would just be to extract the parts that I need and copy them into my branch.

I've done this now: other things had changed in `plutus-ledger-api` and just adding the changes that I needed (the numbers in the JSON files and the models and the explanatory Note in `BuiltinMemoryModels.hs`) saved me from some conflicts.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
ana-pantilie: If the intercept is 0 then for an empty result the memory will be `0` which I don't think we want. Also, I don't really like the way I had to approximate things here, ideally we'd have a way to differentiate between the two cases (empty and non-empty) but I don't think that's possible with the current costing framework.

ana-pantilie:
@kwxm should I just close this PR then?

kwxm:
> @kwxm should I just close this PR then?

@ana-pantilie Yes please: I think everything we needs in the other branch now.  Thanks for working all this out!

kwxm:
These changes were incorporated in #7522 (now merged), so I'll close this.

PR #7521, opened 2026-01-14
test(conformance): add valueData and unValueData builtin tests

Unisay:
## Context

- **What**: Add conformance tests for `valueData` and `unValueData` builtins
- **Why**: Ensure cross-implementation compatibility and validate conversion behavior between Value and Data formats
- **Issue**: Closes IntersectMBO/plutus-private#2023

## Approach

Implemented comprehensive conformance test suite covering the `valueData` and `unValueData` builtins introduced in the Plutus Core specification. These builtins enable conversion between the native `Value` type and its `Data` representation, which is critical for on-chain script interoperability.

The test suite follows the existing conformance testing patterns:
- Input test programs (`.uplc` files) created manually with representative test cases
- Expected outputs (`.uplc.expected`) and execution budgets (`.uplc.budget.expected`) auto-generated by running the evaluator with the `--accept` flag
- Tests organized by builtin function with subdirectories for each test case

Since the Agda metatheory implementation does not yet support these builtins, all tests are added to the `failingEvaluationTests` and `failingBudgetTests` lists in the Agda conformance spec. This prevents CI breakage while maintaining test coverage for the Haskell implementation.

### Review Comments Addressed

**Round-trip tests (requested by both reviewers):**
- ✅ Added `valueData/roundtrip-from-value`: tests `unValueData ∘ valueData`
- ✅ Added `unValueData/roundtrip-from-data`: tests `valueData ∘ unValueData`

**Data normalization tests for `unValueData` (requested by kwxm):**
- ✅ `data-duplicate-tokens`: duplicate token entries sum quantities
- ✅ `data-duplicate-currencies`: duplicate currency entries merge token maps
- ✅ `data-zero-quantity`: zero quantities are removed
- ✅ `data-zero-sum`: tokens summing to zero are removed
- ✅ `data-empty-tokens`: empty token lists are removed
- ✅ `data-unordered`: unordered Data is sorted

**Test organization (kwxm's comment on max-key-len):**
- ✅ Removed `valueData/max-key-len`: the `maxKeyLen` constraint is enforced at the Value type level (via the `K` smart constructor), not by the builtin. Existing `constant/value/max-key-length-*` tests already cover this.

### Original Test Suite

**`valueData` (Value → Data) - 5 test cases:**
- `empty`: Empty value converts to empty Map
- `single-entry`: Single currency/token entry
- `multi-token`: Multiple tokens within single currency
- `multi-currency`: Multiple currencies with tokens
- `negative-quantity`: Negative token quantities

**`unValueData` (Data → Value) - 18 test cases:**

*Success cases (6 tests):*
- Empty Map → empty Value
- Single entry conversion
- Multi-token/multi-currency conversions
- Negative quantities
- Maximum key lengths (32 bytes)

*Failure cases (12 tests):*
- Non-Map data types (Integer, Constr, List, ByteString)
- Currency keys exceeding 32 bytes
- Token keys exceeding 32 bytes
- Quantity overflow (> 2^127-1)
- Quantity underflow (< -2^127)
- Invalid nested structure types (non-ByteString keys, non-Integer quantities, non-Map token maps)

### Agda Conformance Configuration

Updated `plutus-conformance/agda/Spec.hs` to mark all 31 tests (24 original + 7 new) as expected failures for the Agda evaluator until the metatheory implementation adds support for these builtins.

## How to Test

**Run all conformance tests:**
```bash
cabal test plutus-conformance
```

**Run specific test suites:**
```bash
# Haskell implementation (should pass all tests)
cabal test plutus-conformance:haskell-conformance

# Agda implementation (new tests marked as expected failures)
cabal test plutus-conformance:agda-conformance
```

## Test Coverage

✅ **Value → Data conversions**
- All success cases with various Value structures
- Proper Map nesting: `Map[(B currency, Map[(B token, I quantity)])]`

✅ **Data → Value conversions**
- All success cases with proper type validation
- All failure cases covering validation paths

✅ **Boundary validation**
- 32-byte key length limit (33 bytes correctly rejected)
- 128-bit signed integer bounds (2^127-1 max, -2^127 min)

✅ **Type validation**
- Rejects non-Map outer data
- Rejects invalid nested types (non-B keys, non-I quantities)

✅ **Roundtrip properties**
- `unValueData(valueData(v)) = v` for all valid Values
- `valueData(unValueData(d)) = d` for all valid Data

✅ **Data normalization in `unValueData`**
- Duplicate entries summed correctly
- Zero quantities removed
- Empty token lists removed
- Unordered maps sorted


plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/unValueData/multi-token/multi-token.uplc.expected
kwxm, line 1: Maybe this should go in the `constant` directory, along with some others.

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/valueData/max-key-len/max-key-len.uplc
kwxm, line 1: Arguably this (and quite a lot of the other tests) is testing the constants rather than the builtins.

Unisay:
Thank you for the review! I've addressed all the feedback:

## Round-trip Tests (ana-pantilie's request)

Added both directions of round-trip testing:
- `valueData/roundtrip-from-value/` - tests `unValueData ∘ valueData`
- `unValueData/roundtrip-from-data/` - tests `valueData ∘ unValueData`

## Data Normalization Tests (kwxm's request)

Added comprehensive normalization coverage:

1. **Duplicate handling**:
   - `data-duplicate-tokens/` - duplicate token names sum quantities (100 + 50 = 150)
   - `data-duplicate-currencies/` - duplicate currency symbols merge token maps

2. **Zero quantity handling**:
   - `data-zero-quantity/` - individual zero quantities removed
   - `data-zero-sum/` - tokens summing to zero remove entire currency

3. **Empty token lists**:
   - `data-empty-tokens/` - currencies with empty token maps removed

4. **Unordered data**:
   - `data-unordered/` - verifies output is properly sorted

## Redundant Test Removal (kwxm's request)

Removed `valueData/max-key-len/` test as it's redundant with existing `constant/value/max-key-length-*` tests. The maxKeyLen constraint is a Value type property, not specific to the valueData builtin.

## Agda Expected Failures

All new tests added to `plutus-conformance/agda/Spec.hs` expected failures lists since these builtins aren't yet implemented in the Agda metatheory.

All conformance tests now pass (1976 each for haskell, haskell-steppable, and agda).

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/valueData/max-key-len/max-key-len.uplc
kwxm: It's a little tricky to keep track of all the tests, but I think you may have missed a case.  In `unValueData`, if you have a currency with multiple entries for the same token then the quantities are added together; but something similar happens one level up for duplicate currencies: consider

```
(program 1.0.0
  [(builtin unValueData)
    (con data (Map [
                     (B #aa, Map [(B #aa, I 123)])
                   , (B #bb, Map [(B #bb, I 2)])
                   , (B #aa, Map [(B #aa, I -80)])
                   , (B #cc, Map [(B #cc, I 2)])
                   , (B #aa, Map [(B #aa, I -43)])]))]
)
```
In this case the entries for currency `aa` should cancel out so you get
```
(con value [(#bb, [(#bb, 2)]), (#cc, [(#cc, 2)])])
```
and if all of the quantitities were positive we'd get 
```
(con value [(#aa, [(#aa, 246)]), (#bb, [(#bb, 2)]), (#cc, [(#cc, 2)])])
```

I don't think those cases are covered.

Unisay: I see what you mean, but I'm not sure the additional test is necessary.

The scenario you describe is a composition of two behaviors we already test independently:
1. `data-duplicate-currencies` — duplicate currency entries get merged
2. `data-zero-sum` — tokens that sum to zero cause the currency to be removed

If the implementation handles these two primitives correctly, the composite case (merge → cancel out) should work automatically. I can't think of a plausible implementation bug where the composite fails but both primitive tests pass.

That said, I'm not strongly against adding it if you feel it adds clarity to the test suite — just questioning whether it's pulling its weight.

kwxm: > The scenario you describe is a composition of two behaviors we already test independently:

@Unisay I see what you mean, but I'm still a little nervous.  It's conceivable that an implementation could perform the removal of tokens that sum to zero inside a single currency _before_ it merges repeated entries for the same currency, and then during the merge you could end up with repeated tokens that don't get merged/removed.  It might also clarify the expected behaviour if we add an explicit test.  Could you just add the [example](https://github.com/IntersectMBO/plutus/pull/7521#discussion_r2697004234) I gave above, and the same thing with the `-` signs deleted to check that tokens do get merged after currencies are merged (`data-duplicate-currencies` merges two entries for the same currency, but doesn't check that duplicate tokens get merged after that happens)?

The costing PR for `valueData` and `unValueData` is merged now, so I think that if you update the budgets for these conformance tests based on that the PR should be ready to merge.

Unisay: I've added 2 test cases as per your request, see [this commit](https://github.com/IntersectMBO/plutus/pull/7521/changes/f6968cacfe038dc81ba9106077afc0c84b20cb82)

PR #7522, opened 2026-01-14
Finalise cost models for `valueData` and `unValueData`

kwxm:
This PR

  * Uses `nf` in the costing benchmarks for `valueData` so that it doesn't look as if it's constant time
  * Uses a quadratic costing function for `unValueData` since it doesn't seem to be quite linear.
  * Updates the memory costs for `valueData` and `unValueData` to use @ana-pantilie's figures from #7520.
  * Updates the cost models to use models obtained from the new benchmarks.
  * Updates the cost model parameter names to include all of the new builtins.

There might be a couple of minor things that I need to update, but I think it's pretty much complete.

### DIFF
--- a/plutus-core/untyped-plutus-core/testlib/Generators/Spec.hs
+++ b/plutus-core/untyped-plutus-core/testlib/Generators/Spec.hs
@@ -8,7 +8,6 @@ module Generators.Spec where
 import PlutusPrelude (display, fold, void, (&&&))
 
 import Control.Lens (view)
-import Control.Monad (unless)
 import Data.Text (Text)
 import Data.Text qualified as T
 import Hedgehog (annotate, annotateShow, failure, property, tripping, (===))


plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -194,18 +208,47 @@ valueContainsArgs gen = runStateGen_ gen \g -> do
...
 unValueDataBenchmark gen =
-  createOneTermBuiltinBench UnValueData [] (Value.valueData <$> generateTestValues gen)
+  createOneTermBuiltinBenchWithWrapper
+    DataNodeCount
+    UnValueData
+    []
+    (fmap (f . Value.valueData) $ generateTestValues gen)
+  where
+    f (Data.Map l) = Data.Map (reverse l)
+    f _ = error "NO"
```
ana-pantilie, line 286: I'd rename `f` to something more suggestive and improve the error mesage.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -194,18 +208,47 @@ valueContainsArgs gen = runStateGen_ gen \g -> do
...
 
+{-
+  where
+    reverseOuterMap (Data.Map l) = Data.Map $ reverse (fmap reverseInnerMap l)
+    reverseOuterMap d = error ("Unexpected item in reverseOuterMap: " ++ show d)
+    reverseInnerMap (t, Data.Map l) = (t, Data.Map $ reverse l)
+    reverseInnerMap (_, d) = error ("Unexpected item in reverseInnerMap: " ++ show d)
+-}
```
ana-pantilie: I think you forgot to remove this commented out code.

```suggestion
```

plutus-core/cost-model/budgeting-bench/Common.hs
```diff
@@ -73,19 +95,32 @@ pairWith f = fmap (\a -> (a, f a))
 
 ---------------- Creating benchmarks ----------------
 
+-- Choose whether to benachmark with `whnf` or `nf`.  Note that to get sensible
```
ana-pantilie:
```suggestion
-- Choose whether to benchmark with `whnf` or `nf`.  Note that to get sensible
```

plutus-core/cost-model/budgeting-bench/Common.hs
```diff
@@ -73,19 +95,32 @@ pairWith f = fmap (\a -> (a, f a))
...
+-- results with 'whnf', we must use an evaluation function that looks at the
+-- result, so e.g. 'evaluateCek' won't work properly because it returns a pair
+-- whose components won't be evaluated by 'whnf'. In general we'll want to use
+-- `whnf` because 'nf' willl do too much work (for instance if it gets back a
```
ana-pantilie:
```suggestion
-- `whnf` because 'nf' will do too much work (for instance if it gets back a
```

plutus-core/cost-model/budgeting-bench/Common.hs
```diff
@@ -276,6 +314,51 @@ createOneTermBuiltinBenchWithWrapper wrapX fun tys xs =
...
+  -> Benchmark
+createOneTermBuiltinBench_NF = createOneTermBuiltinBenchWithWrapper_NF id
+
+{- Note [Adjusting the memory usage of arguments of costing benchmarks] In some
+  cases we want to measure the (so-called) "memory usage" of a builtin argument
+  in a nonstandard way for benchmarking and costing purposes. This function
+  allows you to supply suitable wrapping functions in the benchmarks to achieve
+  this.  NB: wrappers used in benchmarks *MUST* be the same as wrappers used in
+  builtin denotations to make sure that during script execution the inputs to
+  the costing functions are costed in the same way as the are in thhe
```
ana-pantilie:
```suggestion
  the costing functions are costed in the same way as the are in the
```

plutus-core/cost-model/data/builtinCostModelA.json
```diff
@@ -1240,28 +1240,38 @@
...
-      "arguments": 194713,
-      "type": "constant_cost"
-    },
+        "cpu": {
+            "arguments": {
+                "intercept": 199604,
+                "slope": 39211
+            },
+            "type": "linear_in_x"
+        },
```
ana-pantilie, line 1249: The formatting is messed up here. It's a bit annoying that we use 2 spaces per tab in the cost model jsons and the executable which generates the new cost model jsons introduces 4 spaces per tab and I always have to fix that.

plutus-core/cost-model/data/builtinCostModelA.json
```diff
@@ -1240,28 +1240,38 @@
...
+            "type": "quadratic_in_x"
+        },
+      "memory": {
+          "arguments": {
+              "intercept": 12,
+              "slope": 11
+          },
+          "type": "linear_in_x"
+      }
+  },
```
ana-pantilie, line 1274: Same as above, too many spaces.

plutus-core/cost-model/data/builtinCostModelB.json
```diff
@@ -1240,27 +1240,37 @@
...
-      "arguments": 1,
-      "type": "constant_cost"
-    }
+      "memory": {
+          "arguments": {
+              "intercept": 12,
+              "slope": 11
+          },
+          "type": "linear_in_x"
+      }
```
ana-pantilie, line 1273: Same as above.

plutus-core/cost-model/data/builtinCostModelC.json
```diff
@@ -1258,26 +1258,36 @@
...
+                "c2": 1
+            },
+            "type": "quadratic_in_x"
+        },
+    "memory": {
       "arguments": {
-        "intercept": 1000,
-        "slope": 43200
+          "intercept": 12,
+          "slope": 11
```
ana-pantilie: Same as above.

plutus-core/cost-model/data/models.R
```diff
@@ -218,6 +218,8 @@ adjustModel <- function (r, fname) {
     ## prevent us from getting models which predict negative costs.  See also
     ## https://stackoverflow.com/questions/27244898/force-certain-parameters-to-have-positive-coefficients-in-lm
 
+    ## cat (sprintf ("Adjusting model for %s\n", fname))
+
```
ana-pantilie, line 222: Should this be removed?

plutus-core/cost-model/data/models.R
```diff
@@ -856,6 +858,7 @@ modelFun <- function(path) {
         # Use mean observed time as constant for this case
         above_diag <- filtered %>% filter(y_mem > x_mem)
         constant <- ceiling(mean(above_diag$t))
+        constant <- -1  ## FIXME
```
ana-pantilie: Does this need to be fixed and the parameters regenerated?

plutus-core/cost-model/data/models.R
```diff
@@ -864,9 +867,16 @@ modelFun <- function(path) {
...
-    valueDataModel            <- constantModel ("ValueData")
-    unValueDataModel          <- linearInX ("UnValueData")
+    valueDataModel <- linearInX ("ValueData")
+
+    ## X wrapped with DataNodeCount
+    unValueDataModel   <- {
+        fname <- "UnValueData"
+        filtered <- data %>%
+            filter.and.check.nonempty(fname)
+        m <- lm(t ~ I(x_mem) + I(x_mem^2), filtered)
```
ana-pantilie, line 876: Isn't this the same as the formula `x_mem^2`?

plutus-core/cost-model/test/TestCostModels.hs
```diff
@@ -284,7 +300,7 @@ testPredictFour costingFunH modelR =
             wD = fromSatInt w :: Double
            in
             microToPico . fromSomeSEXP
-              <$> [r|predict(modelR_hs$model, data.frame(x_mem=xD_hs, y_mem=yD_hs, z_mem=zD_hs, w_mem=wD_hs))[[1]]|]
+              <$> [r|predict(modelR_hs$model, data.frame(x_mem=xD_hs, y_mem=yD_hs, z_mem=zD_hs, u_mem=wD_hs))[[1]]|]
```
ana-pantilie, line 303: Thanks for catching this!

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
```diff
@@ -744,7 +766,7 @@ runThreeArgumentModel
       scaleLinearly intercept slope costs3
 runThreeArgumentModel
   (ModelThreeArgumentsQuadraticInZ f) =
-    lazy $ \_ _ costs3 -> CostLast $ evaluateOneVariableQuadraticFunction f $ sumCostStream costs3
+    lazy $ \_ _ costs3 -> scaleQuadratically f costs3
```
ana-pantilie, line 769: Does this change the parameters for the functions which use this model?

plutus-core/cost-model/budgeting-bench/Benchmarks/Nops.hs
```diff
@@ -34,7 +42,7 @@ import System.Random (StdGen)
    the evaluator. -}
 benchUnitTerm :: Benchmark
 benchUnitTerm =
-  bgroup "UnitTerm" [benchWith nopCostParameters (showMemoryUsage ()) $ mkUnit]
+  bgroup "UnitTerm" [benchWithWHNF (showMemoryUsage ()) $ mkUnit]
```
Unisay:
```suggestion
  bgroup "UnitTerm" [benchWithWHNF (showMemoryUsage ()) mkUnit]
```

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -194,18 +208,47 @@ valueContainsArgs gen = runStateGen_ gen \g -> do
...
+-- `unValueData`.  We reverse the lists to make the inputs less favourable (and
+-- possibly worst-case).
 unValueDataBenchmark :: StdGen -> Benchmark
 unValueDataBenchmark gen =
-  createOneTermBuiltinBench UnValueData [] (Value.valueData <$> generateTestValues gen)
+  createOneTermBuiltinBenchWithWrapper
+    DataNodeCount
+    UnValueData
+    []
+    (fmap (f . Value.valueData) $ generateTestValues gen)
```
Unisay:
```suggestion
    (f . Value.valueData <$> generateTestValues gen)
```

plutus-core/cost-model/budgeting-bench/Common.hs
```diff
@@ -8,20 +8,42 @@
 module Common
 where
 
-import PlutusCore hiding (Constr)
+import PlutusCore hiding
+  ( Constr
+  )
```
Unisay, line 13: I wonder if you manually formatted it as my `fourmolu` doesn't change the original formatting.

plutus-core/cost-model/budgeting-bench/Common.hs
```diff
@@ -276,6 +314,51 @@ createOneTermBuiltinBenchWithWrapper wrapX fun tys xs =
...
+     , uni `HasTermLevel` a
+     , ExMemoryUsage a'
+     , NFData a
+     )
+  => (a -> a')
+  -> fun
+  -> [Type tyname uni ()]
+  -> [a]
+  -> Benchmark
+createOneTermBuiltinBenchWithWrapper_NF wrapX fun tys xs =
```
zliu41, line 359: There's no need to duplicate the code for `createOneTermBuiltinBenchWithWrapper`, if you make `Normaliser` an argument to `benchDefault`. Can you add a FIXME?

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
```diff
@@ -164,7 +164,96 @@ size must be allocated:
...
+From this we can derive that the memory allocated for the resulting 'Value' is:
+
+    Memory = 21*((n - 3)/2) + 12 = 10.5*n - 19.5 --approx--> 11*n - 19
+
+where 'n' is the number of 'Data' nodes in the input structure.
+
+However, this formula can yield negative results for small 'n' (specifically n < 4).
+Due to limitations in our costing framework, we cannot case on the value of 'n' to handle
+such situations. To ensure non-negative memory costs, we further overapproximate the formula
+to remove the negative intercept:
```
zliu41, line 253: Can we make the intercept 0 or 1, instead of 12?

plutus-core/cost-model/data/models.R
kwxm: > Does this need to be fixed and the parameters regenerated?

Yes. I was waiting for some new benchmark results (there were some missing numbers) and I've updated it now.

plutus-core/cost-model/data/models.R
kwxm: > Isn't this the same as the formula `x_mem^2`?

No: that's just R trying to confuse us with its model formulas.  If you put `t ~ I(x_mem^2)` then you get a result of the form `a*(x_mem)^2 + b`, and if you put `t ~ x_mem^2` you get a result of the form `a+b*(x_mem)`.  I think if you put `x^2` it interprets it in the same way as `x*x`, which gives you something with an interaction between `x` and `x`, but that's zero for some reason.

plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs
kwxm: > Does this change the parameters for the functions which use this model?

No: this performs the cost computation on the cost stream incrementally and the previous version summed up the cost stream up front then performed the cost computation.  All of the types we used this on before had cost stream with just a single entry though: it's just `data` that we have to worry about.

plutus-core/cost-model/data/models.R
kwxm: I think I'll leave that in.  There are a few other things like this as well.  If something goes wrong when the R code's being executed this helps to work out where it's going wrong, and it's easier to leave the commented code there rather than trying to remember what you need to add.  It might be better to have a global `debug` variable that turns these on though.

plutus-core/cost-model/budgeting-bench/Common.hs
kwxm: Not on purpose. I need to make my emacs setup deal with fourmolu properly though.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
kwxm: Thanks!

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
kwxm: > Can we make the intercept 0 or 1, instead of 12?

I'm not sure. @ana-pantilie ?

plutus-core/cost-model/budgeting-bench/Common.hs
kwxm: I had some reason for doing that, but it may just have been that I was in a hurry, or possibly I didn't want to change the existing code because we only need this for one case out of 100.  I've added a FIXME to remind us to look at this again though.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
ana-pantilie: As I said in the [original PR](https://github.com/IntersectMBO/plutus/pull/7520#discussion_r2689967884): If the intercept is 0 then for an empty result the memory will be 0 which I don't think we want. Also, I don't really like the way I had to approximate things here, ideally we'd have a way to differentiate between the two cases (empty and non-empty) but I don't think that's possible with the current costing framework.

Making the intercept 1 will probably be closer to the actual result, but I don't like any of these approaches honestly.

zliu41: If the goal is to fix negative cost due to the intercept being negative, then naturally the intercept should be raised to the smallest possible positive number. 12 seems totally arbitrary.

> ideally we'd have a way to differentiate between the two cases (empty and non-empty)

What does "empty" mean?

kwxm:
I've now added @Unisay's improved benchmark inputs for `ValueContains` (PR #7527) in [this commit](https://github.com/IntersectMBO/plutus/pull/7522/commits/a25243de8a1d400bccaf56bcdfa9df4d4599ea5b).  I've added a check for total sizes in `valueContains` [here](https://github.com/IntersectMBO/plutus/blob/a25243de8a1d400bccaf56bcdfa9df4d4599ea5b/plutus-core/plutus-core/src/PlutusCore/Value.hs#L385): it speeds things up in the case when the second input has more nodes in total than the first one.

kwxm:
I think this is almost ready to merge, but we should sort out the question about the memory models first.

plutus-core/cost-model/create-cost-model/BuiltinMemoryModels.hs
ana-pantilie: When the intercept is negative for a CPU model, we replace that with 1000. I don't know what we want to do for memory models. Maybe Kenneth has a better idea about this.

ana-pantilie: > What does "empty" mean?

When the result is an empty map

kwxm: > Making the intercept 1 will probably be closer to the actual result, but I don't like any of these approaches honestly.

Earlier it says that `11*n - 19` is an approximation to the memory usage of `unValueData`, but then adjusts it to a positive over-approximation `11*n+12`; but `11*n+1` will still overapproximate `11*n-19`, so I think I agree with @zliu41 that an intercept of 1 would  be safe.  For n > 2 that'll cost 20 units more than we'd ideally like, but given that we're charging 32 units for things like `FstPair` and `unBData`  that's not too bad.  I doubt if people will be calling `unValueData` dozens of times, so an extra 20 units a few times won't be too bad.

> ideally we'd have a way to differentiate between the two cases (empty and non-empty)

We could do something like that.  For example, for `divideInteger` and related functions we need a two-variable quadratic function with a negative coefficient, which means that there's a danger that the costing function could return a negative number.  To [guard against that](https://github.com/IntersectMBO/plutus/blob/0d0dfa40521adff32f27fa87b7a6fb0e6294728a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/CostingFun/Core.hs#L417) the cost model also includes a minimum value that's returned if the quadratic function returns something too small.  We could do something similar here: evaluate a linear function like `11n-19` and if the result's less than or equal to zero, return 10 or something.   However, I think that's probably not a good idea: we should keep negative numbers out of the cost parameters unless it's absolutely necessary and in this case it's probably worth paying an extra 20 units now and again instead of trying to be too precise.  I admit that this isn't ideal, but the whole memory costing thing's a crude overapproximation anyway and I'm not sure if it can be improved significantly given that we can't take account of GC or anything.

kwxm: > 11*n+1 will still overapproximate 11*n-19

In fact we could just make it 11n.  The size of the input `data` object is measured by the number of nodes it contains, and every `data` object has at least one node (in the case of the empty `value`, it'll be `Map []`), so we don't have to worry about the n=0 case.

plutus-core/cost-model/data/builtinCostModelA.json
kwxm: Yes, that annoys me too.  I'm not sure why it happens.

kwxm:
> I think this is almost ready to merge, but we should sort out the question about the memory models first.

We eventually settled on a memroy cost of 11n+1 memory units for `unValueData`.  This overestimates the cost by about n/2+20.5, but that seems acceptable given that a single CEK step costs 100 units.

PR #7523, opened 2026-01-14
Plinth guide: more details on versions

basetunnel:
This includes a section about protocol parameters, including protocol version which was missing. I've also extended the intro a bit.


doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -5,6 +5,12 @@ sidebar_position: 10
 # Different Notions of Version
 
 There are several different notions of version that Cardano smart contract developers must distinguish.
+For each Plutus script that is deployed on-chain, you can determine its associated __protocol parameters__ (including __protocol version__), __ledger language version__ and __Plutus Core version__.
```
zliu41: - A script does not have associated protocol parameters. A script _execution_ has associated protocol parameters. A deployed script can be executed many times with different protocol parameters.
- It is strange to say "you can determine" without saying how. I suggest removing "you can determine".

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -5,6 +5,12 @@ sidebar_position: 10
...
+These all influence which language features are available and how the script is executed.
+
+The ledger language version and Plutus Core version can be chosen by the script developer, while the protocol parameters are only changed by by on-chain governance.
```
zliu41:
```suggestion
The ledger language version and Plutus Core version can be chosen by the script developer, while the protocol parameters are only changed by on-chain governance.
```

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -50,6 +56,26 @@ See [CIP-85](https://cips.cardano.org/cip/CIP-0085) for more details.
 Note that adding new builtin functions does not require a new Plutus Core version.
 Once a new builtin function is added, one can simply start using the new builtin function with an existing Plutus Core version.
 
+## Protocol Parameters
+Protocol parameters influence the overall behaviour of the Cardano blockchain, such as validation of blocks which includes script execution.
+Each produced block includes the protocol parameters under which it was created, so that scripts in that block will always be executed in the same way, regardless of future protocol parameters.
+Two of the most important parameters that influence script exection are __protocol version__ and __cost model parameters__.
```
zliu41: I don't think any of the bold emphasis is necessary. If we want to emphasize certain things, I think many other words in this document are equally, if not more, worth emphasizing.

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -50,6 +56,26 @@ See [CIP-85](https://cips.cardano.org/cip/CIP-0085) for more details.
...
+For more detail, see the [Cardano Developer Portal](https://developers.cardano.org/docs/governance/cardano-governance/)
+
+### Protocol Version
+The protocol version (PV) is a protocol parameter of the form `MAJOR.MINOR`, which is the main factor for determining available features and behaviour of the chain.
+That behaviour is implemented in the node software (which includes the Plutus Core evaluator).
+For example, starting with PV 11, the semantics of PlutusV3 + UPLC 1.1.0 is extended with [casing on constants](../delve-deeper/casing-constants).
+This feature does not affect execution of older scripts, as the UPLC interpreter takes the associated PV into account.
+
+### Cost model parameters
+Plutus script execution is subject to an execution budget for CPU and memory usage, which influences whether a script can successfully finish execution (See [Cost Model](../delve-deeper/cost-model) for more detail).
```
zliu41: Cost model parameters have nothing to do with versions. We have another page about costing.

doc/docusaurus/docs/essential-concepts/versions.md
```diff
@@ -50,6 +56,26 @@ See [CIP-85](https://cips.cardano.org/cip/CIP-0085) for more details.
 Note that adding new builtin functions does not require a new Plutus Core version.
 Once a new builtin function is added, one can simply start using the new builtin function with an existing Plutus Core version.
 
+## Protocol Parameters
+Protocol parameters influence the overall behaviour of the Cardano blockchain, such as validation of blocks which includes script execution.
+Each produced block includes the protocol parameters under which it was created, so that scripts in that block will always be executed in the same way, regardless of future protocol parameters.
+Two of the most important parameters that influence script exection are __protocol version__ and __cost model parameters__.
+
+Changes to protocol parameters are agreed upon by the decentralized governance model, and a change to protocol version requires a hard fork.
```
zliu41: - A hard fork bumps the major protocol version. Bumping the minor protocol version does not require a hard fork.
- It's not that changing the major protocol version "requires" a hard fork - there cannot be a hard fork that does not change the major protocol version. It's that changing protocol version is part of a hard fork.

zliu41 (changes requested):

doc/docusaurus/docs/essential-concepts/versions.md
basetunnel: I got this from reading the plutus core spec, which states:

> The protocol version is part of the history of the chain, as are all protocol parameters. That means that all blocks are associated with the protocol version from when they were created, so that they can be interpreted correctly.

So I understood this as: all scripts are executed with the protocol parameters of the block they are part of. How are scripts run under different protocol parameters?

doc/docusaurus/docs/essential-concepts/versions.md
basetunnel: Good point. I wrote that part because I wanted to know what factors determine UPLC semantics on-chain. I've removed this section and linked to the costing page at the start of the page, when mentioning script execution.

doc/docusaurus/docs/essential-concepts/versions.md
basetunnel: Removed "determine"

doc/docusaurus/docs/essential-concepts/versions.md
basetunnel: Fixed, thanks.

doc/docusaurus/docs/essential-concepts/versions.md
zliu41: Take the auction smart contract as an example. You have a script that validates bids. Alice submits a bid in PV10. Your script runs in PV10 to validate her bid. Then Bob submits a bid in PV11, and your script runs in PV11 to validate his bid.

basetunnel: Thanks, it makes sense to me now.

PR #7527, opened 2026-01-15
fix(cost-model): add true above-diagonal samples to valueContains benchmark

### LINKED ISSUE #7525: fix(cost-model): valueContains benchmark missing above-diagonal samples
## Problem

In PR #7476, the `valueContains` cost model was changed to use `const_above_diagonal` which expects benchmark data in two regions:
- **Above diagonal** (y > x): contained size > container size → constant cost (early exit)
- **Below diagonal** (x >= y): container size >= contained size → linear cost

However, the `valueContainsArgs` benchmark function only generates samples in the **below-diagonal region** (contained is always a subset of container), leaving the above-diagonal constant-cost path **untested**.

## Root Cause

The code has `aboveDiagonalCombos` which refers to a different diagonal:
- **Value structure diagonal**: `numPolicies` vs `tokensPerPolicy` (internal map shape)  
- **Costing diagonal**: container total size vs contained total size (what the cost model needs)

Lines 155-191 always select entries FROM the container, so `totalSize(contained) <= totalSize(container)` is always true.

## Solution

Add true above-diagonal samples where contained Value has more entries than container Value:
1. Remove misleading `aboveDiagonalCombos` (lines 143-145)
2. Add independent (container, contained) pairs where `totalSize(contained) > totalSize(container)`
3. Update comments to clarify the two diagonal concepts

## Expected Impact

- Better test coverage for the constant-cost early-exit path
- More accurate cost model parameters for the above-diagonal region
- Clearer code with properly named variables

Unisay:
## Summary

Fixes #7525

The `valueContains` cost model uses `const_above_diagonal` which requires benchmark data in two regions:
- **Above diagonal** (y > x): contained size > container size → constant cost (early exit)
- **Below diagonal** (x >= y): container size >= contained size → linear cost

However, the benchmark function only generated samples in the below-diagonal region (contained was always a subset of container).

### Benchmark Generation
- **Removed** misleading `aboveDiagonalCombos` variable (referred to Value structure diagonal, not costing diagonal)
- **Added** 20 true above-diagonal samples where contained.totalSize > container.totalSize with varied size ratios:
  - Small containers (4-16 entries) with larger contained values
  - Medium containers (25-100 entries) with larger contained values
  - Large containers (150-1024 entries) with much larger contained values
  - Very large containers (800-1024 entries) with extreme contained values
- **Updated** comments to clarify two different diagonal concepts:
  - Value structure diagonal: numPolicies vs tokensPerPolicy (internal map shape)
  - Costing diagonal: container totalSize vs contained totalSize (used by cost model)

### Cost Model Updates
- Re-ran remote costing benchmarks with expanded above-diagonal coverage (20 samples)
- Updated `builtinCostModelA/B/C.json` with refined parameters:
  - Above-diagonal constant: 233391 → 246085 (+5.4%)
  - Below-diagonal intercept: 658617 → 686640 (+4.3%)
  - Slope parameters adjusted for better fit
- Updated `benching-conway.csv` with new benchmark data (~570 total points)


plutus-core/cost-model/data/benching-conway.csv
```diff
@@ -13696,610 +13696,591 @@ ScaleValue/2/50402,1.2019950465447722e-2,1.1754087469952004e-2,1.217488858969444
...
+ValueContains/2047/1632,4.673374789208302e-5,4.650501178542745e-5,4.697197069681256e-5,8.195761325310597e-7,7.034095640975434e-7,9.462357582886929e-7
+ValueContains/2047/1836,4.9493219291175485e-5,4.9455588505146145e-5,4.952493697785179e-5,1.1212742406599245e-7,8.537164258094893e-8,1.5512835180690427e-7
+ValueContains/2047/2040,5.3266545259874884e-5,5.308681884957415e-5,5.354249992828979e-5,7.248979908504095e-7,4.954387722128181e-7,9.363178973784903e-7
+ValueContains/2047/2047,5.2236242145021454e-5,5.207812274691822e-5,5.232629881721979e-5,3.8864128477745657e-7,1.9087089282066684e-7,7.420374628401684e-7
+ValueContains/6/19,1.0699887015083838e-6,1.0687662261607064e-6,1.0708544970959695e-6,3.4522065657181432e-9,2.604000047664845e-9,4.511497764208202e-9
+ValueContains/14/39,1.0717634719239387e-6,1.0709872578402148e-6,1.0724089953427136e-6,2.3616473734478717e-9,1.9008164318281284e-9,2.8433237981270886e-9
+ValueContains/19/69,1.0726533995053781e-6,1.0712197114128875e-6,1.073944554452795e-6,4.460072405313118e-9,3.9215494067991e-9,5.1790811627754355e-9
```
Unisay: Above diagonal.

plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
```diff
@@ -123,76 +123,97 @@ valueContainsArgs gen = runStateGen_ gen \g -> do
...
+       (used by const_above_diagonal cost model)
 
      Size calculation: containerSize = tokensPerPolicy + (numPolicies - 1)
      Max size: 1024 + 1023 = 2047 (symmetric with LookupCoin range)
 
-     Grid: 10×10 power-of-2, below-diagonal (55) + 2 above-diagonal (57 total)
+     Grid: 10×10 power-of-2 below-diagonal (55 combos)
      Samples per container: 10 evenly distributed contained sizes
-     Total points: ~570
+     Above-diagonal: 3 independent (container, contained) pairs
```
kwxm: Maybe 10 or 20 with some of them containing larger values would be better, just to make sure that we get a decent average: 3 points seems a bit small.

Unisay: Done.

kwxm:
Thanks for this.  I've incorporated the updated benchmarking code in #7522, so I'll close this.

PR #7528, opened 2026-01-20
Let uplc support .hex files in addition to flat files

zeme-wana:
Currently, parsing UPLC .flat file using uplc utility expects given file to be encoded as raw binary containing flat encoded UPLC term. Yet, it is more common to have some script source code in textual hex string format as that is how it is often used in various offchain frameworks and cardano-cli.

This PR adds support for .hex files in `uplc`: when the input file ends in `.hex`, then hex encoding is expected and the file is parsed as text. All other extensions fallback to flat.

### DIFF
--- a/plutus-core/executables/src/PlutusCore/Executable/AstIO.hs
+++ b/plutus-core/executables/src/PlutusCore/Executable/AstIO.hs
@@ -38,9 +38,6 @@ import Control.Lens (traverseOf)
 import Data.ByteString qualified as BS
 import Data.ByteString.Base16 qualified as Base16
 import Data.ByteString.Lazy qualified as BSL
-import Data.Text qualified as T
-import Data.Text.Encoding qualified as TE
-import Data.Text.IO qualified as T
 import PlutusCore.Flat (Flat, flat, unflat)
 import System.FilePath (takeExtension)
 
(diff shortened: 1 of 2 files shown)


kwxm:
Thanks, I'll take a look.  This is probably the same as https://github.com/IntersectMBO/plutus-private/issues/2000.

plutus-core/executables/src/PlutusCore/Executable/AstIO.hs
```diff
@@ -175,9 +178,18 @@ serialiseUplcProgramFlat =
...
 getBinaryInput StdInput = BSL.getContents
-getBinaryInput (FileInput file) = BSL.readFile file
+getBinaryInput (FileInput file)
+  | takeExtension file == ".hex" = do
+      hexBytes <- BS.readFile file
+      case Base16.decode (BS.filter (not . isHexWhitespace) hexBytes) of
+        Left err -> error $ "Hex decode failure for " ++ show file ++ ": " ++ err
+        Right bs -> pure $ BSL.fromStrict bs
+  | otherwise = BSL.readFile file
+  where
```
kwxm, line 191: I think this could also be attached to the `case` above, but it's not a big deal.

PR #7531, opened 2026-01-20
Add `QuadraticInX` to metatheory

zliu41:


ana-pantilie:
Don't we need to add `WithInteractionInXAndY` and `LinearInU` as well? These are new cost models added in https://github.com/IntersectMBO/plutus/pull/7435.

zliu41:
They are already added

ana-pantilie:
> They are already added

~Can you point out where exactly? I don't see them in the definition of `convertRawModel`. I also didn't add them to the metatheory in #7435 (but maybe someone else did in a separate PR).~ Nvm, I can't read.

ana-pantilie (approved):

PR #7534, opened 2026-01-21
update unValueData to reject non-canonical input

zliu41:
based on https://github.com/cardano-foundation/CIPs/pull/1134

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Value.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Value.hs
@@ -44,8 +44,6 @@ import Data.Bifunctor
 import Data.ByteString (ByteString)
 import Data.ByteString qualified as B
 import Data.ByteString.Base64 qualified as Base64
-import Data.DList (DList)
-import Data.DList qualified as DList
 import Data.Foldable (find)
 import Data.Hashable (Hashable (..))
 import Data.IntMap.Strict (IntMap)
@@ -449,13 +447,14 @@ unValueData :: Data -> BuiltinResult Value
 unValueData = \case
   Map cs -> do
     (outerList, sizes, total, neg) <- goCurrencies cs
-    pure $ Value (Map.fromDistinctAscList (DList.toList outerList)) sizes total neg
+    pure $ Value (Map.fromDistinctDescList outerList) sizes total neg
   _ -> fail "unValueData: non-Map constructor"
   where
     unB :: Data -> BuiltinResult K
     unB = \case
       B b -> maybe (fail $ "unValueData: invalid key: " <> show (B.unpack b)) pure (k b)
       _ -> fail "unValueData: non-B constructor"
+    {-# INLINEABLE unB #-}
 
     unQ :: Data -> BuiltinResult Quantity
     unQ = \case
@@ -464,6 +463,7 @@ unValueData = \case
             fail "unValueData: invalid quantity"
         | otherwise -> pure (UnsafeQuantity i)
       _ -> fail "unValueData: non-I constructor"
+    {-# INLINEABLE unQ #-}
 
     -- Returns the inner map and the number of negative quantities in it.
     unTokens :: Data -> BuiltinResult (Map K Quantity, Int)
@@ -471,19 +471,20 @@ unValueData = \case
       Map ts -> do
         when (null ts) $ fail "unValueData: empty inner map"
         (innerList, neg) <- goTokens ts
-        pure (Map.fromDistinctAscList (DList.toList innerList), neg)
+        pure (Map.fromDistinctDescList innerList, neg)
       _ -> fail "unValueData: non-Map constructor"
+    {-# INLINEABLE unTokens #-}
 
     -- Returns outer map's list, plus stats (sizes, total, neg).
     goCurrencies
       :: [(Data, Data)]
-      -> BuiltinResult (DList (K, Map K Quantity), IntMap Int, Int, Int)
+      -> BuiltinResult ([(K, Map K Quantity)], IntMap Int, Int, Int)
     goCurrencies = go Nothing mempty mempty 0 0
       where
         go !prev !acc !sizes !total !neg = \case
           [] -> pure (acc, sizes, total, neg)
           (cData, tsData) : rest -> do
-            c <- unB cData
+            !c <- unB cData
             -- Verify that currencies are strictly ascending
             whenJust
               prev
@@ -492,21 +493,21 @@ unValueData = \case
                     (p < c)
                     (fail "unValueData: currency symbols not strictly ascending")
               )
-            (inner, innerNeg) <- unTokens tsData
+            (!inner, !innerNeg) <- unTokens tsData
             let sizes' = IntMap.alter (maybe (Just 1) (Just . (+ 1))) (Map.size inner) sizes
                 total' = total + Map.size inner
                 neg' = neg + innerNeg
-                acc' = DList.snoc acc (c, inner)
+                acc' = (c, inner) : acc
             go (Just c) acc' sizes' total' neg' rest
 
     -- Returns inner map's list, plus the number of negative quantities in the inner map.
-    goTokens :: [(Data, Data)] -> BuiltinResult (DList (K, Quantity), Int)
+    goTokens :: [(Data, Data)] -> BuiltinResult ([(K, Quantity)], Int)
     goTokens = go Nothing mempty 0
       where
         go !prev !acc !neg = \case
           [] -> pure (acc, neg)
           (tData, qData) : rest -> do
-            t <- unB tData
+            !t <- unB tData
             -- Verify that token names within an inner map are strictly ascending
             whenJust
               prev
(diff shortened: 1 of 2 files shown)


kwxm:
@zliu41 You'll need to remove` f` and the `where` clause [here](https://github.com/IntersectMBO/plutus/blob/5c39a7adce358ab30754471c0310ff9491c71e32/plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs#L283) (that's supposed to take you to line 283, but doesn't work for me).  We'll need to re-run the costing benchmark too: a quick experiment suggests that it's still quadratic, but only slightly.

ana-pantilie:
Should we release 1.57.0.0 without this modification or should we wait until this is merged? @zliu41

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -440,17 +441,15 @@ valueData = Map . fmap (bimap (B . unK) tokensData) . Map.toList . unpack
     tokensData = Map . fmap (bimap (B . unK) (I . unQuantity)) . Map.toList
 {-# INLINEABLE valueData #-}
 
-{-| \(O(n \log n)\). Decodes `Data` into `Value`, in the same way as non-builtin @Value@.
+{-| \(O(n)\). Decodes `Data` into `Value`.
 This is the denotation of @UnValueData@ in Plutus V1, V2 and V3. -}
 unValueData :: Data -> BuiltinResult Value
 unValueData =
-  fmap pack . \case
+  fmap pack' . \case
```
kwxm: Can't you remove `fmap pack' .` and put `pure . pack' . Map.fromDistinctAscList $ cs'` or something similar later?

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -460,15 +459,33 @@ unValueData =
...
-      Map ts -> fmap (Map.fromListWith unsafeAddQuantity) (traverse (bitraverse unB unQ) ts)
+      Map ts -> do
+        when (null ts) $ fail "unValueData: empty inner map"
+        ts' <- traverse (bitraverse unB unQ) ts
+        ensureDistinctAsc "unValueData: token names not strictly ascending" (fst <$> ts')
+        pure $ Map.fromDistinctAscList ts'
       _ -> fail "unValueData: non-Map constructor"
 {-# INLINEABLE unValueData #-}
 
+ensureDistinctAsc :: String -> [K] -> BuiltinResult ()
```
kwxm: You could use `Data.List.Ordered.isSortedBy` here (or not).

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -440,17 +441,15 @@ valueData = Map . fmap (bimap (B . unK) tokensData) . Map.toList . unpack
...
-  fmap pack . \case
+  fmap pack' . \case
     Map cs -> do
-      -- Use unchecked addition during construction
-      outerMap <-
-        Map.fromListWith (Map.unionWith unsafeAddQuantity) <$> traverse (bitraverse unB unTokens) cs
-      -- Validate all quantities are within bounds
-      validateQuantities outerMap
+      cs' <- traverse (bitraverse unB unTokens) cs
+      ensureDistinctAsc "unValueData: currency symbols not strictly ascending" (fst <$> cs')
```
kwxm: Can we somehow merge this into the traversal?  It looks as if we have to traverse the list three times here: once to unwrap the entries, once to check that it's ascending, and then once again when we convert it into a map.  Maybe laziness means that that doesn't actually happen though (or GHC's able to optimise some of it away)?

plutus-conformance/test-cases/uplc/evaluation/builtin/semantics/unValueData/data-duplicate-currencies-cancel/data-duplicate-currencies-cancel.uplc.budget.expected
kwxm, line 1: It's convenient that all of these tests are still testing something ...

kwxm (approved):
This all looks good.  It'll need to be re-costed though, and I think that we should look carefully at the benchmark results: there's a possiblity that it'll round up a really small (small enough to ignore) quadratic coefficient to 1 and charge too much.  Something similar happened in #7491 (not yet merged).

zliu41:
> Should we release 1.57.0.0 without this modification or should we wait until this is merged? 

@ana-pantilie we should release 1.57 without this

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -440,18 +443,14 @@ valueData = Map . fmap (bimap (B . unK) tokensData) . Map.toList . unpack
     tokensData = Map . fmap (bimap (B . unK) (I . unQuantity)) . Map.toList
 {-# INLINEABLE valueData #-}
 
-{-| \(O(n \log n)\). Decodes `Data` into `Value`, in the same way as non-builtin @Value@.
+{-| \(O(n)\). Decodes `Data` into `Value`.
```
kwxm, line 444: I'm not convinced that it is O(n), but I don't know what it should be.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -440,18 +443,14 @@ valueData = Map . fmap (bimap (B . unK) tokensData) . Map.toList . unpack
...
-      -- Use unchecked addition during construction
-      outerMap <-
-        Map.fromListWith (Map.unionWith unsafeAddQuantity) <$> traverse (bitraverse unB unTokens) cs
-      -- Validate all quantities are within bounds
-      validateQuantities outerMap
-    _ -> fail "unValueData: non-Map constructor"
+unValueData = \case
+  Map cs -> do
+    (outerList, sizes, total, neg) <- goCurrencies cs
+    pure $ Value (Map.fromDistinctAscList (DList.toList outerList)) sizes total neg
```
kwxm: We could do the `fromDistinctAscList` inside `goCurrencies`, and similarly for `goTokens`.  I don't know if that's any better though.

kwxm: I wonder if using `DList` is any better than using a standard (strict) list as an accumulator and reversing it at the end.  The [documentation](https://hackage-content.haskell.org/package/dlist-1.0/docs/Data-DList.html#g:2) for `DList.toList` says 

> Evaluating toList xs may “collapse” the chain of function composition underlying many [DList](https://hackage-content.haskell.org/package/dlist-1.0/docs/Data-DList.html#t:DList) functions ([append](https://hackage-content.haskell.org/package/dlist-1.0/docs/Data-DList.html#v:append) in particular) used to construct xs. This may affect any efficiency you achieved due to laziness in the construction.

I see that the accumulators are strict anyway, so I have no idea how efficient it'll be.  It's probably not very important in comparison with the gains from fusing the traversals, so this should be fine.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -460,13 +459,66 @@ unValueData =
...
     unTokens = \case
-      Map ts -> fmap (Map.fromListWith unsafeAddQuantity) (traverse (bitraverse unB unQ) ts)
+      Map ts -> do
+        when (null ts) $ fail "unValueData: empty inner map"
+        (innerList, neg) <- goTokens ts
+        pure (Map.fromDistinctAscList (DList.toList innerList), neg)
       _ -> fail "unValueData: non-Map constructor"
+
+    -- Returns outer map's list, plus stats (sizes, total, neg).
+    goCurrencies
```
kwxm, line 479: I briefly wondered if we could use something like `foldlM` to do this, but decided that that was a bad idea: it's much clearer (and maybe more efficient) to make everything explicit.

plutus-core/plutus-core/src/PlutusCore/Value.hs
```diff
@@ -460,13 +459,66 @@ unValueData =
 
     unQ :: Data -> BuiltinResult Quantity
```
kwxm, line 459: This only gets used once, so it could be inlined.  It's maybe not worth it though.

plutus-core/plutus-core/test/Value/Spec.hs
```diff
@@ -494,6 +503,9 @@ tests =
     , testProperty
         "oppositeScaleIsInverse"
         prop_oppositeScaleIsInverse
+    , testProperty
```
kwxm, line 508: We could also check the round trip in the other direction; it looks as if `genValueDataWithMixedQuantities` would provide suitable inputs, including failing ones.

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: Actually we don't need to reverse it because we can simply use `fromDistinctDescList`

plutus-core/plutus-core/src/PlutusCore/Value.hs
zliu41: I'm pretty sure GHC inlines it

plutus-core/plutus-core/test/Value/Spec.hs
zliu41: I've added that check in the same test.

PR #7539, opened 2026-01-23
Add plutus-tx tests for compiling to case on built-ins

basetunnel:
As suggested by @Unisay in https://github.com/IntersectMBO/plutus/pull/7481, the Plinth guide page about casing on constants should also have some Plinth example code. I've written one for each supported built-in type and also added them as plutus-tx tests.


Unisay:
Rebased on master to get preview action working.

PR #7548, opened 2026-01-27
fix(cek): prevent variable capture in dischargeCekValue (#7526)

### LINKED ISSUE #7526: `dischargeCekValue` doesn't handle open terms correctly
`dischargeCekValue` has the following code:

```haskell
                  -- var is free, leave it alone
                  (Var () named)
```

which means that

`(\0 \0 var 2) (delay (var 1))`

evaluates to

`[1 :-> delay (var 1)] (\0 var 2)`

which then discharges to

`\0 delay (var 1)`

which is variable capture.

Unisay:
## Context

- **What**: Fix variable capture bug in CEK machine's `dischargeCekValue` function
- **Why**: Free variables in environment values were incorrectly bound when discharged under lambda binders, causing semantic errors in term reconstruction
- **Issue**: Fixes #7526

## Approach

The bug occurred in `dischargeCekValue` when reconstructing terms from CEK runtime values. When an environment value containing free variables was discharged under lambda binders, those free variables would incorrectly capture to the lambda binder instead of remaining free.

**Root Cause:** `goValue` was called directly on environment values without accounting for the `shift` parameter tracking lambda depth. This meant free variables in environment values weren't shifted to preserve their "freeness" relative to the enclosing lambda context.

**Solution:** Instead of a two-pass approach (discharge + separate `shiftTermBy` post-pass), we fuse the shifting directly into the discharge traversal by threading a `global` shift parameter through `goValue` and `goValEnv`:

- `goValue` gains a `Word64` parameter (`global`) that accumulates shifts from outer discharge contexts
- When a variable is found in the env, `goValue (global + shift)` passes the accumulated shift into the recursive discharge
- When a variable is truly free (not in the env), `shiftNamedDeBruijn (global + shift)` shifts it directly
- A new `shiftNamedDeBruijn` utility is added to `PlutusCore.DeBruijn`
- The standalone `shiftTermBy` function is removed entirely

This single-pass approach avoids a redundant traversal and handles truly free variables (not found in the environment) consistently.

## Tests

Comprehensive discharge tests in `Evaluation.FreeVars`:
- Variable capture scenarios from the original bug (empty env with free vars under lambdas)
- Nested closures with accumulated shifts across multiple env lookups
- Truly free variables past non-empty environments (with and without lambdas, nested, and mixed bound/free vars)
- Boundary condition: `shift == idx` verifying bound variable detection
- Constructor (`VConstr`) arguments containing free variables under lambdas

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/DeBruijn/Internal.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/DeBruijn/Internal.hs
@@ -108,7 +108,9 @@ newtype Index = Index Word64
 deBruijnInitIndex :: Index
 deBruijnInitIndex = 0
 
--- | Shift a 'NamedDeBruijn' index by a given amount.
+{-| Shift a 'NamedDeBruijn' index by a given amount.
+The addition is unchecked and will silently wrap on 'Word64' overflow,
+which is safe in practice since terms with @2^64@ nested binders cannot be constructed. -}
 shiftNamedDeBruijn :: Word64 -> NamedDeBruijn -> NamedDeBruijn
 shiftNamedDeBruijn i (NamedDeBruijn t (Index n)) = NamedDeBruijn t (Index (n + i))
 
--- a/plutus-core/plutus-core/src/PlutusCore/DeBruijn/Internal.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/DeBruijn/Internal.hs
@@ -42,6 +42,7 @@ module PlutusCore.DeBruijn.Internal
   , deBruijnInitIndex
   , toFake
   , fromFake
+  , shiftNamedDeBruijn
   ) where
 
 import PlutusCore.Name.Unique
@@ -106,6 +107,10 @@ newtype Index = Index Word64
 -- | The LamAbs index (for debruijn indices) and the starting level of DeBruijn monad
 deBruijnInitIndex :: Index
 deBruijnInitIndex = 0
+
+-- | Shift a 'NamedDeBruijn' index by a given amount.
+shiftNamedDeBruijn :: Word64 -> NamedDeBruijn -> NamedDeBruijn
+shiftNamedDeBruijn i (NamedDeBruijn t (Index n)) = NamedDeBruijn t (Index (n + i))
 
 -- The bangs gave us a speedup of 6%.
 
--- a/plutus-core/plutus-core/src/PlutusCore/DeBruijn/Internal.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/DeBruijn/Internal.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
@@ -112,7 +113,7 @@ deBruijnInitIndex = 0
 The addition is unchecked and will silently wrap on 'Word64' overflow,
 which is safe in practice since terms with @2^64@ nested binders cannot be constructed. -}
 shiftNamedDeBruijn :: Word64 -> NamedDeBruijn -> NamedDeBruijn
-shiftNamedDeBruijn i (NamedDeBruijn t (Index n)) = NamedDeBruijn t (Index (n + i))
+shiftNamedDeBruijn !i (NamedDeBruijn t (Index n)) = NamedDeBruijn t (Index (n + i))
 
 -- The bangs gave us a speedup of 6%.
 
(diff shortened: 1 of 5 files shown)


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -690,8 +690,8 @@ dischargeCekValue value0 = DischargeNonConstant $ goValue value0
                 maybe
                   -- var is free, leave it alone
                   (Var () named)
-                  -- var is in the env, discharge its value
-                  goValue
+                  -- var is in the env, discharge its value and shift free vars
+                  (shiftTermBy shift . goValue)
```
basetunnel: Is there a way to measure how this call impacts overall CEK performance?

basetunnel: Since it does a full traversal over each value (at least when shift > 0), I'd be interested if this logic could be fused in goValue instead.

basetunnel: Here's an attempt which passes the new tests: https://github.com/IntersectMBO/plutus/compare/yura/issue-7526-discharge-cek-value-open-terms...basetunnel/issue-7526-discharge-cek-value-open-terms

It maintains a "global" shift, so that any free variable encountered in `goValEnv` can directly be shifted. I don't know how to test it performance-wise, but it avoids doing an extra pass.

basetunnel:
Is there a reason to fix this immediately? We have the scope check in place to reject open terms before CEK execution, and I'm not aware of any code path that evaluate open terms. As this could slow down the CEK machine, it only seems necessary if we ever remove the scope check.

effectfully:
@basetunnel 

> We have the scope check in place to reject open terms before CEK execution

Only in prod where it doesn't matter, because discharging doesn't happen in prod (it would be a major bug if it did).

I.e. this doesn't introduce any slowdown on the critical path and a slowdown in tests and whatever is completely fine. Perhaps even what looks like a quadratic-in-case-of-free-variables version implemented in this PR.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
Unisay: Thanks! I reworked the PR to use your global shift approach.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -653,55 +653,61 @@ dischargeResultToTerm (DischargeConstant val) = Constant () val
...
-      VDelay body env -> Delay () $ goValEnv env 0 body
+      VDelay body env -> Delay () $ goValEnv env global 0 body
       VLamAbs (NamedDeBruijn n _ix) body env ->
         -- The index on the binder is meaningless, we put @0@ by convention, see 'Binder'.
-        LamAbs () (NamedDeBruijn n deBruijnInitIndex) $ goValEnv env 1 body
+        LamAbs () (NamedDeBruijn n deBruijnInitIndex) $ goValEnv env global 1 body
       -- We only return a discharged builtin application when (a) it's being returned by the
       -- machine, or (b) it's needed for an error message.
       -- @term@ is fully discharged, so we can return it directly without any further discharging.
       VBuiltin _ term _ -> term
```
basetunnel: This term should also be shifted, see the problematic example at https://github.com/IntersectMBO/plutus/issues/7526#issuecomment-3823154953. Perhaps it can be turned in a test too?

plutus-core/plutus-core/src/PlutusCore/DeBruijn/Internal.hs
```diff
@@ -107,6 +108,10 @@ newtype Index = Index Word64
 deBruijnInitIndex :: Index
 deBruijnInitIndex = 0
 
+-- | Shift a 'NamedDeBruijn' index by a given amount.
+shiftNamedDeBruijn :: Word64 -> NamedDeBruijn -> NamedDeBruijn
+shiftNamedDeBruijn i (NamedDeBruijn t (Index n)) = NamedDeBruijn t (Index (n + i))
```
basetunnel: Do you think we need a bang pattern on `i`? It seems like this is usually done in the CEK code dealing with indices.

basetunnel:
> @basetunnel
> 
> > We have the scope check in place to reject open terms before CEK execution
> 
> Only in prod where it doesn't matter, because discharging doesn't happen in prod (it would be a major bug if it did).

Thanks for pointing that out. This relies on laziness, right? `dischargeCekValue` is called in two places in the CEK machine, but the resulting term under `DischargeNonConstant` is never forced. What is the reason we don't just return a `CekValue` from the machine?

effectfully:
> This relies on laziness, right?

Yes.

> What is the reason we don't just return a `CekValue` from the machine?

Just to have a tidy uniform API between various evaluators? They take a term and return a term. Maybe it's not worth it, but then you have the same issue with builtins within `CekValue`, so you'd still end up with a similar situation.

PR #7553, opened 2026-01-30
Replace use of tail with drop 1 in PlutusTx/Enum/TH.hs

zeme-wana:
Fixes this error:
```
src/PlutusTx/Enum/TH.hs:53:59: error: [GHC-63394] [-Wx-partial, -Werror=x-partial]
    In the use of ‘tail’
    (imported from Prelude, but defined in GHC.List):
    "This is a partial function, it throws an error on empty lists. Replace it with drop 1, or use pattern matching or Data.List.uncons instead. Consider refactoring to use Data.List.NonEmpty."
   |
53 |       [ funD 'succ (fmap (deriveSuccPred Succ) (zip cons (tail (Just <$> cons) ++ [Nothing])))
   |     
```

### DIFF
--- a/plutus-tx/src/PlutusTx/Enum/TH.hs
+++ b/plutus-tx/src/PlutusTx/Enum/TH.hs
@@ -50,7 +50,7 @@ deriveEnum name = do
     <$> instanceD
       (pure [])
       (pure instanceType)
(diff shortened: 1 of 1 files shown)


kwxm:
Is there any danger that this will actually get an empty list, or is the input guaranteed to be non-empty -in this situation?  `drop 1 []` is `[]`, which might not be what we want.  Matching on the list and failing for the empty list (which may be impossible) might be safer.

PR #7558, opened 2026-02-02
Enable plinth-tx-plugin to be built with ghc912

zeme-wana:


plutus-tx-plugin/plutus-tx-plugin.cabal
```diff
@@ -43,8 +43,9 @@ common lang
 
 common ghc-version-support
   -- See the section on GHC versions in CONTRIBUTING
-  if (impl(ghc <9.6) || impl(ghc >=9.7))
-    buildable: False
+  -- if (impl(ghc <9.6) || impl(ghc >=9.7))
+  --   buildable: False
+  buildable: True
```
zeme-wana: Temporary: proper `if` guard to be added later

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -1017,7 +1017,7 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
     -- We can safely commit to this match as soon as we've seen fromString -
     -- we won't accept any applications of fromString that aren't creating literals of
     -- the types we support.
-    (strip -> GHC.Var (GHC.idDetails -> GHC.ClassOpId cls))
+    (strip -> GHC.Var (GHC.idDetails -> GHC.ClassOpId cls True))
```
zeme-wana: [Haddock for ClassOpId](https://hackage-content.haskell.org/package/ghc-9.14.1/docs/src/GHC.Types.Id.Info.html#ClassOpId):
```
True <=> given a non-bottom dictionary, the class op will
definitely return a non-bottom result
and Note [exprOkForSpeculation and type classes]
in GHC.Core.Utils
```

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -1017,11 +1019,11 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
     -- We can safely commit to this match as soon as we've seen fromString -
     -- we won't accept any applications of fromString that aren't creating literals of
     -- the types we support.
-    (strip -> GHC.Var (GHC.idDetails -> GHC.ClassOpId cls))
+    (strip -> GHC.Var cls)
```
zliu41: What's the reason for this change?

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -883,6 +884,7 @@ compileHaskellList = buildList . strip
         consume li >>= traverse compileExpr
     buildList _ = err
 
+{- FOURMOLU_DISABLE -} -- Fourmolu can't parse non-top-level #if directives
```
zliu41: Disabling fourmolu for such a large function isn't ideal. You can create a `PlutusTx.Compiler.Compat` module and put all the CPP there.

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
zeme-wana: @SeungheonOh

SeungheonOh: GHC912 changed how overloaded strings are compiled, they use `fromString @_... ""` which I wasn't able to match if I had it match `ClassOpId`.

zliu41: Then this variable should not be called "cls" anymore, since it is no longer a `Class`. It is now an `IdDetails`.

zliu41 (changes requested):
Looks good, just two minor comments

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -1017,11 +1019,11 @@ compileExpr e = traceCompilation 2 ("Compiling expr:" GHC.<+> GHC.ppr e) $ do
     -- We can safely commit to this match as soon as we've seen fromString -
     -- we won't accept any applications of fromString that aren't creating literals of
     -- the types we support.
-    (strip -> GHC.Var (GHC.idDetails -> GHC.ClassOpId cls))
+    (strip -> GHC.Var cls)
       `GHC.App` GHC.Type ty
       `GHC.App` _dict
       `GHC.App` (strip -> content)
-        | GHC.getName cls == GHC.isStringClassName -> do
+        | GHC.getName cls == GHC.isStringClassName || GHC.getName cls == GHC.fromStringName -> do
```
zliu41: `cls` can no longer possibly match `GHC.isStringClassName` since it is no longer a `Class`.

zliu41: You should be able to simply bind this variable on the `GHC.Var` itself, i.e., `(strip -> var)`

PR #7560, opened 2026-02-03
Fix uconv to allow an arbitrary number of variables

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/2055


plutus-metatheory/src/FFI/Untyped.hs
```diff
@@ -43,22 +43,22 @@ conv (Force _ t) = UForce (conv t)
 conv (Constr _ i es) = UConstr (toInteger i) (toList (fmap conv es))
 conv (Case _ arg cs) = UCase (conv arg) (toList (fmap conv cs))
 
-tmnames :: String
-tmnames = ['a' .. 'z']
+tmnames :: [String]
```
zliu41: `tnames :: Int -> String` would be a lot more efficient than `!!`. In general, CAFs are bad.

zliu41 (approved):

PR #7562, opened 2026-02-04
feat: native builtin data vs th builtin data benchmark

colll78:
Added a benchmark using very crude and barely optimized native builtin data functions to compare against the existing Vesting contract implemented with TH AsData stuff.

Added script context builder for constructing realistic script contexts which obey ledger invariants.

### DIFF
--- a/plutus-benchmark/cardano-loans/src/CardanoLoans/Test.hs
+++ b/plutus-benchmark/cardano-loans/src/CardanoLoans/Test.hs
@@ -9,14 +9,26 @@
 module CardanoLoans.Test where
 
 import PlutusTx
-import PlutusTx.Prelude
+import PlutusTx.Prelude hiding ((<>))
 
 import CardanoLoans.Validator (LoanDatum (..), LoanRedeemer (..), loanValidatorCode)
+import PlutusLedgerApi.Test.ScriptContextBuilder.Builder
+  ( buildScriptContext
+  , withAddress
+  , withInlineDatum
+  , withOutRef
+  , withOutput
+  , withSigner
+  , withSpendingScript
+  , withTxOutAddress
+  , withTxOutValue
+  , withValidRange
+  , withValue
+  )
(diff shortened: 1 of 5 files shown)


plutus-benchmark/linear-vesting/src/LinearVesting/Validator.hs
```diff
@@ -139,20 +142,37 @@ validateVestingFullUnlock ctx =
...
+      (not (txSignedBy txInfo beneficiaryKey))
+      (\_ -> traceError "Missing beneficiary signature")
+      ( \_ ->
+          BI.ifThenElse
+            (vestingPeriodEnd vestingDatum >= currentTimeApproximation)
+            (\_ -> traceError "Unlock not permitted until vestingPeriodEnd time")
+            (\_ -> True)
+            BI.unitval
+      )
+      BI.unitval
```
Unisay, line 155: I did an experiment and results demonstate that multiway-if compiled by the plugin is actually better than hand-rolled lazy ifthenelse pattern you employ here:
https://github.com/IntersectMBO/plutus/pull/7578

colll78: Good point, I guess similar could be achieved with `ifThenElseLazy predicate ~a ~b = BI.ifThenElse predicate a b` right? And if the branches don't contain an error and the terms are small (ie. for `&&` which has only `True` and `False` as branches, then we would want to use `BI.ifThenElse` because multiway if isn't smart enough to know whether it can be safely optimized to strict (and whether it should based on the terms in the branches). 

I'll update the implementation to use multiway if, I'm sure it could be further optimized, but I just did a rough translated to BuiltinData and didn't put much effort into optimization and it is already optimized to around 35% of the budget of the TH Data version.

Unisay:
Optimized linear vesting validator https://github.com/IntersectMBO/plutus/pull/7658

PR #7573, opened 2026-02-06
Fix flaky uplc-evaluator-integration-tests and re-enable

Unisay:
## Summary

- **Fix measurement sharing bug**: GHC's full laziness transformation shared the `evaluateWithBudget term` result between the initial validation call and the timing loop. After validation forced the result to WHNF, subsequent calls in the loop measured a no-op (~0ns). Fixed with two layers of defense:
  - `{-# NOINLINE evaluateWithBudget #-}` prevents inlining the function body
  - `{-# NOINLINE measureSingleExecution #-}` (new) prevents GHC from inlining the timing function into the loop, keeping `evaluateWithBudget t` dependent on the lambda-bound `t` — full laziness **cannot** float expressions that depend on lambda-bound variables, providing a semantic guarantee (not just a heuristic) that CEK evaluation runs fresh each iteration
- **Fix file-write race condition**: The service could read partially-written input files. Fixed by writing to a temp file then atomically renaming.
- **Restore strict timing assertions**: With the robust NOINLINE fix, `cpu_time_ns > 0` assertions serve as a canary — if sharing prevention ever breaks, tests fail loudly rather than silently passing with 0ns measurements.
- **Re-enable test suite**: Removed `buildable: False` from `uplc-evaluator-integration-tests`.

## Test plan

- [x] `cabal build plutus-benchmark:uplc-evaluator` succeeds
- [x] `cabal build plutus-benchmark:uplc-evaluator-integration-tests` succeeds
- [x] All 28 integration tests pass
- [x] Tests pass consistently across 5 consecutive runs (with `> 0` assertions)

### DIFF
--- a/plutus-benchmark/uplc-evaluator/test/Spec.hs
+++ b/plutus-benchmark/uplc-evaluator/test/Spec.hs
@@ -41,908 +41,908 @@ main = withUtf8 do
     ( testGroup
         "uplc-evaluator integration tests"
         [ testGroup
-      "Infrastructure"
-      [ testCase "Basic scaffolding test" do
-          -- This is a basic test to verify the test infrastructure is set up correctly
-          let expected :: Integer = 2 + 2; actual = 4
-          actual @?= expected
-      , testCase "Service lifecycle - start and stop" do
-          -- Test that the service can start and stop cleanly
-          execPath <- findEvaluatorExecutable
-          withEvaluatorService execPath \handle -> do
-            -- Verify input directory exists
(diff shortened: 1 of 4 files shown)


plutus-benchmark/uplc-evaluator/test/Spec.hs
Unisay, line 1: the only real change is one hunk: the >= 0 → > 0 fix. The other ~880 "changed" lines were fourmolu normalizing indentation.

PR #7595, opened 2026-02-16
Fix out of date comment

mietek:
Small omission in https://github.com/IntersectMBO/plutus/pull/7486.

cc @basetunnel


basetunnel:
Thanks! The repo requires signed commits, could you set that up? Otherwise I can always commit and merge it myself.

mietek:
Ready with signed commit.

PR #7598, opened 2026-02-17
Generate and use static .flat files for Marlowe benchmark

basetunnel:
https://github.com/IntersectMBO/plutus/issues/7293

### DIFF
--- a/plutus-benchmark/plutus-benchmark.cabal
+++ b/plutus-benchmark/plutus-benchmark.cabal
@@ -721,10 +721,10 @@ benchmark marlowe-agda-cek
     , agda-internal
     , base                     >=4.9   && <5
     , criterion
+    , filepath
     , marlowe-internal
     , plutus-benchmark-common
     , plutus-ledger-api        ^>=1.60
-    , plutus-tx                ^>=1.60
 
 -------------------- bitwise -----------------------
 


plutus-benchmark/marlowe/exe/Main.hs
```diff
@@ -5,21 +5,35 @@ module Main (main) where
...
   , semanticsBenchmarks
   , tabulateResults
+  , writeFlat
   , writeFlatUPLCs
   )
 import PlutusBenchmark.Marlowe.RolePayout qualified as RolePayout
+import PlutusBenchmark.Marlowe.Scripts.Data.RolePayout qualified as DataRolePayout (rolePayoutValidator)
+import PlutusBenchmark.Marlowe.Scripts.Data.Semantics qualified as DataSemantics (marloweValidator)
+import PlutusBenchmark.Marlowe.Scripts.RolePayout qualified as RolePayout (rolePayoutValidator)
+import PlutusBenchmark.Marlowe.Scripts.Semantics qualified as Semantics (marloweValidator)
```
SeungheonOh: nitpick: 
Maybe `DataRolePayout` and `SOPRolePayout`; `DataSemantics` and `SOPSemantics` would be clearer

SeungheonOh:
On a side note, since now we generate flat files for malrowe, we can also put this for `validation` benchmark. I think this would be helpful since test scripts in `validation` doesn't really represent the SOP terms much as many scripts there were created before introduction of Case/Constr

plutus-benchmark/marlowe/exe/Main.hs
basetunnel: Done

PR #7599, opened 2026-02-17
[certifier] Add a five script benchmark

zliu41:
Each script runs for 12 simplifier iterations, plus 4 more simplifier iterations interleaved with CSE. The benchmark counts the total time taken by the certifier.

Currently the certifier doesn't work so well - it fails to certify the first four scripts, and does not terminate on Marlowe. More investigation will be needed.


effectfully:
image

ana-pantilie:
https://github.com/IntersectMBO/plutus-private/issues/2054 is probably one reason the certifier is currently failing, which @effectfully implemented in the compiler without changing the certifier as well 😛 .

ana-pantilie (approved):
On a more serious note, I think this is a good idea! Should the tests be added to a `failing` bucket until we fix them? It will be very helpful to try them out while working on the set of known issues we have with the translation relations (including performance fixes).

zliu41:
> Should the tests be added to a failing bucket until we fix them?

I'd rather enable the certifiers one by one, instead of enabling the tests one by one. There are some certifiers that are clearly broken.

PR #7605, opened 2026-02-18
Additional property tests for bitwise

zliu41:
Plus some improvements and refactoring.


plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Bitwise/CIP0123.hs
```diff
@@ -436,6 +440,37 @@ rotateHomomorphism =
               ]
       evaluateTheSame lhs rhs
 
+rotateInverse :: Property
```
kwxm, line 444: These tests could maybe do with brief comments to say what they do.

kwxm (approved):
Looks good.  Thanks!

PR #7608, opened 2026-02-19
Integrate the metatheory certifier in plutus-tx-plugin 

zeme-wana:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1891

### DIFF
--- a/nix/outputs.nix
+++ b/nix/outputs.nix
@@ -122,8 +122,6 @@ let
       (packages) //
       { ghc96 = project-variants-hydra-jobs.ghc96; } //
       { ghc912 = project-variants-hydra-jobs.ghc912; } //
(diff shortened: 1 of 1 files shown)


plutus-metatheory/plutus-metatheory.cabal
```diff
@@ -48,7 +48,7 @@ common lang
 
 common ghc-version-support
   -- See the section on GHC versions in CONTRIBUTING
-  if (impl(ghc <9.6) || impl(ghc >=9.7))
+  if !((impl(ghc >=9.6) && impl(ghc <9.7)) || (impl(ghc >=9.12) && impl(ghc <9.13)))
```
ana-pantilie, line 51: I missed the `!` at first and was very confused! Should we instead apply the negation by hand, or at least comment what this means?

PR #7609, opened 2026-02-19
Wrap integer/string builtins in BuiltinResult

zliu41:
This makes things much more consistent and uniform.


zliu41:
> ???

Any concerns with this?

effectfully:
> Any concerns with this?

There were two kinds of builtins:

1. pure => no `BuiltinResult`
2. impure => `BuiltinResult`

Now there's 

1. some pure => no `BuiltinResult`
2. some pure => `BuiltinResult`
3. impure => `BuiltinResult`

How does that make anything more consistent and uniform?

Like, scroll down, `Data`-construction builtins don't have `BuiltinResult`, most BLS builtins don't have `BuiltinResult`, most logical and bitwise builtins don't have `BuiltinResult` etc.

You just basically erased info from a builtin on whether it's considered pure or not at the cost of introducing more noise. I don't see any point in doing so, but if you want to do it, do it for all builtins then and update `PlutusCore.Builtin.Meaning` to make `BuiltinResult` mandatory, 'cause otherwise you're gonna need to tell people contributing code to add that pointless `BuiltinResult` to pure builtins.

PR #7611, opened 2026-02-20
use timesInt2# in satint

zliu41:
`mulIntMayOflo#` appears to work fine at least on GHC 9.6 and 64 bit architectures (which is all we care about), but just to be future proof - GHC often does unpredictable things in a new version - it's better to switch to `timesInt2#`.


kilhacmaron-byte:
Great

PR #7614, opened 2026-02-22
Make Delay and Force by-need

zliu41:
This realizes the full potential of UPLC.


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs
```diff
@@ -1080,6 +1091,23 @@ enterComputeCek = computeCek
         (coerce varIx)
     {-# INLINE lookupVarName #-}
 
+    newThunk :: CekM uni fun s (CekThunk uni fun ann)
+    newThunk = CekM . unsafeIOToST . fmap CekThunk $ newIORef Nothing
```
effectfully, line 1095: Just use an `STRef`?

zliu41: Maybe in the final version. I'm just experimenting, and using STRef requires adding `s` in a lot of places

effectfully: > using `STRef` requires adding s in a lot of places

Right, didn't realize that.

PR #7618, opened 2026-02-24
Reuse bitwise primitives for Data

zliu41:
Reduces code duplication. The primitive implementation is also highly optimized.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Bitwise.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Bitwise.hs
@@ -3,7 +3,6 @@
 {-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE UnboxedTuples #-}
-{-# OPTIONS_GHC -ddump-simpl #-}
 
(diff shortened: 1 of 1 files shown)


plutus-core/plutus-core/src/PlutusCore/Bitwise.hs
```diff
@@ -289,6 +290,14 @@ unsafeIntegerToByteString requestedByteOrder requestedLength input = case input
           poke pEnd (byteSwap16 wStart)
           poke pStart (byteSwap16 wEnd)
           finishUp (plusPtr ptr 2) (remaining - 4)
+{-# INLINEABLE unsafeIntegerToByteString #-}
+
+integerToBytesBE :: Integer -> ByteString
```
kwxm, line 298: I'd suggest adding a comment saying that because of the `error` this must never be called on the chain with a negative input (and maybe reflect that in the name: `nonNegativeIntegerToBytesBE` or something equally horrible).  Fortunately `encodeInteger` avoids that case, but we should be careful in case we use it for anything else in future.  Maybe even get it to return "" or "\000" for negative numbers just to be sure.

kwxm:
This reminds me that we forgot to provide conformance tests for `serialiseData`.  We should make sure that we test serialisation of `I` objects with negative contents when we do that.

PR #7623, opened 2026-02-25
Add a certifer test suite using real validators

zliu41:
Added a test suite and 7 more real validators (3 Marlowe, 4 guardrail), for a total of 12.


basetunnel:
Why is the test suite so much slower than the benchmark? E.g. on my machine criterion reports:

```
benchmarking marlowe-semantics.uplc
time                 1.195 s    (1.122 s .. 1.296 s)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 1.351 s    (1.279 s .. 1.476 s)
std dev              119.1 ms   (3.121 ms .. 145.7 ms)
variance introduced by outliers: 22% (moderately inflated)
```

while the tasty reports:

```
marlowe-semantics.uplc:       OK (9.71s)
```

zliu41:
> Why is the test suite so much slower than the benchmark?

Good question. This helps: https://github.com/IntersectMBO/plutus/pull/7623/commits/91183b36a3e8330587235d23bc1d0ddec93d3f6a

PR #7624, opened 2026-02-25
Don't shell out when testing the certifier

zliu41:
There's no reason to shell out, and it's a complete mess. Cleaning it up.

Edit: fixes https://github.com/IntersectMBO/plutus-private/issues/1582


ana-pantilie:
I changed the description because we already had an issue targeting this. Thanks for looking at this code.

plutus-metatheory/src/Certifier.hs
```diff
@@ -77,27 +77,17 @@ mkCertifier
   -- ^ The name of the certificate to be produced
   -> Certifier CertifierSuccess
 mkCertifier simplTrace certName = do
-  certName' <- validCertName certName
```
ana-pantilie, line 82: Good catch that this check was redundant now that it's part of `mkCertificate`.

plutus-executables/test/certifier/Test/Certifier/Executable.hs
```diff
@@ -138,10 +52,10 @@ runAgda file = do
 
 agdaTestCert :: String -> Assertion
 agdaTestCert name = do
-  certDir <- makeUplcCert name
+  certDir <- loadAndMakeCert name
   oldDir <- getCurrentDirectory
   setCurrentDirectory certDir
-  (resCode, resText) <- runAgda ("src" </> fstToUpper name <> ".agda")
+  (resCode, resText) <- runAgda ("src" </> name <.> "agda")
   setCurrentDirectory oldDir
```
ana-pantilie, line 63: I think this stateful hopping between directories is really hacky, I was hoping we could fix this as part of this issue. Do you (or maybe @zeme-wana ) have any ideas how to handle this in a nicer way?

zliu41: Something like https://hackage.haskell.org/package/temporary-1.3/docs/System-IO-Temp.html#v:withTempDirectory

PR #7626, opened 2026-02-26
Certifier: generalized views

basetunnel:
- Added generalized views that also support predicates on any of the constructor fields.
  - I've kept the existing views, to not have to refactor all existing code that uses them. But this could eventually be done.
- Added type class `Inhabitant` for predicates that have a trivial inhabitant
- Added example to demonstrate generalized views
- Added fixity declaration for  `·`, we only have a limited supply of parentheses on this planet


ana-pantilie:
@zliu41 Jacco's views make "matching" on the structure of the term much nicer: he added infix operations and generalized the predicates. He noticed that when writing decision procedures you have to do this "matching" 3 times: once in the `with` clause, another time in the negative proof and another time in the positive proof. The `Inhabitant` typeclass removes the duplication in the negative proof, which speeds up writing decision procedures and makes them more readable IMO. I'll let @basetunnel to add any more details.

basetunnel:
> What is the goal - to make the decision procedures more concise? Or more flexible? Or faster? 

More concise and flexible. It allows you to combine more of the standard decision procedures like `_≟_` without introducing extra `with` clauses that introduce more cases. And `Inhabitant` makes negative cases shorter as Ana mentioned.

I've added some more detail to the comments in the file. Also see the example at the end of the file, which is more verbose using the other views.

zliu41 (approved):

PR #7630, opened 2026-02-27
Allow the uplc executable to take serialised scripts

zliu41:
Most scripts out there are serialised, so this should be very useful.

Since the code for serialising and deserialising scripts are in `plutus-ledger-api`, this requires moving `plutus-core:plutus-core-execlib` to `plutus-ledger-api`. And this in turn requires moving `plutus-ledger-api:plutus-ledger-api-plugin-test` to `plutus-tx-plugin`. The tests in `plutus-tx-plugin` run with GHC 9.12 (in addition to GHC 9.6), which is why there are a bunch of new golden files.

Now you can do `cabal run uplc -- optimise --certify Marlowe --input ./plutus-benchmark/certifier/data/marlowe-semantics.uplc --print-mode Classic --input-format serialised`.


plutus-ledger-api/executables/src/PlutusCore/Executable/Common.hs
```diff
@@ -206,7 +215,7 @@ printBudgetStateTally term model (Cek.CekExTally costs) = do
         -- Memory usage figures are meaningless in this case
         Unit -> printf "%15s" (show cpu) :: String
     printStepCost constr =
-      printf "%-10s %20s\n" (tail $ show constr) (budgetToString . getSpent $ Cek.BStep constr)
+      printf "%-10s %20s\n" (drop 1 $ show constr) (budgetToString . getSpent $ Cek.BStep constr)
```
kwxm, line 222: Why the change?  Do we not like `tail`?

plutus-ledger-api/executables/src/PlutusCore/Executable/Parsers.hs
```diff
@@ -72,6 +72,7 @@ formatReader :: String -> Maybe Format
 formatReader =
   \case
     "textual" -> Just Textual
+    "serialised" -> Just Serialised
```
kwxm, line 75: Can you update `formatHelp` as well please?

plutus-ledger-api/executables/src/PlutusCore/Executable/Common.hs
```diff
@@ -320,6 +332,13 @@ writeFlat outp flatMode prog = do
     StdOutput -> BSL.putStr flatProg
     NoOutput -> pure ()
 
+writeSerialised :: Output -> B.ByteString -> IO ()
+writeSerialised outp prog = do
+  case outp of
```
kwxm, line 341: This is very nearly the same as [writeFlat](https://github.com/IntersectMBO/plutus/blob/d2973d488052aeba76bc905d61d2d36300173720/plutus-core/executables/src/PlutusCore/Executable/Common.hs#L318), so it might make sense to abstract out a `writeByteString` function or something and get it to handle both `flat` and `serialised` scripts.

zliu41:
> it might make sense to move everything into plutus-executables

I don't think this works. The problem is that `plutus-metatheory` depends on `plutus-core-execlib`. `plutus-metatheory` is released on CHaP, so `plutus-core-execlib` must also be in a cabal package that is released on CHaP, and `plutus-executable` isn't one of them.

The other parts of this PR also improve things:
- The `plutus` executable shouldn't be in `plutus-core`. This PR moved it to `plutus-executables`, which is where it should be.
- Moving `plutus-ledger-api-plugin-tests` into `plutus-tx-plugin` ensures that nothing in `plutus-ledger-api.cabal` depends on `plutus-tx-plugin`.

kwxm:
> I don't think this works. The problem is that `plutus-metatheory` depends on `plutus-core-execlib`. `plutus-metatheory` is released on CHaP, so `plutus-core-execlib` must also be in a cabal package that is released on CHaP, and `plutus-executable` isn't one of them.

:( I'd forgotten about all that.  This makes `uplc` take even longer to compile though!  We have to compile `plutus-metatheory` and `plutus-ledger-api` as well.

zliu41:
> We have to compile plutus-metatheory and plutus-ledger-api as well.

It already requires compiling `plutus-metatheory`. `plutus-ledger-api` is a fairly small and quick to compile.

PR #7631, opened 2026-02-28
Remove duplicate Notes and fix references

kwxm:
There were some problems with the Notes in the code, and this should fix them all.

There were many duplicates that had been pasted from one file into another and in those cases I've removed the duplicates to leave one definitive version of the Note to avoid things getting out of sync.  

There were also a number of references to non-existent Notes:  these were generally due to mistakes in the name of the Note being referred to so I've fixed them.  

I found all of these using `scripts/check-Notes.sh` and `scripts/list-Notes.sh`.  There were a number of references to Notes in the GHC codebase as well and I've changed the format of those because the scripts couldn't tell that they weren't referring to code outside our codebase.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/MachineParameters.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/MachineParameters.hs
@@ -71,25 +71,27 @@ instance (NoThunks machinecosts, Bounded fun, Enum fun) => NoThunks (MachinePara
     allNoThunks [noThunks ctx caser, noThunks ctx varPars]
 
 {- Note [The CostingPart constraint in mkMachineVariantParameters]
-Discharging the @CostingPart uni fun ~ builtincosts@ constraint in 'mkMachineParameters' causes GHC
-to fail to inline the function at its call site regardless of the @INLINE@ pragma and an explicit
-'inline' call.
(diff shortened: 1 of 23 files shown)


plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/MachineParameters.hs
```diff
@@ -71,25 +71,27 @@ instance (NoThunks machinecosts, Bounded fun, Enum fun) => NoThunks (MachinePara
...
+https://gitlab.haskell.org/ghc/ghc/-/blob/1f02b7430b2fbab403d7ffdde9cfd006e884678e/compiler/prelude/PrelRules.hs#L1442-1449
+And if @f@ is compiled into a @Cast@ like 'mkMachineVariantParameters' with the
+discharged constraint, then inlining won't work. We don't know why it's
+implemented this way in GHC.
 
 It seems fully applying @f@ helps, i.e.
 
-    - inline mkMachineParameters unlMode <$> <...>
-    + (\x -> inline (mkMachineParameters unlMode x)) <$> <...>
+    - inline mkMachineVariantParameters unlMode <$> <...>
```
kwxm, line 93: I'm not sure whether this still applies (and the same goes for other parts of the note), but I don't want to risk changing it.

PR #7634, opened 2026-03-03
[certifier] Add notImplemented constructor to Transformation

zliu41:
This represents all passes whose certifiers haven't been implemented.


basetunnel:
This will be unnecessary after merging https://github.com/IntersectMBO/plutus/pull/7574

PR #7660, opened 2026-03-11
Add Agda certifier integration with source location tracking

zeme-wana:
- Refactor source location tracking in plutus-tx-plugin to flow through ReaderT
  CompileContext (ccCurrentLoc) instead of invasive function parameters on
  compileExpr.
  - Add --certify[=DIR] plugin option that invokes the Agda certifier for each
  compiled Plutus script, embedding package name, module name, and source location in
  the certificate metadata.
  - Simplify compileMarkedExprs (remove stateful go lastLoc loop), stripAnchors
  (single pattern match instead of recursive go), and package name resolution
  (lookupUnit / thisPackageName / stripVersionFromUnitId fallback chain).
  - Improve Certifier.hs: include certifier report in error messages, use
  createDirectoryIfMissing, remove noisy console output.
  - Encode TH source locations in structured format (encodeTHLoc) so the plugin can
  recover RealSrcSpan from the type-level string.
  - Add CI workflow (plinth-certifier-tests.yml) and test script
  (scripts/plinth-certifier-tests.py).


plutus-tx-plugin/test/CallTrace/9.6/func03.golden.eval
```diff
@@ -6,4 +6,4 @@ Trace:
 -> func (test/CallTrace/Spec.hs:79:1-79:4)
```
SeungheonOh, line 6: All test golden in test/CallTrace/* needs to be reverted. The call srcspan gets erased when we build from compilation cache. 

Also, I think this can be fixed now. made an issue https://github.com/IntersectMBO/plutus/issues/7722

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -893,7 +894,7 @@ compileHaskellList = buildList . strip
           | otherwise = err
         consume _ = err
        in
-        consume li >>= traverse (compileExpr Nothing)
+        consume li >>= traverse (compileExpr)
```
SeungheonOh, line 897: this didn't get caught by hlint? I'm surprised

plutus-tx-plugin/src/PlutusTx/Compiler/Types.hs
```diff
@@ -68,6 +68,10 @@ data CompileContext uni fun = CompileContext
   , ccDebugTraceOn :: Bool
   , ccRewriteRules :: PIR.RewriteRules uni fun
   , ccSafeToInline :: Bool
+  , ccCurrentLoc :: Maybe GHC.RealSrcSpan
+  , ccPackageName :: String
+  , ccModuleName :: String
+  , ccSourceLoc :: Maybe GHC.RealSrcSpan
```
SeungheonOh, line 74: what's for `ccSourceLoc` and what's for `ccCurrentLoc`? 

I see ccSourceLoc is only used for verifier related stuff; can we not only have ccCurrentLoc and use that for verifier stuff?

plutus-tx-plugin/src/PlutusTx/Plugin/Common.hs
```diff
@@ -918,12 +984,14 @@ stripTicks = \case
...
 
-stripCoreAnchors :: GHC.Name -> GHC.CoreExpr -> GHC.CoreExpr
-stripCoreAnchors marker = \case
-  GHC.Tick _ e -> stripCoreAnchors marker e
-  GHC.App (GHC.App (GHC.App (GHC.Var f) _locTy) _codeTy) code
-    | GHC.getName f == marker -> stripCoreAnchors marker code
-  other -> other
+{-| Strip a single @anchor@ application (and any surrounding ticks),
+returning the decoded source span if present. -}
+stripAnchors :: GHC.Name -> GHC.CoreExpr -> (Maybe GHC.RealSrcSpan, GHC.CoreExpr)
```
SeungheonOh, line 989: Maybe it's better to have separate `getAnchorSrcSpan` function instead of adding it to `stripAnchors`; or, give it some better name.

SeungheonOh (commented):
Left some comments, mainly over the compiler part. Second pair of eyes on certifier part would be nice since I'm not too familiar with that area

ana-pantilie (approved):
The certifier changes are awesome, thanks @zeme-wana! The only issue I have with this PR is, as @SeungheonOh pointed out:

> All test golden in test/CallTrace/* needs to be reverted. The call srcspan gets erased when we build from compilation cache.

plutus-tx-plugin/src/PlutusTx/Plugin/Common.hs
```diff
@@ -518,17 +535,15 @@ compileMarkedExprs expr = do
...
+        | markerName == GHC.idName fid ->
+            compileMarkedExprOrDefer (decodeSrcSpan (GHC.unpackFS fs_locStr)) codeTy inner
+    GHC.App (GHC.App fnExpr tyExpr) inner
+      | (stripAnchors anchorGhcName -> GHC.Var fid) <- fnExpr
+      , (stripAnchors anchorGhcName -> GHC.Type codeTy) <- tyExpr
+      , plinthcName == GHC.idName fid ->
+          compileMarkedExprOrDefer (getAnchorSrcSpan anchorGhcName fnExpr) codeTy inner
     e@(GHC.Var fid)
-      | markerName == GHC.idName fid ->
+      | markerName == GHC.idName fid Prelude.|| plinthcName == GHC.idName fid ->
```
SeungheonOh: I don't think we want to use plc and plinthc at the same time here. 

I assume this was done because in `GHC.App (GHC.App fnExpr tyExpr) inner`, `fnExpr` isn't anchored? I think you can make it so that they typecheceking plugin inserts anchor to `plinthc` so that it becomes 
`(anchor @loc plinthc) @a ...code to be compiled...` With this, you can use `getAnchorSrcSpan` on the fid to get source location.

PR #7681, opened 2026-03-18
Mystery

zliu41:
This PR simply adds `anotherUnitval` in `PlutusTx.Builtins.Internal`, and define it in `defineBuiltinTerms`. Somehow, this changes the UPLC of the constitution scripts, but not the PIR (at least when printed in readable format without uniques)!! I'm completely dumbfounded.

Whoever solves this mystery gets a mystery prize.


SeungheonOh:
It's probably due to inliner pass count/limit. Adding new definition to the compiler will add extra let binding on the pre-optimized pir. Probably having extra unitVal bind makes it take up extra inliner pass and prevent further inlining

SeungheonOh:
I'm trying to test it but nix started to rebuild ghc again... 😢

SeungheonOh:
I'm right: https://github.com/IntersectMBO/plutus/compare/master...sho/reasons

zliu41:
How is it the case, though, that the PIRs are the same but UPLCs are different? Indeed PIR's DCE should immediately remove `anotherUnitval`. Is it somehow re-introduced during the PIR to UPLC transformation? I don't recall that this is a thing?

zliu41:
Also, changing the simplifier options on master also changes the outcome. So this doesn't completely explain why introducing `anotherUnitval` matters. It is supposed to be eliminated by PIR's DCE pass - end of story.

SeungheonOh:
It actually makes sense because having extra builtin definition changes unique name allocations at PIR. so even if DCE removes anotherUnitval, it causes other binds to be shifted. 

This actually causes problem in CSE since it doesn't use alpha equivalence so it is sensitive to how uniques are set. With this, all of the compilation/optimization pipeline uses same QuoteT, so adding/changing how builtins are set effects CSE even if the actual binds are removed by DCE. 

Maybe we should make independent optimization pass run on fresh QuoteT environment..?

PR #7688, opened 2026-03-24
Mutually recursive PIR function inlining

SeungheonOh:
New optimization pass that will inline mutually recursive PIR functions when they satisfy below criteria.

Given chain of mutually recursive functions f, g, and h, we inline when
- only one of the function is used in the body of recursive let
- the other functions(helper) are used only single time the body of other function
- All of the mutually recursive functions are not self recursive


cardano-constitution/test/Cardano/Constitution/Validator/Data/GoldenTests/unsorted.golden.uplc
```diff
@@ -805,7 +797,15 @@ program
                                                                                                            (constr 3
                                                                                                               [ (constr 1
                                                                                                                    [ cse
-                                                                                                                   , cse ]) ]))
+                                                                                                                   , (constr 1
```
SeungheonOh: I think we have issue similar to #7681 here. CSE not doing it's job probably due to some rename happening

plutus-core/plutus-ir/test/PlutusIR/Compiler/Recursion/even3.golden
```diff
@@ -10,69 +10,28 @@ program
...
-                  (/\Q ->
-                     \(choose : (Nat -> Bool) -> (Nat -> Bool) -> Q)
-                      (even : Nat -> Bool)
-                      (odd : Nat -> Bool) ->
-                       choose
-                         (\(n : Nat) ->
-                            match_Nat n {Bool} True (\(p : Nat) -> odd p))
-                         (\(n : Nat) ->
-                            match_Nat n {Bool} False (\(p : Nat) -> even p)))))
+             (\(s :
```
SeungheonOh, line 13: this was the existing test cases that was originally inefficient due to mutual recursion but now more efficient.

plutus-core/plutus-ir/src/PlutusIR/Transform/RecInline.hs
```diff
@@ -0,0 +1,267 @@
...
+          ]
+      roots =
+        Map.keysSet bindingMap
+          `Set.intersection` (bodyUsed `Set.union` passthroughUsed)
+  pure (buildGraph (fmap key eligible) bindingMap roots, passthrough)
+  where
+    asRecBinding = \case
+      TermBind bindAnn strictness decl rhs
+        | let arity = rhsArity rhs
+        , not (null arity) ->
```
zliu41: Why do you require this condition? Can you just use the purity analysis instead?

plutus-core/plutus-ir/src/PlutusIR/Transform/RecInline.hs
```diff
@@ -0,0 +1,267 @@
...
+inlineCallsOf helperName helperRhs = go
+  where
+    go term = do
+      term' <- traverseOf termSubterms go term
+      case splitApplication term' of
+        (Var _ name, args) | name == helperName -> do
+          rhs' <- PLC.rename helperRhs
+          pure $ case saturates args (rhsArity rhs') of
+            Just Saturated -> fillAppContext rhs' args
+            _ -> term'
```
zliu41: I don't think you need to give up if it is OverSaturated?

plutus-core/plutus-ir/src/PlutusIR/Transform/RecInline.hs
```diff
@@ -0,0 +1,267 @@
...
+
+{-| Mutual recursion inlining.
+
+Given a @let rec@ group, this pass identifies helper bindings — those not
+used by the body (non-roots), not self-recursive, called from exactly one
+sibling, and used exactly once there — and inlines them into their caller.
+This works across independent subgroups within the same @let rec@, e.g.
+@{even, odd, f, g}@ where @{even, odd}@ and @{f, g}@ are separate cycles.
+
+No beta reduction is performed here; the resulting unsaturated applications
```
zliu41:
```suggestion
No beta reduction is performed here; the resulting applications
```

The resulting application isn't necessarily unsaturated, is it?

plutus-core/plutus-ir/src/PlutusIR/Transform/RecInline.hs
```diff
@@ -0,0 +1,267 @@
...
+collapseRecGroup group =
+  case findCandidate group of
+    Nothing -> pure group
+    Just (hostKey, helperKey) ->
+      tryInline group hostKey helperKey >>= \case
+        Just group' -> collapseRecGroup group'
+        Nothing -> pure group
+
+{-| Find a helper eligible for inlining: not a root, not self-recursive,
+called from exactly one sibling, and used exactly once in that sibling. -}
```
zliu41: These "exactly one" conditions seem quite restrictive. Can you use similar criteria as the main inliner - e.g., if it is used exactly once, then it's always OK. Otherwise, it is OK if `costIsAcceptable` and `sizeIsAcceptable`.

plutus-core/plutus-ir/src/PlutusIR/Transform/RecInline.hs
```diff
@@ -0,0 +1,267 @@
...
+        [h] -> Just h
+        _ -> Nothing
+      helperBinding <- Map.lookup helper (rgBindings group)
+      hostBinding <- Map.lookup host (rgBindings group)
+      -- Must appear exactly once in the host so inlining doesn't duplicate work.
+      guard (Usages.getUsageCount (rbName helperBinding) (rbUsages hostBinding) == 1)
+      pure (host, helper)
+
+{-| Inline all saturated calls to the helper within the host's RHS.
+If successful (all references eliminated), remove the helper from the group. -}
```
zliu41: > Inline all saturated calls to the helper

Doesn't `findCandidate` guarantee that there's exactly one callsite?

And why must the callsite be saturated?

plutus-core/plutus-ir/src/PlutusIR/Transform/RecInline.hs
```diff
@@ -0,0 +1,267 @@
...
+  :: MonadQuote m
+  => RecGroup uni fun a
+  -> m (RecGroup uni fun a)
+collapseRecGroup group =
+  case findCandidate group of
+    Nothing -> pure group
+    Just (hostKey, helperKey) ->
+      tryInline group hostKey helperKey >>= \case
+        Just group' -> collapseRecGroup group'
+        Nothing -> pure group
```
zliu41: Is it possible that another candidate in the group can be inlined?

zliu41 (commented):
Can you check if this solves the problem described in Note [Making the cons case non-strict in caseList and caseList']?

plutus-core/plutus-ir/src/PlutusIR/Transform/RecInline.hs
SeungheonOh: I think if it's a non-lambda term, it should be handled by inliner instead.

plutus-core/plutus-ir/src/PlutusIR/Transform/RecInline.hs
SeungheonOh: nice. I'll do this

plutus-core/plutus-ir/src/PlutusIR/Transform/RecInline.hs
```diff
@@ -0,0 +1,267 @@
...
+
+-- | Walk the term bottom-up, attempting to collapse each @let rec@ group.
+recInline
+  :: MonadQuote m
+  => Term TyName Name uni fun a
+  -> m (Term TyName Name uni fun a)
+recInline = go
+  where
+    go term = do
+      term' <- traverseOf termSubterms go term
```
basetunnel, line 95: Doesn't `traverseOf termSubterms` already do the bottom-up traversal itself? I don't understand why recursion with `go` is needed.

Unisay (approved):

PR #7711, opened 2026-04-09
Add --certified-opts-only

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1477

This new CL option adds a mode in which the UPLC optimizer does not run those optimizations which are not yet certified.

We also do not want `plutus-core` to depend on `plutus-metatheory`, therefore we have to enforce on the Haskell side whether a pass is certified or not. These types need to share the same representation on the Agda side, which is enforced by Agda's Haskell (MAlonzo) backend.

This PR also fixes the naming problem between which passes are "simplifications" and which aren't. Therefore, many modules/types have now been renamed from simplify/simplifier to optimize/optimizer etc.

I tested by switching `--certified-opts-only` on for the `n-queens` report, and I can confirm that no uncertified passes are run. **Question to reviewers:** should we add this version of the test as well?


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Simplify/Opts.hs
```diff
@@ -11,15 +11,25 @@ module UntypedPlutusCore.Simplify.Opts
...
   , defaultSimplifyOpts
+  , CseWhichSubterms (..)
   ) where
 
 import Control.Lens.TH (makeLenses)
 import Data.Default.Class
 
 import PlutusCore.Annotation (InlineHints (..))
 import PlutusCore.AstSize
-import UntypedPlutusCore.Transform.Cse (CseWhichSubterms (..))
```
ana-pantilie, line 22: I made this change because having this dependency here broke the hierarchy of the modules. The reason is it could easily produce a cyclic dependency if any of the optimization passes would require access to the simplifier options, which is a very natural dependency which may arise in the future.

plutus-metatheory/test/certifier-report/golden/n-queens.golden.report
```diff
@@ -16,7 +16,7 @@ Pass 2: Float Force into Case Branches  ✅
   Program Size: 3504 (after)
   Execution Cost: CPU = 288404181782, MEM = 1752469055 (before)
   Execution Cost: CPU = 283210581782, MEM = 1720009055 (after)
-
+  Optimization sites: 18
```
ana-pantilie, line 19: These additions come from a bug fix included in this PR: the report wasn't producing numbers for this optimization pass.

plutus-ledger-api/executables/src/PlutusCore/Executable/Parsers.hs
```diff
@@ -270,6 +269,14 @@ simplifyOpts = do
       ( long "opt-no-apply-to-case"
           <> help "Disable apply-to-case optimization"
       )
+  _soSafeOpts <-
+    flag
+      False
+      True
+      ( long "safe-optimisation"
```
zliu41: We already have `conservative optimisation`, so adding `safe-optimisation` is very confusing. A better name is `certified-only` or `certified-opts-only`

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Certify/Trace.hs
```diff
@@ -9,20 +10,77 @@ import UntypedPlutusCore.Transform.Certify.Hints qualified as Certify
...
 import GHC.Generics
 
-data SimplifierStage
+{-| Datatype which represents optimization passes which are also
+certified (__I__mplemented in the __C__ertifier).
+
+This means that these passes are formalized as part of the certifier,
+and adding a new pass constructor to this type means that it is expected
+the pass will be also certified in the same PR. -}
+data ICSimplifierStage
```
zliu41: We need to fix incorrect uses of word "Simplifier". Simplifier refers to a subset of optimizaitons. CSE and ApplyToCase are not part of the simplifier. Almost all uses of "simplifier" in Agda is wrong. It should be called "optimizer" "optimization" or just "opt".

Also, "SimplifierStage" is fully spelled out, but "IC" is an acronym, which sounds strange. Consider: "OptStageWithCertifier" or "CertifiedOptStage"

plutus-metatheory/src/VerifiedCompilation/Trace.lagda.md
```diff
@@ -19,23 +19,50 @@ open Maybe
...
+  forceDelayT : ICSimplifierTag
+  forceCaseDelayT : ICSimplifierTag
+  caseReduceT : ICSimplifierTag
+  inlineT : ICSimplifierTag
+  cseT : ICSimplifierTag
+  applyToCaseT : ICSimplifierTag
+
+SimplifierTag = Utils.Either NICSimplifierTag ICSimplifierTag
+
+floatDelayTag : SimplifierTag
```
zliu41: The existence of both `floatDelayT` and `floatDelayTag` is not good naming, since "T" also means "Tag". You can use different capitalization: `floatDelayT` vs. `FloatDelayT`. Or `floatDelayTᵣ` and `letFloatOutTₗ`.

plutus-tx-plugin/src/PlutusTx/Options.hs
```diff
@@ -85,6 +85,7 @@ data PluginOptions = PluginOptions
     _posRemoveTrace :: Bool
   , _posDumpCompilationTrace :: Bool
   , _posCertify :: Maybe String
+  , _posSafeOpts :: Bool
```
zliu41: I don't see any use of this variable outside of this module.

Also, `plinth-compiler-options.md` needs to be regenerated.

plutus-metatheory/src/FFI/AgdaUnparse.hs
```diff
@@ -50,17 +50,18 @@ instance AgdaUnparse AgdaFFI.UTerm where
...
-  agdaUnparse ForceCaseDelay = "forceCaseDelayT"
-  agdaUnparse CaseOfCase = "caseOfCaseT"
-  agdaUnparse CaseReduce = "caseReduceT"
-  agdaUnparse Inline = "inlineT"
-  agdaUnparse CSE = "cseT"
-  agdaUnparse ApplyToCase = "applyToCaseT"
-  agdaUnparse LetFloatOut = "letFloatOutT"
-  agdaUnparse Unknown = "unknown"
+instance AgdaUnparse ICSimplifierStage where
+  agdaUnparse FloatDelay = "(inj₂ floatDelayT)"
```
zliu41: Why do you need `inj₂` if the type is `ICSimplifierStage`, which isn't an `Either`?

zliu41 (changes requested):

plutus-tx-plugin/src/PlutusTx/Options.hs
ana-pantilie: Good catch, I completely forgot to wire this to the plugin!

plutus-metatheory/src/FFI/AgdaUnparse.hs
ana-pantilie: Good catch, thanks!

zliu41 (approved):

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Optimize.hs
```diff
@@ -0,0 +1,168 @@
...
+      runStage FloatDelayStage
+        >=> runStage ForceCaseDelayStage
+        >=> runStage LetFloatOutStage
+        >=> runStage ForceDelayStage
+        >=> runStage CaseOfCaseStage
+        >=> runStage CaseReduceStage
+        >=> runStage InlineStage
+
+    runStage stage' =
+      let isIC = either (const False) (const True) stage'
```
zliu41: `let certified = isRight stage'`

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Optimize.hs
```diff
@@ -0,0 +1,168 @@
...
+          withCertifiedOptsOnly action
```
zliu41, line 115: This can be simplified (I thought I commented on this yesterday, but I probably accidentally deleted it)

```haskell
runStage stage'
  | certifiedOnly && not certified = pure
  | otherwise = action
```

zliu41 (commented):
A minor comment

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Optimize.hs
zliu41: No need to have `withCertifiedOptsOnly `:

```haskell
certifiedOnly = _ooCertifiedOptsOnly opts

runStage stage'
  | certifiedOnly && not certified = pure
  | otherwise = action
  where
  certified = isRight stage'
  action = case stage' of FloatDelayStage -> floatDelay; ...
```

zliu41:
Can you address https://github.com/IntersectMBO/plutus/pull/7711#discussion_r3099539307?

ana-pantilie:
@zliu41 I saw your comment after already fixing the conflicts with master. I'll address your comment in a subsequent PR, I don't want to block this one any longer because CI is already taking ages

zliu41:
I can force merge

ana-pantilie:
@zliu41 Pushed the change, thanks for comment!

plutus-metatheory/src/VerifiedCompilation.lagda.md
```diff
@@ -59,54 +59,52 @@ open import VerifiedCompilation.Certificate hiding (_>>=_)
...
-mRelationOf forceDelayT     = just UFD.ForceDelay
-mRelationOf caseReduceT     = just UCR.UCaseReduce
-mRelationOf cseT            = just UCSE.UntypedCSE
-mRelationOf inlineT         = just (UInline.Inline (λ()) UInline.□)
-mRelationOf unknown         = nothing
-mRelationOf caseOfCaseT     = nothing -- FIXME: https://github.com/IntersectMBO/plutus-private/issues/2054
-mRelationOf forceCaseDelayT = just UFCD.ForceCaseDelay
-mRelationOf applyToCaseT    = just UA2C.UApplyToCase
-mRelationOf letFloatOutT    = nothing
+-- TODO: this should be enforced at the type level somehow
```
basetunnel, line 71: Since this function is called in the type of a certificate, it is already enforced at the type level.

PR #7712, opened 2026-04-09
Plinth INLINEABLE no more

zliu41:
Added a `typeCheckResultAction` that automatically inserts INLINEABLE pragmas. Requires compiling with `plinthc` or `plc` instead of `compile`.

These caused some changes to unrelated golden tests, presumably because the added INLINEABLE pragmas caused GHC to inline more than before.


basetunnel:
That's great! Why doesn't it work with `compile`?

zliu41:
> That's great! Why doesn't it work with compile?

None of the frontend plugins we recently added work with `compile`, because frontend plugins cannot be installed via TH - TH runs after the frontend.

Unisay (approved):

PR #7714, opened 2026-04-10
Split plutus-tx-plugin-tests

zliu41:
This test suite is very big and takes a long time to run.


Unisay:
The new test suites `assocmap-plugin-tests` and `list-plugin-tests` contain golden tests, but `nix/project.nix` only sets `--no-create` for the original suite:

```nix
plutus-tx-plugin.components.tests.plutus-tx-plugin-tests.testFlags = [ "--no-create" ];
plutus-tx-plugin.components.tests.size.testFlags = [ "--no-create" ];
```

Should we add the same for the two new suites?

```nix
plutus-tx-plugin.components.tests.assocmap-plugin-tests.testFlags = [ "--no-create" ];
plutus-tx-plugin.components.tests.list-plugin-tests.testFlags = [ "--no-create" ];
```

Without this, if a golden file is accidentally missing, CI would silently create it instead of failing.

---

As a minor note: most of these test groups (`List property tests`, `Map property tests`, `BuiltinList/Budget`, `DataList/Budget`, `Array`, etc.) can already be run in isolation via `--test-options='-p "group name"'` without splitting into separate suites. The exception is AssocMap golden tests — their Tasty group is named `Budget` (via `testNested "Budget"` in `AssocMap/Spec.hs`), which collides with the main `Budget` group. If the test groups were given unique names (e.g. `AssocMap/Budget`), all of them could be filtered with `-p`. That said, separate suites do give the added benefit of Hydra-level parallelism and independent failure reporting.

zliu41:
> The new test suites assocmap-plugin-tests and list-plugin-tests contain golden tests, but nix/project.nix only sets --no-create for the original suite

Good catch. I've addressed this.

plutus-tx-plugin/list-test/Spec.hs
```diff
@@ -0,0 +1,24 @@
...
+import Test.Tasty.Extras (embed, runTestNested)
+
+main :: IO ()
+main = defaultMain tests
+
+tests :: TestTree
+tests =
+  runTestNested
+    ["list-test"]
+    [ Array.smokeTests
```
ana-pantilie, line 19: Why are we running array tests in the list test suite?

zliu41: lists and arrays are similar structures, and I didn't want to call it `array-and-list-plugin-tests`

ana-pantilie: It's confusing, I'd prefer the longer name

PR #7715, opened 2026-04-13
Simplify inline hints by reordering `check` arguments

zliu41:
This allows removing the `ƛ↓` constructor of `InlineHints`, simplifying inliner implementation. To do so, the case tree of `check` must first split on the pre-term, then `zz`, hints, and post-term. So `check`'s arguments should be in this order.


plutus-metatheory/src/VerifiedCompilation/UInline.lagda.md
basetunnel, line 203: It might be good to put a comment here about why the order of arguments in `check` matters here.

basetunnel (approved):

PR #7718, opened 2026-04-14
Enable UPLC `LetFloatOut` on `Apply` nodes

zliu41:
Turns out the Marlowe test failed previously not because there was an optimizer bug, but because the Marlowe validators are sensitive to script hashes. We just needed to run `cabal run marlowe-validators` to regenerate the flat.


plutus-benchmark/nofib/test/9.6/knights10-4x4.golden.eval
```diff
@@ -1,6 +1,6 @@
-CPU:             1_010_094_754
-Memory:              5_311_458
-AST Size:                1_659
-Flat Size:               1_711
+CPU:             1_368_945_656
```
basetunnel, line 1: This is a significant change

zliu41: I think I know why: by floating bindings out of `apply`, we turn `(\x y. x + y) X Y` into `(\x. (\y. x + y) Y) X`. The inliner can inline `x` and `y` in the first term, but not the second, because of how `extractApps` works.

It's a bit tricky to update the inliner to inline `x` and `y` in the second term, because in the second term, it is possible that `Y` mentions `x`, which complicates things. I think we should update the `LetFloatOut` pass to float out multi-lets.

zliu41: Well, turns out this is not the reason. Then I don't know what caused this - need to dig further.

zliu41: Ah, this could also interact badly with CSE. By turning `(\x. body)  X Y` into `(\x. body Y) X`, we put `Y` under a lambda, which makes it ineligible for CSE. Perhaps we should update CSE to "see through" immediately applied lambdas.

basetunnel: Makes sense. But if CSE can see through let, wouldn't it also pull out `L` here? L could be expensive:
```uplc
constr 0
  [ let x = error in L
  , let y = error in L
  ]
```

zliu41: That's no different than pulling `L` out of `constr 0 [error L, error L]` which it already does. If it's going to error anyway then I don't think it matters.

basetunnel: I see. But that still makes an erroring program unnecessarily slower, right? Could we improve CSE by noticing there is `error` on the evaluation path, and then similar to `LamAbs`, `Delay` and `Case` branches, not pull out expressions across the error?

plutus-tx-plugin/test/DataList/Budget/9.12/makeList.golden.uplc
```diff
@@ -2,10 +2,12 @@
...
       (\cons ->
-         (\f g x -> f (g x))
-           (cons (\i -> iData i) x)
-           ((\f g x -> f (g x))
-              (cons (\i -> iData i) y)
-              (cons (\i -> iData i) z)))
-        (\`$dToData` h eta -> force mkCons (`$dToData` h) eta)
-        []))
\ No newline at end of file
+         case
```
SeungheonOh, line 5: Nice, CaseApply gets applied here after float out

SeungheonOh (approved):
Looks good. Most small improvements comes from let floatout enabling CaseApply pass to engage.

plutus-benchmark/nofib/test/9.6/knights10-4x4.golden.eval
zliu41: This is fixed in #7720.

zliu41: > But that still makes an erroring program unnecessarily slower, right?

If something guarantees to fail, we don't care about the cost. If it _may_ fail (e.g., `error` is in one of the branches), then we shouldn't give up CSE opportunities just because it could make the `error` branch more expensive.

basetunnel:
I think we shouldn't have to regenerate the marlowe uplc files, but instead update the logic that injects the hashes: https://github.com/IntersectMBO/plutus-private/issues/2168

PR #7719, opened 2026-04-14
Fix BuiltinCasing crash on GHC.Prim.Addr# (#7716)

### LINKED ISSUE #7716: BuiltinCasing crashes on GHC.Prim.Addr# when a join point exposes Data in its type
The plugin crashes when compiling a function with `datatypes=BuiltinCasing` if the GHC Core contains a join point with `PlutusCore.Data.Data` in its type:

```
Context: Compiling type: PlutusCore.Data.Data
Context: Compiling data constructor type: PlutusCore.Data.B
Context: Compiling type: Data.ByteString.Internal.Type.ByteString
Context: Compiling data constructor type: Data.ByteString.Internal.Type.BS
Context: Compiling type: GHC.Prim.Addr#
Error: Unsupported feature: Type constructor: GHC.Prim.Addr#
```

## What happens

`BuiltinData` wraps `PlutusCore.Data.Data`. Normally the plugin treats `BuiltinData` as a builtin and never looks inside. But GHC's simplifier can produce join points where `Data` appears naked in the type signature:

```
$j :: Data -> BuiltinUnit
```

The plugin with BuiltinCasing tries to compile `Data` as a regular ADT, follows `B ByteString` → `BS Addr# Int`, and crashes because `Addr#` has no Plutus Core equivalent.

## Reproduction

Branch with a failing test: [`yura/builtin-casing-addr-joinpoint`](https://github.com/IntersectMBO/plutus/tree/yura/builtin-casing-addr-joinpoint)

Minimal trigger — `caseList'` applied twice to the same value in a separate module without BuiltinCasing:

```haskell
failsToCompile :: BuiltinData -> BuiltinUnit
failsToCompile bd =
  case toBuiltinData (firstOf items) of
    _ -> case toBuiltinData (firstOf items) of
      _ -> unitval
  where
    items = unsafeFromBuiltinData bd
    firstOf = caseList' Nothing (\(h :: BuiltinData) _t -> Just h)
```

Compiled from `Spec.hs` (which has `datatypes=BuiltinCasing`) via `$$(compile [||Lib.failsToCompile||])`.

```bash
cabal build plutus-tx-plugin-tests
```

No dependencies beyond `plutus-tx` and `plutus-tx-plugin`.

## Originally discovered as

A cross-library compilation failure: a Plinth validator using `findOwnInput` + `getContinuingOutputs` from `plutus-ledger-api` crashed when compiled with BuiltinCasing. `getContinuingOutputs` calls `findOwnInput` internally, producing the double-call pattern that triggers the join point.

Unisay:
## Summary

- Add unreachable second constructor to `BuiltinData` to prevent GHC's case-of-single-constructor optimization from exposing `PlutusCore.Data.Data` in join point types
- Add `COMPLETE` pragma so existing pattern matches remain exhaustive
- Add failing test (`BuiltinCasing/Lib.hs`) that reproduces the crash

Closes #7716

Alternative to #7863, which fixes the same issue by compiling `BuiltinData` transparently in the plugin instead of changing the type. The two PRs are mutually exclusive — review both and choose one; whichever lands, the other closes.

## The problem

`data BuiltinData = BuiltinData ~Data` has one constructor. GHC's simplifier always unwraps single-constructor types, producing `case bd of { BuiltinData d -> ... }` in Core. This leaks `d :: Data` into join point type signatures. The plugin then tries to compile `Data` as a regular ADT, follows `B ByteString` to `BS Addr# Int`, and crashes — `Addr#` has no Plutus Core equivalent.

## The fix

Add a second constructor `BuiltinDataUnreachable` that is never constructed. GHC won't case-simplify multi-constructor types, so `Data` stays behind the `BuiltinData` wrapper. A `COMPLETE` pragma marks the original constructor as exhaustive.

Details in Note [Opaque builtin types] in `PlutusTx.Builtins.Internal`.

## Alternatives I tried

- Hiding the constructor via export lists — GHC includes all constructors in `.hi` files regardless of what the module exports
- `newtype BuiltinData = UnsafeBuiltinData Any` — newtypes are transparent in Core (coercions), so the plugin just crashes on `Kind: forall k. k` instead
- Mapping `Data` to the builtin type in the plugin — fixes the Addr# crash but needs deeper changes to the expression compiler; now implemented as the alternative PR #7863


zliu41:
I'll need to understand what happened a bit more. There may be simpler solutions.

The issue description says "But the plugin's own simplifier pass (mkSimplPass in Plugin/Common.hs) can produce join points" - that's false. The plugin cannot produce join points.

In general I'd recommend against copy-pasting large amount of AI text. I generally find it verbose, with low signal-to-noise ratio, and unpleasant to read. It's better to use AI to understand the issue, then write the description on your own.

In this particular case, what would be a useful issue description is: "for this Haskell function, without builtin casing, it generates this GHC Core, which the plugin can compile. But with builtin casing, the GHC Core becomes this, which is problematic". It's very useful to include GHC Core, and anything else is unnecessary.

Unisay:
Thanks for the feedback — you're right, I was wrong about `mkSimplPass` producing join points. Fixed the issue and PR description.

Here's the minimal Core showing the problem. Test function:

```haskell
useTwiceData :: BuiltinData -> BuiltinUnit
useTwiceData bd =
  case toBuiltinData (firstOf items) of
    _ -> case toBuiltinData (firstOf items) of
      _ -> unitval
  where
    items = unsafeFromBuiltinData bd
    firstOf = caseList' Nothing (\(h :: BuiltinData) _t -> Just h)
```

GHC Core produced (without BuiltinCasing, stored in the `.hi` file):

```
useTwiceData :: BuiltinData -> BuiltinUnit
useTwiceData
  = \ (bd :: BuiltinData) ->
      case bd of bd1 { BuiltinData ipv ->        -- ← BuiltinData unwrapped
      case unsafeDataAsList of g1 { __DEFAULT ->
      case g1 bd1 of nt { BuiltinList ipv1 ->
      join {
        $j :: Data -> BuiltinUnit                 -- ← Data in join point type
        $j _
          = case caseList'
                   @BuiltinData
                   @(Maybe BuiltinData)
                   (Nothing @BuiltinData)
                   (\ (x :: BuiltinData) (eta :: BuiltinList BuiltinData) ->
                      case x of x1 { BuiltinData ipv3 ->
                      case eta of { BuiltinList ipv4 ->
                      Just @BuiltinData x1
                      }})
                   nt
            of {
              Nothing -> case mkConstr (IS 1#) (mkNilData unitval) of
                           { BuiltinData ipv3 -> unitval };
              Just arg -> case mkConstr (IS 0#) (mkCons @BuiltinData arg ...) of
                            { BuiltinData ipv3 -> unitval }
            }
      } in ...
```

The plugin (with BuiltinCasing) reads this Core via the interface file and tries to compile the type `Data` that appears naked in `$j :: Data -> BuiltinUnit`. It hits `B ByteString -> BS Addr# Int` and crashes.

Without the second constructor, GHC's simplifier unwraps `BuiltinData` via case-of-single-constructor at every use site, and `$j`'s captured free variables pull the unwrapped `Data` into the join point's type signature. With two constructors, GHC can't apply case-of-single-constructor, so `Data` stays hidden behind the `BuiltinData` wrapper and the plugin never sees it.

PR #7720, opened 2026-04-15
Make CSE see through bindings

zliu41:
Normally if an expr occurs under two different lambdas, they shouldn't be considered common subexpressions, but it's OK to ignore immediately applied lambdas.

For example, in `f ((\x -> ...expr...) X) ((\y -> ...expr...) Y)`, the two occurrences of `expr` can be considered common subexpressions.

This is sometimes slightly better and sometimes slightly worse, but this fixes the regression in https://github.com/IntersectMBO/plutus/pull/7718#discussion_r3080818105 so it's worth doing.


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
```diff
@@ -421,10 +421,81 @@ applyCse c = mkLamApp . transformOf termSubterms substCseVarForTerm
...
+                  ]
+            | otherwise -> wrapWithCse node
+          Case ann scrut branches
+            | not (any (cseName `occursIn`) branches) ->
+                -- Descend into the scrutinee, if no branch uses `cseName`.
+                Case ann (placeCseBinding scrut) branches
+            | otherwise -> wrapWithCse node
+          _ -> wrapWithCse node
+          where
+            cseName = ccFreshName c
```
ana-pantilie: I had trouble understanding the code because I thought `ccFreshName` would produce a new variable name based on the current context! Instead, it just reads the variable name for the current cse candidate. Maybe `ccFreshName` should be named differently, and the `c` argument here should be renamed to `candidate` or something more descriptive.

ana-pantilie (commented):
Now that cse has changed, we should put cse in the "not certified" category.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Cse.hs
zliu41: It is a fresh name that was made earlier. Renaming to `candidate` sounds good, although "context" is rarely abbreviated as "c".

zliu41 (commented):
> Now that cse has changed, we should put cse in the "not certified" category.

I don't think this breaks the certifier.

ana-pantilie:
> I don't think this breaks the certifier.

Won't it now do optimizations which are not described in the translation relation for cse?

There's also this TODO here https://github.com/IntersectMBO/plutus/blob/edf4c582f4c245922c5dc3285482454716599edb/plutus-metatheory/src/VerifiedCompilation/UCSE.lagda.md#L44 and I think we realized last week that we probably need to review the current translation relation and possibliy redo it.

ana-pantilie:
Looking at the translation relation I think it should indeed handle this new case you're adding, but I'm still worried about the TODO there, hmm.

zliu41:
The CSE certifier needs some work indeed, as the TODO says. It currently doesn't care if we move things across lambdas, which is why this PR doesn't break it.

ana-pantilie (approved):

PR #7721, opened 2026-04-16
Add CLAUDE.md

ana-pantilie:
### From Claude:

 What CLAUDE.md does:
  - It's loaded automatically into my context at the start of every conversation in this repo.                                                
  - It gives future Claude Code sessions immediate awareness of things that would otherwise require exploring many files to discover — build  
  commands, architecture, conventions, etc.                                                                                                   
                                                                                                                                              
  How that affects my work:
  - When you ask me to add a feature, I already know the compilation pipeline (Haskell → PIR → TPLC → UPLC) without having to rediscover it.  
  - When I write new Haskell, I know to use StrictData, StandaloneKindSignatures, deriving strategies, write type signatures, etc. without you
   having to correct me.                                                                                                                      
  - When tests change, I know to suggest regenerating goldens with --accept and to do it per-GHC-version for the plugin.                      
  - When I suggest a PR, I know changelogs are required via scriv.                                                      
                                                                                                                                              
  What it doesn't do:                                                                                                                         
  - It doesn't auto-generate code, docs, or tests on its own.                                                                                 
  - It doesn't replace reading the actual source — for any specific task I still read the relevant files. CLAUDE.md gives orientation, not    
  complete knowledge.                                                                                                                     
                                                                                                                                              
  Think of it as the onboarding doc you'd give a new team member: it doesn't write code for them, but it means they don't spend the first week
   asking "where does this live?" and "how do I run tests?".


zliu41:
Can you describe how the file was produced. 100% claude? Claude plus something something you wrote?

ana-pantilie:
Discussed during meeting, decided to close. Claude itself changes too often, we should instead modify the human documentation whenever it's hard for Claude/a new team member to figure something out from the code.

PR #7723, opened 2026-04-17
fix(bench): prevent ValueData benchmark crash

Unisay:
Closes https://github.com/IntersectMBO/plutus-private/issues/2161

- [x] Exported `valueDataMaxSize` from `PlutusCore.Value`.
- [x] Bound `generateTestValues` to `Value.valueDataMaxSize` (replaces the local `maxValueEntries = 50_000`), so every generated `Value` satisfies `totalSize v <= valueDataMaxSize` by construction.
- [x] Removed the now-obsolete `-- FIXME: ...` comment in `Benchmarks/Values.hs`.
- [x] Enabled `NumericUnderscores` in `plutus-core`'s shared `default-extensions` and rewrote `valueDataMaxSize = 40000` as `40_000`.

### DIFF
--- a/plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
+++ b/plutus-core/cost-model/budgeting-bench/Benchmarks/Values.hs
@@ -281,9 +281,7 @@ unValueDataBenchmark gen =
     DataNodeCount
     UnValueData
     []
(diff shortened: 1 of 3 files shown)


kwxm (approved):
Yes, that looks as if it should fix our problem.

PR #7725, opened 2026-04-19
Remove PlutusTx.Plugin and `plc`; make `injectAnchors` optional for `plinthc`

zliu41:
`plinthc` is `plc` plus `injectAnchors`. By making `injectAnchors` optional, there's no longer a need for `plc`.

`plinthc` is more user friendly since it doesn't have the `Proxy loc` argument.

`injectAnchors` is an experimental feature, so it is off by default for now.


doc/docusaurus/docs/using-plinth/compiling-plinth.md
```diff
@@ -53,15 +53,15 @@ An alternative way to compile `myPlutusTxCode` is by using the `-fplugin` GHC fl
...
 import PlutusTx.Code (CompiledCode)
-import PlutusTx.Plugin (plc)
+import Plinth.Plugin (plinthc)
 
 myPlutusTxCodeCompiled :: CompiledCode (Integer -> Integer)
-myPlutusTxCodeCompiled = plc (Proxy @"location info") myPlutusTxCode
+myPlutusTxCodeCompiled = plinthc myPlutusTxCode
 ```
 
 Here you can manually provide the location info to be included in error messages, but it's not essential, especially if `plc` is only called once in the module, since there won't be any confusion about which `plc` is causing the issue if the module fails to compile.
```
Unisay, line 67: This paragraph is now outdated after replacing `plc` with `plinthc`:
- "manually provide the location info" — `plinthc` takes no location argument; source locations are now opt-in via the `preserve-source-locations` plugin flag (experimental)
- "which `plc` is causing the issue" — `plc` no longer exists

The section heading on line 54 also still reads "Use this flag with the `plc` function".

Suggested replacement: *"With `plinthc` there is no need to manually specify location info. Source locations can optionally be preserved via the `preserve-source-locations` plugin flag (experimental)."*

Unisay (commented):
- `plutus-benchmark/plutus-benchmark.cabal` line 32: the comment still refers to `PlutusTx.Plugin()` — should be updated to `Plinth.Plugin()`.
- `UntypedPlutusCore/Evaluation/Machine/Cek/EmitterMode.hs` line 76: comment says `PlutusTx.Plugin:profile-all` — should be `Plinth.Plugin:profile-all`.
- `doc/docusaurus/docs/using-plinth/compiling-plinth.md`: the paragraph after the code example still refers to manually providing location info via `plc`, but `plinthc` doesn't accept a location argument. The section heading also still says "Use this flag with the `plc` function". Both need updating.

doc/docusaurus/docs/using-plinth/compiling-plinth.md
zliu41: User guide will be updated separately.

plutus-tx-plugin/test/AsData/Budget/9.12/destructSum-manual.golden.pir
```diff
@@ -31,8 +31,6 @@ let
             (unIData (headList {data} l))
             (unIData (headList {data} l))
             (unIData (headList {data} (tailList {data} l)))
-  data (Tuple :: * -> * -> *) a b | Tuple_match where
-    Tuple2 : a -> b -> Tuple a b
```
basetunnel, line 35: Why did this change? A result of disabling anchors by default?

zliu41: Yes. Anchor wrappers make things more opaque and sometimes change GHC optimizations, which was [observed when adding anchors](https://github.com/IntersectMBO/plutus/pull/7640), and that is the main reason why anchors should be optional.

basetunnel (approved):

PR #7726, opened 2026-04-20
Introduce HasAnn typeclass; fix Apply span regression

zeme-wana:
## Summary

Refactors annotation access across the `plutus-core`, `plutus-ir`, and `untyped-plutus-core` ASTs to use a single `HasAnn` typeclass, replacing the three ad-hoc functions `termAnn`, `typeAnn`, `modifyTermAnn`, and `bindingAnn`. Also fixes a `SrcSpan` regression in the parsers introduced by commit 99fa5e8 (\"Todo The Parser\") where inner `Apply` nodes in iterated applications shared the outermost bracket span instead of carrying their per-argument spans.


Unisay (approved):
Nice consolidation,
Please add a changelog entry.

PR #7728, opened 2026-04-20
Add CaseReduce to the unimplemented-in-the-certifier category

ana-pantilie:
`CaseReduce` is currently not implemented correctly in the certifier (https://github.com/IntersectMBO/plutus-private/issues/2082). We need to turn it off until it is fixed.

### DIFF
--- a/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Certify/Trace.hs
+++ b/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Certify/Trace.hs
@@ -20,7 +20,6 @@ data CertifiedOptStage
   = FloatDelay
   | ForceDelay
   | ForceCaseDelay
-  | CaseReduce
   | Inline
   | CSE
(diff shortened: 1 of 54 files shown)


zliu41:
The CaseReduce certifier is a bit more restrictive than it should, but it's almost never an issue. Have you encountered an actual instance where it fails to certify a valid optimization?

ana-pantilie:
@zliu41 Since the certifier could give incorrect results, it should be turned off until it's fixed.

I haven't confirmed whether this is the reason some of the tests on Lorenzo's branch are failing, but having this merged will make that easier.

zliu41:
I'll defer to @basetunnel - I think he already has a fix.

> I haven't confirmed whether this is the reason some of the tests on Lorenzo's branch are failing

That would be very useful to do.

basetunnel (approved):

basetunnel:
It's better to disable it while it's failing, until my fix is reviewed and merged. It's not much work for me, I'll have to rebase anyway.

PR #7729, opened 2026-04-20
Add `inline-unconditional-growth` flag, controlling aggressiveness of unconditional inlining

zliu41:
We already have the `inline-callsite-growth` flag that controls aggressiveness of callsite inlining. Adding a similar flag for unconditional inlining. This replaces the `sizeIsAcceptable` check. The default value is 1.

The flag is added to both the plugin and the uplc executable.


doc/docusaurus/docs/delve-deeper/plinth-compiler-options.md
```diff
@@ -35,12 +35,13 @@ For each boolean option, you can add a `no-` prefix to switch it off, such as `n
...
+|`inline-callsite-growth`|Int|5|Sets the inlining threshold for callsite inlining. 0 disables inlining a binding at a callsite if it increases the AST size; `n` allows inlining if the AST size grows by no more than `n`. Keep in mind that doing so does not mean the final program will be bigger, since inlining can often unlock further optimizations.|
 |`inline-constants`|Bool|True|Always inline constants. Inlining constants always reduces script costs slightly, but may increase script sizes if a large constant is used more than once. Implied by `no-conservative-optimisation`.|
 |`inline-fix`|Bool|True|Always inline fixed point combinators. This is generally preferable as it often enables further optimization, though it may increase script size.|
+|`inline-unconditional-growth`|Int|1|Sets the inlining threshold for unconditional inlining. `n` allows unconditional inlining if the AST size grows by at most `n` at each callsite (i.e., the size of the binding's RHS is at most `n+1`).|
```
basetunnel: The word "callsite" is a bit confusing here, because it is not related to callsite inlining. I would just write "variable occurrence"

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/Inline.hs
```diff
@@ -508,9 +513,15 @@ acceptable
   -- ^ inline constants
   -> Term name uni fun a
   -> InlineM name uni fun b Bool
-acceptable inlineConstants t =
+acceptable inlineConstants t = do
+  unconditionalGrowth <- view iiInlineUnconditionalGrowth
   -- See Note [Inlining criteria]
-  pure $ costIsAcceptable t && sizeIsAcceptable inlineConstants t
+  let costOk = costIsAcceptable t
+      sizeOk = termAstSize t - 1 <= unconditionalGrowth
```
basetunnel: Somehow, this logic is much easier to understand for me when it's written like  `termAstSize t <= 1 + unconditionalGrowth`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Optimize/Opts.hs
```diff
@@ -56,6 +58,7 @@ defaultOptimizeOpts =
     , _ooConservativeOpts = False
     , _ooInlineHints = def
     , _ooInlineConstants = True
+    , _ooInlineUnconditionalGrowth = 1
```
basetunnel, line 61: How did you arrive at this default value? I guess the old logic can't be exactly replicated (e.g. it accepted terms with arbitrary amounts of force/delay)

basetunnel:
> The default value is 1, i.e., does not allow AST size growth at inlining sites.

Shouldn't the default be `0` if you want to disallow any growth? Right now `termAstSize t - 1 <= unconditionalGrowth` allows a term `t` of size 2.

zliu41:
> Shouldn't the default be 0 if you want to disallow any growth?

Good catch - 1 seems the right default value but I forgot to update the description.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Optimize/Opts.hs
zliu41: 1 seems a good default value based on testing (I think the results are universally better in both size and cost). 0 is too conservative - it's more conservative than it is now, since `sizeIsAcceptable` accepts terms like `Delay x` and `Constr 0 [x]`.

zliu41: > I guess the old logic can't be exactly replicated (e.g. it accepted terms with arbitrary amounts of force/delay)

That's correct, and the old logic also wasn't good - you aren't supposed to accept arbitrary layers.

basetunnel (approved):

PR #7730, opened 2026-04-21
Add binding anchors; make `profile-all` use srcspan from anchors

### LINKED ISSUE #7722: Make profiling output include better source span

https://github.com/IntersectMBO/plutus/blob/a8a5a3519db6a95d25fff6b5be54e0067206cc4a/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs#L662-L687 

This function is used to insert call profiling information to each function calls which is used to run profiling and also call trace information. Currently, it gets source span from `GHC.Var`; source span attached to `GHC.Var` is only available if the module `GHC.Var` resides in gets compiled from scratch without compilation cache. So, to get accurate source spans for these profiling traces, one must always run `cabal clean` and if not it will add traces without source span. 

It should be possible now to get accurate srcspan without needing `cabal clean` with https://github.com/IntersectMBO/plutus/pull/7640

_No response_

SeungheonOh:
closes #7722

Added anchors to bindings; added anchors can be used in `profile-all` which is used for profiling and call tracing to mark function entrance/exit source span. 

Previously, function entrance and exit source span was extracted from GHC.Var which only gave source span when the module was freshly compiled. When module was not compiled and cache was used from previous compilation, GHC.Var will not provide any source span. New anchor based source span fixes this and provide correct source span even when module compilation is cached.

### DIFF
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
@@ -617,6 +617,7 @@ hoistExpr
 hoistExpr var t = do
   wrapUnsafeDataAsConstrName <-
     lookupGhcName 'PlutusTx.AsData.Internal.wrapUnsafeDataAsConstr
+  bindingAnchorName <- lookupGhcName 'PlutusTx.Plugin.Utils.bindingAnchor
   let name = GHC.getName var
       lexName = LexName name
 
@@ -635,7 +636,7 @@ hoistExpr var t = do
   -- See Note [Dependency tracking]
   modifyCurDeps (Set.insert lexName)
   maybeDef <- PIR.lookupTerm lexName
-  let varSpan = getVarSourceSpan var
+  let varSpan = findBindingAnchorLoc bindingAnchorName t <|> getVarSourceSpan var
       addSpan = case varSpan of
         Nothing -> id
         Just src -> fmap . fmap . addSrcSpan $ src ^. srcSpanIso
@@ -650,7 +651,7 @@ hoistExpr var t = do
         (PIR.Def var' (PIR.mkVar var', PIR.Strict))
         mempty
 
-      t' <- maybeProfileRhs var var' =<< addSpan (compileExpr Nothing t)
+      t' <- maybeProfileRhs varSpan var var' =<< addSpan (compileExpr Nothing t)
       -- See Note [Non-strict let-bindings]
       PIR.modifyTermDef lexName (const $ PIR.Def var' (t', PIR.NonStrict))
       pure $ PIR.mkVar var'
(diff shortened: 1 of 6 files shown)


plutus-tx-plugin/src/PlutusTx/Plugin/Common.hs
```diff
@@ -185,13 +188,64 @@ anchorExpr :: GHC.Id -> GHC.LHsExpr GHC.GhcTc -> GHC.LHsExpr GHC.GhcTc
...
           anchor = GHC.mkHsWrap wrapper (GHC.HsVar GHC.noExtField $ GHC.noLocA anchorId)
        in GHC.noLocA (Compat.hsAppTc (GHC.noLocA anchor) le)
-  | otherwise = le
+
+{-| Wrap the body of every 'GRHS' of a value binding with an @anchor@ carrying
+the binder's source span. This gives 'findAnchorLoc' a reliable outer-most
+anchor for every hoisted binding, so profile-trace output identifies which
+function was entered (the span points to the binder, not an arbitrary
+sub-expression). See issue #7722. -}
+anchorBinding
```
SeungheonOh, line 217: we are using same `anchor` function that's being used for anchoring expressions. It's working okay at the moment, but maybe in the future we want to separate binding anchors with some other anchor function like `bindingAnchor` so that it can be distinguished more easily

plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs
```diff
@@ -1729,6 +1736,32 @@ extractLoc anchorName modBreaks = go Nothing
             go (Just ss) e
       other -> (acc, other)
 
+{-| Find the first @anchor@-carried 'RealSrcSpan' anywhere inside a 'CoreExpr',
+descending through applications, lambdas, lets, cases, casts, and ticks.
+Useful when the immediate head of the expression is not an anchor call
```
zliu41: Do you have an example where the immediate head of the expression is not an anchor call, because it was hoisted?

SeungheonOh: Looking back, I don't think this is the case anymore after adding binding anchor.

SeungheonOh:
I redid everything with a different approach: 

instead of using original `anchor`, I created a new `bindingAnchor` specialized for creating anchors for definition bindings, which seems to work better.

zliu41:
Does `bindingAnchor` also serve the same purpose as `anchor`, i.e., improving line number reporting in error messages? Or is it only useful in profiling mode?

zliu41:
It would be useful to expand the Haddock for `anchor` and `bindingAnchor`, to make the answer to this question clear by reading them.

SeungheonOh:
`bindingAnchor` is only used for profiling mode since we only need binding information for profiling. This is unless we want to add some binding information to error messages, but I'm not sure which error message would benefit from this extra information.

PR #7733, opened 2026-04-22
Certifier: fix the CSE translation relation and re-enable it

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/2163


plutus-metatheory/src/VerifiedCompilation/UCSE.lagda.md
```diff
@@ -27,43 +27,204 @@ open import Data.Empty using (⊥)
...
+   as witnessed by `OccursOnExecPath`, a predicate that holds when a variable lies
+   on at least one execution path through a term. This guarantees soundness: the
+   transformed term is no less strict than the original.
 
 ```
+
+private variable
+  n : ℕ
+
+data OccursOnExecPath (var : Fin n) : n ⊢ → Set where
```
zliu41: This is just strictness analysis. You can reuse the definitions in the `Untyped.Strictness` module.

ana-pantilie: Great point. I added the new case to the module and rewrote the decision procedure.

plutus-metatheory/src/Untyped/Strictness.lagda.md
```diff
@@ -11,63 +11,129 @@ module Untyped.Strictness where
...
-... | no ¬ps = no λ { (constr ps) → ¬ps ps }
-x ∈↓? case M Ns with x ∈↓? M
-... | yes p = yes (scrut p)
-... | no ¬p = no λ { (scrut p) → ¬p p }
-x ∈↓? ƛ M = no λ ()
-x ∈↓? delay t = no λ ()
-x ∈↓? con c = no λ ()
-x ∈↓? builtin b = no λ ()
-x ∈↓? error = no λ ()
+_∈↓?_ : (v : Fin n) (M : n ⊢) → Dec (v ∈↓ M)
```
zliu41, line 72: This refactoring doesn't seem like an improvement. It's more verbose now.

zliu41 (approved):

plutus-metatheory/src/Untyped/Strictness.lagda.md
ana-pantilie: @zliu41 Once you add the `Let...` case, the application case becomes ambiguous in multiple ways:
- either it's a `lam body . rhs`, and you have to check whether the variable is in the body or in the rhs
- either it's not, and you check if the variable is on the left or if it's on the right and pick the first place you find it

The problem is that you need a way to write that match first, and keep the information that the match doesn't hold in the next case. I spent a lot of time yesterday trying to find a way to write this without Agda complaining. When I tried View patterns, it became much easier to keep track of the proofs that the term matches some structure or not, and it was the only way I could find to write this cleanly and have Agda be happy with it. 

Perhaps there is a cleaner way but it seemed good enough and I had already spent too much time on it.

PR #7734, opened 2026-04-22
Move CSE to the unimplemented-in-the-certifier category

ana-pantilie:
See https://github.com/IntersectMBO/plutus-private/issues/2163#issuecomment-4295734599 for the reasoning behind this change.


plutus-metatheory/test/certifier-report/golden/n-queens.golden.report
```diff
@@ -46,7 +46,7 @@ Pass 5: Case-of-Case  ⚠ (certifier unavailable)
 
 
 ──────────────────────────────────────────────────────
-Pass 6: Case-Constr and Case-Constant Cancellation  ⚠ (certifier unavailable)
+Pass 6: Common Subexpression Elimination  ⚠ (certifier unavailable)
```
basetunnel: Why did the passes change in the report?

ana-pantilie: Huh, good catch. Looks like CaseReduce became CSE and the other way around. Trying to figure out what happened...

ana-pantilie: @basetunnel I found the culprit: the positions of `CSE` and `CaseReduce` were switched in the `COMPILE` pragma, therefore Agda associated the Haskell `CaseReduceStage` to `cseT` and vice-versa. I've fixed it and added warnings in the documentation that order matters. Ideally we'd automate this, will think about if there's a convenient way to check this automatically.

basetunnel: Ah, makes sense! I guess that these kind of bugs are somewhat benign (similar to incorrect hints): worst case the certifier rejects the trace.

basetunnel (approved):

PR #7735, opened 2026-04-23
Use independent seeds in makeSizedByteStrings

zeme-wana:
## Summary

Fixes the `-- FIXME: this is terrible` in `plutus-core/cost-model/budgeting-bench/Generators.hs`.

`makeSizedByteStrings` passed the same `H.Seed` to every call of `makeSizedByteString`, so each generated ByteString was a prefix of the same deterministic byte sequence — smaller ones were literal prefixes of larger ones. This produced correlated inputs for benchmarks that depend on byte content (equality, hashing, string ops).

### DIFF
--- a/plutus-core/cost-model/budgeting-bench/Generators.hs
+++ b/plutus-core/cost-model/budgeting-bench/Generators.hs
@@ -12,13 +12,14 @@ import Control.Monad
 import Data.Bits
 import Data.ByteString (ByteString)
 import Data.Int (Int64)
-import Data.List as List (foldl')
+import Data.List as List (foldl', unfoldr)
 import Data.Text (Text)
 import Data.Word (Word64)
 
 import Hedgehog qualified as H
 import Hedgehog.Internal.Gen qualified as G
 import Hedgehog.Internal.Range qualified as R
+import Hedgehog.Internal.Seed qualified as Seed
 import Hedgehog.Internal.Tree qualified as T
 import System.IO.Unsafe (unsafePerformIO)
 import System.Random (StdGen, randomR)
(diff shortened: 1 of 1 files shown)


zliu41 (commented):
> Use independent seeds

Looks reasonable, but are you sure this is the reason for the "FIXME: this is terrible"? Who added the FIXME?

zliu41:
> /benchmark nofib

This has nothing to do with `nofib`. It updates the costing benchmark.

Unisay (commented):
Two suggestions, neither blocking:

1. Could `seeds` be a top-level binding? `unfoldr (Just . Seed.split)` is enough of a head-scratcher that I'd give it a name, and the same stream is useful for the next sized-list helper:

   ```haskell
   -- | Infinite stream of independent seeds derived from a root seed.
   splitSeeds :: H.Seed -> [H.Seed]
   splitSeeds = unfoldr (Just . Seed.split)
   ```

2. The same bug is in `makeSizedTextStrings` (line 108) and `makeSizedUtf8ByteStrings` (line 118). Both `fmap` the same seed across all sizes, so string and `decodeUtf8` benchmarks get correlated inputs too. With `splitSeeds` extracted, each fix is one line:

   ```haskell
   makeSizedTextStrings :: H.Seed -> [Integer] -> [Text]
   makeSizedTextStrings seed sizes =
     zipWith makeSizedTextString (splitSeeds seed) (fmap fromInteger sizes)

   makeSizedUtf8ByteStrings :: H.Seed -> [Integer] -> [ByteString]
   makeSizedUtf8ByteStrings seed sizes =
     zipWith makeSizedUtf8ByteString (splitSeeds seed) (fmap fromInteger sizes)
   ```

If you'd rather keep this PR scoped, I'm happy to do these in a follow-up.

kwxm:
> Looks reasonable, but are you sure this is the reason for the "FIXME: this is terrible"? Who added the FIXME?

That was me.  It was a quick implementation and I meant to go back and improve it some time. This seems to do the job though.

kwxm:
> If you'd rather keep this PR scoped, I'm happy to do these in a follow-up.

@Unisay Please do that!

kwxm (approved):
Sorry for the long delay in reviewing this!  It seems to do what's required.  I don't think the original version was actually problematic, but it certainly wasn't ideal, and this should make it less easy to get things wrong in the future.  I ran costing benchmarks for some of the bytestring functions with both this branch and the version in `master` and the results weren't siginficantly different.

The generators for the costing benchmarks have grown over time and they're a bit repetitive and unsystematic.  There a [long-standing issue](https://github.com/IntersectMBO/plutus/issues/6012) to rework them all and we should do that when we get the chance.  This improves things for the time being though.

kwxm:
(I just want to check the benchmark results.)

kwxm:
Nope, that's not going to work.

PR #7241, opened 2025-07-28
Experiment with OPAQUE in `builtinsIntroducedIn`

kwxm:
Trying @effectfully's [suggestion](https://github.com/IntersectMBO/plutus/pull/7223#discussion_r2229334293) to see if it improves the efficiency of `builtinsAvailableIn`.

PR #7736, opened 2026-04-23
Use UArray instead of Set for builtin availability check

zeme-wana:
## Summary

Addresses the `TODO` in [`SerialisedScript.hs:215`](https://github.com/IntersectMBO/plutus/blob/d7123ec65dc4d06286c1e7ad857afa49de1bd8c9/plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs#L215) suggesting a better data structure for the builtin availability check in `scriptCBORDecoder`.

Previously `checkBuiltin` did an O(log n) `Set.member` on a `Set DefaultFun`. `DefaultFun` already derives `Ix`, so a `UArray DefaultFun Bool` gives a true O(1) unboxed array index with no `fromEnum`/hash at the lookup site (GHC's `Ix` instance handles the offset).


zliu41:
Unclear this is an improvement. Try `validation-decode` and `validation-full` benchmarks.

Unisay (commented):
Nice cleanup on the TODO.

Worth noting that the lookup is O(1) now, but the array itself gets rebuilt on every `scriptCBORDecoder` call. `(ll, pv)` ranges over a small finite set (four ledger languages times the handful of known protocol versions), so caching the array at module level costs almost nothing. Each decode then does a `Map` lookup instead of `newArray` plus a `mapM_` over the `Set`.

Roughly:

```haskell
availableBuiltinsArr
  :: PlutusLedgerLanguage -> MajorProtocolVersion -> UArray DefaultFun Bool
availableBuiltinsArr =
  let cache = Map.fromList
        [((ll, pv), build ll pv) | ll <- [minBound .. maxBound], pv <- knownPVs]
      build ll pv = runSTUArray $ do
        arr <- newArray (minBound, maxBound) False
        mapM_ (\f -> writeArray arr f True) (builtinsAvailableIn ll pv)
        pure arr
  in \ll pv -> Map.findWithDefault emptyArr (ll, pv) cache
```

`MajorProtocolVersion` doesn't have a meaningful `Bounded` instance, so `knownPVs` has to be listed by hand from [`Common/Versions.hs`](https://github.com/IntersectMBO/plutus/blob/master/plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs): `alonzoPV`, `vasilPV`, `changPV`, `plominPV`, `vanRossemPV`.

Whichever variant lands, it would be nice to see the win show up in `plutus-benchmark:validation` numbers before merging.

zliu41:
I'll defer this to @kwxm

plutus-ledger-api/src/PlutusLedgerApi/Common/SerialisedScript.hs
```diff
@@ -198,11 +199,19 @@ scriptCBORDecoder
   -> CBOR.Decoder s ScriptNamedDeBruijn
 scriptCBORDecoder ll pv =
   -- See Note [New builtins/language versions and protocol versions]
-  let availableBuiltins = builtinsAvailableIn ll pv
+  let available = builtinsAvailableIn ll pv
+
+      availableArr :: UArray DefaultFun Bool
```
kwxm: I think that this is the only place (except maybe in tests) that [builtinsAvailableIn](https://github.com/IntersectMBO/plutus/blob/b6e724e9577419e0cddfaf861ce2d77ea7526d09/plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs#L334) is used, so you could try making that and `builtinsIntroducedIn` return a UArray instead of a set.  That might be more efficient than introducing the set only to get rid of it again later.  For consistency you'd probably need to do something similar for some of the other functions in that function.

kwxm (changes requested):
This looks promising, but I think it could be made more efficient.  Also the decoding benchmark results are a bit strange: on average this seems to speed things up a bit, but if you you at the results for individual scripts it looks as if it gets quicker to decode some scripts and slower to decode others.  Maybe that's just random though: the speedups/slowdowns don't seem to be consistent over different benchmark runs.  It could just be Criterion being weird.

Relatedly, I have an issue to look again at the whole `builtinsAvailableIn` thing.  It may be possbile to compute a lot more of this stuff at compile time, but I'm not totally confident about that.  I did some experiments in #7232 , #7233, and #7241 but then had to go and do other stuff.  I'll take another look some time.

effectfully:
it starting with -10.9% and ending with +10.9% is almost poetic

image

zeme-wana:
Results are inconclusive, closing this.

PR #7739, opened 2026-04-23
Certifier for case-reduce

basetunnel:
This adds 
- A translation relation and decision procedure for case-reduce pass, see `CaseReduce.lagda.md` for documentation on how it is structured.
- Reusable definitions/properties for translation relations in `Untyped.Relation.Binary.{Core,Properties,Structures}`. This mimics the module structure of `agda-stdlib`
- Modular rules for building translation relations (`Untyped.Relation.Binary.Modular`), used by the equivalence relation of case-reduce, see the end of the module for an example how to build decidable relations.


plutus-metatheory/src/Untyped/Relation/Binary/Core.lagda.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+## Pointwise
+
+Variant of Data.List.Relation.Binary.Pointwise that works well for well-scoped
+terms, which have an implicit scope parameter. The polarity annotation helps for
+constructing relations using `Untyped.Relation.Binary.Modular`.
+
+```
+data Pointwise {X} (@++ R : X ⊢ → X ⊢ → Set) : List (X ⊢) → List (X ⊢) → Set where
```
ana-pantilie: Nitpicking: can we do something about these very long lines? It makes it hard to review Agda code in general. I think in the Haskell code we have a 100 length line limit or something of the sort.

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
```diff
@@ -0,0 +1,538 @@
...
+
+
+## Term compatibilty rules
+
+These are typical rules that are part of a translation relation.
+
+```
+data CompatVar (@++ R : Relation) : Relation where
+  `F_ :
+    ∀ {X} (x : Fin X)
```
ana-pantilie, line 105: I find this notation confusing. I'm used to seeing `X` as representing any type (of type `Set` usually). Whereas, `X : ℕ` here, so it's actually just a number! `x` is also just a number, but wrapped by `Fin`.

Would the following notation be better?

```suggestion
    ∀ {n} (nᶠ : Fin n)
```

or

```suggestion
    ∀ {n} (n↑ : Fin n)
```

or whatever you think is better.

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
```diff
@@ -0,0 +1,538 @@
...
+pattern p0 p = inl p
+pattern p1 p = inr (p0 p)
+pattern p2 p = inr (p1 p)
+pattern p3 p = inr (p2 p)
+pattern p4 p = inr (p3 p)
+pattern p5 p = inr (p4 p)
+pattern p6 p = inr (p5 p)
+pattern p7 p = inr (p6 p)
+pattern p8 p = inr (p7 p)
+pattern p9 p = inr (p8 p)
```
ana-pantilie, line 196: Should we have better names for `p0...9`?

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
```diff
@@ -0,0 +1,538 @@
...
+  ; compat-ƛ       = λ RM → inj (compat-lambdaF RM)
+  ; compat-·       = λ RM RN → inj (compat-applyF RM RN)
+  ; compat-force   = λ RM → inj (compat-forceF RM)
+  ; compat-delay   = λ RM → inj (compat-delayF RM)
+  ; compat-constr  = λ RMS → inj (compat-constrF RMS)
+  ; compat-case    = λ RM RMS → inj (compat-caseF RM RMS)
+  ; compat-con     = inj compat-conF
+  ; compat-builtin = inj compat-builtinF
+  ; compat-error   = inj compat-errorF
+  }
```
ana-pantilie, line 228: I don't understand this. In `CompatTerm R ⊆ R`, is `⊆` the regular subset relation? How is it defined on `Relation`?

Also, this is a function which returns a record of type `TermCompatible`. I found it's defined in `Structures`, but I can't find an import for that. I did, see, however that it's reexported in `Untyped.Relation.Binary`, and we import that here. I don't think we should, because it's one level higher. Why not import `Untyped.Relation.Binary.Structures` here directly?

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
```diff
@@ -0,0 +1,538 @@
...
+  DecidableRel R →
+  ∀ {X} (Ms Ns : List (X ⊢)) →
+  Dec (Pointwise R Ms Ns)
+pointwise? R? [] []         = yes []
+pointwise? R? (x ∷ xs) (y ∷ ys)
+  with R? x y ×-dec pointwise? R? xs ys
+... | yes (Rxy , Rxsys) = yes (Rxy ∷ Rxsys)
+... | no ¬R = no λ {(R ∷ Rs ) → ¬R (R , Rs)}
+pointwise? R? (_ ∷ _) []    = no λ ()
+pointwise? R? [] (_ ∷ _)    = no λ ()
```
ana-pantilie: Why is this decision procedure defined here instead of where the `Pointwise` type is defined?

Actually, why are the decision procedures for all of the relations defined in this module? Would it be better to have a Untyped.Relation.Decidable module?

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
```diff
@@ -0,0 +1,538 @@
...
+  +-dec compatLam?
+  +-dec compatApply?
+  +-dec compatForce?
+  +-dec compatDelay?
+  +-dec compatCon?
+  +-dec compatConstr?
+  +-dec compatCase?
+  +-dec compatBuiltin?
+  +-dec compatError?
+  +-dec empty?
```
ana-pantilie, line 389: Awesome!

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
```diff
@@ -0,0 +1,538 @@
...
+  → Refinement? (G R)
+  → Refinement? ((F + G) R)
+(f <|> g) M
+  with f M
+... | just (N , RMN) = just (N , inl RMN)
+... | nothing
+  with g M
+... | just (N , SMN) = just (N , inr SMN)
+... | nothing = nothing
+```
```
ana-pantilie, line 411: Should this be moved elsewhere? I saw that the Refinement type is defined in another module, and `<|>` doesn't seem to be used anywhere here.

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
```diff
@@ -0,0 +1,538 @@
...
+
+
+## Example relation: remove force/delay
```
ana-pantilie, line 414: Awesome that you added an example!!

plutus-metatheory/src/Untyped/Relation/Binary/Properties.lagda.md
```diff
@@ -0,0 +1,163 @@
...
+Symmetric : Relation → Set
+Symmetric _~_ = ∀ {X} {M N : X ⊢} →
+  M ~ N →
+  -------
+  N ~ M
+
+Idempotent : Relation → Set
+Idempotent R = ∀ {X} {L M N : X ⊢} → R L M → R M N → M ≡ N
+
+_⊆_ : Relation → Relation → Set
```
ana-pantilie, line 52: Ah, this is the `⊆` I was asking about earlier!

plutus-metatheory/src/Untyped/Equality.lagda.md
```diff
@@ -139,6 +139,9 @@ instance
   DecEq-ℤ : DecEq ℤ
   DecEq-ℤ ._≟_ = Data.Integer.Properties._≟_
 
+  DecEq-Fin : ∀ {n} → DecEq (Fin n)
+  DecEq-Fin ._≟_ = Data.Fin.Properties._≟_
+
```
ana-pantilie, line 144: Thanks for adding this, I stumbled upon the issue as well.

plutus-metatheory/src/Untyped/Transform.lagda.md
```diff
@@ -0,0 +1,154 @@
...
+
+If `f` refines `R`, then so does `f ↑`.
+
+```
+module Refines
+  (R : Relation)
+  (~-trans : Transitive R)
+  (~-compat : TermCompatible R)
+  (f : Transform)
+  (f-refines : Refines f R)
```
ana-pantilie, line 97: Nice, module parameters!

plutus-metatheory/src/VerifiedCompilation/UApplyToCase.lagda.md
```diff
@@ -21,7 +21,7 @@ module VerifiedCompilation.UApplyToCase where
 open import Untyped using (_⊢; case; constr)
 open import Untyped.Reduction using (iterApp)
 open import VerifiedCompilation.Certificate using (ProofOrCE; ce; proof; ApplyToCaseT)
-open import VerifiedCompilation.UCaseReduce using (justEq)
+open import VerifiedCompilation.UCaseReduce using ()
```
ana-pantilie: Why not just remove the import?

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
```diff
@@ -4,122 +4,428 @@ layout: page
...
+          (case (con (tmCon (list A) (cons x xs))) (N₁ ∷ N₂ ∷ []))
+          (N₁ · con (tmCon A x) · con (tmCon (list A) xs))
+
+  data CaseNil (@++ R : Relation) : Relation where
+    case-nil :
+      ∀ {X} {N₁ N₂ : X ⊢} {A}
+      -----------------------------------------------------------
+      → CaseNil R
+          (case (con (tmCon (list A) nil)) (N₁ ∷ N₂ ∷ []))
+          N₂
```
ana-pantilie, line 133: This looks right, first branch `N₁` for non-empty case and second branch `N₂` for empty. I hope I'm reading https://github.com/IntersectMBO/plutus/blob/7918873ead2ce8cfe8b4bb4d91c1426d32549f90/plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs#L831 right

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
```diff
@@ -4,122 +4,428 @@ layout: page
...
-open import Untyped.CEK using (stepper; step)
-open import Builtin using (Builtin; addInteger; subtractInteger)
 
-ex1 : 0 ⊢
-ex1 = (((ƛ (ƛ (((builtin subtractInteger) · (` zero)) · (` (suc zero)))))) · (con-integer 2)) · (con-integer 3) --- \× . \y . x - y ==>  2 - 3
+The decision procedure is currently not used, since it accepts unwanted compiler
+behaviour, such as case-reducing in the opposite direction.
+
+Completeness requires for each reduction rule in `_~_` a lemma that it is
+admissible in `Computaitonal`. Here is one proof for the `case-constr` rule.
```
ana-pantilie:
```suggestion
admissible in `Computational`. Here is one proof for the `case-constr` rule.
```

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
```diff
@@ -4,122 +4,428 @@ layout: page
...
+
+```
+module Decide
+  {X : ℕ}
+  -- TODO: completeness, by induction on the `_~_` derivation, requires a lemma for each
```
ana-pantilie, line 426: Do you have an issue for this?

ana-pantilie (approved):
This is really cool! My comments only regard structure/formatting/notation, so I can approve this now

plutus-metatheory/src/Untyped/Relation/Binary/Core.lagda.md
basetunnel: We could move the part of the type signature after `:` to the next line. The only thing I dislike from that solution is that the whole block with constructors has to be indented one more level.

basetunnel: Oh this could actually be simplified to
```
data Pointwise {X} (@++ R : Relation) : Relation* where
```

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
basetunnel: Agreed that `X` is awkward. It is a left-over from when the metatheory used nested maybe's for representing scope size. Right now this is consistent with all other uses in the codebase like `X ⊢`. We should change this project-wide.

basetunnel: I actually don't mind `x` personally, since it represents an (arbitrary) variable name in the ast, it just happens to be represented by a natural. But perhaps `v` is also good?

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
basetunnel: Perhaps we should :) Right now this is just boilerplate for the compatibility pattern synonyms. Ideally we would generate them with meta-programming of some form.

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
basetunnel: Right, it's the relation you found later on in your review. `Untyped.Relation.Binary` just re-exports the most useful sub-modules of that hierarchy, including the `Structures` one. So it just saves me a few import lines when I'm using multiple of those modules anyway.

I believe we do this style of re-exporting in the haskell code base too, for convenience.

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
basetunnel: It is an operator specific to relation transformers, so I've opted to keep all of those types/combinators inside the `Modular` module. Refinement is more generally applicable on translation relations and I don't want to pollute namespaces if you are not using the `Modular` approach.

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
basetunnel: Done, https://github.com/IntersectMBO/plutus-private/issues/2181

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
ana-pantilie: Both `x` and `v` are ok, although tying it to the `n` in some way is more helpful to me since it indicates we're using Debruijn

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
ana-pantilie: Having a module which reexports other modules is fine. My issue is with the directory hierarchy. `Modular` is inside the `Binary` directory whereas `Untyped.Relation.Binary` is outside of it.

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
basetunnel: I've opened https://github.com/IntersectMBO/plutus-private/issues/2182, it's probably better to do that in a separate PR.

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
basetunnel: Oh, do you mean that the import looks cyclic? Even if it isn't the case. Yeah, I guess that should change.

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
basetunnel: I've moved `pointwise?` now to `Untyped.Binary.Core`. The decision procedures are here because they are tied to the `Modular` approach, I'd rather have them contained here. We could move them to `Untyped.Relation.Binary.Modular.Decidable` perhaps, but I don't think it's too bad to keep it all in a single module for now.

plutus-metatheory/src/Untyped/Relation/Binary/Core.lagda.md
basetunnel: I've updated it to 
```
data Pointwise {X} (@++ R : Relation) : List (X ⊢) → List (X ⊢) → Set where
```

Because otherwise it would require a lot of implicit quantification/application in the constructors.

plutus-metatheory/src/Untyped/Relation/Binary/Modular.lagda.md
basetunnel: I've changed it to now explicity import the necessary submodules.

plutus-metatheory/src/VerifiedCompilation/UApplyToCase.lagda.md
basetunnel: Done.

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
basetunnel: Yes I think so, the order has cons first because that's what's expected when case on list is given a single branch.

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
```diff
@@ -4,122 +4,428 @@ layout: page
...
+  + CaseNil
+  + CasePair
+```
+
+The equivalence is closed under the reduction rules and compatibility
+rules
+
+```
+_~_ : Relation
+_~_ = Fix (Reduction + CompatTerm + Transitivity + Symmetry + Reflexivity)
```
zliu41, line 169: What's the point of having this relation, and why does it need to be symmetric or reflexive?

zliu41 (commented):
If I understand correctly this is a quite complex way of saying: "If `M` reduces to `M'` via case-reduce, pass, else fail".

As we discussed, this has the disadvantage of tight coupling with the compiler: for example, if the compiler is changed such that it stops reducing `case-constr` in certain cases, then the certifier would break. I thought we've discussed an approach that doesn't have such tight coupling?

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
basetunnel: For the same reason why the force-delay paper has an equivalence: it is (more) obviously semantics preserving, and since the case-reduce pas in Agda is a re-implementation of the Haskell implementation, a mistake there could carry over to Agda. It also becomes easier to count the optimization sites using a relation vs a function.

zliu41: I'd agree with the value of `_~_` if its definition is much simpler and much more obviously correct than `CaseReduce`, but their complexities seem similar.

Also, since `CaseReduce` performs the actual reduction that the compiler does, doesn't it already include compatibility closure, making the `CompatTerm` redundant?

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
```diff
@@ -4,122 +4,428 @@ layout: page
...
+  <|> red-pair
+```
+
+The pass is implemented as a bottom-up traversal that applies the reduction
+rules:
 
 ```
-## An Example:
+reduceM : X ⊢ → Maybe (X ⊢)
+reduceM = refine? (reduce {R = _~_})
```
zliu41: I find it confusing that the definition of `CaseReduce` mentions `_~_`. I thought they are supposed to be two completely separate relations, one is more of an actual reduction algorithm, and the other is a specification that is more obviously correct.

basetunnel:
> If I understand correctly this is a quite complex way of saying: "If `M` reduces to `M'` via case-reduce, pass, else fail".

It also guarantees by construction that it is sound with respect to the inductive reduction rules.

> As we discussed, this has the disadvantage of tight coupling with the compiler: for example, if the compiler is changed such that it stops reducing `case-constr` in certain cases, then the certifier would break. I thought we've discussed an approach that doesn't have such tight coupling?

Yes, I looked at two ways of doing that:

1. reduce both pre- and post-term:

```
_≈_ : Relation
M ≈ M' = case-reduce M ≡ case-reduce M'
```

This is equally easy to decide, and less coupled because it allows case-reduction rules to be applied selectively: the compiler is allowed to not touch some case-constant/case-constructor. However, it also accepts the reverse application of reduction rules, which we deemed outside of the spec of the pass (this relation is sound/complete with the `~` equivalence).

2. a decision procedure for
```
Fix (Reduction + CompatTerm + Transitivity)
```
This would also allow selective application of the reduction rules, but the decision procedure needs to do some reducing itself. Initially that made it hard to construct the proof of the relation, but with the sound-by-construction rules it may now not be that difficult anymore.

Anyway, I'm happy to look into 2 more, but I wanted to get working certifiers for all passes first before generalizing more.

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
basetunnel: Good point, that's not necessary here. `R` just needs to be instantiated to resolve ambiguity for the type checker, it's not actually used.

zliu41 (approved):
This seems overcomplicated for a simple pass like CaseReduce. I feel like some small changes on top of the original approach would be fine.

This approach with two relations - one easier to decide and the other more obviously correct - would be more useful for the inliner, since its current relation is quite complex and not obviously correct.

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
basetunnel: > I'd agree with the value of _~_ if its definition is much simpler and much more obviously correct than CaseReduce, but their complexities seem similar.

You're right that their complexity doesn't differ much. Still I think it's useful to have a more general and declarative relation like `_~_`, since `case-reduce` is just a re-implementation of the compiler logic (potentially having the same bugs etc.)

basetunnel: > Also, since CaseReduce performs the actual reduction that the compiler does, doesn't it already include compatibility closure, making the CompatTerm redundant?

`CompatTerm` is part of `_~_`, not `CaseReduce`, so I don't understand exactly what you're asking.

`CompatTerm` is not redundant in `_~_` because it relates the parts of terms that are not changed by the pass.

plutus-metatheory/src/VerifiedCompilation/UCaseReduce.lagda.md
basetunnel: > Good point, that's not necessary here. `R` just needs to be instantiated to resolve ambiguity for the type checker, it's not actually used.

I was wrong. It's needed because I defined `reduce` to be sound wrt `~` by construction, i.e. it computes a Maybe pair of the reduced term and a proof of its soundness. The reduced term is completely independent of the proof, so reduceM does not depend on `~`, it just additionally shows it is sound.

To make the connection clearer, we could also set it up like this:
```
red-unit : X ⊢ → Maybe (X ⊢)

red-unit-sound : (M M' : X ⊢) → red-unit M ≡ just M' → CaseUnit R M M'
```

But this has to be done for each reduction rule and resulted in more boilerplate code which disappears if you combine them.

basetunnel: I've updated the comments in the code with this explanation

PR #7744, opened 2026-04-24
Clearer parse error for identifiers with a '-' in the middle (#7742)

### LINKED ISSUE #7742: UPLC textual parser (plutus-core 1.45.0.0): fails to parse large (case (constr 0 ...) (lam ...)) when branch body is ~490 lines
## Summary

The megaparsec-based UPLC textual parser in `plutus-core 1.45.0.0` fails to parse a syntactically valid `(case (constr 0 b1..b8) (lam ...))` expression when the single case branch contains a body of approximately 490 lines. The error position reported is misleading (points to the outer `(lam` start) due to megaparsec backtracking.

## Reproduction

**File:** the HTLC scenario from [Unisay/scalus-cape-submissions](https://github.com/Unisay/scalus-cape-submissions), commit `0846510184d98b054c4405c4d036154444467d9f`, file `src/htlc/htlc.uplc`.

- Size: 51 700 bytes
- SHA-256: `abd5609f2da11b37cdb71d8314b34d94012bee1f3b2a3b43793a715228818493`
- Program header: `(program 1.1.0 ...)`

**Trigger structure** (outer program body, lines 1–5 of the file):

```uplc
(program 1.1.0 (case (constr 0
      (force (force (builtin chooseList)))
      (force (force (builtin fstPair))) (force (builtin headList))
      (force (builtin ifThenElse)) (force (builtin mkCons))
      (force (force (builtin sndPair))) (force (builtin tailList))
      (force (builtin trace))) (lam __builtin_ChooseList
    (lam __builtin_FstPair
      ...~490 lines of body...
    ))))
```

**How to reproduce:**

```haskell
import UntypedPlutusCore.Parser (parseProgram)
import Data.Text.IO qualified as T

main :: IO ()
main = do
    src <- T.readFile "htlc.uplc"
    case parseProgram src of
        Left err -> print err
        Right _  -> putStrLn "OK"
```

**Error:**

```
test:5:32:
5 |       (force (builtin trace))) (lam __builtin_ChooseList
  |                                ^
  unexpected '('
  expecting ')'
```

## What was ruled out

- **Paren balance:** 437 opens = 437 closes. File is well-formed.
- **`case`/`constr` support:** parser version 1.45.0.0 does support `case`/`constr` (language version 1.1.0). fibonacci Scalus 0.16.0 UPLC in the same repo — which also uses `(force (case (constr 0 ...) ...))` but with a small body — **parses and verifies fine**.
- **Constants:** removing `(con string …)` and `(con data …)` constants does not change the failure.
- **Structure in isolation:** a minimal `(case (constr 0 b1..b8) (lam x1 (lam x2 ... (lam x8 small-body))))` parses fine.
- **Hidden characters / encoding:** no BOM, no CRLF, plain UTF-8.

## Root cause hypothesis

`caseTerm` in `UntypedPlutusCore/Parser.hs` uses `many term` for branches. When `term` starts parsing the `(lam __builtin_ChooseList ...)` branch and then fails somewhere deep inside the ~490-line body, megaparsec's `many` backtracking causes the error to be reported at the outermost `(` of the failed branch (line 5:32) rather than at the actual failure site. The actual parser error is lost and the reported position is misleading.

The underlying issue may be one of:
1. A missing `try` combinator on one of the sub-parsers inside `term`, causing megaparsec to incorrectly drop progress and backtrack to the `many term` boundary.
2. An error inside the branch body that is not recoverable — the parser cannot finish parsing `(lam __builtin_ChooseList BODY)` due to some construct inside BODY — but the reported location is wrong.

## Impact

Any UPLC file generated by Scalus 0.16.0's `toUplcOptimized()` for contracts with 8+ forced builtins (e.g. HTLC, other Cardano V3 validators) will fail to parse under `plutus-core ^>=1.45`. The workaround is to use `toUplc()` (no CaseConstrApply optimization), which produces larger but parseable UPLC.

This blocks UPLC-CAPE benchmark submissions for complex Scalus 0.16.0 contracts.

## Environment

- `plutus-core`: 1.45.0.0
- Parser: `UntypedPlutusCore.Parser.parseProgram` (megaparsec-based)
- UPLC language version: `1.1.0` (Plutus V3 / Conway)
- Confirmed via `cabal repl` in UPLC-CAPE repo

Unisay:
## Summary

Closes #7742.

A UPLC file emitted by Scalus 0.16.0's `toUplcOptimized` — the HTLC sample from the issue — fails to parse with a confusing message hundreds of lines from the real site. The shape reduces to one bad identifier:

```
(program 1.1.0 (lam pubKeyHash-305478r71 (lam xIn72 xIn72)))
```

`-` isn't an identifier character in UPLC (see `isIdentifierChar` in `PlutusCore.Name.Unique`); it exists only as the separator between the name text and its numeric unique-suffix (`foo-123` → `Name "foo" (Unique 123)`). So `pubKeyHash-305478r71` isn't a valid unquoted identifier. The old parser still "succeeded" on it in the worst possible way: ate `pubKeyHash` as the name, `-305478` as the unique, left `r71` in the stream. `r71` then got picked up as the body of the enclosing `(lam ...)`, leaving the closing-paren bookkeeping off by one, and megaparsec eventually blew up on an unrelated `(` with "unexpected '(' expecting ')'" at line 448 col 39 — nowhere near the bad name on line 447.

This PR keeps the grammar as-is (it should stay narrow — see the discussion in the issue comments) but makes the diagnostic point at the real problem.

## Before

```
test:1:42:
  |
1 | (program 1.1.0 (lam pubKeyHash-305478r71 (lam x x)))
  |                                          ^
unexpected '('
expecting closing parenthesis ')'
```

## After

```
test:1:32:
  |
1 | (program 1.1.0 (lam pubKeyHash-305478r71 (lam x x)))
  |                                ^
Malformed unique suffix '305478r71' for name 'pubKeyHash' at test:1:32.
A unique suffix must be a non-empty sequence of digits (e.g. '-123').
To use this text as a name verbatim, quote it with backticks: `pubKeyHash-305478r71`
```

### DIFF
--- a/plutus-core/changelog.d/20260424_182803_yuriy.lazaryev_issue_7742_uplc_parser_large_case.md
+++ b/plutus-core/changelog.d/20260424_182803_yuriy.lazaryev_issue_7742_uplc_parser_large_case.md
@@ -1,12 +1,7 @@
 ### Changed
 
-- The UPLC/PLC/PIR textual parser now rejects unquoted identifiers that
-  contain a `-` anywhere other than as the terminal numeric unique-suffix
-  separator (e.g. `pubKeyHash-305478r71`, `foo-bar`, `foo-123-456`) with
-  a dedicated `InvalidIdentifier` diagnostic that points directly at the
-  offending name and shows the full bad text. Previously the same inputs
-  silently mis-parsed — the prefix was taken as a name plus unique-suffix
-  and the remainder was picked up as an adjacent term — which surfaced as
-  a confusing "unexpected '(' expecting ')'" message far from the real
-  site (see #7742). To use such a string as a name verbatim, wrap it in
-  backticks: `` `pubKeyHash-305478r71` ``.
+- The UPLC/PLC/PIR textual parser now rejects an unquoted name whose unique
+  suffix (the region after `-`) is not a non-empty sequence of digits,
+  raising a dedicated `MalformedUniqueSuffix` diagnostic pointing at the
+  `-` and showing the offending suffix text (#7742). Wrap such a string
+  in backticks to use it as a name verbatim: `` `pubKeyHash-305478r71` ``.
--- a/plutus-core/changelog.d/20260424_182803_yuriy.lazaryev_issue_7742_uplc_parser_large_case.md
+++ b/plutus-core/changelog.d/20260424_182803_yuriy.lazaryev_issue_7742_uplc_parser_large_case.md
@@ -1,0 +1,12 @@
+### Changed
+
+- The UPLC/PLC/PIR textual parser now rejects unquoted identifiers that
+  contain a `-` anywhere other than as the terminal numeric unique-suffix
+  separator (e.g. `pubKeyHash-305478r71`, `foo-bar`, `foo-123-456`) with
+  a dedicated `InvalidIdentifier` diagnostic that points directly at the
+  offending name and shows the full bad text. Previously the same inputs
+  silently mis-parsed — the prefix was taken as a name plus unique-suffix
+  and the remainder was picked up as an adjacent term — which surfaced as
+  a confusing "unexpected '(' expecting ')'" message far from the real
+  site (see #7742). To use such a string as a name verbatim, wrap it in
+  backticks: `` `pubKeyHash-305478r71` ``.
--- a/plutus-core/plutus-core/src/PlutusCore/Error.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Error.hs
@@ -2,13 +2,11 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+-- appears in the generated instances:
 {-# OPTIONS_GHC -fno-warn-orphans #-}
-
--- appears in the generated instances
 
 module PlutusCore.Error
   ( ParserError (..)
@@ -52,12 +50,7 @@ data ParserError
   = BuiltinTypeNotAStar !T.Text !SourcePos
   | UnknownBuiltinFunction !T.Text !SourcePos ![T.Text]
   | InvalidBuiltinConstant !T.Text !T.Text !SourcePos
-  | {-| An unquoted identifier that violates the grammar: a '-' appeared
-    anywhere other than as the separator of a terminal numeric unique-suffix
-    (e.g. @pubKeyHash-305478r71@, @foo-bar@, @foo-123-456@). The 'Text'
-    carries the full offending text as it appeared in the source, so the
-    user sees their own name back in the diagnostic. -}
-    InvalidIdentifier !T.Text !SourcePos
+  | MalformedUniqueSuffix !T.Text !T.Text !SourcePos
   deriving stock (Eq, Ord, Generic)
   deriving anyclass (NFData)
 
@@ -198,18 +191,22 @@ instance Pretty ParserError where
       <+> squotes (pretty s)
       <+> "at"
       <+> pretty loc
-  pretty (InvalidIdentifier txt loc) =
-    "Invalid identifier"
-      <+> squotes (pretty txt)
+  pretty (MalformedUniqueSuffix base suffix loc) =
+    "Malformed unique suffix"
+      <+> squotes (pretty suffix)
+      <+> "for name"
+      <+> squotes (pretty base)
       <+> "at"
       <+> pretty loc
       <> "."
       <> hardline
-      <> "A '-' inside a name is the numeric unique-suffix separator and must be"
-      <+> "followed only by digits and a word boundary."
+      <> "A unique suffix must be a non-empty sequence of digits"
+      <+> "(e.g."
+      <+> squotes "-123"
+      <> ")."
       <> hardline
       <> "To use this text as a name verbatim, quote it with backticks:"
-      <+> pretty ("`" <> txt <> "`")
+      <+> pretty ("`" <> base <> "-" <> suffix <> "`")
 
 instance ShowErrorComponent ParserError where
   showErrorComponent = show . pretty
--- a/plutus-core/plutus-core/src/PlutusCore/Error.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Error.hs
@@ -52,6 +52,12 @@ data ParserError
   = BuiltinTypeNotAStar !T.Text !SourcePos
   | UnknownBuiltinFunction !T.Text !SourcePos ![T.Text]
   | InvalidBuiltinConstant !T.Text !T.Text !SourcePos
+  | {-| An unquoted identifier that violates the grammar: a '-' appeared
+    anywhere other than as the separator of a terminal numeric unique-suffix
+    (e.g. @pubKeyHash-305478r71@, @foo-bar@, @foo-123-456@). The 'Text'
+    carries the full offending text as it appeared in the source, so the
+    user sees their own name back in the diagnostic. -}
+    InvalidIdentifier !T.Text !SourcePos
   deriving stock (Eq, Ord, Generic)
   deriving anyclass (NFData)
 
@@ -192,6 +198,18 @@ instance Pretty ParserError where
       <+> squotes (pretty s)
       <+> "at"
       <+> pretty loc
+  pretty (InvalidIdentifier txt loc) =
+    "Invalid identifier"
+      <+> squotes (pretty txt)
+      <+> "at"
+      <+> pretty loc
+      <> "."
+      <> hardline
+      <> "A '-' inside a name is the numeric unique-suffix separator and must be"
+      <+> "followed only by digits and a word boundary."
+      <> hardline
+      <> "To use this text as a name verbatim, quote it with backticks:"
+      <+> pretty ("`" <> txt <> "`")
 
 instance ShowErrorComponent ParserError where
   showErrorComponent = show . pretty
--- a/plutus-core/untyped-plutus-core/testlib/Generators/Spec.hs
+++ b/plutus-core/untyped-plutus-core/testlib/Generators/Spec.hs
@@ -60,6 +60,9 @@ test_parsing =
         , propMissingConOperands
         , propInvalidKeyword
         , propBracketMismatch
+        , propInvalidIdentifierHyphenLetters
+        , propInvalidIdentifierHyphenWord
+        , propInvalidIdentifierDoubleUnique
         ]
     ]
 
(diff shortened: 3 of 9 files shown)


Unisay:
The single Hydra failure (`ci/hydra-build:x86_64-linux.ghc96-mingsW64:checks:plutus-core:test:flat-test`) is unrelated to this change. Log says:

> stable byte encodings: FAIL
> Exception: diff: startProcess: does not exist (No such file or directory)

It's the Windows-cross builder missing the `diff` binary that `tasty-golden` shells out to. The same check failed on the recently-merged #7734 with the same message. All native x86_64-linux / aarch64-darwin / x86_64-darwin Hydra checks pass, and locally `./scripts/regen-goldens.sh` (~12 800 tests, 7 subprojects) produces zero golden updates outside the three this PR adds.

plutus-core/untyped-plutus-core/testlib/Generators/Spec.hs
```diff
@@ -241,6 +244,49 @@ propBracketMismatch =
...
+and today the parser mis-parses them in a way that surfaces as a confusing
+error hundreds of lines away from the offending name — see issue #7742.
+
+The goldens below freeze the *current* (unhelpful) error output so that a
+future diagnostic improvement shows up as an explicit golden-file diff.
+When the parser is taught to point at the bad name itself, accept the new
+goldens with 'scripts/regen-goldens.sh' (or '--accept'). -}
+
+{-| @pubKeyHash-305478r71@ — the exact shape Scalus 0.16.0 produces, inside a
+binder. Current behaviour: the parser eats @pubKeyHash-305478@ as name+unique,
```
zliu41: None of these comments is necessary. Also, once this is merged, "current behavior" will no longer be "current", and that will be quite confusing.

plutus-core/untyped-plutus-core/testlib/Generators/Spec.hs
```diff
@@ -241,6 +244,49 @@ propBracketMismatch =
...
+picks up @r71@ as the lam body, then fails far away on the next paren. -}
+propInvalidIdentifierHyphenLetters :: TestTree
+propInvalidIdentifierHyphenLetters =
```
zliu41: I don't see any value of this test case given the previous one.

plutus-core/untyped-plutus-core/testlib/Generators/Spec.hs
```diff
@@ -241,6 +244,49 @@ propBracketMismatch =
...
+  testParseErrorGolden
+    "Invalid identifier: hyphen followed by digits then letters"
+    "invalid-identifier-hyphen-letters"
+    "(program 1.1.0 (lam pubKeyHash-305478r71 (lam x x)))"
+
+{-| @foo-bar@ — hyphen followed by non-digits. Current behaviour: the parser
+stops at '-' (it is not in 'isIdentifierChar'), takes @foo@ as the name, and
+then explodes on @-bar@ which is not a valid continuation anywhere. -}
+propInvalidIdentifierHyphenWord :: TestTree
+propInvalidIdentifierHyphenWord =
```
zliu41: I don't see any value of this test case given the previous one. Also, you need to test some valid identifiers, such as `foo-123`.

plutus-core/plutus-core/src/PlutusCore/Error.hs
```diff
@@ -52,6 +52,12 @@ data ParserError
   = BuiltinTypeNotAStar !T.Text !SourcePos
   | UnknownBuiltinFunction !T.Text !SourcePos ![T.Text]
   | InvalidBuiltinConstant !T.Text !T.Text !SourcePos
+  | {-| An unquoted identifier that violates the grammar: a '-' appeared
+    anywhere other than as the separator of a terminal numeric unique-suffix
+    (e.g. @pubKeyHash-305478r71@, @foo-bar@, @foo-123-456@). The 'Text'
+    carries the full offending text as it appeared in the source, so the
+    user sees their own name back in the diagnostic. -}
+    InvalidIdentifier !T.Text !SourcePos
```
zliu41: What's the current error thrown if there's an invalid character (like `^` or whatever) in the identifier? If you are adding an `InvalidIdentifier` error, it shouldn't just apply to `-`.

plutus-core/untyped-plutus-core/testlib/Generators/Spec.hs
```diff
@@ -241,6 +244,49 @@ propBracketMismatch =
...
+not allowed by the unquoted grammar (see 'isIdentifierChar' in
+'PlutusCore.Name.Unique'). Several tools in the wild (e.g. Scalus 0.16.0's
+'toUplcOptimized') emit names like 'pubKeyHash-305478r71' that violate this,
+and today the parser mis-parses them in a way that surfaces as a confusing
+error hundreds of lines away from the offending name — see issue #7742.
+
+The goldens below freeze the *current* (unhelpful) error output so that a
+future diagnostic improvement shows up as an explicit golden-file diff.
+When the parser is taught to point at the bad name itself, accept the new
+goldens with 'scripts/regen-goldens.sh' (or '--accept'). -}
```
zliu41: This is over documentation - way too verbose. Please avoid using large blocks of AI-generated text - review and revise so it’s concise and to the point.

This can be summarized in one sentence: "`-` is allowed in unquoted names only if there's a single `-`, and it is followed by numbers".

plutus-core/changelog.d/20260424_182803_yuriy.lazaryev_issue_7742_uplc_parser_large_case.md
```diff
@@ -0,0 +1,12 @@
...
+- The UPLC/PLC/PIR textual parser now rejects unquoted identifiers that
+  contain a `-` anywhere other than as the terminal numeric unique-suffix
+  separator (e.g. `pubKeyHash-305478r71`, `foo-bar`, `foo-123-456`) with
+  a dedicated `InvalidIdentifier` diagnostic that points directly at the
+  offending name and shows the full bad text. Previously the same inputs
+  silently mis-parsed — the prefix was taken as a name plus unique-suffix
+  and the remainder was picked up as an adjacent term — which surfaced as
+  a confusing "unexpected '(' expecting ')'" message far from the real
+  site (see #7742). To use such a string as a name verbatim, wrap it in
+  backticks: `` `pubKeyHash-305478r71` ``.
```
zliu41: Please shorten

zliu41 (changes requested):

plutus-core/plutus-core/src/PlutusCore/Error.hs
Unisay: This is the crux: there's an asymmetry in how identifier-related errors are handled.

1. Characters that are invalid for an identifier (`not isIdentifierChar` and not `-`), like `^` or `&`: the parser stops at the valid prefix, returns it as a `Name`, and the bad character produces a generic Megaparsec "unexpected" error at its own position. So `foo^bar` reports `^` at column N, with no link back to the identifier the user actually meant.
2. Characters that belong to the identifier-char set, or another `-` (i.e. `isNameExtensionChar = isIdentifierChar || '-'`): when they appear right after a parsed `name + optional -NNN` and would extend it, the parser has already over-committed. The `-NNN` it consumed as a unique-suffix was actually part of the intended name. It eats the remainder and raises `InvalidIdentifier` with the full offending text.

The new error only fires for case 2, but the constructor name `InvalidIdentifier` is misleading as it sounds to cover both cases. It's really the hyphen-misparse case, not "any invalid identifier".

That said I see two possible ways forward:
1. [Local solution] Keep this error specific to the second case and rename it accordingly, i.e. `MalformedIdentifierExtension`
2. [Global solution] Rework the way identifiers are parsed to emit identifier-related errors in a uniform, consistent way.

I chose local fix: its an improvement that doesn't stop us from reworking the whole identifier parser later if we want (much bigger scope)

plutus-core/untyped-plutus-core/testlib/Generators/Spec.hs
Unisay: Replaced with a single property test for all possible invalid identifier suffixes.

plutus-core/untyped-plutus-core/testlib/Generators/Spec.hs
Unisay: Replaced with a single property test for all possible invalid identifier suffixes.

plutus-core/untyped-plutus-core/testlib/Generators/Spec.hs
Unisay: Removed.

plutus-core/plutus-core/src/PlutusCore/Error.hs
Unisay: In the end I have re-implemented the middle-ground solution:
I didn't rework the way whole identifiers are parsed, only touched the unique suffix part `-123`
The change turned out not that big and we have a more consistent error handling for suffixes now.

zliu41: > This is the crux: there's an asymmetry in how identifier-related errors are handled.

Do you mean this is how they are handled currently on master, or how you are handling it in your PR?

> I didn't rework the way whole identifiers are parsed, only touched the unique suffix part -123

I don't know what this means. "Touched" in what sense?

And to step back a bit - why do we need to allow parsing both names-with-unique and names-without-unique? In any given textual UPLC, don't all names either have uniques or don't?

Btw, as a general rule, avoid posting Claude answers verbatim. It comes across to me as not being respectful of other people's time. They are often quite verbose and of questionable quality - inaccuracy is common. You are of course free to use Claude, but make sure to quality-check, shorten and rephrase, make it suitable for human conversation before asking others to read it. This applies to PR descriptions, code reviews, and other communication.

Unisay: > Do you mean this is how they are handled currently on master

yes, I meant master.

> I don't know what this means. "Touched" in what sense?

by "touched" I meant "changed" the suffix (numeric extension) parser.

> And to step back a bit - why do we need to allow parsing both names-with-unique and names-without-unique? 

Idk, this decision was made before me.

> In any given textual UPLC, don't all names either have uniques or don't?

I agree that most likely its either all with suffixes or all without. Depends on the machinery that produced the UPLC.

> Btw, as a general rule, avoid posting Claude answers verbatim. It comes across to me as not being respectful of other people's time. 

So far when producing public texts (GH descriptions, comments, etc.) I tried to come close to how I want to receive them (Golden rule: treat others same way you want them to treat you). 

I didn't mean to disrespect anyone, and since you criticized the original Claude produced PR description - I simplified it significantly and now its in the shape in which I'd want others to create PRs for my review. I think it also serves your request to provide the Before and After examples.

Other than the original PR description - I typed all comments on keyboard myself without asking LLM to write them. 

Anyway, I got your point: 

> They are often quite verbose and of questionable quality - inaccuracy is common. You are of course free to use Claude, but make sure to quality-check, shorten and rephrase, make it suitable for human conversation before asking others to read it. This applies to PR descriptions, code reviews, and other communication.

PR #7748, opened 2026-04-29
Certifier: bundle README.md with generated Agda certificate

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/2185


zliu41 (commented):
All these backslashes makes this quite hard to maintain. This is mostly static text, the only dynamic part is `name`, but I don't think you really need it to be dynamic. You can just make a static readme and load it with `file-embed.`

zliu41:
Even if you really want the dynamic `name`, you can easily do a `Text.replace`.

zliu41 (commented):
I'm not sure if `data-files` works. @zeme-wana What do you think? You mentioned this before about `data-files`:

> They do in CI. But once the executable is built and deployed to some machine, the files are nowhere to be found

In any case, this is exactly what `file-embed` is designed to do, and that definitely works.

zliu41 (changes requested):
Yeah, since we release `uplc` as a single executable, this can't work. You need to use `file-embed`.

ana-pantilie:
@zliu41 I'm now getting https://ci.iog.io/build/12767781/nixlog/18 which I'm unable to reproduce locally.

@zeme-wana said that there might be platform specific issues with this approach, so I'm going to just use the original version, which was to embed the file into a Haskell string. It's much simpler, the file isn't that large, we won't have to change it too often and I think we've already spent too much time on this.

zliu41:
> platform specific issues

What platform issues - can you elaborate? We use TH to load the cost model JSON files and there's no problem with that.

ana-pantilie:
@zliu41 The message I got from @zeme-wana regarding platform specific issues is that it might break CI on Windows. I did not get to see if this error would happen or not, because I got a different error (https://ci.iog.io/build/12767781/nixlog/18) which I also mentioned in the previous message. I wasn't able to reproduce this error locally, and it seemed to me like, given the issue and other potential issues, we should opt for the simplest solution.

zliu41:
> I did not get to see if this error would happen or not, because I got a different error

Your error is because you didn't add the md file to `extra-source-files`.

> it might break CI on Windows

@zeme-wana this is no longer a problem, which is why we decided to keep `readJSONFromFile`, right? And `file-embed` is also much less likely to cause the same problem than `readJSONFromFile`.

zeme-wana:
Yes, @zliu41 is right about the error. Worth trying with `extra-source-files` first. It should work, even when xcompiling on windows.

ana-pantilie:
> Your error is because you didn't add the md file to extra-source-files.

I completely forgot about that, sorry 🤦‍♀️ 

> Yes, @zliu41 is right about the error. Worth trying with extra-source-files first. It should work, even when xcompiling on windows.

👍

zliu41 (approved):

PR #7749, opened 2026-04-29
Refactor AgdaUnparse to use prettyprinter

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/1224

### DIFF
--- a/plutus-metatheory/src/FFI/AgdaUnparse.hs
+++ b/plutus-metatheory/src/FFI/AgdaUnparse.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# OPTIONS_GHC -Wall #-}
 
 module FFI.AgdaUnparse where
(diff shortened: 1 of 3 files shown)


plutus-metatheory/src/FFI/AgdaUnparse.hs
```diff
@@ -155,59 +158,54 @@ instance AgdaUnparse (UPLC.DefaultUni (PLC.Esc a)) where
...
-                   agdaUnparse val
-               )
-        PLC.ValueOf (PLC.DefaultUniApply _ _) _ ->
-          error "Application of an unknown type is not supported."
-      ++ ")"
+    parens $
+      "tagCon"
+        <+> case valOf of
+          PLC.ValueOf PLC.DefaultUniInteger val ->
+            "integer" <+> agdaUnparse val
```
basetunnel: Perhaps `agdaUnparse` can also be used to print the type tags, just like in the case of list, pair etc? There seems to be some duplication now.

basetunnel (approved):

plutus-metatheory/src/FFI/AgdaUnparse.hs
ana-pantilie: GHC has trouble typechecking it because it doesn't know if there's an `AgdaUnparse` instance for the `DefaultUni` constructors, however I did use `bring` in some of the cases, and it seems I could actually use it at higher level (without having to go down into the `case` at all) as well!

PR #934, opened 2019-04-23
all: remove sizes

michaelpj:
This does the job of removing sizes from PLC and the knock-on effects.
It doesn't *completely* expunge the concept, since there are many pieces
of code that are now redundant, or poorly named, or poorly commented.
However, I thought it was best to get something working in first and
then do some cleanup later.

There are a couple of generators commented out (in `Language.PlutusCore.Generators.Interesting`), since weirdly they seem to loop now. My debugging has failed to work out why, so putting this up in case some more eyes (especially @effectfully). I'll keep trying to figure out what's going on.

Ideally I'd also like to include the spec changes, but I'll leave that for a little since there's some ongoing activity.

PR #7752, opened 2026-04-30
Remove redundant recursive alias in genTerm

zeme-wana:
## Summary

In `genTerm`'s inner `go` helper (`PlutusCore/Generators/Hedgehog/Entity.hs`), the `recursive` where-binding was byte-for-byte identical to `variables`:

```haskell
variables = map proceed $ lookupInContext tr context
recursive = map proceed $ lookupInContext tr context
```

These were not always aliases. Back when PLC had a size system, `variables` looked up the type as written while `recursive` looked it up with sizes erased, so two different keys could yield two different result sets — `variables ++ recursive` was actually mixing them.

PR #934 (0ed93972a6) ripped sizes out of the language. The size-erasure step turned into a no-op, both bindings collapsed onto the same lookup, and a FIXME was left behind instead of a real fix. The duplication then rode along through later refactors (2064f49cea, 65b2c34d5a, e9c420da4d) untouched.

So today `variables ++ recursive` is just `xs ++ xs`, and the `Gen.frequency` calls were silently double-weighting the same set. This PR removes the duplication: one binding (`variables`), referenced once in each guard.

If anyone wants the doubled weight back at depth 2, a single `map (8,) variables` would say so honestly — but that doubling was an accident, not a design choice, so this PR drops it.

## Test plan
- [ ] `cabal build plutus-core:test:plutus-core-test`
- [ ] `cabal test plutus-core:test:plutus-core-test`

### DIFF
--- a/plutus-core/testlib/PlutusCore/Generators/Hedgehog/Entity.hs
+++ b/plutus-core/testlib/PlutusCore/Generators/Hedgehog/Entity.hs
@@ -195,11 +195,10 @@ genTerm genBase context0 depth0 = Morph.hoist runQuoteT . go context0 depth0
       -> TypeRep r
       -> GenT (QuoteT m) (TermOf (Plain Term uni fun) r)
     go context depth tr
-      -- FIXME: should be using 'variables' but this is now the same as 'recursive'
       | depth == 0 = choiceDef (liftT $ genBase tr) []
-      | depth == 1 = choiceDef (liftT $ genBase tr) $ variables ++ recursive
-      | depth == 2 = Gen.frequency $ stopOrDeeper ++ map (3,) variables ++ map (5,) recursive
-      | depth == 3 = Gen.frequency $ stopOrDeeper ++ map (3,) recursive
+      | depth == 1 = choiceDef (liftT $ genBase tr) variables
+      | depth == 2 = Gen.frequency $ stopOrDeeper ++ map (5,) variables
+      | depth == 3 = Gen.frequency $ stopOrDeeper ++ map (3,) variables
       | otherwise = Gen.frequency stopOrDeeper
       where
         stopOrDeeper = [(1, liftT $ genBase tr), (5, lambdaApply)]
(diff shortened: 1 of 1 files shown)


ana-pantilie (approved):

plutus-core/testlib/PlutusCore/Generators/Hedgehog/Entity.hs
```diff
@@ -195,11 +195,10 @@ genTerm genBase context0 depth0 = Morph.hoist runQuoteT . go context0 depth0
       -> TypeRep r
       -> GenT (QuoteT m) (TermOf (Plain Term uni fun) r)
     go context depth tr
-      -- FIXME: should be using 'variables' but this is now the same as 'recursive'
       | depth == 0 = choiceDef (liftT $ genBase tr) []
-      | depth == 1 = choiceDef (liftT $ genBase tr) $ variables ++ recursive
-      | depth == 2 = Gen.frequency $ stopOrDeeper ++ map (3,) variables ++ map (5,) recursive
-      | depth == 3 = Gen.frequency $ stopOrDeeper ++ map (3,) recursive
+      | depth == 1 = choiceDef (liftT $ genBase tr) $ variables ++ variables
```
Unisay: `variables` and `recursive` weren't always aliases. Back when PLC had a size system, `variables` looked up the type as written, while `recursive` looked it up with sizes erased, so a generator for `integer 2` could reuse entries typed `integer s`. Two different keys, two different result sets, and `++` was actually mixing them.

Then PR #934 (`0ed93972a6`) ripped sizes out of the language. The size-erasure step turned into a no-op, both bindings collapsed onto the same lookup, and the author left a FIXME instead of fixing it. The duplication rode along through later refactors: `2064f49cea` (static builtins gone), `65b2c34d5a` (TypeRep rewrite), `e9c420da4d` (fourmolu). Nobody touched it.

So today `variables ++ recursive` is just `xs ++ xs`. It doesn't add diversity, it only doubles the weight of variable-based generators in the `Gen.frequency` calls. If that weighting were actually wanted, one binding with a bigger frequency number would say so honestly.

So please remove the unnecessary duplication.

Unisay (changes requested):

PR #7756, opened 2026-04-30
Add json for semvar D and E plus sync up related code

zliu41:


plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs
```diff
@@ -712,6 +713,21 @@ instance
   readKnown = readKnownCoerce @Integer
   {-# INLINE readKnown #-}
 
+deriving newtype instance
+  KnownTypeAst tyname DefaultUni TextCostingByteLength
```
kwxm: The name `TextCostingByteLength` is a little inconsistent with things like `IntegerCostedLiterally`.  `TextCostedByByteLength` might be more consistent, but that's just nitpicking.

kwxm:
The `print-cost-model` executable should know about variants D and E: just add them [here](https://github.com/IntersectMBO/plutus/blob/73e8f007a8eba2415e89aec41bbd9dbc57968bb6/plutus-core/cost-model/print-cost-model/Main.hs#L239).

plutus-core/plutus-core/src/PlutusCore/DataFilePaths.hs
```diff
@@ -36,10 +36,10 @@ cekMachineCostsFileC :: FilePath
...
 cekMachineCostsFileD :: FilePath
-cekMachineCostsFileD = costModelDataDir </> "cekMachineCostsB" <.> "json"
+cekMachineCostsFileD = costModelDataDir </> "cekMachineCostsD" <.> "json"
 
 cekMachineCostsFileE :: FilePath
-cekMachineCostsFileE = costModelDataDir </> "cekMachineCostsC" <.> "json"
+cekMachineCostsFileE = costModelDataDir </> "cekMachineCostsE" <.> "json"
 
 latestMachineCostsFile :: FilePath
 latestMachineCostsFile = cekMachineCostsFileC
```
kwxm, line 45: This should be probably `cekMachineCostsFileE` now (line 45).  The files are identical so it won't actually change anything.

kwxm (approved):
Looks good to go, modulo one or two small comments.

plutus-core/plutus-core/src/PlutusCore/DataFilePaths.hs
zliu41: Will change this in a separate PR.

kwxm: @zliu41 You'll also need to update `latestBuiltinCostModelFile` [here](https://github.com/IntersectMBO/plutus/blob/f92b7d7d82622a26caf456a6be33859f697e2cfc/plutus-core/plutus-core/src/PlutusCore/DataFilePaths.hs#L26).

PR #7764, opened 2026-05-04
Add a constant folding pass to UPLC

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/2172


plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Contexts.hs
```diff
@@ -36,3 +39,64 @@ fillAppCtx term = \case
...
+      AppCtxType _ ctx -> go (acc + 1) ctx
+
+appendAppCtx
+  :: AppCtx name uni fun a
+  -> AppCtx name uni fun a
+  -> AppCtx name uni fun a
+appendAppCtx ctx1 ctx2 = go ctx1
+  where
+    go AppCtxEnd = ctx2
+    go (AppCtxTerm arg ann ctx') = AppCtxTerm arg ann $ go ctx'
```
zliu41: AppCtxTerm's first argument is `ann`

plutus-core/plutus-core/src/PlutusCore/Compiler.hs
```diff
@@ -1,40 +1,41 @@
...
      )
   => Term tyname name uni fun a
   -> m (UPLC.Term name uni fun a)
 compileTerm t = do
-  optimizeOpts <- view coOptimizeOpts
-  builtinSemanticsVariant <- view coBuiltinSemanticsVariant
+  builtinSemanticsVariant <- view (ooBuiltinsInfo . biSemanticsVariant)
   let erased = eraseTerm t
   renamed <- rename erased
+  optimizeOpts <- view id
```
zliu41:
```suggestion
  optimizeOpts <- ask
```

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Optimize.hs
```diff
@@ -108,6 +109,7 @@ termOptimizer opts builtinSemanticsVariant =
         >=> runStage ForceDelayStage
         >=> runStage CaseOfCaseStage
         >=> runStage CaseReduceStage
+        >=> runStage ConstantFoldingStage
```
zliu41: This should run after inlining.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Contexts.hs
```diff
@@ -36,3 +39,64 @@ fillAppCtx term = \case
   AppCtxEnd -> term
   AppCtxTerm ann arg ctx -> fillAppCtx (Apply ann term arg) ctx
   AppCtxType ann ctx -> fillAppCtx (Force ann term) ctx
+
+dropAppCtx :: Int -> AppCtx name uni fun a -> AppCtx name uni fun a
```
zliu41: Doesn't seem to be used anywhere?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Contexts.hs
```diff
@@ -23,7 +26,7 @@ splitAppCtx :: Term nam uni fun a -> (Term nam uni fun a, AppCtx nam uni fun a)
 splitAppCtx = go AppCtxEnd
   where
     go appCtx = \case
-      Apply ann function argument -> go (AppCtxTerm ann argument appCtx) function
+      Apply ann function argument' -> go (AppCtxTerm ann argument' appCtx) function
```
zliu41: Why?

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/EvaluateBuiltins.hs
```diff
@@ -0,0 +1,87 @@
...
+      -> AppCtx name uni fun a
+      -> Maybe (Term name uni fun ())
+    eval (BuiltinCostedResult _ getFXs) AppCtxEnd =
+      case getFXs of
+        BuiltinSuccess y -> Just y
+        -- Evaluates successfully, but does logging. If we're being conservative
+        -- then we should leave these in, so we don't remove people's logging!
+        -- Otherwise `trace "hello" x` is a prime candidate for evaluation!
+        BuiltinSuccessWithLogs _ y -> if preserveLogging then Nothing else Just y
+        -- Evaluation failure. This can mean that the evaluation legitimately
```
zliu41: Rather than duplicating these comments, it's better to just add a comment on the top, says this mirrors the PIR pass.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Analysis/Builtins.hs
```diff
@@ -0,0 +1,109 @@
...
+
+-- See Note [Unserializable constants]
+defaultUniUnserializableConstants :: Some (ValueOf DefaultUni) -> Bool
+defaultUniUnserializableConstants = \case
+  Some (ValueOf DefaultUniBLS12_381_G1_Element _) -> True
+  Some (ValueOf DefaultUniBLS12_381_G2_Element _) -> True
+  Some (ValueOf DefaultUniBLS12_381_MlResult _) -> True
+  _ -> False
+
+{- Note [Unserializable constants]
```
zliu41: There's no need to duplicate this entire note. If it's different from the one in PIR, just point out the dfiference.

plutus-core/plutus-core/src/PlutusCore/Compiler.hs
```diff
@@ -44,15 +45,15 @@ compileProgram (Program a v t) = UPLC.Program a v <$> compileTerm t
...
      )
   => Program tyname name uni fun a
   -> m (UPLC.Program name uni fun a, UPLC.OptimizerTrace name uni fun a)
 compileProgramWithTrace (Program a v t) = do
-  optimizeOpts <- view coOptimizeOpts
-  builtinSemanticsVariant <- view coBuiltinSemanticsVariant
+  builtinSemanticsVariant <- view (ooBuiltinsInfo . biSemanticsVariant)
   let erased = eraseTerm t
   renamedProgram <- UPLC.Program a v <$> rename erased
+  optimizeOpts <- view id
```
zliu41:
```suggestion
  optimizeOpts <- ask
```

plutus-tx/src/PlutusTx/Lift.hs
```diff
@@ -69,28 +63,23 @@ import Prelude as Haskell
 {-| Get a Plutus Core term corresponding to the given value. Allows configuring
 PIR and UPLC optimization options. -}
 safeLiftWith
-  :: forall a uni fun m
-   . ( Lift.Lift uni a
-     , PLC.GEq uni
-     , PLC.Everywhere uni Eq
-     , MonadError (PIR.Error uni fun (Provenance ())) m
+  :: forall a m
+   . ( Lift.Lift UPLC.DefaultUni a
```
zliu41: I don't see why everything in this module needs to be hardcoded to `DefaultUni`. If PIR constant folding doesn't require this, then neither should UPLC constant folding.

zliu41 (changes requested):

plutus-tx/src/PlutusTx/Lift.hs
ana-pantilie: Is Plinth ever used with anything other than `DefaultUni` and `DefaultFun`? Are there any plans to add other universes of builtins? As far as I know there aren't any, and I believe abstraction just for the sake of abstraction is a code smell.

zliu41: If we decide that only `DefaultUni` is needed, the right thing to do is to open a PR and remove the `uni` type parameter altogether. That doesn't seem like a good idea though, since other universes are being used for testing and benchmarking.

If we keep the `uni` parameter, then this monomorphization just makes it inconsistent with everything else.

plutus-tx-plugin/test/Budget/9.6/toFromData.golden.uplc
```diff
@@ -1,82 +1 @@
...
-                                                                         , (\ds ->
-                                                                              True) ]
-                                                                         args) ]) ])
-                                                          (force tailList ds)) ]
-                                                   args) ]) ])
-                                , (\ds -> constr 1 []) ]
-                                args) ]) ]) ]
-             args) ]))
\ No newline at end of file
+(program 1.1.0 (constr 1 [(constr 0 [(constr 0 [True, 1, False])])]))
```
ana-pantilie, line 1: Looks like running this pass at the UPLC level as well unlocked further optimization opportunities.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Contexts.hs
ana-pantilie: Shadowing an import, renamed it to `arg`.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Contexts.hs
ana-pantilie: Good catch, also this wasn't used so I removed it.

I started with all of the logic of the PIR pass ported to UPLC, and removed what wasn't necessary 1-by-1 so that's why this (and other functions) ended up being left over. I've removed it now.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Optimize.hs
ana-pantilie: I wasn't sure where the best place for it was, looks like moving it after inlining did make more passes fire in some tests but in one it didn't.

plutus-tx/src/PlutusTx/Lift.hs
ana-pantilie: > the right thing to do is to open a PR and remove the uni type parameter altogether.

I don't think removing it from Plinth means we need to remove it from all of the other packages at the same time. Although, I do need to double check if they're used in some other Plinth code as well.

> That doesn't seem like a good idea though, since other universes are being used for testing and benchmarking.

I asked Claude where we use other universes, this is the response:

### Custom Builtin Function Sets
`ExtensionFun` — plutus-core/examples/PlutusCore/Examples/Builtins.hs
Used as `Either DefaultFun ExtensionFun` in example programs and tests. Includes factorial, polymorphic identity functions, cost-tracking builtins, unsafe coercions, and semantic variant testing.

### NopFun — cost-model/budgeting-bench/Benchmarks/Nops.hs
No-op builtins used exclusively for cost model calibration benchmarks — measure the raw overhead of calling a builtin function with varying argument types/counts.
Custom Builtin Function Sets `ExtensionFun` — plutus-core/examples/PlutusCore/Examples/Builtins.hs
Used as `Either DefaultFun ExtensionFun` in example programs and tests. Includes factorial, polymorphic identity functions, cost-tracking builtins, unsafe coercions, and semantic variant testing.

If Claude is correct, we're keeping the whole universe machinery just for an example and for cost model calibration. I think it would be worth looking into whether the cost model calibration can be represented in a different way. It doesn't make sense to keep all of the complicated universe abstraction around just for that. The current representation would be good if UPLC was used to target other platforms (not just Cardano), but that doesn't seem grounded in reality. Also, there might be performance benefits for the CEK machine. I think this would be worth exploring.

ana-pantilie: Btw, I'm reverting the change to `Lift` here since I tend to agree this would be better done separately, but I still think we should discuss ^.

zliu41: > If Claude is correct, we're keeping the whole universe machinery just for an example and for cost model calibration.

Cost model calibration sounds quite important to me?

Btw, as a general rule, avoid posting Claude answers verbatim. They are often quite verbose and of questionable quality - inaccuracy is common. You are of course free to use Claude, but make sure to quality-check, shorten and rephrase, make it suitable for human conversation before asking others to read it.

zliu41 (approved):

plutus-tx/src/PlutusTx/Lift.hs
ana-pantilie: > Cost model calibration sounds quite important to me?

Of course it's important, I'm wondering if there's some other way to represent it.

> Btw, as a general rule, avoid posting Claude answers verbatim. They are often quite verbose and of questionable quality - inaccuracy is common. You are of course free to use Claude, but make sure to quality-check, shorten and rephrase, make it suitable for human conversation before asking others to read it.

I'm 100% on the same page on this, so I'm not sure where this is coming from? If I ever post/send something for review which looks like AI slop I probably forgot to check it myself so please let me know.

PR #7766, opened 2026-05-04
Add detailed error message for unsupported casing on Integer

zliu41:
```
Error: Unsupported feature: Cannot pattern match on a value of type 'Integer'.

       This usually happens when pattern matching on an integer literal,
       for example:

         f 42 = ...
         f n  = ...

       Plinth does not support pattern matching on 'Integer'. Use equality
       from 'PlutusTx.Prelude' instead, for example:

         import PlutusTx.Prelude ((==))

         f n | n == 42   = ...
             | otherwise = ...

       Note: GHC can generate these unexpectedly, you may need '-fno-strictness', '-fno-specialise', '-fno-spec-constr', '-fno-unbox-strict-fields', or '-fno-unbox-small-strict-fields'.
   |ᴾᴸᴵᴺᵀᴴ
27 |      let y = PlutusTx.divideInteger 42 (f x)
```

### DIFF
--- a/plutus-tx-plugin/changelog.d/20260504_112339_unsafeFixIO_case_on_integer.md
+++ b/plutus-tx-plugin/changelog.d/20260504_112339_unsafeFixIO_case_on_integer.md
@@ -1,0 +1,3 @@
+### Added
+
+- Detailed error message for unsupported casing on Integer.
--- a/plutus-tx-plugin/src/PlutusTx/Compiler/Type.hs
+++ b/plutus-tx-plugin/src/PlutusTx/Compiler/Type.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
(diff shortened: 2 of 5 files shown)


Unisay (approved):
Nice error message!

Its peculiar that with GHC 9.6 we have 1 golden file change (caseInt) and with 9.12 we have two (caseInt, rangeEnumFromThenTo). 

It turns out that:
1. Plugin [tries to detect](https://github.com/IntersectMBO/plutus/blob/master/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs#L1206-L1225) if a symbol has to do with the `enumFromTo` or other functions from this family:
https://github.com/IntersectMBO/plutus/blob/master/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs#L365
2. This detection stopped working in GHC 9.12 - I suspect they moved symbols like `rangeEnumFromThenTo` to another namespace, like `GHC.Internal.Enum` instead of `GHC.Enum`.

It would be nice if you can fix this regression in this PR since its touching the corresponding test.

zliu41:
> Plugin [tries to detect](https://github.com/IntersectMBO/plutus/blob/master/plutus-tx-plugin/src/PlutusTx/Compiler/Expr.hs#L1206-L1225) if a symbol has to do with the enumFromTo

> This detection stopped working in GHC 9.12

Neither is accurate. `enumFromTo` is a result of list range syntax `[a..b]`. I don't think it has anything to do with pattern matching on Integer.

The plugin does try to detect `integerEq`, which is a result of pattern matching on Integer. But this is fragile, since GHC could easily decide to inline `integerEq`. This happens in 9.6 as well as 9.12.

Unisay:
> I don't think it has anything to do with pattern matching on Integer.

I think it does: in this PR the golden file `rangeEnumFromThenTo.golden.uplc` has changed (it contains integer casing error), and it comes from this code:

```
rangeEnumFromThenTo :: CompiledCode [Integer]
rangeEnumFromThenTo = plinthc [1, 7 .. 50]
```

PR #7767, opened 2026-05-05
Improve error reporting for a few unsupported types

zliu41:


plutus-tx-plugin/src/PlutusTx/Compiler/Type.hs
```diff
@@ -104,6 +115,51 @@ compileType t = traceCompilation 2 ("Compiling type:" GHC.<+> GHC.ppr t) $ do
...
+  Map.fromList
+    . mapMaybe
+      ( \(name, alt) -> do
+          modu <- TH.nameModule name
+          pure ((modu, TH.nameBase name), alt)
+      )
+    $ [ (''Prelude.Float, Just "Integer or PlutusTx.Ratio.Rational")
+      , (''Prelude.Double, Just "Integer or PlutusTx.Ratio.Rational")
+      , (''Data.Ratio.Ratio, Just "PlutusTx.Ratio.Rational")
+      , (''Data.Text.Text, Just "BuiltinString")
```
SeungheonOh, line 159: We made changes some time ago so that `BuiltinInteger ~ Integer`, `BuiltinBool ~ Bool` and etc. Do you remember if there were any reasons we didn't make that changes for `Text` and `ByteString`? `BuiltinByteString` and `BuiltinString` is just a strict wrapper around `ByteString` and `Text` so I think we can make similar change here. 

This is slightly of topic and would need other PR

SeungheonOh (approved):
Looks good to me

plutus-tx-plugin/src/PlutusTx/Compiler/Type.hs
zliu41: > BuiltinByteString and BuiltinString is just a strict wrapper around ByteString and Text

They are not strict wrappers - they are opaque wrappers, so that GHC doesn't mess with them in unpredictable ways. Maybe raw `ByteString` and `Text` can be made to work, but it might be infeasible since GHC may turn them into something else. `Integer` and `Bool` are much simpler.

PR #7769, opened 2026-05-06
[uplc] Use typical key order in optimized blueprint

zliu41:
Previously when the `uplc` executable optimizes a blueprint JSON, the field order would be shuffled in the new JSON.

While it technically possible to remember the order in the input JSON, aeson doesn't seem to provide helpers to do this, and it's nontrivial (a JSON object is fundamentally a map).

So I'm doing the simple thing: hardcode the order obtained from a typical blueprint file compiled from Aiken. This should be the field order used in the vast majority of blueprints.

### DIFF
--- a/plutus-ledger-api/executables/src/PlutusCore/Executable/Blueprint.hs
+++ b/plutus-ledger-api/executables/src/PlutusCore/Executable/Blueprint.hs
@@ -13,7 +13,7 @@ import PlutusLedgerApi.Common
 
 import Data.Aeson (Value (..))
 import Data.Aeson qualified as Aeson
-import Data.Aeson.Encode.Pretty (encodePretty)
+import Data.Aeson.Encode.Pretty (Config (..), Indent (..), defConfig, encodePretty', keyOrder)
 import Data.Aeson.Key qualified as Key
 import Data.Aeson.KeyMap qualified as KM
 import Data.ByteString.Lazy qualified as BSL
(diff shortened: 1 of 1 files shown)


plutus-ledger-api/executables/src/PlutusCore/Executable/Blueprint.hs
```diff
@@ -79,11 +79,39 @@ getPlutusVersion _ = error "Blueprint: top-level value is not an object"
...
    in case outp of
-        FileOutput file -> BSL.writeFile file (encodePretty updated)
-        StdOutput -> BSL.putStr (encodePretty updated)
+        FileOutput file -> BSL.writeFile file bytes
+        StdOutput -> BSL.putStr bytes
         NoOutput -> pure ()
 
+-- | Key order in a typical blueprint compiled from Aiken.
+blueprintKeyOrder :: [Text]
+blueprintKeyOrder =
```
ana-pantilie, line 91: Can you also add some documentation pointing to the blueprint schema?

ana-pantilie (approved):
Approving but please add docs so that it's clear why these are the blueprint keys.

plutus-ledger-api/executables/src/PlutusCore/Executable/Blueprint.hs
zliu41: The blueprint is described in CIP-57. It doesn't stipulate a particular key order. I just obtained this by compiling a simple Aiken script into a blueprint file.

PR #7774, opened 2026-05-08
[Chore] Extract shared term constructors into Transform.Lib

Unisay:
Preparatory refactor before changing CSE tests in a follow-up PR.

`Transform.Inline.Spec` already had a section of UPLC term-construction helpers (`var`, `lam`, `app`, `delay`, `case_`, `builtin`, `addInteger`). 

- [x] Lift them into a shared `Transform.Lib` module reusable across other test specs in the same testlib.
- [x] Make Simplifier Spec use these helpers.

### DIFF
--- a/plutus-core/plutus-core.cabal
+++ b/plutus-core/plutus-core.cabal
@@ -491,6 +491,7 @@ library untyped-plutus-core-testlib
     Transform.CaseOfCase.Spec
     Transform.EvaluateBuiltins.Spec
     Transform.Inline.Spec
+    Transform.Lib
     Transform.Simplify.Lib
     Transform.Simplify.Spec
 
--- a/plutus-core/plutus-core.cabal
+++ b/plutus-core/plutus-core.cabal
@@ -491,10 +491,10 @@ library untyped-plutus-core-testlib
     Transform.CaseOfCase.Spec
     Transform.EvaluateBuiltins.Spec
     Transform.Inline.Spec
-    Transform.Lib
     Transform.Simplify.Lib
     Transform.Simplify.Spec
 
+  other-modules:      Transform.Lib
   build-depends:
     , base                             >=4.9   && <5
     , base16-bytestring
--- a/plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
+++ b/plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
@@ -3,8 +3,8 @@
 {-| Common UPLC term-construction helpers shared across the
 @Transform.*.Spec@ test modules.
 
-Variables are referred to by a textual hint: @var "x"@ is an occurrence, @lam "x"
-body@ a binder, and @name "x"@ the bare 'Name'. See Note [Names from textual hints]. -}
+Variables are referred to by name: @var "x"@ is an occurrence, @lam "x"
+body@ a binder, and @name "x"@ the bare 'Name'. See Note [Names from strings]. -}
 module Transform.Lib
   ( T
   , var
@@ -35,22 +35,18 @@ import PlutusCore.MkPlc (mkConstant)
 import PlutusCore.Name.Unique (Name (..), Unique (..))
 import UntypedPlutusCore.Core.Type (Term (..))
 
-{- Note [Names from textual hints]
-These test modules build UPLC terms purely, without threading a 'Quote' supply to
-allocate fresh uniques. A variable is identified entirely by a textual hint:
-'name' maps a hint to a 'Name' whose 'Unique' is derived deterministically (and
-injectively) from the hint text via 'uniqueFromText', and 'var' / 'lam' wrap
-'name' for occurrences and binders.
+{- Note [Names from strings]
+These test modules build UPLC terms purely, without using the 'Quote' monad to
+allocate fresh uniques. A variable is identified entirely by its name: 'name'
+maps a 'String' to a 'Name' whose 'Unique' is derived injectively from it via
+'uniqueFromText', and 'var' / 'lam' wrap 'name' for occurrences and binders.
 
-The invariant is: one hint denotes one variable. Terms built from the same hint
-share a 'Unique' and so are the same variable; distinct hints yield distinct
-variables. This is what lets a binder and its occurrences be written
-independently (@lam "a" (var "a")@) yet still refer to the same variable, and
-lets an assertion mention a free variable by hint (@isStrictIn (name "a") term@)
-with no plumbing.
+This lets a binder and its occurrences be written independently
+(@lam "a" (var "a")@) yet still refer to the same variable, and lets an assertion
+mention a free variable by name (@isStrictIn (name "a") term@) with no plumbing.
 
 The trade-off against 'QuoteT' is that freshness is no longer correct by
-construction: reusing one hint for two variables meant to differ silently aliases
+construction: reusing one name for two variables meant to differ silently aliases
 them, since 'Name' equality compares only the 'Unique'. For these small,
 hand-written test terms that is an acceptable price for dropping the monadic
 plumbing; where capture is a genuine concern, 'QuoteT' remains available.
@@ -61,21 +57,21 @@ Alternatives considered: an 'IsString' or 'IsLabel' instance for 'Name'
 read as if @x@ were a bound Haskell variable. The plain @String -> Name@ used
 here is the simplest: no orphan instance and no language extension at all.
 
-'uniqueFromText' is injective only on short ASCII hints, so it fails fast on
-non-ASCII characters or on hints long enough to overflow the 'Int' rather than
+'uniqueFromText' is injective only on short ASCII names, so it fails fast on
+non-ASCII characters or on names long enough to overflow the 'Int' rather than
 aliasing silently.
 -}
 
 -- | Convenient alias used throughout the test modules.
 type T = Term Name DefaultUni DefaultFun ()
 
-{-| A 'Var' occurrence of the variable with the given hint.
-See Note [Names from textual hints] -}
+{-| A 'Var' occurrence of the variable with the given name.
+See Note [Names from strings] -}
 var :: String -> T
 var = Var () . name
 
-{-| A lambda binding the variable with the given hint.
-See Note [Names from textual hints] -}
+{-| A lambda binding the variable with the given name.
+See Note [Names from strings] -}
 lam :: String -> T -> T
 lam = LamAbs () . name
 
@@ -129,18 +125,18 @@ text = mkConstant @Text () . Text.pack
 err :: T
 err = Error ()
 
--- | Build a 'Name' from a hint. See Note [Names from textual hints]
+-- | Build a 'Name' from a 'String'. See Note [Names from strings]
 name :: String -> Name
 name s = Name t (Unique (uniqueFromText t))
   where
     t = Text.pack s
 
-{-| Pack the hint's bytes big-endian into an 'Int' (> 7 bytes overflow).
-See Note [Names from textual hints] for the injectivity and fail-fast contract. -}
+{-| Pack the string's bytes big-endian into an 'Int' (> 7 bytes overflow).
+See Note [Names from strings] for the injectivity and fail-fast contract. -}
 uniqueFromText :: Text -> Int
 uniqueFromText t
   | Text.any ((> 127) . fromEnum) t =
-      error ("Transform.Lib: non-ASCII name hint: " <> show t)
+      error ("Transform.Lib: non-ASCII name: " <> show t)
   | Text.length t > 7 =
-      error ("Transform.Lib: name hint too long (would overflow Unique): " <> show t)
+      error ("Transform.Lib: name too long (would overflow Unique): " <> show t)
   | otherwise = Text.foldl' (\acc c -> acc * 256 + fromEnum c) 0 t
--- a/plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
+++ b/plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
@@ -1,0 +1,146 @@
+{-# LANGUAGE TypeApplications #-}
+
+{-| Common UPLC term-construction helpers shared across the
+@Transform.*.Spec@ test modules.
+
+Variables are referred to by a textual hint: @var "x"@ is an occurrence, @lam "x"
+body@ a binder, and @name "x"@ the bare 'Name'. See Note [Names from textual hints]. -}
+module Transform.Lib
+  ( T
+  , var
+  , lam
+  , app
+  , force
+  , delay
+  , case_
+  , builtin
+  , constr
+  , sopTrue
+  , sopFalse
+  , builtinTrue
+  , builtinFalse
+  , ite
+  , con
+  , text
+  , err
+  , name
+  ) where
+
+import Data.Text (Text)
+import Data.Text qualified as Text
+import Data.Word (Word64)
+import GHC.Exts (fromList)
+import PlutusCore.Default (DefaultFun (IfThenElse), DefaultUni)
+import PlutusCore.MkPlc (mkConstant)
+import PlutusCore.Name.Unique (Name (..), Unique (..))
+import UntypedPlutusCore.Core.Type (Term (..))
+
+{- Note [Names from textual hints]
+These test modules build UPLC terms purely, without threading a 'Quote' supply to
+allocate fresh uniques. A variable is identified entirely by a textual hint:
+'name' maps a hint to a 'Name' whose 'Unique' is derived deterministically (and
+injectively) from the hint text via 'uniqueFromText', and 'var' / 'lam' wrap
+'name' for occurrences and binders.
+
+The invariant is: one hint denotes one variable. Terms built from the same hint
+share a 'Unique' and so are the same variable; distinct hints yield distinct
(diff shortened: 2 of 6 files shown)


plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
```diff
@@ -0,0 +1,128 @@
...
+err :: T
+err = Error ()
+
+-- | Generate a fresh 'Name' from a textual hint.
+name :: Text -> Name
+name = runQuote . freshName
+
+-- | Generate two fresh 'Name's with sequential 'Unique's.
+freshNames2 :: Text -> Text -> (Name, Name)
+freshNames2 a b = runQuote $ (,) <$> freshName a <*> freshName b
```
SeungheonOh: I think having some warning that freshNameN must be used exactly once within term you are generating here would be nice. doing something like 

```hs
foo = ... 
  where 
    (a, b) = freshName2 "a" "b"
    c = name "a"
```

it would generate such that `a` and `c` is actually same. 

though, all of the changes made are not making this mistake.

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
```diff
@@ -0,0 +1,128 @@
...
+  runQuote $ (,,,) <$> freshName a <*> freshName b <*> freshName c <*> freshName d
+
+-- | Generate five fresh 'Name's with sequential 'Unique's.
+freshNames5 :: Text -> Text -> Text -> Text -> Text -> (Name, Name, Name, Name, Name)
+freshNames5 a b c d e =
+  runQuote $
+    (,,,,) <$> freshName a <*> freshName b <*> freshName c <*> freshName d <*> freshName e
+
+-- | Generate eight fresh 'Name's with sequential 'Unique's.
+freshNames8
```
SeungheonOh: Honestly just making `T = Quote (Term Name DefaultUni DefaultFun ())` would be better just to not have this 🙃

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
```diff
@@ -0,0 +1,128 @@
...
+case_ scrut branches = Case () scrut (fromList branches)
+
+builtin :: DefaultFun -> T
+builtin = Builtin ()
+
+-- | A 'Constr' term tagged with the given index.
+constr :: Word64 -> [T] -> T
+constr = Constr ()
+
+-- | The standard UPLC encoding of @True@ as @Constr 0 []@.
```
basetunnel: Isn't the standard encoding `constant bool true`?

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
effectfully: You can have an infinite list of pure functions returning fresh names and a bit of pattern synonym sugar to apply those to strings. Smth like

```haskell
let N "a" a : N "f" f : N "z" z : _ = freshNames
in <...>
```

Or you can use this https://github.com/effectfully-ou/sketches/tree/master/better-syntax-for-eDSLs

SeungheonOh: It's beautiful but at the same time it's pattern synonym 🙃

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
Unisay: Good catch, and it ended up reshaping the design. The `freshNames2…8` ladder is gone. Names now come from `fresh.<hint>` (via `OverloadedRecordDot`), with the `Unique` derived from the hint text, so "one hint denotes one variable" is an invariant rather than something you have to keep in mind. The case you described, mixing a `freshNames2` call with a separate `name` and aliasing `Unique` 0, can't happen anymore. The invariant is documented on the module header and on `Fresh`.

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
Unisay: Right, the builtin encoding is `(con bool True)`. These helpers use the sum-of-products form (`Constr 0 []` / `Constr 1 []`) on purpose, because the case-of-case and simplifier passes under test work on `Constr`/`Case`, not on the builtin `bool`. I reworded the comment to call it the SOP/datatype encoding of `Bool` and to contrast it with the builtin `bool` constant, so "standard" isn't carrying misleading weight.

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
Unisay: Went with the `OverloadedRecordDot` approach from your sketch, @effectfully: `fresh.a` instead of the `freshNames2…8` ladder. A single `HasField` instance covers everything here, since these terms only ever need a `Name` (no separate `var`/`lam` prefixes). The literal pattern-synonym form (`let N "a" a : _ = freshNames`) doesn't typecheck: a pattern synonym can't feed its own argument into its view, so `N "a" a` fails with `Not in scope: hint`. That's what settled it on the dot syntax. The `Unique` is derived from the hint text (injective for the short hints these tests use), which also closes the shared-supply aliasing from the other thread. Net −95 lines across the transform specs, goldens unchanged.

basetunnel: > Went with the OverloadedRecordDot approach from your sketch

This solution doesn't look ideal to me: `OverloadedRecordDot` isn't a very common extension and I think you could get the same functionality using `OverloadedStrings`. It also doesn't generate fresh variable names, so the name `fresh` feels incorrect. @effectfully 's approach seems quite powerful once you incorporate the type class onstraints for well-scopedness, but it also has its own syntactic noise.

What was the motivation for changing this? If the goal was to have a simpler way to bind multiple uniques, could we perhaps define a type class + method that generates nested tuples of uniques in the `QuoteT` monad?

I quite like the `QuoteT` because you don't have to worry about variable capture

Unisay: Ok, my goal was to simplify code by getting rid of monadic contexts for pure term construction by sacrificing a little of correct-by-construction properties. This isn't something one would want to do in the production code, but in the test code IMO the sacrifice is worth it as we're dealing with a ton of simple expressions that I don't worry much about accidentally capturing names or scope violations. For the more complex cases where this is a concern you can still use QuoteT.
Also, for negative test cases we want to construct invalid terms.

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
```diff
@@ -0,0 +1,114 @@
...
+
+instance (res ~ Name, KnownSymbol hint) => HasField hint Fresh res where
+  getField _ = Name txt (Unique (uniqueFromText txt))
+    where
+      txt = Text.pack (symbolVal (Proxy @hint))
+
+{-| Pack the hint's bytes big-endian into an 'Int'; injective for the short
+ASCII hints used in these tests. -}
+uniqueFromText :: Text -> Int
+uniqueFromText = Text.foldl' (\acc c -> acc * 256 + fromEnum c) 0
```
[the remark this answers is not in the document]
Unisay: Addressed in cfac61ab10: `uniqueFromText` now fails fast (calls `error`) on non-ASCII characters or hints longer than 7 bytes, instead of silently producing a colliding `Unique`. The hints used in these tests are all short ASCII identifiers, so this only guards against future misuse.

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
Unisay: > I think you could get the same functionality using OverloadedStrings

Good point, and `OverloadedStrings` would genuinely cut code here: it drops the extension pragma (enabled in cabal already) plus the whole `HasField`/`Fresh` apparatus. The catch is that `"a" :: Name` needs an `IsString Name` instance, which is an orphan on a core type and exported from the test sublibrary for every dependent to see. I'd rather keep the derivation localized behind `fresh` than hang an orphan off `Name`. `OverloadedLabels` (`#a`) carries the same orphan cost and is an equally uncommon extension, so it doesn't really address your concern either. That said, if we're collectively fine with an orphan `IsString Name` in the test lib, I'm happy to switch, since it's strictly less code.

basetunnel: What about a plain smart constructor `:: String -> Name` (or from `Text`) that is injective? That would avoid extensions and instances so it's much easier to understand while also avoiding the Quote monad. I find overloading `x.y` quite confusing, because it looks like `y` is a regularly bound Haskell identifier.

basetunnel: > Also, for negative test cases we want to construct invalid terms.

What invalid terms can `QuoteT` not construct?

Unisay: Thanks, this is the simplification I was missing. I went with the plain injective smart constructor: `name :: String -> Name`, with the `Unique` derived deterministically from the hint, and `var` / `lam` composing it (`var "x"`, `lam "x" body`). That drops the whole `OverloadedRecordDot` + `HasField` apparatus, and since it takes `String` rather than `Text` it needs no orphan instance and no language extension at all (not even `OverloadedStrings`). Expressiveness is kept: `name` still hands you a bare `Name` for free-variable lists and the `isStrictIn` / `effectSafe` assertions. I documented the scheme and the alternatives in a `Note [Names from textual hints]`.

On invalid terms: you're right, scratch that. I confused it with the well-scopedness guarantee from @effectfully's scoped sketch (the type-level `InScope` trick), which `QuoteT` doesn't provide. `QuoteT` only manages unique freshness and is just as capable of building ill-formed terms, so that wasn't an argument for dropping it.

Unisay:
The red `ci/hydra-build:x86_64-linux.required` is an unrelated flake, not this PR: the `uplc-evaluator-integration-tests` leaf (ghc912) hit Hydra's 2 h cap and timed out ([build 13254785](https://ci.iog.io/build/13254785)). This PR only touches `untyped-plutus-core` testlib; the suite passes locally in seconds on both 9.6 and 9.12. Tracked in IntersectMBO/plutus-private#2257. Restarting the build clears the gate.

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
```diff
@@ -0,0 +1,146 @@
...
+import GHC.Exts (fromList)
+import PlutusCore.Default (DefaultFun (IfThenElse), DefaultUni)
+import PlutusCore.MkPlc (mkConstant)
+import PlutusCore.Name.Unique (Name (..), Unique (..))
+import UntypedPlutusCore.Core.Type (Term (..))
+
+{- Note [Names from textual hints]
+These test modules build UPLC terms purely, without threading a 'Quote' supply to
+allocate fresh uniques. A variable is identified entirely by a textual hint:
+'name' maps a hint to a 'Name' whose 'Unique' is derived deterministically (and
```
basetunnel: "deterministically" seems a bit superfluous, as any pure function is deterministic. The property of interest here is injectivity.

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
```diff
@@ -0,0 +1,146 @@
...
+import Data.Text qualified as Text
+import Data.Word (Word64)
+import GHC.Exts (fromList)
+import PlutusCore.Default (DefaultFun (IfThenElse), DefaultUni)
+import PlutusCore.MkPlc (mkConstant)
+import PlutusCore.Name.Unique (Name (..), Unique (..))
+import UntypedPlutusCore.Core.Type (Term (..))
+
+{- Note [Names from textual hints]
+These test modules build UPLC terms purely, without threading a 'Quote' supply to
```
basetunnel: I think you mean "threading a fresh variable supply" or "using the QuoteT" monad here?

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
```diff
@@ -0,0 +1,146 @@
...
+allocate fresh uniques. A variable is identified entirely by a textual hint:
+'name' maps a hint to a 'Name' whose 'Unique' is derived deterministically (and
+injectively) from the hint text via 'uniqueFromText', and 'var' / 'lam' wrap
+'name' for occurrences and binders.
+
+The invariant is: one hint denotes one variable. Terms built from the same hint
```
basetunnel: I think this paragraph doesn't add much to the last one, because it restates injectivity. The property that same names identify the same variable is already true for any representation of named variables.

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
```diff
@@ -0,0 +1,146 @@
...
+share a 'Unique' and so are the same variable; distinct hints yield distinct
+variables. This is what lets a binder and its occurrences be written
+independently (@lam "a" (var "a")@) yet still refer to the same variable, and
+lets an assertion mention a free variable by hint (@isStrictIn (name "a") term@)
+with no plumbing.
+
+The trade-off against 'QuoteT' is that freshness is no longer correct by
+construction: reusing one hint for two variables meant to differ silently aliases
```
basetunnel: Perhaps then "textual hint" is not the right terminology to use here. I would already expect that `var "x"` denotes the same variable as `var "x"` in a pure setting. The terminology seems appropriate in a `QuoteT` context: I'd associate "hints" with suggestions, where the actual resulting variable name may be different depending on the situation/freshness etc.

plutus-core/plutus-core.cabal
```diff
@@ -491,6 +491,7 @@ library untyped-plutus-core-testlib
     Transform.CaseOfCase.Spec
     Transform.EvaluateBuiltins.Spec
     Transform.Inline.Spec
+    Transform.Lib
```
basetunnel: If this module is to be used internally in this library, should it be in `other-modules`?

plutus-core/untyped-plutus-core/testlib/Transform/Simplify/Spec.hs
```diff
@@ -1,14 +1,26 @@
...
 module Transform.Simplify.Spec where
 
-import Data.Text (Text)
-import Data.Vector qualified as V
 import PlutusCore qualified as PLC
-import PlutusCore.MkPlc (mkConstant, mkIterApp, mkIterAppNoAnn)
-import PlutusCore.Quote (Quote, freshName, runQuote)
+import PlutusCore.MkPlc (mkIterApp, mkIterAppNoAnn)
 import Test.Tasty (TestTree, testGroup)
+import Transform.Lib
```
basetunnel, line 6: Is it necessary to import all identifiers of this module explicity?

basetunnel:
The tests become generally much more readable, nice. I've added some minor comments.

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
Unisay: Reworded to "without using the 'Quote' monad to allocate fresh uniques".

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
Unisay: Dropped "deterministically", kept "injectively".

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
Unisay: Trimmed the restatement, kept just the consequence: writing a binder and its occurrences independently, and naming a free variable in an assertion.

plutus-core/untyped-plutus-core/testlib/Transform/Lib.hs
Unisay: Agreed. Here the text is the variable's identity, not a QuoteT-style suggestion, so "hint" is misleading. Dropped it throughout for "name"/"string" and renamed the Note to [Names from strings].

plutus-core/plutus-core.cabal
Unisay: Moved Transform.Lib to other-modules; it's imported only by the sibling Spec modules within this library. I'd like this reusable beyond the testlib eventually, but there's no consumer outside it yet, so I'm keeping it internal for now and will move it to exposed-modules once a concrete need appears (a one-line change).

plutus-core/untyped-plutus-core/testlib/Transform/Simplify/Spec.hs
Unisay: Not necessary, but a deliberate best practice for me: explicit imports let a reader trace a symbol's origin from the source alone, without HLS or having to know every module in scope. I'd rather pay the verbosity than lose that, so I'm keeping the explicit list (and the other three specs use the same style).

basetunnel: Okay, I see. Perhaps we should think about what should be in our styleguide, which discourages it (https://github.com/IntersectMBO/plutus/blob/master/STYLEGUIDE.adoc#avoid-complex-imports-statements).

basetunnel (approved):

PR #7777, opened 2026-05-12
Resolve stale TODO in PlutusIR DeadCode liveBinding

zeme-wana:
## Summary

Resolves a stale `-- TODO: HasUnique instances for VarDecl and TyVarDecl?` comment in [`PlutusIR.Transform.DeadCode.liveBinding`](plutus-core/plutus-ir/src/PlutusIR/Transform/DeadCode.hs). Those `HasUnique` instances already exist in [`PlutusCore.Core.Type`](plutus-core/plutus-core/src/PlutusCore/Core/Type.hs#L238-L244), so the local `liveVarDecl`/`liveTyVarDecl` helpers that pattern-matched to extract the inner name were unnecessary. Inlined the calls to `live` directly on the decls and dropped the helpers and the TODO.


Unisay (approved):
👍🏼

PR #7779, opened 2026-05-12
UPLC optimization pass: hoist polymorphic builtins

zliu41:
Before: `...force (force fstPair)...`

After: `(\fstPairForced -> ...fstPairForced...) (force (force fstPair))`


plutus-tx-plugin/test/BuiltinList/Budget/9.6/cons.golden.uplc
```diff
@@ -1 +1 @@
-(program 1.1.0 (\xs -> force mkCons 0 xs))
```
SeungheonOh, line 1: Hoisting builtin that was used once(or even twice) is worse for performance. I don't think it would be too difficult to check how many times it was used. 

Also, this might not even need to be manually handled if we run inline pass after builtin poly at least for single-use cases.

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/PolyBuiltin.hs
```diff
@@ -0,0 +1,107 @@
...
+      checkFullyForced t = case t' of
+        Builtin _ fun
+          | n >= 1
+          , n == numForces fun ->
+              Just (getAnn t, fun, n)
+        _ -> Nothing
+        where
+          (n, t') = peelForces t
+
+      candidates :: HashMap fun Int
```
SeungheonOh, line 59: I don't think it's useful to keep track of number of forces applied here since we already know that from builtin itself and we are not hoisting any partially instantiated builtins. 


However, I think the `Int` here can be used to track occurrences so that only builtins that were used more than twice gets hoisted. This way we don't have to rely on inliner to do this.

SeungheonOh (commented):
adjusting optimization step could yield better result. Currently, BuiltinPoly runs at the very end, so hoisted builtin bindings won't get turned into more optimal `Case/Constr`. Also, inliner won't run on hoisted builtins that were used only once, which is the reason why there are some performance regressions. 

This is how it runs
```
  simplifyNTimes (_ooMaxSimplifierIterations opts)
    >=> runStage CseStage
    >=> runStage ApplyToCaseStage
    >=> runStage PolyBuiltinStage
```
would be better to do 
```
  runStage PolyBuiltinStage
    >=> simplifyNTimes (_ooMaxSimplifierIterations opts)
    >=> runStage CseStage
    >=> runStage ApplyToCaseStage
```

plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Transform/PolyBuiltin.hs
zliu41: This `Int` just keeps track of how many forces a builtin need, so that `applyForces` applies the right number of forces. We can call `builtinArity` in `applyForces` again, but that's more costly.

SeungheonOh:
I was wrong, we don't want to inline soley based on occurrence count, since it could be inlined into places that will be run multiple times. This would need more complicated fix-point detection to prevent inlining accross fixpoint combinator which we don't have now.

zliu41 (commented):
The idea is to hoist forced builtins blindly, like what Plutarch does.

It is run at the end of the pipeline to avoid the inliner inlining `let forcedFstPair = force (force fstPair) in List.any (\x -> ...forcedFstPair...) xs`.

SeungheonOh (approved):

PR #7781, opened 2026-05-14
Resolve TODO: rename TermOf to TermWith in Hedgehog generators

zeme-wana:


Unisay (approved):
Idk why this rename makes sense, but it looks ok (pretty mechanistic)

PR #7782, opened 2026-05-14
Resolve TODO: move typeUsages to plutus-core

zeme-wana:
## Summary
Moves `typeUsages` (plus shared `Usages` type, `getUsageCount`, `allUsed`) out of `PlutusIR.Analysis.Usages` into a new `PlutusCore.Analysis.Usages` module.
`PlutusIR.Analysis.Usages` gets deleted.
This addresses a TODO


plutus-core/plutus-core.cabal
```diff
@@ -612,7 +612,6 @@ library plutus-ir
     , mmorph
     , monoidal-containers
     , mtl
-    , multiset
```
Unisay, line 615: Nit: it's good practice to declare direct dependencies explicitly rather than rely on transitive ones. `PlutusIR.Subst` still imports `Data.MultiSet.Lens` after this change, so `multiset` is still a direct dep of `library plutus-ir` in spirit — it just happens to resolve transitively via `plutus-core`. Worth keeping it in `build-depends`.

Unisay (commented):
Clean, narrowly-scoped refactor. Resolves the TODO and improves layering as advertised. One small cabal-hygiene nit inline — non-blocking.

plutus-core/plutus-core.cabal
zeme-wana: build fails though with `unused package` when building with `-Werror`

Unisay: Ah, makes sense — `Data.MultiSet.Lens` lives in `lens`, not in `multiset`, so `library plutus-ir` has no direct import from `multiset` and `-Wunused-packages` is right to flag it. Withdrawing the nit.

Unisay (approved):

PR #7785, opened 2026-05-19
INLINE `decodeListOf`

zeme-wana:
Add `{-# INLINE decodeListOf #-}` to the small helper in [`decodeListOf `](plutus-core/plutus-core/src/PlutusCore/Data.hs) that drives every list-of-`Data` CBOR decode. This function is on the per-script-execution hot path and inlining it will yield performance improvements.


effectfully:
Welcome to my world haha.

zeme-wana:
This seems to actually worsen performance: closing.

PR #7788, opened 2026-05-21
Certifier: deduplicate ASTs in agda certificate

basetunnel:
Fixes https://github.com/IntersectMBO/plutus-private/issues/2162

Agda certificates can be slow to type-check for large programs. The current compilation trace records all passes, even if the pass did not change the AST. This means that agda is running unnecessary decision procedures during type checking.

For example, the marlowe semantics validator has a trace of 133 passes, while there are only 33 of those that change something about the AST. This PR deduplicates the trace so that only real transformations are certified.

Also
- Adds `Data.List.NonEmptySep` for representing the certifier trace properly on both Haskell/Agda side
- Move hints to their corresponding AST modules
- Print the top-level trace in a readable way

### DIFF
--- a/plutus-metatheory/src/Certifier.hs
+++ b/plutus-metatheory/src/Certifier.hs
@@ -18,21 +18,23 @@ import Data.Foldable
 import Data.List.Extra (replace)
 import Data.List.NonEmpty (NonEmpty (..))
 import Data.List.NonEmpty qualified as NE
-import Data.Maybe (fromMaybe)
+import Data.List.NonEmptySep as NES
 import Data.Text qualified as Text
 import Data.Text.IO qualified as T
 import System.Directory (createDirectoryIfMissing)
 import System.FilePath (takeBaseName, (</>))
+import Text.Printf (printf)
 
 import FFI.AgdaUnparse (AgdaUnparse (..), renderAgdaUnparse)
 import FFI.CostInfo
-import FFI.OptimizerTrace (Trace, mkFfiOptimizerTrace, toEvalResult)
+import FFI.OptimizerTrace
 import FFI.Untyped (UTerm)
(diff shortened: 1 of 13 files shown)


plutus-metatheory/src/Certifier.hs
```diff
@@ -358,12 +308,14 @@ writeCertificateProject
           certName = takeBaseName mainModulePath
       createDirectoryIfMissing True certDir
       createDirectoryIfMissing True (certDir </> "src")
+      for_ astModules $ \(modName, _) -> do
+        createDirectoryIfMissing True (certDir </> "src" </> moduleDir modName)
```
basetunnel, line 312: This directory per pass is anticipating a later change where the translation relation proof per pass is in a separate module like `Proof.agda`

plutus-metatheory/src/Data/List/NonEmptySep.hs
```diff
@@ -0,0 +1,34 @@
...
+  deriving (Show, Generic)
+  deriving anyclass (NFData)
+
+instance Functor (NonEmptySep sep) where
+  fmap f (Singleton x) = Singleton (f x)
+  fmap f (Cons x y xs) = Cons (f x) y (fmap f xs)
+
+instance Foldable (NonEmptySep sep) where
+  foldr f e (Singleton a) = f a e
+  foldr f e (Cons x _ xs) = f x (foldr f e xs)
```
ana-pantilie, line 26: Deriving these using generics doesn't work?

ana-pantilie (approved):
Thanks!

plutus-metatheory/src/Data/List/NonEmptySep.hs
basetunnel: Thanks, I didn't think of that. Will do that soon.

PR #7789, opened 2026-05-21
Fix the FloatDelay compiler optimization pass

### LINKED ISSUE #7193: Fix `isEssentiallyWorkFree`
`UntypedPlutusCore.Transform.FloatDelay.isEssentiallyWorkFree` doesn't make any sense:

```haskell
isEssentiallyWorkFree :: Term name uni fun a -> Bool
isEssentiallyWorkFree = \case
  LamAbs{}   -> True
  Constant{} -> True
  Delay{}    -> True
  Constr{}   -> True
  Builtin{}  -> True
  Var{}      -> False
  Force{}    -> False
  -- Unsaturated builtin applications should also be essentially work-free,
  -- but this is currently not implemented for UPLC.
  -- `UntypedPlutusCore.Transform.Inline.isPure` has the same problem.
  Apply{}    -> False
  Case{}     -> False
  Error{}    -> False
```

Why do we consider any `Constr` node to be essentially work-free if it may literally contain arbitrary computations?

And unsaturated builtin applications should get better treatment, as per the comment (I don't think the part about `isPure` is correct though, but not sure).

This issue is for fixing all of that, most likely simply by moving to `isWorkFree`.

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/2232
Fixes https://github.com/IntersectMBO/plutus/issues/7193

This PR fixes two problems: 
1. See the test case added. On master, the input program is transformed into `(\a -> 1) (constr 0 [error])` which would evaluate to `error`, whereas without the transformation the program evaluates to `1`. This proves that the optimization pass is not sound as it doesn't preserve semantics. Enabling the compiler transformation only when the argument to the delay is pure fixes this.
2. [This PR](https://github.com/IntersectMBO/plutus/pull/5527) replaced `isEssentiallyWorkfree` for PIR `FloatDelay` pass, but we forgot to replace it for UPLC as well. Checking that the argument of the delay is workfree is necessary for the transformation to actually be an optimization, since removing the delay from a costly argument will result in its evaluation.


effectfully:
Related: https://github.com/IntersectMBO/plutus/issues/7193

ana-pantilie:
Thanks @effectfully! I think the comment about unsaturated builtin applications is true, I'll just open a separate issue for it.

SeungheonOh (approved):
Looks good

PR #7796, opened 2026-05-26
Move evaluateCekLikeInProd to plutus-ledger-api

kwxm:
I want to move `evaluateCekLikeInProd`  from `plutus-benchmark` to `plutus-ledger-api` so that I can use it without depending on `plutus-benchmark`.  Let's check that it doesn't affect the benchmark results first though.


kwxm:
Hmmmm...

kwxm:
That seems to cause a definite slowdown.  Let's try adding `INLINE`.

kwxm:
I'm not sure if this is making a difference or not.  I'll come back to it later.

PR #7797, opened 2026-05-27
Optimal non-builtin valueOf in plutus-ledger-api Data.Value

Unisay:
## Summary

`PlutusLedgerApi.V1.Data.Value.valueOf` now walks the outer and inner `BuiltinList`s directly: `unsafeDataAsMap` for the maps, `unsafeDataAsB` for the keys, `unsafeDataAsI` for the result, `equalsByteString` to compare. The loop returns `0` from the nil-case, so there is no `Maybe` constructed and immediately scrutinised. Same semantics, no public API change.

Closes [plutus-private#2242](https://github.com/IntersectMBO/plutus-private/issues/2242).

## Evidence

The lookup matrix and the `unsafeDataAsValue` baseline are on a separate, never-merge PR ([#7798](https://github.com/IntersectMBO/plutus/pull/7798)) so the 96 golden files do not sit in this PR's CI surface. Those goldens regenerate on any upstream plugin or cost-model shift, which makes them noise for a regression gate.

Old vs new `valueOf`, CPU in thousands, GHC 9.6, plutus-ledger-api 1.65.0.0. Old = `master` (`Map.lookup` + `withCurrencySymbol` continuation). New = this PR.

Shape legend: **S1** = ada only (1 token); **S3** = ada plus 2 single-token policies (3 tokens); **S8** = ada plus 7 single-token policies (8 tokens); **S100** = ada plus 10 policies of 10 tokens each (101 tokens). Lookup position is the row index of the matching `(currency, token)` inside the outer/inner maps: **ada** = first, **middle** = around the centre, **last** = final entry, **miss** = absent key (full scan).

| shape | lookup at | old CPU | new CPU | speedup |
|-------|-----------|---------|---------|---------|
| S1    | ada       |   5 423 |   3 772 | 1.44× |
| S1    | miss      |   3 180 |   2 500 | 1.27× |
| S3    | ada       |   5 423 |   3 772 | 1.44× |
| S3    | middle    |   7 905 |   5 010 | 1.58× |
| S3    | last      |  10 065 |   6 247 | 1.61× |
| S3    | miss      |   7 500 |   4 974 | 1.51× |
| S8    | ada       |   5 423 |   3 772 | 1.44× |
| S8    | middle    |  14 385 |   8 721 | 1.65× |
| S8    | last      |  20 865 |  12 432 | 1.68× |
| S8    | miss      |  18 301 |  11 159 | 1.64× |
| S100  | ada       |   5 423 |   3 772 | 1.44× |
| S100  | middle    |  25 186 |  15 162 | 1.66× |
| S100  | last      |  46 787 |  27 852 | 1.68× |
| S100  | miss      |  24 781 |  14 870 | 1.67× |

First-position lookups (`ada`) are constant in both versions because both short-circuit on the first key match; the new path is 1.44× faster on the constant cost. Linear scans (`middle`, `last`, `miss`) get a 1.6–1.7× speedup that holds across shapes, because the per-step cost dropped from `Map.lookup`-with-dictionary-conversion to a single `equalsByteString` plus three raw `unsafeDataAs*` builtins. Memory dropped roughly in proportion.

For comparison against the builtin path (`unsafeDataAsValue` + `lookupCoin`), see [#7798](https://github.com/IntersectMBO/plutus/pull/7798); the headline is that the new `valueOf` beats the builtin path on first-position lookups for large values (S100 ada: 3.77 M vs 22.11 M, ~5.9× cheaper).

### DIFF
--- a/plutus-ledger-api/changelog.d/20260527_123507_yuriy.lazaryev_issue_2242_optimal_valueof.md
+++ b/plutus-ledger-api/changelog.d/20260527_123507_yuriy.lazaryev_issue_2242_optimal_valueof.md
@@ -1,0 +1,3 @@
+### Changed
+
+- `PlutusLedgerApi.V1.Data.Value.valueOf` rewritten to walk the underlying `BuiltinList` directly via `unsafeDataAsMap` / `unsafeDataAsB` / `unsafeDataAsI` and short-circuit on the first match. The previous implementation went through `Map.lookup`, which materialised a `Maybe` only to deconstruct it immediately. Semantics are unchanged.
--- a/plutus-tx-plugin/test-ledger-api/Spec/Data/Value.hs
+++ b/plutus-tx-plugin/test-ledger-api/Spec/Data/Value.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BlockArguments #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TemplateHaskell #-}
@@ -12,9 +13,13 @@ import Prelude qualified as Haskell
 
 import PlutusLedgerApi.V1.Data.Value
 
+import Plinth.Plugin (plinthc)
 import PlutusTx.Base
+import PlutusTx.Builtins qualified as B
+import PlutusTx.Builtins.Internal qualified as BI
 import PlutusTx.Code (CompiledCode, getPlc, unsafeApplyCode)
 import PlutusTx.Data.AssocMap qualified as AssocMap
+import PlutusTx.IsData qualified as Tx
 import PlutusTx.Lift
 import PlutusTx.List qualified as List
 import PlutusTx.Maybe
@@ -22,6 +27,7 @@ import PlutusTx.Numeric
 import PlutusTx.Prelude hiding (integerToByteString)
 import PlutusTx.Show (toDigits)
 import PlutusTx.TH (compile)
+import PlutusTx.Test.Run.Code (evalResult, evaluateCompiledCode)
 import PlutusTx.Traversable qualified as Tx
 
 import PlutusCore.Builtin qualified as PLC
@@ -31,12 +37,16 @@ import UntypedPlutusCore qualified as PLC
 import UntypedPlutusCore.Evaluation.Machine.Cek qualified as PLC
 
 import Control.Exception qualified as Haskell
+import Data.ByteString qualified as BS
 import Data.Functor qualified as Haskell
 import Data.List qualified as Haskell
 import Data.Map qualified as Map
+import PlutusLedgerApi.Test.V1.Data.Value qualified as ListToValue
 import Prettyprinter qualified as Pretty
+import Test.QuickCheck (Arbitrary (arbitrary), forAll, (===))
 import Test.Tasty
 import Test.Tasty.Extras
+import Test.Tasty.QuickCheck (testProperty)
 
 scalingFactor :: Integer
 scalingFactor = 4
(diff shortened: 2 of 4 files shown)


plutus-tx-plugin/test-ledger-api/Spec/Data/Value.hs
```diff
@@ -258,3 +264,24 @@ test_EqValue =
     $ [ test_EqCurrencyList "Short" currencyListOptions
       , test_EqCurrencyList "Long" currencyLongListOptions
       ]
+
+{-| Check that running the compiled 'valueOf' on CEK produces the same
+'Integer' as the host-Haskell 'valueOf', for arbitrary 'Value's and
+arbitrary @(CurrencySymbol, TokenName)@ pairs. -}
+test_valueOf :: TestTree
```
zliu41, line 285: I don't think this test is very useful. If the implementation of `valueOf` is buggy, then both the CEK evaluation and the Haskell evaluation will just produce the same buggy result. It's much better to test this against the `valueOf` implementation on the builtin Value.

zliu41 (commented):
Looks good, just one comment on the test.

plutus-tx-plugin/test-ledger-api/Spec/Data/Value.hs
Unisay: Reworked. The test now compares the non-builtin `valueOf` against the builtin lookup path (`unsafeDataAsValue` + `lookupCoin`) on the CEK machine; they share no source code, so disagreement is a real bug in one of them. The arbitrary `Value` generator can produce values that violate the builtin path's well-formedness invariants (empty inner maps, zero quantities, unsorted entries), so each generated value is normalised before comparison.

plutus-tx-plugin/test-ledger-api/Spec/Data/Value.hs
```diff
@@ -258,3 +267,65 @@ test_EqValue =
...
+compiledValueOf = $$(compile [||valueOf||])
+
+{-| Compiled builtin lookup: @\\bd cs tn -> lookupCoin cs tn (unsafeDataAsValue bd)@.
+Used as the independent oracle in the differential property test for 'valueOf'. -}
+compiledBuiltinLookup
+  :: CompiledCode (BI.BuiltinData -> BI.BuiltinByteString -> BI.BuiltinByteString -> Integer)
+compiledBuiltinLookup =
+  $$(compile [||\bd c t -> B.lookupCoin c t (B.unsafeDataAsValue bd)||])
+
+{-| Check that the non-builtin 'valueOf' agrees with the builtin lookup path
```
zliu41, line 283: Just keep the first sentence here. It is clear enough. Remove everything else. Like I said earlier today, always try to make Claude output more succinct before presenting it to others.

plutus-tx-plugin/test-ledger-api/Spec/Data/Value.hs
```diff
@@ -258,3 +267,65 @@ test_EqValue =
     $ [ test_EqCurrencyList "Short" currencyListOptions
       , test_EqCurrencyList "Long" currencyLongListOptions
       ]
+
+-- | Compiled non-builtin 'valueOf', evaluated on CEK by the property test.
+compiledValueOf :: CompiledCode (Value -> CurrencySymbol -> TokenName -> Integer)
+compiledValueOf = $$(compile [||valueOf||])
```
zliu41: Use `plinthc valueOf`. We are phasing out `compile`.

plutus-ledger-api/changelog.d/20260527_123507_yuriy.lazaryev_issue_2242_optimal_valueof.md
```diff
@@ -0,0 +1,3 @@
+### Changed
+
+- `PlutusLedgerApi.V1.Data.Value.valueOf` rewritten to walk the underlying `BuiltinList` directly via `unsafeDataAsMap` / `unsafeDataAsB` / `unsafeDataAsI` and short-circuit on the first match. The previous implementation went through `Map.lookup`, which materialised a `Maybe` only to deconstruct it immediately, and paid `ToData k` / `UnsafeFromData a` dictionary work per element. Semantics are unchanged.
```
zliu41: > and paid `ToData k` / `UnsafeFromData a` dictionary work per element

Your new implementation still pays this, via `unsafeDataAsB`, `unsafeDataAsMap` and `unsafeDataAsI`. Remove this claim.

plutus-tx-plugin/test-ledger-api/Spec/Data/Value.hs
Unisay: Done — kept only the first sentence.

plutus-tx-plugin/test-ledger-api/Spec/Data/Value.hs
Unisay: Done — switched both compiled bindings to `plinthc`.

plutus-ledger-api/changelog.d/20260527_123507_yuriy.lazaryev_issue_2242_optimal_valueof.md
Unisay: Done — dropped that claim; the new walk pays the same `unsafeDataAs*` cost per element.

zliu41 (approved):

PR #7799, opened 2026-05-27
Fuse unionWith and checkBinRel in PlutusLedgerApi.V1.Data.Value

Unisay:
`PlutusLedgerApi.V1.Data.Value.unionWith` now runs one outer `Map.union` and one outer `Map.map` that collapses the `These` shape inline per currency-symbol entry. The previous implementation built a `Map CurrencySymbol (Map TokenName (These Integer Integer))` intermediate via a `unionVal` helper and then re-walked it to apply `f`: three outer passes for a single conceptual merge. `checkBinRel` is refactored along the same shape with `Map.union` + `Map.all`; the inner per-currency merge now sits inside the short-circuiting `Map.all` walk, so a failing pair skips the inner merges of every later currency (`geq` / `leq` / `gt` / `lt` inherit the early exit). `unionVal` and `checkPred` are dropped (both were module-internal, not exported). Same semantics, no public API change.

The `Map.union` lookup-and-merge core (`O(L×R)`) is untouched, so this is a constant-factor change: one of the two outer `Map.map` passes over the post-union map goes away, and a currency present on only one side has its token map mapped once instead of wrapped and then re-mapped.

Closes [plutus-private#2243](https://github.com/IntersectMBO/plutus-private/issues/2243).


SeungheonOh (approved):
Nice optimization. maybe it could be faster if `Map.all` gets inlined as well but I'm not sure if that will be worth it

Unisay:
Thanks. Leaving this out of the PR for now. `Map.all` is `INLINEABLE`, but PIR keeps it as one shared binding; forcing the inline at the four `checkBinRel` sites might shave CPU at the cost of flat size (which this PR currently reduces), so the net sign is unclear. It's also shared across the module, so changing its pragma isn't local. Better as a measured follow-up on the evidence branch. Can open a ticket if it's worth chasing.

PR #7805, opened 2026-06-02
ci: harden workflows against script injection + fix draft-PR filtering

zeme-wana:
Several workflows interpolated untrusted `${{ github.event.* }}` values (and step outputs) directly into `github-script` bodies and shell `run:` blocks.  The fix is to pass these values through `env:` and read them as inert data: `process.env.X` in JS, quoted `"$X"` in shell.


Unisay (approved):
LGTM

PR #7809, opened 2026-06-08
Make default builtins 32-bit-correct (wasm32): bounds-check index/length narrowing

### LINKED ISSUE #7808: Default builtins should behave correctly when compiled to 32-bit targets (wasm32)
## Summary

Plutus can be compiled to 32-bit targets — notably **wasm32** (GHC's wasm backend), which is increasingly used for browser/edge transaction building, simulation, and signing. It compiles, but several **default builtins assume a 64-bit machine word**: they narrow an `Integer`/`Int64` index or length to the platform `Int` (via `fromIntegral`/`fromInteger`) **before bounds-checking**, in order to call `Data.ByteString` / `Data.Vector` (which are `Int`-indexed). On a 64-bit platform that conversion is lossless; on **wasm32 (`Int`/`Word` = 32-bit)** it **wraps**, so evaluation **silently diverges** from native.

The result is a determinism break for any off-chain consumer that evaluates Plutus on a 32-bit target: the same script + arguments can produce different results than a 64-bit node. Block-producing nodes run natively (64-bit) and are **unaffected** — this is **not** a consensus/on-chain issue — but it breaks the guarantee that off-chain wasm32 evaluation matches on-chain behavior.

**Ask:** make the default builtins **32-bit-correct** — identical observable behavior on 32-bit and 64-bit targets.

## Root cause

Two patterns, both at the builtin boundary (the CEK machine itself is fine — it carries integers as unbounded `Integer`):

1. **Unchecked narrowing of indices/lengths.** Builtins `fromIntegral`/`fromInteger` an `Integer`/`Int64` argument to the platform `Int` before validating it is in range, then index a `bytestring`/`vector`. On 32-bit, values `>= 2^31` (or `>= 2^32`) wrap. The mismatch is fundamental: Plutus indices are arbitrary-precision `Integer`, but the underlying containers are machine-`Int`-indexed.
2. **Hardcoded 64-bit word.** `integerToByteString` computes the minimal positive byte length as `8 - clz#/8`, hardcoding an 8-byte word; on wasm32 `clz#` operates on a 32-bit word, so the length is over-counted.

## Confirmed divergences (native x86_64 vs wasm32 via wasmtime, differential-tested)

- **`sliceByteString`** — `BS.take (fromIntegral n) (BS.drop (fromIntegral start) xs)` with `Int64` start/len.
  `sliceByteString 0 4294967296 0x4299` → native `4299`, wasm `<empty>`.
  `sliceByteString 2147483648 1 0x4299` → native `<empty>`, wasm `42`.
- **`indexByteString`** — `BS.indexMaybe xs (fromIntegral n)`.
  `indexByteString [0x42,0x99] 4294967296` → native **error (out of bounds)**, wasm `0x42` (index wrapped to 0).
- **`readBit`** — bit index `fromIntegral`.
  `readBit [0x01] 4294967296` → native **error**, wasm `true` (bit 0 read).
- **`integerToByteString`** (small-int / `IS` branch) — hardcoded 64-bit word.
  `integerToByteString BigEndian 0 1` → native `0x01`, wasm `0x0000000001` (5 bytes); a fixed-width request like `integerToByteString BigEndian 2 1` errors on wasm.

All four are **silent-wrong-output** (or native-error-vs-wasm-success), the most dangerous class for determinism, and are present on `master` (the PlutusTx wrappers in `PlutusTx/Builtins/Internal.hs` still narrow unconditionally; the evaluator denotations' `Int` use is gated behind `#if WORD_SIZE_IN_BITS == 64`, which is not an affirmative 32-bit-correct path).

## Suspected (static audit, not yet differential-confirmed)

- **`IndexArray`** — `vec Vector.!? fromIntegral n`.
- **`PlutusTx.Builtins.Internal.caseInteger`** — `b !! fromIntegral i`.
- **bit-length arithmetic** `8 * len` for impractically large bytestrings (`> 2^28` bytes).

## Fix direction

- For indexing/slicing/bit builtins: validate the index/length against bounds in `Integer`/`Int64` space **before** narrowing to `Int` (reject out-of-range as the native semantics already do, or clamp where the spec says clamp).
- For `integerToByteString`: derive the word size from the target (`finiteBitSize (0 :: Int)`) instead of hardcoding `8`.

## Environment / method

- GHC wasm backend (`wasm32-wasi`), where `Int`/`Word` are 32-bit and `clz#`/`fromIntegral` behave correctly *for a 32-bit word* — the bug is in the builtins' assumptions, not in the primops.
- Differential method: evaluate each builtin natively and under wasmtime over a boundary corpus (`0`, `±1` around powers of 256, around `2^31` and `2^63`, large bignums, both endiannesses, fixed- and minimal-length modes).

A draft PR with the fixes and a native-vs-wasm differential test follows and references this issue.

paolino:
Fixes #7808.

Several default builtins narrow an `Integer` index/length to the platform `Int` **before** bounds-checking. On 64-bit targets the narrowing is lossless; on 32-bit targets (notably wasm32, GHC's wasm backend) it wraps, so the same script and arguments can evaluate differently than on a 64-bit node. Affected: `sliceByteString`, `indexByteString`, `readBit`, `indexArray`, `integerToByteString` (hardcoded 8-byte word), and the corresponding PlutusTx wrappers.

### Approach

All changes are gated behind `WORD_SIZE_IN_BITS` (the codebase's existing idiom, cf. `PlutusCore.Default.Universe` and `PlutusCore.Flat.*`):

- **64-bit targets compile the original source unchanged.** No semantics, costing, codegen or golden change on any platform a block producer runs on; there is nothing to re-benchmark.
- On 32-bit targets, indices and lengths are bounds-checked in `Integer` space before narrowing, restoring the 64-bit observable behaviour.

The test cases added to `Evaluation.Builtins.Definition` are unconditional: they pass on 64-bit before and after this change, and fail on an unpatched 32-bit build.

### Verification

- `plutus-core-test`, `untyped-plutus-core-test` and `plutus-tx-test` pass on this branch (x86_64-linux, ghc96) with **no golden changes**.
- The same fix applied to a release base passes the **full `plutus-core-test` suite (2697 tests, `--hedgehog-tests 1000`) cross-compiled to wasm32-wasi and run under wasmtime**, i.e. with `WORD_SIZE_IN_BITS == 32`: https://github.com/lambdasistemi/plutus/tree/wasm32-fixes. The same tree also passes all 37 nix flake-check test suites on x86_64-linux.

Follow-up branches exist that (a) enable building and running the test suites on wasm32 and (b) add a nightly wasm32 CI job; happy to open them as separate PRs if there is interest.

### DIFF
--- a/plutus-core/plutus-core/src/PlutusCore/Bitwise.hs
+++ b/plutus-core/plutus-core/src/PlutusCore/Bitwise.hs
@@ -1,8 +1,11 @@
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE MagicHash #-}
 {-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE UnboxedTuples #-}
+
+#include "MachDeps.h"
 
 {-| Implementations for CIP-121, CIP-122 and CIP-123. Grouped because they all operate on
 'ByteString's, and require similar functionality. -}
@@ -57,7 +60,9 @@ import GHC.Exts
   , intToInt8#
   , isTrue#
   , neWord8#
+#if WORD_SIZE_IN_BITS == 64
   , quotInt#
+#endif
   , quotRemInt#
   , sizeofByteArray#
   , word2Int#
@@ -170,7 +175,16 @@ unsafeIntegerToByteString requestedByteOrder requestedLength input = case input 
           -- can then figure out the number of unused bytes (all-zero)
           -- by taking the quotient of the leading zero count by 8.
           let counted# = clz# (int2Word# i#)
+#if WORD_SIZE_IN_BITS == 64
               minLength = 8 - I# (quotInt# (word2Int# counted#) 8#)
+#else
+              -- `clz#` counts leading zeroes of the platform word, so the
+              -- unused byte count must be derived from the actual word size
+              -- rather than a hardcoded 8 bytes.
+              wordSizeInBits = Bits.finiteBitSize (0 :: Int)
+              bitLength = wordSizeInBits - I# (word2Int# counted#)
+              minLength = (bitLength + 7) `quot` 8
+#endif
            in if
                 | requestedLength == 0 -> Right (mkSmall minLength i#)
                 | requestedLength < minLength -> Left NotEnoughDigits
@@ -509,6 +523,7 @@ complementByteString bs = unsafeDupablePerformIO . BS.useAsCStringLen bs $ \(src
 {-# INLINEABLE complementByteString #-}
 
 -- | Bit read at index, as per [CIP-122](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0122)
+#if WORD_SIZE_IN_BITS == 64
 readBit :: ByteString -> Int -> BuiltinResult Bool
 readBit bs ix
   | ix < 0 = do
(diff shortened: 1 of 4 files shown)


paolino:
Maintained at https://github.com/lambdasistemi/plutus (wasm32-fixes) instead.

PR #7816, opened 2026-06-15
Kwxm/parser/value ascending keys

kwxm:
Previously the UPLC parser would accept literal `value`s where currency and token IDs could be unordered and could have repetitions, which would be merged.

This adds a check to the parser to ensure that literal `value`s satisfy

* All currency IDs are in strictly ascending order
* For each currency, all token IDs are in strictly ascending order
* No currency or token ID is more than 32 bytes long
* All quantities are nonzero and lie in the range [-2^127, 2^127-1]

This makes the parser consistent with the flat decoder and the specification.  

I've also updated and reorganised the conformance tests a bit to reflect the changes.


plutus-conformance/test-cases/uplc/evaluation/builtin/constant/value/currencyID-too-long-2/currencyID-too-long-2.uplc
```diff
@@ -0,0 +1,16 @@
+-- Reject a currency ID that is too long: note that the currencies are in the wrong order here.
+(program 1.0.0
+  (con value
+    [ (#bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb,
+        [ (#0011, 12345)
+        , (#0022, 45678)
+        ]
+      )
+    , (aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
```
Unisay: The second currency here is missing its leading `#`, so the parser rejects it while parsing the bytestring literal rather than on the length check. The test still yields `parse error`, but it no longer exercises the intended path: an over-long currency ID, plus the "bounds are checked before ordering" interaction the comment describes. Adding the `#` restores that coverage. Compare `currencyID-too-long-1`, where it's present.

Unisay (changes requested):
The change is right: it brings the textual `value` parser in line with the flat/CBOR decoders and the spec, so strictly ascending IDs, no zero quantities, no empty token maps, and the length bounds are all enforced at parse time. The validation logic checks out and the unit tests cover the success/failure matrix for both currencies and tokens. I also confirmed that `ill-formed`/`overflow`/`underflow` produce genuine parse errors on this branch, so dropping them from the Agda failing-list is correct.

Two things block approval:

- The changelog check is red: there's no fragment under `plutus-core/changelog.d/` and no `No Changelog Required` label. Since the parser now rejects literals it used to accept, I'd add a `Changed` entry rather than the label.
- `currencyID-too-long-2.uplc` is missing the leading `#` on its second currency, so it fails while parsing the bytestring literal instead of on the length check. It no longer exercises the case its comment describes (see the inline note).

One point to discuss rather than a blocker: the parser reimplements the canonical-form invariants that `buildValueWith` already enforces, and then still routes the result through `Value.fromList`, where the merge and normalise steps are now unreachable. Sharing one validating constructor between the parser and the decoders would remove the duplication. The manual checks do produce better error messages, so if that's the reason for keeping them, it's worth a comment in the code.

plutus-conformance/test-cases/uplc/evaluation/builtin/constant/value/currencyID-too-long-2/currencyID-too-long-2.uplc
kwxm: > The second currency here is missing its leading #

Good point: I've fixed that.

kwxm: > Since the parser now rejects literals it used to accept, I'd add a Changed entry rather than the label.

Yeah, I debated whether to mention it or not.  I'll add a changelog entry for completeness.

> The changelog check is red:

No it isn't!  I added the label moments after opening the PR.  I'll add a changelog entry anyway though.

kwxm:
> still routes the result through `Value.fromList`, where the merge and normalise steps are now unreachable.

I was going to bring that up separately.  `Value.fromList` is used elsewhere, and doesn't include all of the checks that we want.  I'm wondering whether to update it (which I think might duplicate some work that's already done in the flat decoder) or change the name to something like `unsafeFromList`.  I though it was better to duplicate some functionality in the parser for the time being rather than start changing the code in `Value.hs`, which is more important than the parser.

SeungheonOh (approved):
Changes looks good, but for sake of uniformity, I think we can use `PlutusCore.Value.buildValueWith` here directly to enforce invariant rather than having a separate validation code

kwxm:
> I think we can use `PlutusCore.Value.buildValueWith` here directly to enforce invariant rather than having a separate validation code

Good point, but that turned out to be more difficult than it appeared to be.  `buildValueWith` takes (a generalised version of) a `[(K, [(K, Quantity)])]` list, so you either have to parse a list in that form or parse a `[(ByteString, [(ByteString, Integer)])]`list and convert it afterwards.  I went for the former, using `k` and `quantity` from `PlutusCore.Value`.  These perform the size checks, which lets you provide more precise error positions (although they still seem to be one parser token too late); the earlier version would only fail after the whole nested list had been parsed, so you'd just get a pointer to the end of that.  I ended up spending quite a long time fiddling with monadic code to reduce duplication, so now it's quite complicated.  I'm not sure if that's any better.  I was also too lazy to make the parser check the ordering, so that still gets checked by `buildValueWith` and you get an error at the end of the value if it's incorrect.

plutus-core/plutus-core/src/PlutusCore/Parser/Builtin.hs
```diff
@@ -19,19 +19,38 @@ import PlutusCore.Value qualified as PLC (Value)
...
+  , getSourcePos
+  , takeWhileP
+  )
 import Text.Megaparsec.Char (char, hexDigitChar, string)
 import Text.Megaparsec.Char.Lexer qualified as Lex
 
 cachedBuiltin :: Map.Map T.Text DefaultFun
 cachedBuiltin = Map.fromList [(display fn, fn) | fn <- [minBound .. maxBound]]
 
+-- | Parser of constants whose type is in 'DefaultUni'.
```
kwxm, line 39: This isn't new: it was further down the file and I moved it up.

SeungheonOh (approved):
having the validation logic linked to the actual one is good; I think it's worth having some extra complication, at least better than these validations being misaligned in the future.

Changes looks good!

Unisay (approved):

PR #7818, opened 2026-06-16
Add Plinth parsed-AST deriving plugin to plutus-tx-plugin

Icelandjack:
## What

Ports the `PlinthPlugin` parsed-result plugin from the ghc-plinth "standalone
Plinth" effort into `plutus-tx-plugin`. It expands data declarations written
with `deriving … via PlinthPlugin` into:

- `AsData` pattern synonyms (a newtype over `BuiltinData` plus bidirectional patterns)
- `Optics` prisms
- `Match` destructor functions

It runs at `parsedResultAction`, alongside the existing core `Plinth.Plugin`.
The deriving sentinel lives in `PlinthPlugin.Via` (`data Via = PlinthPlugin`).

## Notes / open items

- Builds on GHC 9.6 and 9.12 via CPP shims (9.12 verified locally; 9.6 relies on CI here).
- No activation/golden test yet: the modules compile, but nothing loads the plugin in plutus (in ghc-plinth it is baked in as a static plugin). A `test-frontend-plugin`-style golden test is the natural follow-up.
- Licensing: the code originated as Evoke (MIT); needs an attribution decision relative to plutus's Apache-2.0.

### DIFF
--- a/plutus-tx-plugin/plutus-tx-plugin.cabal
+++ b/plutus-tx-plugin/plutus-tx-plugin.cabal
@@ -21,6 +21,7 @@ source-repository head
 common lang
   default-language:   Haskell2010
   default-extensions:
+    BlockArguments
     DeriveFoldable
     DeriveFunctor
     DeriveGeneric
@@ -31,6 +32,8 @@ common lang
     FlexibleContexts
     GeneralizedNewtypeDeriving
     ImportQualifiedPost
+    LambdaCase
+    PatternSynonyms
     ScopedTypeVariables
     StandaloneDeriving
 
@@ -59,8 +62,10 @@ library
     PlutusTx.Compiler.Error
     PlutusTx.Options
     PlutusTx.Plugin.Common
+    PlutusTx.Plugin.Deriving.Via
 
   other-modules:
+    Paths_plutus_tx_plugin
     PlutusTx.Compiler.Binders
     PlutusTx.Compiler.Builtins
     PlutusTx.Compiler.Compat
@@ -75,8 +80,20 @@ library
     PlutusTx.PIRTypes
     PlutusTx.PLCTypes
     PlutusTx.Plugin.Boilerplate
+    PlutusTx.Plugin.Deriving
+    PlutusTx.Plugin.Deriving.Constant.Module
+    PlutusTx.Plugin.Deriving.Generator.AsData
+    PlutusTx.Plugin.Deriving.Generator.Common
+    PlutusTx.Plugin.Deriving.Generator.Match
+    PlutusTx.Plugin.Deriving.Generator.Optics
+    PlutusTx.Plugin.Deriving.Hs
+    PlutusTx.Plugin.Deriving.Hsc
+    PlutusTx.Plugin.Deriving.Type.Constructor
+    PlutusTx.Plugin.Deriving.Type.Field
+    PlutusTx.Plugin.Deriving.Type.Type
     PlutusTx.Plugin.Unsupported
 
+  autogen-modules: Paths_plutus_tx_plugin
   build-depends:
     , array
     , base                   >=4.9   && <5
@@ -242,8 +259,10 @@ test-suite frontend-plugin-tests
   hs-source-dirs: test-frontend-plugin
   main-is:        Spec.hs
   other-modules:
+    AsData.Spec
     Inlineable.Lib
     Inlineable.Spec
+    Match.Spec
     NoStrict.Spec
     Strict.Spec
 
@@ -254,6 +273,7 @@ test-suite frontend-plugin-tests
     , plutus-tx-plugin                 ^>=1.66
     , plutus-tx:plutus-tx-testlib
     , tasty
+    , tasty-hunit
 
   ghc-options:    -threaded -rtsopts -with-rtsopts=-N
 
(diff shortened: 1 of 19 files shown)


plutus-tx-plugin/src/PlutusTx/Plugin/Deriving/Generator/Common.hs
```diff
@@ -0,0 +1,365 @@
...
+--
+-- where @PlinthPlugin@ is the promoted constructor of @data Via = PlinthPlugin@.
+-- Using a real constructor (rather than a @Symbol@ string literal) means the
+-- name must be in scope and well-kinded, so when the plugin is not loaded GHC
+-- reports a clean type error instead of a confusing one.
+--
+-- This function is responsible for analyzing a deriving strategy to determine
+-- if the plugin should fire or not.
+parseDerivingStrategy ::
+  Maybe (Ghc.LDerivStrategy Ghc.GhcPs) -> Maybe [String]
```
hsyl20: Shouldn't it return a Bool?

plutus-tx-plugin/src/PlinthPlugin.hs
```diff
@@ -0,0 +1,347 @@
...
+  let config = Config.fromFlags flags
+
+  Monad.when (Config.help config)
+    . Hsc.throwError srcSpan
+    . Ghc.vcat
+    . fmap Ghc.text
+    . lines
+    $ Console.usageInfo ("PlinthPlugin version " <> version) Flag.options
+  Monad.when (Config.version config) . Hsc.throwError srcSpan $
+    Ghc.text version
```
hsyl20: Why do we need this command-line stuff?

It adds quite a lot of complexity (`Paths_` module, `Flag`, `Config`) and I'm not sure why?

plutus-tx-plugin/test-frontend-plugin/AsData/Spec.hs
```diff
@@ -0,0 +1,53 @@
+{-# OPTIONS_GHC -fplugin PlinthPlugin #-}
+-- The plugin-generated code uses 'head'/'tail' internally and emits pattern
+-- synonyms without top-level signatures; suppress those warnings here (the
+-- -Wno-unrecognised-warning-flags keeps -Wno-x-partial harmless pre-GHC 9.8).
```
hsyl20: Could we generate explicit pattern matches instead of head/tail?

Could we generate the top-level signatures too?

plutus-tx-plugin/src/PlinthPlugin.hs
```diff
@@ -0,0 +1,347 @@
...
+--
+-- > data Shape = Point | Circle Integer Integer
+-- >   deriving AsData via PlinthPlugin
+-- >   deriving Optics via PlinthPlugin
+plugin :: Ghc.Plugin
+plugin =
+  Ghc.defaultPlugin
+    { Ghc.parsedResultAction = parsedResultAction,
+      Ghc.pluginRecompile = Ghc.purePlugin
+    }
```
hsyl20: I thought we would merge this plugin's features into plutus-tx-plugin directly so that users of plutus-tx-plugin automatically benefit from it. Is it possible?

plutus-tx-plugin/src/PlinthPlugin/Generator/Common.hs
```diff
@@ -0,0 +1,365 @@
...
+  lHsSigType <- case Ghc.unLoc lDerivStrategy of
+    Ghc.ViaStrategy (Ghc.XViaStrategyPs _ x) -> Just $ Ghc.unLoc x
+    _ -> Nothing
+  lHsType <- case lHsSigType of
+    Ghc.HsSig _ _ x -> Just x
+  rdrName <- case Ghc.unLoc lHsType of
+    Ghc.HsTyVar _ _ x -> Just $ Ghc.unLoc x
+    _ -> Nothing
+  case Ghc.occNameString $ Ghc.rdrNameOcc rdrName of
+    "PlinthPlugin" -> Just []
```
hsyl20:
```suggestion
    "Plinth" -> Just []
```

Users don't care that it is a plugin or not.

plutus-tx-plugin/plutus-tx-plugin.cabal
```diff
@@ -56,11 +56,27 @@ library
...
+    PlinthPlugin.Generator.Match
+    PlinthPlugin.Generator.Optics
+    PlinthPlugin.Hs
+    PlinthPlugin.Hsc
+    PlinthPlugin.Options
+    PlinthPlugin.Type.Config
+    PlinthPlugin.Type.Constructor
+    PlinthPlugin.Type.Field
+    PlinthPlugin.Type.Flag
+    PlinthPlugin.Type.Type
```
hsyl20: I'd move these into `PlutusTx.Plugin.Deriving.*` or something like this

plutus-tx-plugin/src/PlinthPlugin.hs
Icelandjack: It's now wired in at `GHC.parsedResultAction = Deriving.parsedResultAction`. Users of `-fplugin Plinth.Plugin` get `deriving ... via Plinth`.

plutus-tx-plugin/test-frontend-plugin/AsData/Spec.hs
Icelandjack: Both done in latest push. No head/tail, `Match` generator and `AsData` use an explicit pattern. Top-level signatures are generated.

PR #7827, opened 2026-07-01
Add a `with-crypto` cabal flag to build Plinth/Plutus scripts without system crypto C libraries

zeme-wana:
**Problem**
Compiling a Plinth script transitively pulls `cardano-crypto-class`, which forces the system C libraries `libsodium/libblst/libsecp256k1` at cabal-solve time (cabal#4087). 
Script authors must install these even though compiling a script never runs any crypto.

**Change**
New `with-crypto` flag on `plutus-core` (default on, mirrored on `plutus-metatheory`). 
The 6 `PlutusCore.Crypto.*` modules are now CPP-gated:

**Tests**
New `crypto-hash-parity-test`: 114 goldens (6 hashes × 19 inputs, incl. block/padding edge cases) generated by the `+with-crypto` build (`cardano-crypto-class`) and reproduced byte-for-byte by the `-with-crypto` build (`crypton`), this ensures hash parity in CI.

**Net effect** 
Default builds are unchanged; script authors can opt into a C-library-free toolchain with flags: `-with-crypto`.


SeungheonOh:
I assume the main goal we want to achieve is being able to compile `plutus-tx-plugin` and `plutus-ledger-api` without the crypto c libraries. While this PR does allow that, I think the cleaner way is to separate the AST definitions from `plutus-core`(perhaps call it `plutus-ast` and to make `plutus-tx-plugin` and `plutus-ledger-api` only depend on `plutus-ast` which does not have any c crypto dependencies(or any dependencies really). 

Doing it that way would allow us to achieve same thing without requiring cabal flags which I'd prefer. 

Update:
Okay, seems like this approach is already been tried and is more difficult than I originally anticipated.

plutus-core/plutus-core/src/PlutusCore/Crypto/BLS12_381/Error.hs
```diff
@@ -1,6 +1,31 @@
...
+{-| Structural validation of a compressed BLS12-381 point, shared by the C-free
+G1 and G2 `uncompress` stubs (@expectedLen@ is 48 for G1, 96 for G2). It checks
+exactly what can be checked without field arithmetic: the length, and the three
+metadata bits of the leading byte -- the compression bit must be set, and if the
+infinity bit is set the encoding must be the canonical @0xc0 00..00@. It
+deliberately does NOT verify that the bytes encode a point on the curve or in the
+subgroup (that needs blst), so it is weaker than the real `blsUncompress`; but it
+rejects the malformed inputs the real decoder rejects on structure alone, rather
+than accepting anything. -}
+checkCompressed :: Int -> BS.ByteString -> Either BLS12_381_Error BS.ByteString
```
SeungheonOh, line 21: I'd rather not have this duplicate validation logic that's originally in the blst c library. `\_ _ -> Left (CryptoDisabled :: BLS12_381_Error)` would be sufficient

zliu41 (approved):
Looks good!

Does every crypto builtins call into one of the C libraries? If so, the name "with-crypto" is appropriate. Otherwise it isn't and we should use a different name.

Also we should update the Plinth user guide and the plinth-template documentation accordingly.

PR #7832, opened 2026-07-05
Metatheory: add Value built-in type with postulated functions

basetunnel:
Adds a postulated type `Value` and corresponding postulated built-in functions. Fixes fixes https://github.com/IntersectMBO/plutus-private/issues/1796.

I've also refactored the pattern matching of `convertModel`, while I was changing it to work for the new built-in functions. `convertModel` had a catch-all clause that broke the Agda CEK machine after a new `RawModel` constructor was added (which is used by one of the value built-ins). Such problems should from now on be caught by the coverage checker.

This also fixes https://github.com/IntersectMBO/plutus-private/issues/2283.


plutus-metatheory/src/Algorithmic/ReductionEC/Determinism.lagda.md
```diff
@@ -21,7 +21,7 @@ open import Relation.Binary.PropositionalEquality
 open import Relation.Binary.HeterogeneousEquality
         using (_≅_;refl;≅-to-≡)
 
-open import Utils hiding (_×_;List)
+open import Utils hiding (_×_;List;Value)
```
ana-pantilie, line 24: Is this import necessary given it's not used anywhere?

plutus-metatheory/src/Algorithmic/ReductionEC.lagda.md
```diff
@@ -29,7 +29,7 @@ open import Relation.Binary.PropositionalEquality
 open import Relation.Binary.HeterogeneousEquality
         using (_≅_;refl;≅-to-≡)
 
-open import Utils hiding (List;length;map)
+open import Utils hiding (List;length;map;Value)
```
ana-pantilie, line 32: Same Q as above.

plutus-metatheory/src/Cost/Model.lagda.md
```diff
@@ -189,38 +189,79 @@ runModel (literalCostIn n m) xs with lookup xs n
...
+  3 → just (twoArgumentsLinearInYAndZ intercept slope1 slope2)
+  _ → nothing
+convertRawModel (LinearInY2 (mkLF intercept slope) _) = λ where
+  (suc (suc n)) → just (linearCostIn (suc zero) intercept slope)
+  _ → nothing
+convertRawModel (LinearInMaxYZ (mkLF intercept slope)) = λ where
+  3 → just (twoArgumentsLinearInMaxYZ intercept slope)
+  _ → nothing
+convertRawModel (QuadraticInX (mkQF1 c0 c1 c2)) = λ where
+  (suc n) → just (quadraticCostIn1 zero c0 c1 c2)
```
ana-pantilie, line 200: In the original, this case matches on `suc (suc n)` while the new version matches on `suc n`. Is there a reason for this change?

```suggestion
  suc (suc n) → just (quadraticCostIn1 zero c0 c1 c2)
```

plutus-metatheory/src/Raw.lagda.md
```diff
@@ -51,7 +51,7 @@ data RawTy where
 data RawTyCon where
   atomic     : AtomicTyCon → RawTyCon
   list       : RawTyCon
-  array       : RawTyCon
+  array      : RawTyCon
```
ana-pantilie, line 54: Why is `value` not added as well?

ana-pantilie (changes requested):
Mostly looks good, but I'm concerned about the budget changes and there's also one potentially problematic change I've pointed out in the review.

Could the budget changes be explained by the fact that they were marked as "failing" and they weren't checked by CI? I know we've made multiple changes to the cost model between first adding the tests and this PR. Or did something in this PR trigger the modifications?

plutus-metatheory/src/Algorithmic/ReductionEC/Determinism.lagda.md
basetunnel: It clashes with `Value` from the CEK machine, which is used in this module.

plutus-metatheory/src/Algorithmic/ReductionEC.lagda.md
basetunnel: See answer above.

plutus-metatheory/src/Cost/Model.lagda.md
basetunnel: Sorry, I forgot to mention that on the PR. The original `suc (suc n)` was too restrictive. It requires at least two arguments while this model can be used for single-argument functions, which is the case for `unValueData`. (https://github.com/IntersectMBO/plutus/blob/master/plutus-core/cost-model/data/models.R#L868).

Without it, most conformance tests were failing because the overall cost model couldn't be constructed properly.

Similarly, I added a clause for `LinearInXAndY`, which is needed for `valueContains`. That made the overall model construction fail because it defaulted to `nothing` here.

plutus-metatheory/src/Raw.lagda.md
basetunnel: It is part of the `AtomicTyCon` (for base-kinded builtin types).

basetunnel:
>  but I'm concerned about the budget changes

> Could the budget changes be explained by the fact that they were marked as "failing" and they weren't checked by CI? I know we've made multiple changes to the cost model between first adding the tests and this PR. 

That's what I assumed too, I'll try to see if I can verify that.

basetunnel:
> > but I'm concerned about the budget changes
> 
> > Could the budget changes be explained by the fact that they were marked as "failing" and they weren't checked by CI? I know we've made multiple changes to the cost model between first adding the tests and this PR.
> 
> That's what I assumed too, I'll try to see if I can verify that.

Thanks for spotting this, the budgets were indeed not supposed to change. The costing in Agda was off, because costing `Value` builtins sometimes requires different measure functions for their arguments. I've postulated those measures and refactored `runModel` a bit:
- `runModel` now takes a vector of argument sizes, rather than a vector of argument values. Computing those sizes now happens in `builtinCost`, because that function still knows what builtin we're costing and can therefore choose the right measures.
- `integerToByteString` also uses a custom measure on the third argument, which was previously done in an ad-hoc way in `runModel`. Now it's just another measure function.

Now the costing matches again :)

ana-pantilie (approved):

kwxm:
@basetunnel I see that adding this allowed us to enable [some of the conformance tests](https://github.com/IntersectMBO/plutus/pull/7832/changes#diff-f45de1f8469124f0ea199851c81019e7d31316362d89b63aef66d319b6d4a84a) that were previously turned off, but you didn't enable the ones for `valueData` and `unValueData`.  Those appear to work properly now though: is there some reason you didn't turn them on, or was it just an oversight?

basetunnel:
> @basetunnel I see that adding this allowed to enable [some of the conformance tests](https://github.com/IntersectMBO/plutus/pull/7832/changes#diff-f45de1f8469124f0ea199851c81019e7d31316362d89b63aef66d319b6d4a84a) that were previously turned off, but you didn't enable the ones for `valueData` and `unValueData`. Those appear to work properly now though: is there some reason you didn't turn them on, or was it just an oversight?

Thanks! I missed those indeed, I'll open a PR.

PR #7837, opened 2026-07-10
Refactor BLS12_381 conformance tests to remove literal curve points

kwxm:
We provide literal syntax for BLS12_381 constants in the textual format, but don't permit such constants in flat-encoded files because deserialisation is quite CPU-intensive.  Instead, if you want to use a constant of this type you have to deserialise it from a bytestring during script execution using an appropriate `uncompress` function so that the user pays for deserialisation cost from the script budget.  Using the textual format made it easy to write conformance tests, but tests using that feature couldn't be converted to `flat`, making it difficult for external users to use the tests.  This PR updates most of the conformance tests to perform explicit deserialisation (in preparation for adding `flat` files to the repository).  I haven't altered the tests in [the `constant` directory](https://github.com/IntersectMBO/plutus/tree/kwxm/conformance/rewrite-bls-constants/plutus-conformance/test-cases/uplc/evaluation/builtin/constant) because they're really tests of the parser, and in hindsight maybe shouldn't be included in the conformance tests; maybe we should move them elsewhere at some point.


kwxm:
~~Hold on: I missed some files.~~

Fixed now.

Unisay (approved):
Uniform, correct refactor. LGTM.

PR #7838, opened 2026-07-10
Add `unsafeLovelaceValueOf`: a positional lovelace accessor for ledger `Value`s

Unisay:
`unsafeLovelaceValueOf` assumes that the first token of the first currency in the Value represents lovelace. Returns a silently wrong answer or fails if that's not the case. It is thus much faster than `lovelaceValueOf`.

## Impact

Isolated budget goldens on a ledger-shaped value (ada first, then three native assets), the same input fed to both functions:

| | Data `lovelaceValueOf` | Data `unsafeLovelaceValueOf` | delta |
|---|---:|---:|---:|
| CPU | 1,295,411 | 595,767 | -54.0% |
| Memory | 7,430 | 2,628 | -64.6% |

| | Scott `lovelaceValueOf` | Scott `unsafeLovelaceValueOf` | delta |
|---|---:|---:|---:|
| CPU | 1,657,760 | 1,232,100 | -25.7% |
| Memory | 10,102 | 7,800 | -22.8% |

Both variants evaluate to the same result. Because both already short-circuit on the ada-first entry, the delta is purely the removed key extraction, comparisons, and lookup recursion, so it is a fixed saving independent of how many other assets the value carries. A validator that reads lovelace on every input and every output, a common shape, compounds it.

## Coverage

A property test in both `Spec.V1.Value` and `Spec.V1.Data.Value` checks that on a canonical ledger-shaped value `unsafeLovelaceValueOf` agrees with `lovelaceValueOf` and returns the exact lovelace amount. The budget numbers above come from new `goldenEvalCekCatchBudget` goldens; I also added PIR-readable goldens so the structural difference is visible in review. Goldens are committed for GHC 9.6 and 9.12.

Implements IntersectMBO/plutus-private#2304.

### DIFF
--- a/plutus-ledger-api/changelog.d/20260709_120000_yuriy.lazaryev_issue_2304_positional_lovelace.md
+++ b/plutus-ledger-api/changelog.d/20260709_120000_yuriy.lazaryev_issue_2304_positional_lovelace.md
@@ -1,3 +1,3 @@
 ### Added
 
-- `unsafeLovelaceValueOf`, in both `PlutusLedgerApi.V1.Value` and `PlutusLedgerApi.V1.Data.Value`: a faster, positional way to read the lovelace quantity of a ledger-provided `Value`. It returns the first amount of the first currency, skipping the ada key lookup and comparisons that `lovelaceValueOf` performs. This is sound only for `Value`s that are canonical as the ledger produces them — ada present and sorted first, as in a `TxOut` value — so it is not a drop-in replacement for `lovelaceValueOf` on mint values or user-constructed `Value`s, where ada may be absent or not first. See its Haddock for the precondition.
+- `unsafeLovelaceValueOf` in `PlutusLedgerApi.V1.Value` and `PlutusLedgerApi.V1.Data.Value`: assumes the first amount of the first currency represents lovelace without verifying this, which makes it much faster than `lovelaceValueOf`. It is sound only when this is true.
--- a/plutus-ledger-api/changelog.d/20260709_120000_yuriy.lazaryev_issue_2304_positional_lovelace.md
+++ b/plutus-ledger-api/changelog.d/20260709_120000_yuriy.lazaryev_issue_2304_positional_lovelace.md
@@ -1,0 +1,3 @@
+### Added
+
+- `unsafeLovelaceValueOf`, in both `PlutusLedgerApi.V1.Value` and `PlutusLedgerApi.V1.Data.Value`: a faster, positional way to read the lovelace quantity of a ledger-provided `Value`. It returns the first amount of the first currency, skipping the ada key lookup and comparisons that `lovelaceValueOf` performs. This is sound only for `Value`s that are canonical as the ledger produces them — ada present and sorted first, as in a `TxOut` value — so it is not a drop-in replacement for `lovelaceValueOf` on mint values or user-constructed `Value`s, where ada may be absent or not first. See its Haddock for the precondition.
--- a/plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs
@@ -49,6 +49,7 @@ module PlutusLedgerApi.V1.Data.Value
   , currencySymbolValueOf
   , lovelaceValue
   , lovelaceValueOf
+  , unsafeLovelaceValueOf
   , scale
   , symbols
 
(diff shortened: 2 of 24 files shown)


plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs
```diff
@@ -392,6 +393,31 @@ lovelaceValueOf :: Value -> Lovelace
...
+first currency, so it can be read as @snd . head@ twice — skipping the two
+'equalsByteString' comparisons and the key unwrapping that 'lovelaceValueOf'
+performs on the way.
+
+PRECONDITION: the 'Value' is canonical as produced by the ledger, i.e. ada is
+present and sorts first. This holds for 'TxOut' values but /not/ for mint values
+or user-constructed 'Value's, where ada may be absent or not sorted first.
+Applied to such a 'Value' this returns the first amount of whichever currency
+happens to be first (a silently wrong answer), or fails on an empty 'Value'.
+Use 'lovelaceValueOf' whenever the invariant is not guaranteed. -}
```
zliu41: Shorten. "Assumes that the first token of the first currency in the Value represents lovelace. Returns a silently wrong answer or fails if that's not the case."

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
```diff
@@ -396,6 +397,28 @@ lovelaceValueOf :: Value -> Lovelace
...
+(min-ada). The lovelace quantity is therefore exactly the first amount of the
+first currency, so it can be read as @snd . head@ twice — skipping the two
+key comparisons that 'lovelaceValueOf' performs via 'Map.lookup' on the way.
+
+PRECONDITION: the 'Value' is canonical as produced by the ledger, i.e. ada is
+present and sorts first. This holds for 'TxOut' values but /not/ for mint values
+or user-constructed 'Value's, where ada may be absent or not sorted first.
+Applied to such a 'Value' this returns the first amount of whichever currency
+happens to be first (a silently wrong answer), or fails on an empty 'Value'.
+Use 'lovelaceValueOf' whenever the invariant is not guaranteed. -}
```
zliu41: ditto

plutus-ledger-api/changelog.d/20260709_120000_yuriy.lazaryev_issue_2304_positional_lovelace.md
```diff
@@ -0,0 +1,3 @@
+### Added
+
+- `unsafeLovelaceValueOf`, in both `PlutusLedgerApi.V1.Value` and `PlutusLedgerApi.V1.Data.Value`: a faster, positional way to read the lovelace quantity of a ledger-provided `Value`. It returns the first amount of the first currency, skipping the ada key lookup and comparisons that `lovelaceValueOf` performs. This is sound only for `Value`s that are canonical as the ledger produces them — ada present and sorted first, as in a `TxOut` value — so it is not a drop-in replacement for `lovelaceValueOf` on mint values or user-constructed `Value`s, where ada may be absent or not first. See its Haddock for the precondition.
```
zliu41: Shorten: `unsafeLovelaceValueOf` assumes the first amount of the first currency represents lovelace without verifying this, which makes it much faster than `lovelaceValueOf`. It is sound only when this is true.

zliu41 (commented):
Was this requested or suggested by someone or was it your own idea?

I shortened the PR description for you. Your version used too many words to say something that can be said very simply.

Unisay:
> Was this requested or suggested by someone or was it your own idea?

The idea I got while optimizing UPLC-CAPE validators and comparing Plinth/Scalus implementations. I saw that "unsafe" Lovelace retrieval gives quite some ex budget boost, especially when its done multiple times (for different tx inputs).

plutus-ledger-api/src/PlutusLedgerApi/V1/Data/Value.hs
Unisay: Done.

plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs
Unisay: Done.

plutus-ledger-api/changelog.d/20260709_120000_yuriy.lazaryev_issue_2304_positional_lovelace.md
Unisay: Done.

zliu41 (approved):
> The idea I got while optimizing UPLC-CAPE validators and comparing Plinth/Scalus implementations. I saw that "unsafe" Lovelace retrieval gives quite some ex budget boost, especially when its done multiple times (for different tx inputs).

Very nice.

PR #7839, opened 2026-07-10
Re-export the full `Value` API from V2/V3 and the `Data.*` modules

Unisay:
`PlutusLedgerApi.V2`, `PlutusLedgerApi.V3`, and `PlutusLedgerApi.Data.V1`/`V2`/`V3` now re-export the full `Value` API that `PlutusLedgerApi.V1` already exposed, and all six umbrellas (including `V1`) additionally expose `geq`, `gt`, `leq`, `lt`, and `withCurrencySymbol`, which previously lived only in `PlutusLedgerApi.V1.Value`/`.Data.Value`. Reading or comparing a `Value` from a version umbrella no longer needs a direct `PlutusLedgerApi.V1.Value` or `PlutusLedgerApi.V1.Data.Value` import. Every V2/V3 module in this repo that reached into those V1 modules now uses the umbrella instead; the remaining direct importers are genuine V1 code (the V1 doc examples, the `script-contexts` V1 variant) and the V1 `Value` test suites.


zliu41 (approved):
There are more cleanups to be done, e.g., the `PlutusLedgerApi.V1.Value` import [here](https://github.com/IntersectMBO/plinth-template/blob/main/src/AuctionValidator.hs) should be replaced by `PlutusLedgerApi.V3`, since it is a V3 script.

Unisay:
Done — swept the repo for the same pattern. The modules that still needed `withCurrencySymbol`/`geq`/`leq` were the reason those names weren't re-exported at all, so I added `geq`/`gt`/`leq`/`lt`/`withCurrencySymbol` to the `V1`/`V2`/`V3` and `Data.V1`/`V2`/`V3` umbrellas too; now the only direct `PlutusLedgerApi.V1.Value` importers left are genuine V1 code and the V1 `Value` tests. plinth-template: IntersectMBO/plinth-template#80 (draft, blocked on this PR merging plus a plutus bump there).

PR #7840, opened 2026-07-10
Fix termination and performance issues with DecEq

ana-pantilie:
Needs https://github.com/IntersectMBO/plutus/pull/7832
Needs https://github.com/IntersectMBO/plutus/pull/7843

## Termination fix

Removes the need for the `TERMINATING` pragma in the `decEq` definitions, Agda can now prove that they terminate.

## Performance fix

The matching on `refl` inside `decEq` triggered normalization of terms when typechecking, which affected performance significantly, to the point of using more than 128GB of memory for the example @basetunnel and I worked on.

The example is an equality proof of the first AST to itself, from `agda-certificates/cardano_constitution_Cardano_Constitution_Validator_Data_Sorted-b48sy0.agda-cert`. This was the certificate produced for `defaultConstitutionCode` from `cardano-constitution/src/Cardano/Constitution/Validator/Data/Sorted.hs`.

For this particular example comparing `con`s was the most expensive, so avoiding matching on `refl` in the recursive call to `decEq-⟦ t ⟧tag x x₁` inside `decEq-TmCon` did the trick.


plutus-metatheory/src/Untyped/Equality.lagda.md
```diff
@@ -163,16 +165,24 @@ listDec _ U.[] (x U.∷ ls₂) = no (λ ())
...
+...     | yes q = yes (cong₂ U._∷_ p q)
 
 pairDec : {A B : Set} → DecidableEquality A → DecidableEquality B → DecidableEquality (A × B)
 pairDec eqA eqB (a₁ , b₁) (a₂ , b₂) with (eqA a₁ a₂) | (eqB b₁ b₂)
-... | yes refl   | yes refl = yes refl
+... | yes p   | yes q = yes (cong₂ U._,_ p q)
 ... | no a₁≠a₂ | _ = no λ { refl → a₁≠a₂ refl }
 ... | _             | no b₁≠b₂ = no λ { refl → b₁≠b₂ refl }
 
+decEq-⊢⋆ [] [] = yes refl
```
basetunnel: The unicode character `⋆` seems to be used as a suffix for things that refer plutus types, like `_⊢⋆_` and `Ctx⋆` (probably referencing base kind). Perhaps we could have a different convention for a suffix for these map-like functions that appease the termination checker. I"ve used `*` in some of the translation relation machinery, but I think historically it was more often `List`, e.g. `dischargeList`.

plutus-metatheory/src/Untyped/Equality.lagda.md
```diff
@@ -91,7 +91,9 @@ decEq-TmCon : DecidableEquality TmCon
 
 decEq-⟦_⟧tag : ( t : TyTag ) → DecidableEquality ⟦ t ⟧tag
 
-decEq-⊢ : ∀{n} → DecidableEquality (n ⊢)
+decEq-⊢ : {n : ℕ} → DecidableEquality (n ⊢)
+
+decEq-⊢⋆ : {n : ℕ} → DecidableEquality (List (n ⊢)) -- {n : ℕ} → (Ms : List (n ⊢)) → (Ns : List (n ⊢)) → Dec (Ms ≡ Ns)
```
basetunnel: Is this comment for documentation purposes or a left-over from writing out DecidableEquality?

plutus-metatheory/src/Untyped/Equality.lagda.md
```diff
@@ -228,7 +238,7 @@ fromDec = record { hsEq = λ x₁ x₂ → isYes (x₁ ≟ x₂) }
 instance
   HsEqBytestring : HsEq U.ByteString
   HsEqBytestring = record { hsEq = U.eqByteString }
-  HsEqArray : {A : Set} {{HE : HasEq A}} {{HS : HsEq A}} → HsEq (U.Array A)
```
basetunnel, line 231: Why are these instance arguments not necessary anymore?

ana-pantilie: They weren't necessary in the first place, because the Agda array type is postulated, so the implementation of `eqArray` doesn't actually check the equality of the terms inside the array.

basetunnel (approved):

PR #7844, opened 2026-07-14
Add multiIndexArray builtin (CIP-0156)

Unisay:
Adds the `multiIndexArray` builtin from [CIP-0156](https://cips.cardano.org/cip/CIP-0156): `forall a. list integer -> array a -> list a`, returning the array elements at the given indices in index-list order (duplicates preserved) and failing the whole call on any out-of-bounds index. Bounds are checked in the `Integer` domain, so a negative index or one exceeding `maxBound::Int` is out-of-bounds rather than wrapping.

Placeholder costing (`unimplementedCostingFun`) and gated under `futurePV`, so it is not available in any released protocol version.

Core language + CEK only; Plinth exposure, costing, conformance, and metatheory are separate follow-ups.


plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -2471,6 +2473,23 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
      in makeBuiltinMeaning
           scaleValueDenotation
           (runCostingFunTwoArguments . paramScaleValue)
+  toBuiltinMeaning _semvar MultiIndexArray =
+    let multiIndexArrayDenotation
+          :: [Integer] -> SomeConstant uni (Vector a) -> BuiltinResult (Opaque val [a])
```
zliu41: 1. Need an explanation of why it uses Integer while indexArray uses Int
2. The argument order is different from indexArray's and I see no good reason for it.

zliu41 (changes requested):

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
Unisay: 1. [`indexArray`](https://github.com/IntersectMBO/plutus/blob/3fd47e185b6e8d0ab4a8eeb655ca440eac68dbf9/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs#L2388-L2401) takes a scalar `Int`, which has a [checked unlifting instance](https://github.com/IntersectMBO/plutus/blob/3fd47e185b6e8d0ab4a8eeb655ca440eac68dbf9/plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs#L583-L585): it [reads an `Integer` and bounds-checks it against `Int`'s range](https://github.com/IntersectMBO/plutus/blob/3fd47e185b6e8d0ab4a8eeb655ca440eac68dbf9/plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs#L553-L574), failing operationally on overflow. There's no equivalent for `[Int]`: list unlifting requires the element type to be in `DefaultUni`, and `Int` isn't there, only `Integer` is (same reason [`Bls12_381_G1_multiScalarMul`](https://github.com/IntersectMBO/plutus/blob/3fd47e185b6e8d0ab4a8eeb655ca440eac68dbf9/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs#L2402-L2409) takes `[Integer]`). So the indices arrive as `[Integer]`, and the `0 <= i && i < len` guard does in the `Integer` domain what the `Int` unlifting plus `Vector.!?` do for `indexArray`, additionally treating an index above `maxBound :: Int` as out-of-bounds rather than overflowing on conversion. Added an in-code comment noting this (282bd5a9).

2. Agreed, switching to `Array a -> List Integer -> List a` to match `indexArray`. The CIP is being amended to match (cardano-foundation/CIPs#1234); the implementation update follows here and in #7845 (Plinth).

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
```diff
@@ -2471,6 +2473,26 @@ instance uni ~ DefaultUni => ToBuiltinMeaning uni DefaultFun where
      in makeBuiltinMeaning
           scaleValueDenotation
           (runCostingFunTwoArguments . paramScaleValue)
+  toBuiltinMeaning _semvar MultiIndexArray =
+    -- Indices are 'Integer', not 'Int' (unlike 'indexArray'), because '[Int]' has no
+    -- unlifting: only 'DefaultUni' element types unlift, and 'Int' is not one. The bounds
+    -- check is therefore done in the 'Integer' domain.
```
zliu41: You should mention that `Int` works because it has a special ReadKnown instance; we could technically write a special instance for `[Int]` as well, but it makes no sense to do that (at least for multiIndexArray) since it just does an extra traversal plus allocations, and buys us nothing.

zliu41 (approved):

plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs
Unisay: Done

PR #7845, opened 2026-07-15
Expose multiIndexArray to Plinth (CIP-0156)

Unisay:
Exposes the `multiIndexArray` builtin ([CIP-0156](https://cips.cardano.org/cip/CIP-0156)) to Plinth: adds the `PlutusTx.Builtins.multiIndexArray` Haskell implementation and the plugin mapping that replaces the no-op arm left by #7844. It mirrors `indexArray`: indices out of range fail, in index-list order with duplicates preserved.

Closes https://github.com/IntersectMBO/plutus-private/issues/2265


plutus-tx/src/PlutusTx/Builtins/Internal.hs
```diff
@@ -675,6 +675,13 @@ indexArray :: BuiltinArray a -> BuiltinInteger -> a
 indexArray (BuiltinArray v) i = v Vector.! fromInteger i
 {-# OPAQUE indexArray #-}
 
+{-| Returns the elements at the given indices, in index-list order with duplicates preserved.
+  Fails if any index is not in the range @[0..j)@, where @j@ is the length of the array. -}
+multiIndexArray :: BuiltinList BuiltinInteger -> BuiltinArray a -> BuiltinList a
+multiIndexArray (BuiltinList is) (BuiltinArray v) =
+  BuiltinList (map (\i -> v Vector.! fromInteger i) is)
```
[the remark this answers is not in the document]
Unisay: Fixed in 8301f47f1c: the wrapper now traverses the index list eagerly with the bounds check done in the Integer domain, mirroring the builtin denotation.

plutus-tx/test/Array/Spec.hs
```diff
@@ -0,0 +1,44 @@
...
+import Control.Exception (ErrorCall)
+import Data.Either (isLeft)
+import PlutusCore.Test (pureTry)
+import PlutusTx.Builtins.Internal qualified as BI
+
+import Test.Tasty (TestTree, testGroup)
+import Test.Tasty.HUnit (Assertion, assertBool, testCase, (@?=))
+import Prelude
+
+-- The wrapper must fail exactly where the on-chain builtin fails: any index outside
```
zliu41: I don't think "wrapper" is standard terminology. It took me a few seconds to realize what it means.

zliu41 (approved):

plutus-tx/test/Array/Spec.hs
Unisay: Gotcha. How would you call those "opaque" Plinth functions?

zliu41: "Haskell implementation of the builtin"?

Issue #7342, opened 2025-09-18
Define PlutusV4 script context


lehins:
This ticket is created to track progress on PlutusV4 context that will include at least two  CIPs:
* [CIP-112 - Observe script type](https://github.com/cardano-foundation/CIPs/pull/749)
* [CIP-118 - Nested Transactions](https://github.com/cardano-foundation/CIPs/pull/862)

Considering that CIP-112 will be required for CIP-118 it makes sense to start with it first.

* [ ] - Define `PlutusV4` version in `plutus-ledger-api`: https://github.com/IntersectMBO/plutus/blob/fa5c9e64ec395661eff8f6a5ce3d784e1100833c/plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs#L89-L92
* [ ] - Create `V4` in `plutus-ledger-api:PlutusLedgerApi` that reuses as much stuff from V1-V3 as possible
* [ ] - Define `TxInfo` for PlutusV4
* [ ] - Add `txInfoGuards :: [Credential]` to `TxInfo` for PlutusV4 field and remove `txInfoSignatories`, since former supersedes the latter (Note that we will be using the term Guards in the implementation instead of Observers as CIP-112. CC @colll78 if you are still up to renaming the CIP to avoid confusion)
* [ ] - TBD: CIP-118 related parts added to TxInfo.

I'd recommend creating a child ticket for each of the bullet points, so the progress can be tracked at more granular level, but I'll leave this up to the Plutus team.

zliu41:
Two specific things to note
1. We'll add two new constructors to `Data` - Array and Value, and they will be used in Plutus V4 script context. The script context for V1-V3 will of course remain the same.
2. We are not yet sure whether we can implement https://cips.cardano.org/cip/CIP-0152 in time for Dijkstra. If we can, then that will also impact the script context.

colll78:
I think a clear non-controversial change is to add a `ownScriptHash` field to the `ScriptContext`. 


```haskell
-- | The context that the currently-executing script can access.
data ScriptContext = ScriptContext
  { scriptContextTxInfo     :: TxInfo
  -- ^ information about the transaction the currently-executing script is included in
  , scriptContextRedeemer   :: V2.Redeemer
  -- ^ Redeemer for the currently-executing script
  , scriptContextScriptInfo :: ScriptInfo
  -- ^ the purpose of the currently-executing script, along with information associated
  -- with the purpose
[...]

PR #7846, opened 2026-07-15
Plutus V4 ledger api types as per #7342

SeungheonOh:
Adds Plutus Ledger Api types for PlutusV4 as per #7342

ledger API types are not fully finalized yet

### DIFF
--- a/plutus-ledger-api/src/PlutusLedgerApi/V4/Contexts.hs
+++ b/plutus-ledger-api/src/PlutusLedgerApi/V4/Contexts.hs
@@ -127,6 +127,7 @@ data AccountBalanceInterval
   = AccountBalanceLowerBound V2.Lovelace
   | AccountBalanceUpperBound V2.Lovelace
   | AccountBalanceBothBounds V2.Lovelace V2.Lovelace
+  | AccountBalanceExact V2.Lovelace
   deriving stock (Generic, Haskell.Show, Haskell.Eq, Haskell.Ord)
   deriving anyclass (HasBlueprintDefinition)
   deriving (Pretty) via (PrettyShow AccountBalanceInterval)
@@ -139,6 +140,7 @@ $( makeIsDataSchemaIndexed
      [ ('AccountBalanceLowerBound, 0)
      , ('AccountBalanceUpperBound, 1)
      , ('AccountBalanceBothBounds, 2)
+     , ('AccountBalanceExact, 3)
      ]
  )
 
@@ -204,8 +206,8 @@ data TxInfo = TxInfo
   , txInfoFee :: V2.Lovelace
   , txInfoMint :: V3.MintValue
   , txInfoTxCerts :: [TxCert]
-  , txInfoWithdrawals :: Map V2.Credential V2.Lovelace
-  , txInfoDirectDeposits :: Map V2.Credential V2.Lovelace
+  , txInfoWithdrawals :: Map AccountId V2.Lovelace
+  , txInfoDirectDeposits :: Map AccountId V2.Lovelace
   , txInfoAccountBalanceIntervals :: AccountBalanceIntervals
   , txInfoValidRange :: V2.POSIXTimeRange
   , txInfoGuards :: [V2.Credential]
@@ -253,11 +255,10 @@ data TopTxInfoSimplified = TopTxInfoSimplified
   , ttisMints :: V3.MintValue
   , ttisBurns :: V3.MintValue
   , ttisTxCerts :: [TxCert]
-  , ttisWithdrawals :: Map V2.Credential V2.Lovelace
-  , ttisDirectDeposits :: Map V2.Credential V2.Lovelace
+  , ttisWithdrawals :: Map AccountId V2.Lovelace
+  , ttisDirectDeposits :: Map AccountId V2.Lovelace
   , ttisValidRange :: V2.POSIXTimeRange
-  , ttisGuards :: [V2.Credential]
-  , ttisRequiredTopLevelGuards :: Map V2.Credential ()
+  , ttisGuards :: Map V2.Credential ()
   , ttisScriptPurposes :: Map ScriptPurpose ()
   , ttisData :: Map V2.DatumHash V2.Datum
   , ttisVotes :: Map Voter (Map GovernanceActionId Vote)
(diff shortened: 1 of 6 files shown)


plutus-ledger-api/src/PlutusLedgerApi/V4/Contexts.hs
```diff
@@ -0,0 +1,454 @@
...
+  => HasBlueprintSchema AccountId referencedTypes
+  where
+  schema =
+    schema @V2.Credential @referencedTypes
+      & withSchemaInfo \info -> info {title = Haskell.Just "AccountId"}
+
+data AccountBalanceInterval
+  = AccountBalanceLowerBound V2.Lovelace
+  | AccountBalanceUpperBound V2.Lovelace
+  | AccountBalanceBothBounds V2.Lovelace V2.Lovelace
```
lehins, line 129: Let's add one more case, which will be a nice optimization, so instead of specifying `AccountBalanceBothBounds x (x+1)`, one can simply say `AccountBalanceExact x`. This seem s like can be a common case and it will be better for Tx serialization size and probably for DApp execution too.
We already have it in the works on the ledger side: https://github.com/IntersectMBO/cardano-ledger/issues/5922
```suggestion
  | AccountBalanceBothBounds V2.Lovelace V2.Lovelace
  | AccountBalanceExact V2.Lovelace
```

plutus-ledger-api/src/PlutusLedgerApi/V4/Contexts.hs
```diff
@@ -0,0 +1,454 @@
...
+  , ttisCurrentTreasuryAmount :: Haskell.Maybe V2.Lovelace
+  , ttisTreasuryDonations :: V2.Lovelace
+  }
+  deriving stock (Generic, Haskell.Show, Haskell.Eq)
+  deriving anyclass (HasBlueprintDefinition)
+  deriving (Pretty) via (PrettyShow TopTxInfoSimplified)
+
+data TopTxInfo = TopTxInfo
+  { topTxInfoSubTransactions :: [TxInfo]
+  , topTxInfoDatums :: Map V3.TxId V2.Datum
```
lehins: Question about this field.
Considering that `topTxInfoSubTransactions` will contain the `TxId`, getting it would be as simple as: 
```haskell
getTxId txIx = txInfoId (topTxInfoSubTransactions txInfos !! txIx)
```
Finding an index is also possible, but arguably is more expensive:
```haskell
getTxIx txId = fromJust $ txInfoSubTxIx =<< find ((== txId) . txInfoId) (topTxInfoSubTransactions txInfos)
```
So, the question is, would it be better to supply the index of sub transaction instead of `TxId` in this Map?

```suggestion
  , topTxInfoDatums :: Map Haskell.Integer V2.Datum
```

plutus-ledger-api/src/PlutusLedgerApi/V4/Contexts.hs
```diff
@@ -0,0 +1,454 @@
...
+  { txInfoId :: V3.TxId
+  , txInfoSubTxIx :: Haskell.Maybe Haskell.Integer
+  , txInfoInputs :: [TxInInfo]
+  , txInfoReferenceInputs :: [TxInInfo]
+  , txInfoOutputs :: [V2.TxOut]
+  , txInfoFee :: V2.Lovelace
+  , txInfoMint :: V3.MintValue
+  , txInfoTxCerts :: [TxCert]
+  , txInfoWithdrawals :: Map V2.Credential V2.Lovelace
+  , txInfoDirectDeposits :: Map V2.Credential V2.Lovelace
```
lehins: Let's change these to `AccountId`
```suggestion
  , txInfoWithdrawals :: Map AccountId V2.Lovelace
  , txInfoDirectDeposits :: Map AccountId V2.Lovelace
```

plutus-ledger-api/src/PlutusLedgerApi/V4/Contexts.hs
```diff
@@ -0,0 +1,454 @@
...
+data TopTxInfoSimplified = TopTxInfoSimplified
+  { ttisIds :: [V3.TxId]
+  , ttisInputs :: [TxInInfo]
+  , ttisReferenceInputs :: [TxInInfo]
+  , ttisOutputs :: [V2.TxOut]
+  , ttisMints :: V3.MintValue
+  , ttisBurns :: V3.MintValue
+  , ttisTxCerts :: [TxCert]
+  , ttisWithdrawals :: Map V2.Credential V2.Lovelace
+  , ttisDirectDeposits :: Map V2.Credential V2.Lovelace
```
lehins: Ditto
```suggestion
  , ttisWithdrawals :: Map AccountId V2.Lovelace
  , ttisDirectDeposits :: Map AccountId V2.Lovelace
```

plutus-ledger-api/src/PlutusLedgerApi/V4/Contexts.hs
```diff
@@ -0,0 +1,454 @@
...
+  , ttisValidRange :: V2.POSIXTimeRange
+  , ttisGuards :: [V2.Credential]
+  , ttisRequiredTopLevelGuards :: Map V2.Credential ()
```
lehins: Actually, now looking at this, makes me realize there is no point in having `ttisRequiredTopLevelGuards`, since they all will be also present in the `ttisGuards` field. So, instead we'll turn `ttisGuards` into a Set and remove `ttisRequiredTopLevelGuards`
```suggestion
  , ttisGuards :: Map V2.Credential ()
```

plutus-ledger-api/src/PlutusLedgerApi/V4/Contexts.hs
```diff
@@ -0,0 +1,454 @@
...
+findContinuingOutputs :: ScriptContext -> [Haskell.Integer]
+findContinuingOutputs ctx
+  | Haskell.Just TxInInfo {txInInfoResolved = V2.TxOut {txOutAddress}} <- findOwnInput ctx =
+      List.findIndices
+        (\V2.TxOut {txOutAddress = otherAddress} -> txOutAddress PlutusTx.== otherAddress)
+        (txInfoOutputs (scriptContextTxInfo ctx))
+findContinuingOutputs _ = PlutusTx.traceError "Le"
+{-# INLINEABLE findContinuingOutputs #-}
+
+getContinuingOutputs :: ScriptContext -> [V2.TxOut]
```
lehins, line 356: It is important to point out that "ContinuingOutputs" functions only look into individual sub-transaction/top level transaction, not at the whole transaction, which is technically all that they can do for the "Spending" purpose

lehins (commented):
Looks great! Thank you!
Just a few last moment changes that I thought of during the review.

zliu41 (approved):
LGTM once Alexey's comments are incorporated.

plutus-ledger-api/src/PlutusLedgerApi/V4/Contexts.hs
SeungheonOh: I made this change, but this might be a good thing to benchmark and find out after merging this PR. cc @zliu41

PR #7850, opened 2026-07-16
Forgot to add test case for duplicate currency IDs in value

kwxm:
In #7816 I got the textual parser to reject literal values where the currency or token IDs weren't in strictly ascending order.  I added some conformance tests at the time but I seem to have forgotten to commit the one for duplicated currency IDs.  This PR adds it.


Unisay (approved):
👍🏼

zliu41 (approved):

PR #7853, opened 2026-07-16
Add flat files to conformance tests

kwxm:
This adds `.flat` and `.flat.expected` files for most of the conformance tests to make things easier for other users.

 * By default tests like `haskell-conformance` still use the textual format; however you can use the `--format flat` option to get them to use the `.flat` files instead.
 * Some of the test cases (eg, for parse errors) don't have flat equivalents.  There are no flat files for these.
 * I've changed the extension for the files containing expected budgets to `.budget.expected` (it was previously `.uplc.budget.expected`, but the flat files will require the same budgets).
 * I've renamed the `constant` directories to `parser` because those tests are really testing the parser and don't have any flat equivalent.
 * I've added a `conformance-consistency` test to check that every .uplc file has a corresponding .flat file and vice versa (but see the next point) and that the uplc and flat files decode to identical ASTs (and similarly for the results).
 * `conformance-consistency` contains a [list of directories/files](https://github.com/IntersectMBO/plutus/blob/b1ed67743ae71210a700101beca5a2d1e6171518/plutus-conformance/consistency/Spec.hs#L24) to skip to deal with the .uplc files that don't have corresponding .flat files.
 * It may in fact be possible to convert some of the files in `test-cases/uplc/evaluation/builtin/parser` (in particular, some of the `value` ones) to flat, but that'll have to be done by manufacturing the flat encodings directly (ie, not using `uplc convert`, since it'll fail to parse them).  This will be a bit tricky, so I'll come back to it in a separate issue.
 * I've had to exclude the test in `test-cases/uplc/evaluation/term/var` because of some problems with free variables: I'll come back and look at this later.

~If something goes wrong during execution of a `.flat` file then the `.flat.expected` file is empty.  I'm undecided about whether this is a good idea: maybe it should just contain text "parse/decode" error or "evaluation failure" like the other two `.expected` files.~

If something goes wrong during execution of a `.flat` file then the `.flat.expected` file contains a textual message saying what went wrong, which should be identical to the contents of the `.uplc.expected` file.  This means that consumers of these files should check whether they contain a textual error message before attempting flat decoding.


plutus-conformance/README.md
```diff
@@ -11,71 +17,136 @@ This suite tests the latest version of Plutus. Testing of older versions may be
...
+The library provides functions that users can import and run conformance tests with their own implementation. At the moment the tests can only be run against another Haskell implementation. More support will be added later. Of course, one can wrap an arbitrary executable in Haskell. See an explanation [here](https://www.fpcomplete.com/blog/2017/02/typed-process/) and [the related documentation](https://www.stackage.org/haddock/lts-19.11/typed-process-0.2.10.1/System-Process-Typed.html).
 
-#### "parse error"
+### Expected outputs for textual `.uplc` files
 
-The input files are expected to have the concrete syntax. The expected output will show "parse error" when the parser fails to parse the input file.
+If a textual `.uplc` file is expected to fail to parse correctly then the `.uplc.expected` file will contain the string  "parse/decode error"
 
-#### "evaluation failure"
+If the file is expected to parse correctly but an error is expected to occur during evaluation then the `.uplc.expected` file will contain the string "parse/decode error".
```
zliu41: evaluation failure?

plutus-conformance/src/PlutusConformance/Common.hs
```diff
@@ -50,10 +88,43 @@ parseTxt resTxt = runQuoteT $ UPLC.parseProgram resTxt
...
+formatExtension Textual = "uplc"
+formatExtension Flat = "flat"
+
+{-| This instance allows `Format` to be used as a `tasty` command-line option
+(`--format=uplc` or `--format=flat`), so that users of the test suites can
+choose which input format the tests are run against.  The default is `Uplc`. -}
+instance IsOption Format where
+  defaultValue = Textual
+  parseValue s = case s of
+    "uplc" -> Just Textual
```
zliu41: Why "uplc" not "textual"? This is inconsistent with the help text below.

plutus-conformance/agda/Spec.hs
```diff
@@ -132,17 +132,17 @@ agdaEvalUplcProg WithoutCosting =
...
-              Left _ -> Nothing
+              Left _ -> EvalFailure
               Right tmEvaluated ->
                 case runQuote $
                   runExceptT $
                     withExceptT FreeVariableErrorE $
                       unDeBruijnTerm tmEvaluated of
-                  Left _ -> Nothing
-                  Right namedTerm -> Just $ UPLC.Program () version namedTerm
+                  Left _ -> EvalFailure -- Shouldn't happen unless there's something wrong with the Agda code.
```
zliu41: DecodeError?

plutus-conformance/src/PlutusConformance/Common.hs
kwxm: OK, I changed it.  That makes it consistent with the `uplc` executable (although there you use `--input-format` or `--if`` instead of `--format`), but inconsistent with the `.uplc` extension.  The default is `textual` anyway, so you never have to type it.

plutus-conformance/agda/Spec.hs
kwxm: Good question.  This is when the evaluation result is being converted back to flat, so it's really an encoding error rather than a decoding one.  I've changed it to `error` for compatibility with the [version in the Haskell tests](https://github.com/IntersectMBO/plutus/blob/e409b7ae087c8d978f609b9bbd828fed65284040/plutus-conformance/src/PlutusConformance/Common.hs#L589) and added some explanatory comments.

zliu41 (approved):

PR #7861, opened 2026-07-21
Reduce the number of MintValue/WithCurrencySymbol test cases

zliu41:
These tests are quite slow and cause pressure on CI resource. The functionalities tested are very simple and we don't need that many test cases.


Unisay (approved):
LGTM.

Did you consider configuring number of samples at a group level like [here](https://github.com/IntersectMBO/plutus/blob/af34bbb9d602aa0376b29e5dc223173868db3fb0/plutus-core/index-envs/test/RAList/Spec.hs#L140) ?

zliu41:
> Did you consider configuring number of samples at a group level like [here](https://github.com/IntersectMBO/plutus/blob/af34bbb9d602aa0376b29e5dc223173868db3fb0/plutus-core/index-envs/test/RAList/Spec.hs#L140) ?

Good point; updated.

PR #7862, opened 2026-07-22
Compile BuiltinData transparently in the plugin

Unisay:
Alternative to the unreachable-constructor workaround in #7719 (stacked on that branch, so the diff shows only the delta; retarget to master once #7719 merges or if it's superseded).

The plugin treats `BuiltinData` as representationally transparent: `PlutusCore.Data.Data` compiles to the builtin `data` type (same as `BuiltinData`), the `BuiltinData` constructor compiles to the identity, and a single-alternative case on it binds both the case binder and the field to the scrutinee. GHC unwrapping the wrapper (the #7716 join-point crash) then becomes harmless instead of fatal, so `BuiltinDataUnreachable` and its COMPLETE pragma are removed. Details in Note [Transparent BuiltinData] in `PlutusTx.Compiler.Expr`.

The GHC 9.12 budget goldens return exactly to master's numbers — the ~1% CPU / ~2.5% memory cost of blocking GHC's unboxing is gone. Both plugin test suites pass under GHC 9.6 and 9.12 (847 + 81).

One diagnostic change: `StageViolation/builtinData` now reports the unwrapped binding (`ipv … No unfolding, Type: PlutusCore.Data.Data`) instead of `validator :: BuiltinData`, because GHC unwraps the where-binding again; the stage-violation hint and remediation notes are unchanged.

### DIFF
--- a/plutus-tx-plugin/changelog.d/20260722_140000_yuriy.lazaryev_transparent_builtin_data.md
+++ b/plutus-tx-plugin/changelog.d/20260722_140000_yuriy.lazaryev_transparent_builtin_data.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- The plugin compiles `BuiltinData` transparently: `PlutusCore.Data.Data` maps
+  to the builtin `data` type and the `BuiltinData` constructor compiles to the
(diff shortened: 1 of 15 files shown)


Unisay:
Superseded — reopening as an independent PR against master rather than stacked on #7719.

PR #7864, opened 2026-07-22
Formalize integer division and CIntegers in the metatheory

ana-pantilie:
Fixes https://github.com/IntersectMBO/plutus-private/issues/2356


plutus-metatheory/src/Algorithmic/CEK.lagda.md
```diff
@@ -186,6 +189,11 @@ discharge (V-con c)  = con c refl
 discharge (V-I⇒ b bt) = dischargeB bt
 discharge (V-IΠ b bt) = dischargeB bt
 discharge (V-constr i Tss s refl) = constr i Tss refl (dischargeStack s)
+
+mkCInteger : ℤ → Either (∅ ⊢Nf⋆ *) CInteger
+mkCInteger i with CInt.minBound ≤? i | i ≤? CInt.maxBound
+mkCInteger i | yes p | yes q = return (cInt i p q)
+mkCInteger i | _ | _ = inj₁ (con (ne (^ (atomic aInteger))))
```
kwxm: `(con (ne (^ (atomic aInteger))))` shows up multiple times. Would it be worth defining some shorthand form, or is that against the spirit of Agda?

plutus-metatheory/src/Algorithmic/CEK.lagda.md
```diff
@@ -194,28 +202,48 @@ If a builtin returns a value, then this function produces a `Value`, otherwise i
...
+  i₂ ← mkCInteger i'
+  return (V-con (CInt.subtract i₁ i₂))
+BUILTIN multiplyInteger (base $ V-con i $ V-con i') = do
+  i₁ ← mkCInteger i
+  i₂ ← mkCInteger i'
+  return (V-con (CInt.multiply i₁ i₂))
+BUILTIN divideInteger (base $ V-con i $ V-con i') = do
+  i₁ ← mkCInteger i
+  i₂ ← mkCInteger i'
+  i₁/i₂ ← maybeToEither (con (ne (^ (atomic aInteger)))) (CInt.div i₁ i₂)
```
kwxm: Maybe using a name like that looks like an operation could lead to confusion, especially where `i₁%i₂` means two different things, both different from Agda's `i₁ % i₂`.  Just call the results `r` or something?  I don't really mind all that much though.

plutus-metatheory/src/Untyped/CEK.lagda.md
```diff
@@ -146,61 +149,86 @@ V-I b {tm = suc tm} bt = V-IΠ b bt
...
   }
 BUILTIN divideInteger = λ
-  { (app (app base (V-con integer i)) (V-con integer i')) -> decIf
-      (i' ≟ ℤ.pos 0)
-      (inj₁ userError)
-      (inj₂ (V-con integer (div i i')))
+  { (app (app base (V-con integer i)) (V-con integer i')) -> do
+      i₁ ← mkCInteger i
+      i₂ ← mkCInteger i'
+      i₁/i₂ ← maybeToEither userError (CInt.div i₁ i₂)
```
kwxm, line 187: Same comment about names as for `Algorithmic/CEK.lagda.md`.

plutus-metatheory/src/Builtin/Integer/Base.lagda.md
```diff
@@ -0,0 +1,55 @@
...
+open import Data.Integer
+import Data.Nat as ℕ
+import Data.Sign as S
+open import Data.Product.Base using (_×_; _,_; proj₁; proj₂)
+```
+
+## Quotient and remainder
+
+The `quot` and `rem` functions are based on the Haskell `quot` and `rem` functions, which perform truncated division.
+This follows the implementation of Haskell's `integerQuotRem#`.
```
kwxm, line 28: For this to do the right thing it's important here that Agda's `/` and `%` behave identically to Haskell's `quot` and `rem` for positive arguments.  Can we safely assume that this is true, or should we try to test it somehow?  Maybe not:  the proved properties later are helpful, and the conformance tests give us some extra assurance, and we're planning to convert more extensive Haskell property tests like [these](https://github.com/IntersectMBO/plutus/blob/master/plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Integer/QuotRemProperties.hs) into conformance tests, which should make things a lot better.

plutus-metatheory/src/Builtin/Integer/Base.lagda.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+## Quotient and remainder
+
+The `quot` and `rem` functions are based on the Haskell `quot` and `rem` functions, which perform truncated division.
+This follows the implementation of Haskell's `integerQuotRem#`. 
+
+```
+quot : (n d : ℤ) .{{_ : NonZero d}} → ℤ
+quot n d = ((sign n S.* sign d) ◃ (∣ n ∣ ℕ./ ∣ d ∣))
```
kwxm: I worried for a bit about what `sign 0` is.  It's `+` though (unlike Haskell's `signum`, which is 0) , so I think this is OK.

plutus-metatheory/src/Builtin/Integer/Properties.lagda.md
```diff
@@ -0,0 +1,109 @@
...
+open ≡-Reasoning
+import Data.Sign as S
+import Data.Sign.Properties as SP
+import Data.Nat.Properties as ℕP
+open import Data.Integer.Properties
+open import Data.Nat.DivMod
+open import Data.Product.Base using (_×_; _,_; proj₁; proj₂)
+```
+
+## Properties
```
kwxm: I didn't check all of the details of the proofs, but if Agda says they're OK I'm willing to believe it.

plutus-metatheory/src/Builtin/Integer/Properties.lagda.md
kwxm, line 1: There are other properties one could prove here (like the ones about the signs of the function results and the sizes of remainders [here](https://github.com/IntersectMBO/plutus/blob/b9d726d7cc957fa154c6ba9f01959952887f1246/plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Integer/DivModProperties.hs#L64) and [here](https://github.com/IntersectMBO/plutus/blob/3e257708aea5705074ae5a9687e0d97d66a954f2/plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Integer/QuotRemProperties.hs#L83)) which would pin down the behaviour of these functions a bit more precisely (maybe completely?).  That might be overkill though, and also time-consuming.

plutus-metatheory/src/Builtin/CInteger.lagda.md
```diff
@@ -0,0 +1,99 @@
...
+
+```
+
+## The CInteger type
+
+The `CInteger` type is a restriction of the `ℤ` type to the range of integers specified by `minBound` and `maxBound`.
+
+This type constitutes the denotational semantics of the Cardano `BuiltinInteger` type for all of the inputs to the `BuiltinInteger` builtin functions, except `equalsInteger` and `expModInteger`.
+
+The inputs to `equalsInteger` are of the unrestricted `ℤ` type. The `expModInteger` function is not yet formalised and is left as future work.
```
kwxm, line 39: `expModInteger` is an interesting case.  We could implement it in Agda easily enough (exponentiate using [square-and-multiply](https://en.wikipedia.org/wiki/Exponentiation_by_squaring) and then use `%`), but that might be pretty slow.  On the other hand, the Haskell version uses `integerPowMod#`, which I think ultimately uses [this C code](https://github.com/gmp-mirror/gmp/blob/master/mpn/generic/powm.c) in GMP, and it'd be out of the question to formalise that in Agda (although we could try asking Claude just to see how it reacts ...).  I suppose we'll have to try something not too complicated and see if it performs reasonably.

kwxm (commented):
Looks good, although we should talk about the conformance tests since they're quite big.  I made some comments, but they're mostly just nitpicking.

plutus-metatheory/src/Builtin/Integer/Properties.lagda.md
ana-pantilie: Thanks, that's a good point. I'll try converting these property tests to proofs.

plutus-metatheory/src/Builtin/Integer/Base.lagda.md
ana-pantilie: There was actually a bug in the implementation which was uncovered when porting the Haskell property tests to Agda proofs. I also added a property test suite which checks that the Agda semantics of the division operators is equivalent to the Haskell semantics.

plutus-metatheory/src/Algorithmic/CEK.lagda.md
ana-pantilie: I think this is standard practice in Agda. For example, in decision procedures we usually name the negative proof term `¬<something>` like `¬p`.

plutus-metatheory/src/Builtin/CInteger.lagda.md
ana-pantilie: I opened an issue about this, thanks. https://github.com/IntersectMBO/plutus-private/issues/2381

plutus-core/untyped-plutus-core/testlib/Evaluation/Builtins/Integer/DivModProperties.hs
```diff
@@ -36,7 +36,7 @@ prop_div_mod_compatible (biginteger -> a) (NonZero (biginteger -> b)) =
   let t = addInteger (multiplyInteger b (divideInteger a b)) (modInteger a b)
    in evalOkEq t a
 
--- (k*b) `div` b = b and (k*b) `mod` b = 0 for all k
+-- (k*b) `div` b = k and (k*b) `mod` b = 0 for all k
```
kwxm, line 39: Well spotted!

plutus-metatheory/src/Builtin/Integer/Base.lagda.md
```diff
@@ -0,0 +1,106 @@
...
+rem n d = sign n ◃ (∣ n ∣ ℕ.% ∣ d ∣)
+```
+
+## Division and modulus
+
+The `div` and `mod` functions are based on Haskell's `div` and `mod`,
+via the same fixup as `integerDivMod#`. This implements floored division.
+
+```
+divModFixup : ℤ → ℤ → ℤ → ℤ × ℤ
```
kwxm: Would it be worth making this conditional on d ≠ 0?  This is only ever used in that case, but I assume that when d=0 here it's covered by the second two cases.

plutus-metatheory/src/Builtin/Integer/Properties.lagda.md
kwxm, line 1: The documentation is very helpful here.  I don't understand the details of the Agda, but the comments make it possible to kind of see what's going on.

plutus-metatheory/src/Builtin/CInteger.lagda.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+minBound : ℤ
+minBound = - ((+ 2) ^ 262143)
```
kwxm: 262143 is 2^18-1. It might be worth writing that explicitly here.  Admittedly the Haskell code doesn't do that ...

kwxm (approved):
This is great.  Now we just have to do the same for all of the cryptographic functions!

PR #7866, opened 2026-07-23
Use list casing in asData pattern synonyms

zliu41:
The `asData` destructors already use list casing, but the pattern synonyms don't, since `unsafeCaseList` is a CPS function that isn't compatible with ViewPatterns. But there's a quite simple solution: introduce an opaque, direct-style `directUnsafeCaseList`, and let the plugin rewrite it into `unsafeCaseList`. See Note [Use list casing in AsData pattern synonyms]. `wrapUnsafeUncons` and `wrapTail` are no longer needed.

This is almost always an improvement for V3, but not for V1/V2, since V1/V2 do not support builtin casing, so list casing is compiled to inefficient code using `head` and `tail`. We could re-introduce the existing codegen for V1/V2, but I think it's a low priority and I'd like to do that in a separate PR.

Also, list casing is easier to optimize using `dropList` (which basically entails a modified `DeadCase` pass), which should improve further.


SeungheonOh (approved):
It would be nice if we can change TH itself so that it doesn't use ViewPattern. Being said, the changes looks good as is

plutus-tx-plugin/test/Budget/9.12/toFromData.golden.uplc
```diff
@@ -1 +1,6 @@
-(program 1.1.0 (constr 1 [(constr 0 [(constr 0 [True, 1, False])])]))
\ No newline at end of file
+(program
+   1.1.0
+   (constr 1
+      [ (constr 0
+           [ (constr 0
+                [True, 1, (case 0 [(\ds -> False), (\ds -> True)] [])]) ]) ]))
```
zliu41, line 6: This is not really a regression - it requires 13 simplifier iterations to reduce to optimal code; unfortunately the default is 12 iterations.

PR #7867, opened 2026-07-25
Expand AsData budget tests for dropList

zliu41:
We should see improvements in these tests once `AsData` starts using `dropList` in the codegen.

### DIFF
--- a/plutus-tx-plugin/test/AsData/Budget/Spec.hs
+++ b/plutus-tx-plugin/test/AsData/Budget/Spec.hs
@@ -24,6 +24,9 @@ tests =
           "onlyUseFirstField-manual"
           onlyUseFirstFieldManual
           (onlyUseFirstFieldManual `unsafeApplyCode` inp)
+      , goldenBundle "richInts1" richInts1 (richInts1 `unsafeApplyCode` inpRich)
+      , goldenBundle "richInts2" richInts2 (richInts2 `unsafeApplyCode` inpRich)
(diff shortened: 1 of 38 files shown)


plutus-tx-plugin/test/AsData/Budget/Spec.hs
```diff
@@ -52,6 +63,27 @@ onlyUseFirstFieldManual =
...
+  plinthc
+    ( \d -> case PlutusTx.unsafeFromBuiltinData d of
+        RichInts {ri9 = x, ri14 = y} -> PlutusTx.addInteger x y
+    )
+
+richInts3 :: CompiledCode (PlutusTx.BuiltinData -> Integer)
+richInts3 =
+  plinthc
+    ( \d -> case PlutusTx.unsafeFromBuiltinData d of
+        RichInts {ri4 = x, ri8 = y, ri15 = z} -> PlutusTx.addInteger x (PlutusTx.addInteger y z)
```
SeungheonOh, line 84: How are these fields selected? seems rather arbitrary; would be nice to have rationale written somewhere in the comments.

SeungheonOh (approved):
fyi, we also have plutus-benchmark/lists which I think would also be a good place to add dropList related test cases. It uses much bigger lists.

zliu41:
Yeah, this is specifically for testing that AsData + CollapseCase + DeadCase generates optimal code for datatypes.

